PR #11, opened 2018-10-03
do_regtest: remove SVN refs, remove self-update functionality

dev-zero:
forgot one line where the `do_regtest` directly references the arch file


alazzaro (approved):

oschuett:
At this occation, could you also please update the Makefile:
https://github.com/cp2k/cp2k/blob/master/Makefile#L130
https://github.com/cp2k/cp2k/blob/master/Makefile#L191

PR #23, opened 2018-10-05
Arch file update + normal copy on AFS

mkrack:


oschuett:
Hi Matthias, you can submit your changes by hitting **Rebase and merge**.
However, note that your first commit has krack@merlin-l-02.psi.ch as author email.

mkrack:
Hi Ole, thanks for the note. Indeed, I forgot to update my .gitconfig on merlin. I tried to amend the author email with an interactive rebase session, but it is likely that I messed up something in that way as git rookie.

oschuett:
Yes, interactive rebase with `git commit --amend --author="..."` would have been my first approach too. Certainly, with git there is always more than one way to skin a cat ;-)

PR #34, opened 2018-10-09
S-ALMO patch (old compilers compatibility)

RubenStaub:
Fermi-Dirac smearing within the ALMO formalism, allowing an ALMO description of metals.
(contains also a quick fix for molecular guess with the diagonalisation algorithm, within ALMO)
(should be compatible with old compilers, with a more elegant solution proposed by @oschuett)


oschuett:
It looks like you did a `git pull` instead of starting fresh a branch. That creates a very complicated git history including merge commits, which we want to avoid.

I would suggest you do the following:

1. Make sure you have the upstream repository setup:
```
git remote add upstream https://github.com/cp2k/cp2k.git
git fetch upstream
```

2. Create a new feature branch:
```
git checkout -b new_branch_name upstream/master
```
3. Commit you changes.

4. Push them to your repo:
```
git push origin new_branch_name
```
5. Open another pull request for new_branch_name.

oschuett:
Was replaced by #35.

PR #35, opened 2018-10-09
S-ALMO bug fix

RubenStaub:
Fermi-Dirac smearing within the ALMO formalism, allowing an ALMO description of metals.
(contains also a quick fix for molecular guess with the diagonalisation algorithm, within ALMO)
(should be compatible with old compilers, with a more elegant solution proposed by @oschuett)

### DIFF
--- a/src/almo_scf_qs.F
+++ b/src/almo_scf_qs.F
@@ -707,34 +707,40 @@ CONTAINS
 !> \param energy_total ...
 !> \param mat_distr_aos ...
 !> \param smear ...
-!> \param kTS ...
+!> \param kTS_sum ...
 !> \par History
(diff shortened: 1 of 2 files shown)


RubenStaub:
I hope the patch is now compatible with old compilers...

Again, thanks for the support (I usually only pull and push with git), and sorry for the trouble. I hope the failed regtests did not cause too much inconvenience for the developers.

oschuett:
Looks good. The minimal test, which formerly failed, now passes too.

Don't worry about breaking stuff. It's inevitable. What's important is that you follow up on your changes in a timely manner - as you did splendidly :-)

Regarding git: Simply avoid `git pull` because it creates merge commits. Instead create a new feature branch for each PR as I outlined above.

PR #43, opened 2018-10-15
SIRIUS integration

mtaillefumier:
Hi,

This pull request contains the necessary modifications to the toolchain for compiling cp2k with the SIRIUS dependency. Note that SIRIUS is NOT activated by default since this preliminary work is missing quite a few functionalities to be fully functional. CP2K will still works with sirius but only the DFT groundstate and relaxation are possible at this stage.

The pull request also contains few test cases that will be extended in the future.

Best

Mathieu

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -2,329 +2,254 @@
 
 ##  1. Acquire the code:
 
-see  https://www.cp2k.org/download
+See  https://www.cp2k.org/download
 
 For users, the preferred method is to download a release.
 For developers, the preferred method is to download it from Git.
 
 ## 2. Install Prerequisites
 
-Sub-points here discuss prerequisites needed to build CP2K.
-Some of these can be conveniently installed by a script see: cp2k/tools/toolchain
-Copies of the recommended versions of 3rd party software can be
-downloaded from https://www.cp2k.org/static/downloads/ .
+Sub-points here discuss prerequisites needed to build CP2K. Most of these can be conveniently installed via the [toolchain script](./tools/toolchain). Copies of the recommended versions of 3rd party software can be downloaded from https://www.cp2k.org/static/downloads/.
 
 ### 2a. GNU make (required, build system)
 
-GNU make should be on your system (gmake or make on linux) and used for
-the build, go to https://www.gnu.org/software/make/make.html
-download from https://ftp.gnu.org/pub/gnu/make/
-also python (2.X) is required for building.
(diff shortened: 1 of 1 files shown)


tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -1071,7 +1151,7 @@ To use the installed tools and libraries and cp2k version
 compiled with it you will first need to execute at the prompt:
   source ${SETUPFILE}
 To build CP2K you should change directory:
-  cd cp2k/
```
oschuett: Actually, the `Makefile` was recently moved into the root directory.

tests/sirius/regtest-1/TEST_FILES
```diff
@@ -0,0 +1,14 @@
...
+# e.g. 0 means do not compare anything, running is enough
+#      1 compares the last total energy in the file
+#      for details see cp2k/tools/do_regtest
+#
+# Sirius will run in serial and parallel with or without cuda
+#
+# testing global tree creation using single exact potential, the deterministic random number generator
+#   with the standard move types, check i
+# fully test non colinear magnetism with spin orbit coupling
+Au.inp                                      85    1.0E-5           -147.171051899887289
```
oschuett: `1e-5` is a pretty low precision which we usually only use for tests that run MD or alike.

oschuett:
Hi Mathieu,

thanks you very much for this PR. It's great that SIRIUS gets some tests.
A few comments:
 - We try to keep a linear git history. Therefore, merge commits are not allowed. You can try to remove the offending commit with `git rebase -i`. Alternativly, you can start with a fresh branch and copy your files over.
 - It seems you did not include some files in `/tools/toolchain/scripts/`. This is probably because our `gitignore` files is a tad too general at the moment (see #44).
- Somewhat unrelated, but do you have any plans to make proper use of CP2K's input structure instead of the json sections?

tests/sirius/regtest-1/TEST_FILES
dev-zero: Do we really want to start testing Sirius for numerical stability?
I would be totally happy with a test which only proves that the interface works in general.

oschuett:
I realize that json is attractive from an implementation point of view. However, I really think this is a missed opportunity. Hence, I opened #45 to keep track and maybe continue the discussion in the open.

oschuett:
From my experience keeping a linear history is not that difficult. Really, one only has to remember to use `git rebase` instead of `git pull`.

Here are some background info:
- https://www.endoflineblog.com/gitflow-considered-harmful
- https://www.endoflineblog.com/oneflow-a-git-branching-model-and-workflow

Let me know if you have any questions. We are still in the process of writing up our own support docs.

PR #44, opened 2018-10-15
Simplify .gitignore

oschuett:
The current gitignore file is very general and ignores e.g. `[Ss]cripts` directories. This is an attempt to reduce the file to what we actually need on a daily basis.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@ obj/
 bin/
 regtesting/
 install/
+tools/toolchain/install/
+tools/toolchain/build/
 
 # ignore local arch files not meant to be shared
 arch/local*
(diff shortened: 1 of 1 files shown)


dev-zero:
@oschuett A large project's `.gitignore` should be a superset of what different developers need, not the smallest common subset. I therefore extended the `.gitignore` on purpose to a rather large set (which is generated, btw) and to also ensure that people who are tempted to use `git add -A` will not commit anything they shouldn't.
Since the `scripts` directory seems to be part of a `virtualenv` now, I would rather suggest that we rename our directory to something a bit more meaningful, like `build.scripts.d` or `package.scripts.d`. The first time I went looking there I expected some helper `scripts` to be in there, and was surprised to find the package build scripts instead.

oschuett:
As always, there is a trade-off. An overreaching gitignore file will lead to mistakes too, as #43 shows. Hence, I think we should start small and add rules on demand. I also think that `scripts` is a perfectly valid directory name which we should not surrender to virtualenv.

dev-zero (changes requested):
please do not edit the generated `.gitignore` part, do not remove the Python related part because there are people who are using `virtualenv` (also inside CP2K for testing scripts or bindings), please do not remove editor-specific ignores since we have people working with them, please do not remove ignores for extra Fortran module files since people may have extra files lying around they compile manually, the same goes for library and executable endings

oschuett:
I'm sorry, but the current gitignore contains a couple of rules which will simply not work for us:

```
*.pot
*.pdb 
[Ss]cripts
[Ii]nclude 
dist/
*.mod*
```

Have a look:
```
find .  | git check-ignore --no-index -v --stdin
```

Also, I don't think we have to catch every conceivable mishap. In general people rarely commit files unintentionally. So, for me gitignore is mostly about filtering noise from `git status`.

dev-zero:
Interesting, it seems that we failed with our own basic definition as well:

`regtesting/` also matches `tools/regtesting/`

See #46 for the fix (plus some more fixes) and the controlled whitelisting of specific files and directories.

oschuett:
While #46 is no exactly the simplification I was looking for, it does solve my original problem ;-)

PR #49, opened 2018-10-17
Fix toolchain scripts

schulkov:
* replace $nprocs with $NPROCS in install_openblas.sh;
* enable C++ MPI binding in install_openmpi.sh.

C++ binding is only needed to compile CP2K with PLUMED. However linking flags "-lmpi_cxx -lstdc++" appear in generated local arch files even if we do not use PLUMED at all.

### DIFF
--- a/tools/toolchain/scripts/install_openblas.sh
+++ b/tools/toolchain/scripts/install_openblas.sh
@@ -64,7 +64,7 @@ case "$with_openblas" in
                  install > install.serial.log 2>&1
             if [ $ENABLE_OMP = "__TRUE__" ] ; then
                make clean > clean.log 2>&1
-               make -j $nprocs \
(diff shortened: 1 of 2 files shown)


dev-zero:
if you want to build explicitly in parallel with `-j`, you should set `MAKE_NB_JOBS=0` as well to turn off OpenBLAS' build parallelization.

schulkov:
You are probably correct. The original line in 'install_openblas.sh'

make -j $nprocs USE_THREAD=1 [...]

still implies external building. Moreover, due to undefined bash variable "$nprocs", the script actually calls the command:

make -j USE_THREAD=1 [...]

which instructs 'make' not to limit the number of jobs that it spawns.

dev-zero:
merged, will fix the mentioned issue and #50 in #46

PR #60, opened 2018-10-23
Fixed (potential) OOBs unveiled by link-time code-optimization (LTO).

hfp:
The issues were listed with `-flto -Wl,-flto` (GNU Fortran). The flag `-Wno-lto-type-mismatch` perhaps suppresses the issues (although this is unexpected as the issues is not a type-mismatch).


dev-zero:
Agreed for `pot%pseudo_type` and `pot%relativistic`, but the other cases could never have happened given the sizes of the buffers copying from and to, but if we can play it safe, then do it.
But, isn't the behavior of `READ` to truncate to the target string/buffer size? If that's the case I would rather prefer to have an error right there than getting weird errors if that truncated string is used in other places. What do you think?

hfp:
Ok, I can see your point. Let me only pick the cases you recommend and leave the other lines as-is.

dev-zero:
@hfp sorry, that's not what I meant: for consistency and safety I think your approach is better. Definitions can change after all, at which point those places may start to generate real errors. The only thing I dislike is the possible silent truncation of strings leading to hard to track down errors on the user side.

hfp:
What is the preference: CPABORT or CPASSERT. I mean given that there was no error so far, CPASSERT would do it in particular since the user cannot do anything about it (perhaps increase `default_string_length`). Of course, CPASSERT would involve default_string_length into the (printed) expression.

hfp:
It appears that I could hit the button to take the PR (I noticed this also earlier). Is this generally the case (with PRs sucessfully passing the necessary tests), or because I am member of the CP2K organization? What is the preference: "Squash and merge" or "Rebase and merge"?

dev-zero:
because you are a member of the CP2K organization. Welcome! ;-)

It depends whether or not you want to retain your commits in the final history or not. In this case I would do `Squash and merge` since the second commit is a fix for the first one.

PR #63, opened 2018-10-30
Allow CPASSERT to be elided if NDEBUG is defined. Introduced CP_VERSIONx.

hfp:
Allow CPASSERT to be elided if NDEBUG is defined. Introduced CP_VERSION3 and CP_VERSION4 to consistently calculate a single (version-)number such that it can be compared using the preprocessor e.g., #if CP_VERSION3(1, 2, 3) <= CP_VERSION3(MYLIB_MAJOR, MYLIB_MINOR, MYLIB_UPDATE).

This change may allow to setup a test with NDEBUG defined aka "release build" (not meant to spawn a discussion of whether or not an actual production-build should contain assertions). Anyhow, a test with NDEBUG can unveil potential side-effects or misuses of CPASSERT. The current status (to my knowledge) is that regression tests fail to a large portion if CPASSERT is elided.

**Note**: `make pretty` appears to fail parsing this change.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -19,7 +19,7 @@ GNU make should be on your system (gmake or make on linux) and used for the buil
 Python 2.x is needed to run the dependency generator. On most system Python is already installed. For more information visit: https://www.python.org/
 
 ### 2c. Fortran and C Compiler (required, build system)
-A Fortran 2003 compiler and matching C compiler should be installed on your system. We have good experience with gcc/gfortran (gcc >=4.6 works, later version recommended). Be aware that some compilers have bugs that might cause them to fail (internal compiler errors, segfaults) or, worse, yield a mis-compiled CP2K. Report bugs to compiler vendors; they (and we) have an interest in fixing them. Always run a `make -j test` (See point 5.) after compilation to identifiy these problems.
+A Fortran 2003 compiler and matching C compiler should be installed on your system. We have good experience with gcc/gfortran (gcc >=4.6 works, later version recommended). Be aware that some compilers have bugs that might cause them to fail (internal compiler errors, segfaults) or, worse, yield a mis-compiled CP2K. Report bugs to compiler vendors; they (and we) have an interest in fixing them. Always run a `make -j test` (See point 5.) after compilation to identify these problems.
 
 ### 2d. BLAS and LAPACK (required, base functionality)
 BLAS and LAPACK should be installed.  Using vendor-provided libraries can make a very significant difference (up to 100%, e.g., ACML, MKL, ESSL), not all optimized libraries are bug free. Use the latest versions available, use the interfaces matching your compiler, and download all patches!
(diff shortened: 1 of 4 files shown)


dev-zero:
any concrete users for the `CP_VERSION{3,4}` macros?

hfp:
CP_VERSION3 can be used for the GNU Compiler Collection (and other codes) e.g., CP_VERSION3(__GNUC__, __GNUC_MINOR__, __GNUC_PATCHLEVEL__). This also works GNU Fortran (despite the fact that it is called "GNUC"):

```bash
gfortran -dM -E - < /dev/null > symbols.txt
grep GNUC symbols.txt
```
The CP_VERSION4 can be used for LIBXSMM (version consists of major, minor, update, and patch). I may have patches in the future that then allow to maintain compatibility with older LIBXSMM.

hfp:
Well, CP_VERSION3 would probably suffice with LIBXSMM as well (unless one intends to conditionally take unreleased changes). I can rework to a single CPVERSION macro if that is more suitable with respect to the macro-name.

dev-zero:
> CP_VERSION3 can be used for the GNU Compiler Collection (and other codes) e.g., CP_VERSION3(**GNUC**, **GNUC_MINOR**, **GNUC_PATCHLEVEL**). This also works GNU Fortran (despite the fact that it is called "GNUC"):
> 
> ```shell
> gfortran -dM -E - < /dev/null > symbols.txt
> grep GNUC symbols.txt
> ```
> 
> The CP_VERSION4 can be used for LIBXSMM (version consists of major, minor, update, and patch). I may have patches in the future that then allow to maintain compatibility with older LIBXSMM.

nice!

The only thing I wondered is whether it would be clearer if the range for major/minor/update is the same for both versions since there are projects which leave away the patch version number on the first release (starting with a `1.2.3` and then making a patch `1.2.3.1.` which would compare differently if I would do `CP_VERSION3(1.2.3) > CP_VERSION4(1,2,3,1)`.
Otherwise a line of documentation that one should never compare a `CP_VERSION3 to a CP_VERSION4` would be sufficient.

hfp:
Indeed CP_ASSERT{3|4} were not compatible since the ranges were tuned to not overflow even with 32-bit integer. Anyhow, let's not worry -- I consolidated into a single CPVERSION. If there is any need ever to check four components, we can add CPASSERT4 by then.

dev-zero:
That's fine too. I guess bit-shifts are not allowed for a preprocessor in compat mode?

hfp:
No idea, but this code works for GNU Fortran and IFORT and the current expression is in use since years (in my fork of CP2K which was previously the "intel" branch of CP2K's Git-mirror).

hfp:
The remaining failure with `sdbg` is related to spglib or specifically this library is missed on the link-line. Perhaps this PR may wait for other fixes (see dashboard failures).

Issue #66, opened 2018-10-31
[toolchain] default configuration broken


dev-zero:
the addition of SIRIUS to the toolchain broke the "default" configuration (running the toolchain without any arguments):

```console
$ ./install_cp2k_toolchain.sh
WARNING: (./install_cp2k_toolchain.sh, line 309) No MPI installation detected on you system. Ignore this message if you are using Cray Linux Environment
nvcc not found, disabling CUDA by default
Not using MPI, so scalapack is disabled.
Not using MPI, so PEXSI is disabled.
ERROR: (./install_cp2k_toolchain.sh) For SIRIUS to work you need a working gsl library use --with-gsl option to specify if you wish to install the library or specify its location.
```

CC @mtaillefumier

hfp:
Also, farming tests for Current Toolchain (psmp) currently [throw an exception](https://storage.googleapis.com/cp2k-reports/test_psmp_report.txt) due to initializing SIRIUS multiple times ("SIRIUS library is already initialized").

mtaillefumier:
I have the patch for the first issue which I think is more important right now. Do you want a pull request or can I send you a patch for this. 

I do not understand how sirius can be initialized more than one's. the only time it is called is in cp2k_run.F which call a wrapper cp_sirius_init() which call sirius_initialize() function.

alazzaro:
I think farming in CP2K (but I'm not sure) calls the init part twice for the master rank, which behaves as master and worker.
The current solution is to finalize the library before going to farming, see

https://github.com/cp2k/cp2k/blob/9a24425f33e985c94adbbb41cb6414874f03d376/src/start/cp2k_runs.F#L261

Therefore I would add there a sirius finalization...

mtaillefumier:
This is indeed a possibility but the real fix would be to call sirius_initialize only when needed. As far as i understand it is not possible because the input preprocessing is done after the initialization of the full library. I do not know however how much work it would be to have this information available. 

I will make the necessary modifications to avoid the double initialization.

mtaillefumier:
Ok I included the modifications ion my pull request.

hfp:
As a side-note, spglib somehow became necessary for at least the sdbg-test (CI test that runs by default). However, spglib's install script added the library only conditionally to the link-line (IF_MPI and IF_OMP). I fixed as part of PR #63 (early version of PR #63 failed due to missing spglib symbols).

dev-zero:
[...]

PR #67, opened 2018-10-31
SIRIUS is not activated by default and is now compiled with and without GPU support if cuda is found

mtaillefumier:
- Fix the build issue [[toolchain] default configuration broken #66](https://github.com/cp2k/cp2k/issues/66)
- Fix the compilation when cuda is found. we generate two different versions of the library with and without cuda and give the right parameters to the toolchain.


dev-zero:
looks good, thanks for the quick fix!

mtaillefumier:
no worries. I still think the double initialization could be handled differently but I do not fully understand how the farming mode really works right now.

PR #75, opened 2018-11-02
Fix building ELPA with OpenBLAS

dev-zero:
The facilities in commit 57e0cd7 can be used everywhere where a build system does a check on the linker flags and uses BLAS (and therefore potentially OpenBLAS).


dev-zero:
@hfp fyi, in case you are looking at similar build errors using the toolchain

oschuett:
Yes, it seems like my rsync is missing the `--executability` flag. I'll take care.

dev-zero:
@oschuett <s>sorry, I don't understand, where/how exactly?</s> nevermind, was looking at an old version of the test report

PR #79, opened 2018-11-04
Add a missing cp_sirius_init() after farming is done.

mtaillefumier:
I miss a sirius init call after the farming is done leading to an abort of cp2k in farming mode.

### DIFF
--- a/src/sirius_interface.F
+++ b/src/sirius_interface.F
@@ -82,7 +82,7 @@ CONTAINS
 
       CHARACTER(len=*), PARAMETER :: routineN = 'cp_sirius_init', routineP = moduleN//':'//routineN
 
-      CALL sirius_initialize(bool(.TRUE.))
+      CALL sirius_initialize(bool(.FALSE.))
 
    END SUBROUTINE cp_sirius_init
 #else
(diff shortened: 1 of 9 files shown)


oschuett:
I think, the number in the first line of [TEST_TYPES](https://github.com/cp2k/cp2k/blob/master/tests/TEST_TYPES#L1) should be `85`.

oschuett:
Looks like your PR contains a merge commit. We want to keep the history linear. So, please always use `git rebase` instead of `git pull`. Here are some background info:

- https://www.endoflineblog.com/gitflow-considered-harmful
- https://www.endoflineblog.com/oneflow-a-git-branching-model-and-workflow

Let me know if you have any questions. We're still in the process of writing up our own support docs.

oschuett:
Look, I'm only trying to be helpful. I know that CP2K's mixture of legacy quirks and tight testing can be frustrating at times.

Since you now closed this PR, are you still planning to follow up on the open issues?

mtaillefumier:
Yes, I will follow up on this. I will prepare a pull request with more than just one thing though.

PR #82, opened 2018-11-05
Toolchain: Fix parse_if.py --selftest

oschuett:


dev-zero:
Does it really make sense to add a non-implemented flag which probably will never be implemented?
Over at [dbcsr](https://github.com/cp2k/dbcsr/commit/f5c8805e0dfc88964fa38301788eae09c1ff5831) I just removed such a flag to replace it with proper unittests instead.

oschuett:
I agree that the `--selftest` flag did not live up to expectation. Nevertheless, I'd like to merge this change now to clear the dashboard. I opened #83 to discuss how we can improve our Python testing in the future.

oschuett:
I disagree. The main purpose of the Python CI test is to check different Python versions. And I think we'll have to support a range of Python version for the foreseeable future (even after 2.x is gone).

PR #86, opened 2018-11-06
DBCSR integration

### LINKED ISSUE #52: DBCSR-Git Integration
Version 1.0 of the DBCSR is ready for the inclusion in DBCSR. Let's start a discussion on how we can achieve that (submodule, copy, ...)

pseewald:
Integration of DBCSR master branch into CP2K, see also cp2k/dbcsr#86.

Open discussion: Method to include DBCSR into CP2K: copy vs. git submodule (see #52) vs. fully external. Right now I use git submodule but I can change this as soon as we've arrived at a conclusion.

This pull request is based on https://github.com/alazzaro/cp2k/tree/master/cp2k

This resolves #52

Building CP2K with DBCSR requires cp2k/dbcsr#87 to be merged into dbcsr master.


alazzaro:
I think the CI doesn't work with submodules.
My assumption is that it requires a

git submodule update --init --recursive

to download the submodules. At least, the output seems to report that DBCSR submodule was not downloaded.

@oschuett @dev-zero  is this the case?

oschuett:
Currently, the CI does the following:
```
git fetch origin pull/${PR_NUMBER}/head 
git checkout ${HEAD_SHA}
```

oschuett:
Ok, I added `git submodule update --init --recursive`. Now it seems to stumble over some other Makefile issue.

alazzaro:
Yes, Patrick fixed that problem with https://github.com/cp2k/dbcsr/pull/87

pseewald:
I changed to develop branch of DBCSR such that we can test with cp2k/dbcsr#87 (and possibly other fixes).

pseewald:
I don't know what goes wrong with the CI, locally on my system all tests pass.

Furthermore there seems to be an issue with libxsmm, I can only compile without -D__LIBXSMM.

hfp:
Hi Patrick,

how to reproduce the issue with libxsmm? I can pull your PR locally, and try to reproduce.

Hans

pseewald:
Yes, you pull this branch by:

`git clone git@github.com:pseewald/cp2k.git -b dbcsr-include --recursive
`

I'm using libxsmm-1.7.1.

hfp:
Ok, even without your PR -- it fails because of libxsmm. The problem is the CMake depends on a .pc-file (package-config), which was only added recently. In fact, no release version comes with support for pkg-config. I think, a check is needed for the CMake stuff to conditionally enable LIBXSMM (only if a package config exists).

hfp:
Quick update:

* Newer GCC (8.2.0 in my case) do not accept LIBXSMM's F-interface without compilation error if LIBXSMM is older than 1.8.2. We did not change integration of CP2K/DBCSR and LIBXSMM since CP2K 3.0. So the issue is really with newer GCC and old LIBXSMM (older than 1.8.2).
* CMake's tests for pkg-config support seems sufficient, it just bails out if `-DUSE_SMM=libxsmm` is used and if no `.pc` files are found in LIBXSMM's lib folder. However, I guess this PR is not relying on CMake for that part, therefore it might be the first point (new compiler + old libxsmm).

hfp:
Hmm, even GCC 4.8.5 (GNU Fortran) does not compile the LIBXSMM-integration if LIBXSMM is older than v1.8.2. So I guess to resolve the issue with LIBXSMM in this PR requires to use at least v1.8.2.

dev-zero:
@hfp CMake is not being used when building DBCSR from within CP2K. For CMake one has to specify `-DUSE_SMM=libxsmm` to actually use libxsmm and it should then fail if libxsmm can't be found.

pseewald:
@hfp If I try to compile cp2k master with libxsmm, everything works fine. So I guess it is rather related to old libsmm and new DBCSR.

In fact, if I try to compile DBCSR develop with libxsmm 1.7.1 I get the same error.

hfp:
Yes, the issue is purely related to libxsmm older than v1.8.2.

pseewald:
So then we should document that DBCSR requires libxsmm v1.8.2 or newer?

hfp:
I guess that is fair to say v1.8.2 or newer. However, the integration between CP2K and LIBXSMM did not change since CP2K 3.0 (initial version) and older versions of LIBXSMM worked for sure. I will check the version-range further (and see if only a specifc range is broken and why), and I will send a PR later this week with updated version requirement.

alazzaro:
I think the libxsmm is something not related to the CI problem. Please open another issue for that.
CI is using libxsmm 1.9.0, so we are OK. 
As far as I can see, the problem is that CP2K is not properly compiling. 
@oschuett Can I have access to the entire log?

I see that unittest have some problems. Is this line OK?

>>> /opt/cp2k-master/regtesting/local/sdbg/TEST-local-sdbg-2018-11-07_07-37-29/UNIT/*_unittest
    *_unittest                                                                 -         RUNTIME FAIL

oschuett:
Unfortunately, it's not so easy at the moment to extract more info from CI runs. However, I prepared a major refactoring of the CI. I was going to hold off, but I think I'll just commit it now. Afterwards, the CI tests can be run locally.

alazzaro:
So the problem is due to the fact that the do_regtest calls make without any target and it gets another level of nested make calls (4 levels). It turns out that in this case the first (default) target is dbcsr, since we have this include first in the targets order:

include $(EXTSHOME)/Makefile.inc 

while we should have the default target as the first one:

default_target: all

So, this is clearly my fault (I didn't test all target). 
The solution is to invert this order of target, i.e. put 

default_target: all

at the very beginning.
This is the patch:

```
diff --git a/Makefile b/Makefile
index 330b3bcc0..4bcd7a85d 100644
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,9 @@ EXTSDIR      := exts
 EXTSHOME     := $(CP2KHOME)/$(EXTSDIR)
 EXTSPACKAGES := $(shell cd $(EXTSHOME) ; find * -maxdepth 0 -type d )
 
+# Common Targets ============================================================
+default_target: all
+
 # Discover programs =========================================================
 ifeq ($(ALL_EXE_FILES),)
 export ALL_EXE_FILES := $(sort $(shell $(TOOLSRC)/build_utils/discover_programs.py $(SRCDIR)))
@@ -95,9 +98,6 @@ ALL_OBJECTS        = $(addsuffix .o, $(basename $(notdir $(OBJ_SRC_FILES))))
 ALL_EXE_OBJECTS    = $(addsuffix .o, $(EXE_NAMES))
 ALL_NONEXE_OBJECTS = $(filter-out $(ALL_EXE_OBJECTS), $(ALL_OBJECTS))
 
-# Common Targets ============================================================
-default_target: all
-
 # stage 1: Call make recursively with each element in $(VERSION) as a target,
 #          The actual target is stored in ORIG_TARGET.
 #          Make can then parallelize over multiple versions targets.

```

Alternatively, we can move the include after the default one...

@oschuett @pseewald @dev-zero Any opinion on that?

pseewald:
@alazzaro Thank you for figuring this out. I think having the default target on top is the best solution, can you commit the fix to this pull request?

pseewald:
@alazzaro the PACKAGE files are not the problem, I can successfully run `test_conventions.sh` locally so it is a problem with the CI. I guess the problem is `--delete` option missing for rsync. I think this is solved by #87, I will update this PR

oschuett:
I'm working on it, but I'm afraid it'll take a few hours.

pseewald:
Sorry, I was not careful and I did not pull @alazzaro`s commit before rebasing and force pushing, thereby I deleted his commit. I recovered that change in 1620e387fde21430e5d1104aad30548508e18973

alazzaro:
Now I'm confused. The CI is still failing. Is this because of the rsync problem mentioned by Patrick? If so, let's wait for Ole fix...

alazzaro:
OK, now sdbg and pdbg run fine. I see a problem with the conventions checker, looks like it doesn't use the proper makefile. @oschuett Could you check it?

oschuett:
The conventions checker seems to have problems when compiling with [this arch-file](https://github.com/cp2k/cp2k/blob/master/arch/Linux-x86-64-gfortran.dumpast). Do you see anything suspicions?

alazzaro:
Well, it compiles locally. Assuming that this is a problem of linking (DBCSR is missing), I would assume that a clean of some sort should trigger the right compilation...

oschuett:
Ok, my fix is in. @pseewald could you once rebase your PR onto the latest cp2k/master branch.

pseewald:
I rebased but there seems to be a different error now.

oschuett:
Dang! Now the conventions checker is confused by removed directories. I'll try to fix it tonight.

oschuett:
@pseewald could you please once more rebase your PR onto the latest cp2k/master branch.

alazzaro:
Great, now the conventions are OK. I have pushed the other tests to run...

alazzaro:
@oschuett OK, the Python CI test is still using some old directories... Could you fix it?

oschuett:
Ok, the Python test went through as well.

alazzaro:
@pseewald 
I put here some of the changes to do before merging:

1. Changes to do in the arch files for the GPU available in CP2K. Now we have to add something like (or variants):

  CXX         = CC
  CXXFLAGS    = -O3 -I${CUDA_PATH}/include -std=c++11
  GPUVER      = K20X
  NVFLAGS     = -g -O3 -Xcompiler='-fopenmp' --std=c++11 $(DFLAGS)
  LIBS       += -lstdc++ -lcudart -lnvrtc -lcuda

in the following files:

[01:02:15] alazzaro@daint102:/store/uzh/uzh1/CP2K_AUTO_REGTEST/DAINT/CRAY_XC50-gfortran_gpu_git/psmp/cp2k/arch > grep NVCC *
CRAY-XC30-gfortran-cuda.psmp:NVCC     = nvcc -D__GNUC_MINOR__=8
CRAY-XK7-gfortran-cuda.psmp:NVCC     = nvcc
Linux-x86-64-cuda-ma.psmp:NVCC     = nvcc
Linux-x86-64-cuda.sopt:NVCC	 = nvcc
Linux-x86-64-cudadp.sopt:NVCC	 = /usr/local/cuda/bin/nvcc
Linux-x86-64-dbcsr-cuda.popt:NVCC     = nvcc
Linux-x86-64-gfortran.warn:NVCC     = nvcc -D__GNUC_MINOR__=6

2. Update the file:

https://github.com/pseewald/cp2k/blob/dbcsr-include/exts/README.md

I think we should write what will be the procedure if people want to update DBCSR from CP2K.

exts/Makefile.inc
```diff
@@ -0,0 +1,10 @@
+
+LIBS     := -L$(LIBEXTSDIR)/dbcsr -ldbcsr $(LIBS)
+EXTSDEPS  = $(LIBEXTSDIR)/dbcsr/libdbcsr.a
```
alazzaro: @pseewald After the discussion with @oschuett, likely here we can relax this dependency by using the dbcsr_api.mod ($(OBJEXTSDIR)/dbcsr_api.mod). Could you try that?

pseewald: Yes but I think we need two things:
- executables should depend on libdbcsr.a to make sure that they are linked against the changed library
- objects should depend on dbcsr_api.mod

Do you agree?

alazzaro: I think it is too much work... Assuming that you change .mod, then I would argue that .a gets recompiled too. In any case, CP2K will get recompiled, so you get it to link again. Therefore, checking for a change in dbcsr_api.mod is enough.

alazzaro:
@pseewald Another thing that we have to update is the toolchain, I mean the arch file generated by the toolchain to run CUDA

exts/Makefile.inc
alazzaro: As pointed by Patrick, we need two dependencies (compilation by dbcsr_api*.mod and linking by libdbcsr.a). Let's postpone it after the merging.

pseewald: We have several DBCSR *.mod files used in cp2k, not only dbcsr_api:
- dbcsr_vector
- dbcsr_tensor_api
- dbcsr_acc_device

I think we should not hardcode dependencies on mod files then.

pseewald: This has been taken care of with db2c1ccfdca9c2a83d97e946e05d89b05058bf5a
@alazzaro maybe you want to take a quick look and review my changes. I removed copying of mod files from $(OBJEXTSDIR) to $(OBJDIR) since this always updates the timestamp.

alazzaro: Thanks Patrick, it makes sense. There is something wrong with the CI (as usual)...

oschuett: The CI is optimized for quick results in the common case. This PR is really the opposite of common. So, please have some mercy ... 

I guess any change to the Makefile should trigger a `make distclean`. I'll look into it.

alazzaro: @pseewald I assume that you want your commit 2398b7141b4a96587dcee72ab3e5adcf396f1cb1 into the master release of DBCSR... That's means we have to make a new RC (@dev-zero )

oschuett: Ok, checks are passing again.

pseewald: > @pseewald I assume that you want your commit 2398b7141b4a96587dcee72ab3e5adcf396f1cb1 into the master release of DBCSR... That's means we have to make a new RC (@dev-zero )

Done

pseewald:
This is now ready to be merged. Should we wait for a proper DBCSR release?

alazzaro:
Please, proceed to merge, but don't delete the branch. 
It is a bit complicated procedure:
1) we merge CP2K with a DBCSR release candidate
2) if all tests on the dashboard are fine, then we make an official release for DBCSR
3) we update CP2K with the official release

PR #87, opened 2018-11-07
Docker: Major refactoring

oschuett:


tools/docker/scripts/ci_entrypoint.sh
```diff
@@ -0,0 +1,100 @@
...
+          --exclude=/tools/toolchain/         \
+          --executability                     \
+          --ignore-times                      \
+          --update                            \
+          --verbose                           \
+          --recursive                         \
+          --checksum                          \
+          /mnt/cp2k/  /workspace/cp2k/
+else
+    echo "Neighter GIT_REF nor /mnt/cp2k found - aborting."
```
dev-zero: `neighter` -> `neither`

tools/docker/scripts/ci_entrypoint.sh
```diff
@@ -0,0 +1,100 @@
...
+    git checkout "${GIT_REF}"
+    git submodule update --init --recursive
+    git --no-pager log -1 --pretty='%nCommitSHA: %H%nCommitTime: %ci%nCommitAuthor: %an%nCommitSubject: %s%n' | tee -a $REPORT
+
+elif [ -d  /mnt/cp2k ]; then
+    echo -e "\n========== Copying Changed Files =========="
+    #TODO: Exclusion rules could also come from .gitigore
+    rsync --exclude="*~"                      \
+          --exclude=".*/"                     \
+          --exclude="*.pyc"                   \
```
dev-zero: `--exclude="*.py[cod]" --exclude="__pycache__"`

tools/docker/scripts/ci_entrypoint.sh
```diff
@@ -0,0 +1,100 @@
...
+    cd /workspace/cp2k
+    git fetch origin "${GIT_BRANCH}"
+    git checkout "${GIT_REF}"
+    git submodule update --init --recursive
+    git --no-pager log -1 --pretty='%nCommitSHA: %H%nCommitTime: %ci%nCommitAuthor: %an%nCommitSubject: %s%n' | tee -a $REPORT
+
+elif [ -d  /mnt/cp2k ]; then
+    echo -e "\n========== Copying Changed Files =========="
+    #TODO: Exclusion rules could also come from .gitigore
+    rsync --exclude="*~"                      \
```
dev-zero: you might want to either `rm -rf /workspace/cp2k/` or use `--delete` to make sure to clear removed files, especially with a build system which picks up files automatically

tools/docker/scripts/ci_entrypoint.sh
```diff
@@ -0,0 +1,100 @@
...
+          --exclude="*.pyc"       \
+          --exclude="/build"      \
+          --exclude="/install"    \
+          --executability         \
+          --ignore-times          \
+          --update                \
+          --verbose               \
+          --recursive             \
+          --checksum              \
+          /workspace/cp2k/tools/toolchain/  /opt/cp2k-toolchain/
```
dev-zero: dito. You might want to factor out common rsync arguments or the rsync call itself

tools/docker/scripts/ci_entrypoint.sh
```diff
@@ -0,0 +1,100 @@
...
+fi
+
+echo -e "\n========== Running Test =========="
+cd /workspace
+"$@" 2>&1 | tee -a $REPORT &  # Launch in the background.
+TEST_PID=$!
+trap '[ -z $TEST_PID] || kill $TEST_PID' SIGHUP SIGINT SIGQUIT SIGTERM
+
+# Upload preliminary report every 30s while test is running.
+while [ -e /proc/${TEST_PID} ]; do
```
dev-zero: why not use the `jobs` builtin instead? Would be more portable.

tools/docker/scripts/test_regtest.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
+
+# author: Ole Schuett
+
+ARCH=$1
+VERSION=$2
+
```
dev-zero, line 12: if you are expecting exactly two arguments, assert for that:

```bash
[ $# -eq 2 ] || (echo "Usage: ..." && exit 1)
```

tools/docker/scripts/test_python.sh
```diff
@@ -1,62 +1,54 @@
...
 # find executable python scripts
-ALL_SCRIPTS=`find ./src/ ./tools/ -name "*.py"  -executable`
-ESSENTIAL_SCRIPTS=`find ./tools/build_utils -name "*.py"  -executable`
+ALL_SCRIPTS=$(find ./src/ ./tools/ -name "*.py"  -executable)
+ESSENTIAL_SCRIPTS=$(find ./tools/build_utils -name "*.py"  -executable)
 ESSENTIAL_SCRIPTS="$ESSENTIAL_SCRIPTS ./src/dbcsr/libsmm_acc/libcusmm/generate.py"
 
 # python 2.6
-run_selftests python2.6 $ESSENTIAL_SCRIPTS
+run_selftests python2.6 "${ESSENTIAL_SCRIPTS}"
```
dev-zero, line 42: Please drop 2.6 testing: Python 2.6 is EOL and popular projects have even started dropping Python 2.7 support

tools/docker/scripts/test_coverage.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash -e
+
+# author: Ole Schuett
+
+ARCH=local_coverage
+VERSION=$1
+
```
dev-zero, line 12: please check number of arguments passed, see comment above

tools/docker/scripts/ci_entrypoint.sh
oschuett: Good catch - thanks!

tools/docker/scripts/test_python.sh
oschuett: It probably comes at no surprise to you that I'm reluctant to drop Python 2.6 support ;-)

However, we should really find a better way to settle these kind of arguments. So, I propose we simply ask our users (#90).

tools/docker/scripts/ci_entrypoint.sh
oschuett: The best I could come up with is `while jobs -p | grep "${TEST_PID}" > /dev/null; do`, which I find less readable. I don't think portability matters that much as we're using Docker.

dev-zero: AFAIK Docker on Mac needs special flags to get a `/proc` in a container

dev-zero: plus you might be checking any PID (after PID reuse) and not necessarily the child PID, while `jobs` should give the PID only if it was spawned from this shell.

oschuett: You're absolutely right. I was actually checking the PID of the `tee` process. I just pushed a new and improved version.

dev-zero:
Unless you plan to squash your commits, please read https://github.com/cp2k/cp2k/wiki/Contribution-Guidelines#commit-messages

oschuett:
Don't worry. I'll squash them. Sorry for the noise.

Issue #71, opened 2018-11-01
Should SIRIUS be activated by default


mtaillefumier:
During a meeting with the others developers of sirius, it was suggested that sirius should be activated by default in cp2k (it is not at the moment because of some others issues hopefully fixed now). SIRIUS integration to cp2k is already in good shape despite the few glitches with the build system. We have few test cases for regression and basic functionalities up and running. The build has passed basic CI tests (I do not the psmp tests activated) but I think it would be a good opportunity to catch others potential problems (like the farming issue) in a more systematic way if SIRIUS is built by default. 

I already have a patch ready for it but I wanted to have your view on this before making a pull request. I know it goes against the changes I made yesterday, but most of the changes where needed anyway.

oschuett:
Addressed via #88.

PR #88, opened 2018-11-07
Bugs related to SIRIUS interface

mtaillefumier:
this pull request fixes a few bugs encountered during the farming and mpi tests. This pull request fixes the following items
- Fix a double initialization of mpi
- Fix a crash of cp2k when the farming tests are done. The crash was due to the fact that sirius was not reinitialized after going out the farming tests
- Activate Sirius by default.

### DIFF
--- a/src/sirius_interface.F
+++ b/src/sirius_interface.F
@@ -82,7 +82,7 @@ CONTAINS
 
       CHARACTER(len=*), PARAMETER :: routineN = 'cp_sirius_init', routineP = moduleN//':'//routineN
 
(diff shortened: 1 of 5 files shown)


dev-zero:
@mtaillefumier @oschuett you can also please take care of the respective issues? -> #71

mtaillefumier:
Actually i temporarily activated sirius in this pull request. I would simply close the issue or put it on standby for now. I do not need to have automatic testing with this part of the code since it is in active development.

PR #98, opened 2018-11-15
GTH pp interface to Sirius

juerghutter:


src/sirius_interface.F
```diff
@@ -224,11 +238,152 @@ SUBROUTINE cp_sirius_create_env(pwdft_env)
...
+         NULLIFY (upf_pot, gth_potential)
+         CALL get_qs_kind(qs_kind_set(ikind), upf_potential=upf_pot, gth_potential=gth_potential)
+         IF (ASSOCIATED(upf_pot)) THEN
+            CALL sirius_add_atom_type(sctx, string(label), fname=string(upf_pot%filename), &
+                                      mass=REAL(mass, KIND=C_DOUBLE))
+         ELSEIF (ASSOCIATED(gth_potential)) THEN
+            !
+            NULLIFY (atom_grid)
+            CALL allocate_grid_atom(atom_grid)
+            nmesh = 929
```
dev-zero, line 252: accidentally hardcoded number of grid points?

juerghutter: No. We didn't wanted to make this an input option, just use a hardcoded default.

oschuett:
I see that you upper-cased the regtest directory in to `/tests/SIRIUS/` (good!). I guess, then `TEST_DIRS`  should be updated and the old lower-case dir removed.

oschuett:
I think `tests/TEST_DIRS` still needs to be updated.

PR #102, opened 2018-11-20
Install LIBXSMM 1.10

hfp:
Updated installer script to LIBXSMM 1.10, which brings at least the following "Zero-mask unused register parts to avoid false positives with enabled FPEs (MM kernels)" (see https://github.com/hfp/libxsmm/releases/tag/1.10). The aforementioned fix was an explicit finding from CP2K's regtests running on Skylake server.


oschuett:
Hi Hans, I just uploaded the tgz to  https://www.cp2k.org/static/downloads/libxsmm-1.10.0.tar.gz. Could you also add it to [checksums.sha256](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/checksums.sha256).

hfp:
Thank you!

oschuett:
It's confused because I renamed your tgz from `1.10` to `1.10.0`, but keep the contained directory name. I now repackaged the archive, but now it's sha does not match your upstream archive anymore. So maybe we should stick to `1.10`?

In any case, I now uploaded both version: 
 -  https://www.cp2k.org/static/downloads/libxsmm-1.10.tar.gz
 -  https://www.cp2k.org/static/downloads/libxsmm-1.10.0.tar.gz

oschuett:
Looks good now.

I'll prepare a separate PR to make sure that CI checks also fail upon toolchain errors in the future.

hfp:
There is some [failed test](https://storage.googleapis.com/cp2k-ci/dashboard_psmp_report.txt) due to tolerance:

```
/workspace/cp2k/regtesting/local/psmp/TEST-local-psmp-2018-11-21_12-29-57/QS/regtest-gpw-4/H2-geo-2.inp.out
 Total energy  :   ref = -1.12365243716354 new = -1.12365243716352
 relative error:   1.77848717e-14 >  numerical tolerance = 1.0E-14
```

It is very likely a false positive.

oschuett:
I adjusted the tolerance as part of #103.

hfp:
Merci

PR #139, opened 2018-12-20
Fixed orbital angular momentum calculation for RTP:

mattiatj:
1. use non-symmetric neighbour list in qs_moments::build_local_magmom_matrix
2. track imaginary part of the density if magnetic moments are requested
3. use imaginary part of the density to calculate the expectation value

### DIFF
--- a/src/qs_moments.F
+++ b/src/qs_moments.F
@@ -48,7 +48,8 @@ MODULE qs_moments
    USE cp_result_types,                 ONLY: cp_result_type
    USE dbcsr_api,                       ONLY: &
         dbcsr_copy, dbcsr_create, dbcsr_deallocate_matrix, dbcsr_distribution_type, &
-        dbcsr_get_block_p, dbcsr_p_type, dbcsr_set, dbcsr_trace, dbcsr_type, dbcsr_type_symmetric
+        dbcsr_get_block_p, dbcsr_p_type, dbcsr_set, dbcsr_trace, dbcsr_type, dbcsr_type_symmetric,&
+        dbcsr_desymmetrize, dbcsr_multiply
    USE distribution_1d_types,           ONLY: distribution_1d_type
    USE kinds,                           ONLY: default_string_length,&
                                               dp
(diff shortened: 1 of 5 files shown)


src/qs_moments.F
```diff
@@ -522,24 +519,14 @@ SUBROUTINE build_local_magmom_matrix(qs_env, magmom, nmoments, ref_point, ref_po
                              sphi_b(1, sgfb), SIZE(sphi_b, 1), &
                              0.0_dp, work(1, 1), SIZE(work, 1))
 
-                  IF (iatom <= jatom) THEN
+                  ! IF (iatom <= jatom) THEN
```
dev-zero: please do not include commented out code, and re-indent the following code properly

src/qs_moments.F
```diff
@@ -1495,11 +1482,14 @@ SUBROUTINE qs_moment_locop(qs_env, magnetic, nmoments, reference, ref_point, uni
...
       TYPE(cp_result_type), POINTER                      :: results
       TYPE(dbcsr_p_type), DIMENSION(:), POINTER          :: magmom, matrix_s, moments, rho_ao
+      TYPE(dbcsr_type), POINTER                          :: rho_magmom_ao
       TYPE(dft_control_type), POINTER                    :: dft_control
       TYPE(distribution_1d_type), POINTER                :: local_particles
       TYPE(particle_type), DIMENSION(:), POINTER         :: particle_set
       TYPE(qs_kind_type), DIMENSION(:), POINTER          :: qs_kind_set
       TYPE(qs_rho_type), POINTER                         :: rho
+      ! TYPE(rt_prop_type), POINTER                        :: rtp
+      ! LOGICAL                                            :: imag_density = .FALSE.
```
dev-zero: dito, but ok'ish if you plan to use those with a follow-up commit

src/rt_propagation_types.F
```diff
@@ -97,6 +97,7 @@ MODULE rt_propagation_types
       REAL(KIND=dp)                               :: mixing_factor
       LOGICAL                                     :: mixing
       LOGICAL                                     :: do_hfx
+      LOGICAL                                     :: magnetic
```
dev-zero, line 100: unused?

src/qs_moments.F
mattiatj: Done.

src/qs_moments.F
mattiatj: Deleted, actually don't need them

src/rt_propagation_types.F
dev-zero: uh, my bad, sorry

dev-zero:
you have to run `make pretty` as well

dev-zero:
... and it seems that your changes broke the test with magnetic moments (`H2O-magnetic`)

mattiatj:
In my opinion the regtest H20-magnetic should give zero orbital angular momentum if there is no magnetic field

dev-zero:
Makes sense, and I guess this is why it says in the output:

```
[...]
  Magnetic Dipole Moment [WRONG]
    X=   -0.00851550 Y=    0.00496533 Z=   -0.65955777     Total=      0.65963143
[...]
```

Could you therefore please correct the test value in `tests/QS/regtest-ot-1/TEST_FILES` and remove the `[WRONG]` string in `src/qs_moments.F` since you fixed it?

Might be good if someone else @juerghutter or @marci73 could take a look as well?

PR #145, opened 2018-12-27
[WIP] Conventions: Remove INTENT(OUT) dummy args

oschuett:
Dummy arguments with `INTENT(OUT)` are a common pitfall because they require a full initialization. 

Unfortunately, these missing initializations are not detectable via Valgrind. And there is no help from the [compiler side](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=31447) either. 

On the other hand, there seems to be very little upside to using `INTENT(OUT)` and the Fortran language really only requires it for `PURE` subroutines. Hence, I was wondering if we could just remove  `INTENT(OUT)` altogether?

### DIFF
--- a/src/almo_scf_optimizer.F
+++ b/src/almo_scf_optimizer.F
@@ -4293,7 +4293,7 @@ CONTAINS
    SUBROUTINE opt_k_apply_preconditioner_blk(almo_scf_env, step, grad, ispin)
 
       TYPE(almo_scf_env_type), INTENT(INOUT)             :: almo_scf_env
-      TYPE(dbcsr_type), INTENT(OUT)                      :: step
+      TYPE(dbcsr_type)                                   :: step
       TYPE(dbcsr_type), INTENT(IN)                       :: grad
       INTEGER, INTENT(IN)                                :: ispin
 
--- a/src/aobasis/ai_angmom.F
+++ b/src/aobasis/ai_angmom.F
@@ -53,7 +53,7 @@ CONTAINS
       INTEGER, INTENT(IN)                                :: la_min, lb_max, npgfb
       REAL(KIND=dp), DIMENSION(:), INTENT(IN)            :: zetb, rpgfb
       REAL(KIND=dp), DIMENSION(3), INTENT(IN)            :: rac, rbc
-      REAL(KIND=dp), DIMENSION(:, :, :), INTENT(OUT)     :: angab
+      REAL(KIND=dp), DIMENSION(:, :, :)                  :: angab
 
       INTEGER                                            :: ax, ay, az, bx, by, bz, i, ipgf, j, &
                                                             jpgf, la, la_start, lb, na, nb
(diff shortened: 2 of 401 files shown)


alazzaro:
I like the idea. I will make the same change in DBCSR...

oschuett:
I'm now working on getting Valgrind support into the CI. Then we can see how many bugs this change would actually uncover.

dev-zero:
I am sorry, but I think this is heading in a completely wrong direction:

* `intent(out)`, even though not being fully enforced by the compiler serves as a documentation on the programmers intent on how to argument is being handled (and can be parsed by documentation generator)
* not having an `intent(out)` means that the caller must fully initialize the parameters (possibly with information it doesn't possess yet) if he doesn't want to assume what is being done in the subroutine/function being called (my assumption would be: any object I pass is re-initialized -- which is actually enforced when derived types are passed)
* the implementations of such subroutines/functions get even more sensitive to changes and refactoring since the interdependence between caller and callee will be increased

Furthermore I am quiet surprised that you didn't mention the most important error class here: passing derived types with initialized pointer components as `intent(out)` arguments may lead to memory leaks since using `intent(out)` on derived types makes the compiler call the default initialization which is often to `nullify` the pointer components. But even this can be easily prevented by writing proper OO derived types with a `final` routine (I would recommend [Modern Fortran by Clerman & Spector](https://www.cambridge.org/core/books/modern-fortran/878E48909A36DAA6BBC384F21174CB7A) at this point).

oschuett:
Yes, I realized that this is a pretty radical proposal. Loosing the documentation is certainly a downside. It has to be weighted against the upside of automatically eliminating an entire class of mistakes. The questions is how common these mistakes are. I want to asses that by running Valgrind on this PR. 

So, maybe we find it's not a big issue after all. Or we discover a more elegant mean of detection. This PR is very much WIP.

I did not mention leaks because we can already reliably detect those via LeakSanitizer.

mkrack:
Regardless of all possible technical implications of compiler and runtime behaviour related to the attribute INTENT(OUT), I suggest to keep the (original) information about the intention of the programmer concerning the input and output parameters at least as a comment. It has been good coding style already in the times before Fortran allowed for the INTENT attribute to specify such information in the header of a subroutine as a comment.

mkrack:
@oschuett I guess you have a script to remove all INTENT(OUT) attributes. Why not preprocessing the source code before a sdbg compilation and then running valgrind?

oschuett:
> I guess you have a script to remove all INTENT(OUT) attributes.

Yes, it's called `sed -i` ;-)

> Why not preprocessing the source code before a sdbg compilation and then running valgrind?

I don't think this will work, because code might actually rely on the default initialization of derived types. We could try [Joost's idea](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=31447#c3), but it won't be easy.

Issue #70, opened 2018-10-31
[toolchain] broken with OpenBLAS


dev-zero:
the first error one usually sees is that building ELPA breaks with the following in the `configure.log`:

```
configure: error: in `/data/tiziano/cp2k/tools/toolchain/build/elpa-2017.05.002/obj_no_thread':
configure: error: C compiler cannot create executables
```

the `config.log` shows the following:

```
configure:3938: mpicc -O2 -ftree-vectorize -g -fno-omit-frame-pointer -march=native -ffast-math  -I'/data/tiziano/cp2k/tools/toolchain/install/openblas-0.2.20/include'   -Wl,--enable-new-dtags  -L'/data/tiziano/cp2k/tools/toolchain/install/openblas-0.2.20/lib' -Wl,-rpath='/data/tiziano/cp2k/tools/toolchain/install/openblas-0.2.20/lib' -L'/data/tiziano/cp2k/tools/toolchain/install/scalapack-2.0.2/lib' -Wl,-rpath='/data/tiziano/cp2k/tools/toolchain/install/scalapack-2.0.2/lib'  conftest.c -lscalapack  IF_OMP(-lopenblas_omp|-lopenblas) >&5
gcc: error: IF_OMP(-lopenblas_omp|-lopenblas): No such file or directory
configure:3942: $? = 1
configure:3980: result: no
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "elpa"
| #define PACKAGE_TARNAME "elpa"
| #define PACKAGE_VERSION "2017.05.002"
| #define PACKAGE_STRING "elpa 2017.05.002"
| #define PACKAGE_BUGREPORT "elpa-library@mpcdf.mpg.de"
| #define PACKAGE_URL ""
| #define PACKAGE "elpa"
| #define VERSION "2017.05.002"
| #define EARLIEST_API_VERSION 20170403
| #define CURRENT_API_VERSION 20170403
| #define WITH_MPI 1
| /* end confdefs.h.  */
| 
| int
| main ()
| {
| 
|   ;
|   return 0;
| }
configure:3985: error: in `/data/tiziano/cp2k/tools/toolchain/build/elpa-2017.05.002/obj_no_thread':
configure:3987: error: C compiler cannot create executables
See `config.log' for more details
[...]
LIBS='-lscalapack  IF_OMP(-lopenblas_omp|-lopenblas)'
[...]
```

[...]

PR #152, opened 2019-01-04
toolchain: fix broken pexsi, quip and superlu build with OpenBLAS

robertschade:
This issue that this commit fixes is very much related to https://github.com/cp2k/cp2k/issues/70.


dev-zero:
while this fixes some cases, I think it breaks the case when the toolchain is called with `--enable-omp=no`, in which case (with OpenBLAS) will the string evaluate to `-lopenblas_omp` which is then not available. It might be safer to use something like

```bash
if [ "$ENABLE_OMP" = __TRUE__ ] ; then
  MATH_LIBS=$(resolve_string "${MATH_LIBS}"  OMP)
else
  MATH_LIBS=$(resolve_string "${MATH_LIBS}")
fi
```

respectively to simply use `$(resolve_string "${MATH_LIBS}")` (without the `OMP`) since this is only used by the configure scripts to validate that we have a valid BLAS implementation but since we are not doing any dynamical linking, only the libraries in the arch file will be linked in.

robertschade:
Sorry, this is something I have overlocked. I will prepare a suitable PR.

PR #156, opened 2019-01-07
do_regtest: explicitly output submodule SHAs

### LINKED ISSUE #154: Regtest logs do not indicate which version of DBCSR was used
The regtest logs accessible via the [CP2K Dashboard](https://dashboard.cp2k.org)) (for example, this [one](https://dashboard.cp2k.org/archive/cray-xc50-daint-gpu-psmp_perf/commit_53314e2d2677c6abcb41e93c64382d35294c83cf.txt)) do not indicate which DBCSR version was used in the build, and this information is cumbersome to retrieve a posteriori now that DBCSR is used via git submodules. 

Could the regtests log be completed to include the DBCSR version (or perhaps even better: SHA) used in the CP2K build?

dev-zero:
Although submodule SHAs can be derived from main repo SHAs, it is
easier to just explicitly mention them in the output, fixes #154

### DIFF
--- a/tools/regtesting/do_regtest
+++ b/tools/regtesting/do_regtest
@@ -720,6 +720,7 @@ done
 echo "--------------------------- GIT ------------------------------------------"
 
 git_sha="<N/A>"
+submodule_sha=()
 
 if [[ -d "${dir_base}/${cp2k_dir}/.git" ]] ; then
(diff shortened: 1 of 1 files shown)


dev-zero:
@shoshijak something like this?

shoshijak:
@dev-zero That looks great, thanks a lot!

PR #158, opened 2019-01-08
Remove __F2008 flag

mkrack:
Use intrinsic erf and erfc functions.

### DIFF
--- a/exts/Makefile.inc
+++ b/exts/Makefile.inc
@@ -17,4 +17,6 @@ dbcsr:
 	   FYPPEXE=$(TOOLSRC)/build_utils/fypp
 
 dbcsrversion:
-	@$(MAKE) -C $(EXTSHOME)/dbcsr version
+	@$(MAKE) -C $(EXTSHOME)/dbcsr \
+	   FYPPEXE=$(TOOLSRC)/build_utils/fypp \
+	   version


dev-zero:
@mkrack looks good, thank you. The errors from the CI indicate that the compiler invocation is missing the `-std=f2008` flag (which is mandatory for older GCCs to enable F2008).

oschuett:
You can simply change this in the [toolchain](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/install_cp2k_toolchain.sh#L1033) as part of this PR.

mkrack:
My fork is now 2 commits ahead, 1 commit behind cp2k:master. "git pull --rebase upstream master" results in the error message
From github.com:cp2k/cp2k
branch                master     -> FETCH_HEAD
Fetching submodule exts/dbcsr
Fetching submodule exts/dbcsr/tools/build_utils/fypp
Fetching submodule exts/dbcsr/tools/fprettify
fatal: bad object 0000000000000000000000000000000000000000

How do I update my fork without merge?

oschuett:
That error looks weird (maybe a git bug?) In any case, you don't need to do anything. Just keep committing and pushing.

mkrack:
Thanks @oschuett, just did so.

mkrack:
That test requires the presence of the wfn restart file from the preceding test case si7c_broy_gapw_a04_atomic.inp. I suggest to run always both cases in a sequence in order to avoid any incremental improvement of the wave function.

oschuett:
I already [modified](https://github.com/cp2k/cp2k/commit/9a686185383a5cdad7bdaac201ed4b7f04b1e045) the test to avoid overwriting the wfn file. My guess is that there is an initialization issue somewhere in the restart code and I'm quietly hoping to find it via #145.

oschuett:
I think, it [did](https://storage.googleapis.com/cp2k-ci/run-cp2k-sdbg-b7e5189d_report.txt) just that. Which log file are you looking at?

mkrack:
OK, so I assume "FAILED" and "WARNINGS" are expected?

oschuett:
Yes, those `FAILED` and `WARNINGS` message coming from [md5sum](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/ci_entrypoint.sh#L89). I though it would be nice to know which files have triggered the cleaning of the build cache. But I agree the messages are quite irritating.

oschuett:
Regarding `si7c_broy_gapw_a04_restart.inp`  the different regtest version [disagree](https://github.com/cp2k/cp2k/pull/158/checks?sha=afc0fed425ef021c029e9c88286c9554f539908c) about the reference value. Can you reproduce this locally and investigate?

mkrack:
I could not reproduce such variations for that test case locally by now using my sdbg arch file with GFortran 7.3. I recognize that you employ some more compiler options like -finit-derived -finit-real=snan -finit-integer=-42 -finit=snan. I get compilation errors for the dbcsr routines when I try to add these flags using GFortran 7.3.

alazzaro:
@mkrack What are the DBCSR errors? Could elaborate more?

Although we have to check for uninitialized values, I don't think these flags are useful for that...

mkrack:
I get the following error with GFortran 7.3, if I add the compiler flags "-finit-derived -finit-real=snan -finit-integer=-42" and only if I add all of them:
/home/krack/github/mkrack_f2008/cp2k/exts/dbcsr/src/utils/dbcsr_btree.F:39:46:
 $:ts
                                              1
Error: The element in the structure constructor at (1), for pointer component ‘p’ should be a POINTER or a TARGET

which looks more like a compiler bug.

mkrack:
I tried the -finit-* flags together fpe-trap with an older GFortran version some time ago. I observed either a strange compilation behaviour or nothing was trapped at runtime even for test cases for which something should have been trapped. Therefore I did not include these flags in my GFortran sdbg arch file at that time. It seems these features did not improve by now, i.e. in v8.2.

oschuett:
I recently managed to find two bugs using `-finit-derived` and `-finit-integer`. And at least with GCC 8.2 they do not cause any trouble. Hence, I'd like to keep these flags.

mkrack:
I have built CP2K locally for this PR using the toolchain based on gcc 8.2 and the regression test shows no wrong result. It is weird that the CP2K CI shows a wrong result. I cannot reproduce that value.

mkrack:
It seems CP2K CI is hanging.

oschuett:
The CI is working. It just takes longer as it needs to recompile everything due to the changed arch files. Furthermore, you now have to rebase because of my commits from yesterday. Hopefully, it works this time...

Too bad you can not reproduce it. However, for an initialization issue this is not totally unexpected. I'll try to have a closer look within the next days.

mkrack:
I always rebase as suggested, but I cannot get rid of the weird fatal error from git even after trying several times. I will cancel this PR and start from scratch.

PR #159, opened 2019-01-08
toolchain: fix broken pexsi, quip and superlu build with OpenBLAS

robertschade:
This PR fixes the issue at a higher level than the previous PR https://github.com/cp2k/cp2k/pull/152/commits/34cc5f62ba4347dee1365a4b328ca67e449a94c1.


dev-zero:
@robertschade I'm sorry, that's not what I meant with the comment on the PR #152

The problem is that the toolchain does not have a clear convention when building its libraries: some of them like OpenBLAS have explicit OpenMP handling: they build the serial version always and the OpenMP version if `--enable-omp=yes` (the default). Others don't. What I meant in #152 was that you could always have used `$(resolve_string "${MATH_LIBS}")"` instead of `$(resolve_string "${MATH_LIBS}" OMP)"` since the serial version will always be available (no matter whether OMP support is enabled) for the `sopt` variant AND since we are linking statically only the final linking step as specified in the CP2K arch file (written by the `install_cp2k_toolchain.sh`) are relevant.

With this PR, even a single-threaded version of CP2K (`sopt`) will link against the threaded OpenBLAS since the file loaded by `install_cp2k_toolchain.sh` will always set `FAST_MATH_LIBS` to the threaded one, likely resulting in multiple threads being spawned since people usually don't set `OMP_NUM_THREADS` to something reasonable when using `sopt`.

@oschuett since you did the review, please sort it out

oschuett:
@dev-zero, Good catch! Obviously, you would have been the better reviewer. Unfortunately, I had no way of knowing that you have the time to spare. I think we should come up with a better process to handle PRs from external contributors to give them a good experience. Obviously reviews take time. So maybe we should try to (self-) assign within a few hours and then aim for a first review within a couple of days?

robertschade:
@dev-zero, Shouldn't my change

`if [ $ENABLE_OMP = "__TRUE__" ] ; then`
`    export OPENBLAS_LIBS="${OPENBLAS_LIBS_OMP}"`
`else`
`    export OPENBLAS_LIBS="${OPENBLAS_LIBS}"`
`fi`

prevent sopt linking against the threaded OpenBLAS or am I overlooking something?

dev-zero:
@robertschade unfortunately not. The toolchain works as follows (assuming OMP support was requested):

1. build OpenBLAS (`libopenblas.a`) + a second set with OMP support (`libopenblas_omp.a`)
2. [load the setup script `setup_openblas`](https://github.com/cp2k/cp2k/blob/56c57bcb920a8a5ff5e5ce2b3381eb100539f51f/tools/toolchain/install_cp2k_toolchain.sh#L974) written by `install_openblas.sh` which now contains:
```
export FAST_MATH_LIBS="${FAST_MATH_LIBS} -lopenblas_omp"
```
3. write `local.sopt` which includes the `FAST_MATH_LIBS` variable content, which is now `-lopenblas_omp` (wrong)
4. write `local.popt`, which includes the correct `-lopenblas_omp`

So, assuming we do only static linking:

* if a package is OpenMP aware (either directly uses or does something different when the other components use OpenMP), we have to build multiple sets of it if requested (ELPA is such an example)
* if a package only checks whether the linker line is valid and some BLAS/LAPACK routines are available, we can always use a single-threaded BLAS implementation since that one is always available (just make sure that the package does not re-export the library flags)

Of course, the second case breaks should we ever do dynamic linking (then both variants of the BLAS library will enter the linker).

PR #166, opened 2019-01-09
Add compiler flag -fPIC

mkrack:
Suppress one error message


oschuett:
Just out of curiosity: What's the advantage of using `-fPIC`?

PR #168, opened 2019-01-10
Accelerate lapack and scalapack lib build

mkrack:
Perform a parallel make followed by a serial one to complete the library build which can reduce the build times significantly.

### DIFF
--- a/tools/toolchain/scripts/install_reflapack.sh
+++ b/tools/toolchain/scripts/install_reflapack.sh
@@ -55,7 +55,10 @@ TMGLIB       = libtmglib.a
 LAPACKELIB   = liblapacke.a
 EOF
             # lapack/blas build is *not* parallel safe (updates to the archive race)
-            make -j 1 lib blaslib > make.log 2>&1
+            # Run first in parallel which will result most likely in an incomplete library
+            make -j $NPROCS lib blaslib > make.log 2>&1
+            # Complete library in non-parallel mode
+            make -j 1 lib blaslib > make1.log 2>&1
             # no make install, so have to do this manually
             ! [ -d "${pkg_install_dir}/lib" ] && mkdir -p "${pkg_install_dir}/lib"
             cp libblas.a liblapack.a "${pkg_install_dir}/lib"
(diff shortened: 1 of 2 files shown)


mkrack:
@oschuett The toolchain update concludes with FAILED, but it is not clear to me from the report what went wrong. How can I get more detailed info?

oschuett:
Where did you see that error message?

Btw, I realized the convention check takes too long when the build cache gets cleared and [doubled](https://github.com/cp2k/cp2k-ci/commit/aab32e67b2fa4eafef624be2f24ff1fec7fdc82e) its CPUs.

mkrack:
Intermediately a sdbg status report file was shown with FAILED (link under Details). Maybe that was a link to wrong file. Eventually the status report was fine.

oschuett:
Did it say something like: `<Message>The specified key does not exist.</Message>` ? That is an error message from google storage because the first version of the report was not yet uploaded. The CI backend should probably upload an [empty report](https://github.com/cp2k/cp2k-ci/issues/31) to avoid confusion.

mkrack:
> Did it say something like: The specified key does not exist.

Yes, that is always shown initially.
I saw, however, a report from a toolchain screen output (with gcc, lapack, and scalapack rebuilds only, all OK), but at the end a status FAILED was shown. Anyway, that report disappeared finally.

oschuett:
Hmm, maybe that `FAILED` was from `md5sum` again? They are indeed confusing. I'll [remove](https://github.com/cp2k/cp2k-ci/issues/32) them. Otherwise the report should simply [grow](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/ci_entrypoint.sh#L100) with old lines remaining the same.

mkrack:
Initially all tests show as "Live Report" something like
`NoSuchKeyThe specified key does not exist.No such object: cp2k-ci/run-cp2k-conventions-488585c4_report.txt`
I think an empty page or just "No report available yet" would be sufficient.

oschuett:
Done via https://github.com/cp2k/cp2k-ci/commit/26915558213a1fb1181fd2d908a339fa1bee625d.

oschuett:
OK, I think I know what you saw. The compute node got preempted and during the shutdown it uploaded that broken report. Later the job got restarted and overwrote the report. I'll see how I can handle this case more gracefully.

mkrack:
Thanks for looking into that issue.

Issue #174, opened 2019-01-12
Limit runtime of individual regtests to 30 seconds


oschuett:
As @mbanck has [pointed out](https://github.com/cp2k/cp2k/issues/153#issuecomment-452487525), some regtest directories take a disproportional amount of time. While we can and should continue to split test directories, we also have to keep the runtimes of the individual regtests under control. Currently, the distributions looks like this ([notebook](https://github.com/cp2k/cp2k/files/2752018/analyse_regtest_runtimes.ipynb.zip)):

| Percentile | Runtime (popt) | # Slower Tests |
| ---------- | ------- | -------------- |
| 99-th | 39.64 sec | 32 |
| 98-th | 26.87 sec | 63 |
| 95-th | 16.12 sec | 156 |
| 90-th | 9.03 sec | 311 |
| 80-th | 4.26 sec | 621 |

I think, a feasible first limit would be 30 seconds. I have added a list to work through below. If you are familiar with any of these tests, please try to speed it up. Oftentimes it's just a matter of reducing the number of iterations or the numerical precision. Once we have established a limit we can easily [enforce](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest#L144) it.

- [x]  QS/regtest-sccs-3/H2O_sccs_td_cd5_geo_opt.inp **(119.11 sec)**
- [x]  QS/regtest-sccs-1/H2O_sccs_td_fft.inp **(95.28 sec)**
- [ ]  QS/regtest-embed/H4_H8_pbe_pbe0_const_pot.inp **(85.39 sec)**
- [x]  QS/regtest-sccs-3/H2O_sccs_td_cd5.inp **(83.35 sec)**
- [x]  QS/regtest-sccs-2/H2O_sccs_otdiis_cd5.inp **(77.50 sec)**
- [ ]  QS/regtest-ls-rtp/Ar_mixed_aa_planar-rtp-osc-field.inp **(73.61 sec)**
- [x]  QS/regtest-mp2-grad/MOM_MP2_geoopt.inp **(69.41 sec)**
- [ ]  QS/regtest-pod/furane-pbe-4A.inp **(67.92 sec)**
- [x]  QS/regtest-sccs-1/H2O_sccs_td_cd5_fg.inp **(66.76 sec)**
- [ ]  QMMM/SE/regtest-force-mixing/Lysozyme_small_NVT.inp **(66.62 sec)**
[...]

PR #175, opened 2019-01-13
Reduce runtime of SCCS regtests

mkrack:
Accelerate SCCS regression tests by using larger threshold values.


oschuett:
Great :-) How about also reducing the GEO_OPT steps to 1?

mkrack:
@oschuett Done. GEO_OPT steps reduced to 1.

oschuett:
Cross link #174.

PR #179, opened 2019-01-23
Build system: minor fixes

shoshijak:
`INSTALL.md`
- Remove an obsolete instruction 

`Makefile` and `exts/Makefile.inc`
- One would expect a build of CP2K after a `realclean` to re-build DBCSR as well, but it does not. Fix target `realclean` so that it cleans DBCSR as well.


dev-zero:
Looks good, merged. Thanks!

shoshijak:
thank you 👍

PR #188, opened 2019-01-30
Print both spin-components of computed properties in case of spin-polarised NEGF.

schulkov:


oschuett:
Hi Sergey, do you need any help with this PR? Otherwise, feel free to hit "Rebase and Merge".

Issue #114, opened 2018-12-05
Numerical instability somewhere in SCF_GUESS RESTART


oschuett:
I believe this race affects at least these tests:

-  `QS/regtest-gpw-5/si7c_broy_gapw_a04_restart.inp`
-  `QS/regtest-gapw-1/c8_broy_gapw_all_xashh.inp`
- `QS/regtest-gpw-7/si8_broy_kry_r.inp`

But probably there are more. After we fix this we should revisit the thresholds of all tests that use `SCF_GUESS RESTART`

The good news is that it's easy to reproduce as it shows up in about 10% of the runs:
```
for i in {1..10} ; do 
  ../../../exe/local/cp2k.ssmp si7c_broy_gapw_a04_restart.inp | grep -e "Total energy:"
done
```

oschuett:
Unfortunately, the problem persists even when compiled with `-finit-derived -finit-local-zero`.

mkrack:
Finally, I could reproduce a wrong result for the test input

`QS/regtest-gpw-5/si7c_broy_gapw_a04_restart.inp`

using the CP2K toolchain when I start with the gcc-8.2.0 compiler and build the toolchain with

`./install_cp2k_toolchain.sh --with-gcc --with-mpich --install-all`

Already the arch file local.sdbg gives reproducibly a wrong result in the order of 1.0E-6. At a first glance that seems to originate from differences in the information read from the restart file, but that turned out to be wrong, though the MO eigenvectors may differ depending on the built. But this is due to MO degeneracy which causes differences in the MO orthonormalization.
Tracing the issue led to the routine make_basis_lowdin which builds (C^T *S*C)^(-1/2) for the Lowdin orthonormalisation of the MOs read from the restart file in the case of smearing. Though the MO coefficient C from the restart file were already orthonormal for the given overlap matrix S, the routine does not return a unit matrix. This is caused by a not fully orthonormal set of eigenvectors returned by the LAPACK solver routine SYEVD. For the specific case above, only one pair of eigenvectors is not orthonormal (0.0044 instead of 0.0), but that is sufficient to spoil the transformation of  S. Thus, eventually, the originally orthonormal MO eigenvectors become slightly non-orthonormal after that "re-orthonomalisation step".
I have introduced a CHECK_DIAG flag in my cp2k fork that checks the orthonormality of the eigenvectors calculated by the routine `choose_eigv_solver`. A full regression tester run with local.sdbg resulted in failures for the following test inputs:
[...]

PR #193, opened 2019-02-05
Use Fortran intrinsics erf and erfc; add CHECK_DIAG flag

mkrack:
See [here](https://github.com/cp2k/cp2k/issues/114#issuecomment-460726765 ) for further details concerning -D__CHECK_DIAG.


dev-zero:
hmpf, the toolchain doesn't get that it has to rebuild everything providing a Fortran module when the compiler changes

oschuett:
Yeah, the toolchain is not really aware of the dependencies between packages. We could of course add them, but it would be a bit of work. Matthias, are you now certain that it's a bug in gcc 8, or is this a one-off experiment?

mkrack:
@oschuett I am quite confident that gcc-8.2.0 does not compile LAPACK properly. I do not get any regtest errors with gcc-7.4.0 using the toolchain locally. In my opinion, we should use the most recent gcc compiler release. The gcc-8.2.0 release is not the newest one (check [here](https://gcc.gnu.org))

Release date  |  gcc version
------------- | ---------
2018-12-06 | gcc-7.4.0
2018-10-26 | gcc-6.5.0
2018-07-19 | gcc-8.2.0 

@dev-zero Is it not possible to compile the selected gcc compiler first and then everything is compiled with that compiler version. If the gcc release is changed, then the whole toolchain has to be compiled from scratch.

dev-zero:
@mkrack not without introducing proper dependency tracking in the toolchain (or a dependency-like workaround just for the compiler).

8.2.0 is the latest release in the release branch 8 (which is maintained alongside with 7), distros [usually backport](https://src.fedoraproject.org/rpms/gcc/tree/f29) changes. An 8.3 [was expected for end of last year or beginning of this](https://lwn.net/Articles/760970/).

oschuett:
> In my opinion, we should use the most recent gcc compiler release.

I disagree. I think we should try to use the release with the highest version number. Otherwise we'd be jumping back and forth between the branches, which have feature gaps.

> gcc-8.2.0 does not compile LAPACK properly.

Huh, than this is a pretty bad regression. We should certainly file a bug with gcc for it. Still, strange that nobody else noticed this.


> If the gcc release is changed, then the whole toolchain has to be compiled from scratch.

Right, but for performance reasons the CI does not build the toolchain from scratch for each test run. 
Instead it simply runs `install_cp2k_toolchain.sh` a second time. And then only the packages that have changed get rebuild, see e.g. [here](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/install_libxc.sh#L25). So, for the CI to handle compiler version changes properly we'd have to teach the toolchain about dependencies between the packages. Given that @dev-zero is working on replacing the toolchain with Spack, I'm not sure if it's worth the effort. So, maybe we just bypass the CI for this PR?

mkrack:
I get 9 failed and 5 wrong regression tests with gcc-8.2.0 and clean sdbg test runs using gcc-7.4.0 as well as with gcc-6.5.0. Therefore I would appreciate to get rid of gcc-8.2.0 as the default toolchain compiler employed by the CI. We have wasted enough resources using an unreliable compiler release. I would indeed appreciate if we could bypass the CI this time, since I am curious how the other regression tester will digest the recent changes.

oschuett:
> We have wasted enough resources using an unreliable compiler release.

I totally understand your frustration. I've been looking at this several times myself without success. It's great that you managed to narrow it down!  Now I would like to spend just a little bit more time to thoroughly understand this bug. My fear is that otherwise it will not get resolved and we are basically cutting our self off from new gcc releases.

> I get 9 failed and 5 wrong regression tests with gcc-8.2.0 and clean sdbg test runs using gcc-7.4.0 as well as with gcc-6.5.0.

So, it is definitely a corner case problem, otherwise it would show up in more regtests. But it could still be a bug in CP2K, right? I think we should try to build a simple reproducer with only the call to `syevd` and all the cp_fm stuff striped away. Either we discover along the way that the problem really resides within CP2K, or we have something that we can file with gcc.

If you're busy, I can take over building the reproducer. Just give me a day or so.

mkrack:
We apply quite a lot of debug flags which are rarely used by others and thus it is indeed likely that we encounter a corner case potentially triggered by a not so well tested compiler flag. I don't think that this specific issue is a CP2K problem, but we need a reliable compiler to find the CP2K specific problems. I already try to further narrow the syevd problem.

mkrack:
The wrong results, caused by non-orthonormal eigenvectors returned by SYEVD, occur when the LAPACK/BLAS libraries are compiled with the -ffast-math flag using the gcc-8.2.0. LAPACK/BLAS relies on the IEEE-754 compliance of the compiler. Therefore the use of the -ffast-math flag might be hazardous.

oschuett:
Nice solution! And it's good that we can keep using the latest gcc.

dev-zero:
Nice find! I guess it's time to ditch `-ffast-math` once and for all then and leave the selection of it to the specific software (like fftw). This is probably also the reason why Linux distros didn't see any of this behaviour.

mkrack:
I would also vote for discarding the -ffast-math flag, since the often minor acceleration does not compensate for the potential risks.

PR #195, opened 2019-02-07
Drop -ffast-math everywhere

dev-zero:
Check the discussion and latest failures with LAPACK using gcc-8.2.0
and `-ffast-math` here:

  https://github.com/cp2k/cp2k/pull/193

Just. Not. Worth. It.


oschuett:
~~I think this is a bad idea.~~
I think we should be careful (sorry for the initial wording).

Just because we stop using fast-math for our tests doesn't mean everybody else does. AFAIK, ifort has similarly aggressive optimization enabled by default. Also, have you measured the performance implications?

oschuett:
Just to clarify: I think fast-math compatibility is a property of a code. Hence, I'm totally fine with not using it in the toolchain for packages that don't support it. However, for CP2K itself we should not simply give this up. In my understanding it's a proof of robustness.

dev-zero:
> I think this is a bad idea.
> I think we should be careful (sorry for the initial wording).
> 
> Just because we stop using fast-math for our tests doesn't mean everybody else does. AFAIK, ifort has similarly aggressive optimization enabled by default.

From what I've seen so far, people are usually copying one of the arch files and modify it until it _just works (tm)_. Stopping to advocate `-ffast-math` will at some point also stop people from using it.

> Also, have you measured the performance implications?

No, I have not. Was that done when `-ffast-math` was introduced in the first place?

But consider this:

* In production runs (supercomputing centers) people should be using the provided optimized math and FFT libraries which should (ideally) be the source of the largest portion of floating point operations. Turning `-ffast-math` on/off for CP2K alone should not make much difference, otherwise we have to re-investigate those parts of the code.
* We have now seen a reference implementation - which I assume to be written by people with a lot of experience - fail with `-ffast-math`. We therefore have to assume that all other libraries in the toolchain may potentially suffer from the same or similar problems and therefore turn it off for all of them.
* Our users do not have the same experience as for example @mkrack: the only thing they see is a failed calculation - be it a wrong number or a crash - and wasted man- and node-hours. And they will likely blame it on the code.
* For practicality we usually do not include long-running calculations with many iterations in our regtests, meaning that we currently don't even test for a possible accumulation of numerical errors over long periods. Akin to tuning a car and testing it on a 10km drive while the customer would like to go on a 1000km trip instead.

mkrack:
@oschuett 
> AFAIK, ifort has similarly aggressive optimization enabled by default.

Please note that we employ in most intel arch files -fp-model precise which disables fast-math as far as I understood (see [here](https://software.intel.com/en-us/node/522979))

alazzaro:
I would like to add my two cents:
1) we dropped fast-math in DBCSR. It was measured to be useless and we don't support it anymore
2) If you rely on fast-math to gain in performance, then you are in the wrong design... This option is compiler specific and imho this flag is used to debug the compiler itself, not the applications...

alazzaro:
> @oschuett
> 
> > AFAIK, ifort has similarly aggressive optimization enabled by default.
> 
> Please note that we employ in most intel arch files -fp-model precise which disables fast-math as far as I understood (see [here](https://software.intel.com/en-us/node/522979))

The -fp-model was needed for 17.x, it is not anymore for 18.x (and some people already tested with 19.x...). Now the claim to be GNU compatible by default...

oschuett:
It seems we're really having two discussions at once:
 1. What is the performance impact? 
 2. How should we deal with the risk?

## Regarding Performance

> libraries which should (ideally) be the source of the largest portion of floating point operations.

Yes, there is a trend towards using more libraries for established methods. And DBCSR is a good example as it has offloaded most of it's FLOPs to libxsmm / libcusmm. However, CP2K still has a lot of FLOPs. And it's going to stay like this because there will always be a new method for which no established library exists. Also, there is legacy code like e.g. our XC functionals.

>  > Also, have you measured the performance implications?

> Was that done when `-ffast-math` was introduced in the first place?

Most probably, but that was before my time. We should measure it again. I wouldn't be surprised to find 10-30% performance drops.

## Regarding Risk

Every compiler optimization comes with a risk. Not so long ago did `ifort -O3` not work for all our modules. I do agree that we should communicate this openly, possibly even discourage `--ffast-math`. However, it is very popular and (many?) people will continue to use `--ffast-math` regardless. Hence, in my opinion the responsible thing todo is to try and keep CP2K fast-math compatible.

So, if the performance loss  turns out to be negligible, maybe the middle ground is to add a dashboard test for fast-math?

mkrack:
I have run some small, not representative tests using the H2O-32 benchmark input using OT and also alternatively diagonalization (TD), since my last performance test on the impact of fast-math is quite long ago:

Test input | -ffast-math | no -ffast-math
---------- | ----------- | ----------------
| H2O-32 (OT) | 485 s | 444 s
| H2O-32 (TD) | 486 s | 440 s

The "-ffast-math" timings (average of several runs) correspond to the current toolchain using local.sopt and the "no -ffast-math" to the complete removal of that flag from the toolchain scripts as done by @dev-zero .

Frankly speaking, I am puzzled by these results, since fast-math is slower by almost 10%. At the same time -ffast-math increases the risk of having numerical issues which rather destabilize the regression testing. 

@oschuett Sorry, I don't see the benefit in supporting that flag given the potential risks.

oschuett:
@mkrack thanks for running these benchmarks! This is in fact surprising. Maybe you could once run cp2k with `-ffast-math` using the toolchain without `-ffast-math`? Just to be sure that a speedup in one of the dependencies does not cover-up a performance-lose within cp2k?

> Sorry, I don't see the benefit in supporting that flag given the potential risks.

I'm starting to agree. However, I still think we have a responsibility towards our users because the flag was advocated for so long. So, I'd propose we either add a dashboard test that runs with `-ffast-math` or we use [`__FAST_MATH__`](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-ffast-math) to abort the compilation.

dev-zero:
@mkrack is it possible that the number of SCF cycles or the number of iterations or quality for the initial guess are different?
@oschuett or adding a fixed warning (and increase the warning counter) if `__FAST_MATH__` is defined?

mkrack:
> 
> 
> @mkrack is it possible that the number of SCF cycles or the number of iterations or quality for the initial guess are different?

@dev-zero SCF_GUESS is ATOMIC and the number of calls to high-level routines (e.g. SCF inner loop calls and calculate_rho_elec) are the same.

oschuett:
> or adding a fixed warning (and increase the warning counter) if `__FAST_MATH__` is defined?

I think it would be best to catch this at compilation time when people fiddle with the flags anyways. Runtime warnings are often ignored and many people don't compile cp2k by themselves.

So, how about adding an overwrite flag like `-D__USE_FAST_MATH_AT_OWN_RISK` ?

dev-zero:
@oschuett is there a portable and supported way to interrupt compilation of Fortran code other than inserting invalid code (which sometimes does not even show the correct line but just the next valid statement)?
Otherwise we would have to shove this into the C code, since pragmas are ignored in Fortran code (strict mode for preprocessor) and this depends on the user specifying the same compiler flags for Fortran and C code, which they not necessarily do (check the Spack recipe for CP2K).
Furthermore we still want to taint the output (like adding a flag) to make sure we get that `-ffast-math` was used when users are sending their output.

dev-zero:
... or we do it in the `Makefile`.

oschuett:
I'd go with inserting invalid code. We already use it eg. for [libxc](https://github.com/cp2k/cp2k/blob/master/src/xc/xc_libxc_wrap.F#L23). AFAIK `__FAST_MATH__` is gnu specific anyways and gfortran prints the offending lines. Unfortunately, it seems there is no equivalent macro in ifort.

If we add an overwrite flag than this would be part of every output through `cp2k_flags()`.

mkrack:
> Maybe you could once run cp2k with `-ffast-math` using the toolchain without `-ffast-math`? Just to be sure that a speedup in one of the dependencies does not cover-up a performance-lose within cp2k?

Here is the updated table:

Test input | -ffast-math | no -ffast-math | -ffast-math (CP2K only) | 
---------- | ----------- | ---------------- | ------------------------ |
| H2O-32 (OT) | 485 s | 444 s | 488 s |
| H2O-32 (TD) | 486 s | 440 s | 491 s |

It has more or less no impact, if only CP2K or the whole toolchain is compiled with -ffast-math.

mkrack:
I get still 3 failing regression tests with the current toolchain using -ffast-math when I run the local.sopt regression test with the -D__CHECK_DIAG flag, even if fast-math is now disabled for the SCALAPACK and LAPACK builts. On the other hand, I don't get any failing test if I drop the -ffast-math flag entirely. It seems that the -ffast-math flag affects also the computation elsewhere.

oschuett:
I think `local.sopt` uses OpenBlas.

Should we maybe split this PR into two parts by first removing `-ffast-math` from all the dependencies in the toolchain?

mkrack:
The local.sopt regression test is also clean, if I put back the -ffast-math flag for the CP2K compilation only. Nevertheless, I consider that flag as potentially detrimental.

oschuett:
Yes, I'm totally fine with removing `-ffast-math` from the toolchain. And since there is apparently no performance advantage, I'm also fine with removing it from the `local.*` files. I just think we should have some kind of protective measure in place first.

dev-zero:
Since we're targeting gfortran only, we can use the `#error` preprocessor macro (tested with gfortran-4.8.5 and 7.3.1) to fail compilation somewhat gracefully. The logic for adding a `cp2k_flag` would be somewhat misleading (adding the `force_fast_math` flag only if both `-ffast-math` and `-D__FORCE_USE_FAST_MATH` where used) plus it would indicate that fast-math is somewhat a feature, which it is not.
Therefore I prefer to add a pretty noisy warning to taint the output instead and if sysadmins still decide they know better and ship with `-ffast-math` enabled, they can patch it out.

dev-zero:
It seems the CI still uses a fast-math enabled toolchain for the conventions checker. @oschuett is there a way we can nuke the CI cache?

oschuett:
I'm a bit confused because it does run `make distclean`. Let me have a look. Cache invalidation is hard...

oschuett:
This is really strange because the actual compiler line does **not** contain `-ffast-math`:
```
mpif90 -c -march=native -fno-omit-frame-pointer -g  -O3 -funroll-loops  -fopenmp -I'/opt/cp2k-toolchain/install/mpich-3.2.1/include'  -I'/opt/cp2k-toolchain/install/openblas-0.3.5/include' -I'/opt/cp2k-toolchain/install/fftw-3.3.8/include' -I'/opt/cp2k-toolchain/install/libint-1.1.6/include' -I'/opt/cp2k-toolchain/install/libxc-4.3.0/include' -I'/opt/cp2k-toolchain/install/libxsmm-1.10.0/include' -I'/opt/cp2k-toolchain/install/elpa-2017.05.003/include/elpa_openmp-2017.05.003/modules' -I'/opt/cp2k-toolchain/install/elpa-2017.05.003/include/elpa_openmp-2017.05.003/elpa' -I'/opt/cp2k-toolchain/install/scotch-6.0.0/include' -I'/opt/cp2k-toolchain/install/parmetis-4.0.3/include' -I'/opt/cp2k-toolchain/install/superlu_dist-5.1.2/include' -I'/opt/cp2k-toolchain/install/pexsi-0.10.2/include' -I'/opt/cp2k-toolchain/install/quip-cc83ceea5776c40fcb5ab224a25ab04d62175449/include' -I'/opt/cp2k-toolchain/install/gsl-2.5/include' -I/opt/cp2k-toolchain/install/spglib-1.12.1/include -I/opt/cp2k-toolchain/install/hdf5-1.10.4/include -I/opt/cp2k-toolchain/install/sirius-6.0.0/include -I'/opt/cp2k-toolchain/install/json_fortran-7.0.0/include' -ffree-form -std=f2008 -fimplicit-none  -Werror=aliasing -Werror=ampersand -Werror=c-binding-type -Werror=intrinsic-shadow -Werror=intrinsics-std -Werror=line-truncation -Werror=tabs -Werror=target-lifetime -Werror=underflow -Werror=unused-but-set-variable -Werror=unused-variable -Werror=unused-dummy-argument -Werror=conversion -Werror=zerotrip -Werror=uninitialized -Wno-maybe-uninitialized -Wuse-without-only -pedantic -Wall -Wextra -Wsurprising -Wunused-parameter -Warray-temporaries -Wcharacter-truncation -Wconversion-extra -Wimplicit-interface -Wimplicit-procedure -Wreal-q-constant -Wunused-parameter -Walign-commons -Wfunction-elimination -Wrealloc-lhs -Wcompare-reals -Wzerotrip -D__LIBXSMM  -D__parallel  -D__FFTW3  -D__LIBINT -D__LIBINT_MAX_AM=6 -D__LIBDERIV_MAX_AM1=5 -D__LIBXC -D__HAS_smm_dnn -D__SCALAPACK -D__ELPA=201705 -D__LIBPEXSI -D__QUIP -D__GSL -D__SPGLIB -D__HDF5 -D__SIRIUS -D__JSON   -D__COMPILE_ARCH="\"local_warn\"" -D__COMPILE_DATE="\"Wed Feb 13 17:20:00 UTC 2019\"" -D__COMPILE_HOST="\"run-cp2k-conventions-41b19c8b-6mscr\"" -D__COMPILE_REVISION="\"git:013aa26\"" -D__DATA_DIR="\"/workspace/cp2k/data\"" -D__SHORT_FILE__="\"environment.F\"" -I'/workspace/cp2k/src/' -I'/workspace/cp2k/obj/local_warn/psmp/exts/dbcsr' environment.F90 2> environment.F.warn
```

My guess is that one of the many warning flags implies `-ffast-math`. So, we either trace that down or add  `-D__FORCE_USE_FAST_MATH` to `local_warn.psmp`.

mkrack:
That's indeed strange. I cannot reproduce that error locally. I doubt that `__FAST_MATH__` is defined by one of the warning flags. Either there is still somewhere a -ffast-math flag hidden (mpif90?) or it is a bug.

oschuett:
Indeed `mpif90` is the culprit:
```
$ mpif90 -show
/opt/cp2k-toolchain/install/gcc-8.2.0/bin/gfortran -O2 -ftree-vectorize -g -fno-omit-frame-pointer -march=native -ffast-math -I/opt/cp2k-toolchain/install/mpich-3.2.1/include -I/opt/cp2k-toolchain/install/mpich-3.2.1/include -L/opt/cp2k-toolchain/install/mpich-3.2.1/lib -lmpifort -Wl,-rpath -Wl,/opt/cp2k-toolchain/install/mpich-3.2.1/lib -Wl,--enable-new-dtags -lmpi
```

mkrack:
@oschuett Great. How can one then enforce a complete rebuilt of the toolchain for CI?

oschuett:
> How can one then enforce a complete rebuilt of the toolchain for CI?

Currently, there is no simple way of doing this. As a workaround you could (pseudo-) change every toolchain script that needs to re-run.

oschuett:
I'm starting to think that this _workaround_ is actually not such a bad solution. Hence, I'd propose we add a comment like this to every toolchain script:
```
# reset: 0  Increase number to trigger CI re-build of this package.
```

mkrack:
That's a viable workaround which requires not too much effort given the fact that obviously a new more advanced system will become available soon.

mkrack:
@dev-zero Are you going to apply the workaround proposed by @oschuett ?

dev-zero:
@oschuett can you please elaborate a bit more why the toolchain does not get completely rebuild when I change `install_cp2k_toolchain.sh` in the first place?
Since we don't have any dependency tracking anyway, I would prefer to have only such version number in the main `install_cp2k_toolchain.sh`.

oschuett:
First off, the toolchain predates the CI, which is why it's lacking proper dependency tracking. 
What we rely on for now is an approximation I've added [very recently](https://github.com/cp2k/cp2k/commit/b4e4f571d04e5978d149bc84c3d9f50d1433bba8). Despite it's simplicity it works pretty well for the most common case of updating libraries like libxc, libxsmm, or Elpa.

We could of course improve the dependency tracking via:

```
if [[ $install_lock_file -nt $SCRIPT_NAME && \
      $install_lock_file -nt $SOME_DEPENDENCY_SCRIPT_NAME &&  ... ]]; then
  echo "foo is already installed, skipping it."
else
```

> I would prefer to have only such version number in the main `install_cp2k_toolchain.sh`.

While this is certainly the safest and simplest solution, it would also be the slowest. Rebuilding the full toolchain takes at least an hour (sometimes more due to preemption).

On the other hand, far-reaching changes like this PR are really rare. Most past modifications to `install_cp2k_toolchain.sh` did not require a full rebuild.

dev-zero:
any ideas about the changed value in the FIST calculation (regtest sopt)?

```
/workspace/cp2k/regtesting/local/sopt/TEST-local-sopt-2019-02-16_13-09-14/Fist/regtest-3/2d_pot.inp.out : 
 Local Curvature            = : ref = -0.3582893449 new = -0.5533958425  
 relative error :   3.52562276e-01 >  numerical tolerance = 5e-03  
```

mkrack:
Looks fine. Maybe just noise from an unstable test?

dev-zero:
@mkrack Yes, seems to be gone after the rebuild, but it seems like a very large fluctuation.

mkrack:
Let's hope that this change helps to stabilise some of the regression tests. The tests might need a general review. In the course of time, the threshold values of many tests increased quite a bit. But that is rather a topic for a new thread. A smaller fast running set of tests which gives results fluctuating by 1.0E-8 or less would be desirable. Threshold values of 1.0E-02 or even 1 are suspicious.

alazzaro:
For the reference, Cray compiler CCE sets __FAST_MATH__ already with -O1. Although we can just set -hfp0 to avoid that, I've just set the flag -D__FORCE_USE_FAST_MATH and left -O2 (thanks @dev-zero for pointing me to this solution). This issue was addressed by PR #208 #209

oschuett:
> I've just set the flag -D__FORCE_USE_FAST_MATH and left -O2

So, if we can't shake off `--fast-math` after all, how about adding a `sopt_fastmath` and `popt_fastmath` test to the dashboard? This would allow us to very quickly spot problems caused by these optimizations.


> In the course of time, the threshold values of many tests increased quite a bit

Indeed they have. I [recently](https://github.com/cp2k/cp2k/commit/4960dbb27c2f5ac98b428ff159acdc2ec242e905) added a column to the [regtest survey](https://dashboard.cp2k.org/regtest_survey.html) for finding loose thresholds.

mkrack:
> Indeed they have. I [recently](https://github.com/cp2k/cp2k/commit/4960dbb27c2f5ac98b428ff159acdc2ec242e905) added a column to the [regtest survey](https://dashboard.cp2k.org/regtest_survey.html) for finding loose thresholds.

@oschuett Thanks, that table is useful. Sorting the `#failed` column gives a kind of "negative" high-score of problematic regression tests.

PR #200, opened 2019-02-12
Update arch files

mkrack:


dev-zero:
@mkrack can you elaborate why libxc-4.3.x is no good? ... just to determine whether we have to downgrade it in the toolchain as well. There I currently have a patched 4.3.2, but I planned to upgrade to 4.3.3 which hopefully doesn't the patches anymore. And regtests didn't reveal anything irregular with that release.

mkrack:
When I upgraded first to libxc 4.3.0 and then to 4.3.1, there had been linking errors. Shortly afterwards, the version 4.3.2 and then 4.3.3 appeared as "stable releases". Therefore, I decided to wait a bit longer until the libxc 4.3 release has eventually stabilised.

dev-zero:
Thanks!
Yes, I encountered the linker errors as well, hence the patch in the toolchain-scripts to fix them.

PR #217, opened 2019-02-25
Improve runtime regtests MP2/RPA

fstein93:


tests/QS/regtest-mp2-3/TEST_FILES
```diff
@@ -1,3 +1,3 @@
 H2O-03.inp                                            11      3e-13            -17.156342200878413
-CH3_GPW.inp                                           11      4e-13            -7.2996767350835619
+CH3_GPW.inp                                           11      2e-5             -7.2996767350835619
```
oschuett: This is a pretty dramatic reduction in precision.

fstein93: Sorry, I was still using the -ffast-math flag. Now, it is the old precision.

oschuett:
Thanks for this PR! Lots of progress on #174 :-)

fstein93:
I had some time left and saw that issue.

PR #222, opened 2019-02-28
bug fix

rybkinjr:


oschuett:
This PR went again into the wrong branch. I now delete that branch an added a new branch protection rule which should prevent this from happening again.

PR #223, opened 2019-02-28
Dfet

rybkinjr:


oschuett:
Try:
```
git fetch upstream
git reset --hard upstream/master
```
Then re-create your bug fix commit.

Finally, do `git push --force origin dfet` .

PR #233, opened 2019-03-08
Annotate the atomic code.

schulkov:


dev-zero:
Awesome work! :+1: 

Personally I wouldn't have made the effort to also fill up `\history` tags since that info is in the git log, just in case you were planning to do more of this ;-)

PR #235, opened 2019-03-11
Input

mtaillefumier:
this pull request contains all necessary changes to configure SIRIUS using cp2k input files syntax. All sirius options are given through the interface which means that if new runtime options are added in sirius they will automatically be present without modification of cp2k code. At worst linking against sirius is needed at best nothing should be done (if shared libraries are used). 
- Sirius options are automatically documented. Documentation might be a bit simple but most parameters are self explanatory.
- test cases are now following cp2k input standard.
- Another important update is the presence of van der Waals corrections in sirius. The implementation uses libvdwxc (originally developed for octopus) as a backend. Forces and stresses are also there but test cases will be needed.


src/input_cp2k_pwdft.F
```diff
@@ -161,15 +161,15 @@ SUBROUTINE fill_in_section(section, section_name)
                CALL sirius_option_get_description_usage(section_name, name, description, usage)
                SELECT CASE (ctype)
                CASE (1)
-                  CALL sirius_option_get_int(section_name, name, dummy_i, vec_length);
+                  CALL sirius_option_get_int(section_name, name, dummy_i, vec_length);
```
dev-zero: I doubt that our convention checker requires a trailing space ;-)

mtaillefumier: Agreed, but the CI tells me that formatting is incorrect. That's why I had to run make pretty and push the changes. Emacs  suppress the trailling whitespace automatically but make pretty put it back.
I can not do much about it :(.
M.

mtaillefumier: I will also need to have psmp regtest activated to know that everything works with the CI. This would be very useful feedback.

oschuett: Ok, I just triggered it. I've actually been contemplating to make `psmp` a required check as well. The downside is that it takes slightly longer than `sdbg`.

oschuett: My guess is that the trailing space originates from the trailing semicolon, which could simply be remove.

dev-zero:
Since the input section is generated on the fly directly from SIRIUS, this means that the output of `cp2k --html-manual` varies depending on whether CP2K was linked against SIRIUS or not, correct?

mkrack:
More importantly, if the SIRIUS input sections and keywords do not appear in the xml dump (--xml flag), these sections and keywords won't appear in the CP2K online manual as well as in the editor support files like cp2k.vim.

oschuett:
I believe we already have some other places where the input depends on the compilation flags. What is new here is that SIRIUS only works with psmp. Hence, we'll have to generate the manual using the psmp binary from now on. I'll take care of that.

mkrack:
@oschuett Right, the flags work fine, as the available features of the actual CP2K executable are documented. For the online manual, however, we should then indeed switch to the psmp version to have all important features shown.

src/input_cp2k_pwdft.F
mtaillefumier: I will fix that. Indeed trailing semicolon are responsible for this.

mtaillefumier:
That's the drawback of this. the advantage is that if we add new options in sirius they are automatically included in cp2k without code modification.

oschuett:
Yes, this solution is a nice separation of concerns. I like it very much.

The manual generation is now using `psmp` (https://github.com/cp2k/cp2k/commit/93b5b51b086a828a750433401ece730954825bd3).

mtaillefumier:
It is just a lazy way to do it ;).

tools/toolchain/scripts/install_libvdwxc.sh
```diff
@@ -0,0 +1,106 @@
...
+        pkg_install_dir="${INSTALLDIR}/libvdwxc-${libvdwxc_ver}"
+        install_lock_file="$pkg_install_dir/install_successful"
+        if verify_checksums "${install_lock_file}" ; then
+            echo "libvdwxc-${libvdwxc_ver} is already installed, skipping it."
+        else
+            if [ -f libvdwxc-${libvdwxc_ver}.tar.gz ] ; then
+                echo "libvdwxc-${libvdwxc_ver}.tar.gz is found"
+            else
+                # do not remove this. They do not publish official version often
+                wget "https://gitlab.com/libvdwxc/libvdwxc/repository/archive.tar.gz?ref=master" -O libvdwxc-master.tar.gz
```
oschuett: I uploaded today's version to here: https://www.cp2k.org/static/downloads/libvdwxc-master-429a80027a2ec2c97e2f6d9a3dc84843f2739865.tar.gz . Please use that URL to make the build reproducible.

oschuett:
The new images are ready, but `install_libvdwxc.sh` still [failes](https://storage.googleapis.com/cp2k-ci/run-cp2k-sdbg-83ed8723_report.txt). Note that I already upgraded spglib (#245). So, you can remove that now from this PR.

mtaillefumier:
no problems. I will update it later today. Thanks

mtaillefumier:
could you do the same with sirius if not already done?

oschuett:
Sure, here you go: https://www.cp2k.org/static/downloads/SIRIUS-6.1.2.tar.gz

tools/toolchain/scripts/install_libvdwxc.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash -e
+[ "${BASH_SOURCE[0]}" ] && SCRIPT_NAME="${BASH_SOURCE[0]}" || SCRIPT_NAME=$0
+SCRIPT_DIR="$(cd "$(dirname "$SCRIPT_NAME")" && pwd -P)"
+libvdwxc_ver=${libvdwxc_ver:-master}
```
oschuett, line 4: The commit sha could be part of the version:
```
libvdwxc_ver=${libvdwxc_ver:-master-429a80027a2ec2c97e2f6d9a3dc84843f2739865}
```

mtaillefumier: - I actually have a problem mpich compiled in the toolchain. Curiously configure does not find the libs despite the detection of the header files. Any idea why it is behaving that way? 

- I will fix this. No worries.

mtaillefumier:
Ok, I will close this pull request because I have trouble debugging the build from the CI and the commits become messy as well. I will debug that locally with the dockerfile included in the toolchain. All comments have a been taken into account and already included. I will submit a new pull request when I have a consistent build

PR #237, opened 2019-03-11
Atomic code: new test routine, accurate dot products, removed few magic numbers.

schulkov:


src/atom_utils.F
```diff
@@ -2338,8 +2340,8 @@ SUBROUTINE atom_basis_condnum(basis, rad, cnum)
       REAL(KIND=dp), ALLOCATABLE, DIMENSION(:)           :: weig, work
       REAL(KIND=dp), ALLOCATABLE, DIMENSION(:, :)        :: smat
       REAL(KIND=dp), DIMENSION(3)                        :: rab
-      REAL(KIND=dp), DIMENSION(7, 7)                     :: sab
       REAL(KIND=dp), DIMENSION(:), POINTER               :: zeta, zetb
+      REAL(KIND=dp), DIMENSION(nco(lmat), nco(lmat))     :: sab
```
dev-zero, line 2344: Ok, I see why you do this, but this means that to call the function `atom_basis_condnum` the function `init_orbital_pointers(lnum)` must have been called before, which is far from ideal and which is why the tests fail (some other code directly calls `atom_basis_condnum`).

One way to fix this is to make `sab` an allocatable, then call `init_orbital_pointers(lnum)` (which is a NOP if it has been called before with `lnum` or bigger), then allocate `sab` with `nco(lmat)`.

Another way would be to drop the global `nco` array completely, make that calculation a function and leave the caching of the `nco` values to the functions which actually need it.

schulkov: You are definitely right. I have just had an impression that these pointers are initialised once when CP2K is started. In fact it is not the case: the subroutine init_orbital_pointers() called 32 times in total from different places of the code.

oschuett: Yes, `init_orbital_pointers()` is a bit of a pain. Since this modifies a global variable one has to be particularly careful with threading.

A pragmatic solution would be to simply call `init_orbital_pointers(20)` during startup. I don't think any method will every use a higher `lmax`. This should of course be check via `CPASSERT`.

PR #243, opened 2019-03-12
add autotools in the docker file

mtaillefumier:
Added some packages needed to build one of the dependency of SIRIUS.


oschuett:
It will take a while for the images to rebuild. They should be ready by tomorrow.

PR #253, opened 2019-03-14
add pkg-config in the docker ci image

mtaillefumier:
Well, unfortunately things that work on all computers I have at disposal do not work with the docker image by default.


mtaillefumier:
thanks :)

oschuett:
Sure, new images are ready.

PR #256, opened 2019-03-18
SIRIUS section follows cp2k input standard

mtaillefumier:
Opefully this pull request will finally work with the CI and include modifications suggested in the diverse comments of my last closed pull request.

This pull request contains all necessary changes to configure SIRIUS using cp2k input files syntax. All  options are given through the interface which means that if new runtime options are added in sirius they will automatically be present without modification of cp2k code. At worst linking against SIRIUS is needed at best nothing should be done (if shared libraries are used).

- Sirius options are automatically documented. Documentation might be a bit simple but most parameters are self explanatory. 
- test cases are now following cp2k input standard. 
- Another important update is the presence of van der Waals corrections in sirius. The implementation uses libvdwxc (originally developed for octopus) as a backend. Forces and stresses are also there but test cases will be needed.

PR #257, opened 2019-03-18
Docker: Disable core dumps and handle preemption gracefully

oschuett:


dev-zero:
Please don't. This changes default behavior on a system without the possibility to get the default behavior back.

oschuett:
Ok, how about a `max_coredump_size` argument that defaults to zero?

It came up this morning with #256. Every regtest failed, keeping the machine busy with writing core dumps until the kubernetes job eventually timed out before it could print the stack traces.

dev-zero:
yes, but use `$(ulimit -c)` as the default, then set the argument for the CI.

For systems with systemd you can easily load a coredump with gdb using

    coredumpctl gdb

oschuett:
Maybe the better solution would be to add `ulimit -c 0` to [ci_entrypoint.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/ci_entrypoint.sh) ?

PR #262, opened 2019-03-20
Docker: Remove leading ./ from artifacts.tgz

oschuett:


tools/docker/scripts/ci_entrypoint.sh
```diff
@@ -136,7 +136,8 @@ fi
 if [ -n "${ARTIFACTS_UPLOAD_URL}" ] &&  [ -d /workspace/artifacts ]; then
     echo "Uploading artifacts..."
     ARTIFACTS_TGZ="/tmp/test_${TESTNAME}_artifacts.tgz"
-    tar -czf "${ARTIFACTS_TGZ}" -C /workspace/artifacts .
+    cd /workspace/artifacts
+    tar -czf "${ARTIFACTS_TGZ}" *
```
dev-zero, line 140: you're aware that you're also excluding dot-files this way?

oschuett: Yes, but all the alternative solutions I found looked quite complicated. And I don't thinks it's an issue because artifacts are currently always in non-hidden sub-directories.

PR #272, opened 2019-03-22
Refactor RPA

fstein93:
Apart from some minor changes, I did some refactoring in the RPA Code. I moved the GW and AXK routines into separate files, added intents and pure keywords. Further, I removed some features that were buggy, left from the development phase or are inefficent compared with other functionalities.


fstein93:
Is still something missing?

oschuett:
Sorry, for the delay. I think, the problem is that currently nobody feels responsible for this part of the code. Maybe @rybkinjr could have a look?

rybkinjr:
I will as soon as I'm back on the 1st.

Vladimir

чт, 28 мар. 2019 г., 19:18 Ole Schütt <notifications@github.com>:

fstein93:
I absolutely agree. And even with that PR it is still ugly. The more I look in that code, the more issues I find that could be changed. But adding ~ 100 commits at once (or more) is probably not the best way to do that.

fstein93:
What about @pappnasenpappmeister (I know who that is)?

JWilhelm:
The refactoring of the GW part looks very good to me and is a highly appreciated step to arrive at a well-arranged RPA/RPA-AXK/SOS-MP2/GW code. Cheers, Jan

rybkinjr:
I'm fine with how AXK looks now.

oschuett:
So, if there are no more objections I'll merge this after lunch.

pseewald:
Please squash and merge s.t. we don't have the full history of copying gw to a new file and then reverting this change back (since this would lead to conflicts with other ongoing development of GW).

Issue #244, opened 2019-03-12
Slim down the toolchain


oschuett:
The aim of the [toolchain](https://github.com/cp2k/cp2k/tree/master/tools/toolchain) is to provide a simple reproducible way of building CP2K's dependencies. Naturally, the toolchain can not replace a full Linux distribution and has to have dependencies of its own. With fb915d00caf535890ffeaf89243b0f3f6474581a we added a few more of such dependencies. Hence, we should revisit if components like `make`, `cmake`, `binutils`, and `lcov` can also become external dependencies.

mtaillefumier:
I think the toolchain should only contain dependencies which cp2k calculations are sensitive of. GCC (binutils to lesser degree) and scientific libraries should be present. 

As for make, binutils, git, sed, cmake, autotools, etc, then they are provided by distributions already and we should rely on that. 

The toolchain has defaults but it is a very practical tool (at least for me). I could built cp2k on openpower just with the toolchain (I built many of the dependencies by hand for simplicity) in a couple of hours, without any specfic teaks to it.  they are probably better tools (spark maybe) but right now it is good enough to build cp2k without too much difficulties.

oschuett:
I moved lcov out of the toolchain via https://github.com/cp2k/cp2k/commit/b436df0f42cf0eef86daf91a6b5e0e7e4484143f. I'll leave valgrind for the time beeing because it's pretty entangled and I have no good way of testing it at the moment.

PR #277, opened 2019-03-28
Toolchain: Remove binutils, make, cmake, and lcov

oschuett:
See discussion at #244 .


dev-zero:
Please don't remove cmake: some distros (EPEL, etc.) provide an oldish version of cmake.

For lcov we explicitly use a version which includes https://github.com/linux-test-project/lcov/commit/94eac0ee870e58630d8052dca1181b0cf802525f. A new version including this commit was released a month ago (https://github.com/linux-test-project/lcov/releases/tag/v1.14), so I'm assuming that most distros do not provide a recent enough version yet.

oschuett:
Wow, you're right. SIRIUS [requires](https://github.com/electronic-structure/SIRIUS/blob/master/CMakeLists.txt#L1) CMake version 3.9. That was released only four months ago! This is either a mistake or CMake is not nearly as mature as I thought. In the later case we'll indeed have to keep it.

The defining property of an external dependency should be that it has stabilized enough so that an oldish version will do. I'm pretty sure CP2K would build just fine with a 10 year old version of GNU Make, Bash, and Python.

As for lcov, I had totally forgotten about that workaround. Thanks for reminding me :-) Nevertheless, I'm inclined to move it's installation into the Dockerfiles, because it's not actually needed for building CP2K. For the same reason we should probably move Valgrind out of the toolchain too.

dev-zero:
> Wow, you're right. SIRIUS [requires](https://github.com/electronic-structure/SIRIUS/blob/master/CMakeLists.txt#L1) CMake version 3.9. That was released only four months ago! This is either a mistake or CMake is not nearly as mature as I thought. In the later case we'll indeed have to keep it.

CMake 3.9 was released in 2017.

oschuett:
Upps, you're [right](https://github.com/Kitware/CMake/commit/f15cfd891d1e01247ed285320ae32b6c3182ac8f) again. I got fooled by the Github release page. Still, that version is too young to have reached all distros. According to our survey only 35% of users have CMake 3.9 or later installed:
```
=============== Which CMake version are you using? ===============
2.6:  1
2.8:  5
3.1:  8
3.5:  7
3.7:  1
3.8:  1
3.9:  3
3.11: 2
3.12: 4
3.13: 3
```

In contrast, json-fortran [requires](https://github.com/jacobwilliams/json-fortran/blob/master/CMakeLists.txt#L12) only CMake 2.8.8. Maybe @mtaillefumier could comment if SIRIUS really requires such a recent version of CMake?

dev-zero:
From DBCSR-side I can say that CMake 3.10 made some important progress, which is why we require it for the CMake-build for DBCSR.

mtaillefumier:
You consider cmake 3.9 as recent. Well it is old to my standard. 
- We use some of the functionalities of cmake 3.9 and I am not sure I want to touch the build system either to accomodate some old distros. Most cluster have updated versions of all these tools either through modules (daint, summit, summitdev, etc...) and distro provide updated package as well. 
- the only build system that works often very well over the years is autotools since it is only based on bash, make, and m4. Unfortunately introduction of libtool made things sometime harder (especially for cuda files)

So my advise is remove cmake, make, etc... and indicate that you need at least these versions. Many packages do that already (if not all). you can not adapt to every possible systems in the world by recompiling everything.

oschuett:
On one hand, I really think CMake should be an external dependency. On the other hand, removing it would break the toolchain for 65% of our users. SIRUS is still at a pretty early stage (few tests, no tutorials,...). So maybe we can simply make SIRUS optional in the toolchain for another year or so?

mtaillefumier:
The default activation of SIRIUS to cp2k will have to be discussed elsewhere. I am creating test cases for sirius (actually we have around 30 test cases already), but I have to convert them to cp2k input file and then make a pull request for that. I also have few fixes for the toolchain.

oschuett:
Ok, I'll settle on just removing make and binutils for today. I'll probably move lcov and valgrind to the CI Dockerfiles later. And for CMake let's just hope that it stabilizes and we can make it an external dependency eventually (in a year or so).

PR #278, opened 2019-03-29
Cubic scaling RPA with DBCSR tensors

pseewald:
This uses DBCSR tensors for cubic scaling RPA. This is not yet ready for merge, we need to wait for cp2k/dbcsr#149.

This does not change the input in `IM_TIME` section but parameters `EPS_FILTER_IM_TIME`, `EPS_FILTER`, `GROUP_SIZE_3C` and `GROUP_SIZE_P` need to be chosen differently for optimal performance and accuracy than with old implementation.

The following features are not yet implemented:
* Cubic scaling GW
* PAOs

This pull request will not remove the old implementation of cubic scaling RPA since Cubic scaling GW is still based on the old code. PAOs appear to be not useful in the current implementation so it should be removed at a later time.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "exts/dbcsr"]
 	path = exts/dbcsr
-	url = https://github.com/cp2k/dbcsr
-	branch = master
+	url = https://github.com/pseewald/dbcsr
+	branch = tensors-pr
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "exts/dbcsr"]
 	path = exts/dbcsr
-	url = https://github.com/pseewald/dbcsr
-	branch = tensors-pr
+	url = https://github.com/cp2k/dbcsr
+	branch = master
--- a/src/input_cp2k_mp2.F
+++ b/src/input_cp2k_mp2.F
@@ -1261,6 +1261,26 @@ CONTAINS
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
+      CALL keyword_create( &
+         keyword=keyword, &
+         name="DO_DBCSR_T", &
+         description="Uses DBCSR tensors for cubic scaling RPA/L-SOS-MP2", &
+         usage="DO_DBCSR_T", &
+         default_l_val=.TRUE., &
+         lone_keyword_l_val=.TRUE.)
+      CALL section_add_keyword(section, keyword)
+      CALL keyword_release(keyword)
+
+      CALL keyword_create( &
+         keyword=keyword, &
+         name="MAX_BLOCK_SIZE_SQRT", &
+         description="Square root of maximum block size for splitting blocks of matrix representation of large tensors "// &
+         "along the combined dimension. "// &
+         "Atomic blocks are split in order to allow for better load balance.", &
+         default_i_val=10)
+      CALL section_add_keyword(section, keyword)
+      CALL keyword_release(keyword)
+
       ! here we generate a RPA imag. time subsection to use for MAOs
       CALL create_mao_section(subsection)
       CALL section_add_subsection(section, subsection)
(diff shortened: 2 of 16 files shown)


oschuett:
A recent addition: https://github.blog/2019-02-14-introducing-draft-pull-requests/

oschuett:
Please rebase, I've recently removed Make from the toolchain.

oschuett:
I believe, I found [a solution](https://github.com/cp2k/cp2k-ci/commit/34c64c2d3b180bb9a35d534132a043dcbad3b551) for these annoying "please rebase" situations.


Nevertheless, the CI is not happy with your PR. Apparently, the DBCSR branch you are referencing does not exists any more:

```
 M	exts/dbcsr
 error: Server does not allow request for unadvertised object  4075e825308fd4db39003890e8699bd901f69c4
 Fetched in submodule path 'exts/dbcsr', but it did not contain 34075e825308fd4db39003890e8699bd901f69c4. Direct fetching of that commit failed.
 ```
Unfortunately, the error message [does not make](https://github.com/cp2k/cp2k-ci/issues/39) it into the uploaded log file.

oschuett:
Ok, now the git submodule error also appears in the logs.

hfp:
Regarding submodule issues: I found forking a project and changing the version of a submodule is ignored for the command line we recommend. For instance, CP2K's recommendation for DBCSR submodule is `git submodule update --init --recursive`. However, I found `git submodule update --init --recursive --remote` solves the afore mentioned issue.

hfp:
I ran into compilation errors with the master branch, and the resolution was to pull this PR into my fork. I guess, I had some dirty state or this PR is already a (partial) prerequisite of master. Running regtests now...

pseewald:
@hfp yes I created a new DBCSR release on master branch, which was not compatible with CP2K before merging in this pull request. However since CP2K was not updated to use the latest DBCSR release before this PR was merged in, everything should have worked if your submodule was in the correct state.

pseewald:
> Regarding submodule issues: I found forking a project and changing the version of a submodule is ignored for the command line we recommend. For instance, CP2K's recommendation for DBCSR submodule is `git submodule update --init --recursive`. However, I found `git submodule update --init --recursive --remote` solves the afore mentioned issue.

So probably this is the reason why you ran into problems: the [`--remote`](https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt---remote) flag checks out the latest commit in master:

> Instead of using the superproject’s recorded SHA-1 to update the submodule, use the status of the submodule’s remote-tracking branch

It is not correct to use `--remote` since this is not necessarily the correct state of the DBCSR submodule.

The submodule issues we saw before were probably due to `git rebase` in the DBCSR pull request changing the commit hashes.

hfp:
Thanks!   The `--remote` however is what I discovered is needed for my fork of CP2K since I want to refer to the tip of my forked DBCSR/develop. This is perhaps special and everyone who needs this behavior also needs to find out how to achieve it. So you are right regarding the regular user or even developer should get the revision that is recorded/permitted by the original master. Thank you for clarifying!

hfp:
It is still surprising that a fork of CP2K that needs a different DBCSR revision (and recorded that different requirement in `.gitmodules`) asks the user who clones the fork to run with the `--remote` flag. I guess this is an oddity of using submodules. I would expect a fork can fully determine its own submodule revisions without requiring the user to do something special.

mkrack:
The PSI regtester failed for the same reason. After adding the flag --remote to git submodule update, it works again, Thanks @hfp

alazzaro:
Yes, indeed, it was not updating to the new DBCSR version. However, now we get a compilation error and we have to provide a new DBCSR alpha release.

oschuett:
Some of these compilation problems might also originate from my https://github.com/cp2k/cp2k/commit/7dfc1a277a00b15088858ff4f64533cb5ab6e8d2. I'm rebuilding the containers now and then I'll rerun all the dashboard tests.

PR #284, opened 2019-04-04
do_regtest: Add timings report

oschuett:


tools/regtesting/timings.py
```diff
@@ -0,0 +1,41 @@
...
+#===============================================================================
+def main():
+    if(len(sys.argv) != 2):
+        print("Usage: timings.py <timings.txt>")
+        sys.exit(1)
+
+    fn = sys.argv[1]
+    timings = sorted([float(line.split()[0]) for line in open(fn).readlines()])
+
+    print('Plot: name="timings", title="Timings Distribution", ylabel="time [s]"')
```
dev-zero, line 20: Please use

```python
from __future__ import print_function
```

to make sure `print()` is actually a function in Python 2.

tools/regtesting/do_regtest
```diff
@@ -528,6 +529,7 @@ function run_regtest_dir() {
      fi
 
      printf "    %-50s %25s %20s %s\n" "${input_file}" "${test_value}" "${this_test}" "${leak_marker}" >> ${dir_out}/status/REGTEST_TASK_TESTS-$task
+     printf "%7.2f  %s\n" "${timing}" "${dir}/${input_file}" >> ${timings}
```
dev-zero: please use `"${timings}"` to avoid problems should `${dir_out}` contain spaces or other strange characters

oschuett: Done. Thanks!

tools/regtesting/timings.py
oschuett: It does actually work, but for the wrong reason. Thanks!

tools/regtesting/timings.py
```diff
@@ -0,0 +1,42 @@
...
+from __future__ import print_function
+
+#===============================================================================
+def main():
+    if(len(sys.argv) != 2):
+        print("Usage: timings.py <timings.txt>")
+        sys.exit(1)
+
+    fn = sys.argv[1]
+    timings = sorted([float(line.split()[0]) for line in open(fn).readlines()])
```
dev-zero: please use

```python
with open(fn) as fhandle:
    timings = sorted(float(line.split()[0]) for line in fhandle.readlines())
```

... because we all know that people copy stuff around and suddenly you have a fd leak in a loop.
Furthermore: `sorted()` can handle generator expressions.

oschuett: I didn't know about sorted(). That's indeed nice.

PR #291, opened 2019-04-09
Make clear that the toolchain provides GNU arch files

alazzaro:


INSTALL.md
```diff
@@ -31,6 +31,8 @@ The basic steps are:
 ```
 
 - Once the script has completed successfully, follow the instructions given at the end of its output.
+Note that the pre-built arch files provided by the toolchain are for GNU compilers, users have to adapt them for other compilers.
+It is possible to use the provided [arch files](https://github.com/cp2k/cp2k/tree/master/arch) as guidance.
```
oschuett: You can just use a relative link like `./arch/`.

PR #300, opened 2019-04-12
update libint -> libint2

pseewald:
After this PR, CP2K will use libint 2.5 (or upwards) instead of outdated libint 1. Since libint 2.5 comes with Fortran bindings, the compilation process of CP2K and the libint interface of CP2K become simpler. In particular the flags `-D__LIBINT_MAX_AM` and `-D__LIBDERIV_MAX_AM1` are no longer needed for CP2K.

After merging this PR, CP2K will no longer support libint 1.

Libint can be easily updated by running the `cp2k/tools/toolchain/install_cp2K_toolchain.sh` script of this pull request branch. If you use a CP2K that was not built with the toolchain script, see below for instructions on how to update libint manually.

## Instructions to compile a CP2K-compatible libint 2:

Download a libint generated for CP2K from [here](https://github.com/cp2k/libint-cp2k/releases/latest) (one of the libint tarballs, not the source code archive). The `lmax-5` library is recommended for most users/developers.

After extracting the tarball, run
```
./configure --prefix=[libint2 install directory] --enable-fortran \
    CXX=[C++ compiler] CXXFLAGS=[C++ compiler flags] FC=[Fortran compiler]

    make
    make install
```
For `CXXFLAGS` don't use `-g` flag since it will increase library size by a large factor (`-g1` seems to be ok though).

More detailed instructions for building libint can be found here:
https://github.com/evaleev/libint/wiki#compiling-libint-library

## Instructions to compile CP2K with Libint 2:

Change CP2K arch file: 
- remove the flags `-D__LIBINT_MAX_AM` and `-D__LIBDERIV_MAX_AM1`
- Add `-I'[libint2 install directory]/include'` to `FCFLAGS`
- Add `-L'[libint2 install directory]/lib'` to `LDFLAGS`
- Replace `-lint` by `-lint2` and remove `-lderiv`.


dev-zero:
Cool :-)

Some questions/remarks:

* does libint build a static or shared library (or both) by default?
* is OpenMP supported in any way?
* for the toolchain we should build a static library, in which case `-Wl,-rpath='<libint2 install directory>/lib64' ` is not needed
* can you explain the configuration flags for libint2?
* could we possibly (or do you already) verify that a provided libint2 is suitable?
* do the initial configuration options have any influence on the API/ABI of libint2? (important if linked dynamically)

pseewald:
> * does libint build a static or shared library (or both) by default?

by default libint builds a static library but a dynamic library can be requested by `./configure --enable-shared`

> * is OpenMP supported in any way?

I don't think so. SIMD vectorization is however a feature of libint 2 that we don't yet use in CP2K.

> * can you explain the configuration flags for libint2?

In short: these flags are roughly equivalent to our previous `-D__LIBINT_MAX_AM=6` and `-D__LIBDERIV_MAX_AM1=5` (so up to h-functions for energy and g-functions for derivatives). Furthermore I configure libint 2 with support for 3-center ERIs. These will be used in CP2K in the future.
 
In detail:
* `--enable-eri=1`: enable 4-center ERIs up to first order derivative
* `--enable-eri3=1`: same as above but for 3-center ERIs
* `--with-max-am=4`: by default, support Gaussians of angular momentum up to 4
* `--with-eri-max-am=5,4`: support 4-center ERIs for Gaussians of angular momentum up to 5 and 4 for first-order derivative
* `--with-eri3-max-am=5,4`: same as above but for 3-center ERIs
* `--with-opt-am=3`: enable optimizations up to ang. mom. 3
* `--enable-generic-code` and `--disable-unrolling`: I guess these flags limit the amount of generated code - I remember that I experienced that libint takes a very long time to compile without these and they are recommended for [other codes](https://github.com/evaleev/libint/wiki#program-specific-notes).

> * could we possibly (or do you already) verify that a provided libint2 is suitable?

Unfortunately there is no preconfigured libint that come with 3-center ERIs so I guess we can not avoid configuring a suitable libint ourselves.

> * do the initial configuration options have any influence on the API/ABI of libint2? (important if linked dynamically)

Yes the API changes with the configuration options. However since libint 2 comes with a Fortran API, CP2K does not require one specific configuration of libint (of course libint should be compiled with the necessary features)

pseewald:
I had to reintroduce `__LIBINT_MAX_AM` flag because we need to know at compile time which of the 

`_aB_s___0__s___1___TwoPRep_s___0__s___1___Ab__up_0`
`_aB_s___0__s___1___TwoPRep_s___0__s___1___Ab__up_1`
`_aB_s___0__s___1___TwoPRep_s___0__s___1___Ab__up_2`
(and so on until 4 * LIBINT2_MAX_AM_ERI)

are defined in libint. We can not easily get the macro `LIBINT2_MAX_AM_ERI` at compile time from libint, so we still need to define `-D__LIBINT_MAX_AM` in CP2K arch file.

oschuett:
> We can not easily get the macro LIBINT2_MAX_AM_ERI at compile time from libint

If I recall correctly, there used to be a header file which only contains pre-processor stuff so that it could actually be include it into a Fortran file.

pseewald:
> > We can not easily get the macro LIBINT2_MAX_AM_ERI at compile time from libint
> 
> If I recall correctly, there used to be a header file which only contains pre-processor stuff so that it could actually be include it into a Fortran file.

Yes you're right, so indeed we can get rid of `__LIBINT_MAX_AM` flag.

I'll try to integrate libint 2 into toolchain install script and after that, this PR is ready.

oschuett:
Unfortunately, the CI is currently in a bit of weird state as I'm moving the docker build into the check run. I'm afraid toolchain changes won't quite work right now. Can you wait maybe a day or two?

pseewald:
In any case we should announce this change a week or so before merging anyway to make sure that developers are prepared that cp2k will no longer compile with libint 1. Do we somehow need to prepare a new release for that as well?

oschuett:
Ok, the CI is back in one piece. It now runs docker build as part of the check_run, which should greatly improve reproducibility. I also had to make some changes on the toolchain itself, ie. please rebase.

pseewald:
> Ok, the CI is back in one piece. It now runs docker build as part of the check_run, which should greatly improve reproducibility. I also had to make some changes on the toolchain itself, ie. please rebase.

Great! Is there any way to see what goes wrong on the CI during the build of libint? The error should be logged in `build/libint-2.5.0/build/configure.log`.

oschuett:
This is something I'm still thinking about. It's not that easy because Docker doesn't not actually create a layer when a RUN command fails. Hence, there is no easy way of extracting only those files.

The simplest solution would be to temporary remove the redirection into the log file. Then configure's output will end up in the CI log.

Alternatively, you could debug interactively by commenting the lower part of the Dockerfile and then building the container locally:

```
cd tools/toolchain
Docker build -t partial_toolchain -f Dockerfile_until_line_65 .
docker run -ti partial_toolchain /bin/bash
```

pseewald:
@oschuett the CI did not checkout the correct CP2K commit (`-D__COMPILE_REVISION="\"git:9b85492\""`), could you have a look?

oschuett:
I see. The culprit is [here](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/install_basics.sh#L27). I'm currently traveling.  Maybe you can try to patch that file.

src/hfx_libint_wrapper.F
```diff
@@ -12,384 +12,410 @@
...
+      CPABORT("This CP2K executable has not been linked against the required library libint.")
+#endif
+   END SUBROUTINE
 
-      CHARACTER(LEN=*), PARAMETER :: routineN = 'initialize_libderiv', &
-         routineP = moduleN//':'//routineN
+   SUBROUTINE cp_libint_set_contrdepth(lib, contrdepth)
+      TYPE(cp_libint_t) :: lib
+      INTEGER           :: contrdepth
+      lib%prv(1)%contrdepth = contrdepth
```
oschuett, line 421: I believe an `#if(__LIBINT)` is missing here.

pseewald: is fixed now!

oschuett:
I just changed the convention checker to ignore Fypp variables. Hence, you can simply drop your last  commit 405d530 .

pseewald:
I realized that libint 2 depends on _boost_ and _GMP_ only for generating the code but not for compiling the library. Therefore I suggest that we provide a github repo with a pre-configured libint source code that has no additional dependencies.

pseewald:
@oschuett Could you create an empty repository `cp2k/cp2k-libint` that I could use to upload a pre-configured libint source code for cp2k?

hfp:
> cp2k/cp2k-libint [...] to upload a pre-configured libint source code for cp2k

A preconfigured library would also `--enable-generic-code`, right?  I guess it's still possible that one can configure manually (at the expense of the these somewhat elaborated dependencies).

oschuett:
Voilà: https://github.com/cp2k/cp2k-libint

pseewald:
Pre-configured libint source code for CP2K is now provided [here](https://github.com/cp2k/libint-cp2k) (thanks to @dev-zero for setting up the CI)

Library size is definitely an issue when it comes to supporting arbitrarily high angular momentum quantum numbers. When compiling with `--with-max-am=6` the library is 771 MB, with  `--with-max-am=5` it is still 409 MB. For comparison: libint1 library was ~30MB for  `--with-max-am=5`.

CP2K (psmp) compiled with libint (`--with-max-am=5`) : 447MB
CP2K (psmp) compiled without libint: 189MB

I suggest to configure libint with `--with-max-am=5` by default which should be enough for most users/developers. Users who need higher angular momenta would need to build their own libint.

oschuett:
Wow, that's more than a 10x increase in binary size. Out of curiosity, do you know why libint2 suddenly needs that much more space? Is all that code actually executed or could this be a target for JIT?

hfp:
@pseewald I wonder if the library can be stripped wrt to debug symbols?

pseewald:
> @pseewald I wonder if the library can be stripped wrt to debug symbols?

I checked again compiling with gfortran and the library is much smaller without `-g` flag: for `--with-max-am=5` it is 55 MB. So we are talking about an increase of library size of factor of 2 compared with libint 1, which should not be a problem.

PR #304, opened 2019-04-15
Add hybrid-meta-GGA functionals from Libxc

fstein93:
I applied the patches from Stanislaw Simko to Libxc such that we can use hybrid meta-GGAs from Libxc plus some regtests. See https://groups.google.com/forum/#!topic/cp2k/ARya5uM-pA4

### DIFF
--- a/src/xc/xc_libxc.F
+++ b/src/xc/xc_libxc.F
@@ -78,7 +78,8 @@ MODULE xc_libxc
       xc_libxc_wrap_version, &
       xc_libxc_wrap_functional_get_number, &
       xc_libxc_wrap_needs_laplace, &
-      xc_libxc_wrap_functional_set_params
+      xc_libxc_wrap_functional_set_params, &
+      xc_libxc_wrap_is_stable
 #endif
 
 #include "../base/base_uses.f90"
@@ -179,6 +180,9 @@ CONTAINS
          CASE default
             CPABORT(TRIM(func_name)//": this XC_FAMILY is currently not supported.")
          END SELECT
+      END IF
+      IF (.NOT. (xc_libxc_wrap_is_stable(xc_info))) THEN
+         CPWARN(TRIM(func_name)//" is not considered stable. Use with caution.")
       END IF
 
       CALL xc_f03_func_end(xc_func)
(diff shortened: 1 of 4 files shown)


fstein93:
Unfortunately, it worked on my machine. Without UKS the code is running in principle. Due to the error it might be an issue within Libxc. I will try a different functional.

fstein93:
I really wonder about that error. I can compile libxc 4.3.4 with the above code locally (gnu 4.9.2) and on Daint (gnu 6.0.4) and the regtests run without issues.

oschuett:
Are you running the debug version? Note that floating point exception are only trapped when you compile with ` -ffpe-trap=invalid,zero,overflow`.

fstein93:
Unfortunately, gnu 4.9.2 does not provide this (and some other) option(s) of your pdbg. With -ffpe-trap=invalid,zero,overflow, I get the same issues with other functionals on Daint. In the above check, the same test without uks and some others use the same lines of code and it is working there. So it is probably not our code causing the numerical issues.
Because a lot of functionals in Libxc are known to show floating point errors, the error in this test might really arise from Libxc, not from my modifications.
Maybe density smoothing helps.

fstein93:
Would it be okay for you to simply remove the failing regtest?

oschuett:
Yeah, floating point exception in libraries can be tricky. The best solution would be of course to fix libxc because this might be an actual bug. Alternatively, you can disable trapping while calling into the library. We do this e.g. for [pexsi](https://github.com/cp2k/cp2k/blob/master/src/pexsi_interface.F#L394).

You shouldn't remove the test, unless you think libxc is actually broken. However, in that case it would be best to remove the functionality from cp2k altogether to protect our users.

fstein93:
I think mGGA hybrids are a quite useful feature of Libxc that is requested. So I don't want to remove that code; the numerical issues of Libxc are known. If they occur, the user can make use of features of CP2K like density smoothing (examples are available even for Libxc), otherwise the Libxc developers have to be informed. (By the way, the Libxc 5.0 comes up with a new API again.)

oschuett:
If you have confidence in the results then temporarily disabling the trapping is the way to go.

Just to clarify: In the past we have encounter a few broken functionals in libxc, which we then disabled to protect our users. Hence, I wasn't sure if this is another one of those cases.

I think, it's great that you're adding these functionals!

fstein93:
I removed the failing regtest. But to give at least a hint to the user, I used the provided flags from Libxc to tell the user if a functional is not considered stable. (Right now (4.3.4) that holds for every functional but they plan to really use the flag later.)
I absolutely understand your position. But I think, this solutions is the best one. It will probably work for later releases, too. (Except they change the API too much)

fstein93:
The failing regtest is not caused by me (cf. CP2K dashboard).

fstein93:
Do you still have some concerns?

PR #306, opened 2019-04-16
updating DBCSR to v2.0.0-alpha1

pseewald:


oschuett:
I have stop the check_runs for this PR because I'm currently rebuilding the CI containers. I'll restart the runs once the containers are ready. I'm sorry for the delay. These changes to the CI are meant to avoid situations exactly like this.

PR #321, opened 2019-04-25
toolchain: set nvcc flags

### LINKED ISSUE #311: [toolchain] doesn't set nvcc optimisation flags
When using the toolchain script to create arch files and activating the `--enable-cuda` option, the generated arch file does not contain any optimisation flags for CUDA (see  `NVFLAGS` variable). 

We should probably add `-O2` (maybe even `fast-math`?) at least in the `?smp` versions.

shoshijak:
Adds `-O3` to the nvcc flags set by the toolchain script. 
Fixes #311.


shoshijak:
I'm not sure why this test is failing. 
Does this PR look okay?

alazzaro:
@shoshijak I think it was a glitch due to Ole's work on docker. Rerunning and now it is OK

shoshijak:
Thanks a lot! 
(I do not have write access to cp2k, so if this seems fine to cp2k developers, you're welcome to merge)

PR #322, opened 2019-04-25
Enable ELPA by default if compiled into CP2K

hfp:
I will add the label "*Needs discussion*" since I can imagine that some objectives are not the same for others...

**This PR enables more reasonable defaults for ELPA** (*opinion*):

1. Enables ELPA by default (if compiled into the binary),
2. Select ELPA-kernel according to CPUID.

The latter point uses LIBXSMM's CPUID functionality to select the kernel.

**Further change** (*can be questioned*): Introduced CP2K_DIAG_LIBRARY environment variable to adjust default of PREFERRED_DIAG_LIBRARY without using the keyword in the input file (see also the note below). This environment variable allows to quickly disable using ELPA even if it is compiled into CP2K (e.g., CP2K_DIAG_LIBRARY=1 selects basic ScaLAPACK instead of ELPA).

**Comments**: CP2K's ELPA support until this patch is somwhat a deviation from how (optional) libraries are handled i.e., ELPA may be compiled-in but it is never used unless specified by a keyword. This is a surprise for the user who compiled with ELPA for performance reasons.

---

**Notes**

* One can imagine that global keywords in CP2K's input may have corresponding environment variables (could be even an option when registering a keyword).
* This change somewhat obscures the established ELPA-tests since now every test potentially exercises ELPA (if compiled into the binary).
* This PR passes regression tests. However, error margins have to be adjusted since ELPA is always used when compiled int the binary.


dev-zero:
On one hand I agree that the defaults should be more user friendly, on the other worry about the combinatorial explosion of our flags plus the change of default behavior based on which packages were linked in.
If we're going this route I suggest we start dropping old versions of ELPA more aggressively to limit the number of combinations. Second (and this is what I'm already working on) we introduce unittests for our wrappers of the libraries we use: even though ELPA has a compat layer for the deprecated API we use currently, the tests show a significant change.

Furthermore, even though there are good reasons to always use `libxsmm`, I would prefer to use an independent library for CPUID detection, like https://github.com/google/cpu_features for example.

hfp:
> I would prefer to use an independent library for CPUID detection

Sounds reasonable! A first step could be a wrapper function in CP2K (using LIBXSMM initially). As a fallback, a generic CPU could be signaled (or a "CPUID" that is derived from compiler's standard defines that are related to the given target-flags e.g., \_\_AVX\_\_ may be only defined on Sandybridge onwards, etc.).

hfp:
I will make a separate PR for extending `base/machine.F`.

hfp:
@dev-zero Do you think PR #325 is an acceptable proposal/design for a separate CPUID functionality?

dev-zero:
@hfp looks good. Thanks a lot!

hfp:
I will merge PR #325 and then update this PR.

oschuett:
I agree that we should drop support for older ELPA version. I also agree that we should not rely on libxsmm for utility functions, although we already started doing so with `libxsmm_timer_duration`. However, I'm not sure if pulling in yet another library for this is necessary as the required code is presumably fairly short. So, maybe we can simply duplicate these snippets of C code in CP2K.

hfp:
As of CP2K/master, CPUID interface/functions are now part of the Machine module (`m_cpuid_static()` and `m_cpuid()` plus some enums). The implementation simply returns `MACHINE_CPU_GENERIC`, or on Intel architecture (IA) the code path according to compiler flags is somewhat discovered (`__AVX__`, etc.), or if LIBXSMM is present the real/dynamic CPUID is returned. So we can (re-)place/add whatever CPUID/code we want without touching anything other but the Machine module, and until then (or in case of unknown CPU) nothing goes slower when compared to having no CPUID functionality.

hfp:
I will wait merging this PR until Dashboard is green again. Merging this PR, will show "wrong results" with some Dashboard systems (those compiling with ELPA) because of running *all tests* with activated ELPA. This will require to calibrate error margins accross the test set.

src/input_cp2k_global.F
```diff
@@ -99,28 +107,57 @@ SUBROUTINE create_global_section(section)
...
          citations=(/Marek2014/))
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
+      default_i_val = 1 ! AUTO (ELPA-kernel)
+#if defined(__ELPA)
+      IF (m_cpuid() .LE. MACHINE_X86) THEN
+         SELECT CASE (m_cpuid ())
+         CASE (MACHINE_X86_SSE4)
+            default_i_val = 8 ! ELPA_2STAGE_REAL_SSE_BLOCK4
```
oschuett: Currently, the enumeration of the ELPA kernels is fully encapsulated in `cp_fm_elpa.F`. Maybe we should keep it that way and add a routine `get_elpa_kernel_index_for_name()` ?

hfp: Good idea!  I also need to look at PR #329....

I will fix the name of the environment variable to match the input spec./keyword. If it does not outgrow, I will also check if the keyword parser can be reused for the environment variable. Later, I think it's worth to check if every global keyword can be optionally accompanied by an environment variable of the same name.

dev-zero: @hfp should I go ahead with merging PR #329 or do I wait for yours?

hfp: Please go ahead!

hfp:
I have adjusted a number of tolerances as part of this PR, but I left some to be adjusted later or to take an action on top of it. I think even when the tolerance goes up, it is hard to judge whether the previous or the new result is more correct.

oschuett:
In general, I think it would be better to adjust the reference values instead of the tolerances. However, I'm a bit surprised that CDFT is so disproportionately affected. Maybe it hits a broken corner case in ELPA? @nholmber, have you ever used CDFT together with ELPA?

dev-zero:
interestingly when I tried SCALAPACK-2's SYEVR I also broke almost all cdft tests, so maybe there's something hidden there

hfp:
@dev-zero @oschuett Shall I merge and then let people pickup the wrong results?

Alternatively, I can follow Ole's suggestion and replace reference values (and keep margins).

dev-zero:
I don't think that changing reference values helps. With this merged we may possibly see a lot of different values due to different configuration of regtesters (both the toolchain used to build them and the different architectures). Or do I miss something?

hfp:
Indeed, I missed to remind myself about the different configurations.

juerghutter:
If the reference values change when (not) using ELPA, we need to have independent tests for the
two cases and have them conditioned with ELPA available/not available.

nholmber:
Sorry for the delayed response. I regularly used ELPA with CDFT and did not encounter any accuracy issues, although I must admit when I first started using ELPA I only tested that ELPA produces the same total (and relative) energy as ScaLAPACK with regular OT and did not do explicit benchmarking with CDFT.

I think the main problem here is that the CDFT tests are by construction quite fragile. In order to run them quickly, all convergence thresholds had to be set to very loose values. Even tiny changes in the electron density will quickly accumulate and cause a noticeable difference in the CDFT energy, which is the tested quantity in most tests. This error is amplified in tests that are restarted from an earlier CDFT calculation (e.g. the tests with *mixed* in the name). As a result, I had to manually tweak the convergence thresholds in order to have the tests passing on all machines.

I fully expect CDFT with ScaLAPACK/ELPA to produce identical results when the convergence criteria are tightened, but I no longer have access to a machine with CP2K installed to confirm this.

hfp:
@oschuett Would this be a feature of the CI infrastructure to test with different reference values based on configuration?

Another option could be an error norm which is more robust. Both absolute and relative error measures have draw backs. I wonder if the relative Froebenius norm (even for scalars) is more robust. What's needed is a norm that blames issues during development. I guess matching a published scientific reference value is a different topic than identifying a faulty commit.

oschuett:
> Would this be a feature of the CI infrastructure to test with different reference values based on configuration?

We already [condition](https://github.com/cp2k/cp2k/blob/master/tests/TEST_DIRS) some test on the presents of certain flags. It should be straightforward to also allow for conditions on the absence of a flag.

hfp:
Regarding error norms, two border cases: (1) a relative error of a reference value close to zero is somewhat emphasized since one devides by a small number, and (2) an absolute error for a relatively large reference can be absolutely large but relatively fine.

What norm is used in CP2K's regtests?

oschuett:
> What norm is used in CP2K's regtests?
https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest#L312

However, I'm not sure if changing the norm will help. In the end some of our tests are unstable and I'd prefer an explicit solution, like Jürg suggested.

oschuett:
I'm still pondering this. Our current tolerances work well across the diverse set of tests on the dashboard. And in my understanding ELPA should compute the same eigenvectors and values up to machine precision. So, why is a test like `QS/regtest-ot/He2_none_full_all.inp` that used to be super robust suddenly so badly affected?

hfp:
I can believe that "values up to machine precision" may not be meaningful for different algorithms (i.e., ELPA vs reference version of ScaLAPACK). If both algorithms would be iterative, then they should lead to the same result otherwise finite precision can cause different results. Perhaps it would be interesting to know the condition number of the matrix of the system in case of these larger differences. An approximation could be the largest angle between all reported Eigenvectors (sort of dot product).

oschuett:
I realized that we don't have to extent the logic in `do_regtest`. We can simply force test to use Scalapack by explicitly setting `GLOBAL%CP2K_PREFERRED_DIAG_LIBRARY`. Hence, I'd propose that we pin the unstable test to Scalapack. We could still copy some of the tests into a separate directory which we condition on ELPA being present.

hfp:
@oschuett Thank you! I think this is a good idea. I will revert some of the adjusted tolerances and instead pin the tests in question to ScaLAPACK.

hfp:
I am getting the following remaining WRONG results (due to exceeded tolerances). However, the inputs causing these WRONG results are already carrying `PREFERRED_DIAG_LIBRARY SL` as part of the GLOBAL section. I wonder if the errors appear even with ScaLAPACK. Note, all other wrong results are fixed by having added the preferred diagonalization library to the input file (mainly CDFT tests).

```text
/workspace/cp2k/regtesting/local/pdbg/TEST-local-pdbg-2019-06-19_12-33-14/QS/regtest-cdft-hirshfeld/HeH-cdft-1.inp.out : 
 Current value of constraint  : ref = 1.135281644085 new = 1.135279896486  
 relative error :   1.53935519e-06 >  numerical tolerance = 3e-11  
/workspace/cp2k/regtesting/local/pdbg/TEST-local-pdbg-2019-06-19_12-33-14/QS/regtest-cdft-hirshfeld/HeH-cdft-2.inp.out : 
 Current value of constraint  : ref = 1.681172869379 new = 1.681171330858  
 relative error :   9.15148249e-07 >  numerical tolerance = 2e-11  
/workspace/cp2k/regtesting/local/pdbg/TEST-local-pdbg-2019-06-19_12-33-14/QS/regtest-cdft-hirshfeld/HeH-cdft-3.inp.out : 
 Current value of constraint  : ref = 1.197176189824 new = 1.197173595146  
 relative error :   2.16733648e-06 >  numerical tolerance = 4e-08  
/workspace/cp2k/regtesting/local/pdbg/TEST-local-pdbg-2019-06-19_12-33-14/QS/regtest-cdft-hirshfeld/HeH-cdft-4.inp.out : 
 Current value of constraint  : ref = 1.567706385953 new = 1.567705512012  
 relative error :   5.57465030e-07 >  numerical tolerance = 2e-11  
```

oschuett:
>  Apparently the (optional) PSMP-target does not run too often for GitHub based regtests.

The dashboard test should run twice per day, and within a PR it will run whenever triggered.

hfp:
@oschuett I wonder why these [4 tests](https://github.com/cp2k/cp2k/pull/322#issuecomment-503559368) are still failing (`pdbg` and `psmp`). I changed the input files to `PREFERRED_DIAG_LIBRARY SL`.

oschuett:
The `HeH-cdft-?.inp` tests are failing because they restart from the wavefunction created by `HeH-noconstraint.inp`. Hence, pinning that test to Scalapack too should do the trick. And then you can also keep its tight threshold.

src/input_cp2k_global.F
```diff
@@ -99,28 +106,41 @@ SUBROUTINE create_global_section(section)
...
+                          "on homogeneous architectures", &
                           usage="BLACS_REPEATABLE", &
                           default_l_val=.FALSE., lone_keyword_l_val=.TRUE.)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
+#if defined(__ELPA)
+      default_i_val = do_diag_elpa
+      ! TODO (issue 391): parse e.g. CP2K_PREFERRED_DIAG_LIBRARY="SL" or CP2K_PREFERRED_DIAG_LIBRARY="ELPA"
+      CALL GET_ENVIRONMENT_VARIABLE("CP2K_DEFAULT_DIAG", env, STATUS=env_i)
```
oschuett: As discussed before, I'm very skeptical of using environment variables to overwrite the input.

hfp: The isuee is that ELPA sometimes fails in decomposing the problem. I think you are right with this particular environment variable. It sort of a quick toggle to check whether an eventual issue is caused by ELPA. On the other hand one could just set `P2K_PREFERRED_DIAG_LIBRARY SL` in the input. I will remove this snippet, but I still think a general solution for runtime settings like mapping GLOBALs to environment variables is desirable.

juerghutter: I think a hierarchy of the type (from strong to weak) for variables in GLOBAL is possible

1) Variable set in input
2) Variable set in environment
3) default set in code

dev-zero: Personally I would use the following order since it's what most open source tools do in my experience:

1. Command line argument (if any, I think it would be very useful)
2. Variable set in environment
3. Variable set in input (configuration file)
4. Default by code

hfp:
Btw, it seems I cannot start the `pdbg` or `psmp` test. Perhaps I am not patient enough, but it looks like it terminates without doing anything. Maybe it determined that it does not need to run...

hfp:
It looks like it determines whether or not to run (manually selected) tests based on the most recent changeset (diff) rather than considering the entire changset of the PR.

oschuett:
Tests like `pdbg` have to be triggered manually by clicking on the blue `Run test` button. Each click triggers a single check run for the latest commit in the PR. If you want to run the test again after pushing more commits then you'll need to click again.

hfp:
That's what I did. For instance if you attempt to repeat this for `psmp` now, you can still see "ran 4 hours ago in less than 5 seconds". The point is I (manually) triggered the tests (both psmp and pdbg) but they terminated immediately ("[...] in less than 5 seconds").

src/input_cp2k_global.F
oschuett: I agree with Jürg. The Input files should have the highest priority, because it is not a configuration file, it's our main UI.

I think, instead of environments variables we should do more PR's like this which uses the hierarchy:
1. Variable set in input
2. Default overwritten by ARCH file flags
3. Default by code

oschuett:
You are right. There are seems to be some network issues between GCP and Github today. Your click event got dropped because of this. Just try again. The "ran ... in less than 5 seconds" message is a bit unfortunate. If the test result is neutral then the test did not actually run. I simply don't know any other way of implementing optional test within the Github API.

PR #325, opened 2019-04-25
base/machine: m_cpuid() function to test cpu/architecture/features

hfp:
Implemented m_cpuid() function, which allows to test for CPU/architecture/features. Instead of testing for certain bits using bit-ops, a range is checked. This is more manageable also in the light of multiple different CPU architectures supported by CP2K (e.g., ARM, Power, etc.).

A fully fledged CPUID functionality is typically based on feature bits for a single vendor. However, this would quickly exhaust the number of bits of an integer number (or make the signature of such function complicated aka vendors-id, etc). Though, CP2K needs to support multiple CPU architectures/vendors. Moreover, such fine granular feature bits would be too much for a scientific code to specialize for. Hence this PR proposes an easy to use enumeration that can be used for CPU, architecture as well as features such as instruction set extensions. A feature would be tested on a range given by lower/upper bound.

### DIFF
--- a/src/base/machine.F
+++ b/src/base/machine.F
@@ -199,7 +199,9 @@ PURE FUNCTION m_cpuid() RESULT (cpuid)
 #endif
     INTEGER :: cpuid
 #if defined(__LIBXSMM)
-    cpuid = MIN(MACHINE_X86_SSE4 + libxsmm_get_target_archid() - LIBXSMM_X86_SSE4, MACHINE_X86)
+    cpuid = libxsmm_get_target_archid()
+    cpuid = MERGE(MIN(MACHINE_X86_SSE4 + cpuid - LIBXSMM_X86_SSE4, MACHINE_X86), &
+                  MACHINE_CPU_GENERIC, LIBXSMM_X86_SSE4 .LE. cpuid)
 #else
     cpuid = m_cpuid_static()
 #endif
--- a/src/base/machine.F
+++ b/src/base/machine.F
@@ -33,7 +33,7 @@ MODULE machine
   INTEGER, PUBLIC, PARAMETER                   :: default_output_unit = output_unit, &
                                                   default_input_unit  = input_unit
 
-  ! Enumerates the target architecture or instruction set extension.
+  ! Enumerates the target architectures or instruction set extensions.
   ! A feature is present if within range for the respective architecture.
   ! For example, to check for MACHINE_X86_AVX the following is true:
   ! MACHINE_X86_AVX <= m_cpuid() and MACHINE_X86 >= m_cpuid().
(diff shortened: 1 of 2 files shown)


hfp:
Implemented m_cpuid_static (companion of m_cpuid). With `m_cpuid_static()` and `m_cpuid()`, CP2K's startup information can be extended (later) and warn if m_cpuid is beyond m_cpuid_static (the binary is not best-targeting the machine it runs on). Suppressed compiler-internal target-flags used in m_cpuid_static (e.g., \_\_AVX\_\_). Sorted conventions.supp (`LC_ALL=C sort tools/conventions/conventions.supp`).

src/base/machine.F
```diff
@@ -33,6 +33,25 @@ MODULE machine
...
+    MACHINE_X86_AVX     = 1001, &
+    MACHINE_X86_AVX2    = 1002, &
+    MACHINE_X86_AVX512  = 1003, &
+    MACHINE_X86 = MACHINE_X86_AVX512 ! marks end of range
+   ! other arch to be added as needed e.g.,
+   !MACHINE_ARM_SOME    = 2000
+   !MACHINE_ARM_ELSE    = 2001
+   !MACHINE_ARM = MACHINE_ARM_ELSE
+   !MACHINE_PWR_????    = 3000
+
```
dev-zero, line 54: personally I would prefer binary over decimal ranges, allowing for easier subdivision of ranges:

```fortran
  INTEGER, PARAMETER :: &
    MACHINE_CPU_GENERIC =    0, &
    MACHINE_X86         = z'1000', &
    MACHINE_X86_SSE4    = z'1001', &
    MACHINE_X86_AVX     = z'1010', &
    MACHINE_X86_AVX2    = z'1011', &
    MACHINE_X86_AVX512  = z'1012'
```

then the following would hold:

```fortran
IAND(m_cpuid(), MACHINE_X86) ! .TRUE. for any X86 flag
IAND(m_cpuid(), MACHINE_AVX) ! .TRUE. for any AVX flag
```

... assuming that a machine implementing AVX512 also implements AVX2 and AVX2, not sure whether that assumption is sensible.

But I guess you'd prefer to re-use what you introduced in `libxsmm` already.

hfp: Well, reusing the stuff from LIBXSMM is not the reason for the decimal ranges. The background is that INTEGER or INTEGER(8) only yield 32- or 64-bits respectively. If we go for binary flags we perhaps exhaust the number of bits. Remember, the bits are shared between Intel, ARM, Power and what not. Changing the design to not rely on an INTEGER on the other hand may reduce the usability of cpuid functions. Though, bits are inefficient in this design since a certain say feature on ARM may not appear at the same time as e.g. Intel AVX (nevertheless both "features" would need distinct bits/flags).

PR #327, opened 2019-04-28
added calculation of the sqrt of the overlap matrix and the density matrix via the general p-th order root-method from Richters et al.

robertschade:


src/dm_ls_scf_create.F
```diff
@@ -346,8 +354,16 @@ SUBROUTINE ls_scf_init_read_write_input(input, ls_scf_env, unit_nr)
...
-         WRITE (unit_nr, '(T2,A,T61,L20)') "Computing s preconditioner ", ls_scf_env%has_s_preconditioner
-         WRITE (unit_nr, '(T2,A,T61,I20)') "sign sqrt order:", ls_scf_env%sign_sqrt_order
+         WRITE (unit_nr, '(T2,A,T61,L20)') "Computing S preconditioner ", ls_scf_env%has_s_preconditioner
+
+         SELECT CASE (ls_scf_env%s_sqrt_method)
+         CASE (ls_s_sqrt_ns)
+            WRITE (unit_nr, '(T2,A,T61,A20)') "S sqrt method:", "NEWTONSCHULZ"
+         CASE (ls_s_sqrt_proot)
+            WRITE (unit_nr, '(T2,A,T61,A20)') "S sqrt method:", "PROOT"
+         END SELECT
```
oschuett, line 361: Please add a default case:

```
CASE DEFAULT
   CPABORT("Unkown sqrt method.")
END SELECT
```

src/dm_ls_scf_methods.F
```diff
@@ -114,15 +119,23 @@ SUBROUTINE ls_scf_init_matrix_S(matrix_s, ls_scf_env)
...
          CALL dbcsr_create(ls_scf_env%matrix_s_sqrt, template=ls_scf_env%matrix_s, &
                            matrix_type=dbcsr_type_no_symmetry)
          CALL dbcsr_create(ls_scf_env%matrix_s_sqrt_inv, template=ls_scf_env%matrix_s, &
                            matrix_type=dbcsr_type_no_symmetry)
-
-         CALL matrix_sqrt_Newton_Schulz(ls_scf_env%matrix_s_sqrt, ls_scf_env%matrix_s_sqrt_inv, &
-                                        ls_scf_env%matrix_s, ls_scf_env%eps_filter, &
-                                        ls_scf_env%sign_sqrt_order, &
-                                        ls_scf_env%eps_lanczos, ls_scf_env%max_iter_lanczos)
+         IF (ls_scf_env%s_sqrt_method .EQ. ls_s_sqrt_ns) THEN
```
oschuett: I think a `SELECT CASE` with a `DEFAULT CASE` is a better fit than `IF ELSE` here.

src/dm_ls_scf_methods.F
```diff
@@ -467,7 +483,12 @@ SUBROUTINE density_matrix_sign_fixed_mu(matrix_p, trace, mu, matrix_ks, matrix_s
 
       ! compute sign(inv(S)*H-I*mu)
       CALL dbcsr_create(matrix_sign, template=matrix_s, matrix_type=dbcsr_type_no_symmetry)
-      CALL matrix_sign_Newton_Schulz(matrix_sign, matrix_sinv_ks, threshold)
+      SELECT CASE (sign_method)
+      CASE (ls_scf_sign_ns)
+         CALL matrix_sign_Newton_Schulz(matrix_sign, matrix_sinv_ks, threshold)
+      CASE (ls_scf_sign_proot)
+         CALL matrix_sign_proot(matrix_sign, matrix_sinv_ks, threshold)
+      END SELECT
```
oschuett, line 498: Again, please add a `DEFAULT CASE`.

src/iterate_matrix.F
```diff
@@ -982,6 +1083,268 @@ SUBROUTINE matrix_sqrt_Newton_Schulz(matrix_sqrt, matrix_sqrt_inv, matrix, thres
...
+      res = PRODUCT((/(i, i=1, n)/))
+
+   END FUNCTION factorial
+! **************************************************************************************************
+!> \brief ...
+!> \param n ...
+!> \param k ...
+!> \return ...
+! **************************************************************************************************
+   FUNCTION choose(n, k) RESULT(res)
```
oschuett: This function is used only once. Maybe simply inline it?

src/iterate_matrix.F
```diff
@@ -982,6 +1083,268 @@ SUBROUTINE matrix_sqrt_Newton_Schulz(matrix_sqrt, matrix_sqrt_inv, matrix, thres
       CALL timestop(handle)
 
    END SUBROUTINE matrix_sqrt_Newton_Schulz
+! **************************************************************************************************
+!> \brief ...
+!> \param n ...
+!> \return ...
+! **************************************************************************************************
+   FUNCTION factorial(n) RESULT(res)
```
oschuett: This one-liner function is used only three times, hence inlining would be shorter.

src/dm_ls_scf_create.F
```diff
@@ -208,9 +210,15 @@ SUBROUTINE ls_scf_init_read_write_input(input, ls_scf_env, unit_nr)
...
       CALL section_vals_val_get(ls_scf_section, "PERFORM_MU_SCAN", l_val=ls_scf_env%perform_mu_scan)
       CALL section_vals_val_get(ls_scf_section, "PURIFICATION_METHOD", i_val=ls_scf_env%purification_method)
+      CALL section_vals_val_get(ls_scf_section, "SIGN_METHOD", i_val=ls_scf_env%sign_method)
       CALL section_vals_val_get(ls_scf_section, "DYNAMIC_THRESHOLD", l_val=ls_scf_env%dynamic_threshold)
       CALL section_vals_val_get(ls_scf_section, "NON_MONOTONIC", l_val=ls_scf_env%non_monotonic)
-      CALL section_vals_val_get(ls_scf_section, "SIGN_SQRT_ORDER", i_val=ls_scf_env%sign_sqrt_order)
+      CALL section_vals_val_get(ls_scf_section, "S_SQRT_METHOD", i_val=ls_scf_env%s_sqrt_method)
+
+      CALL section_vals_val_get(ls_scf_section, "SIGN_SQRT_ORDER", i_val=ls_scf_env%s_sqrt_order, explicit=explicit)
+      IF (.NOT. explicit) THEN
```
oschuett: This fall back might confuse some people as it's not documented. Maybe a single `SQRT_ORDER` keyword is sufficient? Alternatively, I'd propose a third keyword `DEFAULT_SQRT_ORDER`.

oschuett:
Thanks for this PR, nice work!
Maybe you could a few regression tests?

robertschade:
What kind of regression tests are you thinking of? The included regtests worked without problems.

src/dm_ls_scf_create.F
robertschade: Up to now SIGN_SQRT_ORDER has been used for the sqrt of the overlap matrix.  
However, I have now included the calculation of the sqrt of the overlap matrix with the pth-root method of Richters which does not consider the sign function. That is why I wanted to rename SIGN_SQRT_ORDER it to S_SQRT_ORDER. However, some of the regtests contain SIGN_SQRT_ORDER and maybe also some in-the-wild input files. Hence, I included the above backward compatibility. If it is ok to just rename parameters in the code and the regtest, I could also simply do that instead.

robertschade: Thomas Kühne has suggested the alias-functionality for the backwards compatibility. I was not aware of this before.

oschuett: Yes, the alias is a very nice solution!

We don't have any hard rules for backwards compatibility as it's always a judgment call. Changing less prominent keywords, like `SIGN_SQRT_ORDER`, it's generally ok.

oschuett:
Thanks for adding the test and I see they have already proven useful :-)
Unfortunately, they take quite some time to run. Most of our tests finish within 5 seconds and I'm trying to make 30s a [hard limit](https://github.com/cp2k/cp2k/issues/174). Have a look at the other `regtest-dm-ls-scf-*` tests, they use either less atoms or e.g. DFTB.

PR #328, opened 2019-04-28
Toolchain: fix avx512 detection in install_elpa.sh 

robertschade:
(check for avx512f instead of avx512)


hfp:
Makes sense to me. The check greps for AVX-512 foundational instructions (AVX512F) rather than grep'ing all AVX-512 extensions. Regarding ELPA, AVX-512F is what's referred to by the "AVX512-kernel".

PR #329, opened 2019-04-29
Feature/cleanup library support

dev-zero:
Given that we gain more options over time (and the combinatorial complexity they introduce) I think it is crucial to cut-down on old library support and make sure that what we provide is also tested.

As an alternative to removing the `dsyervx` implementation of `cp_fm_syevr`, we could also keep it as an implementation for `cp_fm_syevr` if `__SCALAPACK` is not set.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -126,7 +126,7 @@ Library ELPA for the solution of the eigenvalue problem
   * A version of ELPA can to be downloaded from http://elpa.rzg.mpg.de/software.
   * During the installation the libelpa.a (or libelpa_mt.a if omp active) is created.
   * Add `-D__ELPA=YYYYMM` to DFLAGS, where `YYYYMM` denotes the release date of the library.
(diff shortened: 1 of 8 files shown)


oschuett:
Thanks for cleaning this up :-)
You can silence the conventions checker's complain about `PDSYEVR`  [here](https://github.com/cp2k/cp2k/blob/master/src/fm/PACKAGE).

dev-zero:
@oschuett thanks. After some more discussion with @marci73 we're removing `cp_fm_syevr` completely (see 83e9de6). I tried to get it to work (https://github.com/dev-zero/cp2k/commit/37da162f68c25c4008d0c4759cabd7cc16a63b90), but in the end with ELPA outperforming it anyway it's not worth it.

hfp:
I will merge this PR now because otherwise the soon revised PR #322 will depend on it.

PR #331, opened 2019-04-30
m_cpuid_static (C impl.): there seems to be no way to determine target flags in Fortran code

hfp:
* Similar to other Fortran compilers, `gfortran -E -dM -mavx - < /dev/null | grep AVX` defines a variety of predefined macros (also similar to C).
* However, with a Fortran translation unit only a subset of these definitions disappears (`gfortran -E -dM -mavx my.F | grep AVX`).

To demonstrate and use `m_cpuid_static()` and `m_cpuid()`, a new (useful) CPWARN is introduced: "_the compiler target flags used to build this binary are insufficiently exploiting the extensions which are available for this CPU model_". The warning is shown after presenting the "CPU model name" (global environment output during CP2K's startup).


src/environment.F
```diff
@@ -800,6 +802,10 @@ SUBROUTINE read_global_section(root_section, para_env, globenv)
          CALL m_cpuinfo(model_name)
          WRITE (UNIT=output_unit, FMT="(T2,A)") &
             start_section_label//"| CPU model name : "//TRIM(model_name)
+         IF (m_cpuid_static() .LT. m_cpuid()) THEN
+            CPWARN("the compiler target flags used to build this binary are insufficiently"//
```
dev-zero: Can we make this a warning which does not increase the warning counter? (maybe introduce a new category `hints`?)
Since this will be triggered in heterogeneous environments and most likely any Linux distro (possibly as well as CP2K built with Spack since we're using `-mtune=native` there instead of `-march=...`), this will lead to people start ignoring the warning counter again and hamper automated calculations and output analysis since we have to filter by specific warnings again to separate real warnings which can significantly affect the result and mere hints (which don't).

hfp: Sure, will do!

hfp: Finished.

PR #341, opened 2019-05-07
Stabilize regtest-dm-ls-scf-4

oschuett:


robertschade:
Hi, why is a "stabilization" needed?

oschuett:
Because it fails for the [minimal](https://dashboard.cp2k.org/archive/minimal-sdbg/index.html) arch file.

robertschade:
Interesting, it worked for me and all other tests.
Let me see if I can reproduce the problem locally. IMHO, changing the SCF-steps is not the solution.

robertschade:
H2O-dft-ls-NEWTONSCHULZ.inp and H2O-dft-ls-DEFAULTS.inp test functionality (calculation of the sign function via a Newton-Schulz iteration) that has been in CP2K before 3be28de2c97549300b10b3aee1245e333069134f.
However, there was no regtest for this till now.

oschuett:
I don't think there is an actual problem with the code. It's more a matter of tuning the test such that they unaffected by numerical noise while not taking too long to run. Hence, I was trying out different `MAX_SCF` settings. The reason the minimal test gives a different results is probably because it's the only one without libxsmm.

robertschade:
To me the issue looks to be connected to fftw3 instead:
(gcc 8.3.0, fftw3 3.3.8)

WITHOUT FFTW3:
$ ./exe/local/cp2k.sdbg tests/QS/regtest-dm-ls-scf-4/H2O-dft-ls-NEWTONSCHULZ.inp | grep SCF
 ------------------------------ Linear scaling SCF ----------------------------- 
SCF     1       -17.031458259       -17.031458259    0.548953                                                                                                                                                                                                 
SCF not converged!                                                                                                                                                                                                                                           
$ ./exe/local/cp2k.sdbg -v                                                                                                                                                                                   CP2K version 7.0 (Development Version)                                                                                                                                                                                                                        Source code revision git:dd18ba31a                                                                                                                                                                                                                            cp2kflags: libint libxc xsmm libderiv_max_am1=5 libint_max_am=6 has_ieee_exceptions spglib json check_diag 

WITH FFTW3:
$ ./exe/local/cp2k.sdbg tests/QS/regtest-dm-ls-scf-4/H2O-dft-ls-NEWTONSCHULZ.inp | grep SCF
------------------------------ Linear scaling SCF -----------------------------                                                                                                                                                                              
SCF     1       -17.031439542       -17.031439542    0.484087                                                                                                                                                                                                 
SCF not converged!                                                                                                                                                                                                                                           
$ ./exe/local/cp2k.sdbg -v                                                                                                                                                                                   
CP2K version 7.0 (Development Version)                                                                                                                                                                                                                        
Source code revision git:dd18ba31a                                                                                                                                                                                                                            
cp2kflags: libint fftw3 libxc xsmm libderiv_max_am1=5 libint_max_am=6 has_ieee_exceptions spglib json check_diag

DCM-Uni-Paderborn:
I think it’s due to the usage of different FFT grids when using different 
FFT libraries in conjunction with a rather small PW cutoff. Why not using 
a larger PW cutoff and simply confining the number of SCF steps via 
MAX_SCF? 

Cheers, 
Thomas

robertschade:
That is a good point and fixes the problem.
With MAX_SCF=1 I get the table [out.txt](https://github.com/cp2k/cp2k/files/3154433/out.txt).
A cutoff of 240 Ry or 175 Ry produces identical results in minimal.sdbg and sdbg with fftw3.

robertschade:
This will fail, because the reference values also have to be modified.

PR #345, opened 2019-05-09
build system: set variable CXXOMPFLAGS needed by DBCSR

### LINKED ISSUE #338: [toolchain] does not set `CXXOMPFLAGS` needed by DBCSR
The arch file generated by the toolchain does not set the variable `CXXOMPFLAGS` (C++ OpenMP flags). As a result, DBCSR's C++ code is silently (!) compiled without OpenMP support (see line https://github.com/cp2k/dbcsr/blob/develop/Makefile#L500), leading to race conditions that are **very** difficult to debug.

shoshijak:
If omitted, DBCSR's C++ code gets compiled without OpenMP support, leading to race conditions (observed on OLCF's Summit). 

Add this variable to the toolchain as well as to the relevant arch files.

Fixes #338

### DIFF
--- a/arch/Linux-x86-64-cuda.sopt
+++ b/arch/Linux-x86-64-cuda.sopt
@@ -26,6 +26,7 @@ DFLAGS   = -D__CUDAPW -D__FFTW3
 CFLAGS   =  -O2
 CPPFLAGS = -traditional -C $(DFLAGS) -P -I/opt/intel/mkl/10.0.1.014/include/fftw
 CXXFLAGS = -O3 -I${CUDA_PATH}/include -std=c++11
+CXXOMPFLAGS = -fopenmp
 NVFLAGS	 = $(DFLAGS) -Xcompiler='-fopenmp' --std=c++11
 FCFLAGS  = $(DFLAGS) -O2 -xW
 MKLPATH  = /opt/intel/mkl/10.0.1.014/lib/em64t/
--- a/arch/Linux-x86-64-cuda.sopt
+++ b/arch/Linux-x86-64-cuda.sopt
@@ -18,7 +18,7 @@ CXX      = CC
 CPP      = cpp
 NVCC	 = nvcc
 FC       = /opt/intel/fce/10.0.025/bin/ifort -FR
-LD       = /opt/intel/fce/10.0.025/bin/ifort -i-static -openmp
+LD       = /opt/intel/fce/10.0.025/bin/ifort -i-static
 AR       = ar -r
 GPUVER   = K20X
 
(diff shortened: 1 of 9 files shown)


arch/Linux-x86-64-cuda.sopt
```diff
@@ -26,6 +26,7 @@ DFLAGS   = -D__CUDAPW -D__FFTW3
 CFLAGS   =  -O2
 CPPFLAGS = -traditional -C $(DFLAGS) -P -I/opt/intel/mkl/10.0.1.014/include/fftw
 CXXFLAGS = -O3 -I${CUDA_PATH}/include -std=c++11
+CXXOMPFLAGS = -fopenmp
 NVFLAGS	 = $(DFLAGS) -Xcompiler='-fopenmp' --std=c++11
```
dev-zero: Can someone else comment on this one: in my understanding should `sopt` never use threads, yet here it seems that `.cu`'s were built with `-fopenmp`. Is that intentional?

And I must have missed it, but why do we need an extra flag `CXXOMPFLAGS` in DBCSR and can't use `CXXFLAGS`?

shoshijak: > And I must have missed it, but why do we need an extra flag CXXOMPFLAGS in DBCSR and can't use CXXFLAGS?

Good question, now that you ask, I can't think of a good reason. `CXXFLAGS` appears in 2 rules in the Makefile: the [general rule for compiling cpp files](https://github.com/cp2k/dbcsr/blob/develop/Makefile#L497) and [the rule for compiling libcusmm.cpp](https://github.com/cp2k/dbcsr/blob/develop/Makefile#L500). Perhaps `CXXFLAGS` and `CXXOMPFLAGS` are separate so that `examples/dbcsr_example_3.cpp` doesn't get compiled with `-fopenmp` ?

alazzaro:
Tiziano is 100% right, we don't need two variables and openmp has to be used only for ssmp and psmp versions. Sorry, this is due to my fault, I have not checked it from the beginning...

shoshijak:
okay thanks, I'll fix both these issues (superflous `CXXOMPFLAGS` variable and use openmp only in ssmp and psmp versions)

shoshijak:
I just realised I cannot get rid of the superfluous `CXXOMPFLAGS` as a part of this PR, because it would break compilation of DBCSR as it currently is.

Once this PR passes review, I suggest we merge it and deal with the `CXXOMPFLAGS` in a different PR.

alazzaro:
I see the problem, indeed it requires a new DBCSR release. Yes, I think it is a good idea to merge this PR and then wait for a new DBCSR release and clean-up the archfiles.

dev-zero:
please open an issue prior to merging to make sure it gets resolved eventually, these kind of things have a tendency to fall of the table...

shoshijak:
You're right, I just opened an issue at https://github.com/cp2k/cp2k/issues/350

shoshijak:
Are any more changes required to this PR? 
If not, could someone merge it?

shoshijak:
Thank you!

PR #356, opened 2019-05-13
[RFC] OO Implementation of SCF Inner Loop/Diagonalization

dev-zero:
This is a Request for Comments and while it compiles it likely contains bugs and subtle changes.

The idea here is to try an Object-oriented approach to structure part of the SCF Inner Loop with the goal to increase cohesion of code, meaning that everything related to a diagonalization method (or concretely the `new_mos` routine) is isolated in a single class, rather than distributed between multiple functions (resulting in a single `CASE`-switch before entering the SCF cycle).

In principle a `new_mos` class has to bring the following:

* a `init` routine, which is currently called before entering the outer SCF
* a `pre` routine which gets called at the beginning of each inner SCF and can be used to (re-)generate the preconditioner (hence `pre`)
* a `run` routine which does the actual diagonalization

### DIFF
--- a/src/qs_scf.F
+++ b/src/qs_scf.F
@@ -50,8 +50,7 @@ MODULE qs_scf
    USE cp_files,                        ONLY: close_file
    USE cp_fm_types,                     ONLY: cp_fm_create,&
                                               cp_fm_release,&
-                                              cp_fm_to_fm,&
-                                              cp_fm_type
+                                              cp_fm_to_fm
    USE cp_log_handling,                 ONLY: cp_add_default_logger,&
                                               cp_get_default_logger,&
                                               cp_logger_release,&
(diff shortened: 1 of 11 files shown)


oschuett:
First of all, I think this is really moving in the right direction. Thanks @dev-zero for pioneering OOP in CP2K!

The obvious concern is compilers. GCC4.8 bails out immediately. In last year's survey 20% were still using it. In the meantime it has probably dropped to a reasonable 10%. Nevertheless, I'd expect subtle issues with the later GCC versions too, as we have seen them e.g. with ALLOCATABLES. And finally there is also Intel, Cray, and IBM....

The code itself does look really nice. So, it's definitely worth the effort. I'm not sure about the camel casing. It certainly feels familiar, but since Fortran is case insensitive the names will probably not show up properly in compiler messages. Hence, I'd propose to extend our existing naming conventions and use `mo_calc_class` instead of `AbstractMOCalc`. I'd also drop the `Abstract` from the name because the consuming code shouldn't have to care. In Java this would have been an interface.

pseewald:
> The obvious concern is compilers. GCC4.8 bails out immediately. In last year's survey 20% were still using it. In the meantime it has probably dropped to a reasonable 10%. Nevertheless, I'd expect subtle issues with the later GCC versions too, as we have seen them e.g. with ALLOCATABLES. And finally there is also Intel, Cray, and IBM....

I'd be in favour of dropping support for old compilers, CP2K developers should be encouraged to use modern Fortran features, should use new compiler versions and should not be hindered by old compilers. Support for object orientation should be fairly complete for all mentioned compilers:
[Fortran 03 status](http://fortranwiki.org/fortran/show/Fortran+2003+status)
[Fortran 08 status](http://fortranwiki.org/fortran/show/Fortran+2008+status)

oschuett:
Unfortunately, this will not only affect the developers but our user base at large. According to the survey 80% compile CP2K by themselves and many are stuck with dated compilers.

That being said, I do believe that the positive impact OOP will have on code health outweighs the costs. Basically, I'm just trying to raise some awareness (and commitment) for what we are about to get ourselves into.

oschuett:
Since we've dropped support for GCC 4.8, can this PR now proceed?

oschuett:
I actually like this PR a lot. Any chance you could revive it @dev-zero?

dev-zero:
Yes.

oschuett:
I contemplated salvaging this PR, because I really like the approach and think we should use more OOP to disentangle our high level code. However, there have been a lot of [changes]( https://github.com/cp2k/cp2k/commits/master/src/qs_scf.F) since and this PR was never complete to begin with.

So I'm now going to close this PR with sorrow.

Issue #83, opened 2018-11-06
Improving Python Tests


oschuett:
Currently all our Python scripts are tested by calling them with `--selftest`. Unfortunately, virtually no script actually implements a proper self test. Still this test is useful as it verifies that syntax and imports work across different versions of Python.

I'd propose that in the future we require all scripts to support `-h` instead. Furthermore, scripts that match `*_unittest.py` are called again without any arguments.

oschuett:
Sure `*_test.py` is fine too. This way it stays close to the script it's testing.

Just for completeness: The original goal of the Python dashboard test was to ensure that our build scripts work across different versions of Python.

dev-zero:
The proper way is to use something like `check-ast` from https://github.com/pre-commit/pre-commit-hooks

Issue #78, opened 2018-11-04
add linters for Shell scripts and Python files


oschuett:
A growing fraction of our code are bash scripts. Hence, we should look into using a linter like 
[ShellCheck](https://github.com/koalaman/shellcheck).

dev-zero:
I also started using `pylint`/`flake8` for the Python scripts, I will come up with a `pre-commit` configuration for all of it.

oschuett:
Yes, flake8 is a good idea. While you build the pre-commits, I'd add this to the CI-system. However, it will probably take a while until all our scripts are made to pass.

dev-zero:
@oschuett you could then use the `pre-commit run -a` feature in the CI

oschuett:
Yes, we should re-use your pre-commit scripts for the CI. I would like to run the flake8 as part of the existing [python test](https://dashboard.cp2k.org/archive/python/index.html) and create a similar test for shell scripts.

oschuett:
We are now using `shellcheck` and `mypy` for new scripts.

PR #357, opened 2019-05-13
Replace selftest with either proper unittests or check-ast from pre-commit

dev-zero:
@oschuett it would be great if we could instead integrate the `pre-commit run --all-files` in the mandatory checks (similar to what we're doing [in DBCSR](https://github.com/cp2k/dbcsr/blob/develop/.travis.yml))


oschuett:
Yes, this looks good. I'm also linking #78 and #83 where we already discussed this a bit.
For a new CI check we basically only need a suitable [Dockerfile](https://github.com/cp2k/cp2k/blob/master/tools/docker/).

dev-zero:
@oschuett I've adapted the present `python_test` Dockerfile and scripts to run `pre-commit`, and with that the `check-ast`.

tools/docker/scripts/test_python.sh
```diff
@@ -19,7 +19,7 @@ function run_selftests {
         echo "Selftesting $i"
         cd "$(dirname "$i")"
         SCRIPT=$(basename "$i")
-        if ! $PYTHON -B "./${SCRIPT}" --selftest ; then
+        if ! $PYTHON -B "./${SCRIPT}" ; then
             echo "Selftest of $i failed"
```
oschuett: better `Test $i failed.`

tools/docker/scripts/test_python.sh
```diff
@@ -28,32 +28,42 @@ function run_selftests {
     done
 }
 
+function run_pre_commit {
+    set +e #disable error trapping
+    if ! pre-commit run --all-files --hook-stage manual check-ast ; then
+        echo "check-ast for at least one Python version"
```
oschuett: better: `Syntax check failed (Python 2 or 3).`

oschuett:
Looks good! Btw, given the survey results I think it is safe to drop Python 2.6 now. Also since you reformatted everything, should we add a check for flake8 or similar?

dev-zero:
> Looks good! Btw, given the survey results I think it is safe to drop Python 2.6 now. 

Ok, done.

> Also since you reformatted everything, should we add a check for flake8 or similar?

I only touched `makedep.py` so far. If we're going to introduce that I'd probably use something like `black` + `flake8+flake8-bugbear` to do the reformatting first and then the checks afterwards.

tools/docker/scripts/install_python.sh
```diff
@@ -20,19 +22,4 @@ rm -rf /var/lib/apt/lists/*
```
oschuett, line 22: The `build-essential` should not be needed any more, and the `python-dev*` packages can probably go as well.

tools/docker/scripts/test_python.sh
```diff
@@ -19,41 +19,48 @@ function run_selftests {
         echo "Selftesting $i"
```
oschuett: Sorry, I overlooked this one:
`"Selftesting $i"`  ->  `"Running $i"`

oschuett:
Yeah, on second thought a Python formatting check could also be a part of the existing formatting check. But then again, maybe we simply combine the various Python, formatting, and eventually shellcheck into one single Precommit test?

> do the reformatting first and then the checks afterwards.

This I don't understand. We do want people to check in formatted code right? It does make sense when run locally as precommit in the literal sense, though.

tools/docker/scripts/install_python.sh
dev-zero: Possibly, but not sure: we're installing `numpy` via pip which may or may not need to build something (depending on whether a wheel is available or not for that distro+version of python).
I'd keep it for now in case they're dropping the wheels for Python 2 at some point.

dev-zero:
> Yeah, on second thought a Python formatting check could also be a part of the existing formatting check. But then again, maybe we simply combine the various Python, formatting, and eventually shellcheck into one single Precommit test?

Yeah, that's what I would do since the `pre-commit` hooks on the user side have all the knowledge already on which files actually changed which speeds things up quiet nicely.

> 
> > do the reformatting first and then the checks afterwards.
> 
> This I don't understand. We do want people to check in formatted code right? It does make sense when run locally as precommit in the literal sense, though.

Ideally, yes. What I meant with that is: I would employ two things here:

* `black` ... as the Python alternative to `prettify` to automatically reformat any Python code
* `flake8` with the `flake8-bugbear` plugin to check the code

tools/docker/scripts/test_python.sh
dev-zero: done, thanks

PR #358, opened 2019-05-13
Refactor MP2

fstein93:
I did some Refactoring in the MP2 code, introduced a new type and removed some debug code.


src/mp2_ri_grad_util.F
```diff
@@ -1665,17 +1668,23 @@ END SUBROUTINE create_dbcsr_gamma
...
 
       INTEGER                                            :: iii, jjj
 
       WRITE (*, *)
       DO iii = 1, SIZE(mat, 1)
          WRITE (*, *) iii
          DO jjj = 1, SIZE(mat, 2), 20
-            WRITE (*, '(1000F10.5)') mat(iii, jjj:MIN(SIZE(mat, 2), jjj+19))
+            IF (PRESENT(unitout)) THEN
+               WRITE (1000+unitout, '(1000F10.5)') mat(iii, jjj:MIN(SIZE(mat, 2), jjj+19))
```
dev-zero: is it intentional that only the inner loop output goes to `1000+unitout` and the empty lines and outer index still to the default?

fstein93: I don't really know. We should ask that the author of these lines (probably Mauro). The routine was part of the debug code.

fstein93:
@dev-zero The claimed issue is solved. Are there other comments?

alazzaro:
@fstein93 you probably better to mention who you are referring to (@dev-zero I assume)

tools/regtesting/do_regtest
```diff
@@ -21,7 +21,7 @@ skip_dirs[1]=""
 ndirstorestrict=0
 restrict_dirs[1]=""
 default_err_tolerance="1.0E-14"
-mpiexec="mpiexec --bind-to none"
+mpiexec="mpiexec --bind-to-none"
```
dev-zero: Care to elaborate? `--bind-to none` is what is documented by OpenMPI

fstein93: Sry, I use this locally and forgot to return to the correct version.

fstein93:
Is it possible to have an additional Regtest with pdbg?

alazzaro:
@fstein93 I've just triggered it...

alazzaro:
@fstein93 Should we merge it now?

fstein93:
From my side: yes.

PR #373, opened 2019-05-21
Fpga interface for pw fft computations

arjunramaswami:
[summary.txt](https://github.com/cp2k/cp2k/files/3202563/summary.txt)
[timings.txt](https://github.com/cp2k/cp2k/files/3202564/timings.txt)

Added APIs for computing single and double precision forward and backward fft3d
with intel fpgas set using `-D__PW_FPGA` and `-D__PW_FPGA_SP` flags. Tested for
FW_R3DC1D and BW_C1DR3D calculations for sizes 16^3, 32^3, 64^3 fft3d. If sizes
not found, falls back to CPU FFT computations.
Included host and kernel code that work with uniform fft3d sizes.

Attached files show the summary of regressions tests executed using the double 
precision fpga build.


alazzaro:
This is great! Thanks for your PR.
I have some questions (mainly curiosity)

1) how can we test this in the CI? Is there any way for you to add an entry in the dashboard? (https://dashboard.cp2k.org/)
2) What is the hardware that you are using and how it compares in performance to CPU?

dev-zero:
> This is great! Thanks for your PR.
> I have some questions (mainly curiosity)
> 
> 1. how can we test this in the CI? Is there any way for you to add an entry in the dashboard? (https://dashboard.cp2k.org/)
> 2. What is the hardware that you are using and how it compares in performance to CPU?

from the PR: `Currently tested only for Intel Stratix 10 and Arria 10 GX1150 FPGAs.`

dev-zero:
@arjunramaswami can you please add a note and possibly a compile-time check to make sure users are aware that `__PW_CUDA` and `__PW_FPGA` are mutually exclusive?

arjunramaswami:
> This is great! Thanks for your PR.
> I have some questions (mainly curiosity)
> 
> 1. how can we test this in the CI? Is there any way for you to add an entry in the dashboard? (https://dashboard.cp2k.org/)
> 2. What is the hardware that you are using and how it compares in performance to CPU?

@alazzaro I doubt this can be tested in the CI unless there is the above mentioned hardware or intel fpga emulation environment installed. I have added some instructions to the latter in a README file in the fpga folder if you are curious to test them.

alazzaro:
Sorry, I was not clear in my question about the dashboard, let me clarify.
Could you provide the hardware such that we can test and keep updated your PR for the dashboard?

https://www.cp2k.org/dev:dashboard#adding_a_tester

In principle, it is a cronjob which can run once a day (this is what we do for some of the tests), or even more than 1 day.

Anything about the performance question?

arjunramaswami:
Sorry for the late reply. 

> @arjunramaswami can you please add a note and possibly a compile-time check to make sure users are aware that `__PW_CUDA` and `__PW_FPGA` are mutually exclusive?

@dev-zero Will make that change in the next commit. Thank you

> Sorry, I was not clear in my question about the dashboard, let me clarify.
> Could you provide the hardware such that we can test and keep updated your PR for the dashboard?
> 
> https://www.cp2k.org/dev:dashboard#adding_a_tester
> 
> In principle, it is a cronjob which can run once a day (this is what we do for some of the tests), or even more than 1 day.
> 
> Anything about the performance question?

@alazzaro  I have been able to synthesize single precision FFTs and optimize them for the older Arria 10 board. The performance is surely not comparable to the state of the art CPU. However, the idea behind this work is to provide an interface for further FPGA development. I believe optimizations for the newer Stratix 10s would surely be interesting with respective to performance, which will be the next point of focus.

Regarding the hardware for the dashboard tests - we are looking into using Singularity containers with double precision emulation binaries, but it is too early and we hope to keep you updated on that.   
Oh and the reason for using emulation for double precision FFT is that DSP IP blocks were available only for the single precision floating point calculations to me. Double precision FFTs have been tested for correctness using emulation (attached with the summary report). Using single precision causes loss in precision to pass the regression tests.

arjunramaswami:
The conventions check seems to be failing partly due to the inclusion of the FPGA kernel code. Therefore, I plan to move them to a separate repository and add a link to it in some NOTE. I hope this is ok!

oschuett:
In general, I believe CP2K's mono-repo approach is a great defense against code rot. However, since your FPGA work is still at an very early stage maybe having the kernels in a separate repo is indeed the better solution for now.

Regarding the conventions:
 - You can fix the header flags by having the macro names [match](https://github.com/cp2k/cp2k/blob/master/tools/conventions/analyze_src.py#L67) the file names.
 - You can fix the other flags by moving the documentaion to [INSTALL.md](https://github.com/cp2k/cp2k/blob/master/INSTALL.md). As these are global flags you should probably prefix them, e.g. with `__PW_FPGA_...`
- The Altera OpenCL playground could be kept as an external dependency or reduced to what you actually need. In the later case you could add our usual CP2K copyright banner.

dev-zero:
> In general, I believe CP2K's mono-repo approach is a great defense against code rot. 

a) If everything relevant is behind a preprocessor definition which is never tested for, there is really nothing stopping it from rotting away.
b) if it is not tested and no test hardware is available to active developers, supporting it becomes almost impossible

> * The Altera OpenCL playground could be kept as an external dependency or reduced to what you actually need. In the later case you could add our usual CP2K copyright banner.

If some code is coming from a third-party, it has to be verified whether this part is under such a permissible license that even their original copyright can be stripped. Otherwise the copyright banner has to be augmented to reflect shared copyright.

alazzaro:
> @alazzaro I have been able to synthesize single precision FFTs and optimize them for the older Arria 10 board. The performance is surely not comparable to the state of the art CPU. However, the idea behind this work is to provide an interface for further FPGA development. I believe optimizations for the newer Stratix 10s would surely be interesting with respective to performance, which will be the next point of focus.

Anyway, this is great. I did a project with FPGA in a different context 10 years ago and it was a mess (VHDL, SystemC), so I'm glad that now it is OpenCL coming to play a role. Thanks for this work!

> Regarding the hardware for the dashboard tests - we are looking into using Singularity containers with double precision emulation binaries, but it is too early and we hope to keep you updated on that.
> Oh and the reason for using emulation for double precision FFT is that DSP IP blocks were available only for the single precision floating point calculations to me. Double precision FFTs have been tested for correctness using emulation (attached with the summary report). Using single precision causes loss in precision to pass the regression tests.

I know this problem, my personal opinion is that experimental code which doesn't support all options should be in a dev branch if you cannot go through all tests, or we can make an ALLERT warning somewhere, something like "This option is still experimental". I leave to the other developer to decide (I see that you got some comments already)...

arjunramaswami:
I agree that the code behind the preprocessor has to have a way to be tested. I just started working on a singularity container for the fpga code that could be added to the dashboard as well as be made available to those who would like to test out the fpga dev environment on their local machine. Hopefully this helps.

arjunramaswami:
There were some rogue vscode files in that last commit. Have to close and do another pull request. Sorry for that

PR #374, opened 2019-05-21
Rely on mpi_comm_set_errhandler instead of mpi_errhandler_set

hfp:
Rely on `mpi_comm_set_errhandler` instead of `mpi_errhandler_set`. The conditional compilation checks for MPIv3 (`__MPI_VERSION`) which is reasonable but may be not exactly related (see for instance [here](https://www.open-mpi.org/faq/?category=mpi-removed)). However, it is reasonable to assume that MPI function names are standardized since MPIv3 (portable MPI-ABI between different implementations of MPI). Fixed a few source code comments (typos).


hfp:
The `pdbg` variant was requested for test (to ensure the new code is actually tested).

hfp:
Apparently, it seems to be safe to merge this PR as https://github.com/cp2k/dbcsr/pull/188 also passes (in addition to my fork which is passing since weeks with this change).

PR #375, opened 2019-05-21
Rely on CONTIGUOUS keyword in MPI wrapper

hfp:
Since CP2K moved to F08, the CONTIGUOUS keyword can be used. MPI requires contiguous buffers hence the wrapper can safely make that statement. In fact, this fixes occurrences of "created temporary array" (unveiled with related GNU Fortran flags). This change can be taken further by additing more uses of the CONTIGUOUS even in the MPI wrapper module.


hfp:
The `pdbg` variant was requested for test (to ensure the new code is actually tested).

PR #382, opened 2019-05-24
Bug fix for LSD kernel, LDA Triplets, update of 6 regtests

juerghutter:
This is apparently a very old bug in CP2K, going back to 2004! It affected all calculations that
needed the LSD kernel functionals (e.g. TDDFT, linear response). I have in my TDDFT branch
a finite difference test code that was used to verify this patch.


mattatlincoln:
Thanks, Juerg!

PR #387, opened 2019-05-27
Dashboard: Enable notifications for everything

oschuett:
With the CI in place people pay less attention to the dashboard. Hence, this PR enables email notifications for all tests.

@alazzaro, is this what you had in mind?

### DIFF
--- a/tools/dashboard/dashboard.conf
+++ b/tools/dashboard/dashboard.conf
@@ -2,7 +2,6 @@
 sortkey:     10
 name:        Formatting
 host:        GCP
-notify:      on
 report_url:  https://storage.googleapis.com/cp2k-ci/dashboard_formatting_report.txt
 info_url:    http://www.cp2k.org/dev:formattingconventions
 
@@ -10,7 +9,6 @@ info_url:    http://www.cp2k.org/dev:formattingconventions
 sortkey:     20
 name:        Coding conventions
 host:        GCP
-notify:      on
 report_url:  https://storage.googleapis.com/cp2k-ci/dashboard_conventions_report.txt
 info_url:    http://www.cp2k.org/dev:codingconventions
 
@@ -248,7 +246,6 @@ report_url:  https://storage.googleapis.com/cp2k-ci/dashboard_python_report.txt
 sortkey:     630
 name:        Linux-x86-64-valgrind.sdbg
 host:        UZH, opt5
-notify:      on
 timeout:     48
 report_url:  http://www.cp2k.org/static/regtest/trunk/uzh-opt5/Linux-x86-64-valgrind.sdbg.out
 
--- a/tools/dashboard/generate_dashboard.py
+++ b/tools/dashboard/generate_dashboard.py
@@ -95,7 +95,7 @@ def gen_frontpage(config, log, status_fn, outdir):
         name        = config.get(s,"name")
         host        = config.get(s,"host")
         report_url  = config.get(s,"report_url")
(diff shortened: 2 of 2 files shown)


alazzaro:
Yes, this is a good starting point, thanks!
General question: How does it work with PR, I mean, where it gets the email, from github user?

oschuett:
The dashboard sends its notifications to the commit author email.
If this is still not noisy enough we could contemplate to also include the committer.

mkrack:
Some time ago I stopped sending automatic notifications to all committers when one of the PSI regtesters failed, since there was an increasing risk of false notifications. It is great that such notification can also be sent by the dashboard. We should however avoid spamming committers with redundant notification from all regtesters in case of an error. I got only mails from the valgrind tester so far. Does this commit activate notifications from all regtest by default in case of any failure?

oschuett:
Yes, I'm in the past the dashboard only send notifications for formatting, conventions, and Valgrind. With the CI the first two are now basically impossible to trigger.

There is indeed a risk of including the wrong persons in those emails. This happens when a test didn't run for a long time, because then it's impossible to tell which commit is at fault. However, most tests run at least twice per day. Hence, this shouldn't happen too often.

> We should however avoid spamming committers with redundant notification from all regtesters in case of an error.

Currently, the notifications are send whenever a tester finishes. I guess we could send them in aggregate once per day. Any other ideas?

alazzaro:
Spamming is not good, I agree. Some tests are running once per day (Daint and Swan), so I understand that it can spam several users... 

Then, let's try to be more experimental.
Is it possible for the dashboard to open an issue and assign it to the people? (instead of direct emails)

Assuming that there are no so many commits from different people (big PR from a single user?), I propose to go for this solution only the first time that any error appears:

1) Commit X --> OK
2) Commit Y --> Something wrong
3) Dashboard opens an issue (reporting problems and links)
4) Dashboard assigns to people involved between the X and Y commits.

I'm not sure if this is a lot of work to do...

alazzaro:
I would like to add that, if this is too much work to do for the implementation (automatic issue and assign on github), we can make it by hand (I volunteer for that). In this case, the dashboard should send to to some "managers" of the dashboard test the list of failures and the people behind the commits and then the managers can open and assign the issues. No spam and we keep track of the issue + we can keep track of the issue on the dashboard.

oschuett:
But isn't opening an issue even more noisy? It depends a bit on how people configured their Github account, but basically a new issue sends an email to everybody.

Note that, the Dashboard only sends a notifications when a test changes from OK to FAILED, ie. there are no subsequent emails for a tests that continues to fail.

I think enabling notifications for all tests is ok. Due to the CI the Dashboard has gotten a lot greener.

alazzaro:
OK, then fine for me. It is better than nothing ;)
In the future, we can consider the "manager" email, such that we can open issues (I still think this is a good idea to keep the problem alive).

oschuett:
Some time the dashboard had a column to show for each test related sourceforge issues. Back then it didn't really catch on because we didn't use the issue tracker. So, maybe we should try to revive that feature? Then one could quickly glance the dashboard and see which long lasting failures are lacking an issues.

mkrack:
@oschuett OK, let's try, we can always adjust things, if it becomes annoying.

oschuett:
Ok, then I'll merge and we'll see how it goes. I'm happy to tweak things along the way.

alazzaro:
Issue tracker column would be really good to have in on the dashboard, but let's go step by step...

PR #410, opened 2019-06-11
Fpga interface for pw fft computations

arjunramaswami:
Added APIs for computing single and double precision forward and backward fft3d
with intel fpgas set using -D__PW_FPGA and -D__PW_FPGA_SP flags. Tested for
FW_R3DC1D and BW_C1DR3D calculations for sizes 16^3, 32^3, 64^3 fft3d. If sizes
not found, falls back to CPU FFT computations.
Included host and kernel code that work with uniform fft3d sizes.

- included OpenCL Utils Library replacing Intel's
- opencl kernel files moved to external repo
- cuda and fpga are mutually exclusive

### DIFF
--- a/src/cp2k_info.F
+++ b/src/cp2k_info.F
@@ -136,6 +136,12 @@ CONTAINS
 #if defined(__PW_CUDA)
       flags = TRIM(flags)//" pw_cuda"
 #endif
+#if defined(__PW_FPGA)
+      flags = TRIM(flags)//" pw_fpga"
+#endif
+#if defined(__PW_FPGA_SP)
+      flags = TRIM(flags)//" pw_fpga_sp"
+#endif
 #if defined(__HAS_PATCHED_CUFFT_70)
       flags = TRIM(flags)//" patched_cufft_70"
 #endif
--- a/src/pw/fpga/fft_fpga.c
+++ b/src/pw/fpga/fft_fpga.c
@@ -21,10 +21,11 @@
 // Function prototypes
 int init();
 void cleanup();
-void init_program(int N[3], char *data_path);
-void queue_setup();
+static void cleanup_program();
+static void init_program(int N[3], char *data_path);
+static void queue_setup();
 void queue_cleanup();
-void fftfpga_run_3d(int inverse, int N[3], cmplx *c_in);
+static void fftfpga_run_3d(int inverse, int N[3], cmplx *c_in);
 
 // --- CODE -------------------------------------------------------------------
 
@@ -52,17 +53,29 @@ int pw_fpga_check_bitstream_(char *data_path, int N[3]){
         (N[0] == 32 && N[1] == 32 && N[2] == 32) ||
         (N[0] == 64 && N[1] == 64 && N[2] == 64)  ){
 
-        // if previously used binary needs to be executed again, otherwise set
-        // the new size and load new binary
-        if( fft_size[0] == N[0] && fft_size[1] == N[1] && fft_size[2] == N[2] ){
+        // if first time
+        if( fft_size[0] == 0 && fft_size[1] == 0 && fft_size[2] == 0 ){
+          fft_size[0] = N[0];
+          fft_size[1] = N[1];
+          fft_size[2] = N[2];
+
+          init_program(fft_size, data_path);
+        }
+        else if( fft_size[0] == N[0] && fft_size[1] == N[1] && fft_size[2] == N[2] ){
+          // if same fft size as previous
+          // dont do anything
         }
         else{
-            fft_size[0] = N[0];
-            fft_size[1] = N[1];
-            fft_size[2] = N[2];
+            // else if different fft size as previous
+            // cleanup and initialize
+          fft_size[0] = N[0];
+          fft_size[1] = N[1];
+          fft_size[2] = N[2];
 
-            init_program(fft_size, data_path);
+          cleanup_program();
+          init_program(fft_size, data_path);
         }
+
         return 1;
     }
     else{
@@ -77,9 +90,6 @@ int pw_fpga_check_bitstream_(char *data_path, int N[3]){
  * \param   din : complex input/output single precision data pointer 
  *****************************************************************************/
 void pw_fpga_fft3d_sp_(int direction, int N[3], cmplx *din) {
-
-  queue_setup();
-
   // setup device specific constructs 
   if(direction == 1){
     fftfpga_run_3d(0, N, din);
@@ -87,8 +97,6 @@ void pw_fpga_fft3d_sp_(int direction, int N[3], cmplx *din) {
   else{
     fftfpga_run_3d(1, N, din);
   }
-
-  queue_cleanup();
 }
 
 /******************************************************************************
@@ -98,8 +106,6 @@ void pw_fpga_fft3d_sp_(int direction, int N[3], cmplx *din) {
  * \param   din : complex input/output single precision data pointer 
  *****************************************************************************/
 void pw_fpga_fft3d_dp_(int direction, int N[3], cmplx *din) {
-  queue_setup();
-
   // setup device specific constructs 
   if(direction == 1){
     fftfpga_run_3d(0, N, din);
@@ -107,8 +113,6 @@ void pw_fpga_fft3d_dp_(int direction, int N[3], cmplx *din) {
   else{
     fftfpga_run_3d(1, N, din);
   }
-
-  queue_cleanup();
 }
 
 /******************************************************************************
@@ -120,6 +124,30 @@ void pw_fpga_fft3d_dp_(int direction, int N[3], cmplx *din) {
 void fftfpga_run_3d(int inverse, int N[3], cmplx *c_in) {
   cl_int status = 0;
   int inverse_int = inverse;
+  cl_kernel fft_kernel = NULL, fft_kernel_2 = NULL;
+  cl_kernel fetch_kernel = NULL, transpose_kernel = NULL, transpose_kernel_2 = NULL;
+  
+// Device memory buffers
+  cl_mem d_inData, d_outData;
+
+  // Create the kernel - name passed in here must match kernel name in the
+  // original CL file, that was compiled into an AOCX file using the AOC tool
+  fft_kernel = clCreateKernel(program, "fft3da", &status);
+  checkError(status, "Failed to create fft3da kernel");
+  fft_kernel_2 = clCreateKernel(program, "fft3db", &status);
+  checkError(status, "Failed to create fft3db kernel");
+  fetch_kernel = clCreateKernel(program, "fetch", &status);
+  checkError(status, "Failed to create fetch kernel");
+  transpose_kernel = clCreateKernel(program, "transpose", &status);
+  checkError(status, "Failed to create transpose kernel");
+  transpose_kernel_2 = clCreateKernel(program, "transpose3d", &status);
+  checkError(status, "Failed to create transpose3d kernel");
+
+  d_inData = clCreateBuffer(context, CL_MEM_READ_WRITE, sizeof(cmplx) * N[0] * N[1] * N[2], NULL, &status);
+  checkError(status, "Failed to allocate input device buffer\n");
+  d_outData = clCreateBuffer(context, CL_MEM_READ_WRITE, sizeof(cmplx) * N[0] * N[1] * N[2], NULL, &status);
+  checkError(status, "Failed to allocate output device buffer\n");
+
   cmplx *h_inData = (cmplx *)alignedMalloc(sizeof(cmplx) * N[0] * N[1] * N[2]);
   if (h_inData == NULL){
     printf("Unable to allocate host memory\n");
@@ -132,6 +160,8 @@ void fftfpga_run_3d(int inverse, int N[3], cmplx *c_in) {
   }
 
   memcpy(h_inData, c_in, sizeof(cmplx) * N[0] * N[1] * N[2]);
+
+  queue_setup();
 
   // Copy data from host to device
   status = clEnqueueWriteBuffer(queue6, d_inData, CL_TRUE, 0, sizeof(cmplx) * N[0] * N[1] * N[2], h_inData, 0, NULL, NULL);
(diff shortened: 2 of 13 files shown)


robertschade:
Hi, the singularity fpga-opencl emulation container works:
Number of FAILED  tests 0
Number of WRONG   tests 0
Number of CORRECT tests 3127
Number of NEW     tests 3
Total number of   tests 3130
GREPME 0 0 3127 3 3130 X
Best Wishes

arjunramaswami:
What do you think about the modifications? Would it be possible to merge the request so that we could provide the hardware for the dashboard tests? @oschuett @dev-zero

src/pw/fpga/fft_fpga.c
```diff
@@ -0,0 +1,353 @@
...
+ * \brief   compute an in-place single precision complex 3D-FFT on the FPGA
+ * \param   data_path_len - length of the path to the data directory
+ * \param   data_path - path to the data directory 
+ * \param   direction : direction - 1/forward, otherwise/backward FFT3d
+ * \param   N   : integer pointer to size of FFT3d  
+ * \param   din : complex input/output single precision data pointer 
+ *****************************************************************************/
+void pw_fpga_fft3d_sp_(int data_path_len, char *data_path, int direction, int N[3], cmplx *din) {
+
+  data_path[data_path_len] = '\0';
```
dev-zero: This assumes that buffer provided is bigger than the string length. If at all, shouldn't this be `data_path[data_path_len-1] = '\0'`?

Wouldn't it be easier to assume that Fortran calls the function with a `TRIM`med string ending with a `C_NULL_CHAR` instead? See also https://gcc.gnu.org/onlinedocs/gfortran/Interoperable-Subroutines-and-Functions.html

src/pw/fpga/opencl_utils.c
```diff
@@ -0,0 +1,370 @@
...
+  fclose(fp);
+  return binary;
+}
+
+// Minimum alignment requirement to use DMA
+const unsigned OPENCL_ALIGNMENT = 64;
+void* alignedMalloc(size_t size){
+
+  void *memptr = NULL;
+  int ret = posix_memalign(&memptr, OPENCL_ALIGNMENT, size);
```
dev-zero, line 258: This temporary can be removed, instead you might want to use something like https://stackoverflow.com/a/26855379 instead (respectively a simple wrapper around it) to prevent the compiler from showing `unused return value` messages.

It might also be good to document the default behavior of `posix_memalign` to leave `memptr` untouched or even explicitly set to `NULL`, leaving it to the caller to make sure the pointer returned is not `NULL`

src/pw/fpga/opencl_utils.c
```diff
@@ -0,0 +1,370 @@
...
+      break;
+
+    case 64 : 
+      strcat(path, subpath);
+      strcat(path, "syn64/");
+      break;
+
+    default:
+      printf("Path not found for the size (%d,%d,%d)", N[0], N[1], N[2]);
+      exit(0);
```
dev-zero: this is an error, probably should return something else than `0`

src/pw/fpga/opencl_utils.c
```diff
@@ -0,0 +1,370 @@
...
+
+#ifdef __PW_FPGA_SP
+  char *subpath = "/../fpgabitstream/fft3d/synthesis_sp/";
+#else
+  char *subpath = "/../fpgabitstream/fft3d/synthesis_dp/";
+#endif
+  
+  switch(N[0]){
+    case 16 : 
+      strcat(path, subpath);
```
dev-zero: possible memory corruption here if `path` is not large enough, same for all other uses of `strcat`

src/pw/fpga/opencl_utils.c
```diff
@@ -0,0 +1,370 @@
...
+/******************************************************************************
+ * \brief   returns the list of all devices for the specfic platform
+ * \param   context created using device
+ * \param   array of devices
+ * \param   number of devices found
+ * \param   size of FFT3d
+ * \retval  created program
+ *****************************************************************************/
+cl_program getProgramWithBinary(cl_context context, const cl_device_id *devices, unsigned num_device, int N[3], char *data_path){
+  char bin_path[500];
```
dev-zero: harcoded buffer size, please use a `const int` at the beginning with a meaningful name

src/pw/fpga/opencl_utils.c
```diff
@@ -0,0 +1,370 @@
...
+  char *binary;
+  char *binaries[num_device];
+
+  size_t bin_size;
+  cl_int bin_status, status;
+
+  strcpy(bin_path, data_path);
```
dev-zero: here you could re-use the length of the `bin_path` and use `strncpy` instead

src/pw/fpga/opencl_utils.c
```diff
@@ -0,0 +1,370 @@
...
+  if( !getBinaryPath(bin_path, N) ){
+    printf("No paths to the binary found\n");
+    return NULL;
+  }
+
+  // Load binary to character array
+  binary = loadBinary(bin_path, &bin_size);
+  if(binary == NULL){
+    printf("Could not load binary\n");
+    free(binary);
```
dev-zero: freeing a `NULL` pointer, probably not fatal, but still

src/pw/fpga/opencl_utils.c
```diff
@@ -0,0 +1,370 @@
...
+
+  if (!fileTestExists(path)){
+    printf("File not found \n");
+    return false;
+  }
+
+  return true;
+}
+
+static char* loadBinary(char *binary_path, size_t *bin_size){
```
dev-zero: I would prefer the interface to be `size_t loadBinary(char *path, char**buf)`, since that's what I see more in APIs, but ok.

src/pw/fpga/opencl_utils.c
```diff
@@ -0,0 +1,370 @@
...
+    printf("\n");
+    va_end(vl);
+
+    queue_cleanup();
+    cleanup();
+    exit(err);
+  }
+}
+
+static void tolowercase(char *p, char *q){
```
dev-zero, line 374: please document the assumptions for this function (`p` must be NULL terminated, `q` must hold `strlen(p)+1` characters)

src/pw/pw_fpga.F
```diff
@@ -0,0 +1,279 @@
...
+
+! **************************************************************************************************
+!> \brief Allocates resources on the fpga device
+! **************************************************************************************************
+   SUBROUTINE pw_fpga_init()
+#if defined (__PW_FPGA)
+      INTEGER :: stat
+
+#if defined (__PW_CUDA)
+#error "CUDA and FPGA cannot be configured concurrently!"
```
dev-zero, line 114: fyi the "#error@ directive may not be defined for all Fortran pre-processors

src/pw/pw_methods.F
```diff
@@ -1376,6 +1381,30 @@ SUBROUTINE fft_wrap_pw1pw2(pw1, pw2, debug)
...
+#else
+               CALL pw_fpga_r3dc1d_3d_dp(n, c_out)
+#endif
+               CALL zdscal(n(1)*n(2)*n(3), norm, c_out, 1)
+
+               CALL pw_gather(pw2, c_out)
+
+               DEALLOCATE (c_out)
+            ELSE
+               ALLOCATE (c_out(n(1), n(2), n(3)))
```
dev-zero: possibly move the `ALLOCATE(c_out..)/DEALLOCATE` outside of the `IF/ELSE`

dev-zero: same below

src/pw/fpga/opencl_utils.c
```diff
@@ -0,0 +1,370 @@
...
+    printf("Query for number of platforms failed\n");
+    exit(0);
+  }
+
+  // Get ids of platforms available
+  cl_platform_id *pids = (cl_platform_id*) malloc(sizeof(cl_platform_id) * num_platforms);
+  status = clGetPlatformIDs(num_platforms, pids, NULL);
+  if (status != CL_SUCCESS){
+    printf("Query for platform ids failed\n");
+    exit(0);
```
dev-zero: maybe use `1` here instead

src/pw/pw_methods.F
arjunramaswami: This I should've noticed.. Thanks :)

src/pw/pw_fpga.F
arjunramaswami: Added CPABORT statements after the error directive so that they fail on runtime in case the preprocessor doesn't catch this.

src/pw/fpga/opencl_utils.c
arjunramaswami: Made this change. Thanks

src/pw/fpga/opencl_utils.c
arjunramaswami: I see this is a better way to make APIs. Thanks for the suggestion, made this change as well 👍

src/pw/fpga/fft_fpga.c
```diff
@@ -0,0 +1,353 @@
...
+ *****************************************************************************/
+
+#if defined ( __PW_FPGA )
+
+// global dependencies
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <stdbool.h>
```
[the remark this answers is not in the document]
arjunramaswami: I removed stdbool and used ints instead. I think that is better than handling more constraints for the users. Thanks !

src/pw/fpga/fft_fpga.c
```diff
@@ -0,0 +1,353 @@
...
+static cl_kernel fft_kernel = NULL, fft_kernel_2 = NULL;
+static cl_kernel fetch_kernel = NULL, transpose_kernel = NULL, transpose_kernel_2 = NULL;
+
+// Device memory buffers
+cl_mem d_inData, d_outData;
+
+// Global Variables
+static cl_int status = 0;
+static int fft_size[3] = {0,0,0};
+bool fft_size_changed = true;
```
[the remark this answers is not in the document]
arjunramaswami: Removed these global states and made use of local variables instead. Had to refactor some code but I think it is more manageable now. Thanks for the suggestion

src/pw/fpga/fft_fpga.c
```diff
@@ -0,0 +1,353 @@
+/*****************************************************************************
+ *  CP2K: A general program to perform molecular dynamics simulations        *
+ *  Copyright (C) 2000 - 2019  CP2K developers group                         *
+ *****************************************************************************/
+
+/******************************************************************************
+ *  Author: Arjun Ramaswami
```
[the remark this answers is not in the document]
arjunramaswami: I removed this.

src/pw/fpga/fft_fpga.c
```diff
@@ -0,0 +1,353 @@
...
+static cl_platform_id platform = NULL;
+static cl_device_id device = NULL;
+static cl_context context = NULL;
+static cl_program program = NULL;
+
+static cl_command_queue queue1 = NULL, queue2 = NULL, queue3 = NULL;
+static cl_command_queue queue4 = NULL, queue5 = NULL, queue6 = NULL;
+
+static cl_kernel fft_kernel = NULL, fft_kernel_2 = NULL;
+static cl_kernel fetch_kernel = NULL, transpose_kernel = NULL, transpose_kernel_2 = NULL;
```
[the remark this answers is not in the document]
arjunramaswami: Moved them within the guard in the `fft_fpga.h` file. Thanks, this got me understanding guards 👍

src/cp2k_info.F
```diff
@@ -136,6 +136,12 @@ FUNCTION cp2k_flags() RESULT(flags)
 #if defined(__PW_CUDA)
       flags = TRIM(flags)//" pw_cuda"
 #endif
+#if defined(__PW_FPGA)
+      flags = TRIM(flags)//" pw_fpga"
+#endif
+#if defined(__PW_FPGA_SP)
+      flags = TRIM(flags)//" pw_fpga_sp"
+#endif
```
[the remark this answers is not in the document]
arjunramaswami: Added an `#error `and `CPABORT` check during initialization if only `__PW_FPGA_SP` is supplied.

src/pw/fpga/opencl_utils.c
arjunramaswami: I let this remain as it is. Seems to be the norm in OpenCL codes as well as this makes it more readable since CL_SUCCESS resolves to 0

dev-zero: The point here is that since this is running in the same address space as CP2K an `exit(0)` will terminate CP2K with an exit code of `0` which means a "successfull termination of the program" (see for example http://www.tldp.org/LDP/abs/html/exit-status.html). Which is clearly not the case.
Should this happen in a script then the script can only determine whether the run was successful (in the sense that CP2K finished gracefully, there is still the possibility that the calculation didn't converge) by checking the log file.

arjunramaswami: Thank you for the explanation ..

src/pw/fpga/fft_fpga.c
```diff
@@ -0,0 +1,353 @@
...
+        else{
+            fft_size[0] = N[0];
+            fft_size[1] = N[1];
+            fft_size[2] = N[2];
+            fft_size_changed = true;
+        }
+        return 1;
+    }
+    else{
+        return 0;
```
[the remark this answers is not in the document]
arjunramaswami: Sorry, I don't understand what you mean here.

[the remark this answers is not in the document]
dev-zero: Nevermind, with the `fft_size` declared here it is clearer what's going on here.

But, what happens if I call this function twice with different but valid `data_path` and/or `N`s without `deinit` in-between? Are we leaking memory if things like `fft_kernel` get simply overwritten?

And what happens if I call it like

```
pw_fpga_check_bitstream_("/tmp/my.dat", {16,16,16});
cleanup();
pw_fpga_check_bitstream_("/tmp/my.dat", {16,16,16}); // due to state tracking here for 'caching' it will properly reinitialize the structures as expected I guess
```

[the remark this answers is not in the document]
arjunramaswami: I see, successive calls with different `N`s causes memory leaks since it does another `init` without cleaning up. Will modify that (done!). 
Calling the function twice with different datapaths having the same `N` would end up using the first `N`'s program that was already loaded. Currently, there are fixed paths to find the binaries, this could be changed in the future to use env variable/ input file/ command line args. 
The code snippet example will cause failure because the binary / program will not be available to create buffers.

dev-zero:
I'd say this is ready to be (squash) merged. Other opinions?

PR #412, opened 2019-06-12
Refactor RPA

fstein93:
I mainly splitted some modules and used the RPA infrastructure for SOS-Laplace-MP2.


hfp:
Btw, GNU Fortran 8.3 says the following:

```text
src/hfx_libint_interface.F:241:14:

          DO i = 2, m_max+1
                          2
             F(i) = 0.0_dp
              1
Warning: Array reference at (1) out of bounds (2 > 1) in loop beginning at (2)
```

fstein93:
Unfortunately, the compiler cannot know that m_max is always small enough. (Compare the size of the array F and the meaning of m_max (sum of the four angular momentum quantum numbers.) The code itself is much older. Are there known issues in this respect?

pseewald:
> Btw, GNU Fortran 8.3 says the following:
> 
> ```
> src/hfx_libint_interface.F:241:14:
> 
>           DO i = 2, m_max+1
>                           2
>              F(i) = 0.0_dp
>               1
> Warning: Array reference at (1) out of bounds (2 > 1) in loop beginning at (2)
> ```

Yes this caught my attention as well, but this was here before (so has nothing to do with this pr). It's pretty sure that this is a coding mistake, for all other cases `i` starts with `1` and this particular case is not tested.

Update: sorry I was wrong it is correct (`F(1)` is manually set before the loop so the loop should start with `2`). Probably the warning is issued without considering that if `m_max < 1` the code inside the loop will never be executed. For the other cases the loop boundary is `m_max+1` as well but the warning is only triggered here, so I guess the warning is bogus.

JWilhelm:
Could you please have a look at https://www.cp2k.org/exercises:2017_uzh_cp2k-tutorial:gw to ensure the inputs there are working?

Could it be worth a check whether massively-parallel GW calculations are still working? E.g. similar to the ones from SI of https://doi.org/10.1021/acs.jpclett.7b02740 and https://doi.org/10.1021/acs.jctc.6b00380

fstein93:
Then, we have to change one line in that tutorial or make an appropriate hint such that the GW keyword has to be set in the G0W0 section instead of IM_TIME section. That will be documented in the manual anyway.

I will make these tests tomorrow.

JWilhelm:
Exactly, could you pls change this line in the tutorial? Maybe, an additional sentence anywhere in the tutorial could be helpful that the input has slightly changed. Thanks for checking that the big GW calcs still give the same numbers :)

fstein93:
@dev-zero How can I add something in the tutorial https://www.cp2k.org/exercises:2017_uzh_cp2k-tutorial:gw?

dev-zero:
@fstein93: past tutorials are locked (and should mention the used CP2K version)

fstein93:
@JWilhelm Are you able to edit the tutorial or could you accept not to change the tutorial, then?

JWilhelm:
@fstein93 Unfortunately, I cannot log into my account and my account seems to be linked to the uzh email adress which is not valid any more. We can then leave the tutorial as it is...

PR #416, opened 2019-06-17
Dfet

rybkinjr:
debug the conditional jump


dev-zero:
@rybkinjr when merging, please consider using the `Squash and merge` option (which merges your commits into one) instead of the `Rebase and merge`. This would help to keep the commit history cleaner and help when generating changelogs and having to disect commits when looking for regressions.

PR #420, opened 2019-06-18
Initial version of the autotools arch generation

mtaillefumier:
This pull request does not change anything in the build system whatsoever and is **not** intended to replace the toolchain to build cp2k dependencies. Instead, it use autotools (a simple generated configure shell script) to detect all dependencies installed on the machine. It generates the different versions of the arch files according to the libraries found on the machine. It is possible to start from these generated arch files and teak them (with optimizations parameters mostly)

It is not perfect but works on fedora, rhel, summit (ornl), and basically should work on any descent unix systems with bash. The configure script will use pkg-config (if it finds it) or rely on the good old fashion way to detect libraries. Note that the arch files are **not** copied in the arch directory to avoid confusion. 

You will need automake, autoconf, m4 and autoconf-archives If you want to generate the configure script but you need to do it only one time. Dependencies can be deactivated with the without-dependency paradigm. CUDA, mkl, and fftw have additional options accessible with a simple

./configure --help

Comments are most welcome.

NB : it is under GPL by default


oschuett:
This is awesome !!! I think it will go a long way in easing build pain.

Regarding the licensing: CP2K is GPL2+ but you uploaded GPL3. Would it be ok for you to contribute this under GPL2+ instead?

mtaillefumier:
Sure, I did not even know it was GPL3 in the copyright file. I will modify it and fix the syntax mistake pkg.m4 (some weard non ascii character).

mtaillefumier:
- I linked cp2k license file in the directory. GPL2+ apply to the all directory unless stated in the files.
- Darn, some of the m4 macros are BSD (mpi one). Not a problem but still !!!!

shoshijak:
@mtaillefumier, very naïve question: 
What does compiling CP2K with your proposed autotools look like concretely? I mean: 
- which commands does the user execute?
- how much time does the configuration take? 
- how does the user specify options? (eg with/without MPI, CUDA, with/without this or that optional dependency?)
- what does the user have to indicate to autotools or tweak?

mtaillefumier:
- you need autotools to generate the configure script. The rest does not depend on autotools afterwards. People that compiled some gnu packages will be in known territory.

- basically, you just have to do

`./autogen.sh` (unless the configure script is already there) then 
`./configure`
Normally it will detect as many libraries as possible. when it is done just copy local_cuda.psmp (for instance) in arch and run make as usual. But if you already have a working arch file just ignore it completely. It is really useful when you do not know where things installed. The teaking part comes afterwards or by specfying CFLAGS, CXXFLAGS and FCFLAGS, for instance `CFLAGS="-O2 -mcpu=power" ./configure`

On summit for instance, you could do
module load gcc/6.2.0
module load netlib-lapack
module load netlib-scalapack
module load cuda
./configure
cp local_cuda.psmp ../../arch/local_cuda_gcc620.psmp
cd ../..
make VERSION=psmp ARCH=local_cuda_gcc620 -j16

done deal +- bugs that people will find.

shoshijak:
Ah, I see, thanks! 
So in your Summit example, why isn't it necessary to run `./autogen.sh` ?

oschuett:
We should check in the generated configure script. Then building CP2K will follow the familiar GNU style:
```
./configure
make
make test
make install
```

I can also extend the formatting test to ensure that the generated configure script remains up to date.

tools/autotools/LICENSE
```diff
@@ -0,0 +1 @@
+/home/mathieut/Projects/codes/cp2k/LICENSE
```
oschuett: Better use a relative symlink.

mtaillefumier:
because I just forgot to write ./autogen.sh. You need to do it only one's or if you change something in configure.ac.
> Ah, I see, thanks!
> So in your Summit example, why isn't it necessary to run `./autogen.sh` ?

tools/autotools/LICENSE
mtaillefumier: > I can also extend the formatting test to ensure that the generated configure script remains up to date.

Yes I think it is a good idea. you will need autoconf-archive installed. Any version will do. I did not want to version the configure script because it is easy to generate from the rest.

PR #422, opened 2019-06-19
Fix issue with Libxc

### LINKED ISSUE #421: libxc warning "unstable" functional for untested
libxc introduced the functionality to get a "stable/unstable" flag per functional.
Unfortunately this seems to be a tri-state at the moment: "stable/unstable/unknown", with almost all functionals being "unknown". In CP2K we're treating "unknown" (the absence of "stable") as "unstable", leading to the simplest (and most likely) stable functionals being reported as "unstable".
We must handle this differently. I see two possibilities at the moment:

1. remove the check again
2. treat "unknown" as "stable"

@fstein93 more ideas?

fstein93:
Fixes #421 
Instead of checking for XC_FLAGS_STABLE, I propose to check for XC_FLAGS_DEVELOPMENT.

Do you think, we should further add a hint to INSTALL.md?

### DIFF
--- a/src/xc/xc_libxc.F
+++ b/src/xc/xc_libxc.F
@@ -102,15 +102,17 @@ CONTAINS
 !> \param needs the components needed by this functional are set to
 !>        true (does not set the unneeded components to false)
 !> \param max_deriv maximum implemented derivative of the xc functional
+!> \param print_warn whether to print warning about development status of a functional
 !> \author F. Tran
 ! **************************************************************************************************
-   SUBROUTINE libxc_lda_info(libxc_params, reference, shortform, needs, max_deriv)
+   SUBROUTINE libxc_lda_info(libxc_params, reference, shortform, needs, max_deriv, print_warn)
 
       TYPE(section_vals_type), POINTER         :: libxc_params
       CHARACTER(LEN=*), INTENT(OUT), OPTIONAL  :: reference, shortform
       TYPE(xc_rho_cflags_type), &
          INTENT(inout), OPTIONAL               :: needs
       INTEGER, INTENT(out), OPTIONAL           :: max_deriv
+      LOGICAL, INTENT(IN), OPTIONAL            :: print_warn
 
       CHARACTER(len=*), PARAMETER :: routineN = 'libxc_lda_info', &
                                      routineP = moduleN//':'//routineN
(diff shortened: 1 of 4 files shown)


src/xc/xc_libxc.F
```diff
@@ -181,8 +181,8 @@ SUBROUTINE libxc_lda_info(libxc_params, reference, shortform, needs, max_deriv)
             CPABORT(TRIM(func_name)//": this XC_FAMILY is currently not supported.")
          END SELECT
       END IF
-      IF (.NOT. (xc_libxc_wrap_is_stable(xc_info))) THEN
-         CPWARN(TRIM(func_name)//" is not considered stable. Use with caution.")
+      IF (.NOT. (xc_libxc_wrap_is_under_development(xc_info))) THEN
```
dev-zero: shouldn't `xc_libxc_wrap_is_under_development` return `.TRUE.` if it is under development?

src/xc/xc_libxc_wrap.F
```diff
@@ -266,31 +266,31 @@ INTEGER FUNCTION xc_libxc_wrap_functional_get_number(func_string) RESULT(func_id
...
+   LOGICAL FUNCTION xc_libxc_wrap_is_under_development(xc_info)
       TYPE(xc_f03_func_info_t)                           :: xc_info
 
-      CHARACTER(LEN=*), PARAMETER :: routineN = 'xc_libxc_wrap_is_stable', &
+      CHARACTER(LEN=*), PARAMETER :: routineN = 'xc_libxc_wrap_is_under_development', &
          routineP = moduleN//':'//routineN
 
       INTEGER                                            :: handle
 
       CALL timeset(routineN, handle)
```
dev-zero: please remove `timeset` and `timestop` here in a second commit while being at it, there is no point in measuring time for this part

dev-zero:
@fstein93 I think the log message is clear enough. If at all, I'd place some explanation at https://www.cp2k.org/faq similar to https://www.cp2k.org/faq:hint_insufficiently_exploiting_cpu_extensions

dev-zero:
otherwise looks good, I think handling it via `is_under_development` makes more sense given they're not (yet) using the `is_stable` property

OndrejMarsalek:
Currently, the warning is printed at every SCF iteration. This clutters the output and also bloats the number of warnings reported at the end, possibly masking other warnings. Would it perhaps be better to print it only at the beginning of the SCF, or even earlier, just once per run?

fstein93:
@OndrejMarsalek I included your proposition.

fstein93:
I have nothing more to add.

PR #423, opened 2019-06-19
Script for planning MPI/OMP-hybrid execution

hfp:
Added CP2K/plan script with documentation (Markdown). The script suggests MPI-rank and OpenMP-thread combinations, which can be helpful for hybrid execution (PSMP). The material is taken from the XCONFIGURE project (https://github.com/hfp/xconfigure/) and hence it is keeping the copyright banner and license as it is used there (BSD 3-clause).

### DIFF
--- a/tools/plan_mpi_omp/plan.md
+++ b/tools/plan_mpi_omp/plan.md
@@ -1,0 +1,60 @@
+# CP2K MPI/OpenMP-hybrid Execution (PSMP)
+
+## Overview
+
+CP2K's grid-based calculation as well as DBCSR's block sparse matrix multiplication (Cannon algorithm) prefer a total rank-count to be a square-number (2d communication pattern). This is not to be obfuscated with a Power-of-Two (POT) rank-count that usually leads to trivial work distribution (MPI). This property can be as dominant as wasting cores is more effective than fully utilizing all cores (sometimes a frequency upside over an "all-core turbo" emphasizes this property further). Counter-intuitively, even an unbalanced rank-count per node i.e., different rank-counts per socket can be an advantage. Pinning MPI processes and placing threads requires extra care to be taken on a per-node basis to load a dual-socket system in a balanced fashion or to setup rank-spacing and OpenMP threads.
+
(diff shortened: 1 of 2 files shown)


hfp:
Any comment whether this PR suits or not is welcome. There is no inference with existing code, and hence this can be merged without problem.

tools/plan_mpi_omp/plan.md
```diff
@@ -0,0 +1,60 @@
...
+
+```bash
+mpirun -np 16 \
+  -genv I_MPI_PIN_DOMAIN=auto -genv I_MPI_PIN_ORDER=bunch \
+  -genv OMP_PLACES=threads -genv OMP_PROC_BIND=SPREAD \
+  -genv OMP_NUM_THREADS=6 \
+  exe/Linux-x86-64-intelx/cp2k.psmp workload.inp
+```
+
+**NOTE**: For hybrid codes, `I_MPI_PIN_DOMAIN=auto` is recommended as it spaces the ranks according to the number of OpenMP thread (`OMP_NUM_THREADS`). It is not necessary or recommended to use `I_MPI_PIN_PROCESSOR_LIST` for hybrid codes (MPI plus OpenMP). To display and log the pinning and thread affinization at the startup of an application, `I_MPI_DEBUG=4` can be used with no performance penalty. The recommended `I_MPI_PIN_ORDER=bunch` ensures that ranks per node are split as even as possible with respect to sockets (imagine e.g., 36&#160;ranks on a 2x20-core system to put 2x18 ranks instead of 20+16 ranks).
```
dev-zero, line 17: are `I_MPI_PIN_ORDER` Intel MPI specific flags or should they work with OpenMPI and MPICH as well? Just wondering since all documentation I found after a quick search point to Intel MPI.

dev-zero:
besides the comment, looks like a very useful tool

tools/plan_mpi_omp/plan.md
hfp: Indeed, this is Intel MPI specific. However, it is mentioned as such "To easily place the ranks, Intel MPI is used" (to my knowledge this is not an emphasized statement). Anyhow, it's not only because of obvious reasons (aka Intel affiliation). I actually worked with OpenMPI and it is somwwhat hard to achieve the same (bind-to, map-by). I believe using OpenMPI and placing less ranks on a node and still balancing between two sockets plus having space for OMP_NUM_THREADS between ranks is only possible with explicit pinning. The latter needs a uttility/helper script to generate the CPUIDs to place on. In turn it must be carefully evaluated if the CPUIDs of the MPI implementation in question rely on logical IDs. Only in the latter case is reliable to map consistently (CPU numbering can vary even for the same kind/SKU).

I think the example in the documentation can be extended to help those people with OpenMPI. However, I don't have it at hand. The goal is to help as many people as possible, no matter which MPI is used plus improving performance "for free". Let me take this task separately.

PR #425, opened 2019-06-20
Improve CPUID hint for cases where CFLAGS are not in sync with FCFLAGS

hfp:
The translation unit base/machine_cpuid.c relies on the (same) target flags as the Fortran code, otherwise it is impossible (with the current implementation) to determine which target flags have been given at compile-time of CP2K.

If m_cpuid_static() returns MACHINE_CPU_GENERIC, then it is likely CFLAGS did not carry the same target flags as FCFLAGS. Instead of hinting "The compiler target flags used to build this binary are insufficiently exploiting the extensions which are available for this CPU model", the hint "Consider compiler target flags as part of FCFLAGS and CFLAGS (ARCH file)" is given.

Prettified the source code.

### DIFF
--- a/src/base/machine.F
+++ b/src/base/machine.F
@@ -210,9 +210,9 @@ END FUNCTION m_cpuid
 ! **************************************************************************************************
 FUNCTION m_cpuid_name(cpuid)
       INTEGER                                            :: cpuid
-      CHARACTER, POINTER                                 :: m_cpuid_name(:)
+      CHARACTER(len=default_string_length), POINTER      :: m_cpuid_name
 
-      CHARACTER, DIMENSION(default_string_length), SAVE, TARGET :: name_generic = "generic", &
+      CHARACTER(len=default_string_length), SAVE, TARGET :: name_generic = "generic", &
          name_unknown = "unknown", name_x86_avx = "x86_avx", name_x86_avx2 = "x86_avx2", &
          name_x86_avx512 = "x86_avx512", name_x86_sse4 = "x86_sse4"
 
--- a/src/base/machine.F
+++ b/src/base/machine.F
@@ -56,10 +56,10 @@ MODULE machine
 
   PRIVATE
 
-  PUBLIC :: m_walltime, m_datum, m_flush, m_flush_internal,&
-            m_hostnm, m_getcwd, m_getlog, m_getpid, m_getarg, m_procrun,&
-            m_memory, m_iargc, m_abort, m_chdir, m_mov, m_memory_details,&
-            m_energy, m_memory_max, m_cpuinfo, m_cpuid_static, m_cpuid
+  PUBLIC :: m_walltime, m_datum, m_hostnm, m_flush, m_flush_internal,&
+            m_getcwd, m_getlog, m_getpid, m_getarg, m_iargc, m_procrun, m_abort,&
+            m_chdir, m_mov, m_memory, m_memory_details, m_memory_max, m_energy,&
+            m_cpuinfo, m_cpuid_static, m_cpuid, m_cpuid_name
 
   INTERFACE
     ! **********************************************************************************************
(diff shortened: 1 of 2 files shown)


hfp:
@mkrack this PR may point to the root cause of the unjustified hint you encountered.

dev-zero:
What about distros which might want to enforce `MACHINE_CPU_GENERIC`?

hfp:
> What about distros which might want to enforce `MACHINE_CPU_GENERIC`?

The current implementation was one example to make use of the new CPUID infrastructure. We meanwhile also use it to select an ELPA kernel. Anyhow, this HINT has its up- and downsides.

We could introduce a preprocessor symbol to omit any hints. If you think that's necessary you can suggest a name for this preprocessor symbol (`__CP2K_NOHINTS`?). Another option is to turn the message about `MACHINE_CPU_GENERIC` into something positive like "This is a generic build".

hfp:
This is not ready for merge yet. I will add some improvement to the main hint-message.

dev-zero:
To be honest, I don't think we should start (resp. continue) trying to catch build problems with runtime checks since they will never be sufficient. If these build inconsistencies continue to be a problem we really have to consider switching build/configure systems and force people to use them to get consistent builds.

oschuett:
>I don't think we should start (resp. continue) trying to catch build problems with runtime checks since they will never be sufficient.

Unfortunately, there is no silver bullet. Building CP2K will always be messy due to the diverse set of architectures and libraries we support. Hence, we should continue to implement checks where and whenever possible.

hfp:
Improving the hint was suggested/asked by mkrack. Even if tests go through I will first check if the message looks "beautiful" (I have not seen it yet with those names coming from the `m_cpuid_name`).

hfp:
An example output with HINT may look like:

```text
 GLOBAL| Force Environment number                                              1
 GLOBAL| Basis set file name                                      GTH_BASIS_SETS
 GLOBAL| Potential file name                                           POTENTIAL
 GLOBAL| MM Potential file name                                     MM_POTENTIAL
 GLOBAL| Coordinate file name                                      __STD_INPUT__
 GLOBAL| Method name                                                        CP2K
 GLOBAL| Project name                                                     H2O-32
 GLOBAL| Preferred FFT library                                             FFTW3
 GLOBAL| Preferred diagonalization lib.                                     ELPA
 GLOBAL| Run type                                                             MD
 GLOBAL| All-to-all communication in single precision                          F
 GLOBAL| FFTs using library dependent lengths                                  F
 GLOBAL| Global print level                                                  LOW
 GLOBAL| MPI I/O enabled                                                       T
 GLOBAL| Total number of message passing processes                            16
 GLOBAL| Number of threads for this process                                    3
 GLOBAL| This output is from process                                           0
 GLOBAL| CPU model name :  Intel(R) Xeon(R) Platinum 8260L CPU @ 2.40GHz
 HINT in environment.F:801 The compiler target flags (x86_avx2) used to build
   this binary cannot exploit all extensions of this CPU model (x86_avx512).
   Consider compiler target flags as part of FCFLAGS and CFLAGS (ARCH file).

 MEMORY| system memory details [Kb]
 MEMORY|                        rank 0           min           max       average
```

alazzaro:
I found a possible problem introduced by this PR.
On Swan (CRAY system) I do cross-compiling. The login node is Broadwell, while I do run on Skylake.
It turns out that my ELPA version compiled with:

```
./configure --prefix=/home/users/p02464/libs/libelpa/intel/2017.05.002 \
--enable-openmp=yes --enable-static=yes --enable-shared=no \
--enable-generic=yes --enable-sse=yes --enable-sse-assembly=yes \
--enable-avx=yes --enable-avx2=yes --enable-gpu=no --enable-avx512=no \
--enable-bgp=no --enable-bgq=no CC=cc FC=ftn CXX=CC \
LDFLAGS="-dynamic"
```
but then when I run CP2K I get the message:

 Cannot set ELPA2 stage real_kernel
 Cannot set ELPA2 stage real_kernel

I assume this is because ELPA is looking for avx512 kernels, right?
I can cross-compile ELPA for AVX512 and see if it works...

Any comment is welcome.

hfp:
> I assume this is because ELPA is looking for avx512 kernels, right?

Yes, that's right.

> I can cross-compile ELPA for AVX512 and see if it works...

Please try.

> Cannot set ELPA2 stage real_kernel

The question is if this could be caught upfront in CP2K's ELPA module...

hfp:
Btw, your case is exactly what this was made for: select the best possible kernel (according to CPUID).

Perhaps the ELPA recipe should be adjusted to build all kernels (in case `--enable-avx512=no` was applied automatically). I remember that by default all ELPA kernels are built, did you request `--enable-avx512=no` explicitly?

alazzaro:
Well, my current problem comes from this line:

            has_AVX512=`grep '\bavx512f\b' /proc/cpuinfo 1>/dev/null && echo 'yes' || echo 'no'`

in the install_elpa.sh, therefore it doesn't install AVX512 since I'm compiling on Broadwell.
I can force the compilation to have AVX512, apart of that it is not a problem...

alazzaro:
OK, everything went fine (https://www.cp2k.org/static/regtest/trunk/swan-skl28/CRAY-XC40-intel.psmp_17.0.4.196.out)

mkrack:
I have the same issue. ELPA compiles by default without avx512 kernels which is fine, since my CPU does not support that feature, but CP2K wants to call
ELPA| Kernel                   :  AVX512_BLOCK4
I guess I am doing something wrong. Is there a way to tell CP2K that the avx512 kernels are unavailable?

hfp:
Hmm, but that's different from what Alfio ran into. I had the impression the installer/build script explicitly disabled AVX-512. So, this time it's a report that it's disabled by default if the system does not have the required AVX-bits.

hfp:
Perhaps the overspec'd build script should be scrubbed:
https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/install_elpa.sh#L70
... of course all AVX-friends not just AVX-512.

mkrack:
I don't use the toolchain for the regression testing. The problem is that my cp2k build returns 0 for cpuid which causes in https://github.com/cp2k/cp2k/blob/3e58b4ddeff8a8fa5f0840d8a89013e17bf63d4f/src/fm/cp_fm_elpa.F#L153 that the default case 17 for index is chosen.

mkrack:
I think the problem is that my pdbg regression tester is compiled without -D__LIBXSMM, but libxsmm is mandatory to have a reasonable cpuid available.

hfp:
> that the default case 17 for index is chosen.

That's a bug!  Thank you for finding this. I will get this sorted out.

hfp:
I am proposing the following condition:

```fortran
IF ((MACHINE_CPU_GENERIC .LT. cpuid) .AND. (cpuid .LE. MACHINE_X86)) THEN
```

hfp:
The cp2k/master revision now implements the suggested/revised condition.

PR #428, opened 2019-06-21
Fix ScaLAPACK settings for MPICH & MKL

OndrejMarsalek:
These changes allow `install_mkl.sh` to find the required libraries and enable ScaLAPACK from MKL. Something is still not quite right, though, as the script still builds Netlib ScaLAPACK, unless I give ` --with-scalapack=no`.

I don't know if there are MKL versions that actually provide `libmkl_blacs_lp64.a`, but I checked the MKL link advisor and it suggests `libmkl_blacs_intelmpi_lp64.a` for MPICH, at least for MKL versions 2017-2019.


OndrejMarsalek:
I am putting the need for `--with-scalapack=no` in a separate issue: #435.

PR #438, opened 2019-06-25
Dashboard: Drop GCC 4.8

oschuett:


alazzaro:
Just to confirm, are we dropping 4.8.x everywhere?
If so, then I will update https://www.cp2k.org/dev:compiler_support

oschuett:
Well, I don't seem much reason for keeping it. Right now we can't run with libint2, which takes out many important features. And once #356 is merged it won't compile at all.

dev-zero:
I guess 4.9 will also be out once we're using `SOURCE=` or `MOLD=`, maybe target 5.4 which was afaik at some point widely used in Linux distros?

alazzaro:
OK, 4.8.x is out now, I'm still keeping 4.9.x (used at UZH).

https://www.cp2k.org/dev:compiler_support

PR #444, opened 2019-06-26
ao_util: optimized exp_radius

hfp:


src/aobasis/ao_util.F
```diff
@@ -190,57 +190,59 @@ FUNCTION exp_radius(l, alpha, threshold, prefactor, epsin) RESULT(radius)
...
+      r1 = scale*r0+1.0_dp
+      DO iter = 1, maxiter
+         ar2 = a*r1*r1
+         g = d*r1**l*EXP(-ar2)
+         IF (g .LT. t) EXIT
+         r0 = r1
+         r1 = scale*r0+1.0_dp
       ENDDO
+      ! avoid system calls in tight/arithmetic loops
+      IF (iter .EQ. maxiter) THEN
```
dev-zero: Shouldn't this rather be: `IF (iter .EQ. maxiter+1)`?
... resp. as before: `IF(iter .GT. maxiter)`

dev-zero:
```console
$ cat iter.f90 
program main
    integer :: idx

    do idx = 1, 10
        print *, idx
    enddo

    print *, "final idx: ", idx
end program
$ ./iter 
           1
           2
           3
           4
           5
           6
           7
           8
           9
          10
 final idx:           11
```

hfp: It's a draft ;-)

... and I think there is a much bigger issue. Let me describe:

* The first iteration captures an interval (r0, r1] such that g(r1) < threshold.
* The second iteration "bisects" the interval such that rd=r1-r0 and rd < epsiter.
* The function is supposed to yield a radius r in (r0, r1] such that r->min.

The original code was supposed to be a bisection i.e., the interval contracts with 0.5. So, I think Golden Section Search (GSS) should work as well and give an egde over bisection (contraction != 0.5). In any case (Bisection or GSS),  unimodal functions are the prerequisite/limitation for finding an extreme value (minimum in this case). The unimodal property is likely ensured by the code that produces the initial interval. The captured function g(r) over (r0, r1] looks like coming from the top and strongly decaying to zero ("function plot").

Now the problem: all tests (e.g., tests/LIBTEST/test_eri_mme_accuracy_longrange.inp) seem to be very sensitive i.e., even if a contraction produces an r that is slightly smaller and even holds g(r) < threshold, it seems the tests will fail. The test calculates reference values that are "theoretical upper bounds" and then claims that (some) test values exceeded the "theoretical upper bound".

I think I will play a bit further with this...

hfp: Indeed, I messed up the final iteration index. Anyhow, that is not a problem at this point and can be fixed later.

hfp: Btw, I am (intentionally) lumping together the iterations from the first part and the latter part:

```fortran
DO iter = iter, maxiter
```

I think what matters are the number of function-evaluations g(r). Btw, in most cases the algorithm does ~10 iterations according to the new counting (as just explained).

pseewald: Is there any particular reason for optimizing this specific routine?

> Now the problem: all tests (e.g., tests/LIBTEST/test_eri_mme_accuracy_longrange.inp) seem to be very sensitive i.e., even if a contraction produces an r that is slightly smaller and even holds g(r) < threshold, it seems the tests will fail. The test calculates reference values that are "theoretical upper bounds" and then claims that (some) test values exceeded the "theoretical upper bound".

If it is just the `eri_mme` tests, I'd need to have a look. I'm aware that `test_eri_mme_accuracy_longrange.inp` is not very stable and even the claim about the theoretical upper bound is not strictly true, so you can ignore that test.

FYI: I wrote a faster implementation of `exp_radius` myself for the `eri_mme` integrals since `exp_radius` was a bottleneck, you can find it [here](https://github.com/cp2k/cp2k/blob/master/src/eri_mme/eri_mme_gaussian.F#L262-L292). Instead of bisection, it directly iterates over equidistant points, but it avoids evaluating exponentials in the loop. However I would not trust that implementation to be stable for general use.

oschuett:
I'm sorry. I broke the CI for the past hour. Should be [fixed](https://github.com/cp2k/cp2k-ci/commit/7e705ed2277cb2a64b0fa8258af9281e1ab5f772) now.

src/aobasis/ao_util.F
hfp: Thank you, I will take a look. Let's call this exercise "without real need" so far. I just came accross this routine since it's all scalar. I only wanted to remove the system-call from the loop with EXP in it (and I will only do that if everything else fails). Anyhow, I am attracted by the fact how instable it makes CP2K although changing the routine from "bisection" to "GSS" should not make a difference for the matter of exp_radius. However, what's called "theoretical upper bound" might be only a set of number that is calibrated to the current (not the tightest) error margins.

Thank you again for sharing your work!

hfp:
@oschuett no worries.

hfp:
Note to myself: one potential problem with the original routine is, it can slide into "zero-land" and then just cut-down the r-interval without further guidance by any function-value g(r) other than zero. It may cut the interval without delivering a minimal radius but just *some* radius that is (potentially far out) in "zero-land".

hfp:
@pseewald I have incorporated your idea of sampling the function.

hfp:
Here is the description for how the new implementation of exp_radius works:
* The sampling based algorithm is only the first stage tightening the initial interval.
* Sampling distance based on estimated radius (to stay within maxiter).
* The second stage ensures preconditions for Golden Section Search ("Bisection") are met.
* The second stage is like the former way to find the initial interval.
* Not many iterations are supposed to be spent in the second stage.
* The last stage is a Golden Section search.

There were several problems with the former implementation:
* It was possible to exceed maxiter in case of too tight bracketting (epsiter)
  even when no change was made to the interval after certain iteration.
* The above issue happened when the left or right interval boundary was hit,
  but numerically there was no further contraction possible (invariant).
* To avoid the issue, many calls/tests supply a large epsiter (e.g., 0.27).
* In turn, large epsiter produced radii anywhere in the remaining interval,
  but exponential function was still strongly decaying...
* The absolute nature of epsiter causes different meaning for different Gaussians
* I have implemented "reliter" ("150") to ensure the remaining interval is
  below 0.75% of the final radius. The value 150 is slightly tighter than what
  was measured for the old implementation (130...145).
* However, "reliter" ensures the "same" meaning no matter what Gaussians
  are processed and effectively prevents to miss relative error margins.
* I tried to make the new exp_radius more stable wrt code changes
  and the related effect on test/error margins.

Replacing the implementation of exp_radius is somewhat a big deal since current reference/test values should be replaced. Though, error margins are not very large but exceed currently recorded margins. Again, it is very likely that the current results are more correct.

hfp:
I will do some stand-alone performance analysis for the previous and the new implementation of exp_radius and report the results here.

pseewald:
Thank you for incorporating the direct sampling. I would like to replace `exp_radius_fast` by this new implementation. How does the new implementation perform if `eps_in` is relatively large compared to the width of the gaussian (the case that was optimized with `exp_radius_fast`)? I guess in that case direct sampling would be most efficient and golden section search would not be needed. Could the new implementation in that case entirely rely on the first stage (if I correctly assume that this is most efficient)?

hfp:
> I guess in that case direct sampling would be most efficient and golden section search would not be needed.

Very likely. Since the radius is estimated upfront, and used to determine an appropriate sampling the new implementation is likely more robust. If on the other hand, the sampling distance is too small the "fast" algorithm can take a very long time.

> Could the new implementation in that case entirely rely on the first stage?

I think "reliter" can become an (optional) argument (currently it is a hard-coded parameter). As "reliter" only applies to the GSS stage, it may also control whether this stage runs or not. It's also possible to replace the fast implementation with the new fast impl., and then call it from the new general exp_radius.

> if I correctly assume that this is most efficient

I will do some performance analysis and report it here.

src/aobasis/ao_util.F
```diff
@@ -281,10 +279,16 @@ FUNCTION exp_radius(l, alpha, threshold, prefactor, epsin) RESULT(radius)
          CPABORT("The requested threshold is zero")
       END IF
 
+      IF (PRESENT(epsin)) THEN
+         epsiter = epsin
+      ELSE ! arbitrary scaled epsilon?
+         epsiter = EPSILON(epsiter)*(1.0E-12_dp/2.22044604925031E-16_dp)
```
dev-zero: isn't this equivalent to simply `epsiter = 1.0E-12_dp` ?

At least `2.22044604925031E-16_dp` is the epsilon of our `dp` type and `gcc -O0` optimizes this away to `epsiter = 1.0E-12_dp`, see https://gcc.godbolt.org/z/65G8Tx

dev-zero: maybe @vondele knows more about this? Based on git history, this seems to be a remnant of the single-precision experiments

hfp: I believe it is a hand-calibrated eps to ensure the algorithm terminates (at least for the test cases). The problem with the old/current implementation is it may not terminate for too small espilon. This is resolved.

hfp: btw, can you recommend a test for eri_mme_lattice_summation? It currently uses exp_radius although exp_radius_fast is already a dependency of eri_mme_lattice_summation.

vondele: I think @dev-zero is right, this was needed for single precision experiments. This stuff is from long ago, so I've forgotten the details. These routines need to be handled with care, i.e. are both critical for performance and accuracy. Underestimating the radius leads to large inaccuracies elsewhere, significantly overestimating leads to larger costs. The target accuracy should be on the order of a grid spacing (IIRC). Nevertheless, these routines are old and might definitely benefit from a new round of TLC.

pseewald: > btw, can you recommend a test for eri_mme_lattice_summation? It currently uses exp_radius although exp_radius_fast is already a dependency of eri_mme_lattice_summation.

It turns out that in the current implementation, `exp_radius_fast` is not used at all and I don't remember whether this is what I intended in pr #369. I will do some testing on my own to find out if it matters whether `exp_radius` or `exp_radius_fast` is used.

hfp: @pseewald Thank you!  Please pull this latest PR into your test branch (I moved exp_radius_fast into ao_util.F because of using it to implement exp_radius; see commit log about cross-dependency).

With your insight we should decide if we want to take the new exp_radius, or if we only want to take exp_radius_fast. I also think with own insight and Joost's input, we may want to adjust tolerances for ccurrently failing tests (since no current/later reference might be "better" or more correct). With this exp_radius in general, it is hard to reproduce the exact same result when the implementation changed (even slightly). Please note, tests also/even fail with a much tighter iteration interval (so reference values are somewhat arbitrary).

Regarding performance, `tests/LIBTEST/test_eri_mme_performance.inp` is not dominated by exp_radius but other ERI related pieces such as error control (I will detail this later).

pseewald: @hfp I know for sure that the error control overhead in `test_eri_mme_performance.inp` is not representative for performance in actual applications since it's only called once. Other than that, any insights are very welcome! Some time ago I identified the `pgf_sum_3c_*` routines as main bottlenecks and optimized them in #369.

hfp:
I will update exp_radius_fast again. The revised version will use the automatically calculated (wide) steps, and then switch to an (eventually) given epsilon/samping.

src/aobasis/ao_util.F
pseewald: @hfp I finally checked that indeed `exp_radius_fast` is no longer needed for me. So I plan to remove it from the `eri_mme_gaussian` module in a later commit.

hfp: @pseewald That's also part of this PR.

I think I could update the tolerances. However, I do not know how to avoid the "failure" with `test_eri_mme_accuracy.inp`. It brings up this ultimate argument `Actual error greater than upper bound estimate`.

vondele: note that you should not update the tolerance, but the reference values for this kind of test. Tolerances are used to take into account variation due to machine architectures, threading etc, the tolerance on most tests is far stricter than all the thresholds in the code actually implies. If one changes the algorithms (like here) reference values need to be updated.

hfp: @vondele Thank you for the comment. I will do so if this PR goes forward.

pseewald:
Some insights into the apparent instability of `test_eri_mme_accuracy` w.r.t. the new implementation of `exp_radius`: there is a parameter [`SUM_PRECISION`](https://manual.cp2k.org/trunk/CP2K_INPUT/TEST/ERI_MME_TEST/ERI_MME.html#SUM_PRECISION) that is directly passed to the `threshold` argument of the `exp_radius` function. By default this threshold is `1.0E-12`. With the new implementation of `exp_radius` this threshold leads to absolute errors as large as `4.50E-03`. In order to obtain the same accuracy as before, I need to change `SUM_PRECISION` to `1.0E-15`. Could it be that the old implementation accidentally returned radii that were larger than required by `threshold`?

hfp:
> Could it be that the old implementation accidentally returned radii that were larger than required by threshold?

Thanks for the analysis!

I need to check with the old implementation to confirm your suspicion. Note, `threshold` applies to g(r) whereas `epsabs` (the former `epsin`) applies to the radius - I guess you refer to epsabs/epsin.

For sure, the old implementation was not able to reach certain small `epsin`, it kept iterating until `maxiter` since it did not check that `r1-r0` did not contract any further (for numeric reasons).

pseewald:
> I need to check with the old implementation to confirm your suspicion. Note, `threshold` applies to g(r) whereas `epsabs` (the former `epsin`) applies to the radius - I guess you refer to epsabs/epsin.

I was really referring to `threshold` - so my guess was that the old implementation returned a larger radius than the minimum `r` satisfying `g(r) < threshold`. This would also imply that we need to tweak the defaults in CP2K that affect the `threshold` parameter rather than updating the reference values.

hfp:
> returned a larger radius than the minimum r satisfying g(r) < threshold.

That's for sure, but so does the new implementation. The problem is that a tighter radius needs more iterations. It's what @vondele pointed out - it was carefully tuned to be sufficient, but to not take too long. It's certainly not feasible to converge to machine precision (tightest radius).

What I did was to add one more convergence criterion (`epsrel`) in an attempt for more stability (better quality measure). However, I think the Bisection/GSS method along with the kind of function (very long decay with a function value of g(r) that is close to machine precision) does not allow to get much better. Perhaps a Newton iteration has faster convergence, but I am not sure if the expense stay in bound given that the derivative is needed.

One hard cut perhaps could be to relax `epsabs` to never hit, and to rely on `epsrel`. Note, `threshold` is only the prerequisite for termination of the algorithm whereas currently `epsabs` mostly determines the quality.

hfp:
Made tail-iteration (GSS) using a deterministic trip-count, which enables vectorization of exp_radius (data-dependent iteration is in exp_radius_fast aka first stage of exp_radius).

hfp:
The `tests/QS/regtest-nonortho/graph*` input files show large deviations with the new exp_radius.

pseewald:
> I think I could update the tolerances. However, I do not know how to avoid the "failure" with test_eri_mme_accuracy.inp.

This is now solved with #479, so after rebase that test will succeed.

dev-zero:
@hfp could you possibly rebase this PR? We have a user in the Google group reporting the following timing in a geoopt who could possibly benefit from this optimization:

```
SUBROUTINE | CALLS | ASD |   | SELF TIME |   | TOTAL TIME
-- | -- | -- | -- | -- | -- | --
  | MAXIMUM |   | AVERAGE | MAXIMUM | AVERAGE | MAXIMUM
CP2K | 1 | 1.0 | 0.560 | 0.562 | 880.431 | 880.44
cp_geo_opt | 1 | 2.0 | 0.060 | 0.060 | 632.220 | 632.22
geoopt_bfgs | 1 | 3.0 | 0.174 | 0.180 | 632.160 | 632.16
cp_eval_at | 35 | 4.0 | 0.036 | 0.078 | 627.687 | 627.71
qs_forces | 34 | 5.0 | 0.113 | 0.118 | 614.752 | 614.76
qs_energies | 35 | 6.0 | 0.012 | 0.015 | 565.518 | 565.55
scf_env_do_scf | 35 | 7.0 | 0.002 | 0.005 | 381.893 | 381.90
scf_env_do_scf_inner_loop | 164 | 8.0 | 0.028 | 0.032 | 260.120 | 261.02
qs_init_subsys | 1 | 2.0 | 0.406 | 0.408 | 245.615 | 245.62
qs_env_setup | 1 | 3.0 | 0.043 | 0.043 | 244.914 | 244.92
qs_env_rebuild_pw_env | 70 | 6.5 | 0.069 | 0.070 | 244.794 | 244.80
pw_env_rebuild | 1 | 5.0 | 0.092 | 0.129 | 244.709 | 244.71
compute_max_radius | 1 | 6.0 | 232.268 | 244.451 | 232.268 | 244.45
rebuild_ks_matrix | 198 | 9.5 | 0.001 | 0.001 | 198.294 | 198.45
qs_ks_build_kohn_sham_matrix | 198 | 10.5 | 0.078 | 0.082 | 198.293 | 198.45
qs_ks_update_qs_env | 199 | 9.0 | 0.002 | 0.002 | 159.038 | 159.18
qs_energies_init_hamiltonians | 35 | 7.0 | 0.033 | 0.034 | 151.114 | 151.13
build_qs_neighbor_lists | 35 | 8.0 | 0.187 | 0.212 | 100.828 | 136.07
init_scf_loop | 35 | 8.0 | 0.001 | 0.002 | 120.711 | 120.72
pw_transfer | 4199 | 13.3 | 0.347 | 0.410 | 93.912 | 95.354
```

hfp:
Will do!

hfp:
The code is updated to current cp2k/master. It misses to update reference values (regtests). For this I would like to ask for some help. It was discussed already, but touching exp_radius is brittle and tightening the interval makes performance worse. The new implementation is more robust/stable and allows additional criterions wrt error margin (absolute, relative, num iter). It did some good wrt to performance (should be double-checked with the forum test case). However, the implementation still does not vectorize. This is in the nature of the algorithm since a singular/scalar variable is iteratively refined, which cannot vectorize naively wrt loop iteration (inherently scalar).

Late comment: I will give it another update and fix the format...

hfp:
This is "ready", i.e., Forum test case can be check for improved performance.

hfp:
I have used the following [script](https://github.com/cp2k/cp2k/pull/916#issuecomment-628448764) to update reference values.

hfp:
When looking at the following [case](https://groups.google.com/d/topic/cp2k/xuezH18qhW4/discussion), I managed to get walltime down by **1.7x**. The workload shared in the forum requires ~**312** seconds based on CP2K/master vs. ~**183** seconds w/ this PR (on the same system using an optimized PSMP-binary). There is further potential to accelerate the related functions (`exp_radius`, `exp_radius_fast`), but this is certainly a separate/later PR.

hfp:
SDBG passed and this is ready to be merged!

However, I triggered the PSMP tests and there (likely unrelated) wrong results, e.g., ELPA and PEXSI tests. This might be also because PSMP tests are non-default and perhaps are lagging some updated tolerances.

mkrack:
Indeed, a few [regtests](https://www.cp2k.org/static/regtest/trunk/Linux-x86-64-intel-regtest/popt) are failing with the current tolerances.

hfp:
@mkrack I will update reference values (rather than tolerances) covering the wrong results.

hfp:
This PR causes a performance regression related to exp_radius_very_extended. I will look into it.

alazzaro:
> This PR causes a performance regression related to exp_radius_very_extended. I will look into it.

The performance tests on Daint are reporting this problem too, see:

https://dashboard.cp2k.org/archive/cray-xc50-daint-gpu-psmp_perf/index.html
https://dashboard.cp2k.org/archive/cray-xc50-daint-mc-psmp_perf/index.html

hfp:
Thank you! I am on it (but it can take a few days). I can also revert and resubmit later.

hfp:
See PR #905

alazzaro:
Thanks Hans for the fix! Just to be sure, could yo confirm that the few seconds slowdown in the H2O-32 is not due to your change? It appears in both plots GPU and MC (first plot):

https://dashboard.cp2k.org/archive/cray-xc50-daint-gpu-psmp_perf/index.html
https://dashboard.cp2k.org/archive/cray-xc50-daint-mc-psmp_perf/index.html

The linear scaling is fine.

hfp:
Let me check this for the exact same (or scaled down) workload!

Algorithmically, the code is no different from before (with current parameterization) except handling some optional arguments and a different modularization/slightly different formulation. So, if it is slower it can be fixed. However, the optional rlow now allows finding maximum radius without fully computing the radius if it cannot reach the given rlow (early exit).

hfp:
I am just waiting for the Dashboard to become green. The regression (RPA/MP2) was in the order of ~20% and it is fixed (already shown by [cray-xc50-daint-gpu-psmp_perf](https://dashboard.cp2k.org/archive/cray-xc50-daint-gpu-psmp_perf/index.html) and [cray-xc50-daint-mc-psmp_perf](https://dashboard.cp2k.org/archive/cray-xc50-daint-mc-psmp_perf/index.html)). Root cause was [https://github.com/cp2k/cp2k/blob/master/src/aobasis/ao_util.F#L204](https://github.com/cp2k/cp2k/blob/master/src/aobasis/ao_util.F#L204).

PR #448, opened 2019-06-27
Dashboard: Add openmpi-psmp tester

oschuett:


dev-zero:
what is the reason we're running this stuff as root?

oschuett:
Convenience mostly because one has to `apt-get install` stuff all the time. I guess, we could create a new user during `install_basics.sh` and give it sudo-powers. Then we could run the rest of the Dockerfile as that user.

oschuett:
It's really unfortunately that `OMPI_ALLOW_RUN_AS_ROOT=1` and `OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1` don't work as advertised.

oschuett:
There also seems to be a memory leak problem. Does anybody want to look into it or should I just switch to psmp?

OndrejMarsalek:
I am not sure I understand the issue. Is there a memory leak in popt but not in psmp?

oschuett:
I think the leaks are in OpenMPI itself. So, we'd need to compile it with debug symbols and then add suppression rules for LeakSanitizer. I'm not sure if it's worth the effort, though. Hence, I switched to `psmp` for now which also tests threading compatibility. However, if you want to further work on this I'm happy to also enable the `pdbg` variant for the OpenMPI toolchain.

mkrack:
Older OpenMPI releases have been full of memory leaks. That was exactly the reason why I changed some years ago from OpenMPI to MPICH for pdbg regression testing, since MPICH has had almost no issues of that kind. So, it seems that not all leaks have been fixed even in the current OpenMPI release.

OndrejMarsalek:
I don't think I have the bandwidth to look into that right now :-/

I was not aware that this is a common issue with OpenMPI. Certainly 4.x is now better in that all the tests pass, at least, but if it is supported by the toolchain at all, perhaps a warning should be printed that there might be memory leaks. I think OpenMPI is quite popular and one should expect people to (want to) use it in production, especially if it is an option in the toolchain.

oschuett:
Ok, then let's stick with `psmp` for the time being. The leaks in MPI libraries are usually benign and we can use the MPICH toolchain to chase down leaks in CP2K.

PR #458, opened 2019-07-07
Refactor collocate / integrate for GPU acceleration

oschuett:


juerghutter:
I started to refactor collocate/integrate routines. I will report on progress soon.

juerghutter:
I have a first version of the collocate routines ready. I was able to partly
factor out the main work routine. It has still about 20 external references, but
most of them a trivial code. This could be a starting point for accelerator programming.

You can find the code in my branch:
https://github.com/juerghutter/cp2k/tree/ACC_COLLOCATE
The file to look at is:
grid_collocate_ref.F

oschuett:
The refactoring by @juerghutter was a great leap! In [qs_collocate_density.F:1811](https://github.com/cp2k/cp2k/blob/b5ddac954f55fa85794d224463af68186a499fe7/src/qs_collocate_density.F#L1811) we can now switch between the old OpenMP and a new simplified reference implementation. It consists of only ~2k lines, mostly in [grid_base_ref.F](https://github.com/cp2k/cp2k/blob/b5ddac954f55fa85794d224463af68186a499fe7/src/grid_base_ref.F) and [qs_modify_pab_block.F](https://github.com/cp2k/cp2k/blob/b5ddac954f55fa85794d224463af68186a499fe7/src/qs_modify_pab_block.F).

The transformations in `qs_modify_pab_block` appear to be stencils. So, maybe we could precompute those on the host and simplify the GPU code?

juerghutter:
I add an initial version of a MiniApp. See the project ColInt on my account.
https://github.com/juerghutter/ColInt

oschuett:
Great! So, should we start to integrate the pieces? I'd propose the following order:

1. Merge the [ACC_COLOCATE](https://github.com/juerghutter/cp2k/commits/ACC_COLLOCATE) branch.  It still needs some [touch up](https://github.com/cp2k/cp2k/pull/458/commits), an input keyword to switch between the different implementations, and add a few regtests.

2. Apply my [refactorings](https://github.com/cp2k/cp2k/pull/526) of the `tast_list` and `rs_distribute_matrix`. I'm also contemplating to replace `pair2int` with normal type fields. 

3. Turn the [miniapp](https://github.com/juerghutter/ColInt) into a unit test.

@juerghutter, should I merge ACC_COLOCATE together with my touch-ups or do you want to do it yourself?

juerghutter:
I think the ACC_COLOCATE branch is not ready for merging. We definitely want to wait until after the next release. Too many shaky changes.

Turning the miniapp into a unit test is certainly a good idea, but has low priority. Main purpose of
the code is to get GPU development going.

oschuett:
I agree that we should hold off with the merge until after the release.

Nevertheless, I would like to reconcile our changes and continue with the C-interface. Since we are both working on the `task_list` we'll need to take turns. So, are you planning any more changes on the `ACC_COLOCATE` branch over the next 2-3 weeks?

juerghutter:
I will not work on this anymore in the near future. I want to see if the GPU code makes progress first.

src/grid_base_ref_c.c
```diff
@@ -0,0 +1,171 @@
...
+                        const int grid_lbound_x,
+                        const int grid_lbound_y,
+                        const int grid_lbound_z,
+                        const int lp,
+                        const int cmax,
+                        const double coef_xyz[(lp+1)*(lp+2)*(lp+3)/6],
+                        const double pol[3][cmax+1][lp+1][2],
+                        const int map[3][2*cmax+1],
+                        const int *sphere_bounds,
+                        double grid[grid_size_z][grid_size_y][grid_size_x]
```
dev-zero: please note that an interface like this requires to build with `-std=c99` (or higher).
Furthermore when looking at other C functions in standard libraries, it seems that the convention is to pass the data first (as `double *`) and then the dimensions.
Also, the `lbound` seems like something which should be handled on the Fortran-side transparently if possible.

src/grid_base_ref_c.c
```diff
@@ -0,0 +1,171 @@
...
+    const int kgmin = sphere_bounds[sci++];
+    for (int kg=kgmin; kg <= 0; kg++) {
+        const int kg2 = 1 - kg;
+        const int k = map[2][kg + cmax];
+        const int k2 = map[2][kg2 + cmax];
+
+        // initialize coef_xy
+        const int n_coef_xy = (lp+1)*(lp+2)/2;
+        double coef_xy[n_coef_xy][2];
+        for (int i=0; i < n_coef_xy; i++) {
```
dev-zero: if going the C99-route, just write: `double coef_xy[n_coef_xy][2] = {0.0};`

oschuett: Unfortunately, this does not to work because C lacks broadcasting:
```
error: variable-sized object may not be initialized
```

src/grid_base_ref_c.c
oschuett: - I added `-std=c99` to the toolchain.
- The dimensions have to be passed before they can be used in a declaration.
- With pointer arguments the bounds are actually passed around as well. And I simply haven't checked yet `grid` always gets allocated with lower bound of 1.

PR #463, opened 2019-07-11
qs_3c_tensors module

pseewald:
This module replaces tensor setup in cubic scaling RPA by more general
methods to build 3-center integral tensors.


pseewald:
@oschuett do you know what's wrong with the CI?

oschuett:
Lauchpad.net was down for a while. As a consequence several job were restarted a few of times, which occupied all machines and led to timeouts of other jobs. For the future #465 should help.

PR #465, opened 2019-07-12
Toolchain: Download all non-Github dependency from cp2k.org

oschuett:
Launchpad.net is down again. Let's reduce the CI's exposure.


dev-zero:
be careful, you have changes in the submodule stashed in this commit

oschuett:
Thanks @dev-zero! I totally missed that.

PR #467, opened 2019-07-12
Set up a NEGF calculation using a single force environment

schulkov:


oschuett:
It seems like the `QS/regtest-negf` directory is not registered in [TEST_DIRS](https://github.com/cp2k/cp2k/blob/master/tests/TEST_DIRS) and therefore ignored by our tests.

PR #472, opened 2019-07-14
Fix minimal arch-file regression tests

schulkov:


oschuett:
Alternatively, you could make `fftw3` a requirement. See e.g. [QS/regtest-ps-implicit-1-1](https://github.com/cp2k/cp2k/blob/master/tests/TEST_DIRS#L69). Nevertheless, you might want to run the two test with different `INTEGRATION_METHOD` to increase coverage.

PR #487, opened 2019-07-23
Enabled GAPW together with ADMM withouht DFT exchange correction

abussy:
Leaves the possibility to compute HFX energy contributions based on reduced basis sets in all electron calculations. This is only allowed if there is no DFT exchange correction, _i.e._ EXCH_CORRECTION_FUNC is set to NONE. Be it GPW or GAPW, if there is no correction, the integration of the aux_fit potential is now skipped.


oschuett:
Since this enables a new combination of features, could you please add some regtests for it.

abussy:
Because the changes only affect the DFT exchange correction part, I only added 2 very simple regtests which do not involve purification.

PR #490, opened 2019-07-25
dashboard: added entry for fft-fpga emulation on Noctua of PC2.

robertschade:


oschuett:
It seems like the test is run twice. Note that the report should only contain the output from one run:
```
Fri Jul 26 01:40:19 CEST 2019
*************************** testing ended ********************************
*************************** testing started ******************************
 started on  Fri Jul 26 01:40:41 CEST 2019
```

robertschade:
Sorry, I will have a look at that immediately.

alazzaro:
@robertschade I see that the dashboard test on Noctua has not been running for a while... 

https://dashboard.cp2k.org/archive/fft-fpga-emulation/index.html

Anything possible to fix?

Issue #351, opened 2019-05-10
Build DBCSR via the toolchain


shoshijak:
Currently, DBCSR is built inside CP2K and their two build systems are very much intertwined. As a result, issues such as #338 can arise, and some fixes inside DBCSR's build system https://github.com/cp2k/cp2k/pull/345#issuecomment-491226981 require a new DBCSR release before they can be reflected in CP2K. 

What is more, DBCSR maintains two build systems (Makefile to be compiled by cp2k and cmake to compile DBCSR as a standalone). with all the errors this could cause.

--- 

What about delegating the build of DBCSR to CP2K's toolchain (i.e. treat DBCSR like all other CP2K dependencies), letting the toolchain use DBCSR's cmake for building, but still keeping DBCSR inside CP2K's tree as a git submodule as is the case now? 

The main motivations for this suggestion are:
- Removes from CP2K the burden of compiling DBCSR correctly and having intimate knowledge of DBCSR's Makefile as well as the variables it needs. The consistency between DBCSR's compilation flags would be handled by DBCSR's cmake
- Avoids having to maintain two build systems in DBCSR (cmake & make) as is currently the case
- DBCSR is kept in CP2K's tree as a git submodule for those who want to develop DBCSR & CP2K together (if I recall correctly, this was the original motivation for having git as a submodule?)
- This suggestion would effectively be a step towards making DBCSR a fully external library, but still allows keeping DBCSR as a git submodule inside CP2K.

If you think this is a useful step, I would be happy to open a PR with a prototype.

shoshijak:
From @dev-zero:

> CP2K and DBCSR are still in a co-development state (see timings between rc-releases of DBCSR and updates of the submodule), except for the accelerator part obviously. At the moment people will get an automatic rebuild of DBCSR whenever the submodule is updated, while the other way round they would simply get build failures in CP2K (best case scenario, worst case it builds and they get wrong results) to remind them to manually update the toolchain as well. Given that there are enough (mostly barely tested as it seems) changes to the toolchain at the moment by 3rd-party projects I really would like to avoid throwing DBCSR into this mix for now.

oschuett:
> CP2K and DBCSR are still in a co-development state

In other words: DBCSR has not yet acquired another big user besides CP2K. Given that this was the main motivation for making DBCSR a library, it seems we have three options:

1. Keep the weird status-quo as we wait a little longer for new users.
2. Make DBCSR fully independent and let it diverge, e.g. by switching to CMake.
[...]

PR #495, opened 2019-07-29
Fix problem of interdependent optionals in construct_domain_preconditioner

rzk1:


oschuett:
Hi Rustam, the `exts/dbcsr` situation is a bit [unfortunate](https://github.com/cp2k/cp2k/issues/351). You'll need to remove that last commit
`git reset --hard 6423a226b`, run `git submodule update --recursive`, and finally `git push --force`.

Btw, [almo_scf_methods.F](https://www.cp2k.org/static/coverage/src/almo_scf_methods.F.gcov.html) and [almo_scf_optimizer.F](https://www.cp2k.org/static/coverage/src/almo_scf_optimizer.F.gcov.html) have pretty low test coverage.

hfp:
There is an additional issue with the submodule (at least for me), the `exts` directory/content appears to be modified for no obvious reason and one may think it must be staged and committed (`git diff` does not show an actual difference). Depending on the Git client, the option to commit into the submodule is offered... anyhow, the issue here was maybe different.

rzk1:
Thanks for the explanation Ole! I followed your instructions and everything seems to be fine now. I guess I do not need to create another pull request. It seems that new checks are triggered automatically.

Low-coverage problem is in my todo list. It will be addressed in another PR.

PR #497, opened 2019-07-30
Use PLUMED 2.5.2 for regtests

mkrack:


oschuett:
I just realized that PLUMED is not part of the toolchain anymore. We must have accidentally dropped it :-/

PR #500, opened 2019-08-01
Adjust tolerance in QS/regtest-rel/h2o-1.inp

rzk1:
The single failed test in my previous PR (38a7112) is not caused by my routines. The reason for the failure is an overly tight in ERROR threshold in TEST_FILES


oschuett:
Thanks for following up nevertheless.

PR #501, opened 2019-08-05
Feature/toolchain version bumps

dev-zero:


dev-zero:
@oschuett can you please take a look at the CI? It seems to pull the changes to toolchain scripts but not the additional files

oschuett:
You'll have to add the new file to the appropriate COPY statement in the [Dockerfile](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/Dockerfile). While this is not really convenient, it helps the build cache.

dev-zero:
@oschuett thanks. This is indeed not very convenient

PR #506, opened 2019-08-08
Bugfix/fm init random

dev-zero:


dev-zero:
This fix seems to introduce more noise in the values :(

oschuett:
Uhh, this seems like a pretty fundamental change. Are you sure the old behavior was wrong?

dev-zero:
@oschuett based on the scarce documentation and the logic previously present in `cp_fm_init_random` I would say so.
But maybe @vondele remembers?

vondele:
not sure I remember. It could be bug, but it could also be that this was intentional, i.e. to make sure tests running under farming would return the same result as tests running standalone. Maybe the commit message of the commit that introduced this code gave a hint?

oschuett:
How should we proceed with this PR?
Are there further validation we could perform, or is it simply not worth the risk?

dev-zero:
Please stop closing valid/unresolved issues.

oschuett:
I don't think we have an *unresolved issues* here, because the code does already work.

I do agree, you've found some code smell that would be nice to clean up.

So, what's your plan for validation?

dev-zero:
The plan is: write an isolated unittest for FM (just basic routines for now), including a test for `fm_init_random`, checking for inequality of first elements  of matrices generated by subsequent calls, which I assume to fail since the initial seed is saved at routine exit rather than the current seed. Then probably tie the local RNG to CP2K's global initial seed, add a `&GLOBAL` option to fix the RNG in FM completely (e.g. make it possible to retain the current behaviour for regtesting).

ltalirz:
Relevant historical commits

https://github.com/cp2k/cp2k/commit/6a53fdccec1fb06e60aafc798ab8adb0298f7d1a
https://github.com/cp2k/cp2k/commit/645155b047f5921b6fe3deb8b874d355988ebe1d

one says "The data in random initial matrices is no longer dependent on # cpus or nrow_block / ncol_block"

oschuett:
This PR has been in limbo for more than two years. I propose that we summarize our discussion as a code comment and then close it.

oschuett:
I've added a comment for posterity: cd2af721ef43522dfdb7fb6268497a2c3e53b7aa.

Issue #433, opened 2019-06-24
Toolchain: Improve error messages


oschuett:
The toolchain is intentionally kept very quiet. As a consequence the [error messages](https://groups.google.com/forum/#!msg/cp2k/RFTmBwSK-uE/7zBqjZWNAAAJ) are usually not very helpful. Maybe we could write a small helper function that runs a command but only prints its output when the command has failed.

oschuett:
Implemented via #1871. Thanks a lot @mkrack!

Issue #13, opened 2018-10-03
Add option to read connectivity/MOLNAME from extra column in XYZ file


dev-zero:
while the `FORCE_EVAL/SUBSYS/COORD` allows to add an optional string after the coordinates to specify  the connectivity, additional columns in a XYZ file are ignored, since the XYZ file format is not standardized. Nevertheless, it would be nice to have an option to tell CP2K to use the N-th additional column in a XYZ for the connectivity (or possibly other things).

@KanySUI is that about right?

he-zilong:
This issue could be retired in favor of #4962 about extended XYZ formats; a general parser for something like the [ovito standard names](https://www.ovito.org/manual/reference/file_formats/input/xyz.html#file-formats-input-xyz-extended-format) may be worth pursuing.

Several other issues are mentioning this issue, not because they are on the same topic, but because they contain improperly formatted error messages with the `#13` string in it.

PR #509, opened 2019-08-11
Fix for sirius options. More tests

mtaillefumier:
this pull request contains a correction to an error introduced in the last SIRIUS pre-release. An include file was mistakenly included in the release which corrupted the list of options. Unfortunately, I realized the error too late and by snow ball effect landed in cp2k. 

- **Updated** the toolchain to the new SIRIUS prerelease version. 
- cp2k now support full potential compuations (through sirius). I had to force cp2k not to interpret the json potential file all the time.
- some improvements in the autotools build (well generate only the arch files)
- Added more tests for the SIRIUS backend. But we need more tests especially for magnetic materials and molecular dynamics computations.


mtaillefumier:
NB : 
- the new SIRIUS prerelease needs to be downloaded here https://github.com/electronic-structure/SIRIUS/archive/v6.3.2.tar.gz
- How do you pass the initial magnetization to cp2k

juerghutter:
Magnetization is a keyword in the &KIND sections:

MAGNETIZATION {Real}
The magnetization used in the atomic initial guess. Adds magnetization/2 spin-alpha electrons and removes magnetization/2 spin-beta electrons.
This keyword cannot be repeated and it expects precisely one real.
Default value: 0.00000000E+000

It is stored in the qs_kind_type and can be accessed by

CALL get_qs_kind(qs_kind, magnetization=XX)

tools/autotools/autogen.sh
```diff
@@ -7,5 +7,5 @@ echo "Running aclocal..." ; aclocal $ACLOCAL_FLAGS -I m4 || exit 1
 echo "Running autoheader..." ; autoheader || exit 1
 echo "Running autoconf..." ; autoconf || exit 1
 echo "Running libtoolize..." ; (libtoolize --copy --automake || glibtoolize --automake) || exit 1
-echo "Running automake..." ; automake --add-missing --copy --gnu || exit 1
+#echo "Running automake..." ; automake --add-missing --copy --gnu || exit 1
```
dev-zero: please remove if it's really not needed anymore

mtaillefumier:
So there is no way to indicate an antiferromagnetic configuration for atom of the same kind ?

> Magnetization is a keyword in the &KIND sections:
> 
> MAGNETIZATION {Real}
> The magnetization used in the atomic initial guess. Adds magnetization/2 spin-alpha electrons and removes magnetization/2 spin-beta electrons.
> This keyword cannot be repeated and it expects precisely one real.
> Default value: 0.00000000E+000
> 
> It is stored in the qs_kind_type and can be accessed by
> 
> CALL get_qs_kind(qs_kind, magnetization=XX)

dev-zero:
> So there is no way to indicate an antiferromagnetic configuration for atom of the same kind ?

There is also the `&BS` ("Broken Symmetry") keyword to set a specific spin configuration per kind: https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/SUBSYS/KIND/BS.html

juerghutter:
You need to define two kinds in that case, one with +M and one with -M

mtaillefumier:
> > So there is no way to indicate an antiferromagnetic configuration for atom of the same kind ?
> 
> There is also the `&BS` ("Broken Symmetry") keyword to set a specific spin configuration per kind: https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/SUBSYS/KIND/BS.html

That's more towards what I wish. In SIRIUS for instace we just add the magnetization after the positions such as

Ni 0.0 0.0 0.0 0.0 0.0 4.0

the last three numbers been the direction of the magnetization. It would be nice to have this in cp2k. Maybe with a dedicated section or simply add the magnetization after the coordinates ?

dev-zero:
> > > So there is no way to indicate an antiferromagnetic configuration for atom of the same kind ?
> > 
> > 
> > There is also the `&BS` ("Broken Symmetry") keyword to set a specific spin configuration per kind: https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/SUBSYS/KIND/BS.html
> 
> That's more towards what I wish. In SIRIUS for instace we just add the magnetization after the positions such as

But it is also in the `KIND` section, so I don't get why you like this better ;-)
`MAGNETIZATION` seems to be much closer to the SIRIUS (and QE) way of specifying it.

> 
> Ni 0.0 0.0 0.0 0.0 0.0 4.0
> 
> the last three numbers been the direction of the magnetization. It would be nice to have this in cp2k. Maybe with a dedicated section or simply add the magnetization after the coordinates ?

While this could be done when loading coords from an XYZ file (https://github.com/cp2k/cp2k/issues/13 requests a similar extension to our XYZ parser, just for a different property), I don't see this happening for the CP2K topology section:
Until now, a clear distinction was made between defining the geometry and the properties for the constituents. I don't think it would be a good idea to change this part now, otherwise you'd also have to start considering adding other properties to be moved.

oschuett:
I uploaded [SIRIUS-6.3.2.tar.gz](https://www.cp2k.org/static/downloads/SIRIUS-6.3.2.tar.gz).

tools/toolchain/scripts/install_sirius.sh
```diff
@@ -127,19 +127,14 @@ case "$with_sirius" in
...
-             make -j $NPROCS -C src >> compile.log 2>&1
-
-            install -d ${pkg_install_dir}/include >> install.log 2>&1
-            install -d ${pkg_install_dir}/lib >> install.log 2>&1
-            cp -R ../src/* ${pkg_install_dir}/include >> install.log 2>&1
-            install -m 644 src/*.a ${pkg_install_dir}/lib >> install.log 2>&1
-            install -m 644 src/mod_files/*.mod ${pkg_install_dir}/include >> install.log 2>&1
+                  ${COMPILATION_OPTIONS} -DCMAKE_INSTALL_PREFIX=${pkg_install_dir} .. #> compile.log 2>&1
+            make -j $NPROCS #>> compile.log 2>&1
+	        make install #>> compile.log 2>&1
```
dev-zero: I guess these redirects are going to change again before the merge?

mtaillefumier: Yes. I try to know why it does not pass your ci while it compile fine on my desktop. Not having access to the log file is a pain.

oschuett: The [log file problem](https://github.com/cp2k/cp2k-ci/issues/48) should be fixed now. Alternatively, you can always run `docker build` locally to reproduce what the CI is doing.

oschuett: Instead of adding/removing redirects we could write a small helper function that runs a command but only prints its output when the command has failed. See also https://github.com/cp2k/cp2k/issues/433.

mtaillefumier: thanks for the mod.

PR #512, opened 2019-08-16
Prettify and remove unused atom_read_upf_json()

oschuett:


oschuett:
I realized that SIRIUS parses pseudo potentials on its own. Hence, I believe we no longer need to depend on JSON-Fortran. @mtaillefumier, do you agree?

PR #514, opened 2019-08-16
Workaround for false gfortran warning

oschuett:


oschuett:
The warning says:
```
/workspace/cp2k/src/hfx_libint_interface.F:241:14:

          DO i = 2, m_max+1
                          2
             F(i) = 0.0_dp
              1
Warning: Array reference at (1) out of bounds (2 > 1) in loop beginning at (2)
```

pseewald (approved):

PR #517, opened 2019-08-22
cp2k_runs/manual: document how to specify units and how they behave, drop --html-manual

dev-zero:


oschuett:
See #518 for discussion.

PR #523, opened 2019-08-23
singles correction (RSE) to RPA

rybkinjr:


alazzaro:
@rybkinjr 
It seems that, as a result of this PR, the tests on Swan are failing with wrong results:

QS/regtest-ri-rpa-rse/RI_RPA_RSE_H2_minus.inp.out
QS/regtest-ri-rpa-rse/Cubic_RPA_RSE_H2.inp.out 

From what I can see, my guess is that this is due OpenMP threads. Swan uses 4 threads, while all other tests use 2. I can test Swan with 2 threads, but I would appreciate if you can a local regtest with 4 threads...

rybkinjr:
Thanks for the comment, Alfio. I will chek the OMP directives myself.

alazzaro:
I can confirm the problem doesn't happen with 1 single thread (PSMP with a single thread). Now I'm checking SSMP with a single thread. I suggest to use the flag:
```
-fsanitize=thread
```
and check the tests with 4 threads...

alazzaro:
Confirm, also SSMP with 1 thread runs fine. So I suggest to use SSMP compiled with `-fsanitize=thread` and check for data races in the two tests...

alazzaro:
For the record, the problem was fixed via #535

PR #534, opened 2019-08-27
Benchmark: STMV for xTB using LS

vondele:
1M atom benchmark of the Satellite Tobacco Mosaic Virus (STMV) to exercise xTB and the linear scaling (LS) code at scale.

More info and some results in README.md


juerghutter:
Hi Joost

You switch of Hirshfeld charges (cannot be calculated in xTB) and remove the GMAX values.
In xTB the Ewald parameters are calculated from a default set.

vondele:
OK, will make the changes, rerun and update the PR

vondele:
updated, should be ready to merge, IMO

PR #559, opened 2019-09-13
vdW-DF functionals data files

juerghutter:


oschuett:
I'm wondering if we really need that special `@XCTYPE` marco? Couldn't we just use the existing `@INCLUDE`?

For example, instead of `@XCTYPE vdW-DF2` we could use `@INCLUDE xc_section/vdW-DF2.sec`. That would be more explicit and the pre-processor could remain cleanly separate from the input structure.

dev-zero:
I'd prefer that too, especially since it will be automatically clear that it follows the same inclusion rules as `@INCLUDE`.

juerghutter:
Original idea was to simplify (and reduce errors) in input files. Final goal was to have a data base with the most common and complex functionals available and allow for a limited setting of parameters in the @XCTYPE structure. 
The first goal is still very much needed (see problems with vdW-DF and many hybrid functionals),

oschuett:
I agree, that the XC section is getting too complicated. However, bolting on a macro language is probably not the best solution. Instead we should evolve the existing input structure. We could e.g. build sth. like this:

```
&XC
   &CANNED vdW-DF2-b86r
      SOME_OPTIONAL_PARAMETER 42.0
   &END CANNED
&END XC
```

dev-zero:
> I agree, that the XC section is getting too complicated. However, bolting on a macro language is probably not the best solution. Instead we should evolve the existing input structure. We could e.g. build sth. like this:
> 
> ```
> &XC
>    &CANNED vdW-DF2-b86r
>       SOME_OPTIONAL_PARAMETER 42.0
>    &END CANNED
> &END XC
> ```

This would mean to place the specification of the functionals in the code with configuration parameters to overwrite coefficients?

oschuett:
> This would mean to place the specification of the functionals in the code with configuration parameters to overwrite coefficients?

Yes, exactly. I'd to write a routine `expand_canned_xc_section()` which make a bunch of calls to `section_vals_val_set()`. Since the expansion logic is in Fortran we could also support more complex mappings. And as a bonus we could automatically include the expanded section in the description of the canned section.

juerghutter:
This solution is an extension of the currently available short-cuts. (e.g. PBE0)
The problems I see for this solution is that it moves the definition of a functional to CP2K code (very obscure code). This is hard to maintain and debug. 
It is also not obvious how to handle the many different XC sections we have in the code.
In fact, we just have an internal new parser for the XC section.

mkrack:
I like the idea of having more xc functional shortcuts very much. In my opinion, a simple @INCLUDE of validated xc input sections in cp2k/data for common functionals is more than sufficient, since I do not see the additional benefit of having a new macro keyword besides stripping the path element "xc_section". I use the same approach via @INCLUDE for force fields, though that is less general, since a force field is not solely defined by the &FORCEFIELD section.

juerghutter:
The open question is how to handle system dependent variables (e.g. the cutoff for vdW functionals or parameters for HF). It is not optimal, if people have to copy the files from data to adjust those parameters. Still looking for an optimal solution.

dev-zero:
> The open question is how to handle system dependent variables (e.g. the cutoff for vdW functionals or parameters for HF). It is not optimal, if people have to copy the files from data to adjust those parameters. Still looking for an optimal solution.

1. extend the preprocessor syntax for variables in such a way that we can set defaults for variables (essentially a subset of the bash parameter substitution syntax, should be easy):

```
   ! evaluates to 0.2 if $ALPHA is not set, otherwise the content of $ALPHA:
   ALPHA ${ALPHA:0.2}

   ! evaluates to 0.2 if $ALPHA is not set or set to empty, otherwise the content of $ALPHA:
   ALPHA ${ALPHA:-0.2}
```

2. refactor our parser framework to provide a fully parsed configuration tree as a first step (opposed to the lazy parsing we're doing in some parts now) and then introduce anchors and references similar to [what YAML provides](https://en.wikipedia.org/wiki/YAML#Advanced_components) ... or switch to YAML ;-)

oschuett:
I'm really worried about usability. Our input structure is already very complex compared to other codes. Forcing new users to also deal with the preprocessor will make things worse.

> This solution is an extension of the currently available short-cuts. (e.g. PBE0)

Exactly, we don't have to reinvent the wheel. These short-cuts are discoverable and documented in the usual way through the manual.

> This is hard to maintain and debug.

While the XC  section could certainly profit from refactoring, the actual expansion code, e.g. for [PBE0](https://github.com/cp2k/cp2k/blob/master/src/input_cp2k_check.F#L154), looks very concise to me.

The translation of your input snippets into Fortran would be one-time effort that could be scripted. For the reverse we can use `section_vals_write()` and include the generated snippets in the descriptions.

juerghutter:
Okay, I checked what happens if shortcuts are combined with explicit input sections. Apparently it works without problems (limited sample space) and the shortcut definitions have priority.  
So it is not possible to change an explicit definition in the shortcut, e.g. the fraction of HFX in PBE0.
I agree, this might be the way to go.

oschuett:
We could probably add a `do_no_overwrite` flag to `section_vals_val_set` to invert the priorities. However, I'm not sure if this gives the best user experience. I think, it would be better to have shortcuts with parameters and then abort upon conflicts.

PR #563, opened 2019-09-17
Longrange potential for MP2 and RPA

fstein93:
This PR enables ...
- ... longrange potentials for MP2 and RPA. It is a first step to enable range-separated double-hybrid functionals with range-separation within the correlation part.
- ... Eigenvalue decompositions as an alternative for the Cholesky decomposition in RI. (necessary for longrange potentials)
- ... lots of regtests.

Misc:
Refactoring and Splitting of some modules in the MP2 code


oschuett:
Thanks for the refactoring the MP2 code. I saw that some of the new test are taking more than a minute to run. Maybe you could try to speed them up a bit?

oschuett:
PS: I'm trying to get all all the regtests to run in less than 30s, see also #174.

fstein93:
Ok, lets see whether they became fast enough.

fstein93:
It seems that all RPA. MP2, GW, and Double hybrid tests run in less than 30 seconds. @oschuett  What do you think?

oschuett:
Great! Thanks for going the extra mile :-)

PR #564, opened 2019-09-18
Fix for sirius options. More tests second try

mtaillefumier:
Something is fishy with the CI. The CI wants to compile thing during the convention check

this pull request contains a correction to an error introduced in the last SIRIUS pre-release. An include file was mistakenly included in the release which corrupted the list of options. Unfortunately, I realized the error too late and by snowball effect landed in cp2k.

- Updated the toolchain to the new SIRIUS prerelease version.

- cp2k now support full potential computations (through sirius). I had to force cp2k not to interpret the json potential file all the time.

- some improvements in the autotools build (well generate only the arch files)

- Added more tests for the SIRIUS backend. But we need more tests especially for magnetic materials and molecular dynamics computations.


dev-zero:
> Something is fishy with the CI. The CI wants to compile thing during the convention check

Nonono, that's perfectly normal 😂
The conventions checker also includes AST and compiler warnings analysis and thus requires compilation.
The conventions-checker run by the `pre-commit` hooks includes only the source code analysis part since I can't reliably only rebuild the changed files and I would like to keep the `pre-commit` lean and fast.

oschuett:
I realized that you are losing a lot of time because the toolchain takes an hour to reach `install_sirius.sh`. Now the CI caches the last successful step after a build error: https://github.com/cp2k/cp2k-ci/commit/bcb4e49cb90768ac08c4b51fb94bb9fb2f40f7c2.

oschuett:
Currently, the CI fails with the following message:
```
Installing from scratch into /opt/cp2k-toolchain/install/sirius-6.3.2
patch: **** Can't open patch file /opt/cp2k-toolchain/scripts/patch-deactivate-libvdwxc-mpi.patch : No such file or directory
ERROR: (./scripts/install_sirius.sh, line 96) Non-zero exit code detected.
```

The file `patch-deactivate-libvdwxc-mpi.patch` is not found because the [Dockerfile](https://github.com/cp2k/cp2k/blob/e30ba8a11931f6e6be41aa7a0c81b41672e44617/tools/toolchain/Dockerfile) does not copy it into the container. Try the following:
```
-  COPY ./scripts/install_sirius.sh ./scripts/
+  COPY ./scripts/patch-deactivate-libvdwxc-mpi.patch \
+       ./scripts/install_sirius.sh ./scripts/
```

PR #565, opened 2019-09-18
(f)prettify update

dev-zero:


oschuett:
While I'd prefer to simply copy the files, I think we should at least point the submodule at a release tag instead of master. I didn't realize we're doing the same for DBCSR. This means we're not keeping track of which versions belong together.

dev-zero:
> While I'd prefer to simply copy the files, I think we should at least point the submodule at a release tag instead of master. I didn't realize we're doing the same for DBCSR. This means we're not keeping track of which versions belong together.

No, that's not how submodules work, Git always tracks a specific commit unless instructed otherwise. `branch =` can not point to a tag and is only there for Git to know which branch to pull when using the `--remote` option for `git submodule update`.
See https://stackoverflow.com/a/18797720

oschuett:
Ok, my bad. Thanks for the link.

dev-zero:
@oschuett the following error from the formatting checker I can't reproduce locally by following the steps from `ci_entrypoint.sh`:

```
#################### Running Image cp2k-formatting ####################

========== Fetching Git Commit ==========
Submodule 'tools/prettify/fprettify' (https://github.com/pseewald/fprettify.git) registered for path 'tools/prettify/fprettify'
fatal: destination path '/workspace/cp2k/tools/prettify/fprettify' already exists and is not an empty directory.
fatal: clone of 'https://github.com/pseewald/fprettify.git' into submodule path '/workspace/cp2k/tools/prettify/fprettify' failed
Failed to clone 'tools/prettify/fprettify'. Retry scheduled
fatal: destination path '/workspace/cp2k/tools/prettify/fprettify' already exists and is not an empty directory.
fatal: clone of 'https://github.com/pseewald/fprettify.git' into submodule path '/workspace/cp2k/tools/prettify/fprettify' failed
Failed to clone 'tools/prettify/fprettify' a second time, aborting
```

I tried to reproduce it with the following and that worked like a charm:

```bash
git clone https://github.com/cp2k/cp2k.git
cd cp2k/
git fetch origin pull/565/head:pr565
git submodule update --init --recursive
git reset --quiet --hard pr565
git submodule update --init --recursive
```

... and for the other images it can update the Git checkout without any problems 🤔

oschuett:
This is caused by the `.pyc` files which are left behind from running `make pretty` during the image build.

You can reproduce this via:
```
$ cd /tools/docker
$ ./build_test.sh formatting
$  docker run  --env GIT_BRANCH=pull/565/head  --env GIT_REF=60fd674a97131f9102cafccccbb7c3ef21427944 img_cp2k_test_formatting:latest
```

How about putting the submodule into `/exts`?

dev-zero:
> This is caused by the `.pyc` files which are left behind from running `make pretty` during the image build.

I think this could be much faster if we'd instead run `prettify` only on the changed files (possibly via `pre-commit`). Maybe something like:

```bash
git diff --name-only HEAD..origin/master src | xargs pre-commit run --files
```
... then we don't need to warm-up the cache.

> You can reproduce this via:
> 
> ```
> $ cd /tools/docker
> $ ./build_test.sh formatting
> $  docker run  --env GIT_BRANCH=pull/565/head  --env GIT_REF=60fd674a97131f9102cafccccbb7c3ef21427944 img_cp2k_test_formatting:latest
> ```

thanks

> How about putting the submodule into `/exts`?

this will likely generate a different problem: given `.pyc` files in the leftover directory Python seems to load them first before a module with the same name in `/exts`. This is something I've encountered in a similar way during testing, likely because the directory of the script itself comes before other module include paths.

But this means manual intervention is needed by the user when upgrading, namely running `git clean -Xfd tools/prettify`

oschuett:
I don't know. The formatting image is already our fastest. Can't we do without the submodule and simply copy the [two files](https://github.com/pseewald/fprettify/tree/master/fprettify) that we actually need?

oschuett:
This new version of fprettify behaves a bit differently, e.g. by putting spaces around some operators. As a result the entire code base is modified. I just realized that this will lead to merge conflicts for any ongoing work. So, do you think it's possible to provide a script which helps people to rebase their local branches after we switched to the new fprettify?

Of course, the easiest solution would be to just stick to our current formatting conventions.

pseewald:
> This new version of fprettify behaves a bit differently, e.g. by putting spaces around some operators. As a result the entire code base is modified. I just realized that this will lead to merge conflicts for any ongoing work. So, do you think it's possible to provide a script which helps people to rebase their local branches after we switched to the new fprettify?
>

`git rebase -Xignore-all-space` should do the job, see 
https://git-scm.com/book/en/v2/Git-Tools-Advanced-Merging#_advanced_merging

dev-zero:
Either people should use @pseewald's suggestion or I can drop it completely now that we're testing only the changed files. The latter would mean that people still running `make pretty` instead of using `pre-commit` would get a large number of changed files, but the `pre-commit` hooks and the CI would only complain for changed but not prettified files.

dev-zero:
@pseewald I've tried `git rebase -Xignore-all-space feature/prettify-update` on a checkout of Augustin's xas_tdp_libint2 and it worked. The only drawback seems to be that it also reverted re-prettified lines not directly touched by Augustin to how they were before (without spaces between `+`/`-`).
So, what should be done is that during the rebase `pre-commit run --files ...` is run on each of the commits after applying them with `-Xignore-all-space`. Have to check, but this should be possible with a filter.

dev-zero:
ok, so for people who do not want to squash their PR (and then apply a `make pretty` on top of it), the following might be a possibility:

```bash
git rebase -Xignore-all-space -Xours \
  --exec 'git diff-tree --no-commit-id --name-only -r HEAD~1 | xargs pre-commit run prettify --files || git commit -a -n --amend --no-edit' \
  feature/prettify-update  # respectively "master" after merging this PR
```

... problem being that the conflict-resolution `-Xours` (which is needed if the commits were not prettified before and for example inserts in `USE` stmts are part of a commit) might overwrite legitimate changes. In that sense it would probably better to leave away `-Xours` and have people deal with rebases (and `prettifier`-induced conflicts) as before.

Nevertheless, the proposed command was tested on PR #570 (to rebase it on this PR).

oschuett:
When I run `make pretty` locally I get several changes. They seem to affect only `SELECT` and `USE` statements:

```
diff --git a/src/task_list_methods.F b/src/task_list_methods.F
index a923245ab..078b52191 100644
--- a/src/task_list_methods.F
+++ b/src/task_list_methods.F
@@ -66,7 +66,7 @@ MODULE task_list_methods
    USE util,                            ONLY: sort
 
 !$ USE OMP_LIB, ONLY: omp_destroy_lock, omp_get_num_threads, omp_init_lock, &
-!$                    omp_lock_kind, omp_set_lock, omp_unset_lock
+!$    omp_lock_kind, omp_set_lock, omp_unset_lock
 #include "./base/base_uses.f90"
 
    IMPLICIT NONE
@@ -1280,7 +1280,7 @@ CONTAINS
             ipair = (iatom - 1)*natom8 + (jatom - 1)
             cost = INT(tasks(5, itask))
 
-            SELECT CASE (tasks(4, itask))
+            SELECT CASE (tasks (4, itask))
             CASE (1)
                bit_pattern = tasks(6, itask)
                nopt = 0
```

... maybe because I'm using Python3?

dev-zero:
@oschuett which version of Python 3 are you using? I can't reproduce it so far, neither with 3.6.5, 3.7.5, 3.8.0, nor with 2.7.15:

```console
tiziano@tcpc18 ~/work/cp2k/cp2k (master=) $ python2 tools/prettify/prettify.py src/task_list_methods.F
tiziano@tcpc18 ~/work/cp2k/cp2k (master=) $ python3 tools/prettify/prettify.py src/task_list_methods.F
tiziano@tcpc18 ~/work/cp2k/cp2k (master=) $ git status
On branch master
Your branch is up to date with 'upstream/master'.
```

oschuett:
I'm sorry, this was a false alert. I had some `.pyc` files left in `tools/prettify/fprettify/`, which prevented the submodule cloning. Strange though, that it did not revert all formatting changes ¯\_(ツ)_/¯.

In any case, I'm very glad that we can now prettify fypp files.

pseewald:
> @oschuett which version of Python 3 are you using? I can't reproduce it so far, neither with 3.6.5, 3.7.5, 3.8.0, nor with 2.7.15:
> 
> ```
> tiziano@tcpc18 ~/work/cp2k/cp2k (master=) $ python2 tools/prettify/prettify.py src/task_list_methods.F
> tiziano@tcpc18 ~/work/cp2k/cp2k (master=) $ python3 tools/prettify/prettify.py src/task_list_methods.F
> tiziano@tcpc18 ~/work/cp2k/cp2k (master=) $ git status
> On branch master
> Your branch is up to date with 'upstream/master'.
> ```

For me there is an issue with python 2 and prettify.
```
python2 tools/prettify/prettify.py src/task_list_methods.F
```
```
diff --git a/src/task_list_methods.F b/src/task_list_methods.F
index a923245ab..a96a46d46 100644
--- a/src/task_list_methods.F
+++ b/src/task_list_methods.F
@@ -66,7 +66,7 @@ MODULE task_list_methods
    USE util,                            ONLY: sort
 
 !$ USE OMP_LIB, ONLY: omp_destroy_lock, omp_get_num_threads, omp_init_lock, &
-!$                    omp_lock_kind, omp_set_lock, omp_unset_lock
+!$    omp_lock_kind, omp_set_lock, omp_unset_lock
 #include "./base/base_uses.f90"
 
    IMPLICIT NONE
```

I test `fprettify` with python2 and python3 (and I checked again on this file), so I assume the issue comes from changes in the cp2k prettify script. Could you check again @dev-zero ?

PR #569, opened 2019-09-25
Calculate molecular electric moments

juerghutter:


src/molecular_moments.F
```diff
@@ -0,0 +1,278 @@
...
+                     j = first_atom+iatom-1
+                     atomic_kind => particle_set(j)%atomic_kind
+                     CALL get_atomic_kind(atomic_kind, kind_number=akind)
+                     CALL get_qs_kind(qs_kind_set(akind), ghost=ghost, floating=floating)
+                     IF (.NOT. ghost .AND. .NOT. floating) THEN
+                        CALL get_qs_kind(qs_kind_set(akind), core_charge=zeff)
+                        ria = particle_set(j)%r-rcc
+                        ria = pbc(ria, cell)
+                        moment_set(i, imol_now) = moment_set(i, imol_now)+ &
+                                                  zeff*ria(1)**lx*ria(2)**ly*ria(3)**lz
```
dev-zero, line 247: not sure whether it would be clearer, but a bit more concise and without need for the temporaries:

```fortran
moment_set(i, imol_now) = moment_set(i, imol_now) + zeff*PRODUCT(ria**indco(1:3, i+1))
```

juerghutter: It's not essential for performance, and this way reflects the basic equation better.

PR #570, opened 2019-09-25
Xas tdp libint2

abussy:
This pull request is for a new feature/method in CP2K named XAS_TDP (X-ray Absorption Sepctroscopy based on Time-Dependent density functional Perturbation theory). It allows for computation of core level excitation energies and oscillator strengths up to the hybrid-GGA TDDFT level of theory, for both molecules and solids. A RI scheme exploiting the locality of the core donor states is used, which should make calculation for large system possible and scalable (benchmarks are yet to be done though).

This method is quite independent from the rest of the code and thus, most of the new code is contained in the following new files: 

- `xas_tdp_types.F` definition of the types used throughout the method
- `xas_tdp_methods.F` general routines, I/O, etc.
- `xas_tdp_utils.F` the more computationally intensive routines 
- `xas_tdp_atom.F` where the numerical XC kernel integration takes place (on atomic grids)
- `xas_tdp_kernel.F` where the kernel matrices are built, the RI 3-center integrals are contracted, and generally all that is kernel related takes place (except for XC kernel integration)

Note that although TDDFT implementations where already around, it made sense to start from scratch as a lot of case specific optimization can be done (due to the nature of XAS). 

Along the way, some other part of the code were modified to fit the need of the method. In particular, the following features were added:
- 3-center integrals from libint2 (new file: `libint_3center.F`, modifications in `hfx_libint_wrapper.F` and `qs_o3c_methods.F` )
- 2-center Obara-Saika  integrals for the truncated Coulomb operator (modifications in `generic_os_integrals.F`, `ai_operators_r12.F`, moved `t_c_g0.F` to common/)
- Special neighbor lists where centers b of a,b pairs only cover a subset of all atoms (modifications in `qs_neighbor_lists.F`) 

The rest of the modifications to the main CP2K code are rather small and mostly involve making some routines public or the infrastructure needed for a new `RI_XAS` basis.

The code was successfully tested for consistency versus other programs (mostly Orca and some Turbomole) on small molecules.

Edit: the commit history was cleaned-up via git rebasing and regtests were added


src/generic_os_integrals.F
```diff
@@ -10,7 +10,8 @@
 !>        ii) (aba) and (abb) s-overlaps
 !> \par History
 !>      created [06.2015]
-!> \author Dorothea Golze
+!>      05.2019: Added truncated coulomb operator (A. Bussy)
+!> \author Doro/hea Golze
```
dev-zero: typo

src/hfx_libint_wrapper.F
```diff
@@ -22,13 +22,15 @@ MODULE hfx_libint_wrapper
...
+                                              C_NULL_PTR,&
+                                              C_FUNPTR
    USE kinds,                           ONLY: dp
 #if(__LIBINT)
    USE libint_f,                        ONLY: &
         libint2_build, libint2_build_eri, libint2_build_eri1, libint2_cleanup_eri, &
         libint2_cleanup_eri1, libint2_init_eri, libint2_init_eri1, libint2_static_cleanup, &
-        libint2_static_init, libint_t, libint2_max_am_eri
+        libint2_static_init, libint_t, libint2_max_am_eri, libint2_init_3eri, libint2_cleanup_3eri!,&
+        !libint2_build_3eri
```
dev-zero, line 33: leftover?

src/hfx_libint_wrapper.F
```diff
@@ -66,6 +69,10 @@ MODULE hfx_libint_wrapper
 #endif
    END TYPE
 
+   !This is a safeguard/hack because 3-center ERIs are bugged on libint 2.5
```
dev-zero: what is the nature of the safeguard here? Can it be removed at some point?

src/input_constants.F
```diff
@@ -1,3 +1,4 @@
+!TEMP COMMENT
```
dev-zero: ?

src/hfx_libint_wrapper.F
pseewald: This is a bug in libint 2.5 and resolved in libint 2.6. In order to preserve backwards compatibility with libint 2.5 (which was the first version of libint 2 compatible with cp2k), I suggested this solution. So we can remove it once we feel that there is no need to support libint 2.5.

src/input_cp2k_dft.F
```diff
@@ -6188,7 +6195,7 @@ SUBROUTINE create_print_embed_pot_cube(section)
                           description="The stride (X,Y,Z) used to write the cube file "// &
                           "(larger values result in smaller cube files). You can provide 3 numbers (for X,Y,Z) or"// &
                           " 1 number valid for all components.", &
-                          usage="STRIDE 1 1 1", n_var=-1, default_i_vals=(/1, 1, 1/), type_of_var=integer_t)
+                          usage="STRIDE 2 2 2", n_var=-1, default_i_vals=(/2, 2, 2/), type_of_var=integer_t)
```
dev-zero: This is not restricted to Xas, correct? Is changing the defaults here intentional? If yes, please factor out to separate commit as this may be relevant to put in the changelog as a separate item.

src/xas_tdp_atom.F
```diff
@@ -0,0 +1,3362 @@
...
+      REAL(dp), DIMENSION(:, :), POINTER                 :: ri_sphi, zet
+      REAL(dp), DIMENSION(:, :, :), POINTER              :: rhoa, rhob
+      TYPE(cell_type), POINTER                           :: cell
+      TYPE(grid_atom_type), POINTER                      :: grid_atom
+      TYPE(gto_basis_set_type), POINTER                  :: ri_basis
+      TYPE(harmonics_atom_type), POINTER                 :: harmonics
+      TYPE(particle_type), DIMENSION(:), POINTER         :: particle_set
+      TYPE(qs_kind_type), DIMENSION(:), POINTER          :: qs_kind_set
+
+      NULLIFY (qs_kind_set, r
```
dev-zero: should be a `parameter`, possibly factor out to `src/common/physcon.F` together with `velit` from `src/dkh_main.F`

src/xas_tdp_kernel.F
```diff
@@ -0,0 +1,1533 @@
...
+
+      ! Case study: UKS or ROKS ?
+      IF (xas_tdp_control%do_uks) THEN
+
+         ! In the case of UKS, donor MOs might be different for different spins. Moreover, the
+         ! fxc itself might change since fxc = fxc_sigma,tau
+         ! => Carfully treat each spin-quadrant separately
+
+         ! alpha-alpha spin quadrant (upper-lefet)
+         quadrants = .FALSE.; quadrants(1) = .TRUE.
```
dev-zero: wondering whether `quadrants = [.TRUE., .FALSE., .FALSE.]` might not be a bit clearer

src/xas_tdp_methods.F
```diff
@@ -0,0 +1,2964 @@
...
+
+!  Logger initialization and XAS TDP banner printing
+      NULLIFY (xas_tdp_section)
+      xas_tdp_section => section_vals_get_subs_vals(qs_env%input, "DFT%XAS_TDP")
+      output_unit = cp_logger_get_default_io_unit()
+
+      IF (output_unit > 0) THEN
+         WRITE (UNIT=output_unit, FMT="(/,T3,A,/,T3,A,/,T3,A,/,T3,A,/)") &
+            "!===========================================================================!", &
+            "!                              XAS TDP                                      !", &
```
dev-zero, line 171: you might want to follow tradition and add a fancy ASCII art banner here ;)

src/xas_tdp_methods.F
```diff
@@ -0,0 +1,2964 @@
...
+!           retrieve MO wave function center coordinates.
+            wfn_center(1:3) = localized_wfn_control%centers_set(ispin)%array(1:3, imo)
+            iat_memo = 0
+
+!           a large enough value to avoid bad surprises
+            dist_min = 10000.0_dp
+            DO iat = 1, n_atoms
+               at_pos = particle_set(iat)%r
+               r_ac = pbc(at_pos, wfn_center, cell)
+               dist = SQRT(r_ac(1)*r_ac(1)+r_ac(2)*r_ac(2)+r_ac(3)*r_ac(3))
```
dev-zero: use `NORM2`

src/xas_tdp_methods.F
```diff
@@ -0,0 +1,2964 @@
...
+      TYPE(dbcsr_p_type), DIMENSION(:), POINTER          :: matrix_ks
+      TYPE(mo_set_p_type), DIMENSION(:), POINTER         :: mos
+
+      NULLIFY (mos, mo_coeff, matrix_ks, para_env, blacs_env, lmo_struct, lmo_fm, ks_struct, ks_fm)
+      NULLIFY (evecs, work)
+
+      ! Get what we need from qs_env
+      CALL get_qs_env(qs_env, mos=mos, matrix_ks=matrix_ks, para_env=para_env, blacs_env=blacs_env)
+
+      nspins = 1; IF (xas_tdp_control%do_uks) nspins = 2
```
dev-zero, line 1215: please make it 2 lines

src/xas_tdp_methods.F
```diff
@@ -0,0 +1,2964 @@
...
+
+            ! Express the MOs in the basis that diagonalizes KS
+            CALL cp_gemm('N', 'N', nao, nlmo, nlmo, 1.0_dp, lmo_fm, evecs, 0.0_dp, work)
+
+            ! Replacing the new MOs back in the MO coeffs
+            i = 0
+            DO ilmo = 1, xas_tdp_control%n_search
+               IF (xas_tdp_env%mos_of_ex_atoms(ilmo, iat, ispin) == -1) CYCLE
+
+               i = i+1
```
dev-zero, line 1265: so, `i` is actually `ilmo-1`?

src/xas_tdp_methods.F
```diff
@@ -0,0 +1,2964 @@
...
+      END IF
+      ALLOCATE (my_mos(n_states, nspins))
+      ALLOCATE (max_overlap(n_states, nspins))
+
+!     Getting the atomic number
+      CALL get_qs_kind(qs_kind_set(donor_state%kind_index), zeff=zeff)
+      zval = INT(zeff)
+
+!     Electronic configuration (copied from MI's XAS)
+      ne =
```
dev-zero: typo (btw, GitHub shows the wrong lines in the overview, go to `View changes` to see them at the correct lines)

src/xas_tdp_utils.F
```diff
@@ -0,0 +1,3926 @@
...
+!> \param ilow ...
+!> \param iup ...
+!> \note  matrix is supposed to be in upper triangular form, and overwritten by this routine
+!>        subsets of eigenvalues/vectors can be selected by
+!>        specifying a range of values or a range of indices for the desired eigenvalues.
+!>        It was once in cp_fm_diag but got removed, moved it back here since it is needed for XAS_TDP
+!>        The eigenvalues/vectors are respectivly the first elements of the array/ columns of the matrix
+!>        Assumes scalapack version 2
+! **************************************************************************************************
+   SUBROUTINE cp_fm_syevr(matrix, eigenvectors, eigenvalues, ilow, iup)
```
dev-zero: this does belong (probably together with the above) in `src/fm/cp_fm_diag.F`, but have you seen 2ac9a3ec8c1cf4c53858129cf9a0fa87453e89a6 ?

src/hfx_libint_wrapper.F
abussy: This is also linked to your previous note about the commented `libint2_build_3eri` at the beginning of the file. In principle it should be imported, but it can be used directly if coming from libint 2.5

dev-zero: I've talked to @pseewald about that: for the CP2K v7.1 release we should drop that completely since it will be the first release requiring libint-2, at which point a requirement for a minimal version of libint-2 is reasonable (and will then be 2.6+).

What is needed here is a description of the workaround, something like: `manual declaration of libint2_build_3eri function since declaration provided by the Fortran bindings in libint-2.5.0 are wrong (fixed in libint-2.6+)`

src/xas_tdp_methods.F
abussy: Not quite. Because of the `IF` condition above, `ilmo` and `i` might evolve at different rates.

dev-zero: ah, that one I overlooked, sorry

src/xas_tdp_utils.F
abussy: Indeed it does. I moved it here (and cleaned it up a bit) after it was removed from master. I am aware that it crashes from time to time, especially for small matrices. This is also why this is not the default option in my code. 

However, I noticed some significant saving in larger systems where the diagonalization starts being relatively expensive and only a few eigenvalues/eigenvectors are needed.

I would personally would like to keep it. I am happy to move it back to `cp_fm_diag.F` too. We can probably keep it around for those specific application without offering it as an option for overall diagonalization in the code ?

src/input_cp2k_dft.F
abussy: Correct, I am not sure how it got there. I'll fix that

pseewald:
@oschuett my guess is that the conventions tester fails in compilation because it comes with a libint which was compiled without 3-center ERI support. 

Unfortunately, now that we use libint features that are not enabled by the default configure of libint, only our custom libint releases on https://github.com/cp2k/libint-cp2k are guaranteed to work with CP2K. I will update the CP2K documentation (INSTALL.md) to mention the exact requirements for libint after this PR was merged.

src/xas_tdp_atom.F
abussy: Done. Also refactored `cspeed` from `atom_operators.F`

abussy:
> @oschuett my guess is that the conventions tester fails in compilation because it comes with a libint which was compiled without 3-center ERI support.
> 
> Unfortunately, now that we use libint features that are not enabled by the default configure of libint, only our custom libint releases on https://github.com/cp2k/libint-cp2k are guaranteed to work with CP2K. I will update the CP2K documentation (INSTALL.md) to mention the exact requirements for libint after this PR was merged.

@pseewald @oschuett It seems that the convention check failed for another reason. I moved the faulty line of code within a `#if(__LIBINT)` construct, where it should have been from the start, and it now goes further.

It still fails, but for some other reason. I will probably have to update the `conventions.supp` file.

dev-zero:
> It still fails, but for some other reason. I will probably have to update the `conventions.supp` file.

yes, `BLAS/LAPACK/SCALAPACK` are always implicit calls since there's no definition for the functions

oschuett:
For LAPACK and BLAS we have a list of used routines in [analyze_gfortran_warnings.py](https://github.com/cp2k/cp2k/blob/master/tools/conventions/analyze_gfortran_warnings.py#L13). However, since `PDSYEVR` is controversial a single exception in [conventions.supp](https://github.com/cp2k/cp2k/blob/master/tools/conventions/conventions.supp) is probably better.

src/libint_3center.F
```diff
@@ -0,0 +1,319 @@
...
+!> \param rc ...
+!> \param lib the libint_t object for evaluation (assume that it is initialized outside)
+!> \param op which operator to use for the integrals
+!> \param omega in case of short range erfc operator
+!> \param r_cutoff the cutoff radius in case of truncated Coulomb
+!> \note Prior to calling this routine, the cp_libint_t type passed as argument must be initialized,
+!>       the libint library must be static initialized, and in case of truncated Coulomb operator,
+!>       the latter must be initialized too
+! **************************************************************************************************
+   SUBROUTINE eri_3center(int_abc, la_min, la_max, npgfa, zeta, ra, &
```
pseewald, line 72: For performance, you should apply screening to each pair of primitive Gaussians, for sure in a,b but also b,c and a,c in case of a truncated/short-range operator (see for instance [overlap3](https://github.com/cp2k/cp2k/blob/master/src/aobasis/ai_overlap3.F#L173-L188) subroutine).

src/qs_o3c_methods.F
```diff
@@ -303,6 +319,234 @@ SUBROUTINE calculate_o3c_integrals(o3c, calculate_forces, matrix_p)
...
+                  ALLOCATE (sabc(ncoa, ncob, ncoc))
+                  sabc = 0.0_dp
+
+                  CALL eri_3center(sabc, la_min(iset), la_max(iset), npgfa(iset), zeta(:, iset), &
+                                   ri, lb_min(jset), lb_max(jset), npgfb(jset), zetb(:, jset), &
+                                   rj, lc_min(kset), lc_max(kset), npgfc(kset), zetc(:, kset), &
+                                   rk, lib, op, omega=my_omega, r_cutoff=my_r_cutoff)
+
+                  IF (do_screen) THEN
+                     IF (MAXVAL(ABS(sabc)) < my_eps_screen) THEN
```
pseewald: Screening should be done on the magnitude of the final values of the contracted integrals. The same kind of screening is done in Hartree-Fock but weighted with the contraction coefficients, have a look at the following lines:
https://github.com/cp2k/cp2k/blob/master/src/hfx_energy_potential.F#L544
https://github.com/cp2k/cp2k/blob/master/src/hfx_libint_interface.F#L1613-L1616

src/qs_environment.F
```diff
@@ -792,6 +793,25 @@ SUBROUTINE qs_init_subsys(qs_env, para_env, subsys, cell, cell_ref, use_ref_cell
...
+         DO ikind = 1, nkind
+            NULLIFY (ri_xas_basis)
+            qs_kind => qs_kind_set(ikind)
+            CALL get_qs_kind(qs_kind, basis_Set=ri_xas_basis, basis_type="RI_XAS")
+            IF (.NOT. ASSOCIATED(ri_xas_basis)) THEN
+               CALL cp_warn(__LOCATION__, "Automatic Generation of RI_XAS basis. "// &
+                            "This is experimental code.")
+               !Because might need nco(l) with high l<=max_orbl in create_ri_aux_basis_set
+               CALL get_qs_kind_set(qs_kind_set, maxlgto=maxlgto)
+               CALL init_orbital_pointers(2*maxlgto)
```
pseewald: Do you think this could be taken care of in `create_ri_aux_basis_set`? It's a bit unfortunate that this needs to be done manually before creating an RI basis.

abussy: Yes indeed, especially since it could happen with the other RI basis sets as well. I'll move that bit of code

src/libint_3center.F
abussy: OK, but let's leave it like this for now as we discussed in real life, until you push your improved version of it.

src/qs_o3c_methods.F
abussy: Yes, indeed. I adapted

abussy:
I made a few changes triggered by the discussions above. Namely I moved back `cp_fm_syevr` to `cp_fm_diag.F` and warned the potential user in the routine description. I also warn the user of the XAS_TDP method that diagonalization might fail if the chosen method is the non-default SYEVR.

As for the 3-center integral part I fixed the pre-contraction screening to match with what is done in HFX. 

Finally I introduced a new `EPS_PGF_XAS` value for setting up neighbor lists in XAS_TDP, as the original `EPS_PGF_ORB` can be very small due to the HFX requirements (which essentially kill the performance of 3-center integral computation) .

I also ran with valgrind on the regtests and fixed a few uninitialized varaibles.

PR #580, opened 2019-10-09
XAS_TDP| setting RUN_TYPE to ENERGY in regtests to avoid forces

abussy:
Because we do not test for it (i.e. forces) and it fails for libint2 compiled with MAX_AM=4


abussy:
Note: I also renamed some !$OMP CRITICAL regions because it seems that intel compilers do not like when they are named after a variable.

PR #581, opened 2019-10-10
RI implementation of Hartree-Fock exchange

pseewald:
This is a working first implementation of RI Hartree-Fock exchange with the following RI metrics: coulomb, truncated coulomb, identity. Forces not yet implemented. Some optimizations (most importantly MO localization) will follow.


src/cp_control_types.F
```diff
@@ -479,7 +479,8 @@ MODULE cp_control_types
                                               auto_basis_aux_fit = 1, &
                                               auto_basis_lri_aux = 1, &
                                               auto_basis_ri_hxc = 1, &
-                                              auto_basis_ri_xas = 1
+                                              auto_basis_ri_xas = 1, &
+                                              auto_basis_ri_hfx
```
abussy: Does that mean that there is no default size for the RI_HFX basis ?

src/libint_3center.F
```diff
@@ -4,14 +4,15 @@
 !--------------------------------------------------------------------------------------------------!
 
 ! **************************************************************************************************
-!> \brief 3-center electron repulsion integral routines based on libint2
+!> \brief 2- and 3-center electron repulsion integral routines based on libint2
```
abussy: Maybe we might want to change the name of this module if it takes care of both 2-center and 3-center

src/libint_3center.F
```diff
@@ -94,13 +109,26 @@ SUBROUTINE eri_3center(int_abc, la_min, la_max, npgfa, zeta, ra, &
...
       NULLIFY (params, p_work)
       ALLOCATE (params)
 
+      dr_ab = 0.0_dp
+      dr_bc = 0.0_dp
+      dr_ac = 0.0_dp
+
+      IF (op == do_potential_truncated .OR. op == do_potential_short) THEN
+         dr_bc = r_cutoff*1.01
+         dr_ac = r_cutoff*1.01
```
abussy: Maybe a comment as to why a larger effective cutoff is required could be useful for the future ?

src/libint_3center.F
```diff
@@ -313,7 +350,214 @@ SUBROUTINE set_params(lib, ri, rj, rk, li_max, lj_max, lk_max, zeti, zetj, zetk,
...
+   SUBROUTINE set_params_2c(lib, rj, rk, lj_max, lk_max, zetj, zetk, &
+                            op, r_cutoff, omega, params_in, params_out)
+
+      TYPE(cp_libint_t), INTENT(INOUT)                   :: lib
+      REAL(dp), DIMENSION(3), INTENT(IN)                 :: rj, rk
+      INTEGER, INTENT(IN), OPTIONAL                      :: lj_max, lk_max
+      REAL(dp), INTENT(IN), OPTIONAL                     :: zetj, zetk
+      INTEGER, INTENT(IN), OPTIONAL                      :: op
+      REAL(dp), INTENT(IN), OPTIONAL                     :: r_cutoff, omega
+      TYPE(params_2c), OPTIONAL, POINTER                 :: params_in, params_out
```
abussy: As far as I can see there is no center permutation for the 2-center, does this `params_in` and `params_out` still make sense in this case ?

src/qs_o3c_methods.F
```diff
@@ -533,6 +537,11 @@ SUBROUTINE calculate_o3c_libint_integrals(o3c, op, t_c_filename, para_env, r_cut
          nsgfc => basis_set_c%nsgf_set
          sphi_c => basis_set_c%sphi
          zetc => basis_set_c%zet
+         rpgf_c => basis_set_c%pgf_radius
+
+         djk = SQRT(SUM(rjk**2))
+         dij = SQRT(SUM(rij**2))
+         dik = SQRT(SUM(rik**2))
```
abussy: You should use `NORM2` instead (also in other parts of your code)

src/cp_control_types.F
pseewald: It's initialized in `read_dft_control` with the same value `1` as the other basis types, but for consistency I updated this.

src/libint_3center.F
pseewald: Yes, I renamed it to `libint_2c_3c` (as well as a couple of other module names that have been generalized).

src/libint_3center.F
pseewald: This is needed due to the discontinuity of the truncated Coulomb potential at the truncation radius, this factor is now a properly documented constant in `libint_2c_3c` (you may want to use this too for the neighbor lists in `xas_tdp_utils`).

PR #586, opened 2019-10-16
bug fix in RPA RSE

rybkinjr:


oschuett:
I saw that you tried to run the `CUDA 2x Kepler (psmp)` check. Since it is not in daily use it often has to build the image first (which in your case got preempted several times). I actually just keep it around for multi-GPU testing.

Our main GPU check is `CUDA Pascal (psmp)`, which is also what I run for the dashboard.

PR #591, opened 2019-10-21
Finite difference for Linres Kernels

juerghutter:


src/input_cp2k_xc.F
```diff
@@ -1230,6 +1230,20 @@ SUBROUTINE create_xc_section(section)
       CALL section_add_keyword(subsection, keyword)
       CALL keyword_release(keyword)
 
+      CALL keyword_create(keyword, __LOCATION__, name="2ND_DERIV_ANALYTICAL", &
+                          description="Use analytical formulas or finite differences for 2nd derivatives of XC", &
+                          usage="2ND_DERIV_ANALYTICAL logical", default_l_val=.TRUE., &
+                          lone_keyword_l_val=.TRUE.)
```
dev-zero: just wondering: in which cases would one want to disable it?

PR #592, opened 2019-10-21
Start better "curating" benchmarks

### LINKED ISSUE #584: Performance test with H2O-DFT-LS and H2O-64-RI-MP2
Hi, I use one node with one Nvidia GPU to run the performance tests.
H2O-64, Fayalite-FIST and LiH-HFX can run finish in 2 hours. However, H2O-DFT-LS and H2O-64-RI-MP2 cannot and may hit some errors. Is there any limitations on these two performance tests if I use one node setting? 
I also found that there is no performance test data with one node setting in the following link.
https://www.cp2k.org/performance:piz-daint-h2o-dft-ls
https://www.cp2k.org/performance:piz-daint-h2o-64-ri-mp2

Here is the attached error log.
[H2O-DFT-LS-1-8.log](https://github.com/cp2k/cp2k/files/3727890/H2O-DFT-LS-1-8.log)
[H2O-64-RI-MP2-2-2.log](https://github.com/cp2k/cp2k/files/3727891/H2O-64-RI-MP2-2-2.log)

Thanks,
Vitesse.

### LINKED ISSUE #589: Curate and document benchmarks better
_This issue is a follow-up to the discussion started in #584._ 

@oschuett wrote:

> I do agree that our benchmarks could use some curating. I suggest we start by moving them into a separate root directory /benchmarks and add a welcoming top-level README.md. Most of the individual benchmarks actually have already a README file. So, these would only need to be converted to Markdown and touched up. For further inspiration we can draw on our [wiki page](https://www.cp2k.org/performance) and [PRACE](https://repository.prace-ri.eu/git/UEABS/ueabs/) incl. [this doc](http://www.prace-ri.eu/IMG/pdf/D7.5_5ip.pdf) starting page 27.

**To Dos**
- [x] create a new root directory `benchmarks/`, and move all the benchmarks out of `tests/` and to `benchmarks/`
- [ ] the root directory `tests/` then contains only regtests, not benchmarks
- [x] "add a welcoming" `benchmarks/README.md`
- [x] touch up existing READMEs and convert them to Markdown

@oschuett does this make sense? Anything you'd do differently?

shoshijak:
- Pull benchmarks out of `cp2k/tests` and into `cp2k/benchmarks`
- Touch up REAMDEs in `cp2k/tests` and `cp2k/benchmarks`
- Fix #589
- and fix #584


shoshijak:
@oschuett  I've moved the files that are obviously benchmarks to a root directory `benchmarks/ `, but there are some input files I'm not too sure about (see below).

#### A couple of questions

- `tests/Fist/*inp`(e.g. aala.inp, argon.inp, ethene.inp, water.inp): regtest/benchmark/other?
  - [ ] regtests
  - [ ] benchmarks
  - [ ] tests (not regtests nor benchmarks)
- `tests/LIBTEST/dbcsr_maxloc.inp`
  - [ ] is a regtest and should be added to `tests/LIBTEST/TEST_FILES`
  - [ ] is not a regtest, but should stay where it is
  - [ ] is not a regtest and should be moved to a different folder
- `tests/QMMM/QS/*.inp` (e.g. Ar-qmmm.inp, qmmm.inp): regtest/benchmark/other?
  - [ ] regtests
  - [ ] benchmarks
  - [ ] tests (not regtests nor benchmarks)
- `tests/QS/C.inp`: regtest/benchmark/other?
  - [ ] regtests
  - [ ] benchmarks
  - [ ] tests (not regtests nor benchmarks)

Also, are all input files in `tests` either a regtest or a benchmark? Or are some just plain tests? 
Should they be in the same directory as regtests or in a separate one?

oschuett:
Thanks for the great work! I don't think it makes sense to keep test inputs around that are not part of our regression suite. Since we never run them they will just bit rot. Hence, I'd suggest you use your best judgement to decided which tests should become regtests and throw the rest out. Then we can even add a convention check for strayed input files in `/tests`, which is actually a common mistake.

benchmarks/README.md
```diff
@@ -0,0 +1,11 @@
+# CP2K Benchmarks
+
+This directory contains input files for CP2K's benchmarks.
+
+For documentation on CP2K's input files, please refer to the [Input Reference Manual](https://manual.cp2k.org/).
+
```
oschuett, line 6: Maybe also add a link to https://www.cp2k.org/performance where we collect measurements from different machines.

shoshijak: Thanks, good point. Done in [de27e43](https://github.com/cp2k/cp2k/pull/592/commits/de27e43599b9e8cbc180093e7cfe6bf348c4b9ae).

oschuett:
Is this ready to be merged?

shoshijak:
Hi Ole, not quite. I still wanted to do what you suggested, namely: 

>  Hence, I'd suggest you use your best judgement to decided which tests should become regtests and throw the rest out. Then we can even add a convention check for strayed input files in /tests, which is actually a common mistake.

As well as copy some text from https://www.cp2k.org/performance to the benchmarks READMEs. 

I am out of office now and not sure I'll be able to complete this work in the coming two weeks. So you could merge this now, and I open a new PR in ~2 weeks with the missing work, or we could wait a little with this PR. Whichever you prefer.

oschuett:
Thanks for the update. Then I'll merge it now as it's already a big improvements.

alazzaro:
I think this PR made the dashboard unhappy:

https://storage.googleapis.com/cp2k-ci/dashboard_scaling_report.txt

oschuett:
Yes, I'm on it. Thanks!

shoshijak:
I'm sorry, I didn't think of the dashboard.

PR #595, opened 2019-10-28
LIBXSMM 1.14

hfp:
LIBXSMM 1.14. Removed Valgrind suppression (fixed since LIBXSMM 1.11).

### DIFF
--- a/tools/toolchain/install_cp2k_toolchain.sh
+++ b/tools/toolchain/install_cp2k_toolchain.sh
@@ -639,7 +639,7 @@ if [ $ENABLE_TSAN = "__TRUE__" ] ; then
         [ "$with_reflapack" = "__DONTUSE__" ] && with_reflapack="__INSTALL__"
         export FAST_MATH_MODE=reflapack
     fi
-    echo "TSAN is enabled, canoot use libsmm"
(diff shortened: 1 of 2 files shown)


hfp:
This PR fails because the source code package is not yet mirrored at CP2K.org.

oschuett:
Fixed.

hfp:
Thank you!

hfp:
Hmm, I cannot look into the generated output/logs. I can only see that all tests fail at runtime. I wonder what the root cause is....

oschuett:
It simply seg faults. I managed to copy a core dump into the artifacts. Hopefully, it provides you with a clue?

hfp:
Best would be the binary (cp2k.sdbg), then I can debug it right away.

oschuett:
Yes, I put the binary into the [artifacts](https://storage.googleapis.com/cp2k-ci/run-cp2k-sdbg-0ea15736_artifacts.tgz) as well.

hfp:
Thank you! I will update this PR with details later on.

hfp:
Now it passes. I think I will not go for a new release (I have still no reproducer) but simply disabled the memory interception in LIBXSMM. The problem seems to be related to some Sanitizer in the sdbg-build.

PR #596, opened 2019-10-28
Dashboard: Remove opt5 testers

oschuett:


alazzaro:
OPT5 was running the valigrind (SDBG).... This is definitely something we need to run once in a while (once per day?). Any plan on how to replace it?

oschuett:
Yes, I also consider the Valgind test a great loss. I have a rough plan for moving it to GCP, but this won't be ready for another couple of months.

alazzaro:
In principle, I can move it to Swan (single node)... I will keep this as an option.
BTW, we have to update the [Compiler support page](https://www.cp2k.org/dev:compiler_support), I think 4.9.2 was only checked bu UZH.... So, now the lowest checked compiler version will be 5. Am I right?

PR #610, opened 2019-11-07
Modernized array_sort and covered one more special case

hfp:


src/common/array_sort.fypp
```diff
@@ -21,19 +21,21 @@ SUBROUTINE ${prefix}$_sort(arr, n, indices)
...
+      indices = (/ (i, i = 1, n) /)
 
-   indices = (/ (i, i = 1, size(arr)) /)
+      CALL ${prefix}$_sort_low(arr(1:n), indices, tmp_arr, tmp_idx)
 
-   CALL ${prefix}$_sort_low(arr(1:n), indices, tmp_arr, tmp_idx)
-
-   DEALLOCATE(tmp_arr, tmp_idx)
+      DEALLOCATE(tmp_arr,tmp_idx)
+   ELSE IF (0 < n) THEN ! can be a frequent case in cp2k
```
oschuett: I'd prefer: 
```
IF (n==1) THEN
  indices(1) = 1
ELSE IF (n>1) THEN
   ...
END IF
```

hfp: Sure, can do. It was actually ordered by "likelyhood" (but mispredicting the branch is the least problem). Sorting appears to be not a hotspots (otherwise we should probably revisit allocating the scratch array recursively and rather only allocate the whole index set upfront and sort without temporary memory).

Although there is a SORT intrinsic in Fortran, it appears to be unable to indirectly sort. A quick C-call would probably cut-it (even C-sort allows taking the comparator as a function pointer and hence it can be used to sort indirectly without any temp storage other than the upfront index array).

If you want, I can go for the latter...

oschuett: Thanks, but I think our current solution is sufficient. I'm also not aware that sort showed up as hotspot.

I was just concerned with readability. If one reads `ELSE IF (0<n) indices(1) = 1` without the context then it looks rather confusing.

hfp: The `IF (0 < n)` is some "good style" adopted from C/C++. The "rule" is to have the constant value in front. The reasoning is `if (a == 1)` is sometimes (incorrectly/typo) written as `if (a = 1)`, which compiles in an unwanted fashion.

dev-zero: https://en.wikipedia.org/wiki/Yoda_conditions

got some attention on [Hackernews](https://news.ycombinator.com/item?id=20704028) lately, there are a lot threads on reddit about it. Personally I'm not a fan, but I don't care enough either.

hfp: @dev-zero Cool, I did not know this!

PR #626, opened 2019-11-14
Feature/curate benchmarks

shoshijak:
... follow up to #589 
(or rather, to: #592)


dev-zero:
@shoshijak add the `benchmarks` directory to the `DEFAULT_EXCLUDED_DIRS` list in `tools/conventions/analyze_src.py` to fix the issue

shoshijak:
Thank you! 
Still leaving this pr as "draft" for the moment, because I'd like to add some reference results for the `H2O-dft-ls` benchmark on daint

oschuett:
Looks great :-)

The convention checker was complaining about the non-ascii chars, e.g. `Å³`. At least in my browser this looks rather strange. Furthermore, not everybody might have the required fonts installed. Hence, I'm wondering if we shouldn't better stick to ascii?

shoshijak:
Ah, I see. Yes, I guess it would be cleaner to stick to ascii.

shoshijak:
Is it clear that `A3` means `Å³`? Or is `angstrom^3` better?

oschuett:
Thanks for making the change. A third option that comes to mind would be to write _cubic angstrom_.

However, this does not solve the general problem that we'll need a way to include formulas in our Markdown files. I hope that Github eventually adds support for it. Until then we could use [MathJax](http://docs.mathjax.org/en/latest/input/tex/html.html) syntax. Our users should all be able to read it and there are browser plugins too:
- https://github.com/orsharir/github-mathjax
- https://github.com/traversaro/github-mathjax-firefox

benchmarks/QS/README.md
```diff
@@ -0,0 +1,32 @@
...
+## Description
+
+Ab-initio molecular dynamics of liquid water using the Born-Oppenheimer approach, using [Quickstep](https://www.cp2k.org/quickstep) DFT. Production quality settings for the basis sets (TZV2P) and the planewave cutoff (280 Ry) are chosen, and the Local Density Approximation (LDA) is used for the calculation of the Exchange-Correlation energy. The configurations were generated by classical equilibration, and the initial guess of the electronic density is made based on Atomic Orbitals.
+
+## Benchmarks
+
+- [`H2O-32.inp`](H2O-32.inp): a system of 32 water molecules (96 atoms, 256 electrons) in a 9.9 cubic angstrom cell and MD is run for 10 steps
+- [`H2O-64.inp`](H2O-64.inp): a system of 64 water molecules (192 atoms, 512 electrons) in a 12.4 cubic angstrom cell and MD is run for 10 steps
+- [`H2O-128.inp`](H2O-128.inp): a system of 128 water molecules (384 atoms, 1'024 electrons) in a 15.6 cubic angstrom cell and MD is run for 10 steps
+- [`H2O-256.inp`](`H2O-256.inp`): a system of 256 water molecules (768 atoms, 2'048 electrons) in a 19.7 cubic angstrom cell and MD is run for 10 steps
```
oschuett: Broken link. See also below.

shoshijak: Thank you for catching this! It is fixed now.

shoshijak:
@oschuett anything missing or to be changed in this PR? 
As far as I can tell, it is ready to be merged.

PR #632, opened 2019-11-19
Unset executable flag [ci skip]

e-kwsm:


hfp:
The failing Formatting test is apparently unrelated to this change. This PR only updates the Git file index wrt the executable flag.

oschuett:
Should we maybe rollback https://github.com/cp2k/cp2k/commit/b7e6611d7a7da1c2fe66dd48e8d8950acd4fd13b? It makes the false assumption that the pre-commit checks passed when the docker image was build. And it also leads to the confusing _"checked 0 files, no violations found"_ message.

dev-zero:
> Should we maybe rollback [b7e6611](https://github.com/cp2k/cp2k/commit/b7e6611d7a7da1c2fe66dd48e8d8950acd4fd13b)?

Please don't: it is important that CI-formatting part and the `pre-commit` hook results are as close as possible.

> It makes the false assumption that the pre-commit checks passed when the docker image was build.

This should be a rather safe assumption to make. Otherwise either the code-base, the Docker generation or the way the code is pulled into the Docker image need to be fixed.

Looking at the error reports: for some reason did the changes in 
c638a72a2160783828cb6efcc554b90922347554 make it pass the previous way of running the conventions check. Any ideas why?

Running `pre-commit run -a analyze_src` reveals more violations. The ones mentioned in `tools/` can be simple ignored (will be gone after a commit included in PR #565), the other need to be either whitelisted or even better: logic added to not trigger when the flag is an inclusion guard.

> And it also leads to the confusing _"checked 0 files, no violations found"_ message.

Then please propose a fix to it.

oschuett:
> it is important that CI-formatting part and the pre-commit hook results are as close as possible.

I agree!

> This should be a rather safe assumption to make.

Generally, yes. However, things can slip through the cracks. For example, when the final commit differs from the merge that the CI used, or when the test scripts themselves change. Since performance doesn't matter here, we should really run the checks on all files. I believe, for Python we're already doing so.

>  logic added to not trigger when the flag is an inclusion guard.

It's already [there](https://github.com/cp2k/cp2k/blob/master/tools/conventions/analyze_src.py#L123), which is why the convention check is still passing. There's probably just a minor glitch in how the filenames are passed.

Since `analyze_src.py` is now run during the formatting check, should we officially move it into [tools/formatting](https://github.com/cp2k/cp2k/tree/master/tools/formatting) or even rename it to `tools/pre-commit`? In the future we'll probably add more checks, e.g. to validate Markdown, or check for broken links.

dev-zero:
> Since `analyze_src.py` is now run during the formatting check, should we officially move it into [tools/formatting](https://github.com/cp2k/cp2k/tree/master/tools/formatting) or even rename it to `tools/pre-commit`? In the future we'll probably add more checks, e.g. to validate Markdown, or check for broken links.

Sure, makes sense, I don't mind either way. Question is where to put the suppression list.
Validating Markdown would be nice, but the only `markdownlint` I found so far is Ruby-based. Broken links checker is a fickle thing: resources may return 429, 503, 504 or even 418 occasionally ;-)

oschuett:
I added #637 to continue our discussion about Markdown and teapots :-)

PR #635, opened 2019-11-19
Docker: Revert b7e6611 to fix formatting test

oschuett:


dev-zero:
I will take a look at the root cause tomorrow, see https://github.com/cp2k/cp2k/pull/632#issuecomment-555652457

PR #636, opened 2019-11-20
pre-commit/analyze_src: only check src/, ignore inclusion guards with filelist

dev-zero:


oschuett:
So, how about having the CI run the pre-commit checks always on all files?

dev-zero:
The issue in PR #632 was exactly related to the files touched in that commit. The cause was that `analyze_src` behaved slightly different when called from `pre-commit` (given an explicit list of files) and when run from the conventions-checker script (full-paths ended up as a source for the include guard check). Hence it works as intended. But I agree that it can be revisited should we find other issues directly related to the way we run it now.

oschuett:
I agree, that this PR will fix the issues encountered with #632.

>  But I agree that it can be revisited should we find other issues directly related to the way we run it now.

As I have [outlined](https://github.com/cp2k/cp2k/pull/632#issuecomment-555705513) there are other issues. So while we're at it, why not start running these checks on all files now? We are already doing exactly that for Python.

dev-zero:
> As I have [outlined](https://github.com/cp2k/cp2k/pull/632#issuecomment-555705513) there are other issues. So while we're at it, why not start running these checks on all files now? We are already doing exactly that for Python.

> It makes the false assumption that the pre-commit checks passed when the docker image was build.

After careful reviewing I still fail to see where this happens: since both the scripts and the sources are in the same repo and it gets updated directly before the run we should be consistent, no?

> confusing "checked 0 files, no violations found" message

IMHO I was always puzzled by the "checked 123 files, ..." when I touched only 2, but since everyone ignores that line anyway if it succeeded, I don't care.

oschuett:
The biggest problem occurs when the test itself or any of its dependencies change. The correct behavior would be to run the test on all files. However, with the [current setup](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_formatting.sh#L14) it might not run at all.

The simple and safe solution is to always run the test on all files. It will still be much quicker than the conventions and sdbg test. So, the users won't have to wait any longer.

PR #640, opened 2019-11-22
Libint 2- and 3-center refactoring + 4-center fix

abussy:
This PR can be seen as an extension of #634 

The 2- and 3-center integrals in `libint_2c_3c.F` have been slightly refactored. Changes include:
- passing a `hfx_potential_type` instead of a bunch of different arguments (omeg, cutoff_radius, etc.)
- primitive gaussian radius based screening for 2-center integrals
- addition of the identity operator (overlap)

The overlap integrals have been carefully tested with `ab_overlap` and `overlap3` as references. 

This development lead to the discovery that the 4-center overlap integrals in the HFX code were wrong, as something else than the (ss|ss) overlap was provided to libint as a starting point. It has been fixed using the knowledge gained through 2- and 3-center overlaps.

Note that this correction does not affect the regtests and should have no effect on previous results since (practically?) nobody actually uses the identity potential for HFX calculations


oschuett:
> Note that this correction does not affect the regtests and should have no effect on previous results since (practically?) nobody actually uses the identity potential for HFX calculations

I think, it would still be good to add a regtest. Alternatively, we could disable the feature via a `CPABORT`.

abussy:
I'm happy to add a new regtest with the identity potential. I am not a big fan of the `CPABORT` route though; it can be frustrating for users to be offered options that immediately trigger crashes (that is unless we also remove that option from the manual)

oschuett:
Yes, then please add a regtest. 

In general, removing an unused feature is also a very legitimate option. The ABORT is really just a last resort to protect our users from broken or unsupported combination of features that can't simply be removed from the input.

abussy:
@pseewald it looks like your `qs_tensors` fix was not quite enough yet

pseewald:
> @pseewald it looks like your `qs_tensors` fix was not quite enough yet

I did not fix anything yet, just a rebase...

abussy:
Sorry, my bad

abussy:
I am not sure what causes prettify to fail in the formatting check, it runs smoothly on my local machine (*i.e.* no file change)

pseewald:
libint interface should not depend on `hfx_types`, shall we rename `hfx_potential_type` and move it somewhere else? If it is libint specific, we could make it part of the libint interface.

abussy:
This seems reasonable, I'll have a look into it

PR #644, opened 2019-11-25
Enable odd number of electrons in S-ALMO

RubenStaub:
This patch allows systems with odd number of electrons to be
treated as closed-shell within the S-ALMO formalism. This is of
critical importance for non spin-polarized metallic slabs with an
odd number of electrons (e.g. p(3x3) Au(111) with 3 layers).

### DIFF
--- a/src/almo_scf.F
+++ b/src/almo_scf.F
@@ -287,26 +287,31 @@ CONTAINS
             multip = almo_scf_env%multiplicity_of_domain(idomain)
             nelec_a = (nelec + multip - 1)/2
             nelec_b = nelec - nelec_a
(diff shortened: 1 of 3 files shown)


dev-zero:
Would you have a sample input you could possibly turn into an additional regtest?

RubenStaub:
Since this is only a minor patch, I did not thought it requires a dedicated regtest (as such input would very much resemble the existing s-almo-no-deloc.inp (just replacing platinum by gold)).

But if you think it is worth some additional computational time for regtesting, then I will update this pull request including the new regtest.

Thanks for the update

PR #645, opened 2019-11-26
Sync input for CUTOFF and MULTIGRID_CUTOFF

mkrack:
- Allow units for the keyword MULTIGRID_CUTOFF in accordance with CUTOFF
- Abort CP2K if the same multi-grid cutoff is specified twice
- Improve error messages

### DIFF
--- a/src/cp_control_utils.F
+++ b/src/cp_control_utils.F
@@ -543,8 +543,9 @@ CONTAINS
 
          CALL section_vals_val_get(mgrid_section, "MULTIGRID_CUTOFF", r_vals=cutofflist)
          IF (ASSOCIATED(cutofflist)) THEN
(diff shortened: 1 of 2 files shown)


oschuett:
If you want to leave a longer abort message you can use:
```
 CALL cp_abort(__LOCATION__, "your lengthy message, "//&
               "which can even span multiple lines")
```

See also: https://www.cp2k.org/dev:error_handling

mkrack:
Yes, I know, but it seems it is not a string length problem. Maybe the "(NGRIDS)" in string poses an obstacle.

mkrack:
Without "(NGRIDS)" it works, but that is not a proof, since the string has been shortened at the same time.

PR #648, opened 2019-11-27
Add cosma as a replacement of the scalapack matrix multiplication

mtaillefumier:
Enclosed is a pull request that replaces the scalapack matrix - matrix calculation with  [cosma](https://github.com/eth-cscs/COSMA). It only affects the toolchain


tools/toolchain/scripts/install_cosma.sh
```diff
@@ -0,0 +1,143 @@
...
+export COSMA_CFLAGS="${COSMA_CFLAGS}"
+export COSMA_LDFLAGS="${COSMA_LDFLAGS}"
+export COSMA_CUDA_LDFLAGS="${COSMA_CUDA_LDFLAGS}"
+export CP_DFLAGS="\${CP_DFLAGS} IF_MPI(IF_OMP(-D__COSMA|)|)"
+export CP_CFLAGS="\${CP_CFLAGS} ${COSMA_CFLAGS}"
+export CP_LDFLAGS="\${CP_LDFLAGS} IF_CUDA(${COSMA_CUDA_LDFLAGS}|${COSMA_LDFLAGS})"
+export COSMA_LIBRARY="-lcosma"
+export COSMA_ROOT="$pkg_install_dir"
+export COSMA_INCLUDE_DIR="$pkg_install_dir/include"
+export PKG_CONFIG_PATH="$PKG_CONFIG_PATH:$pkg_install_dir/lib64/pkgconfig:$pkg_install_dir/lib/pkgconfig"
```
dev-zero: `PKG_CONFIG_PATH` (and some other variables) seems to end up twice in there, correct?
And you once define a `COSMA_LIBRARY` but there is also a `COSMA_LIBS` with a slightly different set?

tools/toolchain/scripts/install_cosma.sh
```diff
@@ -0,0 +1,143 @@
...
+            mv cosma cosma-${cosma_ver}
+            cd cosma-${cosma_ver}
+            mkdir build-cpu
+            cd build-cpu
+            case "$FAST_MATH_MODE" in
+                mkl)
+                cmake -DCMAKE_INSTALL_PREFIX="${pkg_install_dir}" \
+                          -DCOSMA_WITH_MKL_BLAS=ON \
+                          -DCOSMA_WITH_MKL_ScaLAPACK=ON .. \
+                          -DCOSMA_WITH_INSTALL=ON > cmake.log 2>&1
```
dev-zero: please either ensure that COSMA is using `lib` as libdir on all systems or adjust the paths referencing `lib/` in this script to also work with `lib64` instead, see other install scripts for similar issues

dev-zero (changes requested):

tools/toolchain/scripts/install_cosma.sh
dev-zero: please cleanup and/or clarify with a comment

oschuett:
Very exciting to see this one land. We should enable it for our [CUDA test](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/Dockerfile.cuda_mkl#L37).

oschuett:
I'm [adding](https://github.com/cp2k/cp2k/pull/652) single node bechmarks to the CI. Ideally, we get those in before this PR so that we'll have a nice record of COSMA's impact. Unfortunately, there seems to be a [problem](https://storage.googleapis.com/cp2k-ci/run-cp2k-benchmarks-cuda-kepler-502cc631_artifacts.tgz) with SIRIUS and (presumably) multi-GPU.

alazzaro:
What's the status of this PR?

kabicm:
This week we will have a new release of COSMA, that brings some performance improvements and bug-fixes. After that, this PR can be finalized.

kabicm:
A new release of [COSMA](https://github.com/eth-cscs/COSMA/releases/tag/2.0) is finally available. We ran the *RPA-RI* benchmark for 128 water molecules with `CP2K+COSMA` on 128, 256, 400, 512, 800 and 1024 nodes (Piz Daint, GPU partition, Cray XC50). Some of the results can be seen in the [README](https://github.com/eth-cscs/COSMA). It was not necessary to change the CP2K code, just linking CP2K to COSMA before linking to scalapack was enough.

alazzaro:
@kabicm Great!
Then, could you proceed with this PR?

tools/toolchain/scripts/install_cosma.sh
```diff
@@ -0,0 +1,155 @@
...
+EOF
+    fi
+    cat << EOF >> "${BUILDDIR}/setup_cosma"
+export COSMA_CFLAGS="${COSMA_CFLAGS}"
+export COSMA_LDFLAGS="${COSMA_LDFLAGS}"
+export COSMA_CUDA_LDFLAGS="${COSMA_CUDA_LDFLAGS}"
+export CP_DFLAGS="\${CP_DFLAGS} IF_MPI(IF_OMP(-D__COSMA|)|)"
+export CP_CFLAGS="\${CP_CFLAGS} ${COSMA_CFLAGS}"
+export CP_LDFLAGS="\${CP_LDFLAGS} IF_CUDA(${COSMA_CUDA_LDFLAGS}|${COSMA_LDFLAGS})"
+export COSMA_LIBRARY="-lcosma_with_scalapack -lgrid2grid -lcosma -loptions"
```
kabicm: `COSMA_LIBRARY` should be: 
`-lcosma_pxgemm -lcosma -lgrid2grid` for the CPU-only version and
`-lcosma_pxgemm -lcosma -lgrid2grid -lTiled-MM` for the GPU-accelerated version
After this, the linking to the chosen BLAS and SCALAPACK, that COSMA will be using, should follow. 

The proper linking to COSMA is explained in the [README](https://github.com/eth-cscs/COSMA), section *Using COSMA in 30 seconds*.

mtaillefumier:
I know. The pull request being open git hub will try to compile it all
the time I make a change in my repository.

mtaillefumier:
This pull request needs to wait for a fix in the cmake build system of cosma. cmake simply does not detect openblas when we want to build cosma with openblas.

hfp:
I also ran RPA-RI benchmark for 128 water molecules with CP2K+COSMA on CPU-only so far, but have consistent crashes. Is there a recommendation or limitation regarding ranks-per-node, memory requirements, etc.?

kabicm:
@hfp can you tell me on how many nodes it crashed? did you run it on daint-mc? which configuration did you run it with (e.g. how many ranks per node, how many threads per rank)? I will investigate it as soon as I have these details.

The RPA-RI benchmark I ran on at least 128 nodes of Piz Daint GPU partition. It is possible that runnning it on less nodes might exhaust the memory. COSMA is able to deal with less available memory, but the user might have to set the memory limit. If the memory limit is not set, COSMA assumes infinite memory, which might result in crashing. 

Also, feel free to open an issue in [COSMA repo](https://github.com/eth-cscs/COSMA).

hfp:
@kabicm I ran on 54 nodes, i.e., 108 sockets with 24 cores per socket (SMP enabled). My rank count was 24 per node, i.e., 12 per socket. This crashed very likely due to memory consumption. I meanwhile ran successfully with 72 node, 144 sockets, and only 4 ranks per node (2 per socket). I am interested in maximizing performance for a given/fixed set of resources, which means more ranks per socket than GPUs per socket.

Is there a way to set the memory limit using environment variables? I think this can be useful especially for the PxGEMM interface.

kabicm:
Thanks @hfp! This definitely sounds like a memory limit problem. At the moment, it is not possible to set it through the environment variable, but I will now open a new issue and will add this feature soon. This would be a small change in the code, so should not take me too much time.

hfp:
Thanks @kabicm!  I can grab your development version or master whenever this comes true and then retry with more ranks per socket. I can privately share some ScaLAPACK vs COSMA results based on this effort. I think the environment variable thing may also extend to tiles sizes. Btw, are these tiles always square, i.e., 1:1 ratio between extents?

kabicm:
Sounds great! Tile sizes that are used for local gemm calls are only relevant for the GPU version and it is already possible to set them by environment variables, e.g. with
`export COSMA_GPU_TILE_M=5000` 
`export COSMA_GPU_TILE_N=5000` 
`export COSMA_GPU_TILE_K=5000`
where `K` is the shared dimension and `MxN` is the dimension of matrix `C`. 

However, these are the maximal sizes that COSMA is allowed to use, and the actual sizes that COSMA will be using might be less or equal than these in specific cases.

It is also possible to set the number of GPU streams:
`export COSMA_GPU_STREAMS=2`

The values given here are the default values.

The device should have enough memory for this many elements:
`n_streams * (tile_m * tile_k + tile_k * tile_n + tile_m * tile_n)` 

This should be described in the documentation soon.

alazzaro:
@mtaillefumier I've just started the P100 and PDBG tests, both are failing:
1. P100 is a link problem
2. PDBG reports some leaks

kabicm:
I am working on those leaks, should be fixed soon.

mtaillefumier:
> @mtaillefumier I've just started the P100 and PDBG tests, both are failing:
> 
>     1. P100 is a link problem
> 
>     2. PDBG reports some leaks

Thanks for activating the pdbg build. I fixed the missing library. I am waiting for a new version of the library to make a commit again.

alazzaro:
@kabicm I'm using this PR for asking questions...
I'm testing COSMA, I have two questions:
1. Assuming that I link to OpenBLAS (no threaded), it is correct to say that when linking CP2K to COSMA, all DGEMM calls will not use OpenMP? (I mean, for instance for the RI-dRPA test) I mean, to be more specific, it is better to use for the RI-dRPA with 16 ranks and 4 threads or 4 ranks and 16 threads? (still, using single-threaded DGEMM).
2. Is there any way to make use of the multi-gpu? DBCSR does round-robin of the local ranks to the GPUs, while for COSMA I have to set CUDA_VISIBLE_DEVICES to the proper device, otherwise I always get device id 0 (and it will crash with DBCSR)...

kabicm:
@alazzaro Thanks for your questions!
1. COSMA uses OpenMP in 2 places: 
  - inside DGEMM calls (if linked to multithreaded DGEMM)
  - transforming the data layouts between the scalapack layout (block-cyclic) and the native COSMA layout (grid-like, but not necessarily cyclic).

So, it's hard to say which configuration is better. In my experience, it was never a good idea to oversubscribe cores (thus, on Piz Daint-mc, I would make sure #ranks*#threads=36). Having more ranks per node would improve the MPI communication performance, but having less ranks per node means that COSMA has more available memory per rank and thus can use a communication-optimal schedule. My guess would be that for small enough problems, 16 ranks x 4 threads might be better, but for larger problems (when memory becomes a bottleneck) 4 ranks x 16 threads might be better. On the other hand, 4 threads should be more or less enough to utilize the local memory bandwidth, so using it with single threaded dgemm makes sense as well.

2. At the moment, the multigpu is only possible if 1:1 ratio of GPUs and ranks is used. The main challange in multigpu is the following:
  - device id does not uniquely identify the GPU. 
  - each rank would need to know the unique device identifier (uuid) of all visible cuda devices and then local ranks would have to communicate to compare the uuid that are available on the node and to decide on how to split them. The function for querying the unique device identifiers is only available in CUDA 10 however. Also, I would have to check if ROCm offers the same functionality. I see people assuming that device id uniquely determines the device, but this approach might not work on all the platforms.

For these reasons, the multigpu support might take a bit longer to be implemented.

mtaillefumier:
Could someone trigger the psmp tests. Thanks.

alazzaro:
So, it seems that COSMA is working on the CPU side (few wrong results, I think they are OK).
For the GPU side, I think the problem is that now DBCSR and COSMA are competing for the same memory... I would say this is OK (need more memory)...

kabicm:
@alazzaro Regarding the CPU results, I checked the tests that have wrong results, for example `cp2k/tests/QS/regtest-ri-rpa/RI_RPA_minimax_H_atom.inp`.

The computed energy with cray-libsci and with cosma is very close to the tollerance. Can it be that the results are actually correct but that the tollerance is too small?

Regarding the GPU side: you are right, there is not enough memory, also because multiple ranks might be sharing the same GPU. This can be fixed by reducing the tile sizes just for the test purposes, e.g. by:
`export COSMA_GPU_TILE_M=100`
`export COSMA_GPU_TILE_N=100`
`export COSMA_GPU_TILE_K=100`
This would allocate little memory on GPU, so should not be a problem.

alazzaro:
For this CPU: I think this is OK. I leave others (@oschuett ?) to decide...

For the GPU: yes, I think this is a single GPU node and we have to reduce memory. @oschuett could you add the suggested variables to the test?

oschuett:
> I think this is a single GPU node and we have to reduce memory.

Our test cases are all very small and should run effortlessly on a single GPU. I therefore expect that users with serious workloads will see the same issue. The root cause is probably that COSMA and DBCSR are both holding on to their GPU buffers. I'm not sure if this is out of convenience or because device allocations are slow? In the later case the allocation might have to be moved into the host application, just like in the [old days](http://www.netlib.org/lapack/lug/node117.html).

In the meantime, we can set `COSMA_GPU_TILE_[MNK]` in the [toolchain](https://github.com/mtaillefumier/cp2k/blob/master/tools/toolchain/scripts/install_cosma.sh#L151).

> Can it be that the results are actually correct but that the tollerance is too small?

For `RI_RPA_minimax_H_atom.inp` I'd be willing to believe that. However, for the other four wrong results I'm not so sure, because they are off by multiple orders of magnitude. My guess would be that they are hitting some kind of special case in `pdgemm`, e.g. a zero sized block.

hfp:
I wonder if the memory consumption on CPU is also a problem. I mean if the expectation is that COSMA works whenever ScaLAPACK worked, it might be wrong (see [here](https://github.com/cp2k/cp2k/pull/648#issuecomment-599208544) and the related [issue](https://github.com/eth-cscs/COSMA/issues/27)).

I am concerned that CP2K+COSMA passes all tests but real workloads may run out of memory (because no test uses high enough local rank-count or may not have enough cores to bring-up rank count). Though, the [system](https://github.com/cp2k/cp2k/pull/648#issuecomment-599208544) mentioned had a reasonable per-core memory capacity.

Perhaps [user-control to hint local memory usage](https://github.com/eth-cscs/COSMA/issues/27) may be not enough and the default can be perhaps according to system metrics (or generally a small amount that can be assumed to be always available aka reasonable memory capacity per-core).

vondele:
let me quickly chime in, the GPU failures due to memory use will be fixed by @kabicm and result from allocating a 'tile' of memory, irrespective of problem size as a buffer. Makes sense with 1 rank per GPU (typically the best way to run cosma), but not if there are ~12 or so. Once this is fixed (or worked around temporarily) there should be no problem to run the regtests. 

Obviously cosma needs some more memory, that's the underlying idea, trade performance for memory. Users that want the same memory use as scalapack should just link scalapack, and that will only be for cases so big that scaling up will be needed for good time to solution anyway, and for which scaling will be efficient. There is no good way to derive default memory usage from system metrics, as we don't know if other processes will use memory (e.g. regtests) or even other parts of the same program will compete (e.g. dbcsr). I think this will anyway only be an issue for very larg scale runs, that have this problem more or less intrinsically. The environment flags will help, but if you need those, you're anyway living dangerously, memory consumption of the code is a tricky thing to control.

@kabicm will also have a look at the remaining wrong results. Could be another corner case, but zero-sized matrices are already dealt with in cosma... maybe more spots remain.

mtaillefumier:
I just added the environment variables for limiting the tile sizes. Please give it a try.

alazzaro:
From my initial experience with COSMA, I must say this library is a must-have to get performance for some benchmarks (very well done!). However, I would say we can make it no-default, so I'm not much concerned about memory...

The P100 test is now reporting some numerical problems (sorry, I hit "re-run" by mistake)...

oschuett:
I think we have two distinct problems at our hands:

1. Getting multiple CUDA accelerated libraries to co-exist. IIRC, allocating device memory requires only a synchronization, while page locking host memory is really expensive. This is the reason why DBCSR holds on to its buffers. Maybe the situation has improved with the recent arrival of [mlock2](http://man7.org/linux/man-pages/man2/mlock.2.html)? Since HPC processes shouldn't swap pages anyways, we could also consider using `mlockall()`.

2. Ensuring that COSMA behaves reasonable across all our use cases. I agree that it will be difficult to come up with good heuristics. Hence, we ultimately might have to provide the user with better controls, similar to what we have for [HFX](https://manual.cp2k.org/trunk/CP2K_INPUT/ATOM/METHOD/XC/HF/MEMORY.html#MAX_MEMORY) and [FM](https://manual.cp2k.org/trunk/CP2K_INPUT/GLOBAL/FM.html#list_TYPE_OF_MATRIX_MULTIPLICATION).

alazzaro:
> I think we have two distinct problems at our hands:
> 
> 1. Getting multiple CUDA accelerated libraries to co-exist. IIRC, allocating device memory requires only a synchronization, while page locking host memory is really expensive. This is the reason why DBCSR holds on to its buffers. Maybe the situation has improved with the recent arrival of [mlock2](http://man7.org/linux/man-pages/man2/mlock.2.html)? Since HPC processes shouldn't swap pages anyways, we could also consider using `mlockall()`.
> 2. Ensuring that COSMA behaves reasonable across all our use cases. I agree that it will be difficult to come up with good heuristics. Hence, we ultimately might have to provide the user with better controls, similar to what we have for [HFX](https://manual.cp2k.org/trunk/CP2K_INPUT/ATOM/METHOD/XC/HF/MEMORY.html#MAX_MEMORY) and [FM](https://manual.cp2k.org/trunk/CP2K_INPUT/GLOBAL/FM.html#list_TYPE_OF_MATRIX_MULTIPLICATION).

So, as far I understood COSMA does wrapping pdgemm calls out-of-the-box (very nice), so there is no need to change CP2K code at all or add any macro. This is what I find very useful. If you want to control that via a CP2K flag, then you have to change the code and you lose the beauty of having cosma out-of-the-box...

oschuett:
I agree, it's great that COSMA implements the pdgemm API. However, in CP2K we essentially only have a [single call site](https://github.com/cp2k/cp2k/blob/master/src/fm/cp_fm_basic_linalg.F#L448). Furthermore, we have a [wrapper](https://github.com/cp2k/cp2k/blob/master/src/cp_gemm_interface.F) to switch between gemm implementations. So, adding COSMA as another back end would require just a few lines of code. I believe this is preferable to disabling COSMA by default or asking users to keep multiple binaries around.

vondele:
@oschuett check email ;-)

kabicm:
A question about testing: why is maximum number of tasks set to 24 (the number of ranks per node is 2 and the number of threads per rank is also 2)?

oschuett:
We're using 24 vCPUs because that's the maximum GCP [allows](https://cloud.google.com/compute/docs/gpus#gpus-list) for a single P4 GPU.
We use only two [ranks](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest#L217) and [threads](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest#L235) in order to be frugal while still testing with some parallelism.

kabicm:
Thanks @oschuett! I've just sent you an email :)

tools/toolchain/scripts/install_cosma.sh
```diff
@@ -0,0 +1,176 @@
...
+
+        [ -d  ${pkg_install_dir}/lib64 ] && COSMA_LIBDIR="${pkg_install_dir}/lib64"
+
+        COSMA_CFLAGS="-I'${pkg_install_dir}/include'"
+        COSMA_LDFLAGS="-L'${COSMA_LIBDIR}' -Wl,-rpath='${COSMA_LIBDIR}'"
+        ;;
+esac
+if [ "$with_cosma" != "__DONTUSE__" ] ; then
+	if [ "$ENABLE_CUDA" = "__TRUE__" ] ; then
+    		COSMA_LIBS="-lcosma_pxgemm -lcosma -lgrid2grid -lTiled-MM"
```
oschuett: Here, the `-lTiled-MM` needs to be wrapped in `IF_CUDA` too.

mtaillefumier:
Can someone trigger the mpi+gpu test.

kabicm:
If P100 are expensive, maybe we can run them after we ensure that other tests have passed. 

What might be causing the docker problem: the thing that changed is that Tiled-MM library, which is a dependency of COSMA has been updated, so, maybe it tries to use the previuos version of Tiled-MM, which does not match with the commit that COSMA is expecting from Tiled-MM?

kabicm:
@vondele just found one more corner case that needs to be added in COSMA, so those wrong results might still be a bug.

oschuett:
The P100 test cost about twice as much as the P4 test. So, we can definitely run it when needed. 

When there are changes to the toolchain, it's advantagous to run the P4 test first. It has more vCPUs and will build the toolchain image quicker, which the P100 test can afterwards reuse.

alazzaro:
@kabicm Well done, you are very close to fix all problems. Now there are only some leaks in the communicators...

kabicm:
Hopefully, the new version fixes the leaks. Is there some systematic reason why the docker image fails to build from time to time?

If this version passes the tests, then I can add the growing tiles, so that we do not need to specify the sizes explicitly.

oschuett:
> Is there some systematic reason why the docker image fails to build from time to time?

It's hard to tell from the short error message what's going on, but I'd be rather certain that the problem is in `install_cosma.sh`. Did you try building the container locally?

mtaillefumier:
I added a typo in the script. It should be fixed now.

tools/toolchain/scripts/install_cosma.sh
```diff
@@ -0,0 +1,176 @@
...
+[ -f "${BUILDDIR}/setup_cosma" ] && rm "${BUILDDIR}/setup_cosma"
+
+! [ -d "${BUILDDIR}" ] && mkdir -p "${BUILDDIR}"
+cd "${BUILDDIR}"
+
+case "$with_cosma" in
+    __INSTALL__)
+	require_env OPENBLASROOT
+	require_env SCALAPACKROOT
+        export
```
oschuett: Left over debug line.

mtaillefumier: thanks. I will correct it next patch. Waiting for the results of the regtests.

alazzaro:
All tests passed, time for the final change... and merge...

mtaillefumier:
What is the final stage ?

alazzaro:
> What is the final stage ?

For the "Final change" I mean what Ole requested today. Is it already done?

tools/toolchain/scripts/install_cosma.sh
oschuett: IIRC, I stumped over this while trying to build `psmp` with a cuda enabled toolchain.

I believe, the simplest solution would be to just remove lines 134-138 and replace it with the definition from line 146:

```
COSMA_LIBS="-lcosma_pxgemm -lcosma -lgrid2grid IF_CUDA(-lTiled-MM|)"
```
Also note that all references to `${COSMA_FOO}` are replaced immediately, because they are not escaped. Hence, the lines starting with `export COSMA_` can be remove.

mtaillefumier:
I do think we have to keep the hack about fixed tiles. Simply because the CI tests are not a real use case for cosma. The growing tile is needed but we can maybe just update cosma to a new version on the fly. The most important is that it gives the answer we expect and that users can benefit from it. Growing tile or not will not change this hopefully.

kabicm:
@vondele I discussed this today shortly with @mtaillefumier. The script definining the tile sizes is just for the CI, and the toolchain will build COSMA with whatever the default tile sizes are. And, as Mathieu said, since the CI is not a representative of a real use, it make sense to update the version after merging. But however you decide, for me the most important part was to pass the tests :)

kabicm:
Actually, I should soon be finished with these growing tiles, so it's probably better to wait until tomorrow or so, to have the proper version of COSMA.

alazzaro:
My two cents.
I will merge this PR as it is now. I agree that as initial work we have to test that it is working. With the merging, we can get more coverage on the dashboard (and I can install the library on Daint).
I leave to @oschuett to decide.

kabicm:
I think I just finished debugging. So, just need to polish up the code and will publish the new version with growing tiles during the day.

kabicm:
COSMA `v2.0.7` with growing tiles is now published.

The environment variables were renamed as follows:
`COSMA_GPU_TILE_M -> COSMA_GPU_MAX_TILE_M` and similar.

By default these will be set to `5000`, and the tiles will grow up to that value depending on the problem sizes. So, if the problem sizes in tests are not too large, there is probably no need to define those at all. On the other hand, defining those is probably a `nice` behaviour :)

oschuett:
> So, if the problem sizes in tests are not too large, there is probably no need to define those at all.
> On the other hand, defining those is probably a nice behaviour :)

If possible we should remove those tile bounds. After all, the toolchain is also commonly used for production builds.

kabicm:
@oschuett I agree :) I just thought that maybe having the variables defined might hint to people looking at the script that it's actually possible to control the maximum GPU memory usage, but in that case, they should be defined to `5000` or so, since that's the default size that is to be used in production.

oschuett:
> I just thought that maybe having the variables defined might hint to people looking at the script that it's actually possible to control the maximum GPU memory usage.

I doubt that many users will look there, but feel free to leave whatever comments you like.
Personally, I still think an input section similar to [DBCSR](https://manual.cp2k.org/trunk/CP2K_INPUT/GLOBAL/DBCSR.html) would be the best solution.

kabicm:
@oschuett that's a good idea, to just comment out the tile sizes and maybe write some small comment as a hint. 

So, you mean that the variable should be renamed to: `MAX_ELEMENTS_PER_BLOCK`? I can certainly do that if that sounds better.

Btw, I explained the variables names in the README: [tunable parameters section](https://github.com/eth-cscs/COSMA#tunable-parameters)

kabicm:
@oschuett Now I realized what you meant. I can write that section later for CP2K and make a PR.

oschuett:
> So, you mean that the variable should be renamed to:

No no, `COSMA_GPU_MAX_TILE_[MNK]` is a good name, please keep it. I was merely advertising (again) a tighter integration between cosma and cp2k. We already discussed this at length, so feel free to ignore that comment ;-)

> comment out the tile sizes and maybe write some small comment as a hint.

Alternatively or in addition, you could also mention the environment variables in `INSTALL.md`

vondele:
so, let me chime in again. COSMA_GPU_MAX_TILE_M should not be set as part of the toolchain. I agree with @oschuett that this is how people run in production. I also expect that peope will not need to set those values, too many performance related knobs are a pain for the user (and actually developer), be it as env variables or as part of the input. I'd propose we remove setting those variables in the toolchain, and see if it passes. If it does, we merge...time to see what the code does in the wild, afterwards we can improve if needed.

mtaillefumier:
I would also remove those variables. I kept them for the regtests only.

vondele:
@mtaillefumier point is the regtests should now pass without these variables, and that we better prove :-)

kabicm:
Wohoo, passed! :)

oschuett:
Great job everybody!

PR #652, opened 2019-11-28
Add single node benchmarks to the CI

oschuett:


alazzaro:
It looks like I need this PR... 
The performance tests on Daint (see dashboard) are now based on the new directory benchmarks/. However, it turns out that some of the input files are outdated, for instance:

https://github.com/cp2k/cp2k/blob/master/benchmarks/QS_mp2_rpa/32-H2O/RI-MP2.inp

gives

```
found an unknown keyword METHOD in section WF_CORRELATION    
```

Need a refresh of these files under benchmarks...

oschuett:
I'm shelving this for the time being.

PR #657, opened 2019-12-01
Toolchain fixes for avx512 and selinux

mkrack:


tools/toolchain/scripts/install_libvdwxc.sh
```diff
@@ -67,15 +68,15 @@ case "$with_libvdwxc" in
                 ./configure \
                     --prefix="${pkg_install_dir}" \
                     --libdir="${pkg_install_dir}/lib" \
-                    --with-fftw3 \
+                    --with-fftw3=${FFTW_ROOT} \
```
dev-zero, line 71: have you seen a specific error triggering this change? I'm asking because we see a strange build failure on CentOS 7 with a custom GCC in libvdwxc due to some missing symbol in `fftw_mpi`.
And in principle this is not needed since we set `CPATH` and `LIBRARY_PATH` in the environment (at least when building fftw ourselves, didn't check what happens if someone provides a system fftw).

tools/toolchain/scripts/install_openblas.sh
```diff
@@ -45,7 +45,7 @@ case "$with_openblas" in
             cd OpenBLAS-${openblas_ver}
 
             for patch in "${PATCHES[@]}" ; do
-                patch -p1 < "${SCRIPT_DIR}/${patch}"
+                patch -p1 < "${SCRIPT_DIR}/${patch}" || [ $(getenforce) = 'Enforcing' ] && true || false
```
dev-zero: why do we have to hide errors from `patch`?

tools/toolchain/scripts/install_libvdwxc.sh
mkrack: All these fixes were needed for a successful toolchain built on our Merlin6 cluster (CPU with AVX512 capabilities and selinux fully enforced). There is also system fftw installed and the libvdwxc configure cannot find a file (I think it was fftw_mpi.h). After I specified the search path for fftw explicity, i.e. the one to be used with the toolchain, as suggested by the help of the libvdwxc configure, the built works fine.

tools/toolchain/scripts/install_openblas.sh
mkrack: The patch command completes with an error when selinux is Enforcing and the toolchain run stops, since the context of the patched files could not be changed. Do you know any better solution for that issue?

tools/toolchain/scripts/install_libvdwxc.sh
dev-zero: ah, so exactly the case with the system-provided fftw, thanks!

tools/toolchain/scripts/install_openblas.sh
schulkov: This is strange. Users' files are usually have the 'unconfined' context, so they are exempt from SELinux rules. Besides, non-SELinux distributives lack the 'getenforce' command.

dev-zero: Do you still have the original error message?
I guess the way distros do it is to make sure that the files have the proper permissions/contexts/ownership after extracting such that when applying a patch no context change should occur.

I couldn't reproduce on our CentOS 7 with SELinux enforcing.

mkrack: The patch command
```
patch -p1 < /data/user/krack/github/mkrack/cp2k/tools/toolchain/scripts/openblas-0.3.6-disable-avx512.patch
```
causes the error message
```
patching file kernel/x86_64/KERNEL.SKYLAKEX
patch: setting attribute security.selinux for security.selinux: Permission denied
```

dev-zero: @mkrack can you check whether running `restorecon -rv tools/toolchain/` prior to `install_cp2k_toolchain.sh` and/or running it explicitly in the OpenBLAS script after unpacking the tarball changes the behaviour?

The proposed patch needs some work, I get `getenforce: command not found`. Maybe combining with `[ -f /sys/fs/selinux ]` might be a way around, unless you want to explicitly check the command first using `command -v getenforce > /dev/null`

mkrack: restorecon returns also a permission denied error
```
restorecon reset /data/user/krack/github/mkrack/cp2k/tools/toolchain/scripts context unconfined_u:object_r:unlabeled_t:s0->unconfined_u:object_r:default_t:s0
restorecon set context /data/user/krack/github/mkrack/cp2k/tools/toolchain/scripts->unconfined_u:object_r:default_t:s0 failed:'Permission denied'
```
and thus does not help.

schulkov: It seems that this a problem of the 'patch' program itself, that tries to modify the protected 'security.selinux' extended attribute directly using setxattr(2). The current method to fix it is to recompile patch using the '--disable-xattr' configure option. By the way, setxattr(2) syscall fails on the 'security.selinux' attribute even when SELinux operates in permissive mode.

Just curiosity, which Linux distribution do you use (the latest Fedora?). The patch program in CentOS 7 seems to be unaffected.

mkrack: @schulkov thanks for the explanation.
The cluster node is running the Red Hat Enterprise Linux Server release 7.6 (Maipo).

mkrack: @dev-zero I added the explicit check for the availability of the getenforce command as suggested, though I do not understand why you get an error on systems without getenforce, since that branch with getenforce is only processed when the patch command fails even in the prior version.

dev-zero: @mkrack thanks. I should have to be more precise: I get the cited error if the patch command fails, but given that in the future people might not know about this whole thing anymore it is a good idea to not add unrelated error output on top...

dev-zero: @mkrack it might be a good idea to open an issue with either RedHat (or your IT dept?). I have the impression that you're missing an SELinux rule allowing a user a context switch or that the `/data` directory is incorrectly tagged: I've tried reproducing the error with a CentOS 7 with various combinations of setting `chcon -t unlabeled_t` on either the patch, the target directory or the base directory (with a default of `default_t`) where I build CP2K, to no avail. But I think that for some reason you're ending up with a mix of contexts (they're usually only preserved, not explicitly set) without the permission to change between them...
Btw, what do you get when you do `ls -lisaZ tools/toolchain/scripts tools/toolchain/build`?

mkrack: All files and folders in `tools/toolchain/scripts` and `tools/toolchain/build` have the context
```
unconfined_u:object_r:unlabeled_t:s0
```
I will check with our Admins.

PR #659, opened 2019-12-03
start: fix compilation with Intel with compiler_options intrinsic

dev-zero:


dev-zero:
@oschuett it is unclear to me where this Git ID is comming from in the Conventions check, any ideas? ... I already tried to restart.

dev-zero:
seems that yet another restart solved the issue

PR #660, opened 2019-12-03
toolchain: libint update

pseewald:


pseewald:
The CI stops at random places during libint build without apparent errors, can you have a look @oschuett?

oschuett:
I'll have a look.

oschuett:
The CI kills the job because it uses more than 24 GB of memory.

oschuett:
Building libint2 with lmax=5 takes now 32min on a beefy machine. I'm getting a bit worried...

mkrack:
@pseewald Why do you want to use a pre-release of libint now?

pseewald:
@oschuett the recommended way to build libint is now cmake together with ninja (i‘m still building with cmake & make) - i‘m not sure whether that will make a difference but I‘ll test it.

@mkrack officially cp2k should be used with a released version of libint, but I‘d like to test new versions of libint ahead of their release with cp2k (since we were having problems with broken releases of libint in the past).

oschuett:
It's a good idea to check pre-release versions!

So, then let's try ninja. After all, the libint wiki says it's _**strongly** recommended_. If that doesn't work, should we maybe tweak the default settings of the toolchain?

Unfortunately, the [problem](https://github.com/evaleev/libint/issues/63) seems to be low priority for the libint developer.

Btw, what's your opinion on [libcint](https://github.com/sunqm/libcint)?

oschuett:
Libint [2.7.0-beta.5](https://github.com/evaleev/libint/releases/tag/v2.7.0-beta.5) is out. Should we give this another try?

pseewald:
Yes I'll take care of it. I'd like to avoid adding ninja as yet another dependency to the toolchain script since I hope we switch to a more maintainable solution (spack) in the future.

dev-zero:
I've just build-tested the current libint-master (internally tagged as 2.7.0-beta6): building with GNU make is still supported, as is the two-stage build process with autoconf+cmake.
But there is [a recently updated PR](https://github.com/evaleev/libint/pull/148) which is going to change that: there will still be a two-stage build process, but only with CMake. E.g. it will be what they already described on their wiki as the build process.
From inspecting the build artefacts it seems that the Fortran library gets built correctly (again).

oschuett:
Since Patrick is off to other adventures I'm closing this PR.

PR #663, opened 2019-12-03
Fix leak

mkrack:


dev-zero:
No, it's not, it's an allocatable. Which compiler and version was complaining?

mkrack:
`cp2k.pdbg -v` built with the current toolchain returns
```
==89841==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 2779 byte(s) in 1 object(s) allocated from:
    #0 0x2aed13e268c8 in __interceptor_malloc /data/user/krack/github/mkrack/cp2k/tools/toolchain/build/gcc-8.3.0/libsanitizer/lsan/lsan_interceptors.cc:52
    #1 0x45e6a6 in cp2k /data/user/krack/github/mkrack/cp2k/src/start/cp2k.F:248
    #2 0x45ec5b in main /data/user/krack/github/mkrack/cp2k/src/start/cp2k.F:45
    #3 0x2aed1756b494 in __libc_start_main (/lib64/libc.so.6+0x22494)
```
It seems that `compiler_options_string` is implicitly allocated. After adding the DEALLOCATE, the message disappears. You relaxed the compiler warnings from lhs-realloc-all to lhs-realloc to suppress the corresponding warning about an LHS allocation. The [template provided by Intel](https://software.intel.com/en-us/fortran-compiler-developer-guide-and-reference-compiler-options-2) for the usage of compiler_options() also includes that DEALLOCATE.

dev-zero:
> `cp2k.pdbg -v` built with the current toolchain returns
> 
> ```
> ==89841==ERROR: LeakSanitizer: detected memory leaks
> 
> Direct leak of 2779 byte(s) in 1 object(s) allocated from:
>     #0 0x2aed13e268c8 in __interceptor_malloc /data/user/krack/github/mkrack/cp2k/tools/toolchain/build/gcc-8.3.0/libsanitizer/lsan/lsan_interceptors.cc:52
>     #1 0x45e6a6 in cp2k /data/user/krack/github/mkrack/cp2k/src/start/cp2k.F:248
>     #2 0x45ec5b in main /data/user/krack/github/mkrack/cp2k/src/start/cp2k.F:45
>     #3 0x2aed1756b494 in __libc_start_main (/lib64/libc.so.6+0x22494)
> ```
Thanks!

> It seems that `compiler_options_string` is implicitly allocated. After adding the DEALLOCATE, the message disappears. You relaxed the compiler warnings from lhs-realloc-all to lhs-realloc to suppress the corresponding warning about an LHS allocation.

Yes, because that's one of the points of `ALLOCATABLE`s, that you don't have to explicitly allocate or deallocate them anymore. I've encountered these warnings before: for some reason does the GNU compiler sometimes omit deallocation code (if the function is only called once, the memory will be cleaned up at exit).
But it seems we're not the only ones encountering this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86754

> The [template provided by Intel](https://software.intel.com/en-us/fortran-compiler-developer-guide-and-reference-compiler-options-2) for the usage of compiler_options() also includes that DEALLOCATE.

That's really unfortunate, because the standard (F2018, 9.7.3.2/2) says: "When the execution of a procedure is terminated by a execution of a RETURN or END statement, an unsaved allocatable local variable of the procedure retains its allocation and definition status if it is a function result or a subobject thereof; otherwise, if it is allocated it will be deallocated."

dev-zero:
It seems these kind of leaks only occur inside a `PROGRAM`, as soon as these kind of `ALLOCATION`s are inside a procedure they get cleaned up properly.

```fortran
program main
    USE iso_fortran_env, ONLY: compiler_options, compiler_version

#ifdef as_function

    call printem()

contains
    subroutine printem()

#endif
        character(len=:), allocatable :: compiler_options_string
        integer :: i
   
        compiler_options_string = compiler_options()
        PRINT "(T2,A,A)", "compiler: ", compiler_version()
        PRINT "(T2,A)", "compiler options:"
        DO i = 0, (LEN(compiler_options_string) - 1)/68
            PRINT "(T4,A)", &
                compiler_options_string(i*68 + 1:MIN(LEN(compiler_options_string), (i + 1)*68))
        END DO

#ifdef as_function
    end subroutine
#endif
end program
```

```console
$ gfortran -Wl,--enable-new-dtags -ffree-form -std=f2008 -fimplicit-none -fsanitize=leak -fcheck=all -fno-omit-frame-pointer -g -o comp_version_leak comp_version_leak.F90 
$ ./comp_version_leak 
 compiler: GCC version 7.4.1 20190905 [gcc-7-branch revision 275407]
 compiler options:
   -cpp -mtune=generic -march=x86-64 -g -std=f2008 -ffree-form -fimplic
   it-none -fsanitize=leak -fcheck=all -fno-omit-frame-pointer

=================================================================
==9382==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 127 byte(s) in 1 object(s) allocated from:
    #0 0x7f0f0558cd48 in malloc (/usr/lib64/liblsan.so.0+0xed48)
    #1 0x4009c8 in MAIN__ /.../comp_version_leak.F90:15
    #2 0x400d26 in main /.../comp_version_leak.F90:26
    #3 0x7f0f04680f89 in __libc_start_main (/lib64/libc.so.6+0x20f89)

SUMMARY: LeakSanitizer: 127 byte(s) leaked in 1 allocation(s).
$ gfortran -Das_function -Wl,--enable-new-dtags -ffree-form -std=f2008 -fimplicit-none -fsanitize=leak -fcheck=all -fno-omit-frame-pointer -g -o comp_version_leak comp_version_leak.F90
$ ./comp_version_leak 
 compiler: GCC version 7.4.1 20190905 [gcc-7-branch revision 275407]
 compiler options:
   -cpp -D as_function -mtune=generic -march=x86-64 -g -std=f2008 -ffre
   e-form -fimplicit-none -fsanitize=leak -fcheck=all -fno-omit-frame-p
   ointer
```

PR #664, opened 2019-12-04
benchmarks: add results on many nodes for DFT-LS and RPA

shoshijak:


shoshijak:
Ping? 
If no one objects, can we merge this?

(I don't have merge rights..)

shoshijak:
Thank you!

PR #677, opened 2019-12-12
Rollback to DBCSR 2.0.1

pseewald:
This partially reverts 2824b54caf63b9adad79685dd3db4f417b4ac058

This pr should be merged as a preparation for a new CP2K release in case DBCSR 2.1.0 is not released by then.

### DIFF
--- a/src/rpa_gw_im_time_util.F
+++ b/src/rpa_gw_im_time_util.F
@@ -401,12 +401,10 @@ CONTAINS
       ALLOCATE (sizes_MO(nmo))
       sizes_MO = 1
 
-      CALL dbcsr_t_pgrid_create(para_env%group, pdims, pgrid, &
-                                tensor_dims=[SIZE(sizes_RI_split), SIZE(sizes_AO_split), SIZE(sizes_MO)])
+      CALL dbcsr_t_pgrid_create(para_env%group, pdims, pgrid)
 
       pdims_2d = 0
-      CALL dbcsr_t_pgrid_create(para_env%group, pdims_2d, pgrid_2d, &
-                                tensor_dims=[SIZE(sizes_AO), SIZE(sizes_MO)])
+      CALL dbcsr_t_pgrid_create(para_env%group, pdims_2d, pgrid_2d)
 
       CALL create_3c_tensor(t_3c_overl_int_ao_ao, dist1, dist2, dist3, pgrid, &
                             sizes_RI_split, sizes_AO_split, sizes_AO, [1, 2], [3], "(RI AO | AO)")
@@ -450,8 +448,7 @@ CONTAINS
 
       IF (do_ic_model) THEN
          pdims = 0
-         CALL dbcsr_t_pgrid_create(para_env%group, pdims, pgrid_ic, &
-                                   tensor_dims=[SIZE(sizes_RI_split), SIZE(sizes_MO), SIZE(sizes_MO)])
+         CALL dbcsr_t_pgrid_create(para_env%group, pdims, pgrid_ic)
 
          CALL create_3c_tensor(t_3c_overl_int_mo_ao, dist1, dist2, dist3, pgrid_ic, &
                                sizes_RI_split, sizes_MO, sizes_AO, [1, 2], [3], "(RI MO | AO)")


oschuett:
I'd actually like to merge this PR soon to revive our [CUDA test](https://dashboard.cp2k.org/archive/cuda-pascal/index.html), which is having the following [problem](https://dashboard.cp2k.org/archive/cuda-pascal/commit_4a9b0ec864bb2e905699cf422adbf3066dfcf5e0.txt):
```
nvcc -c -arch sm_60 -O3 -Xcompiler='-fopenmp' --std=c++11 -D__ACC -D__DBCSR_ACC -D__PW_CUDA  -D__LIBXSMM  -D__parallel  -D__SCALAPACK -D__FFTW3  -D__SPFFT -D__LIBINT -D__LIBXC -D__GSL -D__SPGLIB -D__HDF5 -D__LIBVDWXC -D__SIRIUS   -I'/workspace/cp2k/exts/dbcsr/src' /workspace/cp2k/exts/dbcsr/src/acc/libsmm_acc/libsmm_acc_benchmark.cpp
In file included from /workspace/cp2k/exts/dbcsr/src/acc/libsmm_acc/libsmm_acc_benchmark.h:16:0,
                 from /workspace/cp2k/exts/dbcsr/src/acc/libsmm_acc/libsmm_acc_benchmark.cpp:14:
/workspace/cp2k/exts/dbcsr/src/acc/libsmm_acc/../hip/acc_hip.h:13:10: fatal error: hip/hip_runtime.h: No such file or directory
 #include <hip/hip_runtime.h>
          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Unfortunately, this PR  surfaced a [another problem](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-pascal-60199164_report.txt) which seems to be related to incremental builds:
```
/workspace/cp2k/exts/dbcsr/src/acc/libsmm_acc/libcusmm/libcusmm_benchmark.cu:15:10: fatal error: parameters.h: No such file or directory
 #include "parameters.h"
          ^~~~~~~~~~~~~~
compilation terminated.
```

I'll try if forcing a complete rebuild is a viable workaround for the second problem.

Nevertheless, both problems are apparently rooted in DBCSR's Makefile. So, instead of fixing it we might as well spend the time on https://github.com/cp2k/cp2k/issues/351.

alazzaro:
Let's go in order:

1. v2.0.1 is there and well tested for inclusion in CP2K, so I agree with the rollback to that version
2. current develop of DBCSR is not yet integrated into CP2K, i.e. it requires more testing from my side
3. if you really want to do that, you just need to add `-D__CUDA`. I've already tested it on Daint. However, this is just a temporary solution, I'm planning to have this macro automatically. Note that the makefile in DBCSR will be only used by CP2K, with 2.1 the official way to build the library will be cmake and the makefile will be deprecated. The plan is to remove it with v3.0 and make DBCSR fully external (yes, #351).

In conclusion, if you want to close CP2K by the end of the year, the only solution I see is to go for v2.0.1...

oschuett:
Yes, I also think we should roll back to v2.0.1 now. Unfortunately, then there will still be the issue with the missing `parameters.h`. That one is actually quite puzzling to me. From the git history this PR would seem like a clean rollback. Maybe some subtle change in the Ubuntu base image?

With my last comment I meant to say that we might have underestimated the maintenance cost for keeping the DBCSR Makefile around when we decided to wait with #351 for Spack.

alazzaro:
I think the parameters part will appear if you compile with -D__CUDA...

oschuett:
But why did it work before? We didn't change the toolchain. Hence, we're using the same arch files.

alazzaro:
This is a new macro, so you have to change the arch file...

pseewald:
> This is a new macro, so you have to change the arch file...

So when was this macro introduced?

Maybe we never tested CUDA with toolchain after this macro was introduced, so the fix would be to add this flag to the toolchain-generated arch files?

shoshijak:
> So when was this macro introduced?

It was introduced with the HIP backend to DBCSR, as a way to indicate whether the HIP or the CUDA backend is to be compiled.

alazzaro:
OK, I think it is clear that we have to use 2.0.1...

oschuett:
I think, I found the cause: The code from `src/acc/libsmm_acc/libcusmm/` was moved one directory level up. After rolling back the generated files `parameters.h` and `smm_acc_kernels.h` were left behind in the new location. The problem should disappear once we merge this PR and rebuild the docker image.

PR #678, opened 2019-12-12
Enable longrange exchange HF potentials for atomic calculations

fstein93:
- Use Fortran intrinsics for Bessel functions
- Add longrange Hartree-Fock for atomic calculations
- Regtests for the different functionalities


src/atom_types.F
```diff
@@ -1157,6 +1180,166 @@ SUBROUTINE release_atom_orbs(orbs)
...
+
+      counter = 0
+      DO ii = 1, 2*nn
+         IF (diag(ii) > 0.0_dp) THEN
+            counter = counter + 1
+            abscissa(counter) = diag(ii)
+            weights(counter) = rootpi*eigenvec(1, ii)**2
+         END IF
+      END DO
+      CPASSERT(counter == nn)
```
dev-zero: I'd probably have used something like `CPASSERT(ALL(diag > 0.0_dp))` prior to the loop to separate error-handling and algorithm, but I guess that's just style. But what I really miss for these kind of checks is a small comment giving the developer/user some ideas on what went wrong on how to fix it.

src/atom_utils.F
```diff
@@ -1438,32 +1455,96 @@ SUBROUTINE potential_numeric(cpot, density, nu, grid)
...
+!> \brief ...
+!> \param cpot ...
+!> \param density ...
+!> \param nu ...
+!> \param grid ...
+!> \param omega ...
+!> \param kernel ...
+!> \param do_gh ...
+! **************************************************************************************************
+   SUBROUTINE potential_longrange_numeric(cpot, density, nu, grid, omega, kernel, do_gh)
```
dev-zero: This function is actually two functions: you could move the branch up the callstack and split this one instead.

src/atom_utils.F
```diff
@@ -1477,172 +1558,290 @@ SUBROUTINE potential_analytic(cpot, la, lb, nu, basis)
...
+         ELSE
+            erfa = 2.0_dp*znFn(z, expa, nu)
+         END IF
+
+         erfa = erfa*oab
+      ELSE
+         erfa = 0.0_dp
+      END IF
+
+      DEALLOCATE (expa, z)
```
dev-zero: possible crash here depending on data & compiler: `ALLOCATE` is conditional so you might end up deallocating unallocated data.

src/atom_utils.F
```diff
@@ -1477,172 +1558,290 @@ SUBROUTINE potential_analytic(cpot, la, lb, nu, basis)
...
+   END SUBROUTINE potential_longrange_analytic
+
+! **************************************************************************************************
+!> \brief Boys' function times z**n
+!> \param z ...
+!> \param expa ...
+!> \param n ...
+!> \return ...
+! **************************************************************************************************
+   ELEMENTAL FUNCTION znFn(z, expa, n) RESULT(res)
```
dev-zero, line 1814: `PURE` maybe?

src/atom_types.F
fstein93: I will add more comments about that! CPASSERT(ALL(diag > 0.0_dp)) is not possible because the eigenvalues will be symmetric wrt the origin in this case and I will need only the positive roots. (This algorithm calculates the roots of the Hermite polynomials that are even or odd functions.)

dev-zero: ah, I see, sorry. But yes, please a comment.

src/atom_utils.F
fstein93: Is solved

src/atom_utils.F
fstein93: Done.

src/atom_utils.F
dev-zero: well, I realized that `ELEMENTAL` already implies `PURE` and there's the `IMPURE` instead

PR #680, opened 2019-12-13
PINT: Change logger iteration level for correct force_eval output

cschran:
This commit restores the correct behaviour for replica output in the pint run_type. This is needed, for example, to correctly print force_eval properties for each path integral bead.

### DIFF
--- a/src/motion/pint_methods.F
+++ b/src/motion/pint_methods.F
@@ -1666,6 +1666,10 @@ CONTAINS
       NULLIFY (logger)
       logger => cp_get_default_logger()
 
+      ! initialize iteration info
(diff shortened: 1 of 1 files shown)


cschran:
The Formatting test fails, because there seems to be a problem on the test machine. The job can't find cp2k-formatting and tries building it, but fails in step 6 with ```unable to execute 'x86_64-linux-gnu-gcc': No such file or directory```

Could someone with access to the machine have a look at that @oschuett or @dev-zero? During the formatting test of pull request #679 cp2k-formatting was still available.

oschuett:
I fixed the formatting check via #683.

PR #690, opened 2019-12-17
Fix GCC 5 (smp)

fstein93:
This PR should fix the issues with the failing Dashboard test.


fstein93:
If everything passes, may I have an additional test "Ubuntu, GCC 5 (ssmp)"?

PR #696, opened 2019-12-18
Makefile: re-use REVISION to allow override, warn user

### LINKED ISSUE #694: obscure Make errors when .git directory is missing
Building CP2K from a recent git checkout, without the .git subdirectory (in my case because I was using [EasyBuild](https://github.com/easybuilders/easybuild)) results in obscure Make errors that took me a long time to figure out.

The cause is the following: `tools/build_utils/get_revision_number` exits with return code 1, but does not print an error message.   As a result, the build fails somewhere in the middle of compilation, Make reports "Error 1", but it's not clear what caused the error.

One of the following changes to `get_revision_number` or the Makefile could make life easier:
- print (on stderr) an error message that explains the problem if the .git directory is missing, or
- keep the script as it is, but use `exit 0` after echoing `<unknown revision>`, so the build can continue (instead of current `exit 1`). Perhaps print a warning on stderr just to inform the user.
- allow the user to proviide GIT_REVISION as a Make variable, and don't run `get_revision_number` if that variable is set.

dev-zero:
fixes #694 by warning the user via stderr and by allowing to override
the revision by setting REVISION explicitly as a `make` argument (the
warnings will still be shown since GNU make seems to execute the shell
commands prior to determining whether it is needed at all, but that is
somewhat intentional)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -379,7 +379,7 @@ doxygen: doxygen/clean
 	@echo "<html><body>Sorry, the Doxygen documentation is currently being updated. Please try again in a few minutes.</body></html>" > $(DOXYGENDIR)/html/index.html
 	cp $(ALL_SRC_FILES) $(DOXYGENDIR)
 	@for i in $(DOXYGENDIR)/*.F ; do mv $${i}  $${i%%.*}.f90; done ;
-	@cat $(TOOLSRC)/doxify/Doxyfile.template | sed "s/#revision#/`$(TOOLSRC)/build_utils/get_revision_number $(CP2KHOME)`/"  >$(DOXYGENDIR)/Doxyfile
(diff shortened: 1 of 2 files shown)


schulkov:
It is because the ':=' assignment (in contrast with the '=' assignment) instructs make to expand the expression immediately. A workaround is to put the ':=' assignment inside an `ifndef .. endif` block:

```
ifndef (REVISION)
REVISION     := $(shell $(CP2KHOME)/tools/build_utils/get_revision_number $(SRCDIR))
endif
```

dev-zero:
@schulkov thanks. If I call it like `make REVISION=123 ...` it will always use that `REVISION`, so there would be no need evaluate the shell at all. But I think I'll leave it as is since people should be aware that they're doing something which is not fully supported.

Issue #673, opened 2019-12-10
xTB + RTP/EMD fail with multiples of 8 MPI ranks in release, 6 MPI ranks in debug


vitesselin:
Dear all,
Is there any limitation on these cases?

[h2o_emd.inp.txt](https://github.com/cp2k/cp2k/files/3944201/h2o_emd.inp.txt)
[h2o_rtp.inp.txt](https://github.com/cp2k/cp2k/files/3944202/h2o_rtp.inp.txt)
[H2O-NMR-4.inp.txt](https://github.com/cp2k/cp2k/files/3944203/H2O-NMR-4.inp.txt)

I set total number of message passing processes to 2, 4, 8, 12 and only hit abort in 8.

Reproduce Steps:
cd tests/xTB/regtest-3
CUDA_VISIBLE_DEVICES=0 OMP_NUM_THREADS=2 /home/vitesse/cp2k/mpich/bin/mpirun -np 8 ../../../exe/local_cuda/cp2k.psmp h2o_rtp.inp
CUDA_VISIBLE_DEVICES=0 OMP_NUM_THREADS=2 /home/vitesse/cp2k/mpich/bin/mpirun -np 8 ../../../exe/local_cuda/cp2k.psmp h2o_emd.inp

cd tests/QS/regtest-nmr-1/
CUDA_VISIBLE_DEVICES=0 OMP_NUM_THREADS=2 /home/vitesse/cp2k/mpich/bin/mpirun -np 8 ../../../exe/local_cuda/cp2k.psmp H2O-NMR-4.inp 2>&1 | tee H2O-NMR-4.inp.txt

On the architecture/host/platform:
Intel(R) Xeon(R) Gold 6142 CPU @ 2.60GHz + V100

CP2K error messages:
h2o_rtp.inp and h2o_emd.inp
```
 *** WARNING in xtb_matrices.F:1088 :: Atomic charges outside chemical   ***
 *** range were detected. Switch-off CHECK_ATOMIC_CHARGES keyword in the ***
 *** &xTB section if you want to force to continue the calculation.      ***

 *******************************************************************************
 *   ___                                                                       *
 *  /   \                                                                      *
 * [ABORT]                                                                     *
 *  \___/                               xTB Charges                            *
 *    |                                                                        *
 *  O/|                                                                        *
 * /| |                                                                        *
 * / \                                                     xtb_matrices.F:1091 *
 *******************************************************************************


 ===== Routine Calling Stack =====

[...]

PR #700, opened 2019-12-20
Disable real-time propergation with xTB

oschuett:
This is only a mitigation. See https://github.com/cp2k/cp2k/issues/673 for details.


oschuett:
No longer needed since  #701 saved the day.

PR #701, opened 2019-12-20
Fix cp_fm_struct_double

schulkov:


schulkov:
It does not fix #673 at the moment, as there is also some issue with `cp_dbcsr_sm_fm_multiply()` when used with very small matrices.

#673 can be fixed by replacing a doubled matrix `fm_tmp1` in `aspc_extrapolate()` by two standard matrices. However, it will introduce communication overhead that we would like to avoid.

schulkov:
The issue #673 should now be fixed.

schulkov:
Is it possible to re-run Linux-x86-64-gfortran.pdbg regression tests? The reported errors seems to be spurious, as none of the amended functions are actually used by the failed tests. Besides, cp2k.pdbg built using the toolchain passes all regression tests locally.

PR #705, opened 2020-01-06
Current

juerghutter:


oschuett:
It seems this PR disabled the OpenMP parallelization in [generate_qs_task_list](https://github.com/cp2k/cp2k/blob/4e2c386c4db90f97de60fb7d02db261e6adf7be2/src/task_list_methods.F#L251).

Furthermore, it introduced an `OMP DO` outside of a parallel region to [sap_sort](https://github.com/cp2k/cp2k/blob/4e2c386c4db90f97de60fb7d02db261e6adf7be2/src/sap_kind_types.F#L242), which makes the Cray compiler [unhappy](https://dashboard.cp2k.org/archive/cce865-psmp/commit_4e2c386c4db90f97de60fb7d02db261e6adf7be2.txt).

PR #711, opened 2020-01-08
Enabled ADMM with GAPW

abussy:
This PR aims at enabling ADMM together with the GAPW method.

The main challenge was to define, compute and store the `AUX_FIT` atomic soft and hard densities. This was achieved  with minimal refactoring by storing alternative types (*e.g.* a new `qs_kind_set`) in the `admm_env` and generalizing GAPW routines to accept external variables (for example by using an external `qs_kind_set` and not the default `qs_env%qs_kind_set`). See `admm_types.F` for the alternative types and the `calculate_vxc_atom` subroutine in `qs_vxc_atom.F` for an example of generalized GAPW routine.

The impact on the rest of the code was kept low. Only an additional (automatically generated) `AUX_FIT_SOFT` basis type and a `exc1_aux_fit` energy contribution were added.

The new feature was thoroughly tested, including for forces (in geometry optimization and MD). Relevant tests were added in `tests/QS/regtest-admm-gapw`

**Notes:** 
- Real time propagation and `admm_dm` methods are not supported
- ADMM as a whole could use some refactoring. I think that all relevant variables (such as `matrix_ks_aux_fit` or `rho_aux_fit`) should be part of the `admm_type` and not the more general `qs_ks_env_type`. Similarly, the calculation of  the aux_fit neighbor lists, task_list, etc. should take place in ADMM specific routines and not the general QS ones. This is material for another PR though.


dev-zero:
I'd suggest to open issues on GitHub (tag them as `enhancement`) for the todos.

PR #717, opened 2020-01-08
Move dumpdcd to src

mkrack:


Makefile
```diff
@@ -111,9 +111,12 @@ ALL_NONEXE_OBJECTS = $(filter-out $(ALL_EXE_OBJECTS), $(ALL_OBJECTS))
 ifeq ($(ONEVERSION),)
 ORIG_TARGET = default_target
 
-fes :
+fes:
 	@+$(MAKE) --no-print-directory -f $(MAKEFILE) $(VERSION) ORIG_TARGET=graph
 
+dcd:
```
oschuett: You actually don't need to change the Makefile when you add a new binary. It get's picked up automatically. There is this special treatment for `fes` because Teodoro insisted on having a Make-target that was different from the binary name.

mkrack: Cool, that is even better.

mkrack: I kept dumpdcd on purpose as a stand-alone program, though it could use some stuff from basic cp2k modules. What do you think about collecting such programs in a separate subfolder like src/standalone-tools?

oschuett: I have actually been working towards partitioning the code base by methods. Since this usually corresponds with team boundaries we should get well defined APIs were we need them.

So, following this approach I'd place `dumpdcd` into the [motion](https://github.com/cp2k/cp2k/tree/master/src/motion) package.

mkrack: Thanks @oschuett! I guessed that from the structure of src/ already and thought that dumpdcd would fit best into motion too.

PR #720, opened 2020-01-09
Report realloc-lhs only as warning

mkrack:


dev-zero:
I guess this will also need blacklisting in the conventions checker?

oschuett:
The [compiler message](https://dashboard.cp2k.org/archive/mkrack-ssmp/commit_cf2ac326d85ff827c428a3723e964c44da7e5dcd.txt) says:

```
/data/user/krack/rt/gnu-ssmp/cp2k/src/core_ppl.F:641:46:

 !$OMP REDUCTION (+ : pv_thread, force_thread )
                                              1
```
Are we sure that this reallocation is really intended?

mkrack:
Thanks @dev-zero, I forgot about the conventions checker. In that case it is presumably better to drop the -Wrealloc-lhs flag completely, since such warnings might pop up unexpectedly for any new commit.

oschuett:
I think, we should just turn `force_thread` into a stack variable:
```
 REAL(KIND=dp), DIMENSION(3, SIZE(particle_set))    :: force_thread
```

oschuett:
I think, we should keep `-Wrealloc-lhs` and continue to avoid automatic reallocations. This is one of those broken features in modern Fortran and that creates more problems than it solves.

mkrack:
@oschuett That works.

mkrack:
While natom is easily available via particle_set in the core_* routines, that is not the case in qs_overlap. It seems that we will have to pass natom explicitly which is a bit ugly.

oschuett:
I guess, in that case going back to allocating a `POINTER` is the best solution.

mkrack:
I tried that first, of course Unfortunately, pointer objects are not allowed in a reduction clause.

oschuett:
@mkrack, sorry for the clever clogs. I didn't know that pointers don't work with `OMP REDUCTION`. I prepared #725 where I indeed created subroutines with the additional `natom` argument to make the arrays explicitly shaped.

I also opened #726 to discuss how we can better detect lhs-reallocs in the future.

mkrack:
@oschuett I see you performed and committed already the interface changes which I have also in my local copy (could not commit them yesterday morning, since Merlin has maintenance until Monday). That duplication of work is a bit unfortunate.

oschuett:
@mkrack I'm sorry. I didn't realize that you already had this prepared. I have not yet merged #725. So, feel free to close it. 

My main motivation was actually to check for roadblocks wrt. enabling OpenMP in `sdbg` and `pdbg`.

mkrack:
@oschuett Please merge your changes.

Issue #187, opened 2019-01-30
Periodic RTP


juerghutter:
Dear Developers

I would like to draw your attention to a possible bug of CP2K. I am trying to make periodic RTP by using hybrid functionals also.
However, when I set the keyword PERIODIC = TRUE in the REAL_TIME_PROPAGATION  block, the perturbed state that is to be propagated is wrong.
Here below I report the electronic energy for an H2 molecule after a Kick (input at the end of this mail). 
As you can see, the propagated state has energies which differ a lot from the ground state. Furthermore, also the calculated electronic spectrum is very strange.
On the other side, when using PERIODIC = FALSE the energy of the propagated state slightly differs from the ground state and the calculated electronic spectrum is the right one.
I have to underline that the bug appears just when one uses hybrid functionals. With GGA everything goes fine. Is this a bug or something already known?.

PERIODIC TRUE :
  Total energy:                                                -1.16466508784488
  Total energy:                                                -0.76496701712800
  Total energy:                                                -0.76311027473197
  Total energy:                                                -0.76310900639003
  Total energy:                                                -0.76310899786364
  Total energy:                                                -0.76310899785477
  Total energy:                                                -0.76325873454457
  Total energy:                                                -0.76316886201251
  Total energy:                                                -0.76316802986251
  Total energy:                                                -0.76316802392354
  Total energy:                                                -0.76316802386850

PERIODIC FALSE:
  GS Total energy:                                          -1.16466508784488
  Total energy:                                                -1.16466508783571
  Total energy:                                                -1.16466508783580
  Total energy:                                                -1.16466508783580
  Total energy:                                                -1.16466508783579
  Total energy:                                                -1.16466508783579
---------------------------------------------------------------------------------

INPUT

&FORCE_EVAL
  METHOD QUICKSTEP
  &DFT
    &REAL_TIME_PROPAGATION
       APPLY_DELTA_PULSE
       DELTA_PULSE_DIRECTION 1 0 0
       DELTA_PULSE_SCALE 1.0E-5
       PERIODIC .TRUE.
       MAX_ITER 10
       MAT_EXP TAYLOR
       PROPAGATOR ETRS
       EXP_ACCURACY 1.0E-10
       EPS_ITER 1.0E-8
       INITIAL_WFN SCF_WFN
    &END
    BASIS_SET_FILE_NAME …./data/BASIS_SET
    POTENTIAL_FILE_NAME …./data/POTENTIAL
    &MGRID
      CUTOFF 240
[...]

PR #734, opened 2020-01-13
Fix Periodic RTP (issue 187)

mattiatj:
Hotfixes issue #187 187
The problem:
If a delta pulse for RTP is applied periodically the inital MOs for the perturbation are calculated via first order perturbation theory on the ground state MOs.
For calculations employing either GAPW or hybrid functionals the qs_env did not contain the correct KS matrix used as starting point for the PT.
This seems to be fixed by the code provided.

However, for a more elegant solution the interface between RTP and a usual ground state SCF run if INITIAL_WFN is set to SCF_WFN should consider all these cases.

### DIFF
--- a/src/qs_energy.F
+++ b/src/qs_energy.F
@@ -102,7 +102,7 @@ CONTAINS
                ENDIF
                ! add MP2 energy if necessary
                IF (ASSOCIATED(qs_env%mp2_env)) THEN
-                  energy%total = energy%total+energy%mp2
+                  energy%total = energy%total + energy%mp2
                END IF
             END IF
          END IF
--- a/src/qs_energy.F
+++ b/src/qs_energy.F
@@ -95,10 +95,14 @@ CONTAINS
 
          IF (PRESENT(consistent_energies)) THEN
             IF (consistent_energies) THEN
-               CALL qs_ks_update_qs_env(qs_env, calculate_forces=.FALSE., just_energy=.TRUE.)
+               IF (dft_control%do_rtp_periodic) THEN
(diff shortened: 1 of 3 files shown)


src/qs_energy.F
```diff
@@ -95,7 +95,11 @@ SUBROUTINE qs_energies(qs_env, consistent_energies, calc_forces)
 
          IF (PRESENT(consistent_energies)) THEN
             IF (consistent_energies) THEN
-               CALL qs_ks_update_qs_env(qs_env, calculate_forces=.FALSE., just_energy=.TRUE.)
+               IF (dft_control%do_rtp_periodic) THEN
```
schulkov: `dft_control%do_rtp_periodic` is just a synonym of the variable `dft_control%rtp_control%periodic`. It may be a better option to use a local logical variable within the subroutine `qs_energies()` instead of introducing yet another component of the `dft_control` derived type :

```
LOGICAL :: do_rtp_periodic
...
do_rtp_periodic = .FALSE.
IF (ASSOCIATED(dft_control%rtp_control)) &
   do_rtp_periodic = dft_control%rtp_control%periodic

CALL qs_ks_update_qs_env(qs_env, calculate_forces=.FALSE., just_energy=.NOT. do_rtp_periodic)
```

mattiatj: I faintly remember trying that first, but the rtp_control was not initiated for every run type, I think

schulkov: In case there is no &REAL_TIME_PROPAGATION section in the input file, `rtp_control` is just a non-associated (NULL) pointer.

Issue #391, opened 2019-05-28
Allow global keywords to be overridden by environment variables


hfp:
I looked at `keyword_create` (perhaps there is better place?) and wanted to allow the environment to supply the value of a keyword (environment variable). To confirm, the value of a keyword should be overridden (rather than the default value). The parser infrastructure of CP2K may already allow an implementation, however it is not completely clear how to implement this when looking at the various candidate functions.

This enhancement can be used e.g., for benchmarks/competitions where modifying the workload/input is not allowed. More general, such a feature allows to quickly tryout a modification and to experiment with different settings. This feature might also enable automated interaction e.g., auto-tuning the performance of a workload.

The following code may illustrate the idea:

```fortran
CHARACTER(LEN=default_string_length) :: env
INTEGER :: env_stat

CALL GET_ENVIRONMENT_VARIABLE("CP2K_"//name, env, STATUS=env_stat)
IF (0 .EQ. env_stat) THEN
  ! parse value of "env" according to keyword-type
END IF
```

The prefix `CP2K_` for the keyword's name can sufficiently ensure an intentional interaction.

oschuett:
Currently, the input file defines the entire calculation. I think, this is a **very** valuable property that helps a lot with reproducibility.

I also don't quite see the use case, because one can always generate the input from a template:
```
cat benchmark.tmpl | sed s/FOO_PLACEHOLDER/${FOO_VALUE}/g > benchmark.inp
```

hfp:
One case is for benchmarks/competitions where modifying the input file is not allowed, another case is to tune behaviour of CP2K. Though, there are many properties in the GLOBAL section that control/tune general operation. This feature is only about [GLOBAL](https://manual.cp2k.org/cp2k-6_1-branch/CP2K_INPUT/GLOBAL.html) and not about mirroring the ability to describe a workload using environment variables (which would be indeed obscure). I believe it is easily possible to detect during `keyword_create` whether a keyword can have a corresponding environment variable or not.

dev-zero:
For parameters which depend on the runtime environment I think it would be very valuable to be able to override them either by explicit CLI arguments and/or environment variables, such that one can use the same input files with different parameters (separation of concerns: the input defines everything algorithmic, the physical system, etc. while the CLI and/or environments can be adjusted to fit the current runtime environment).

hfp:
I think more issues to make CP2K easier to use (or to make it more similar to what other package allow) can be reported.

For example:

[...]

PR #735, opened 2020-01-13
Add default values and CLI arguments for setting preprocessor vars

dev-zero:
possible solution for #391

### DIFF
--- a/tools/manual/cp2k_input.xsl
+++ b/tools/manual/cp2k_input.xsl
@@ -64,6 +64,7 @@
      Assigns the text <i>value</i> to the preprocessing variable <i>VAR</i>. <i>value</i>
      is the text following <i>VAR</i> with the outer whitespace removed. The variable can be
      recalled with a <i>${VAR}</i> (or  <i>$VAR</i>) statement. There can be only one @SET statement per line.
+     A variable name can only contain alphanumerical characters plus the underscore and must not start with a number.
     </dd>
     <dt><b>${VAR}</b> or <b>$VAR</b></dt>
     <dd>
--- a/tools/manual/cp2k_input.xsl
+++ b/tools/manual/cp2k_input.xsl
@@ -72,6 +72,12 @@
      There can be multiple variable statements per line. The expansion process is repeated
      until no more variables are found.
     </dd>
+    <dt><b>${VAR-DEFAULT}</b></dt>
+    <dd>
+     Expand the variable <i>VAR</i> if defined. The text <i>${VAR-DEFAULT}</i> is replaced
+     with the value assigned to <i>VAR</i> in the last @SET directive if defined,
+     otherwise with the value specified after the dash <i>-</i> (here: <i>DEFAULT</i>).
+    </dd>
     <dt><b>@IF / @ENDIF</b></dt>
     <dd>
      Conditional block. The text from the @IF line up to the next line with a valid
--- a/src/f77_interface.F
+++ b/src/f77_interface.F
@@ -538,7 +538,7 @@ CONTAINS
 !>        instead of parsing from file
 !> \param ierr will return a number different from 0 if there was an error
 !> \param work_dir ...
-!> \param initial_variables ...
+!> \param initial_variables key-value list of initial preprocessor variables
 !> \author fawzi
 !> \note
 !>      The following routines need to be synchronized wrt. adding/removing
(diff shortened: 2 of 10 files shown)


tools/manual/cp2k_input.xsl
```diff
@@ -72,6 +72,12 @@
      There can be multiple variable statements per line. The expansion process is repeated
      until no more variables are found.
     </dd>
+    <dt><b>${VAR-DEFAULT}</b></dt>
```
oschuett, line 76: This syntax can lead to ambiguities because variable names can contain pretty much any character. The only save choice for a separator is probably whitespace, ie. `${VAR default}`. Alternatively, we could use the bash syntax which uses two chars `:-`. Furthermore, this feature should probably be disabled for variables without braces, ie. `$VAR`.

oschuett:
The problem with this change is that the input file would no longer be self-contained. Instead one would also have to know the full command line to figure out what calculation was actually performed. This will hurt reproducibility and thereby effectively science.

On the other hand, I still fail to see the benefit for the user. Whoever wants to build parametrizabile inputs will be better served with a proper template- or workflow-engine.

The use case from #391 is about the peculiar situation _''where modifying the workload/input is not allowed''_. It certainly does not justify taking a hit on reproducibility.

tools/manual/cp2k_input.xsl
dev-zero: > This syntax can lead to ambiguities because variable names can contain pretty much any character. The only save choice for a separator is probably whitespace, ie. `${VAR default}`. 

Mildly. If someone wants to use default vars he can't use dashes anymore.
But yes, it might be a good idea to restrict the character set for variable names.

> Alternatively, we could use the bash syntax which uses two chars `:-`.

It is [bash syntax](https://www.tldp.org/LDP/abs/html/parameter-substitution.html), `:-` has a slightly different behavior I didn't want to implement (yet): using the default value if the content of the var is empty.
Besides, it doesn't actually solve the ambiguity, just makes it a bit less likely.

> Furthermore, this feature should probably be disabled for variables without braces, ie. `$VAR`.

It already is only available for vars with braces.

dev-zero:
> The problem with this change is that the input file would no longer be self-contained. Instead one would also have to know the full command line to figure out what calculation was actually performed. This will hurt reproducibility and thereby effectively science.

Given that the behavior (and numbers if not converged) changes depending on number of MPI ranks, OpenMP threads and possibly other environment-affected variables, this has always been the case.

> On the other hand, I still fail to see the benefit for the user. Whoever wants to build parametrizabile inputs will be better served with a proper template- or workflow-engine.

Yes, they would, unfortunately they don't. Instead they build convoluted scripts using `sed`, `awk` and ugly hierarchies of folders with symlinks to snippets being included through `@INCLUDE`, making the run depend on the folder where they are running the calculation.

> The use case from #391 is about the peculiar situation _''where modifying the workload/input is not allowed''_. It certainly does not justify taking a hit on reproducibility.

For reproducibility one always would have had to archive the output of `cp2k --echo`. The initial proposal in #391 to introduce environment variables at the parser level would have made this impossible. Here we stop the magic after a `cp2k --echo` and from there on it is only used libraries and runtimes which will be directly affected by the environment.

tools/manual/cp2k_input.xsl
oschuett: The cleanest solution would probably be `@SETDEFAULT`, still we shouldn't do it.

oschuett:
> this has always been the case.

Yes, we are not doing great on reproducibility. So, let's not make it worse.

> Instead they build convoluted scripts using sed, awk and ugly hierarchies of folders with symlinks...

You have a point. However, I don't think extending our preprocessor is the right answer. Unless you want to spend the next 3 years implementing a full-fledged template engine, people will continue to use *sed* and *awk*. Hence, making our preprocessor more magical will just add to the mess.

Instead, we should write a tutorial about how to use a proper template engine like [Cheetah](https://pythonhosted.org/Cheetah/) or [Jinja](https://jinja.palletsprojects.com/).

tools/manual/cp2k_input.xsl
dev-zero: btw, this extension solves the issue of providing XC includes with default values

oschuett: As discussed in https://github.com/cp2k/cp2k/pull/559 we shouldn't really be using the preprocessor for canned XC sections. Instead we should build on what we have e.g. for [PBE0](https://github.com/cp2k/cp2k/blob/master/src/input_cp2k_check.F#L154). That solution is clean, robust, extendable, and user-friendly.

Btw. I'm marking this as unresolved, because my original concerns about ambiguities and backwards incompatibility have not yet been address.

dev-zero: OK, restriction added for the variable names. With the example given in the documentation we have already implied that the variable name is a word, so I don't see a problem of defining the supported character set now (also, `$` and ` ` were already not part of the set).

dev-zero:
After some more discussion here at UZH I'm going to merge this. To summarize: preprocessor features have always been something only advanced users were using since we usually don't advertise it. Both commits solve a very specific need power users have in a way which is still very obvious (compared to e.g. pulling configuration vars from environment variables directly).

PR #744, opened 2020-01-21
Update to DBCSR 2.1.0

pseewald:
Revert "Rollback to DBCSR 2.0.1"

This reverts commit c88089a1d54fc2c1ee02284298bf9759df3280a6.

### DIFF
--- a/src/f77_interface.F
+++ b/src/f77_interface.F
@@ -43,9 +43,7 @@ MODULE f77_interface
                                               cp_subsys_set,&
                                               cp_subsys_type,&
                                               unpack_subsys_particles
-   USE dbcsr_api,                       ONLY: dbcsr_acc_get_ndevices,&
-                                              dbcsr_acc_set_active_device,&
-                                              dbcsr_finalize_lib,&
+   USE dbcsr_api,                       ONLY: dbcsr_finalize_lib,&
                                               dbcsr_init_lib
    USE eip_environment,                 ONLY: eip_init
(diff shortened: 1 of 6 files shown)


pseewald:
Judging from the discussion #677 we need to either add `-D__CUDA` in the CP2K arch file, or adapt DBCSR such that this macro is not needed. Probably we should go for the DBCSR-internal solution (as suggested already by @alazzaro).

shoshijak:
I agree that we should go with the DBCSR-internal solution. We could just update [these lines of DBCSR's Makefile](https://github.com/cp2k/dbcsr/blob/develop/Makefile#L57-L67) to add the `-D__CUDA` flag if we're building with nvcc. 

Like this: 

```diff
─────────────────────────────────────────────
modified: Makefile
─────────────────────────────────────────────
@ Makefile:59 @ endif
 ifneq ($(ARCH_NUMBER),)
 # If compiling with nvcc
 ifneq (,$(findstring nvcc,$(NVCC)))
+NVFLAGS += -D__CUDA # add the flag for compilation with CUDA
 #if "-arch" has not yet been set in NVFLAGS
 ifeq ($(findstring "-arch", $(NVFLAGS)), '')
 NVFLAGS += -arch sm_$(ARCH_NUMBER)
```

I'll open a PR in DBCSR once I've tested this ...

shoshijak:
Tested on Daint. PR https://github.com/cp2k/dbcsr/pull/288 should solve the issue.

PR #746, opened 2020-01-23
Trust-region methods for XALMO optimization

rzk1:


src/almo_scf_optimizer.F
```diff
@@ -5678,24 +5486,34 @@ SUBROUTINE compute_preconditioner(domain_prec_out, m_prec_out, m_ks, m_s, &
...
       IF (special_case .EQ. xalmo_case_block_diag) THEN ! non-overlapping diagonal blocks
 
-         CALL pseudo_invert_diagonal_blk( &
-            matrix_in=m_tmp_nn_1, &
-            matrix_out=m_prec_out, &
-            nocc=nocc_of_domain(:) &
-            )
+         IF (skip_inversion) THEN
+            CPABORT("NYI: impose blk structure on m_prec_out first")
+            CALL dbcsr_copy(m_prec_out, m_tmp_nn_1, keep_sparsity=.TRUE.)
```
dev-zero, line 5491: There is little point in calling another function after a `CPABORT`.
If this is needed once the function gets implemented then its fine, but you might want to comment it out like other cases, otherwise it looks like a misplaced statement.

src/almo_scf_optimizer.F
```diff
@@ -7605,5 +7427,2016 @@ SUBROUTINE hessian_diag_apply(matrix_grad, matrix_step, matrix_S_ao, &
...
+      ENDDO !ispin
+
+      DEALLOCATE (nocc)
+      DEALLOCATE (m_temp_no)
+
+      coef(:) = coef(:)*spin_factor
+      coef(3) = coef(3) - trust_radius_in*trust_radius_in
+
+      ! solve the quadratic equation
+      discrim
```
dev-zero, line 9216: `trim` not necessary in many cases, same further below and other functions

dev-zero, line 8937: might make sense to turn the `contravariant_matrix_norm` subroutine into a function, but I guess that's a matter of style

dev-zero (approved):
LGTM, only minor remarks

src/almo_scf_optimizer.F
rzk1: This subroutine will be modified in the future to return a matrix. That is why it is not implemented as a function.

src/almo_scf_optimizer.F
rzk1: This block is being implemented.

dev-zero:
@rzk1 thanks

PR #747, opened 2020-01-23
toolchain: fix SIRIUS build with SpFFT on distros with lib64 as libdir

dev-zero:


dev-zero:
@mtaillefumier

dev-zero:
the error in pdbg is unrelated to this change, proceeding

Issue #699, opened 2019-12-19
toolchain: version bumps


dev-zero:
list of packages with a newer version available than currently being installed by the toolchain:

* [x] CMake: 3.15.1 -> 3.15.6/3.16.1
* [x] ELPA: 2019.05.001 -> 2019.05.002 (2019.11.001.rc1 fixes the parallel build issues)
* [x] GCC: 8.3 -> 9.2 (or 8.4 to be released soon)
* [x] GSL: 2.5 -> 2.6
* [x] HDF5: 1.10.4 -> 1.10.5
* [x] MPICH: 3.3 -> 3.3.2
* [x] OpenBLAS: 0.3.6 -> 0.3.7
* [x] OpenMPI: 4.0.1 -> 4.0.2
* [x] PEXSI: 0.10.2 -> 1.2.0
* [x] PLUMED: 2.5.2 -> 2.5.3
* [x] QUIP: cc83ceea5776c40fcb5ab224a25ab04d62175449 -> 641caeefdbeae5f98d2c498bc4ae01a8967453eb ?
* [x] LAPACK: 3.8.0 -> [3.9.0](http://performance.netlib.org/lapack/lapack-3.9.0.html) ([closed bugs](https://github.com/Reference-LAPACK/lapack/milestone/4?closed=1))
* [x] ScaLAPACK: 2.0.2 -> 2.1.0
* [x] SIRIUS: 6.4.1 -> 6.4.4
* [x] SpFFT: 0.9.8 -> 0.9.10
* [x] spglib: 1.12.1 -> 1.12.2/1.14.1
* [x] SuperLU: 5.1.2 -> [6.2.0](https://github.com/xiaoyeli/superlu_dist/releases) (depends on what PEXSI supports)
* [x] valgrind: 3.14.0 -> 3.15.0 (check whether MPI-3+ support is now available)

for most packages minor releases should be easy.

dev-zero:
resolved with #962 and follow-ups

PR #756, opened 2020-01-31
Upgrade packages for toolchain

mkrack:


dev-zero:
cool, that takes 2 off from #699. What's the reasoning behind dropping `-ftree-vectorize` (not that I would know why it's there in the first place)?

mkrack:
I introduced the -ftree-vectorize flag recently and it troubles for OpenBLAS. So I removed it again.

dev-zero:
> I introduced the -ftree-vectorize flag recently and it troubles for OpenBLAS. So I removed it again.

Good to know, thanks. Will be cherry-picking d8af157 to the 7.1 release branch then.

Issue #750, opened 2020-01-27
Formatting check not working


oschuett:
In the current master branch 42 files are not formatted correctly.

For earlier discussions see [here](https://github.com/cp2k/cp2k/pull/636#issuecomment-555996021) and [here](https://github.com/cp2k/cp2k/pull/632#issuecomment-555705513).

Reproducer:
```
$ git clone git@github.com:cp2k/cp2k.git
$ cd cp2k
$ git submodule update --init --recursive
$ make -j10 pretty
...
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   src/arnoldi/arnoldi_data_manipulation.f90
	modified:   src/arnoldi/arnoldi_data_methods.F
	modified:   src/arnoldi/arnoldi_data_selection.f90
	modified:   src/arnoldi/arnoldi_methods.F
	modified:   src/common/callgraph.F
	modified:   src/common/cp_array_sort.F
	modified:   src/common/cp_array_utils.F
	modified:   src/common/list.F
	modified:   src/common/list_callstackentry.F
	modified:   src/common/list_routinereport.F
	modified:   src/common/list_routinestat.F
	modified:   src/common/list_timerenv.F
	modified:   src/common/memory_utilities.F
	modified:   src/common/routine_map.F
	modified:   src/cp_dbcsr_operations.F
	modified:   src/dbcsrx/dbcsr_vector.F
	modified:   src/eri_mme/eri_mme_lattice_summation.F
	modified:   src/ewalds_multipole.F
	modified:   src/fm/cp_fm_elpa.F
	modified:   src/fm/cp_linked_list_fm.F
	modified:   src/gauss_colloc.F
	modified:   src/grid/xyz_to_vab_optimised.F
	modified:   src/input/cp_linked_list_input.F
	modified:   src/input/cp_parser_inpp_types.F
	modified:   src/input_optimize_input.F
	modified:   src/libint_wrapper.F
	modified:   src/mp2_eri.F
	modified:   src/mpiwrap/message_passing.F
	modified:   src/mpiwrap/message_passing.f90
	modified:   src/negf_integr_utils.F
	modified:   src/pw/cp_linked_list_pw.F
	modified:   src/qmmm_gpw_forces.F
	modified:   src/se_fock_matrix_integrals.F
[...]

PR #74, opened 2018-11-02
Improved code comment. Fixed some typos.

hfp:

PR #763, opened 2020-02-05
Fix typo

e-kwsm:


tools/build_utils/fypp
```diff
@@ -37,7 +37,7 @@ Python, one usually interacts with the following two classes:
 * `Fypp`_: The actual Fypp preprocessor. It returns for a given input
   the preprocessed output.
 
-* `FyppOptions`_: Contains customizable settings controling the behaviour of
+* `FyppOptions`_: Contains customizable settings controlling the behaviour of
```
dev-zero: please send fixes for Fypp directly to https://github.com/aradi/fypp

dev-zero (changes requested):
Changes to Fypp should be sent directly to https://github.com/aradi/fypp and we will pull them from there.

dev-zero:
Some of the `prettify` formatting checks probably fail due to changes made to our prettifier without re-running on the complete tree, see #750 
The changes requested by the `black` formatter on the other hand need to be incorporated.

@mtaillefumier can you please make sure fixes to the `m4` files end up at upstream (and document where those files were pulled from originally)?

Otherwise LGTM, thanks!

mtaillefumier:
do not reformat the m4 files please.

tools/build_utils/fypp
e-kwsm: Submitted: aradi/fypp#6

e-kwsm:
> do not reformat the m4 files please.

The changes were reverted.

oschuett:
Seems like you added merge commit. With all the reverts the git history is pretty scrambled anyways. Maybe start over with a single clean commit? You can keep this PR open and simply use `git push --force`.

e-kwsm:
CI failure (`./tools/prettify/prettify.py src/pw/ps_wavelet_base.F` on a local machine) is due to:
```diff
diff --git a/src/pw/ps_wavelet_base.F b/src/pw/ps_wavelet_base.F
index 3e773e1730cb..ac7b2f136479 100644
--- a/src/pw/ps_wavelet_base.F
+++ b/src/pw/ps_wavelet_base.F
@@ -699,7 +699,6 @@ CONTAINS
 !> \author S. Goedecker, L. Genovese
 ! **************************************************************************************************
    SUBROUTINE P_multkernel(n1, n2, n3, lot, nfft, jS, i3, zw, hx, hy, hz)
-      !Arguments
       INTEGER, INTENT(in)                                :: n1, n2, n3, lot, nfft, jS, i3
       REAL(KIND=dp), DIMENSION(2, lot, n2), &
          INTENT(inout)                                   :: zw
@@ -758,7 +757,6 @@ CONTAINS
 !> \author S. Goedecker, L. Genovese
 ! **************************************************************************************************
    SUBROUTINE multkernel(nd1, nd2, n1, n2, lot, nfft, jS, pot, zw)
-      !Arguments
       INTEGER, INTENT(in)                                :: nd1, nd2, n1, n2, lot, nfft, jS
       REAL(KIND=dp), DIMENSION(nd1, nd2), INTENT(in)     :: pot
       REAL(KIND=dp), DIMENSION(2, lot, n2), &
```

oschuett:
Interesting!

Prettify leaves the following unchanged:
```
SUBROUTINE P_multkernel(n1, n2, n3, lot, nfft, jS, i3, zw, hx, hy, hz)
       !Argments
       INTEGER, INTENT(in)                                :: n1, n2, n3, lot, nfft, jS, i3
```

but when changed to:

```
SUBROUTINE P_multkernel(n1, n2, n3, lot, nfft, jS, i3, zw, hx, hy, hz)
       !Arguments
       INTEGER, INTENT(in)                                :: n1, n2, n3, lot, nfft, jS, i3
```

prettify suddenly wants to remove the comment line.

This smells like a bug in fpretty. However, I'm not sure if it's worth investigating. So, mentioning @pseewald just in case he is curious.

As for this PR, you can simply delete that comment line and move on.

alazzaro:
I would assume that there is a conflict with FORD comment here...

pseewald:
Interesting indeed. I think this is done on purpose (not by me): https://github.com/cp2k/cp2k/blob/master/tools/prettify/prettify_cp2k/normalizeFortranFile.py#L123-L126

Shall we remove this?

oschuett:
@pseewald, thanks for the quick response. Yes, I think we can remove that. I've opened #826.

oschuett:
Thanks for fixing these spelling mistakes.

I was wondering if we could maybe turn this into a permanent solution. We could compile a list of commonly misspelled words from this and similar PRs (#74, #302, #609). Then we could let [analyze_src.py](https://github.com/cp2k/cp2k/blob/master/tools/conventions/analyze_src.py) check the code against that list automatically.

Issue #737, opened 2020-01-14
Lots of tools not ported to python3 yet


mbanck:
python2 is EOL and it seems to me that lots of tools (also in dbcsr) are still python2:

```
$ egrep -r '(bin|env)'.python  | grep -v python3
exts/dbcsr/tests/generate_libcusmm_unittest_multiply.py:#!/usr/bin/env python
exts/dbcsr/tests/generate_libcusmm_timer_multiply.py:#!/usr/bin/env python
exts/dbcsr/.pre-commit/check_header.py:#!/usr/bin/env python
exts/dbcsr/tools/fprettify/run_tests.py:#!/usr/bin/env python
exts/dbcsr/tools/fprettify/fprettify/tests/__init__.py:#!/usr/bin/env python
exts/dbcsr/tools/fprettify/fprettify/__init__.py:#!/usr/bin/env python
exts/dbcsr/tools/fprettify/fprettify.py:#!/usr/bin/env python
exts/dbcsr/tools/fprettify/setup.py:#!/usr/bin/env python
exts/dbcsr/tools/build_utils/test_makedep.py:#!/usr/bin/env python
exts/dbcsr/tools/build_utils/check_archives.py:#!/usr/bin/env python
exts/dbcsr/tools/build_utils/fypp/tools/waf/fypp_preprocessor.py:#!/usr/bin/env python
exts/dbcsr/tools/build_utils/fypp/tools/waf/fypp_fortran.py:#!/usr/bin/env python
exts/dbcsr/tools/build_utils/fypp/bin/fypp:#!/usr/bin/env python
exts/dbcsr/tools/build_utils/makedep.py:#!/usr/bin/env python
exts/dbcsr/Makefile:	/usr/bin/env python --version
exts/dbcsr/src/acc/libsmm_acc/libcusmm/generate_parameters.py:#!/usr/bin/env python
exts/dbcsr/src/acc/libsmm_acc/libcusmm/generate_kernels.py:#!/usr/bin/env python
tools/package_planner/plan_packages.py:#!/usr/bin/env python
tools/prettify/prettify.py:#!/usr/bin/env python
tools/prettify/fprettify/run_tests.py:#!/usr/bin/env python
tools/prettify/fprettify/fprettify/tests/__init__.py:#!/usr/bin/env python
tools/prettify/fprettify/fprettify/__init__.py:#!/usr/bin/env python
tools/prettify/fprettify/fprettify.py:#!/usr/bin/env python
[...]

PR #766, opened 2020-02-06
New-Style array handling

oerc0122:
Just a quick fix to allow handling of new-style array syntax
```
integer, dimension(3), parameter :: a = [1, 2, 3]
```


dev-zero:
I guess this is fine, thanks a lot!
Would you mind converting the regex to a verbose one (e.g. using `re.VERBOSE`)?

oerc0122:
How legible would you like it? I can make it a regex built from sub-elements rather than a larger less comprehensible mass, or I can maintain its current style and just add newlines where relevant.

Using sub-elements:
```
NOT_PUNC = r"[^()\"']"
DQUOTED = r'"[^"]*"'                                                               
SQUOTED = r"'[^']*'"
QUOTED = rf"(?:{DQUOTED}|{SQUOTED})"
VAR_RE = re.compile(rf"""
\s*(?P<var>[a-zA-Z_0-9]+)
\s*(?P<rest>(?:\((?P<param>(?:[^()]+|\((?:[^()]+|\([^()]*\))*\))*)\))?
\s*(?:=\s*(?P<value>(:?{NOT_PUNC}+|\[(?:{NOT_PUNC}+|\({NOT_PUNC}*\)|{QUOTED})*\]|\((?:{NOT_PUNC}+|\({NOT_PUNC}*\)|{QUOTED})*\)|{QUOTED})+))?)?
\s*(?:(?P<continue>,)|\n?)
\s*""", re.IGNORECASE|re.VERBOSE)
```

dev-zero:
@oerc0122 that looks much better, thanks! When switching to f-strings we have to update the shebang in `tools/prettify/prettify.py` to match `python3`, otherwise it will fail for some people. See also #737

oerc0122:
If you'd prefer to keep backwards compatibility (pre 3.5/6), it's perfectly viable to use `.format` rather than f-strings.

dev-zero:
No, that's fine. Python 2.7 is EOL, Ubuntu/Debian have LTS/stable releases with Python 3.6 and this is a developer tool.

oerc0122:
Right, I've updated a chunk of them and they should now work as expected. Though the switch from Python2 -> 3 may cause some hiccups in the testing, might need to be double checked, but the framework's there.

dev-zero:
@oerc0122 cool, thanks a lot! Formatting test currently fails due to the black formatter it seems, but that's something I can push on this PR.

dev-zero:
hmm, something's wrong now:

```
 ./tools/prettify/prettify.py src/start/cp2k.F 
  File "/data/tiziano/cp2k/tools/prettify/prettify_cp2k/normalizeFortranFile.py", line 1293, in rewriteFortranFile
    inc_fn = COMMON_USES_RE.match(modulesDict["commonUses"]).group(1)
IndexError: no such group

During handling of the above exception, another exception occurred:

  File "./tools/prettify/prettify.py", line 484, in main
    replace=args.replace,
  File "./tools/prettify/prettify.py", line 284, in prettifyInplace
    outfile = prettifyFile(infile=infile, filename=filename, **kwargs)
  File "./tools/prettify/prettify.py", line 239, in prettifyFile
    orig_filename=filename,
  File "/data/tiziano/cp2k/tools/prettify/prettify_cp2k/normalizeFortranFile.py", line 1301, in rewriteFortranFile
    inc_absfn
UnboundLocalError: local variable 'inc_absfn' referenced before assignment

During handling of the above exception, another exception occurred:

  File "./tools/prettify/prettify.py", line 494, in <module>
    sys.exit(main(sys.argv[1:]))
  File "./tools/prettify/prettify.py", line 487, in main
    logger.exception("processing file failed", extra={"ffilename": filename})
NameError: name 'logger' is not defined
```

... well, the latter two are unrelated issues fixed in #772

oerc0122:
I am an idiot and obviously the `re.VERBOSE` makes the `#` in `#include` a comment marker, thus the entire line is commented. Sorry about that.

PR #767, opened 2020-02-07
Improved time/frequency treatment for low-scaling GW

JWilhelm:
The LUMOs computed with the low-scaling GW algorithm of all 100 molecules from the GW100 set deviate now on average 0.007 eV from the reference data. HOMO is slightly worse due to multiple GW solutions. HOMOs in practice are expected to be as well behaved as LUMOs in GW100 because in larger molecules and materials, the self-interaction is less severe than in small molecules as in GW100. High accuracy is also gained using RI with truncated Coulomb metric as implemented by Patrick Seewald. In the long term, high accuracy is especially important for applying and benchmarking the k-point low-scaling GW code for periodic systems.

### DIFF
--- a/src/mp2_grids.F
+++ b/src/mp2_grids.F
@@ -155,7 +155,7 @@ CONTAINS
       END IF
 
       IF (my_do_mixed_minimax_clenshaw) THEN
-         CALL compute_tj_from_clenshaw(tj, a_scaling)
+         tj(:) = a_scaling/TAN(tj(:))
       END IF
 
       IF (.NOT. do_ri_sos_laplace_mp2 .AND. (.NOT. my_do_mixed_minimax_clenshaw)) THEN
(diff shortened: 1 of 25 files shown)


pseewald (approved):

src/mp2_grids.F
```diff
@@ -267,7 +279,30 @@ SUBROUTINE get_minimax_weights(para_env, unit_nr, homo, Eigenval, num_integ_poin
...
+      REAL(KIND=dp), ALLOCATABLE, DIMENSION(:), &
+         INTENT(INOUT)                                   :: tj
+      REAL(KIND=dp), INTENT(IN)                          :: a_scaling
+
+      CHARACTER(LEN=*), PARAMETER :: routineN = 'compute_tj_from_clenshaw', &
+         routineP = moduleN//':'//routineN
+
+      INTEGER                                            :: handle
+
+      CALL timeset(routineN, handle)
```
fstein93: Is it really necessary to time such a short function? What do you think of totally removing this helper function bc it is used nowhere else?

JWilhelm: Thanks for pointing this out, the function is removed.

PR #769, opened 2020-02-10
Some more small preparations for #458

oschuett:


dev-zero:
Please note that in the [recent past](https://github.com/cp2k/cp2k/pull/410#discussion_r298520649) we've avoided the use of C99 (voluntarily).
So, this requirement should be properly documented in the installation documents.

oschuett:
Good point. Will do.

PR #776, opened 2020-02-13
Remove unused input variables, optional printing of self-energy for real frequency

JWilhelm:


src/rpa_gw.F
```diff
@@ -3621,6 +3558,43 @@ SUBROUTINE continuation_pade(vec_gw_energ, vec_omega_fit_gw, &
...
+            WRITE (filename, "(A21,I1)") "self_energy_level_000", n_level_gw_ref
+         ELSE IF (n_level_gw < 100) THEN
+            WRITE (filename, "(A20,I2)") "self_energy_level_00", n_level_gw_ref
+         ELSE IF (n_level_gw < 1000) THEN
+            WRITE (filename, "(A19,I3)") "self_energy_level_0", n_level_gw_ref
+         ELSE
+            WRITE (filename, "(A18,I4)") "self_energy_level_", n_level_gw_ref
+         END IF
+
+         OPEN (UNIT=n_level_gw_ref*1000, FILE=TRIM(filename))
```
fstein93: You can remove the warning from the convention checker by using the routines open_file and close_file from the module cp_files. Further, why do you have different units when you are opening and closing exactly one file in the routine?

JWilhelm: Thanks a lot for your hint! Will be changed very soon. 

The routine prints the self-energy Sigma(w) for every level that is corrected by GW, so several files are printed. Each level is unique to a single MPI rank and this MPI rank prints the specific file. I hope this answers the question...

src/rpa_util.F
```diff
@@ -747,6 +756,38 @@ SUBROUTINE calc_fm_mat_S_rpa(fm_mat_S, first_cycle, count_ev_sc_GW, virtual, Eig
...
+      REAL(KIND=dp), DIMENSION(:), INTENT(IN)            :: Eigenval, Eigenval_scf
+      INTEGER, INTENT(IN)                                :: iocc, avirt, homo, iter_sc_GW0
+
+      CHARACTER(LEN=*), PARAMETER :: routineN = 'get_eigen_diff', routineP = moduleN//':'//routineN
+
+      INTEGER                                            :: handle
+
+      CALL timeset(routineN, handle)
+
+      IF (iter_sc_GW0 == 1) THEN
```
fstein93: What about moving this if-clause outside the function (maybe to calc_mat_Q)? Then you have to check that condition only once, and you don't need the argument iter_sc_GW0 and only two of the Eigenval arrays in calc_fm_mat_S_rpa.

JWilhelm: Thanks, good point, will be changed!

PR #783, opened 2020-02-19
Simplify input

JWilhelm:
see commit messages


tests/QS/regtest-gw/TEST_FILES
```diff
@@ -1,11 +1,11 @@
...
+evGW_OH_PBE.inp                                       11      1e-08            -16.414920515710477
+evGW_H2O_PBE_RI_HFX.inp                               11      1e-08            -13.112474185087502
+G0W0_H2O_PBE_periodic.inp                             11      1e-07            -17.144843116457857
 G0W0_H2O_PBE_pade.inp                                 78    1.0E-14                          16.22
-evGW_OH_PBE_svd.inp                                   11      1e-08            -16.414906390703532
-evGW_H2O_PBE_RI_HFX_svd.inp                           11      1e-08            -13.112371113806535
-scGW0_H2O_PBE.inp                                     11      1e-08            -17.118381127539799
+evGW_OH_PBE_svd.inp                                   11      1e-08            -16.452864482288359
+evGW_H2O_PBE_RI_HFX_svd.inp                           11      1e-08            -13.176028670946073
+scGW0_H2O_PBE.inp                                     11      1e-08            -17.118441200805574
```
pseewald, line 10: Why does this commit change regtest values?

src/mp2_gpw.F
```diff
@@ -330,6 +329,25 @@ SUBROUTINE mp2_gpw_main(qs_env, mp2_env, Emp2, Emp2_Cou, Emp2_EX, Emp2_S, Emp2_T
...
+      END IF
+      eri_method = mp2_env%eri_method
+
+      PRINT *, 'mp2_env%eri_method =', mp2_env%eri_method
+      PRINT *, 'do_eri_gpw =', do_eri_gpw
+      PRINT *, 'do_eri_os =', do_eri_os
+      PRINT *, 'eri_default =', eri_default
+      PRINT *, 'mp2_env%method =', mp2_env%method
+      PRINT *, 'mp2_method_gpw =', mp2_method_gpw
+
```
pseewald: Probably you want to remove the print statements

src/input_cp2k_mp2.F
```diff
@@ -1190,9 +1190,18 @@ SUBROUTINE create_low_scaling(section)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
+      CALL keyword_create( &
+         keyword, __LOCATION__, name="EPS_FILTER", &
+         description="Determines a threshold for the DBCSR based multiply."// &
+         "Normally, this EPS_FILTER determines accuracy and timing of low-scaling RPA and GW calculations.", &
+         usage="EPS_FILTER 1.0E-10 ", type_of_var=real_t, &
+         default_r_val=1.0E-12_dp)
```
pseewald: I don't think there is a need to change this default. I think the previous default (`EPS_FILTER` from `&WFC_GPW`, `1.0E-09_dp`) is reasonably safe.

JWilhelm: Thanks for pointing this out, let's better keep it as you write.

tests/QS/regtest-gw/TEST_FILES
JWilhelm: It's because of the new defaults for the integrals: GPW for MP2 in any case, Obara-Saika integrals for non-periodic systems, GPW for periodic, Gamma-only calculations, Obara-Saika for periodic systems with k-points. For non-GW regtests, I have added ERI_METHOD GPW in the input to keep the regtest values, for GW regtests, I kept the input as is with the default ERI_METHOD while updating the reference values.

PR #784, opened 2020-02-20
Fixed a bug of reading VIB_EIGS_FILE_NAME

zjing7:


oschuett:
Thanks for the patch!

It seems we don't have a test for this feature yet. Would you like to add one?

oschuett:
We want to keep our git history linear and therefore don't allow merge commits. You should be able to recover by doing the following:

```
git reset --hard 857f74a77d39098fa0351ad551a323d9ea647547
git push --force
```

oschuett:
Upps, now your [original commit](https://github.com/cp2k/cp2k/commit/197ee6ee8dd58d4c83b6cbb98c6614f6526d4c86) is gone. The "official solution" would be to use `git rebase`, but since it's very small I'd just create a new commit on top of what you have now.

oschuett:
Thanks for contributing!

zjing7:
My pleasure:)

PR #788, opened 2020-02-26
Add implementation of submatrix method

michaellass:
This pull request adds an implementation of the submatrix method, as described in:

> Michael Lass, Stephan Mohr, Hendrik Wiebeler, Thomas D. Kühne, and Christian Plessl. 2018. A Massively Parallel Algorithm for the Approximate Calculation of Inverse p-th Roots of Large Sparse Matrices. In Proceedings of the Platform for Advanced Scientific Computing Conference (PASC ’18). Association for Computing Machinery, New York, NY, USA, Article 7, 1–11. DOI:https://doi.org/10.1145/3218176.3218231

The submatrix method allows to approximately solve matrix functions for large, sparse matrices by solving the same function on a set of smaller, dense matrices. While the original publication mainly targets inverse p-th roots, the method can similarly applied to other matrix functions.

As part of this pull request, an option is added to calculate the sign function within dm_ls_scf using the submatrix method. Solving the sign function for the submatrices can be either done using a Newton-Schulz iteration or a direct approach based on an explicit eigendecomposition.


src/iterate_matrix.F
```diff
@@ -1072,6 +1078,228 @@ SUBROUTINE matrix_sign_proot(matrix_sign, matrix, threshold, sign_order)
...
+      CALL dbcsr_get_info(matrix=matrix, nblkcols_total=nblkcols, distribution=dist, group=group)
+      CALL dbcsr_distribution_get(dist=dist, mynode=myrank)
+
+      CALL dissection%init(matrix)
+      CALL dissection%get_sm_ids_for_rank(myrank, my_sms)
+
+      !$OMP PARALLEL PRIVATE(sm, sm_sign, sm_size) SHARED(dissection, myrank, my_sms)
+      !$OMP DO SCHEDULE(GUIDED)
+      DO i = 1, SIZE(my_sms)
+         PRINT *, "Rank", myrank, "processing submatrix", my_sms(i)
```
dev-zero: this bare print statement should probably be converted to using a logger facility

src/submatrix_dissection.F
```diff
@@ -0,0 +1,895 @@
...
+   USE dbcsr_api, ONLY: dbcsr_create, dbcsr_distribution_get, dbcsr_distribution_type, dbcsr_finalize, &
+                        dbcsr_get_block_p, dbcsr_get_info, dbcsr_get_stored_coordinates, &
+                        dbcsr_iterator_blocks_left, dbcsr_iterator_next_block, dbcsr_iterator_start, &
+                        dbcsr_iterator_stop, dbcsr_iterator_type, dbcsr_put_block, dbcsr_type
+   USE dbcsr_mpiwrap, ONLY: mp_allgather, mp_alltoall, mp_irecv, mp_isend, mp_send, mp_wait
+   USE kinds, ONLY: dp
+   USE submatrix_methods, ONLY: qsort_two
+   USE submatrix_types, ONLY: buffer_type, bufptr_type, intBuffer_type, set_type, setarray_type
+
+#ifdef _OPENMP
```
oschuett: Please use
```
!$ USE OMP_LIB, ONLY:  omp_get_max_threads, omp_get_thread_num
```

src/dm_ls_scf_methods.F
```diff
@@ -486,23 +493,60 @@ SUBROUTINE density_matrix_sign_fixed_mu(matrix_p, trace, mu, sign_method, sign_o
...
+
+      IF (sign_symmetric) THEN
+         CALL dbcsr_create(matrix_s_sqrt, template=matrix_s, matrix_type=dbcsr_type_no_symmetry)
+         CALL dbcsr_create(matrix_s_sqrt_inv, template=matrix_s, matrix_type=dbcsr_type_no_symmetry)
+         CALL matrix_sqrt_Newton_Schulz(matrix_s_sqrt, matrix_s_sqrt_inv, matrix_s, threshold, sign_order, &
+                                        1.0E-10_dp, 100, symmetrize, converged) !FIXME
+         CALL dbcsr_release(matrix_s_sqrt)
+         IF (.NOT. converged) THEN
+            PRINT *, "sqrt inv did not converge"
+            STOP
```
oschuett: Please use `CPABORT("sqrt inv did not converge")` instead.

src/submatrix_dissection.F
```diff
@@ -0,0 +1,895 @@
...
+                result_blocks_for_rank_t(0:numthreads - 1), &
+                blocks_from_rank_t(0:numthreads - 1), &
+                nonzero_rows_t(0:numthreads - 1))
+
+      ! Figure out which blocks we need to receive. Blocks are identified here as indices into our COO representation.
+      ! FIXME: This currently shows bad parallel efficiency.
+      !$OMP PARALLEL NUM_THREADS(numthreads) DEFAULT(NONE) &
+      !$OMP          PRIVATE(i,j,k,l,m,l_limit_left,l_limit_right,cur_col,cur_row,mytid) &
+      !$OMP          SHARED(result_blocks_from_rank_t,result_blocks_for_rank_t,blocks_from_rank_t,this,numthreads,nonzero_rows_t)
+#ifdef _OPENMP
```
oschuett: Please instead use
```
mytid = 0
!$  mytid = omp_get_thread_num()
```

src/submatrix_dissection.F
```diff
@@ -0,0 +1,895 @@
...
+      CALL mp_alltoall(num_blockids_send, num_blockids_recv, 1, this%group)
+#else
+      num_blockids_recv(:) = num_blockids_send
+#endif
+      IF (DEBUG) PRINT *, this%myrank, "Number of block ids to receive from each rank:", num_blockids_recv
+      DO i = 0, this%numnodes - 1
+         CALL blocks_for_rank(i)%alloc(num_blockids_recv(i))
+      ENDDO
+      DEALLOCATE (num_blockids_send, num_blockids_recv)
+#ifdef __parallel
```
oschuett: It should not be necessary to have `#ifdef __parallel` in such high-level code. Our MPI wrappers will handle the serial case.

src/dm_ls_scf_methods.F
```diff
@@ -454,28 +456,33 @@ END SUBROUTINE density_matrix_sign
...
       TYPE(dbcsr_type), INTENT(INOUT)                    :: matrix_ks, matrix_s, matrix_s_inv
       REAL(KIND=dp), INTENT(IN)                          :: threshold
+      LOGICAL                                            :: sign_symmetric
+      INTEGER                                            :: submatrix_sign_method
 
       CHARACTER(LEN=*), PARAMETER :: routineN = 'density_matrix_sign_fixed_mu', &
          routineP = moduleN//':'//routineN
 
       INTEGER                                            :: handle, unit_nr
+      LOGICAL                                            :: converged = .FALSE., symmetrize = .TRUE.
```
oschuett: Unfortunately, in Fortran inline initialization likes these will implicitly add the `SAVE` attribute. This makes the variable persistent across invocations and thread-shared. The solution is to simply put the initialization on a separate line.

src/iterate_matrix.F
```diff
@@ -1072,6 +1078,228 @@ SUBROUTINE matrix_sign_proot(matrix_sign, matrix, threshold, sign_order)
...
+         order = 2
+      ENDIF
+
+      CALL dbcsr_get_info(matrix=matrix, nblkcols_total=nblkcols, distribution=dist, group=group)
+      CALL dbcsr_distribution_get(dist=dist, mynode=myrank)
+
+      CALL dissection%init(matrix)
+      CALL dissection%get_sm_ids_for_rank(myrank, my_sms)
+
+      !$OMP PARALLEL PRIVATE(sm, sm_sign, sm_size) SHARED(dissection, myrank, my_sms)
```
oschuett: Please add `DEFAULT(NONE)`.

src/submatrix_dissection.F
```diff
@@ -0,0 +1,895 @@
...
+      CALL dbcsr_iterator_stop(iter)
+
+      IF (DEBUG) THEN
+         PRINT *, this%myrank, "Rows:", this%coo_rows_local
+         PRINT *, this%myrank, "Cols:", this%coo_cols_local
+      ENDIF
+
+      ! We only know how many blocks we own. What's with the other ranks?
+#ifdef __parallel
+      CALL mp_allgather(msgout=this%local_blocks, msgin=blocks_per_rank, gid=this%group)
```
fstein93, line 213: mp_allgather is enough. You can remove the handling with __parallel.

src/submatrix_dissection.F
```diff
@@ -0,0 +1,895 @@
...
+      IF (DEBUG) PRINT *, this%myrank, "Displacements:", coo_dsplcmnts
+
+      ! Get a global view on the matrix
+      this%nblks = SUM(blocks_per_rank)
+      ALLOCATE (this%coo_cols(this%nblks), this%coo_rows(this%nblks), this%coo_col_offsets(this%nblkcols + 1), &
+                this%coo_dptr(this%nblks))
+#ifdef __parallel
+      CALL mp_allgather(msgout=this%coo_rows_local, msgin=this%coo_rows, gid=this%group, rcount=blocks_per_rank, &
+                        rdispl=coo_dsplcmnts)
+      CALL mp_allgather(msgout=this%coo_cols_local, msgin=this%coo_cols, gid=this%group, rcount=blocks_per_rank, &
```
fstein93, line 225: mp_allgather is enough. You can remove the handling with __parallel.

src/submatrix_dissection.F
```diff
@@ -0,0 +1,895 @@
...
+         ENDIF
+      ENDDO
+      IF (DEBUG) PRINT *, "Submatrix owners:", this%submatrix_owners
+
+      ! Prepare data structures for multithreaded loop
+#ifdef _OPENMP
+      numthreads = omp_get_max_threads()
+#else
+      numthreads = 1
+#endif
```
fstein93: Use
    numthreads = 1
!$ numthreads = omp_get_max_threads()

src/submatrix_methods.F
```diff
@@ -0,0 +1,112 @@
...
+
+CONTAINS
+
+! **************************************************************************************************
+!> \brief sort integer array using quicksort
+!> \param arr_a - input array
+!> \param left - left boundary of region to be sorted
+!> \param right - right boundary of region to be sorted
+! **************************************************************************************************
+   RECURSIVE PURE SUBROUTINE qsort(arr_a, left, right)
```
fstein93: I think we have already some sorting routines (module util, routine sort)

src/submatrix_dissection.F
```diff
@@ -0,0 +1,895 @@
...
+
+         IF (DEBUG) PRINT *, "nonzero rows: ", nonzero_rows_t(mytid)%sets(1)%getall()
+
+         ! In the following, we determine all blocks required to build the submatrix. We interpret nonzero_rows_t(mytid)(j) as
+         ! column and nonzero_rows_t(mytid)(k) as row.
+         DO j = 1, nonzero_rows_t(mytid)%sets(1)%elements
+            cur_col = nonzero_rows_t(mytid)%sets(1)%get(j)
+            l_limit_left = this%coo_col_offsets(cur_col)
+            l_limit_right = this%coo_col_offsets(cur_col + 1) - 1
+            ! IF (DEBUG) print *, this%coo_rows(l_limit_left:l_limit_right)
```
pseewald: commented out debug code

src/submatrix_dissection.F
```diff
@@ -0,0 +1,895 @@
...
+               l = l_limit_left
+               DO WHILE (l .LE. l_limit_right)
+                  IF (this%coo_rows(l) .GE. cur_row) EXIT
+                  l = l + 1
+               ENDDO
+               l_limit_left = l
+               IF (l .LE. l_limit_right) THEN
+                  IF (this%coo_rows(l) .EQ. cur_row) THEN
+                     ! We found a valid block. Figure out what to do with it.
+                     ! IF (DEBUG) print *, "Found row", this%coo_rows(l)
```
pseewald: commented out debug code

pseewald (commented):
There are a few todo/fixme comments left. If they concern trivial changes, they should probably be fixed as a part of this pr.

src/submatrix_dissection.F
michaellass: I am indeed able to get rid of those around `mp_allgather` and `mp_alltoall`. However, `mp_isend` and `mp_irecv` raise an error unless `__parallel` is defined:
https://github.com/cp2k/cp2k/blob/master/src/mpiwrap/message_passing.f90#L3015

fstein93: Then you can use something like
```
if (this%numnodes > 1)
   ... code with mp_isend/mp_irecv...
else
   recvbuffer = sendbuffer
end if
```
That's what is done for instance in the MP2 code.

src/submatrix_methods.F
michaellass: Thanks for the hint! I now replaced qsort by sort from util. However, qsort_two behaves differently as it sorts two lists simultaneously and uses the content of both to determine the order. Therefore I kept qsort_two for now.

src/energy_corrections.F
```diff
@@ -1542,7 +1542,8 @@ SUBROUTINE ec_ls_solver(qs_env, ls_method, matrix_ks, matrix_s, matrix_p, matrix
          CASE (ec_matrix_sign)
             CALL density_matrix_sign(matrix_p(ispin, 1)%matrix, mu_spin(ispin), fixed_mu, &
                                      sign_method, sign_order, matrix_ks(ispin, 1)%matrix, &
-                                     matrix_s(1, 1)%matrix, matrix_s_inv, nelectron, eps_filter)
+                                     matrix_s(1, 1)%matrix, matrix_s_inv, nelectron, eps_filter, &
+                                     sign_symmetric=.FALSE., submatrix_sign_method=0)
```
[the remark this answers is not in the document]
michaellass: I made those new arguments optional now so that unrelated code such as energy_corrections does not need to be changed at all.

src/dm_ls_scf_methods.F
```diff
@@ -486,23 +493,60 @@ SUBROUTINE density_matrix_sign_fixed_mu(matrix_p, trace, mu, sign_method, sign_o
...
-      CASE DEFAULT
-         CPABORT("Unkown sign method.")
-      END SELECT
-      CALL dbcsr_release(matrix_sinv_ks)
+
+      IF (sign_symmetric) THEN
+         CALL dbcsr_create(matrix_s_sqrt, template=matrix_s, matrix_type=dbcsr_type_no_symmetry)
+         CALL dbcsr_create(matrix_s_sqrt_inv, template=matrix_s, matrix_type=dbcsr_type_no_symmetry)
+         CALL matrix_sqrt_Newton_Schulz(matrix_s_sqrt, matrix_s_sqrt_inv, matrix_s, threshold, sign_order, &
+                                        1.0E-10_dp, 100, symmetrize, converged) !FIXME
```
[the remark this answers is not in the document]
michaellass: Good catch. This code is now removed and `matrix_s_sqrt_inv` from outside of this routine is used.

michaellass:
Thanks everybody for your reviews! I hope to have addressed all of your feedback in the updated code. The remaining TODOs are mainly entry points for later improvement. They are not to be addressed right now.

michaellass:
Thanks for merging! I see from the dashboard that the code causes an issue with old gfortran versions. I will try to reproduce this locally and send a fix.

michaellass:
This looks like a GCC bug to me which has been solved in later versions. The missing symbol `__vtab_submatrix_types_Set_type` in [submatrix_dissection.F](https://github.com/cp2k/cp2k/blob/5bfd09c5c14741b6ab79a6bf5ac87a0a1b62a14e/src/submatrix_dissection.F#L311) clearly is an internal one. Unfortunately it cannot simply be specified in the shared() clause as it starts with an underscore. The only workaround that I found so far is to specify `DEFAULT(SHARED)` instead of `DEAULT(NONE)`. This however lets the convention check fail.

Side note: The same issues appears in [iterate_matrix.F](https://github.com/cp2k/cp2k/blob/5bfd09c5c14741b6ab79a6bf5ac87a0a1b62a14e/src/iterate_matrix.F#L1289) with the internal symbol `__vtab_submatrix_dissection_Submatrix_dissection_type`.

What are your opinions on this? Would it be acceptible to use DEFAULT(SHARED) and add exceptions to the conventions check? We would still keep the SHARED() and PRIVATE() clauses so it would be easy to go back in future.

oschuett:
I see your made use of [OOP Fortran](https://github.com/pc2/cp2k/blob/e58aa579c42e182a647407accd7dbfe4a2a2d9e7/src/submatrix_dissection.F#L67). While I'm a big fan of modern Fortran, it's unfortunately only usable with recent compilers. So, we have to walk a narrow line between modernizing CP2K and not alienating too many users.

It seems we'd have to drop support for GCC < 8, and possibly for [other compilers](https://www.cp2k.org/dev:compiler_support) as well. So, let's wait for the full dashboard results and then open a new issue to discuss our options.

alazzaro:
If I can add, dropping GCC 7 is too early... I agree with Ole, let's see the dashboard.

dev-zero:
> What are your opinions on this? Would it be acceptible to use DEFAULT(SHARED) and add exceptions to the conventions check? We would still keep the SHARED() and PRIVATE() clauses so it would be easy to go back in future.

Sure, conventions can be ignored if necessary. Maybe use the `__GFORTRAN__` and `__GNUC__` [preprocessor macros](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html) to limit it directly to the affected compiler and version?

alazzaro:
I've just got a problem due to this PR with a new DBCSR version.
It turns out that you are using some internal DBCSR function, namely `dbcsr_mpiwrap`. 
This is internal in DBCSR (I know, we don't have a way to mask it), could you replace these calls with the standard MPI wrappers inside CP2K? The fix is quite easy:
in the file `submatrix_dissection.F` just replace:

`USE dbcsr_mpiwrap, ONLY`

with

`USE message_passing, ONLY`

PR #807, opened 2020-03-09
Update DBCSR v2.1.0.rc7

alazzaro:


alazzaro:
@oschuett This PR should fix the initial problem of the multi-gpu (need one more PR for resetting the GPU). I cannot try the 4x Kepler (it seems a problem with the installation of the toolchain). Is it something you can easily fix?

oschuett:
I'm sorry. #808 should take care of that.

PR #809, opened 2020-03-09
Update DBCSR to v2.1.0.rc7

alazzaro:
This DBCSR provides a first fix for the multi-gpu runs.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -95,23 +95,24 @@ FFTW can be used to improve FFT speed on a wide range of architectures. It is st
   * `-D__MAX_CONTR=4` (default=2) can be used to compile efficient contraction kernels up to l=4, but the build time will increase accordingly.
 
 ### 2h. libsmm (optional, improved performance for matrix multiplication)
-  * A library for small matrix multiplies can be built from the included source (see tools/build_libsmm/README).  Usually only the double precision real and perhaps complex is needed.  Link to the generated libraries. For a couple of architectures prebuilt libsmm are available at https://www.cp2k.org/static/downloads/libsmm/.
+  * A library for small matrix multiplies can be built from the included source (see exts/dbcsr/tools/build_libsmm/README).  Usually only the double precision real and perhaps complex is needed.  Link to the generated libraries. For a couple of architectures prebuilt libsmm are available at https://www.cp2k.org/static/downloads/libsmm/.
   * Add `-D__HAS_smm_dnn` to the defines to make the code use the double precision real library.  Similarly use `-D__HAS_smm_snn` for single precision real and `-D__HAS_smm_znn` / `-D__HAS_smm_cnn` for double / single precision complex.
   * Add `-D__HAS_smm_vec` to enable the new vectorized interfaces of libsmm.
 
 ### 2i. libxsmm (optional, improved performance for matrix multiplication)
   * A library for matrix operations and deep learning primitives: https://github.com/hfp/libxsmm/
-  * Add `-D__LIBXSMM` to enable it (with suitable include and library paths)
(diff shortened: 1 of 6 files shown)


alazzaro:
@oschuett there is still a problem with the 4x Kepler...

oschuett:
The 4x Kepler benchmarks are not yet ready. This is PR #652, which I put on hold because multi-GPU wasn't working. Looking at the 2x Kepler regtest, it [seems](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-kepler-e14fb9b1_report.txt) there still is a problem.

alazzaro:
The reported error message is in PW_CUDA, namely this output:

https://github.com/cp2k/cp2k/blob/8bd161b65d4595e660da4bb714d97cfe7163ab74/src/pw/cuda/pw_cuda_utils.cu#L42-L50

I would assume that the device is not properly set.
Who has time to debug it?

alazzaro:
OK, it seems my last commit gave a first solution to the problem. Currently, there is no mechanism in CP2K to set the current GPU device to use. Threfore, CP2K replies on DBCSR to set the device (i.e. DBCSR init has to be the first call). Let's see if it passes the CI...

oschuett:
>  Threfore, CP2K replies on DBCSR to set the device (i.e. DBCSR init has to be the first call).

I'm afraid we need to set the active device more than just once, because other accelerated libraries are probably using a different GPU assignment. For the time being this might not be a problem because PW_CUDA is not used together with SIRIUS or spFFT. However, before we merge COSMA (#648 ) this should be looked into.

alazzaro:
@oschuett I cannot get what's wrong with the 2x Kepler test. It starts to run, but then it stops. Any clue?

alazzaro:
> > Threfore, CP2K replies on DBCSR to set the device (i.e. DBCSR init has to be the first call).
> 
> I'm afraid we need to set the active device more than just once, because other accelerated libraries are probably using a different GPU assignment. For the time being this might not be a problem because PW_CUDA is not used together with SIRIUS or spFFT. However, before we merge COSMA (#648 ) this should be looked into.

I see two complementary problems here:
1. DBCSR has to set its device (issue https://github.com/cp2k/dbcsr/issues/205). This is my task.
2. All the other GPU parts included in CP2K must have a way to set the GPU device before any call (this is not in place). Voluntaries?

Then, all other libraries (ELPA, SIRIUS, COSMA,...) must have a way to reset the GPU device too, but this is out of our control... Are we sure this is the case?

In principle, DBCSR can reset the device at the beginning and the end (back to the call value), but I would like to avoid the second case...

oschuett:
I agree, it should be the responsibility of each library to setup the GPU.
Let's open an issue for this, It will probably keep us busy for a while.

oschuett:
The 2x Kepler regtest runs out of memory. Interestingly, it doesn't happen immediately. It might just be one test case. So, maybe adding a ulimit to do_regtest would help?

alazzaro:
> The 2x Kepler regtest runs out of memory. Interestingly, it doesn't happen immediately. It might just be one test case. So, maybe adding a ulimit to do_regtest would help?

I see we have `ulimit -t` Could you elaborate a bit more? Can we set `ulimit -v 2097152` too?

alazzaro:
Update:

- `ulimit -v` doesn't work with sanitizer, so I don't use it for *dbg runs
- the 8GB memory limit is too low for the Pascal test, I'm trying to tune it to a better value (12GB?)
- 2x Kepler run OOM with FFT tests (no surprise), I will investigate more
- Overall the GPU tests are very slow

alazzaro:
The 2x Kepler is driving me crazy... It keeps rebooting and the compilation is very slow... the problem is that I cannot reproduce any of error on my multi-gpu system (memory is under control) and `ulimit -v ` seems to give more problems than anything else...
So, I will merge here and will continue the discussion on #814 . PW_CUDA needs a way to set the GPU device (and of course DBCSR too)...

PR #821, opened 2020-03-11
Implement tools/toolchain/scripts/install_intelmpi.sh

e-kwsm:
Support `--mpi-mode=intelmpi` and `--with-intelmpi`


oschuett:
Thanks for this contribution :-)

Could you please also add `install_intelmpi.sh` to our [Dockerfile](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/Dockerfile#L52). (I know, it's not obvious).

Furthermore, it would be good to test this on a regular basis, which could be done by this [other Dockerfile](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/Dockerfile.cuda_mkl).

PR #837, opened 2020-03-17
Fix thresholds of regtests QS/regtest-hfx-wfn-fitting/CH4-rsLDA.inp a…

fstein93:
…nd QS/regtest-hfx-ri/H2O-hfx-coulomb.inp to match Dashboard regtester Ubuntu, GCC 8 (ssmp)


fstein93:
@pseewald : Your regtest will also be fixed.

PR #861, opened 2020-04-13
Pr843 addendum

hfp:
Use more flexible "rank-remapping" (non-F08). In fact, this pattern can "cast" without restrictions regarding the source/target ranks unlike F08's rank-remapping syntax. The expression, "mat => omat(1:ii, 1:ii)" on the other hand produces an array-temporary.

### DIFF
--- a/src/atom_operators.F
+++ b/src/atom_operators.F
@@ -311,7 +311,7 @@ CONTAINS
                END DO
                integrals%nne(l) = ii
                IF (ii > 0) THEN
-                  mat(1:ii, 1:ii) => omat
+                  mat => omat(1:ii, 1:ii)
                   mat = MATMUL(TRANSPOSE(integrals%utrans(1:n, 1:ii, l)), integrals%utrans(1:n, 1:ii, l))
 
                   DO i = 1, ii
--- a/src/atom_operators.F
+++ b/src/atom_operators.F
@@ -11,6 +11,8 @@
 !>
 ! **************************************************************************************************
 MODULE atom_operators
+   USE ISO_C_BINDING,                   ONLY: C_F_POINTER, &
+                                              C_LOC
    USE ai_onecenter,                    ONLY: &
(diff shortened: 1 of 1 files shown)


hfp:
See also this [comment](https://github.com/cp2k/cp2k/commit/470af4cea81688a0e1a1e35e4925c681871bf8a5#commitcomment-38447840).

oschuett:
Is there really no other way than `C_F_POINTER(C_LOC(foo))`?
I'd hate for this to become a common pattern.

alazzaro:
Well, I don't how critical is that region, but I would propose to leave the temporary array and drop C_* functions...

hfp:
I will just revert to the original code for atom_operators, but keep the other (minor) changes of #843. In general, there are several messages about temporary arrays depending on which input file is running. An example looks like:

```text
At line 322 of file /home/hfp/projects/cp2k/src/atom_operators.F
Fortran runtime warning: An array temporary was created for argument 'b' of procedure 'dgesv'
```

These messages are generated when compiling with CP2K's default debug flags. Note, reverting to before PR843 will create temporaries for both the `a` and the `b` argument of `dgesv`.

I wonder if CP2K's dashboard should collect a status for this sort of runtime warning (grep'ping this message from the log/output of all regtests followed by unique sort). On the other hand, this "runtime warning" could be almost a false positive uniquely attributed to (unoptimized) debug builds. If it is true for optimized builds, it's worrysome as even low-level routines are affected (MPI wrapper, etc.).

oschuett:
@hfp, I'm grateful that you're looking into these warnings!

I was just expecting that one can avoid these copies without resorting to `C_F_POINTER`, e.g. by refactoring the called function. If in some cases a copy is unavoidable or worth the convenience, then we should create it explicitly to document it and eliminate the warning.

Of course, if you think that these warnings are mostly false alerts, then we should simply turn them off.

PR #864, opened 2020-04-15
eri_3center optional extreme value and improved pgf_sum_2c_gspace_1d_deltal

hfp:
* libint_2c_3c.F: introduced calculating the extreme value as part of eri_3center, which avoids revisiting the cube of integrals again once it is out of cache. The extreme value is calculated as part of eri_3center's flow in a cache-friendly and chunk-wise fashion. The extreme value is regularly used to prune integrals (screening).

* eri_mme_lattice_summation.F: removed inner condition from loop (routine yields ~10% faster code with GNU Fortran).

The optional extreme value for eri_3center makes subsequent code (screening) much faster. However, the latter is a separate PR.

### DIFF
--- a/src/eri_mme/eri_mme_lattice_summation.F
+++ b/src/eri_mme/eri_mme_lattice_summation.F
@@ -669,26 +669,40 @@ CONTAINS
       CHARACTER(LEN=*), PARAMETER :: routineN = 'pgf_sum_2c_gspace_1d_deltal', &
                                      routineP = moduleN//':'//routineN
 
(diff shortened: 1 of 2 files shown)


hfp:
@pseewald Thank you for the thumbs up. Will merge asap but wait a bit more for objections. @abussy This code is actually used in this [branch](https://github.com/hfp/cp2k/tree/libxsmm_contract), and improves fill_pqX_tensor quite a bit (xas_tdp_integrals.F).

PR #868, opened 2020-04-16
Make use of extreme value determined by revised eri_3center

hfp:


abussy:
https://github.com/cp2k/cp2k/blob/cb19e6298c1aceee2f0781274c8d6957ee803ad4/src/libint_2c_3c.F#L193-L205

At line 202, it should probably be `p_work(p2 + 1:p2 + ncoa)`, I missed it in yesterday's PR

edit: at line 236 too

hfp:
@abussy Thank you very much, I was probably sleeping when writing this code! When it passes, I will add a similar change for the overlap3 routine, i.e., update once more.

PR #869, opened 2020-04-16
collect code optimizations on top of #864 and #868

hfp:
* Revised #864 and #868 at the expense of some (minor) code duplication.
* Attempt to avoid zeroing (intermediate) array when setup as a whole.

### DIFF
--- a/src/common/gamma.F
+++ b/src/common/gamma.F
@@ -209,41 +209,42 @@ CONTAINS
       ELSE
 
 !     *** t > 12 ***
+         tmp = 1.0_dp/t ! reciprocal value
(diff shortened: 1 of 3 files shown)


hfp:
@abussy Thank you!  I will revert that line, but meanwhile I triggered a non-debug build just for curiosity. Apparently, the case you pointed out does not seem to be covered by any of the tests.

abussy:
@hfp Yes indeed. Since the regtests are all rather small, it is possible that this specific radius based screening is never triggered. It does play an important role when systems get larger and sparser though.

PR #875, opened 2020-04-20
Optimization of 3-center integrals contraction using LIBXSMM

abussy:
In this PR, LIBXSMM is used to accelerate the contraction of 3-center integrals from pgf to sgf. The present code together with #869 allows for a cost reduction of 30% to 50% on the 3-center integrals calculation in XAS_TDP (depending on the simulated system and machine used).

@pseewald If you think that this can be useful for your tensor code and/or other methods using 3-center integrals, let me know and I can make the changes there too.

@hfp I stopped short of writing a version dependent code and simply check for the -D__LIBXSMM flag with a fallback on BLAS. I tested the code with versions 1.9.0 and the current development branch, assuming everything in between would work.

### DIFF
--- a/src/aobasis/ai_contraction_sphi.F
+++ b/src/aobasis/ai_contraction_sphi.F
@@ -254,7 +254,7 @@ CONTAINS
       INTEGER                                            :: handle, i
 #if(__LIBXSMM)
       INTEGER(libxsmm_blasint_kind)                      :: m, n, k
-      TYPE(libxsmm_dmmfunction), DIMENSION(2)            :: xmm
+      TYPE(libxsmm_dmmfunction)                          :: xmm1, xmm2
 #endif
 
       CALL timeset(routineN, handle)
@@ -265,14 +265,14 @@ CONTAINS
 
       !pre-fetch the kernels
       m = nsgfa; n = ncob; k = ncoa
-      CALL libxsmm_dispatch(xmm(1), m, n, k, beta=0.0_dp, prefetch=LIBXSMM_PREFETCH_NONE)
+      CALL libxsmm_dispatch(xmm1, m, n, k, beta=0.0_dp, prefetch=LIBXSMM_PREFETCH_NONE)
       m = nsgfa; n = nsgfb; k = ncob
-      CALL libxsmm_dispatch(xmm(2), m, n, k, beta=0.0_dp, prefetch=LIBXSMM_PREFETCH_NONE)
+      CALL libxsmm_dispatch(xmm2, m, n, k, beta=0.0_dp, prefetch=LIBXSMM_PREFETCH_NONE)
 
       !contractions over a and b
       DO i = 1, ncoc
-         CALL libxsmm_dmmcall(xmm(1), tsphi_a, sabc((i - 1)*ncoa*ncob + 1), cpp_buffer)
-         CALL libxsmm_dmmcall(xmm(2), cpp_buffer, sphi_b, ccp_buffer((i - 1)*nsgfa*nsgfb + 1))
+         CALL libxsmm_dmmcall(xmm1, tsphi_a, sabc((i - 1)*ncoa*ncob + 1), cpp_buffer)
+         CALL libxsmm_dmmcall(xmm2, cpp_buffer, sphi_b, ccp_buffer((i - 1)*nsgfa*nsgfb + 1))
       END DO
 
 #else
@@ -289,7 +289,7 @@ CONTAINS
 
 #endif
 
-      !last contraciton, over c, as a larfer MM
+      !last contraciton, over c, as a larger MM
       CALL dgemm("N", "N", nsgfa*nsgfb, nsgfc, ncoc, 1.0_dp, ccp_buffer, nsgfa*nsgfb, sphi_c, ncoc, &
                  0.0_dp, abcint, nsgfa*nsgfb)
 
--- a/src/aobasis/ai_contraction_sphi.F
+++ b/src/aobasis/ai_contraction_sphi.F
@@ -14,8 +14,6 @@ MODULE ai_contraction_sphi
 
    USE kinds, ONLY: dp
 #if(__LIBXSMM)
-   USE ISO_C_BINDING, ONLY: C_F_POINTER, &
-                            C_LOC
    USE libxsmm, ONLY: LIBXSMM_PREFETCH_NONE, &
                       libxsmm_blasint_kind, &
                       libxsmm_dgemm, &
@@ -225,12 +223,12 @@ CONTAINS
 !> \brief 3-center contraction routine from primitive cartesain Gaussians to spherical Gaussian
 !>        functions. Exploits LIBXSMM for performance, falls back to BLAS if LIBXSMM not available.
 !>        Requires pre-allocation of work buffers and pre-transposition of the sphi_a array. Requires
-!>        the LIBXSMM libreary to be initialized somewhere before this routine is called.
+!>        the LIBXSMM library to be initialized somewhere before this routine is called.
 !> \param abcint contracted integrals
 !> \param sabc uncontracted integrals
-!> \param tsphi_a ...
-!> \param sphi_b ...
-!> \param sphi_c ...
+!> \param tsphi_a assumed to have dimensions nsgfa x ncoa
+!> \param sphi_b assumed to have dimensions ncob x nsgfb
+!> \param sphi_c assumed to have dimensions ncoc x nsgfc
 !> \param ncoa ...
 !> \param ncob ...
 !> \param ncoc ...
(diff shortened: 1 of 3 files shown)


hfp:
@abussy looks all good. I think you could cleanup/shrink `mnk`  or replace by scalars. That array was just introduced to tryout the user-defined dispatch thing which needed a binary "blob" as a key to retrieve the set of kernels.

hfp:
I would also just call `dgemm` rather than `libxsmm_dgemm` unless you found it makes a difference; I would be (positively ;-) surprised if libxsmm_dgemm makes an impact. Assuming the last thing is always dgemm, you could even shrink the region of code that depends on LIBXSMM...

abussy:
From the result of my experiments, I can say that in the large matrix limit, standard `dgemm` and `libxsmm_dgemm` are equivalent. However, for small matrices, I do observe a speed-up with `libxsmm_dgemm`.

src/aobasis/ai_contraction_sphi.F
```diff
@@ -210,4 +221,90 @@ SUBROUTINE abcd_contract(abcdint, sabcd, sphi_a, sphi_b, sphi_c, sphi_d, ncoa, n
...
+
+      !contractions over a and b
+      DO i = 1, ncoc
+         CALL libxsmm_dmmcall(xmm(1), tsphi_a, sabc((i - 1)*ncoa*ncob + 1), cpp_buffer)
+         CALL libxsmm_dmmcall(xmm(2), cpp_buffer, sphi_b, ccp_buffer((i - 1)*nsgfa*nsgfb + 1))
+      END DO
+
+      !last contraction, over c, as a larger MM
+      !note: libxsmm_dgemm falls back to BLAS if matrix too large
+      CALL C_F_POINTER(C_LOC(ccp_buffer), ccp, [nsgfa*nsgfb, ncoc])
```
oschuett: I think, we should reserve `ISO_C_BINDING` for when we're directly interfacing with a C code. Here we're using the Fortran binding provided by libxsmm.

hfp:
> I do observe a speed-up with libxsmm_dgemm.

Perfect, then just ignore my related comment.

src/aobasis/ai_contraction_sphi.F
abussy: I'll see if I can circumvent it. This is clearly better than a expensive `RESHAPE` call though.

hfp: If it creates temporaries, it will erode the value of this PR.

I found it hard to provide a Fortran interface in LIBXSMM that is somewhat "generic" or flexible. If one passes something as `array(*)` it accepts any rank including scalar. However, that does not apply anymore when involving such a subroutine/signature into a MODULE PROCEDURE interface, then the call-side must exactly match in rank (which is rank-1 in case of `array(*)`). So it looks like one either get all ranks _or_ type-overload (not both). Though, passing mixed ranks to the afore mentioned signature may cause array-repack.

src/xas_tdp_integrals.F
```diff
@@ -547,11 +607,10 @@ SUBROUTINE fill_pqX_tensor(pq_X, ab_nl, ac_nl, basis_set_list_a, basis_set_list_
                   END IF
 
                   ALLOCATE (work(nsgfa(iset), nsgfb(jset), nsgfc(kset)))
-                  !zeroing work is superfluous here (work = 0.0_dp)
 
-                  CALL abc_contract(work, sabc, &
-                                    sphi_a(:, sgfa:), sphi_b(:, sgfb:), sphi_c(:, sgfc:), &
-                                    ncoa, ncob, ncoc, nsgfa(iset), nsgfb(jset), nsgfc(kset))
+                  CALL libxsmm_abc_contract(work, sabc, tspa(iset, ikind)%array, spb(jset, jkind)%array, &
```
dev-zero, line 604: `libxsmm_abc_contract` is a no-op without libxsmm, so this results in no contraction then? Or did I miss something?

abussy: There is a fallback on BLAS in the case where libxsmm is not linked, which yields the same result, just not as fast. Edit: the fallback is implemented within the `libxsmm_abc_contract` routine.

src/aobasis/ai_contraction_sphi.F
abussy: I am not sure I see the problem with using `ISO_C_BINDING` here. Now if it is actually problematic, then the simplest solution is to use BLAS' dgemm routine instead, like in the fallback code (line 301-302). This would come at a slight performance cost when the MM is actually small though.

src/xas_tdp_integrals.F
dev-zero: ok, I missed the `#else` in `libxsmm_abc_contract`. But this kind of stuff is just one huge anti-pattern, we have to come up with something better at some point. Maybe an optional attribute `preferred_driver` for routines which can use different backends to select the preferred driver for the operation in this call? Well, but not your issue here.

src/aobasis/ai_contraction_sphi.F
dev-zero: @abussy the problem is that we should use one common interface towards a different library for consistency reason (which is the provided Fortran interface in the case of libxsmm). But if it can't be avoided, please add some documentation why it was done this way.

oschuett: Looking at it again, I don't understand why we can't simply call `libxsmm_dgemm` on the 1D array. Isn't [this wrapper](https://github.com/hfp/libxsmm/blob/master/src/template/libxsmm.f#L1371) suitable?

Generally I think that, abusing `ISO_C_BINDING` in this way has pretty bad code-smell. It circumvents the Fortran bounds checks and it fails in subtle ways when used on a non continues pointer.

The proper way is to use F2008 rank remapping:
```
REAL,TARGET :: x(100,100)
REAL,POINTER :: x1(:)
x1(1:Size(x)) => x
```

abussy: The wrapper you mention together with 1D arrays does work, although only with the development version of the library. I believe some of the fortran wrappers have been fixed/modified by @hfp lately. 

The potential failure of `C_F_POINTER` for rank remapping with non-contiguous arrays is probably the strongest argument against using it. I was not aware of F2008 pointer rank remapping, does it prompt the creation of temporary arrays though ?

hfp: No, it does not. It's a viable solution too. However, some compilers seem to have problems with this F08 feature in some if not all cases. @alazzaro recently came accross such case (CCE). Also, it's limited in terms of what ranks can be remapped (only 1:n allowed, i.e., from rank-1 to rank-n array).

> rank remapping with non-contiguous arrays [...] argument against using it

It can be fixed marking the related arrays as `CONTIGUOUS`. I could argue against ISO_C_BINDING (makes Fortran not Fortran anymore), but in the end it should be/is well-defined (if not part of Fortran ;-). I would rather take C_F_POINTER/CONTIGUOUS than spurious temporary arrays. Perhaps this pattern can be permitted by sealing/wrapping it into some utility ("cp_rank_remap").

oschuett: @abussy, thanks for making that change. The regtests passed, so I guess it does work with the latest libxsmm release too.

>  I would rather take`C_F_POINTER/CONTIGUOUS` than spurious temporary arrays. 

My [understanding](https://www.ibm.com/support/knowledgecenter/SSGH4D_15.1.2/com.ibm.xlf151.aix.doc/language_ref/contiguous.html?view=kc#contiguous__contig_rules) is that `CONTIGUOUS` will causes the compiler to create a temporary copy when necessary. And unless the pointer is passed as `INTENT(IN)` it will even copy it back when the subroutine returns. 

One simply can not avoid a copy when non-contiguous data has to be accessed in a contiguous fashion. Fortunately, most of the pointers in CP2K are contiguous, because we [don't actually need them](https://github.com/cp2k/cp2k/pull/482#issuecomment-519695424).

Once an array is contiguous one can reshape it without moving the data, which is called rank remapping in Fortran. Until F2008 the remapping target had to be of rank one and some compilers [still struggle](https://github.com/cp2k/cp2k/commit/470af4cea81688a0e1a1e35e4925c681871bf8a5#r38447840) with the more general case.

That being said, I believe reshaping is hardly ever needed, because it's not a mathematical operation. You will find that the CP2K code bases uses it only in very few places.

alazzaro: > My understanding is that CONTIGUOUS will causes the compiler to create a temporary copy when necessary. And unless the pointer is passed as INTENT(IN) it will even copy it back when the subroutine returns.

It is actually the opposite. You can test it by adding the flag `-Warray-temporaries`. With CONTIGUOUS the compiler doesn't do any test and it will assume the data to be contiguous (you are in trouble if they are not)

abussy: > @abussy, thanks for making that change. The regtests passed, so I guess it does work with the latest libxsmm release too.

I actually opted out of LIBXSMM for that last MM and used BLAS's `dgemm`.

hfp:
Looks good to me!  If you like, you can turn the `xmm` array into two scalars and mark the PR as ready.

src/aobasis/ai_contraction_sphi.F
oschuett: @alazzaro, you are right. The back and forth copying only happens when `CONTIGUOUS` is added to an assumed-shape argument.

> With CONTIGUOUS the compiler doesn't do any test and it will assume the data to be contiguous (you are in trouble if they are not)

No, adding `CONTIGUOUS` to a pointer makes it different type, for which the compiler *guarantees* that it will only ever be associated with a contiguous target.

The downside is that one has to annotate all pointers along the call chain, which is exactly what you're currently doing with https://github.com/cp2k/dbcsr/pull/341. While for DBCSR this makes sense, in CP2K we make regular use of submatrices. Hence, I'm going to disable the warnings in CP2K (#878).

pseewald (approved):

mkrack:
@abussy this commit has issues. Check the traceback [here](https://dashboard.cp2k.org/archive/psi-gnu-psmp/commit_603c96205b10feebbc36ce2b1df3460fb2e1a18d.txt) and more recent [here](https://www.cp2k.org/static/regtest/trunk/Linux-x86-64-gfortran-regtest/psmp/regtest-0)

Issue #317, opened 2019-04-24
[sirius] WARNING in input/input_keyword_types.F:284 :: keyword XC_FUNCTIONALS


dev-zero:
When building CP2K with support for SIRIUS, I get the following when running `cp2k --help`:

```console
$ cp2k.popt --help

 cp2k.popt [-c|--check] [-e|--echo] [-h|--help] [--html-manual]
           [-i] <input_file>
           [-mpi-mapping|--mpi-mapping] <method>
           [-o] <output_file>
           [-r|-run] [--xml]

 starts the CP2K program, see <https://www.cp2k.org/>

 The easiest way is cp2k.popt <input_file>

 The following options can be used:

  -i <input_file>   : provides an input file name, if it is the last
                      argument, the -i flag is not needed
  -o <output_file>  : provides an output file name [default: screen]

 These switches skip the simulation, unless [-r|-run] is specified:

  --check, -c       : performs a syntax check of the <input_file>
  --echo, -e        : echos the <input_file>, and make all defaults explicit
                      The input is also checked, but only a failure is reported
  --help, -h        : writes this message
  --html-manual     : writes a HTML reference manual of the CP2K input
                      in the current directory. The file index.html is a good
                      starting point for browsing
  --license         : prints the CP2K license
  --mpi-mapping     : applies a given MPI reordering to CP2K
  --run, -r         : forces a CP2K run regardless of other specified flags
  --version, -v     : prints the CP2K version and the revision number
  --xml             : dumps the whole CP2K input structure as a XML file
                      xml2htm generates a HTML manual from this XML file
  

 *** WARNING in input/input_keyword_types.F:284 :: keyword XC_FUNCTIONALS ***
 *** assumed undefined type by default                                    ***
```

@mtaillefumier unfortunately this only happens when building CP2K with SIRIUS support.

dev-zero:
this is with SIRIUS v6.1.5 and libvdwxc v0.4.0 and CP2K  9b85492

mtaillefumier:
I will have to find a way to fix it. XC_FUNCTIONALS should take a list of strings but I do not know how to do that with the cp2k parser. It has no implication though since I use the cp2k XC_FUNCTIONAL information instead.

mkrack:
[...]

PR #877, opened 2020-04-21
Update of sirius

mtaillefumier:
- Update sirius to latest version (bug fix in calculation of stress tensor)
- fix issue #317

### DIFF
--- a/src/input_cp2k_pwdft.F
+++ b/src/input_cp2k_pwdft.F
@@ -169,106 +169,114 @@ CONTAINS
 
 !     we need to null char since SIRIUS interface is basically C
          name = TRIM(ADJUSTL(name))//CHAR(0)
-         description = CHAR(0)
-         usage = CHAR(0)
-         CALL sirius_option_get_description_usage(section_name, name, description, usage)
-         SELECT CASE (ctype)
-         CASE (1)
-            CALL sirius_option_get_int(section_name, name, dummy_i, vec_length)
-            CALL keyword_create(keyword, __LOCATION__, &
-                                name=TRIM(name1), &
-                                description=TRIM(ADJUSTL(description)), &
-                                usage=TRIM(ADJUSTL(usage)), &
-                                repeats=.FALSE., &
-                                default_i_val=dummy_i)
-         CASE (11)
-            CALL sirius_option_get_int(section_name, name, ivec(1), vec_length)
-            CALL keyword_create(keyword, __LOCATION__, &
-                                name=name1, &
-                                description=TRIM(ADJUSTL(description)), &
-                                usage=TRIM(ADJUSTL(usage)), &
-                                repeats=.FALSE., &
-                                n_var=vec_length, &
-                                default_i_vals=ivec(1:vec_length))
-         CASE (2)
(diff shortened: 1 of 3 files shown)


mtaillefumier:
Wait before running the parallel regtests.

mtaillefumier:
Can someone launch the psmp regtests. Thanks

src/input_cp2k_pwdft.F
```diff
@@ -169,106 +169,112 @@ SUBROUTINE fill_in_section(section, section_name)
...
+                  CALL keyword_create(keyword, __LOCATION__, &
+                                      name=name1, &
+                                      description=TRIM(ADJUSTL(description)), &
+                                      usage=TRIM(ADJUSTL(usage)), &
+                                      repeats=.FALSE., &
+                                      enum_i_vals=enum_i_val(1:num_possible_values), &
+                                      enum_c_vals=possible_values(1:num_possible_values), &
+                                      default_i_val=0)
+               ELSE
+                  CALL keyword_create(keyword, __LOCATION__, &
```
oschuett, line 267: I think, this will still trigger the warnings from #317. You need to specify the type of the keyword, either explicitly via `type_of_var` or implicitly by providing a default value.

mtaillefumier: Actually, what trigger it is somewhere else. I excluded these two keywords from the list because they are useless for cp2k. Information about xc_functional are taken from libxc section, and memory_usage is only useful for devs.

I fixed it yesterday but I had a problem with github which means it did not go through. Should be fixed now.

oschuett: Yeah, Github is a bit flaky lately...

My point is that, every call to `keyword_create` should have either `default_?_val` or  `type_of_var` as an argument. 
And currently [this one](https://github.com/cp2k/cp2k/blob/bd5662a07fca9e0db6e8fa5041f9b7f1b32a08ed/src/input_cp2k_pwdft.F#L267-L271)  has neither.

Btw, once this is fixed, I'll turn [that warning](https://github.com/cp2k/cp2k/blob/master/src/input/input_keyword_types.F#L286) into an abort so our regtests can catch it.

mtaillefumier: We should document that ;). I just fixed it.

oschuett:
This PR includes many existing commits. I'm surprised it doesn't have merge conflicts. Maybe you could do some `git rebase -i`.

mtaillefumier:
> This PR includes many existing commits. I'm surprised it doesn't have merge conflicts. Maybe you could do some `git rebase -i`.

mtaillefumier:
Sorry got lost with the github interface.

mtaillefumier:
I will revert back to the sirius update commit and push force from there. it will be cleaner

oschuett:
I launched the pdbg check run. Seems like there is a compilation issue.

oschuett:
Sorry for neglecting this PR. All tests are passing. Should I merge this now?

mtaillefumier:
I updated SIRIUS to the latest version. If everything pass then you can merge it.

PR #879, opened 2020-04-22
Fix def2-SVP basis for hydrogen

e-kwsm:
Polarization function was not included, i.e., def2-SV(P)

https://www.basissetexchange.org/basis/def2-sv(p)/format/cp2k/?version=1&elements=1
https://www.basissetexchange.org/basis/def2-svp/format/cp2k/?version=1&elements=1


data/EMSL_BASIS_SETS
```diff
@@ -32659,15 +32659,17 @@ FE NASA-Ames-cc-pVTZ
 H  Ahlrichs-def2-SVP
   2
```
fstein93, line 32660: That has to be a "3" then. Otherwise, the parser will only read the first two sets.

fstein93 (changes requested):

data/EMSL_BASIS_SETS
e-kwsm: Thanks.

fstein93 (approved):

oschuett:
This PR slipped through our CI tests, because the [data](https://github.com/cp2k/cp2k/tree/master/data) directory is missing from the [trigger pattern](https://github.com/cp2k/cp2k-ci/blob/2a3f98dfaa9efcd7325ce354e7af675a03e73c11/backend/cp2k-ci.conf#L146).
I'm going to fix this.

PR #880, opened 2020-04-23
parsing: add options to mark keywords as deprecated/removed

dev-zero:


dev-zero:
Hmm, why exactly wasn't that triggered before:

https://github.com/cp2k/cp2k/blob/55f6b15fd96a0fce25305a0e74b0b20b39f63b9f/src/qs_kind_types.F#L1596-L1601

oschuett:
Because [that test](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/check_inputs.py) only runs `cp2k.x --check`. This show that moving the deprecation check to the parsing stage is a really good idea :-)

dev-zero:
> Because [that test](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/check_inputs.py) only runs `cp2k.x --check`. This show that moving the deprecation check to the parsing stage is a really good idea :-)

Thanks.

oschuett:
Btw, are you also planning to surface the deprecation notices in the [html manual](https://github.com/cp2k/cp2k/blob/master/tools/manual/cp2k_input.xsl)?

PR #881, opened 2020-04-23
Update regtests after fixing of def2-SVP Hydrogen

abussy:
Basis set was changed without checking regtests depending on it


abussy:
Branch did not originate from master, sorry.

Issue #872, opened 2020-04-17
Reference to dangling pointer


vzecca:
Running the QS/regtest-sos-mp2-lr/H2O-sos-mp2-cubic-lr test case at line 802 of dbcsr_tensor.F
the dangling pointer TENSOR_CONTR_1 is referenced.

PR #886, opened 2020-04-27
RI HFX: memory and load balance optimizations

pseewald:


pseewald:
This pr includes an update of DBCSR to see whether the failed test is due to #872

pseewald:
There is a problem with this DBCSR release (see regtests), could you have a look @alazzaro?

The line where CP2K aborts is: https://github.com/cp2k/cp2k/blob/master/src/cp_dbcsr_operations.F#L230
I found that `nfullcols_total` is not correct (strangely, this only happens for 4 regtests).

I tried tracing back the issue to a specific DBCSR commit since the latest release but I failed since most of them don't compile with CP2K. I verified that none of my commits introduced this issue.

valgrind does not detect any uninitialized variables or any other issues.

PR #904, opened 2020-05-07
Kab param

ducryf:
The KAB_PARAM in the XTB section is ignored in the current version. The value is read in and printed to the restart file, but never used in the calculation.  This is fixed and KAB_PARAM is now correctly applied.


dev-zero:
maybe I missed it, but is it possible that `xtb_control%kab_vals` and `xtb_control%kab_types` don't get deallocated?

juerghutter:
Yes, kab_vals and kab_types have to be deallocated in cp_control_types.F in routine
xtb_control_release.

juerghutter:
Could you also add an example input in cp2k/tests/xTB/regtest-_ ?

ducryf:
I have added the regtest cp2k/tests/xTB/regtest-1/h2_kab.inp
and fixed the missing DEALLOCATE  inf xtb_control_release. I have also added NULLIFYs to xtb_control_create.

PR #906, opened 2020-05-08
Add ARM GNU regression tester

mkrack:


oschuett:
Wohoo, very nice :-)
It's also awesome that you managed to get so many libraries to work on ARM.

mkrack:
It was less painful than expected. It works with the current toolchain under Raspbian. I thought why not running a CP2K regression tester when the kodi media center is not used, though a full regression test run takes half a day.

PR #912, opened 2020-05-12
Drop Intel compiler flag -fp-model precise

mkrack:


dev-zero:
For the Spack package maintenance I need some more information about this change: As far as I remember, the Intel compiler may perform unsafe optimization (like commutative reordering) without the `-fp-model precise` and therefore possibly have similar impact to the `-fast-math` flag.
Has that changed? Is there a noticeable speedup?

Can you please add additional comments for such changes in the future? Since the arch files serve as a template for example for the Spack package it would help to know why things get changed. This is what we currently have in Spack:

https://github.com/spack/spack/blob/a44f487a1d5b2f88675d4053485c338c64340a76/var/spack/repos/builtin/packages/cp2k/package.py#L244-L252

mkrack:
I got a note from @alazzaro yesterday that the -fp-model precise flag is most likely dispensable with the Intel 18 and 19 compilers. I ran the PSI Intel 18 and 19 regtests and the results were clean. The default for `-fp-model` is `fast=1` which is indeed a kind of "fast-math" but possibly less aggressive. I would keep for the time being the `-fp-model precise` flag in the Spack configuration. Let's see how the dashboard results evolve.

alazzaro:
This is correct, Intel changed since v18. For a legacy reason I would keep the precise too, it is mostly harmless...

PR #914, opened 2020-05-13
Libxc 5.1

### LINKED ISSUE #909: libxc-5.1 support
Hi all,

Are you planning to add Libxc 5 support? I tried to build cp2k 7.1.0 with this version of the library, but got a lot of errors like:  
`Error: Type mismatch in argument ‘np’ at (1); passed INTEGER(4) to INTEGER(8)`
when compiling src/xc/xc_libxc.F

Thanks in advance

fstein93:
This PR pushes Libxc 4.x to Libxc 5.1.
After merging this PR, Libxc will no longer work with Libxc 4.x.
CP2K needs the Hotfix 5.1 due to the interface problem.
Fixes #909 .
I have enabled Libxc support for LDA-based hybrids from Libxc.


tools/toolchain/scripts/install_libxc.sh
```diff
@@ -30,7 +30,8 @@ case "$with_libxc" in
                 echo "libxc-${libxc_ver}.tar.gz is found"
             else
                 download_pkg ${DOWNLOADER_FLAGS} ${libxc_sha256} \
-                             https://www.cp2k.org/static/downloads/libxc-${libxc_ver}.tar.gz
+                             http://www.tddft.org/programs/libxc/down.php?file=5.0.0/libxc-5.0.0.tar.gz \
```
dev-zero: please use `${libxc_ver}` instead of hardcoding the version

dev-zero:
so, to summarize: they changed the prefix from `f03` to `f90` and are now using `ISO_C_BINDING` datatypes in a couple of places which was introduced in F2003?
About the reference value: would it make sense to update the reference value rather than the threshold?

fstein93:
About the prefix: In 4.x, the f90 interface became deprecated and should have been removed with 5.0. Unfortunately, they removed the f03 interface. The hotfix will revert the removal of the f03 interface and remove the f90 bindings instead.
About the reference value: I have updated the reference value. But as it seems the calculation using TPSS provides similar but no numerically equal values across the different executables. There has been a change in the implementation of TPSS in LibXC. That is why I am using a rather large threshold for this regtest. I will check an alternative mGGA-hybrid, then.

fstein93:
I've replaced the functional in the respective regtest. It is more stable than TPSSh. I don't know what the actual issue with TPSSh is because we didn't have these problems with Libxc 4.3.4.

mkrack:
I suggest to wait with moving to libxc 5 until the f03 interfaces are available again in the latest libxc 5 release.

fstein93:
I have adjusted the wrapper to the hotfix. I will push it to my repository after the hotfix will have been published (probably by the end of next week).

src/xc/xc_libxc_wrap.F
```diff
@@ -23,11 +23,14 @@ MODULE xc_libxc_wrap
...
 #include <xc_version.h>
 ! check for LibXC version
-#if (XC_MAJOR_VERSION < 4)
-   This version of CP2K ONLY works with libxc versions 4.0.3 and above.
+#if (XC_MAJOR_VERSION < 5)
+   This version of CP2K ONLY works with libxc versions 5.1.0 and above.
    Furthermore, -I${LIBXC_DIR}/include needs to be added to FCFLAGS.
 #else
 
+#if (XC_MAJOR_VERSION == 5 && XC_MINOR_VERSION < 1)
```
dev-zero: maybe merge with the above?

src/xc_write_output.F
```diff
@@ -48,7 +48,7 @@ SUBROUTINE xc_write(output_unit, xc_section, lsd)
 
       CHARACTER(LEN=2*default_string_length)             :: shortform
       CHARACTER(LEN=20)                                  :: tmpStr
-      CHARACTER(LEN=20*default_string_length)            :: reference
+      CHARACTER(LEN=200*default_string_length)           :: reference
```
dev-zero: is there a way we can inspect the length needed at runtime and use `ALLOCATE(...)` instead?

tools/toolchain/scripts/install_libxc.sh
```diff
@@ -30,12 +30,13 @@ case "$with_libxc" in
...
             else
                 download_pkg ${DOWNLOADER_FLAGS} ${libxc_sha256} \
-                             https://www.cp2k.org/static/downloads/libxc-${libxc_ver}.tar.gz
+                             https://gitlab.com/libxc/libxc/-/archive/${libxc_ver}/libxc-${libxc_ver}.tar.gz
             fi
             echo "Installing from scratch into ${pkg_install_dir}"
             [ -d libxc-${libxc_ver} ] && rm -rf libxc-${libxc_ver}
             tar -xzf libxc-${libxc_ver}.tar.gz
             cd libxc-${libxc_ver}
+            autoreconf -i > autoreconf.log 2>&1
```
dev-zero: The tarball at https://www.tddft.org/programs/libxc/download/ seems to contain `configure` script, could we use that one instead? This requires the user to have a full autoconf/automake/m4 toolchain installed.

dev-zero: ... but that's what you probably meant by "adapted to the hotfix" I guess?

oschuett:
Since you are working on the libxc integration....

I think, it would be nice if we could tread libxc functional similar to our builtin ones and list them alongside in the [manual](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/XC/XC_FUNCTIONAL.html).

fstein93:
@oschuett What do you have in mind? Do you mean the reference in the reference list or a shortcut for libxc as a section parameter?

oschuett:
I'd like to turn our current user interface:

```
&XC_FUNCTIONAL
  &LIBXC
    FUNCTIONAL XC_HYB_GGA_XC_B3LYP  ! name has to be looked up on libxc homepage
    PARAMETERS 123.4 42.0           ! just an anonymous bunch of numbers 
  &END LIBXC
&END XC_FUNCTIONAL
```

into this:
```
&XC_FUNCTIONAL
  &HYB_GGA_XC_B3LYP                 ! well documented within the cp2k manual
    ALPHA 123.4
    BETA 42.0
  &END HYB_GGA_XC_B3LYP
&END XC_FUNCTIONAL
```

Bascially, we'd create one section per functional and populate it with the [metadata from libxc](https://gitlab.com/libxc/libxc/-/blob/master/src/func_info.c).

tools/toolchain/scripts/install_libxc.sh
fstein93: No. First, I was using the github tarball. The other one seemed to be published this morning. It works locally, now with the other tarball.

fstein93:
To my understanding, you have to look up the available functionals in LibXC anyway because of their sheer amount anyway. Plus, you have to create the sections (with all parameters) by the preprocessor in case the user wants to compile the code without LibXC. I am not sure how to achieve that yet. And the manual page would become too long to read (ca. 400 entries).

EDIT: I have found the routines to create the sections.

src/xc_write_output.F
fstein93: I can only use a smaller value for the length as it was just a number which worked (20*default_string_length does not). Our reference of a LibXC functional may have up to 5 different references, a DOI, external parameters with description and scaling factor. So theoretically, the string can be very long.

oschuett:
> And the manual page would become too long to read (ca. 400 entries).

Sure, the page will be lengthy but so is the [libxc homepage](https://www.tddft.org/programs/libxc/functionals/). At least our list will be in a format familiar to our users.

> by the preprocessor in case the user wants to compile the code without LibXC.

While we [already](https://github.com/cp2k/cp2k/blob/master/src/xc/xc_libxc.F#L44) do that, there is actually a non-zero chance that we'll have to make libxc a required dependency when we start using its GPU acceleration.

> I have found the routines to create the sections.

Great! While the details might take a bit of tinkering, the basic approach should look like this:
```fortran
n = xc_number_of_functionals()
ALLOCATE(func_ids(n))
CALL xc_available_functional_numbers(func_ids)

DO i = 1, n
   func_name = xc_functional_get_name(func_ids(i))
   ! also get description, DOI, etc.
   CALL section_create(section, __LOCATION__, func_name, description, ...)
   ! add keywords for ext parameters
END DO
```

src/xc_write_output.F
oschuett: There should be a safeguard in [xc_libxc_wrap_info_refs](https://github.com/cp2k/cp2k/blob/master/src/xc/xc_libxc_wrap.F#L138) that prevents it from writting out of bounds. Then we can pass in a reasonably sized buffer, e.g. `10*default_string_length`. I think, 10 lines per functional ought to be enough.

src/xc/xc_libxc_wrap.F
fstein93: Done.

fstein93:
I think I have managed it to have separate sections for the LibXC functionals. Two questions to you:

1) In the old versions of LibXC, you were allowed to use two different keys to request a certain functional: KEY and XC_KEY . I can add the second version but it would just blow up the number of available sections.

2) LibXC is not consistent with the naming scheme of the parameters. Some functionals use an underscore, some not. I am using the names without any modifications. Do you have any concerns about it?

oschuett:
> I think I have managed it to have separate sections for the LibXC functionals.

Wohoo :-)

> In the old versions of LibXC, you were allowed to use two different keys to request a certain functional: KEY and XC_KEY .

Yes, let's just use "KEY" without the "XC_".

> LibXC is not consistent with the naming scheme of the parameters. Some functionals use an underscore, some not.

It says [here](https://gitlab.com/libxc/libxc/-/blob/master/src/xc.h#L176) that parameter starting with an underscore are internal. So, maybe we should simply not expose those?

Btw, can we keep the old [LIBXC](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/XC/XC_FUNCTIONAL/LIBXC.html) section for backwards compatibility and only mark its keywords [deprecated](https://github.com/cp2k/cp2k/blob/master/src/input/input_keyword_types.F#L192) for now?

fstein93:
> > LibXC is not consistent with the naming scheme of the parameters. Some functionals use an underscore, some not.
> 
> It says [here](https://gitlab.com/libxc/libxc/-/blob/master/src/xc.h#L176) that parameter starting with an underscore are internal. So, maybe we should simply not expose those?

Ah. I will expose them but add a hint a manual. For some functionals like LDA_X_ERF (long-ranged LDA exchange) it is useful.
 
> Btw, can we keep the old [LIBXC](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/XC/XC_FUNCTIONAL/LIBXC.html) section for backwards compatibility and only mark its keywords [deprecated](https://github.com/cp2k/cp2k/blob/master/src/input/input_keyword_types.F#L192) for now?

Ok. But I will adjust the regtests to the new format.

fstein93:
Unfortunately, there is a memory leak in one of the routines of the Fortran interface I need for the separate sections. Thus, I will not push this commit to the repository and add them in a later PR.

EDIT: Compare [https://gitlab.com/libxc/libxc/-/issues/287](url)

src/xc_write_output.F
fstein93: I have added a variable length reference character and a safe-guard.

tools/toolchain/scripts/stage3/install_libxc.sh
```diff
@@ -29,8 +29,8 @@ case "$with_libxc" in
             if [ -f libxc-${libxc_ver}.tar.gz ] ; then
                 echo "libxc-${libxc_ver}.tar.gz is found"
             else
-                download_pkg ${DOWNLOADER_FLAGS} ${libxc_sha256} \
-                             https://www.cp2k.org/static/downloads/libxc-${libxc_ver}.tar.gz
+                download_pkg -o libxc-${libxc_ver}.tar.gz ${DOWNLOADER_FLAGS} ${libxc_sha256} \
```
oschuett: Please continue to download from our own server: https://www.cp2k.org/static/downloads/libxc-5.1.0.tar.gz
... because any external dependency can bring down our CI.

fstein93: Thank you for arranging it.

ltalirz:
Thanks for adding libxc 5.1 support!

By the way, just out of curiosity: what is the policy concerning the next cp2k release, i.e. will this be officially released some time in December 2021 or can one hope for an earlier release?

dev-zero:
Given that we have a couple of issues with the current release (performance, python 3 compat, macOS compat, etc.) I guess we should try to get another release out sooner.

dev-zero:
@fstein93 when building `sdbg` without `__LIBXC` I now get:

```
/data/tiziano/cp2k.otherstuff/src/xc/xc_libxc.F:116:56:

    FUNCTION libxc_get_reference_length(libxc_params, lsd) RESULT(length)
                                                        1
Error: Unused dummy argument ‘lsd’ at (1) [-Werror=unused-dummy-argument]
f951: some warnings being treated as errors
```

dev-zero:
@fstein93 will fix in #1328

PR #915, opened 2020-05-13
Csr write

ducryf:
Fix ks_csr_write and s_csr_write not implemented for k-points

Compute the KS or overlap matrix at the k-points given in the &KPOINT section and print them to a .csr file analogous to the non-k-point case.
Also adds a call to ks_csr_write and s_csr_write in sq_scf_post_tb.F to allow printing the matrices from TB calculations.


ducryf:
The code works fine as it is. But there may be room for improvement. In particular the following three issues:

There is a lot of copying involved to combine the symmetric/antisymmetric real and imaginary matrices into a single non-symmetric complex matrix. Is there a more elegant way to achieve that?
CALL dbcsr_desymmetrize(rmatrix, rmatrix_nosym)
CALL dbcsr_desymmetrize(imatrix, imatrix_nosym)
CALL dbcsr_copy(cmatrix, rmatrix_nosym)
CALL dbcsr_copy(tmatrix, imatrix_nosym)
CALL dbcsr_add(cmatrix, tmatrix, cone, ione)

I have also noticed that S_CSR_WRITE and KS_CSR_WRITE do not appear in any regtest. Should I change this and if yes, can I add the print statement to an existing regtest or should I create a new one?

Lastly, I call these two functions from qs_scf_post_tb.F. This works fine but creates a dependency of qs_scf_post_tb.F on qs_scf_post_gpw.F. Is that acceptable or is there a better solution?

juerghutter:
Hi, you can avoid the dependencies by moving the routines into a new module. 
wfn_mix is an example where this has been done for qs_scf_post....

ducryf:
I have moved the two functions to a new module qs_scf_csr_write and added the PRINT%KS_CSR_WRITE and PRINT%S_CSR_WRITE sctions to three regtests, one each for QS, DFTB and xTB.

PR #932, opened 2020-05-28
Make OpenMP mandatory

### LINKED ISSUE #729: Drop support for non-OpenMP builds
We now got CPUs with [64 cores](https://www.amd.com/de/products/cpu/amd-ryzen-threadripper-3970x) per socket. Hence, running serial binaries will become very rare. On the other hand supporting non-OpenMP binaries has a real maintenance cost to it, in particular for the toolchain. 

Users who still want to run with a single thread, e.g. the high throughput folks,  can use `OMP_NUM_THREADS=1`. We know from our [scaling tests](https://dashboard.cp2k.org/archive/scaling/index.html) that the performance overhead for OpenMP is negligible.

oschuett:
Closes #729.


dev-zero:
Despite the popularity of `sopt` and `popt` I would rather drop them and keep the `ssmp` and `psmp` (or switch to completely new acronyms instead) rather than simply making the `sopt` and `popt` the OpenMP-threaded variants. The reason is that we don't manage the number of OpenMP threads explicitly and people need to be aware that they have to set `OMP_NUM_THREADS` now explicitly, otherwise they will likely end up with too many threads when running the new `popt` on a single machine.

Another option would be to have symlinks for `sopt` and `popt` and based on the exe-name-suffix set `OMP_NUM_THREADS` explicitly.

mkrack:
I am still using mostly popt binaries, since OpenMP does not provide a better efficiency for many cases, but the risk that all cores of a node are allocated inadvertently. I vote for keeping a pure MPI binary type.

oschuett:
> keep the `ssmp` and `psmp` ... otherwise they will likely end up with too many threads when running the new `popt` on a single machine.

I agree that this will likely confuse some users. However, the results will remain correct and the fix is trivial. So, I'd argue it's actually an efficient way to get the news out ;-)

> or switch to completely new acronyms instead

Yeah, maybe that's the way to go. Do you have any proposals? Ideally, we could also include Cuda / Hip.

> OpenMP does not provide a better efficiency for many cases

That is true and we urgently have to do something about it as we are heading for 100+ cores / socket.

> I vote for keeping a pure MPI binary type.

I think, a pure MPI binary adds little value for the users. Many libraries (pexsi, sirius, cosma) already  require OpenMP. On the other hand, if we keep supporting pure MPI binaries then we are not getting the benefits of a simplified toolchain and reduced testing.

Furthermore, I expect that our performance critical routines will eventually all be migrated to C. And in C it's really cumbersome to support non-OpenMP because one has to use `#ifdef _OPENMP`.

mkrack:
I have no problem with having OpenMP activated by default, if its greedy behavior could be controlled by CP2K internally without setting OMP_NUM_THREADS explicitly to 1 externally.

juerghutter:
I am strongly in favor to have sopt and popt executables available. These might be the corresponding ssmp and psmp executables with OMP_NUM_THREADS = 1 enforced.
I don't know how to achieve this seamless, but it should be done through the Makefile.

oschuett:
Alright, I'll look into preserving `sopt` and `popt` as aliases. The simplest solution is probably to generate shell script wrappers. Alternatively, we could use symlinks as @dev-zero  suggested.

oschuett:
I went with the symlink solution after all. If there are now more comments I'll merge this today.

mkrack:
@oschuett It seems that the symlinks for the unittests are missing. Alternatively, the do_regtest script has to be adapted, too.

PR #935, opened 2020-06-02
libcp2k: add set_cell to libcp2k

holly-t:
Allows changing of the cell dimensions through libcp2k.


oschuett:
The formatting check failed. You can run `make -j pretty` to fix it.

PR #937, opened 2020-06-04
toolchain: collection of small fixes

michaellass:
* openblas: Detect if there is a separate openblas_omp library instead of guessing. This change is for example required when linking against a system installation of openblas which provides only libopenblas.so.
* Remove checks for lib64 where lib64 is not strictly required. Many installation do not contain a separate lib64 folder but still provide everything required (observed for example with easybuild). Unless lib64 is really required, we should not check for its existence.

### DIFF
--- a/tools/toolchain/scripts/install_openblas.sh
+++ b/tools/toolchain/scripts/install_openblas.sh
@@ -83,11 +83,16 @@ case "$with_openblas" in
         OPENBLAS_CFLAGS="-I'${pkg_install_dir}/include'"
         OPENBLAS_LDFLAGS="-L'${pkg_install_dir}/lib' -Wl,-rpath='${pkg_install_dir}/lib'"
         OPENBLASROOT='${pkg_install_dir}'
+        OPENBLAS_LIBS="-lopenblas"
   ;;
     __SYSTEM__)
         echo "==================== Finding LAPACK from system paths ===================="
         # assume that system openblas is threaded
         check_lib -lopenblas "OpenBLAS"
+        OPENBLAS_LIBS="-lopenblas"
+        # detect separate omp builds
+        check_lib -lopenblas_openmp 2>/dev/null && OPENBLAS_LIBS="-lopenblas_openmp"
+        check_lib -lopenblas_omp 2>/dev/null && OPENBLAS_LIBS="-lopenblas_omp"
         add_include_from_paths OPENBLAS_CFLAGS "openblas_config.h" $INCLUDE_PATHS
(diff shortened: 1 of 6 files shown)


tools/toolchain/scripts/install_openblas.sh
```diff
@@ -120,6 +122,7 @@ case "$with_openblas" in
         echo "==================== Finding LAPACK from system paths ===================="
         # assume that system openblas is threaded
         check_lib -lopenblas "OpenBLAS"
+        check_lib -lopenblas_omp 2>/dev/null && OPENBLAS_LIBS_OMP="-lopenblas_omp"
```
dev-zero: while at it you may want to check for `-lopenblas_openmp` as well. On an openSUSE system:

```console
$ ls -1 /usr/lib64/libopenblas*
/usr/lib64/libopenblas_openmp.so
/usr/lib64/libopenblas_openmp.so.0
/usr/lib64/libopenblas_pthreads.so
/usr/lib64/libopenblas_pthreads.so.0
/usr/lib64/libopenblas.so
/usr/lib64/libopenblas.so.0
```

michaellass: Sounds good. I guess it mainly makes sense here and not in the "user paths" section?

oschuett: I don't think we need this in the "user path" section, because `-lopenblas_omp` is the default after all.

michaellass: Not if this change is applied. If `OPENBLAS_LIBS_OMP` is not set, `-lopenblas` would be used:
https://github.com/cp2k/cp2k/pull/937/files#diff-27faa3da481dac7fc5280ec3a0117d58R144

oschuett: Maybe we should just wait until I've merged #932 (hopefully sometime this week)?

michaellass: Ah, I did not notice that #932 affects this. Waiting for that makes sense. If I see correctly, that will make this change even more straightforward as we only need to deal with one linker flag.

oschuett: I've merged #932. As expected, this PR now has some conflicts.

michaellass:
I just force-pushed a new version of this PR. The current OpenBLAS install script in master only looks for openblas.so. This PR now adds back using openblas_omp.so and additionally openblas_openmp.so if any of those are available.

PR #938, opened 2020-06-04
submatrix method: exploit eigendecomposition for adjustment of chemical potential

michaellass:
When using the submatrix method and performing eigendecompositions of the submatrices to compute the sign function, we can use these decompositions to modify the chemical potential µ directly instead of doing repeated sign computations within a bisection loop.

This PR adds this functionality and provides two variants, selectable via SUBMATRIX_SIGN_METHOD:
* DIRECT_MUADJ: Keep eigenvalues and eigenvevectors of all submatrices to be able to find a valid value for µ and to directly derive a solution for the sign function.
* DIRECT_MUADJ_LOWMEM: Keep only eigenvalues and eigenvectors that are required to find a valid value for µ. This consumes severly less memory but requires a second round of eigendecompositions after µ has been determined to compute the sign function.


robertschade:
H Michaeli,
I have a few questions/concerns/suggestions:
- It feels very contradictory to have a new argument nelectron and new functionality that changes the chemical potential for density_matrix_sign_fixed_mu which is supposed to do only grand canonical calculations.
-  Could you please add an author, history, and description for the function trace_from_eigdecomp in src/iterate_matrix.F.
- Could you please add a history for matrix_sign_submatrix_mu_adjust.
- Could you please extend the history (!> \par History) and authors for example for density_matrix_sign_fixed_mu if new functionality was added to a function.
- Could we possibly get rid of the distinction for gfortran before 8.3 in matrix_sign_submatrix_mu_adjust? For example by using DEFAULT(SHARED) always?
- Could you remove the line "! molecular clustering is an interesting parameter" from the new regtests.

michaellass:
>     * It feels very contradictory to have a new argument nelectron and new functionality that changes the chemical potential for density_matrix_sign_fixed_mu which is supposed to do only grand canonical calculations.

Yes, modifying µ within a function suffixed with _fixed_mu seems contradictory. My intention here was to avoid code duplication by reusing what is already there. However, since the new methods only work in the symmetric case anyway, the duplication should not be too bad. I will have another look at that.

>     * Could we possibly get rid of the distinction for gfortran before 8.3 in matrix_sign_submatrix_mu_adjust? For example by using DEFAULT(SHARED) always?

While it is not pretty, I would rather leave it in for now. It gives the regtests a chance to check for missing data sharing clauses while working around the issue on GCC <= 8.3.

oschuett:
@michaellass, are you still planing to add changes or should this PR be merged as is?

michaellass:
Thanks for asking. A revised draft is nearly finished and I will update this PR soon.

michaellass:
I hope I have addressed all remaining points in the new version:

>     * It feels very contradictory to have a new argument nelectron and new functionality that changes the chemical potential for density_matrix_sign_fixed_mu which is supposed to do only grand canonical calculations.

`density_matrix_sign_fixed_mu` now stays unchanged and instead a similar routine `density_matrix_sign_internal_mu` is introduced for methods with internal µ adjustment.

>     * Could you please add an author, history, and description for the function trace_from_eigdecomp in src/iterate_matrix.F.

Done.

>     * Could you please add a history for matrix_sign_submatrix_mu_adjust.

Done.

>     * Could you please extend the history (!> \par History) and authors for example for density_matrix_sign_fixed_mu if new functionality was added to a function.

Skipped for `density_matrix_sign_fixed_mu` since it is now unchanged but done for others.

>     * Could you remove the line "! molecular clustering is an interesting parameter" from the new regtests.

Done.

PR #946, opened 2020-06-11
RPA: minimal output on progress of calculation

pseewald:


fstein93:
I like the idea. (We should add it to MP2 as well.)

fstein93 (approved):

PR #948, opened 2020-06-12
Fix extension for unittests

mkrack:


oschuett:
Yes, the symlink is only created for main cp2k binary. Since the Makefile is already very complex, I didn't see a way to do it generally for all  binaries.

As for the regtests, my plan was to just remove the `*opt` tests. So, I'm not sure we really need this change.

mkrack:
I would like to keep the intel popt regtest. Building and regtesting sopt and popt versions is then still possible, though the corresponding *smp versions are built as well.

oschuett:
> Building and regtesting sopt and popt versions is then still possible

I'm confused. It's no longer possible to build `*opt` binaries. What you could test instead are the `*smp` binaries with a single thread. In my opinion, the proper way of doing that would be:

```make test ARCH="Linux-x86-64-intel-regtest" VERSION="psmp" TESTOPTS="-numthreads 1"```

Is that what you had in mind?

Generally, while I agree that the symlinks are convenient, we shouldn't try to be full backwards compatible.

mkrack:
My intention is that
`make ARCH=target VERSION=psmp`
and
`make ARCH=target VERSION=popt`
generate the same: a target.psmp and the target.popt symlink.
How do you want to regtest the symlink functionality without a popt regtest?

oschuett:
> My intention is that `make ARCH=target VERSION=psmp` and `make ARCH=target VERSION=popt` generate the same: a target.psmp and the target.popt symlink.

Ok, I guess we could create the symlinks in both cases.
However, in order for the regtests to work we'd also have to create the symlinks for the unittests binaries. While this is possible, only the cp2k binary will actually call [omp_set_num_threads](https://github.com/cp2k/cp2k/blob/master/src/start/cp2k.F#L120). So, then we'd have to switch to a shell-script wrapper after all, which might have other issues.

So, I'm really not sure if this is all worth it. In my opinion people should get used to setting `OMP_NUM_THREADS`, e.g. in their bashrc. And the purpose of the aliases is only to smoothen this transition.

> How do you want to regtest the symlink functionality without a popt regtest?

I actually didn't think of testing the symlink itself. We could just use it in some [existing test](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_manual.sh#L11) to ensure it doesn't get accidentally removed.

Note that, we do regtest with a single thread via the `*dbg` runs.

mkrack:
> Ok, I guess we could create the symlinks in both cases.
> However, in order for the regtests to work we'd also have to create the symlinks for the unittests binaries. While this is possible, only the cp2k binary will actually call [omp_set_num_threads](https://github.com/cp2k/cp2k/blob/master/src/start/cp2k.F#L120). So, then we'd have to switch to a shell-script wrapper after all, which might have other issues.

That should work already with this commit. No need to create any further symlinks for regtesting popt binaries.

> So, I'm really not sure if this is all worth it. In my opinion people should get used to setting `OMP_NUM_THREADS`, e.g. in their bashrc. And the purpose of the aliases is only to smoothen this transition.

Here I am less optimistic how people behave. For instance, a careless launch of a cp2k.psmp binary on a Merlin login nodes, let's say just to check a CP2K input, creates 88 threads (the node has 44 cores with HT). As a result the CP2K run proceeds super slow and the node becomes more or less blocked due to over-commitment which affects also the interactive work of all other users logged in to that node.

> I actually didn't think of testing the symlink itself. We could just use it in some [existing test](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_manual.sh#L11) to ensure it doesn't get accidentally removed.
> Note that, we do regtest with a single thread via the `*dbg` runs.

Though that's true, I will try this commit and in case it turns out to be silly, I will revert it.

PR #951, opened 2020-06-18
Gal19

pclabaut:
Hello,
Here is the implementation of the novel forcefield GAL19, described by https://pubs.acs.org/doi/10.1021/acs.jctc.0c00091 to evaluate noble metal/water interactions.
The implementation is working and I hope that you will find it suitable for your code.
Best regards,

Clabaut Paul

[acs.jctc.0c00091(1).pdf](https://github.com/cp2k/cp2k/files/4797372/acs.jctc.0c00091.1.pdf)


pseewald (changes requested):
you accidentally changed the DBCSR submodule (`exts/dbcsr`). Can you revert this? Let me know if you need help doing this.

pclabaut:
Hello, thanks for the answer,
I figured It could be that but then I tried to download the ext/dbscr directory from the cp2k main page again but git doesn't detect any difference with the one one of my branch.

pseewald:
I added a commit to reset the submodule

pclabaut:
Oh, I see, thanks !
Did you detect any change then ?
I'm really surprised that git status coulnd't see any

pclabaut:
Apparently, it failed again in compiling the dbcsr part.
Any idea ?

pseewald:
> Oh, I see, thanks !
> Did you detect any change then ?
> I'm really surprised that git status coulnd't see any

You changed the commit of DBCSR submodule (see `git show 9687c635a2ee9094caf10d49e918a8b95fa026f1 exts`). This should be fixed now.

pclabaut:
But the testing fails again.
Your commit did apply to the gal19 branch, right?
So there is still a problem in the dbscr part of the branch after you reseted it ?

pseewald:
This is a wild guess: there seem to be a non-ascii character in this pull request leading to a failure in our `makedep.py` python script.

pclabaut:
It would be consistent with the error message. But how come it fails in this branch and not in the master of cp2k ?

pseewald:
> It would be consistent with the error message. But how come it fails in this branch and not in the master of cp2k ?

because you added the character (in CP2K, not DBCSR). I fixed it and hope that now tests run through.

pseewald (approved):

pclabaut:
Oh, I see, so the non-ascii character was in my files but detected by a python script from the ext/dbcsr part. 
Thank you a lot for spotting that!

pseewald:
> Oh, I see, so the non-ascii character was in my files but detected by a python script from the ext/dbcsr part.
> Thank you a lot for spotting that!

No this had nothing to do with DBCSR, the python script is in CP2K.

the tests report 2 leaks that should be fixed.

pclabaut:
Yes, I'm currently loocking into that. Thanks again!

pclabaut:
Hello again, 
The leak is fixed and the tests have passed. What is the next step?
Thanks again for the help !

PR #952, opened 2020-06-19
Update of sirius

mtaillefumier:
- update of sirius
- fix issue 317
- disable elpa
- bug fixes


oschuett:
All tests are passing :-)

I just noticed that [He-full-potential.inp](https://github.com/cp2k/cp2k/blob/master/tests/SIRIUS/regtest-1/He-full-potential.inp) takes 97 seconds for pdbg. Generally, we aim for individual regtests to finish within 30s. Is there any way to speed this one up? Doesn't have to be part of this PR.

mtaillefumier:
It is probably related to debugging nothing else since we check out of bounds when compiled in debug mode. removing the out of bound checks will help for sure.

PR #956, opened 2020-06-22
Add D3_EXCLUDE_KIND_PAIR keyword

ducryf:
The D3_EXCLUDE_KIND_PAIR keyword allows to exclude specific
interactions from the D3 dispersion correction.


juerghutter:
In a future refactoring of the code, one can move all the different exclusions into the generation of the pair list. This would be more efficient and leaves the actual pair potential code clean.

Issue #922, opened 2020-05-21
review assertions and avoid side effects


hfp:
Input files `benchmarks/QS_mp2_rpa/*/*.inp` stopped working and cause program abort with:

```text
 *******************************************************************************
 *   ___                                                                       *
 *  /   \                                                                      *
 * [ABORT]                                                                     *
 *  \___/    The requested basis set <RI_AUX> for element <O> was not found in *
 *    |                      the basis set files <./BASIS_H2O>                 *
 *  O/|                                                                        *
 * /| |                                                                        *
 * / \                                          aobasis/basis_set_types.F:1540 *
 *******************************************************************************


 ===== Routine Calling Stack =====

            3 read_qs_kind
            2 create_qs_kind_set
            1 CP2K
```

The complained basis set seems to be present in `BASIS_H2O` ("`O    RI-cc-QZ`").

hfp:
This is caused by a miscompiled binary. I was using GCC 8.3 and 9.3 with quite some optimization flags. I will check which flag(s) is(are) causing the issue.

hfp:
This is caused by `NDEBUG` applied to CP2K´s source code (which happened accidentally on my end). The problem with `NDEBUG` was discovered soon after its introduction but remained unspecific. However, this issue confirms/narrows a potential problem with eliding assertions (side effects). Side note: DBCSR´s source code is fine with eliding assertions (it is currently hard to apply `NDEBUG` separately to CP2K and DBCSR).

oschuett:
Yes, it's unfortunate that some of our assertions have side-effects. Since there is no automatic way to detect side-effects, should we maybe just execute `CPASSERT` regardless of `NDEBUG`?

oschuett:
It's certainly good to once review all assertions. However, unless we find an automatic way to detect side-effects, they will reappear. So, I think unconditional execution is our safest option.

>  Was the latter considered or is calling exit the way to go?

The `EXIT` statement is never actually executed, because we call [cp_abort_handler](https://github.com/cp2k/cp2k/blob/master/src/common/cp_error_handling.F#L54) before. The `EXIT` was merely meant as a compiler hint (not sure any compiler every picked up on it).

I'm not familiar with breakpoints. Would we still get the same error message and stack trace when no debugger is attached?

hfp:
[...]

PR #957, opened 2020-06-23
fixed side-effect about advancing an iterator

hfp:
Issue #922: fixed side-effect about advancing an iterator inside of an assertion (src/common/cp_linked_list.fypp). Prettified all fypp sources.


oschuett:
I have a hard time figuring out what actually changed.
Maybe you could split the prettifying and the actual changes into two separate commits?

hfp:
@oschuett Should be better now. Thanks for pointing out!

PR #961, opened 2020-06-25
Toolchain: Update GCC to version 9.3.0

oschuett:
This is needed for building the toolchain with Ubuntu 20.04.


oschuett:
I'm withdrawing this PR in favor of #962.

oschuett:
It seem upgrading to GCC 10.1 will be a longer undertaking.
Hence, I'm reopening this PR to help Ubuntu 20.04 users because GCC 9.2 doesn't work there.

PR #962, opened 2020-06-25
toolchain: update all packages

dev-zero:


dev-zero:
* [x] libint triggers a bug in gcc-10s cpp: https://github.com/evaleev/libint/issues/173 resp. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95889, patch for CPP added and sent to upstream
* [x] mpich does not build with gcc-10 yet: https://github.com/pmodels/mpich/issues/4300 ... workaround added (`-fallow-argument-mismatch`)
* [x] dbcsr is fixed for single-node in the develop branch
* [x] scalapack is fixed (except its tests) and PR accepted by upstream
* [x] plumed-2.6.1 triggers an ICE in 10.2.0, reported here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96354 ... workaround added (stripping `-g` which upstream does not recommend for production builds due to binary size)
* [x] QUIP (also latest master) is currently broken with gcc-10: https://github.com/libAtoms/QUIP/issues/220 ... workaround added (`-fallow-argument-mismatch`)

dev-zero:
@hfp I had to add `INTRINSICS=1` to make libxssm-1.16.1 build with gcc-10.2 (and binutils 2.32 from openSUSE Leap 15.1) to avoid `Error: no such instruction: `vcvtne2ps2bf16 %zmm0,%zmm12,%zmm15'`. Since the toolchain anyway builds for a specific target (`-march=native`) I guess there's no harm in doing so?

dev-zero:
@juerghutter can you please confirm that 56f52dd is indeed fixing a typo? It very much looks like one to me, but then I didn't dive into this part of the code too much (yet).

hfp:
> @hfp I had to add `INTRINSICS=1` to make libxssm-1.16.1 build with gcc-10.2 (and binutils 2.32 from openSUSE Leap 15.1) to avoid `Error: no such instruction: `vcvtne2ps2bf16 %zmm0,%zmm12,%zmm15'`. Since the toolchain anyway builds for a specific target (`-march=native`) I guess there's no harm in doing so?

This is normally bad if you generally update the recipe to build libxsmm like this (it should not harm too much w/ CP2K/DBCSR). However, the root cause is that Binutils are outdated on that platform while deploying a new GCC. Basically the compiler's capabilities are beyond the capabilities of Binutils. If you build binutils-2.34 (latest) from source and prepend `BINUTILSROOT/bin` to PATH environment variable you should be good.

For example, our environment module for Binutils looks like (you can of course do similar with `export`):

```text
#%Module1.0

module-whatis GNU Binutils 2.34

set             VERSION         "2.34"
set             ROOT            "/nfs/home/tools/binutils/$VERSION"

prepend-path    CPATH           "$ROOT/include"
prepend-path    LIBRARY_PATH    "$ROOT/lib"
prepend-path    MANPATH         "$ROOT/share/man"
prepend-path    PATH            "$ROOT/bin"
```

Let me check if I can find some better build-option. I will report back.

dev-zero:
@vondele I know it has been a long time since you wrote that code, but I am a bit unsure about cf26a29. I figure I would have to adjust the `inca` parameter to `1` as well if I'm going to switch to `zcopy` instead?

dev-zero:
> > @hfp I had to add `INTRINSICS=1` to make libxssm-1.16.1 build with gcc-10.2 (and binutils 2.32 from openSUSE Leap 15.1) to avoid `Error: no such instruction: `vcvtne2ps2bf16 %zmm0,%zmm12,%zmm15'`. Since the toolchain anyway builds for a specific target (`-march=native`) I guess there's no harm in doing so?
> 
> This is normally bad if you generally update the recipe to build libxsmm like this (it should not harm too much w/ CP2K/DBCSR). However, the root cause is that Binutils are outdated on that platform while deploying a new GCC. Basically the compiler's capabilities are beyond the capabilities of Binutils. If you build binutils-2.34 (latest) from source and prepend `BINUTILSROOT/bin` to PATH environment variable you should be good.
[...]
> Let me check if I can find some better build-option. I will report back.

Ok, thanks. I guess except for Arch, Fedora and macOS most distros will feature an outdated binutils for some time still and I would really like to avoid carrying a binutils script as well.

hfp:
> Let me check if I can find some better build-option. I will report back.

I think `INTRINSICS=1` is the best general option for CP2K/DBCSR install script and should not harm performance since only block-ops somewhat rely on dynamic target using Intrinsic based code path (non-JIT). I will likely improve this in our build system like "INTRINSICS=auto" by default.

juerghutter:
Yes, looks fine to me.
Universität Zürich                   E-mail: hutter@chem.uzh.ch
Winterthurerstrasse 190
CH-8057 Zürich, Switzerland

-----"Tiziano Müller" <notifications@github.com> wrote: -----
To: "cp2k/cp2k" <cp2k@noreply.github.com>

vondele:
> @vondele I know it has been a long time since you wrote that code, but I am a bit unsure about [cf26a29](https://github.com/cp2k/cp2k/commit/cf26a29c7a59cb2ca0b0305304559520c5e61d17). I figure I would have to adjust the `inca` parameter to `1` as well if I'm going to switch to `zcopy` instead?

I think that's different.... trying to recall, I think the original code will just copy the real part of the complex numbers to buffer of reals, I don't think you can achieve that with zcopy. Probably best to write the Fortran equivalent.

oschuett:
> the root cause is that Binutils are outdated on that platform while deploying a new GCC.

I was planing to update our [Docker base image](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/Dockerfile#L1) to `ubuntu:20.04` soon. Nevertheless, the toolchain should also work in older environments.

> I would really like to avoid carrying a binutils script as well.

 The binutils were actually once part of the toolchain. Maybe we should just install them during `install_gcc.sh`?

dev-zero:
> I think that's different.... trying to recall, I think the original code will just copy the real part of the complex numbers to buffer of reals, I don't think you can achieve that with zcopy. Probably best to write the Fortran equivalent.

You're right, thanks! I went with an explicit `TRANSFER` as a replacement instead (which is what that was implicitly), see 1047d2f59e9f5f77e3f2e358e5f73c3401ee8d5e

oschuett:
> I went with an explicit `TRANSFER` as a replacement instead (which is what that was implicitly), see 1047d2f

Couldn't you just use `sendbuf = REAL(pw%cc3d, dp)` ?

dev-zero:
> Couldn't you just use `sendbuf = REAL(pw%cc3d, dp)` ?

Rank-mismatch, would have required a `RESHAPE` as well.

PR #965, opened 2020-06-26
Fixe a problem for gal19 and the test

pclabaut:
Hello, I fixed a bug in the code that was preventing from actually working. I also adapted the test for the new result.
Sorry for the inconvenience...

### DIFF
--- a/src/manybody_gal.F
+++ b/src/manybody_gal.F
@@ -14,7 +14,8 @@ MODULE manybody_gal
    USE cell_types,                      ONLY: cell_type,&
                                               pbc
    USE cp_log_handling,                 ONLY: cp_get_default_logger,&
-                                              cp_logger_type
+                                              cp_logger_type,&
+                                              cp_to_string
    USE cp_output_handling,              ONLY: cp_print_key_finished_output,&
                                               cp_print_key_unit_nr
    USE cp_para_types,                   ONLY: cp_para_env_type
@@ -52,10 +53,11 @@ CONTAINS
 !> \param jparticle second index of the atom of the evaluated pair
 !> \param cell dimension of the pbc cell
 !> \param particle_set full list of atoms of the system
+!> \param mm_section ...
 !> \author Clabaut Paul - 2019 - ENS de Lyon
 ! **************************************************************************************************
    SUBROUTINE gal_energy(pot_loc, gal, r_last_update_pbc, iparticle, jparticle, &
-                         cell, particle_set)
+                         cell, particle_set, mm_section)
 
       REAL(KIND=dp), INTENT(OUT)                         :: pot_loc
       TYPE(gal_pot_type), POINTER                        :: gal
@@ -63,14 +65,17 @@ CONTAINS
       INTEGER, INTENT(IN)                                :: iparticle, jparticle
       TYPE(cell_type), POINTER                           :: cell
       TYPE(particle_type), DIMENSION(:), POINTER         :: particle_set
+      TYPE(section_vals_type), POINTER                   :: mm_section
 
       CHARACTER(LEN=*), PARAMETER :: routineN = 'gal_energy', routineP = moduleN//':'//routineN
 
       CHARACTER(LEN=2)                                   :: element_symbol
+      INTEGER                                            :: index_outfile
       REAL(KIND=dp)                                      :: anglepart, cosalpha, drji2, gcn_weight, &
                                                             gcn_weight2, nvec(3), rji(3), &
                                                             sinalpha, sum_weight, Vang, Vgaussian, &
                                                             VTT, weight
+      TYPE(cp_logger_type), POINTER                      :: logger
 
       pot_loc = 0.0_dp
       CALL get_atomic_kind(atomic_kind=particle_set(iparticle)%atomic_kind, &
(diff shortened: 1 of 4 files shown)


oschuett:
While you're at it, could you also please replace the writes to stdout with a [print key](https://www.cp2k.org/dev:printkey). This will make our [convenction checker](https://dashboard.cp2k.org/archive/conventions/commit_b3d5a8484a3591abac4ebe4533d4b337ccefeee7.txt) happy.

pclabaut:
Sure!
I will take a look

pclabaut:
Well, I took a look, tried, but because I am not entirely getting the way this printkey functions, I keep getting weird errors, even on part of the code I did not touched (ex:
/home/pclabaut/Softs/cp2k_compiled/src/hfx_pw_methods.F:244:54:
                                               greenfn=greenfn)
Error: Keyword argument ‘greenfn’ at (1) is not in the procedure)

oschuett:
Hmm, sounds strange. You could try to run `make distclean` to clean the build cache.
Generally, you can just use the print key like you already did in [print_nr_ions_gal](https://github.com/cp2k/cp2k/blob/master/src/manybody_gal.F#L773).

src/manybody_gal.F
```diff
@@ -313,16 +323,16 @@ FUNCTION angular(gal, r_last_update_pbc, iparticle, cell, particle_set, nvec, en
...
-         WRITE (*, *) "Faulty atom :", iparticle
+         !WRITE (*, *) "Faulty atom :", iparticle
          CPABORT("Only one H atom of O atom found")
       ELSEIF (count_h == 3) THEN
-         WRITE (*, *) "Faulty atom :", iparticle
+         !WRITE (*, *) "Faulty atom :", iparticle
          CPABORT("Three H atoms for O atom found")
       ELSEIF (count_h > 3) THEN
-         WRITE (*, *) "Faulty atom :", iparticle
+         !WRITE (*, *) "Faulty atom :", iparticle
```
oschuett: You can still include the atom number in the error message like this:

```
USE cp_log_handling, ONLY: cp_to_string
...
IF (count_h /= 2) THEN
   CALL cp_abort(__LOCATION__, &
                 "Found "//cp_to_string(count_h)//" H atoms for O atom "//cp_to_string(iparticle))
ENDIF
```

pclabaut:
Thanks ! 
I've been strugling for this but didn't found the right way to do it.

src/manybody_gal.F
```diff
@@ -15,6 +15,7 @@ MODULE manybody_gal
                                               pbc
    USE cp_log_handling,                 ONLY: cp_get_default_logger,&
                                               cp_logger_type
```
oschuett: Missing: `,&`

src/manybody_gal.F
```diff
@@ -312,18 +323,9 @@ FUNCTION angular(gal, r_last_update_pbc, iparticle, cell, particle_set, nvec, en
...
-      ELSEIF (count_h == 1) THEN
-         WRITE (*, *) "Faulty atom :", iparticle
-         CPABORT("Only one H atom of O atom found")
-      ELSEIF (count_h == 3) THEN
-         WRITE (*, *) "Faulty atom :", iparticle
-         CPABORT("Three H atoms for O atom found")
-      ELSEIF (count_h > 3) THEN
-         WRITE (*, *) "Faulty atom :", iparticle
-         CPABORT("Error in gal part: too many H atoms for O")
+      IF ((count_h /= 2)) THEN
```
oschuett: Double brackets not needed.

src/manybody_gal.F
```diff
@@ -312,18 +323,9 @@ FUNCTION angular(gal, r_last_update_pbc, iparticle, cell, particle_set, nvec, en
...
-         CPABORT("Three H atoms for O atom found")
-      ELSEIF (count_h > 3) THEN
-         WRITE (*, *) "Faulty atom :", iparticle
-         CPABORT("Error in gal part: too many H atoms for O")
+      IF (count_h /= 2) THEN
+         CALL cp_abort(__LOCATION__, &
+                 " Error: Found "//cp_to_string(count_h)//" H atoms for O atom "//cp_to_string(iparticle))
       ENDIF
 
       IF (count_h == 2) THEN
```
oschuett: One more nit: This could be an `ELSE` branch.

```
IF (count_h /= 2) THEN
   CALL cp_abort(__LOCATION__, ...
ELSE 
   rih1(:) = ...
   ...
END IF
```

The same applies to `angular_d` below.

pclabaut: Yes, I corrected both.
Thanks, I went to quick

pclabaut:
Well, sorry to bother again but I could use some help here.
I'm having a function that is called from two different parts of the code (A and B).
- Called from A, it should print something, and in A, the right printing section (mm_section) is known and can be pass to the function
- Called from B, it should not print anything and in B, there is no printing section known.
That shouldn't be a problem but still, the function need its section argument to be passed.
I tried to pass "None" has section argument when the function is called from B since it is not used anyway but I got a "Symbol 'none' at (1) has no IMPLICIT type" error.
Any idea about how I could operate without rewriting the whole function almost similarly but without the printing line, just for B?
Thanks in advance !

mkrack:
You have to pass a (nullified) pointer variable of proper type instead of the undefined variable NONE. You can also make that pointer argument OPTIONAL and use it only if it is PRESENT.

pclabaut:
Thanks a lot! 
It worked fine!

src/manybody_gal.F
```diff
@@ -339,8 +353,15 @@ FUNCTION angular(gal, r_last_update_pbc, iparticle, cell, particle_set, nvec, en
 
          ! For fit purpose
          IF (gal%express .AND. energy) THEN
-            WRITE (*, *) "Fourier", costheta, COS(2.0_dp*theta), COS(3.0_dp*theta), &
+            logger => cp_get_default_logger()
+            index_outfile = cp_print_key_unit_nr(logger, mm_section, &
```
oschuett: Passing a null pointer for the section to `cp_print_key_unit_nr` will not work because of [this assert](https://github.com/cp2k/cp2k/blob/master/src/input/cp_output_handling.F#L902). This is currently not caught by the regtests, because `gal%express` is always false. A simple solution would be to pass in the output unit `index_outfile` instead of the section. Furthermore, it would be good to have some tests covering those branches.

src/manybody_gal.F
```diff
@@ -312,18 +335,9 @@ FUNCTION angular(gal, r_last_update_pbc, iparticle, cell, particle_set, nvec, en
...
-         CPABORT("Three H atoms for O atom found")
-      ELSEIF (count_h > 3) THEN
-         WRITE (*, *) "Faulty atom :", iparticle
-         CPABORT("Error in gal part: too many H atoms for O")
+      IF (count_h /= 2) THEN
+         CALL cp_abort(__LOCATION__, &
+                       " Error: Found "//cp_to_string(count_h)//" H atoms for O atom "//cp_to_string(iparticle))
       ENDIF
 
       IF (count_h == 2) THEN
```
oschuett: This `IF` statement is not needed because you are already aborting in case `count_h /= 2`. The same goes for the occurrence below.

oschuett (changes requested):

src/manybody_gal.F
pclabaut: Here, I do not think it is a problem because "energy" is never TRUE while mm_section is nullified (that is actually the point of this "energy" boolean: to avoid the printing when this function is called from the force evaluation routine instead of the energy evaluation routine). So when the function is called from the energy routine, energy is TRUE and mm_section is a real section, while when it is called from the force evaluation, energy is FALSE and mm_section is nullified but cp_print_key_unit_nr is never used. I checked on my computer and it works indeed fine even with gal%express being TRUE.
About this testing, this option is only intended to print some quantities needed for a linear fit of the forcefield so I didn't consider it was worth regtesting because it would only be useful for a person that would want to redevelop the forcefield.

src/manybody_gal.F
pclabaut: Indeed, I will change that.

src/manybody_gal.F
oschuett: Ok

src/manybody_gal.F
oschuett: Thanks.

PR #966, opened 2020-06-26
Fix CELL_OPT ignoring CONSTRAINT for convergence criterion

ducryf:
The internal pressure is calculated including dimensions fixed
with the CELL_OPT%CONSTRAINT keyword. This can result in
CELL_OPT simulations that never converge because the 
direction with remaining stress is not allowed to change. With
this patch only directions that are not constrained contribute
to the internal pressure when checking for convergence.

### DIFF
--- a/src/motion/cell_opt_types.F
+++ b/src/motion/cell_opt_types.F
@@ -60,7 +60,7 @@ MODULE cell_opt_types
       INTEGER                                    :: constraint_id, id_nr, ref_count
       LOGICAL                                    :: keep_angles, &
                                                     keep_symmetry
-      REAL(KIND=dp)                              :: pres_ext, pres_int, pres_tol
+      REAL(KIND=dp)                              :: pres_ext, pres_int, pres_tol, pres_constr
       REAL(KIND=dp), DIMENSION(3, 3)             :: mtrx
       REAL(KIND=dp), DIMENSION(3, 3)             :: rot_matrix
       TYPE(cell_type), POINTER                   :: ref_cell
--- a/src/motion/cell_opt_utils.F
+++ b/src/motion/cell_opt_utils.F
@@ -255,11 +255,12 @@ CONTAINS
 !> \param keep_angles ...
 !> \param keep_symmetry ...
 !> \param pres_int ...
+!> \param pres_constr ...
 !> \param constraint_id ...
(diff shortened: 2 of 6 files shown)


mkrack:
@ducryf  A warning or at least a note should be printed in that case, since the printed internal stress is no longer the true one.
I think, it would be better to consider the constraint in the calculation of pres_diff in gopt_f_io for the convergence check and to leave the internal pressure tensor untouched.

ducryf:
@mkrack From what I can tell cell_env%pres_int is only used for printing and checking convergece. That's why I though it is the proper variable to change.
In my opinion the same value that is checked for convergence should be printed in the output. Otherwise there could be a case where a large pressure is reported while also stating that it is converged, e.g.
  Pressure Deviation [bar]   =      1094.4742388630
  Pressure Tolerance [bar]   =        100.0000000000
  Conv. for  PRESSURE        =                             YES
Printing a notice is a good idea. Probably in cell_opt_env_create where "CELL_OPT| Constraint: xy" is printed.

Let me know what you think, then I can change the code accordingly.

mkrack:
The printed "Pressure Deviation" should, of course, match the pressure after applying the constraint to match the convergence criterion, but this could be done also via a local variable in gopt_f_io without changing gopt_env%cell_env%pres_int. On the other hand, you are right, gopt_env%cell_env%pres_int is not used for anything else yet. You could also print the "Pressure Deviation" with and without applying the constraint after each optimisation step if a constraint is active.
```
Pressure Deviation [bar] = 1000.00000000 (without constraint)
Pressure Deviation [bar] =   20.00000000 (with constraint)
Pressure Tolerance [bar] =  100.00000000
``

ducryf:
Thanks for the suggestion! Instead of computing the internal pressure again in gopt_f_io I added a the pressure after applying the constraint to the cell_opt_env_type.
The output is now
  Pressure Deviation [bar]   =     -5782.6407057199
  (without constraint)
  Pressure Deviation [bar]   =        -0.3333246503
  (with constraint)
  Pressure Tolerance [bar]   =      10.0000000000
  Conv. for  PRESSURE        =                  YES

I have also changed the pressure tolerance of the two regtests with the CELL_OPT%CONSTRAINT keyword. They converge without an extremely large tolerance value.

PR #968, opened 2020-06-27
Added mp_sync before and after local data operations, avoid default printing of EXX/RPA in GW

JWilhelm:
- added mp_sync before and after local data operations for proper matrix operations; ensures correct RPA/GW results for large calculations
- avoid default printing of EXX/RPA energy in GW: when doing RI for the exchange self-energy, no EXX energy is computed and total energy is meaningless. EXX/RPA energy printing can be enabled by flag UPDATE_XC_ENERGY in GW


fstein93:
Interestingly, the commits have been added to the repository but due to some issue with Github (?), the PR hasn't been closed. Does anybody know about it?  @dev-zero @oschuett

oschuett:
Yeah, Github has some issues (again) https://news.ycombinator.com/item?id=23675864

PR #975, opened 2020-06-30
Add parameter routineP when routineN is defined

mkrack:


oschuett:
Thanks for taking care of those.

PR #987, opened 2020-07-10
Mixed bag: Threshold adjustment, Docker cleanup, QM/MM error message

oschuett:


pseewald:
@oschuett my cp2k fails with a segfault due to the change in f048338a4f395105ed73d1f245e5f5cb8ae21469 (checked with valgrind). Regarding testing with valgrind I opened a separate issue #994.

Valgrind output:
```
==31000== Invalid read of size 8
==31000==    at 0xD3CAA5: __qs_environment_MOD_qs_init (qs_environment.F:384)
==31000==    by 0xCC9C54: __f77_interface_MOD_create_force_env (f77_interface.F:786)
==31000==    by 0x462083: __cp2k_runs_MOD_cp2k_run (cp2k_runs.F:282)
==31000==    by 0x45D4DE: __cp2k_runs_MOD_run_input (cp2k_runs.F:970)
==31000==    by 0x44E3DA: MAIN__ (cp2k.F:335)
==31000==    by 0x44E51C: main (cp2k.F:42)
==31000==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
```

oschuett:
Uhh, thanks. I'll have a look.

oschuett:
Do you have an input file or regtest to reproduce this?

pseewald:
I think any QS regtest will reproduce this, if not let me know

edit: I just checked again, for instance `tests/QS/regtest-gpw-1/H2.inp` reproduces this.

oschuett:
Strange, I can not reproduce this. After switching to `-O2` the `ssmp` binary from the latest toolchain passes with zero errors.

pseewald:
I'll have a look then

pseewald:
@oschuett the issue is triggered only with `-O0` (what I usually use for debugging). With optimization enabled, probably only the first condition in the `if` statement is executed and the rest is optimized away. Maybe we should also test with `-O0` then? I'll change issue #994 to suggest this.

oschuett:
Oh wow. I really thought Fortran does short circuit logic, but apparently the standard says [otherwise](https://www.scivision.dev/fortran-short-circuit-logic/). Thanks for catching this!

PR #988, opened 2020-07-12
Fix dihedral double counting in Amber topologies

### LINKED ISSUE #984: Incorrect parsing of Amber prmtop files
Classical molecular mechanics module gives incorrect energy and forces when using Amber `prmtop` files.

To investigate the origin the problem, I run several simulations of the GTP molecule with Amber force field in various MD software ([cp2k-amber.zip](https://github.com/cp2k/cp2k/files/4893050/cp2k-amber.zip)). The same `prmtop` file was used in Amber and CP2K.

This analysis shows, that the dihedral energy in CP2K is exactly two times smaller than needed. Interestingly, converting topology and force field parameters to CHARMM format results in correct dihedral energy in CP2K.

| Program                        | Dihedral energy        |
|---------------------------|-----------------------|
| CP2K (Amber prmtop)   |   7.866500 kcal/mol |
| CP2K (PSF)                     | 15.727700 kcal/mol |
| Amber 18                       | 15.733000 kcal/mol |
| Gromacs  2018               | 15.791368 kcal/mol |

Source code inspection shows that force constants of all torsions are halved when parsing `prmtop` files:
https://github.com/cp2k/cp2k/blob/51267ca7aef7d7e6e416ad6faf8961c06b61ae0e/src/topology_amber.F#L1616

Removing the factor 0.5 fixed the problem for me.

The issue was observed with CP2K versions 7.1 and 6.1, and it most probably exists in the development version.

vamironov:
- Fix double counting of two definitions A-B-C-D and D-C-B-A of
  same dihedral by re-ordering atom names
- Remove 0.5 factor from dihedral force constants

Possibly fixes #984

### DIFF
--- a/src/topology_amber.F
+++ b/src/topology_amber.F
@@ -1588,15 +1588,15 @@ CONTAINS
                                             id3=name_atm_c, id4=name_atm_d)
             ! Possibly reorder names to avoid double counting of the dihedral
             IF (name_atm_a < name_atm_d) THEN
-                work_label(1, i) = name_atm_a
-                work_label(2, i) = name_atm_b
-                work_label(3, i) = name_atm_c
-                work_label(4, i) = name_atm_d
+               work_label(1, i) = name_atm_a
+               work_label(2, i) = name_atm_b
+               work_label(3, i) = name_atm_c
+               work_label(4, i) = name_atm_d
             ELSE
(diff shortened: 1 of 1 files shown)


vamironov:
This PR does not fix all problems with dihedral parsing so I close it

PR #991, opened 2020-07-15
Use raw Amber parameters for dihedrals

### LINKED ISSUE #984: Incorrect parsing of Amber prmtop files

vamironov:
Previously, Amber dihedral parameters were assigned in two steps:
1. Find all unique dihedral parameters in the topology. Uniqueness is
defined by combination of kinds of involved atoms
2. Walk along the full dihedral list and for every dihedral pick matching
set of parameters based on the kinds of corresponding atoms

This patch change this behavior to use dihedral parameter assignment
found in Amber parameter+topology file. It uses absolute atomic
indices instead of kinds and does not lead to ambiguity.

Possibly fix #984 

This is re-created PR #990 due to Github errors

### DIFF
--- a/src/force_field_types.F
+++ b/src/force_field_types.F
@@ -151,6 +151,10 @@ MODULE force_field_types
       CHARACTER(LEN=default_string_length), POINTER :: nonbond_a(:)
       REAL(KIND=dp), POINTER                        :: nonbond_eps(:)
       REAL(KIND=dp), POINTER                        :: nonbond_rmin2(:)
+      INTEGER, POINTER                               :: raw_torsion_id(:, :)
+      REAL(KIND=dp), POINTER                         :: raw_torsion_k(:)
+      REAL(KIND=dp), POINTER                         :: raw_torsion_m(:)
+      REAL(KIND=dp), POINTER                         :: raw_torsion_phi0(:)
    END TYPE amber_info_type
 
 ! **************************************************************************************************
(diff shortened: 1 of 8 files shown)


oschuett:
Hi @vamironov, are you still working on this?

vamironov:
> 
> 
> Hi @vamironov, are you still working on this?

This patch fixed the issue to me - now dihedral energies are consistent with what I observe in Amber/Gromacs. The optimizations also converge to nearly same structures for all these programs. If you need tests, I can look for them.

oschuett:
So, if you're confident that the code now behaves correctly, could you please update the reference values of the [four failing tests](https://storage.googleapis.com/cp2k-ci/run-cp2k-sdbg-642b6c99_report.txt).

vamironov:
Done

oschuett:
Great, thanks for contributing!

PR #999, opened 2020-07-22
Enhance BSSE Documentation

patrickmelix:
The input description of the BSSE section was not giving the information needed to understand it. Found the information in the forum and put it here for reference.

Tried to follow the guidelines, but since this is the first contribution to cp2k please let me know if I missed something.


oschuett:
Thanks for contributing!

PR #1000, opened 2020-07-23
NDEBUG: grid/collocate sample/driver code

hfp:
Introduced GRID_COLLOCATE_ASSERT to substitute "assert" in order to avoid eliding side-effects (NDEBUG).

### DIFF
--- a/src/grid/grid_collocate_replay.h
+++ b/src/grid/grid_collocate_replay.h
@@ -8,6 +8,12 @@
 
 #include <stdbool.h>
 
+#if defined(NDEBUG)
+# define GRID_COLLOCATE_ASSERT(EXPR) (void)(EXPR)
+#else
+# define GRID_COLLOCATE_ASSERT(EXPR) assert(EXPR)
+#endif
+
 void grid_collocate_record(const bool orthorhombic,
                            const bool use_subpatch,
                            const int subpatch,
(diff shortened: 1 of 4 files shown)


src/grid/grid_collocate_replay.h
```diff
@@ -8,6 +8,12 @@
 
 #include <stdbool.h>
 
+#if defined(NDEBUG)
+# define GRID_COLLOCATE_ASSERT(EXPR) (void)(EXPR)
+#else
+# define GRID_COLLOCATE_ASSERT(EXPR) assert(EXPR)
+#endif
```
oschuett, line 15: I'm confused. Isn't this already the [standard behavior](https://en.cppreference.com/w/c/error/assert) of `assert` in C?

oschuett: Ah, now I realize the difference. Your definition actual executes the expression regardless.

hfp: No, `assert` completely elides the given expression when `NDEBUG` is defined.

hfp: ( This PR is somewhat similar to issue #922 )

oschuett:
I now realize that my use of `assert` is indeed problematic. Thanks for the hint.
However, I don't like to introduce a macro either. I'll refactor my code to ensure it does not rely on assert side-effects.

src/grid/grid_collocate_replay.h
hfp: > 
> 
> Ah, now I realize the difference. Your definition actual executes the expression regardless.

Yes, the `(void)` cast in front is just to suppress warnings about unused result (comparison expression).

PR #1009, opened 2020-07-31
toolchain: pass -Wno-error=uninitialized to eri_mme_lattice_summation…

dev-zero:
fixes build errors with toolchain-generated arch-files and gcc-10

### DIFF
--- a/tools/toolchain/scripts/generate_arch_files.sh
+++ b/tools/toolchain/scripts/generate_arch_files.sh
@@ -182,6 +182,15 @@ EOF
     eval "printf \"${__TMPL}\n\"" > $__filename
     # pass this to parsers to replace all of the IF_XYZ statements
     "${SCRIPTDIR}/parse_if.py" -i -f "${__filename}" $__flags
+
+    if [ "$with_gcc" != "__DONTUSE__" ] ; then
+        cat << 'EOF' >> $__filename
+# Until https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96312 is fixed
+eri_mme_lattice_summation.o: eri_mme_lattice_summation.F
+	$(TOOLSRC)/build_utils/fypp $(FYPPFLAGS) $< $*.F90
+	$(FC) -c $(FCFLAGS) -Wno-error=uninitialized -D__SHORT_FILE__="\"$(subst $(SRCDIR)/,,$<)\"" -I'$(dir $<)' $(OBJEXTSINCL) $*.F90 $(FCLOGPIPE)
+EOF
+    fi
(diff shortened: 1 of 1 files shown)


oschuett:
Wouldn't it be better to change [eri_mme_lattice_summation.F](https://github.com/cp2k/cp2k/blob/master/src/eri_mme/eri_mme_lattice_summation.F) instead? Some people actually write their own ARCH files.

dev-zero:
> Wouldn't it be better to change [eri_mme_lattice_summation.F](https://github.com/cp2k/cp2k/blob/master/src/eri_mme/eri_mme_lattice_summation.F) instead?

The code is correct and changing it is not as straight forward as it seems.

oschuett:
> The code is correct and changing it is not as straight forward as it seems.

We work around compiler bugs all the time. Removing a few `INTENT(OUT)` might be enough. Otherwise we can simply switch to `POINTER`. Who knows when this GCC bug will be fixed.

pseewald:
> Wouldn't it be better to change [eri_mme_lattice_summation.F](https://github.com/cp2k/cp2k/blob/master/src/eri_mme/eri_mme_lattice_summation.F) instead? Some people actually write their own ARCH files.

People writing their own ARCH file would not expect that CP2K (or any other code) compiles with `-Werror` flags, so they'd use `-Werror=uninitialized` at their own risk. IMO we should workaround compiler bugs only when there is no other way to get CP2K compiled.

mkrack:
Not sure, if that is related to this issue, but sdbg and pdbg CP2K binaries cannot be built based on the current toolchain (using gcc 8.3 and mpich 3.3), if the flag Werror=uninitialized is present:
```
/data/user/krack/github/mkrack/cp2k/src/hfx_ri.F:353:0:

       TYPE(dbcsr_t_type), DIMENSION(1, 1)                :: t_3c_int_batched

Error: ‘MEM[(struct dbcsr_t_type *)&dbcsr_t_type.484 + 72B]’ is used uninitialized in this function [-Werror=uninitialized]
/data/user/krack/github/mkrack/cp2k/src/hfx_ri.F:353:0: Error: ‘MEM[(struct dbcsr_t_type *)&dbcsr_t_type.484 + 80B]’ is used uninitialized in this function [-Werror=uninitialized]
...
```
I think that it would be more user-friendly to change "uninitialized" from error to warning, i.e. to -Wuninitialized

oschuett:
I agree. We are getting too many false alerts with `-Werror=uninitialized`. And disabling it one file at a time creates toil.
Hopefully, in the not-too-distant future we'll have Valgrind again to detect uninitialized variables.

mkrack:
Resolved via #1017

PR #1011, opened 2020-08-01
Fix ASE and i-Pi, add GCC-9 test, Upgrade to Ubuntu 20.04

oschuett:


oschuett:
I'm not upgrading [Dockerfile.cuda_mkl](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/Dockerfile.cuda_mkl) for the time being because of https://github.com/electronic-structure/SIRIUS/issues/513.

PR #1012, opened 2020-08-02
Toolchain: Upgrade to SIRIUS 6.5.5 and SpFFT 0.9.12

oschuett:

PR #1013, opened 2020-08-02
Make cp2k + sirius cuda 11-compatible

haampie:
This is required for cuda 11 + cusolver support, but might still not be sufficient.

There seems to be an issue where SIRIUS is creating too many cuda streams. Locally I can make the tests pass with `TESTOPTS="-maxtasks 8"`

Ping @oschuett, you might want to give this a spin


haampie:
Oh, I really had not seen #1012 before creating the PR. This can be closed.

PR #1018, opened 2020-08-06
Fix ADMM in combination with GW

OndrejMarsalek:
We came across a segfault in a calculation that combines ADMM for the SCF (but not for HFX in RPA) and GW.

This led me to a pointer that was not associated with this combination of settings. This PR fixes the issue - the pointer `admm_env` is now assigned even in this case. I also added an assert and a regtest (based on `G0W0_H2O_PBE0.inp` with minimal changes) in case things change in the future.


oschuett:
Thanks for the fix :-)

PR #996, opened 2020-07-21
Allow non-ASCII characters

edditler:

PR #1024, opened 2020-08-10
Feature/docker enable unicode in Ubuntu

dev-zero:
... includes commits from #996 for testing


mkrack:
@dev-zero The [regtester](https://www.cp2k.org/static/regtest/trunk/Linux-x86-64-gfortran-regtest/psmp/regtest-0) is failing, since setlocale does not work (is not allowed). How to change the locale without being admin?

PR #1032, opened 2020-08-14
Drop -pedantic flag

mkrack:
- Drop -pedantic flag
- Remove redefinition of LC_ALL in the arch file
- Split error output
- Remove unused routineP and routineP parameters


oschuett:
It seem Gfortran has slightly change the format of its warning messages. In particular it now uses backticks (`` ‘routinep’ ``) instead of single quotes (`'routinep'`).

So, we'll have to adopt analyze_gfortran_warnings.py in [line 31](https://github.com/cp2k/cp2k/blob/master/tools/conventions/analyze_gfortran_warnings.py#L31), [line 90](https://github.com/cp2k/cp2k/blob/master/tools/conventions/analyze_gfortran_warnings.py#L90), [line 92](https://github.com/cp2k/cp2k/blob/master/tools/conventions/analyze_gfortran_warnings.py#L92), and [line 104](https://github.com/cp2k/cp2k/blob/master/tools/conventions/analyze_gfortran_warnings.py#L104).

mkrack:
@oschuett Many thanks for the hint.
Unfortunately, these are not backticks, i.e. grave accents (ACSII code 96), which are available on keyboards, but [left](https://www.codetable.net/decimal/8216) and [right](https://www.codetable.net/decimal/8217) quotation marks (ASCII codes 8216 and 8217 or 0x2018 and 0x2019). So, it turns out to be good, that we moved already to Python 3 and UTF-8 support.

mkrack:
@oschuett The formatting run fails due to my changes in analyze_gfortran_warnings.py. I cannot find or get a more detailed error output why it fails. Please, could you have a look?

oschuett:
Yes, these messages from [pre-commit](https://github.com/cp2k/cp2k/wiki/Git-Tips-&-Tricks#git-pre-commit-hooks) are not very self-explanatory. You have to install and run the [black](https://github.com/psf/black) Python Formatter.

Or you simply apply this diff:
```

diff --git a/tools/conventions/analyze_gfortran_warnings.py b/tools/conventions/analyze_gfortran_warnings.py
index 219af6772..82b10ef6d 100644
--- a/tools/conventions/analyze_gfortran_warnings.py
+++ b/tools/conventions/analyze_gfortran_warnings.py
@@ -6,7 +6,7 @@ import argparse
 import re
 import ast
 from os import path
-from argparse import RawTextHelpFormatter # enable newline
+from argparse import RawTextHelpFormatter  # enable newline
 
 blas_re = re.compile(
     r"[SDCZ]"
@@ -139,7 +139,8 @@ For generating the warn-files run gfortran with all warning flags and redirect t
 This can be achieved by putting
     FCLOGPIPE   = 2>&1 | tee $(notdir $<).warn
 in the cp2k arch-file.
-""", formatter_class=RawTextHelpFormatter
+""",
+        formatter_class=RawTextHelpFormatter,
     )
     parser.add_argument(
         "files",
```

mkrack:
@oschuett Many thanks! Now I understand the meaning of 'black'.

PR #1046, opened 2020-08-19
Format C code with clang-format

oschuett:
I decided to bite the bullet and reformatted the comments manually, which allows us to use the default style after all.
See also the discussions in https://github.com/cp2k/dbcsr/pull/358 and https://github.com/cp2k/dbcsr/issues/359.


e-kwsm:
It would be preferable to perform `clang-format --dump-config > .clang-format` and track the file.

oschuett:
> It would be preferable to perform clang-format --dump-config > .clang-format and track the file.

I explicitly went with the default style to avoid an endless stream of style discussions and changes.
I think the default style is just fine (it's good enough for LLVM after all).

e-kwsm:
> I explicitly went with the default style to avoid an endless stream of style discussions and changes.
I think the default style is just fine (it's good enough for LLVM after all).

Agreed but I'm worried that `.clang-format` under one of the parental directories, e.g. `$HOME`, may affect the formatting results.

oschuett:
> I'm worried that .clang-format under one of the parental directories, e.g. $HOME, may affect the formatting results.

Well, `$HOME` would be a rather unfortunate place for `.clang-format` because the formatting style is usually project specific.
Nevertheless, I'll make sure our precommit script explicitly uses `clang-format --style=llvm`. Furthermore, I've added `.clang-format` to `.gitignore` so as a last resort folks can overwrite their local settings.

PR #1054, opened 2020-08-25
Makefile: Replace pretty and doxify with precommit

oschuett:


dev-zero:
Please don't remove the `doxygen` target. A `pre-commit` framework should not be used to generate artifacts (separation of concerns).

Plus, why exactly do we need our own pre-commit framework? One of the reasons why I introduced the `pre-commit` was explicitly to avoid the NIH-Syndrome and re-use what thousands of other projects are using as well, in an effort to make it maintainable and self-documenting.

oschuett:
> Please don't remove the doxygen target.

Indeed, that should stay. I accidentally lumped it in with doxify.

> , why exactly do we need our own pre-commit framework? 

Because the existing frameworks don't satisfy all our requirements.

> ... re-use what thousands of other projects are using as well

The problem with the [pre-commit](https://pre-commit.com) framework which you chose is that all tools need to be install on every developers machine. Basically, it only works well when all your tools are written in Python. While this is true for many projects, we are unfortunately not one of them.

> in an effort to make it maintainable and self-documenting.

My new framework consist of less than 400 lines of Python. I think we'll manage.

Issue #1061, opened 2020-08-28
make pretty with new pre-commit introduces changes after clean checkout


mkrack:
```
diff --git a/src/arnoldi/PACKAGE b/src/arnoldi/PACKAGE
index 57bfcf7..f82a40a 100644
--- a/src/arnoldi/PACKAGE
+++ b/src/arnoldi/PACKAGE
@@ -2,5 +2,7 @@
     "description": "Arnoldi iteration eigenvalue algorithm",
     "archive": "libdbcsrarnoldi",
     "requires": ["../mpiwrap", "../base", "../dbcsrx", "../common"],
-    "public": ["arnoldi_api.F",],
+    "public": [
+        "arnoldi_api.F",
+    ],
 }
diff --git a/tools/fix_unused_public.py b/tools/fix_unused_public.py
index fbec77b..7a7a6b5 100755
--- a/tools/fix_unused_public.py
+++ b/tools/fix_unused_public.py
@@ -127,7 +127,10 @@ def parse_file(fn):
     matches = re_use.findall(content)
     for m in matches:
         uses.append((m.strip(), ("*",)))
-        if m.strip() not in ("iso_c_binding", "f77_blas",):
+        if m.strip() not in (
+            "iso_c_binding",
+            "f77_blas",
+        ):
             print("missing ONLY-clause: ", fn, m)
 
     matches = re_useonly.findall(content)
diff --git a/tools/toolchain/scripts/parse_if.py b/tools/toolchain/scripts/parse_if.py
index 03e7484..4d357d6 100755
--- a/tools/toolchain/scripts/parse_if.py
+++ b/tools/toolchain/scripts/parse_if.py
@@ -61,9 +61,9 @@ class Parser:
 
     def ParseIf(self, string, switch):
         """
-ParseIf(string, switch) will replace in string recursively
-occurance of IF_key(A|B) statements with A if val = True;
-B if val = False
+        ParseIf(string, switch) will replace in string recursively
+        occurance of IF_key(A|B) statements with A if val = True;
+        B if val = False
         """
         result = string
         while result.find("IF_" + switch[0]) > -1:
@@ -72,9 +72,9 @@ B if val = False
 
     def ParseString(self, string):
         """
-ParseString(string) will parse in string recursively
-all of IF_key(A|B) statements for all (key, val) pairs
-in dictionary self.mSwitches
+        ParseString(string) will parse in string recursively
+        all of IF_key(A|B) statements for all (key, val) pairs
[...]

PR #1066, opened 2020-08-29
Cubic GW: fix memory / load balancing issue

pseewald:


oschuett:
The formatting check is failing because of #1061. I have made it non-required for the time beeing so we can continue to merge PRs.

PR #1067, opened 2020-08-30
Revert "Cubic scaling GW: resolve memory bottleneck"

pseewald:
This reverts commit 205154d613b27beb6a32f4908136b3dcb0e42c4c.

This was a premature optimization.

### DIFF
--- a/src/rpa_gw.F
+++ b/src/rpa_gw.F
@@ -117,9 +117,6 @@ CONTAINS
 !> \param mo_coeff ...
 !> \param t_3c_overl_int_gw_RI ...
 !> \param t_3c_overl_int_gw_AO ...
-!> \param t_3c_overl_int_AO_AO ...
-!> \param t_3c_overl_int_AO_MO ...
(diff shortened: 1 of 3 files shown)


pseewald:
There is an issue with the precommit hooks:
`Command '('./tools/prettify/prettify.py', './src/hfxbase/hfx_contraction_methods.F')' timed out after 300 seconds`

oschuett:
Apparently the 5min timeout is a bit tight - I'll increase it.

pseewald:
Shouldn't the precommit hooks check only those files that were changed? I did not change `hfx_contraction_methods.F`.

oschuett:
> Shouldn't the precommit hooks check only those files that were changed?

Yes, by default `prcommit.py` skips files that haven't changed. However, the CI is more thorough and always [checks every file](https://github.com/cp2k/cp2k/blob/master/tools/precommit/entrypoint.sh#L25), because the caching heuristic might fail e.g. when the tools themselves change.

PR #1068, opened 2020-08-30
Restore old `make pretty` as `make fpretty`, add pre-commit installation script and hook into Makefile

dev-zero:
This reverts parts of commit 1525887a18b79688d8ee2d66285ece64b699e213
due to regressions:

- the cloud-based replacement does not scale with the number of local
processors while the old (local) behavior did, meaning that it takes
some key devs 4-5 times longer now to run `make pretty`, depending on
the number of local cores
- since the tools run on the complete codebase (including Python and
C/C++ files), developers relying on the previous behavior of `make pretty`
now get possibly more unrelated changes, interrupting their workflow
further
- `make pretty` is used outside of a commit scenario, for example when
when developers are refactoring code to automatically remove unused
variables, sorting the declaration section, etc. hence pre-commit is not a drop-in replacement

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,16 +23,11 @@ repos:
     rev: master
     hooks:
     -   id: markdownlint
--   repo: git://github.com/doublify/pre-commit-clang-format
-    rev: master
+-   repo: https://github.com/shellcheck-py/shellcheck-py
+    rev: v0.7.1.1
     hooks:
-    -   id: clang-format
-        args: ['--style=llvm']
-        exclude: >-
-          (?x)^(
-            tools/manual/collapsibleList.js|
-            tools/manual/toggle_folding.js|
-          )$
+    -   id: shellcheck
+        files: '^tools/pre-commit-install.sh$'
 -   repo: local
     hooks:
     -   id: doxify
@@ -61,3 +56,16 @@ repos:
         language: system
         files: 'data/*'
         pass_filenames: false
+    -   id: clang-format
+        name: clang-format
+        description: Format files with ClangFormat.
+        entry: clang-format
+        language: python
+        files: \.(c|cc|cxx|cpp|frag|glsl|h|hpp|hxx|ih|ispc|ipp|java|js|m|mm|proto|textproto|vert)$
+        args: ['-i', '-fallback-style=none', '--style=llvm']
+        additional_dependencies: ['clang-format']
+        exclude: >-
+          (?x)^(
+            tools/manual/collapsibleList.js|
+            tools/manual/toggle_folding.js|
+          )$
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
     -   id: check-added-large-files
     -   id: check-ast
 -   repo: https://github.com/ambv/black
-    rev: stable
+    rev: "19.10b0"
     hooks:
     -   id: black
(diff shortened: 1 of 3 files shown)


oschuett:
> the cloud-based replacement does not scale with the number of local processors

It does - up to a [reasonable limit](https://github.com/cp2k/cp2k/blob/master/tools/precommit/precommit.py#L50). In any case, the slow part is formatting Fortran files, which in some cases takes several minutes. In comparison the few remote invocations finish within seconds.

If performance is really the issues then we should look into [fprettify](https://github.com/pseewald/fprettify).

> developers relying on the previous behavior of make pretty now get possibly more unrelated changes

Depends on what you consider _the old behavior_. My interpretations was always that `make pretty` should make the code comply with our formatting conventions.

> make pretty is used outside of a commit scenario,... pre-commit is not a drop-in replacement

Actually `make pretty` calls `precommit.py --allow-modifications`. So, it does serve both use cases.

Obviously, the new precommit system is still very young. Nevertheless, it solves some real issues, like e.g. #1061. Hence, I don't think this PR should be merged.

dev-zero:
> > the cloud-based replacement does not scale with the number of local processors
> 
> It does - up to a [reasonable limit](https://github.com/cp2k/cp2k/blob/master/tools/precommit/precommit.py#L50). In any case, the slow part is formatting Fortran files, which in some cases takes several minutes. In comparison the few remote invocations finish within seconds.
> 
> If performance is really the issues then we should look into [fprettify](https://github.com/pseewald/fprettify).

Let's look at it like this: core count and SSD speed are going up, a data transfer and a request in between will always add latency on top of that, even if you would scale-out on the cloud side.

@pseewald  did an excellent job in tuning [fprettify](https://github.com/pseewald/fprettify), it's mostly our extensions to it which make it slow. And yes, it would be nice if someone worked on that, but I guess it is always easier to simply add some new code rather than trying to understand and fixing old one.

> > developers relying on the previous behavior of make pretty now get possibly more unrelated changes
> 
> Depends on what you consider _the old behavior_. My interpretations was always that `make pretty` should make the code comply with our formatting conventions.

No interpretation needed: it reformatted Fortran code and cleaned up variable declarations. The new `make pretty` does even more unrelated things, breaking peoples workflow.

oschuett:
> Let's look at it like this: core count and SSD speed are going up, a data transfer and a request in between will always add latency on top of that, even if you would scale-out on the cloud side.

As I said, the bulk of the work is formatting the Fortran files which still happens [locally](https://github.com/cp2k/cp2k/blob/master/tools/precommit/precommit.py#L223). If you don't like the limit to [16 workers](https://github.com/cp2k/cp2k/blob/master/tools/precommit/precommit.py#L50) then feel free to change or remove it.

> @pseewald did an excellent job in tuning fprettify, it's mostly our extensions to it which make it slow.

I wasn't aware that out extensions are the actual bottleneck. I certainly agree that @pseewald did a great job with fprettify.

> but I guess it is always easier to simply add some new code rather than trying to understand and fixing old one.

Again, the new precommit system is not about Fortran. It's about adding robust support for C, Python, Markdown, and soon Shell.

> No interpretation needed: it reformatted Fortran code and cleaned up variable declarations.

Well, the old help text said otherwise:

https://github.com/cp2k/cp2k/blob/518f42e8105daa30dc97a3cb3e7180c6def77ba3/Makefile#L340

So, how about we add a new target for formatting only Fortran files: `make fpretty` ?

Issue #1072, opened 2020-08-31
Retire Apidoc and bring back Doxygen


mkrack:
All links to the source files, like to [cp_fm_types](https://github.com/cp2k/cp2k/blob/master/cp2k/src/fm/cp_fm_types.F) linked from [here](https://apidoc.cp2k.org/index.html?cp_fm_types.html), are broken in the  [documentation](https://apidoc.cp2k.org/index.html) since a while.

oschuett:
Yes, the apidoc has been without updates for a while and it's not easy to salvage because it relies on some unlicensed code.
I think we should just return to Doxygen because we're getting more C code and DBCSR (the main use case) has moved away.

PR #1074, opened 2020-09-01
Add SPDX headers and make C banners Doxygen compatible

### LINKED ISSUE #89: Add SPDX License Headers
I think we should add [SPDX License Identifier](https://spdx.org/using-spdx-license-identifier) to our file headers. DBCSR already took [the lead](https://github.com/cp2k/dbcsr/pull/56).

oschuett:
Closes #89 and prepares for #1072.


mkrack:
It seems that the SPDX-License-Identifier `GPL-2.0+` has been [deprecated](https://spdx.org/licenses/GPL-2.0+.html) recently and `GPL-2.0-or-later` is used instead.

dev-zero:
you might want to check https://reuse.software/ and change the copyright line too while being at it

oschuett:
>  SPDX-License-Identifier GPL-2.0+ has been deprecated

Upps, thanks for the hint!

> change the copyright line too while being at it

So, how would you suggest we change it?

dev-zero:
> So, how would you suggest we change it?

    ! SPDX-FileCopyrightText: 2000-2020 CP2K developers group <https://cp2k.org>

based on the most canonical form from https://reuse.software/spec/#comment-headers

But given the option further down, even the following would be conformant:

    ! Copyright 2000-2020 CP2K developers group <https://cp2k.org>

... and I would follow their examples by adding an empty line between copyright notice and SPDX license header:

    ! SPDX-FileCopyrightText: 2000-2020 CP2K developers group <https://cp2k.org>
    !
    ! SPDX-License-Identifier: GPL-2.0-or-later

PR #1075, opened 2020-09-02
grid: collection of minor code changes

hfp:
Capitalized and prefixed macro names (highly recommended), which significantly reduces risk of name-clash for instance with arbitrary variable names (even from common/standard/other headers). Replaced `ceill` with `ceil` (long double is not required; prevents micro-coded instruction flow). Rely on `fmin` and `fmax` when appropriate. Updated help message of driver programs. Fixed some spelling.

### DIFF
--- a/src/grid/common/grid_common.h
+++ b/src/grid/common/grid_common.h
@@ -40,18 +40,18 @@ static const double fac[] = {
     0.26525285981219105864E+33};
 
 //******************************************************************************
-// \brief Macros for integer minimum and maximum (missing from the C standard)
+// \brief Minimum and maximum for integers (missing from the C standard)
 // \author Ole Schuett
 //******************************************************************************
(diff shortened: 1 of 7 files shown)


oschuett:
Thanks the PR. That `ceill` is a great catch!

I also agree that my macros are somewhat risky. However, with capitalization and prefixes it looks kinda ugly. Hence, I'm wondering if we could use inline functions instead?

hfp:
Good suggestion!  With our C99 baseline we can rely on `inline`. In any case, we can use `static` instead of `inline` as well. Let me fix this and reintroduce more readable names. What about calling min/max actually `imax` and `imin`? Though, we now become type-specific, i.e., something like:

```bash
inline int imin(int x, int y) { return (x < y ? x : y); }
```

oschuett:
Yes, `imin` and `imax` are better names. Thanks for making the change!

PR #1077, opened 2020-09-02
Add missing cp2k copyright banners

oschuett:
After this PR all our files will have the same copyright banner.

I also discovered BSD licenses in [t_c_g0.F](https://github.com/cp2k/cp2k/blob/master/src/common/t_c_g0.F), [t_sh_p_s_c.F](https://github.com/cp2k/cp2k/blob/master/src/common/t_sh_p_s_c.F), and [sockets.c](https://github.com/cp2k/cp2k/blob/master/src/sockets.c). I'm actually not sure if the BSD license still applies to the entire file, because subsequent changes within the CP2K project are probably under GPL-2+. To resolve this confusing situation, I'd like to remove the BSD licenses with the permissions of the original authors.

So, @vondele and @ceriottm are you ok with removing these BSD license statements?

### DIFF
--- a/src/sockets.c
+++ b/src/sockets.c
@@ -1,38 +1,41 @@
-/* A minimal wrapper for socket communication.
+/*----------------------------------------------------------------------------*/
+/*  CP2K: A general program to perform molecular dynamics simulations         */
+/*  Copyright 2000-2020 CP2K developers group <https://cp2k.org>              */
+/*                                                                            */
+/*  SPDX-License-Identifier: GPL-2.0-or-later                                 */
+/*----------------------------------------------------------------------------*/
 
-Copyright (C) 2013, Joshua More and Michele Ceriotti
+/*----------------------------------------------------------------------------*/
+/*  Copyright (C) 2013, Joshua More and Michele Ceriotti                      */
+/*                                                                            */
+/*  Permission is hereby granted, free of charge, to any person obtaining     */
+/*  a copy of this software and associated documentation files (the           */
+/*  "Software"), to deal in the Software without restriction, including       */
+/*  without limitation the rights to use, copy, modify, merge, publish,       */
+/*  distribute, sublicense, and/or sell copies of the Software, and to        */
+/*  permit persons to whom the Software is furnished to do so, subject to     */
+/*  the following conditions:                                                 */
+/*                                                                            */
+/*  The above copyright notice and this permission notice shall be included   */
+/*  in all copies or substantial portions of the Software.                    */
+/*                                                                            */
+/*  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,           */
+/*  EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF        */
+/*  MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.    */
+/*  IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY      */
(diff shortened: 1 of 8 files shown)


src/common/xml_parser.F
```diff
@@ -1,49 +1,54 @@
-! xmlparse.f90 - Simple, limited XML parser in Fortran
-!
-! $Id: xmlparse.f90,v 1.14 2007/12/07 10:10:19 arjenmarkus Exp $
```
dev-zero: please list Arjen as a copyright holder here instead of moving his name to just authors

dev-zero: https://sourceforge.net/projects/xml-fortran/ lists the BSD license on the project, so we MUST maintain a copyright notice.

oschuett: I don't want to change the top banner, because the copyright of today's version is shared among several people.

So how about an explicit statement in the Doxygen box:
```
!> \note  Arjen Markus holds the copyright for the initial version of this file.
```

src/sockets.c
```diff
@@ -1,38 +1,18 @@
-/* A minimal wrapper for socket communication.
-
-Copyright (C) 2013, Joshua More and Michele Ceriotti
```
dev-zero, line 3: Do we have an authoritative AUTHORS list? We seem to be referring to cp2k.org, but I couldn't find it on a quick glance. Otherwise without an authoritative AUTHORS file where both of them are explicitly listed, we're essentially removing them for the list of copyright holders.

src/common/xml_parser.F
dev-zero: > I don't want to change the top banner, because the copyright of today's version is shared among several people.

https://reuse.software/faq/#copy-work

dev-zero: So, the solution is simply to add for the files in question the proper copyright headers as well as the original licenses. Since they are compatible, the resulting license of a built cp2k executable will still be GPL2-or-later. Then we also don't have to ask the people to relicense (which as @vondele pointed out in earlier discussions is problematic).
The good thing is that now we're using a standardized SPDX header it should be easy to write a generic file header validator.

dev-zero: the FSFE even provides a [copyright linter tool](https://pypi.org/project/reuse/), together with a [pre-commit hook](https://github.com/fsfe/reuse-tool/blob/master/.pre-commit-hooks.yaml)

src/sockets.c
oschuett: No we don't maintain an AUTHORS list. The authorship is partially documented in the Doxygen boxes and perfectly tracked by the git history.

> we're essentially removing them for the list of copyright holders.

This I don't understand. The copyright assignment doesn't change just because we document it differently.

dev-zero: https://reuse.software/faq/#copyright-holder-author

dev-zero: > This I don't understand. The copyright assignment doesn't change just because we document it differently.

based on the above we are unfortunately not allowed to document it differently

src/common/xml_parser.F
oschuett: > the FSFE even provides a copyright linter tool, together with a pre-commit hook

The point of this PR is to simplify the situation. Hence, I don't want to build out the infrastructure to support even more license mixups.

I found that `xml_parser.F` is completely unused and `glob_matching.F` is barley used. So, can just remove those two.

oschuett:
Alright, I removed the code from Arjen Markus and restored the BSD statements.

I still think it might be legit to remove the BSD licenses in this case, because the authors explicitly submitted the code to a GPL project and presumably intended a dual-licensing. Nevertheless, it would require some clarification by @vondele and @ceriottm.

oschuett:
I just left the original BSD statement as is, because that's what it demands and I'm not actually bothered by the space it takes up.

My goal was to simplify the situation by not having multiple licenses to deal with.

PR #1079, opened 2020-09-03
fix linker errors due to inline definitions

hfp:
Rely on `static` to also get a copy of the respective function into the library/archives (otherwise we see linker error for code that does not include the header with `inline` definitions).


oschuett:
With `static` we get warnings about unused functions. How about making them `static inline` then?

hfp:
Thank you -- yes, that´s what we need. I will update the PR.

oschuett:
The `qs_chargemol.F` file slipped through the CI because of the the delay between push and merge. The remedy would be to enable [Require branches to be up to date before merging](https://docs.github.com/en/github/administering-a-repository/types-of-required-status-checks). However, I feel that would cause more harm than good.

hfp:
Of course, I have updated the branch but the commit still sneaked-in while I was sorting my bits ;-)

PR #1092, opened 2020-09-16
Initial commit of Spack-based toolchain bootstrap

dev-zero:
the basic options are compatible with the original toolchain script with the following restrictions:

* presetting for a Cray environment is not yet available
* CUDA is not yet available
* packages which are not direct dependencies of CP2K are removed
* enforcing an updated GCC is not yet supported (requires an explicit build and an update of the list of compilers)
* additional tools not used in the build process can not yet be installed (valgrind)

there are already some advantages:

* Many build-only dependencies can be reused from the running environment if supported, Spack will bootstrap required packages automatically as-needed
* Spack can reconfigure an environment and consistently rebuild packages as-needed
* ~5000 LOC less to maintain
* the new script is POSIX-shell compatible, and should together with Spack also work on macOS (although not yet tested)

To install packages with Spack we build an environment for each VERSION (ssmp, psmp) with different presets and install `cp2k@master` in each of the environments with `dev-build` which we let only run until after the arch-file is generated. Using `build-env` we generate the file to source to get all the required environment variables setup for building CP2K.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,31 @@ repos:
     rev: v0.7.1.1
     hooks:
     -   id: shellcheck
-        files: '^tools/pre-commit-install.sh$'
+        exclude: >-
+          (?x)^(
+            benchmarks/QS_single_node/do_benchmark|
+            tools/autotools/autogen.sh|
+            tools/autotune_grid/(generate_makefile|get_results).sh|
+            tools/build_utils/get_revision_number|
+            tools/conventions/test_conventions.sh|
+            tools/doxify/(generate_doxygen.sh|doxify.sh)|
+            tools/formatting/test_formatting.sh|
+            tools/hfx_tools/(contraction/create|block_contraction/generate_hfx_contract_block)|
+            tools/toolchain/(
+                install_cp2k_toolchain.sh|
+                scripts/(common_vars|get_openblas_arch|setup_buildtools|signal_trap|tool_kit|generate_arch_files).sh|
+                scripts/install_(
+                    acml|sirius|libsmm|libxc|mkl|cmake|openmpi|spfft|hdf5|intelmpi|
+                    mathlibs|mpich|pexsi|ptscotch|superlu|cosma|quip|reflapack|
+                    libxsmm|libint|elpa|fftw|libvdwxc|openblas|plumed|valgrind|gcc|
+                    gsl|spglib|scalapack
+                    ).sh
+                )|
+            tools/manual/(update_manual|xml2htm)|
+            tools/plan_mpi_omp/plan.sh|
+            tools/precommit/deploy.sh|
+            tools/regtesting/do_regtest|
+          )$
 -   repo: local
     hooks:
     -   id: doxify
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 #
-# make -j 16 sopt popt ssmp psmp
+# make -j 16 ssmp psmp
 #
 # will now perform a parallel build of 4 cp2k executables
 #
@@ -18,7 +18,7 @@ else
  ARCH         := local
 endif
 
-export VERSION=sopt
+export VERSION=ssmp
 
 MAKEFILE     := $(CP2KHOME)/Makefile
 ARCHDIR      := $(CP2KHOME)/arch
--- a/Makefile
+++ b/Makefile
@@ -11,13 +11,7 @@ CP2KHOME     := $(abspath $(shell pwd))
 export CP2KHOME
 endif
 
-ifneq ($(SPACK_COMPILER_SPEC),)
- # SPACK_COMPILER_SPEC is set when running in a Spack build-env
- ARCH         := $(shell spack arch)-$(shell echo $${SPACK_COMPILER_SPEC%%@*})
-else
- ARCH         := local
-endif
-
+ARCH         := local
 export VERSION=ssmp
 
 MAKEFILE     := $(CP2KHOME)/Makefile
--- a/tools/pre-commit-install.sh
+++ b/tools/pre-commit-install.sh
@@ -7,7 +7,7 @@ set -o nounset
 
 PYTHON=$(command -v python3 || true)
 
-if [ x"${PYTHON}" = "x" ] ; then
+if [ -z "${PYTHON}" ] ; then
     echo "ERROR: the python3 executable could not be found, but is required for the complete build process"
     exit 1
 fi
@@ -39,18 +39,18 @@ ENVDIR=$(CDPATH='' cd -- "${ENVDIR}" && pwd)
 PYTHON=$(command -v python)
 PIP=$(command -v pip || true)
 
-if [ x"${PIP}" = "x" ] ; then
+if [ -z "${PIP}" ] ; then
     echo "Bootstrapping the pip command inside the virtual environment..."
 
     FETCHER="$(command -v curl || true)"
     FETCHER_OPTS=""
 
-    if [ x"${FETCHER}" = "x" ] ; then
+    if [ -z "${FETCHER}" ] ; then
         FETCHER="$(command -v wget || true)"
         FETCHER_OPTS="-O-"
     fi
 
-    if [ x"${FETCHER}" = "x" ] ; then
+    if [ -z "${FETCHER}" ] ; then
         echo "ERROR: neither wget nor curl seem to be available, please download"
         echo "    https://bootstrap.pypa.io/get-pip.py"
         echo "manually and run:"
--- a/tools/toolchain-spack/.gitignore
+++ b/tools/toolchain-spack/.gitignore
@@ -1,0 +1,2 @@
+envs
+spack
--- a/tools/toolchain-spack/install_cp2k_toolchain.sh
+++ b/tools/toolchain-spack/install_cp2k_toolchain.sh
@@ -1,0 +1,485 @@
+#!/bin/sh -l
+# author: Tiziano Müller
+# SPDX-License-Identifier: MIT
+
+set -o errexit
+set -o nounset
+
+PYTHON=$(command -v python3 || true)
+if [ -z "${PYTHON}" ] ; then
+    echo "ERROR: the python3 executable could not be found, but is required for the complete build process"
+    exit 1
+fi
+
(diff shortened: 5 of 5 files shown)


oschuett:
This is really nice!

I'm wondering if we should keep the development version of [package.py](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/cp2k/package.py) in our repository? This would make the build reproducible and simplify upgrades. Of course, we should upstream the spack package on a regular basis, e.g. whenever we make a release.

dev-zero:
> I'm wondering if we should keep the development version of [package.py](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/cp2k/package.py) in our repository? This would make the build reproducible and simplify upgrades. Of course, we should upstream the spack package on a regular basis, e.g. whenever we make a release.

I thought about this as well, but haven't reached a conclusion yet. Here are some thoughts:

* wrt releases: a Spack package supports multiple versions, when running the script on a release version of CP2K we should therefore install `cp2k@X.Y.Z` rather than `cp2k@master`. But for a version depending on a tarball we have to know the hash. This could be solved by adding additional version targets for the different CP2K release branches. On the other hand one could argue that people installing a released version of CP2K should rather be using `spack install cp2k@X.Y` and let Spack handle the fetching from the start.
* wrt maintaining our own `package.py`: Providing our own CP2K package file will most likely lead to inconsistencies with the rest of the packages (or Spack itself) the longer it diverges (e.g. for a release I assume that after a couple of years it would not be buildable anymore). For the master we can of course avoid that by syncing changes often. So far I've planned only to have our own `package.py` on a case-by-case basis (for example when adding new dependencies, development changes, etc.).
* wrt reproducibility: A Spack package by default does not pin the version nor the providers (although provider-pinning is done by the script). For reproducibility we would either have to start pinning the versions OR pinning the Spack version pulled from the repo. The latter has again the drawback that the user will not receive fixes. The former requires that we maintain a package list with versions. For pinning versions we do not have to maintain our own `package.py`, but this can be achieved by using the `specs` attribute in the environments `spack.yaml` (with the advantage or disadvantage that Spack would not try to upgrade them).

So, the question is rather: should we do version pinning or not. The way to do this is then probably not via maintaining our own `package.py` but via the environments [abstract specs](https://spack.readthedocs.io/en/latest/environments.html#adding-abstract-specs).

oschuett:
I think, we should approach Spack like Ubuntu. We rely on Ubuntu packages during development and after we make a release the official cp2k package in Ubuntu gets updated so that our end-users can simply type `apt-get install cp2k`.

So, you're right we probably don't need a copy of [package.py](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/cp2k/package.py). In fact `install_cp2k_toolchain.sh` should not install the cp2k package at all, but only its dependencies. This should allow us to make combined code + toolchain changes, like e.g. https://github.com/cp2k/cp2k/commit/e8d986985345a4b1678d598c4fd2f44d4418fcd9, which the CI can test before we merge them.

> So, the question is rather: should we do version pinning or not.

It's probably not a yes or no question. Some dependencies we simply have to pin because of API incompatibilities. For others we might want to offer a default version. Currently, we have the `--install-all` flag for this purpose. Certainly, within the CI we want to pin as much as possible to get reproducible results.

> pinning the Spack version pulled from the repo

We could just use the regular [Spack releases](https://github.com/spack/spack/releases) and upgrade on a regular basis like we do with [Ubuntu](https://github.com/cp2k/cp2k/commit/fd9a017e65d11858faf233015eb83e4266891b15).

dev-zero:
> I think, we should approach Spack like Ubuntu. We rely on Ubuntu packages during development and after we make a release the official cp2k package in Ubuntu gets updated so that our end-users can simply type `apt-get install cp2k`.

Ok. Just for understanding, our CP2K release version installation instructions would then be:

1. if you're happy with a non-optimized binary and are running Fedora/RHEL/Ubuntu/Arch or using Snap/FlatPak/AppImage, use either the distribution-provided package or our statically-linked binary
2. if you need an optimized binary with additional features, install Spack, then run `spack install cp2k@7.1`

?

> So, you're right we probably don't need a copy of [package.py](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/cp2k/package.py). In fact `install_cp2k_toolchain.sh` should not install the cp2k package at all, but only its dependencies.

That's exactly what the new script does: it installs all required packages for the CP2K features selected by the user and uses the Spack CP2K package to generate a matching arch file and an environment loading script.

> This should allow us to make combined code + toolchain changes, like e.g. [e8d9869](https://github.com/cp2k/cp2k/commit/e8d986985345a4b1678d598c4fd2f44d4418fcd9), which the CI can test before we merge them.

Well, for this scenario we may actually need our own CP2K package. With Spack, this commit could have been done in the following way:

First bump the SIRIUS version in Spack upstream and introduce [upper-version-bounds in the CP2K package for older releases](https://github.com/spack/spack/blob/7fb83047fdea6d47b310b60ec5c9058057c54b81/var/spack/repos/builtin/packages/cp2k/package.py#L122-L123) to not break them. Then add a custom CP2K package with [the updated SIRIUS dependency](https://github.com/spack/spack/blob/7fb83047fdea6d47b310b60ec5c9058057c54b81/var/spack/repos/builtin/packages/cp2k/package.py#L125), together with the required patches. Afterwards submit the changes to the CP2K package to Spack and drop our copy of it again.

> > So, the question is rather: should we do version pinning or not.
> 
> It's probably not a yes or no question. Some dependencies we simply have to pin because of API incompatibilities. For others we might want to offer a default version. Currently, we have the `--install-all` flag for this purpose. Certainly, within the CI we want to pin as much as possible to get reproducible results.

With "pin" I mean exact versions, not lower or upper version-bounds (which are usually part of the package dependencies).
Let's take libxc as an example: the initial dependency in the CP2K package was: `depends_on('libxc@4.0.3:', when='+libxc@7.0:')` (= "depend on libxc >= 4.0.3 if CP2K is to be built with libxc-support and the CP2K version is 7 or higher). The addition of libxc-5.0 in Spack broke the CP2K package in Spack since new installations and updates pulled in libxc-5. Until it was patched our builds would have failed (e.g. build was not reproducible). Pinning the libxc version would prevent that, but would also have required manual intervention to upgrade libxc to 4.0.4 and later 4.0.5.
 
> > pinning the Spack version pulled from the repo
> 
> We could just use the regular [Spack releases](https://github.com/spack/spack/releases) and upgrade on a regular basis like we do with [Ubuntu](https://github.com/cp2k/cp2k/commit/fd9a017e65d11858faf233015eb83e4266891b15).

That's an option, but it would delay or complicate certain workflows (for example the one outlined above for the SIRIUS upgrade) and prevent upstream fixes to flow in automatically.

The longer I am thinking about this the more I have the impression that an exact version pinning of the direct dependencies of CP2K within the Spack-based toolchain causes less headache for most devs. The reason why I would like to avoid it, is that we would be partially replicating the dependency-logic already present in the CP2K Spack package (the current version of the script does it only implicitly by specifying the variants/features to enable for CP2K).
But in any case, that is something which can be done as an improvement on top of this commit.

oschuett:
> 2. if you need an optimized binary with additional features, install Spack, then run `spack install cp2k@7.1`

Yes, it would be great if Spack became our main distribution channel for end-users.

> First bump the SIRIUS version in Spack upstream and introduce upper-version-bounds in the CP2K package for older releases to not break them. Then add a custom CP2K package with the updated SIRIUS dependency, together with the required patches. Afterwards submit the changes to the CP2K package to Spack and drop our copy of it again.

That's much more complicated that what we currently have. The problem is that you want to use the upstream package already during development, which leads to a circular dependency. To avoid this the toolchain should not rely on the upstream cp2k package at all. It should only use Spack to install dependencies - just like we do with Ubuntu.

>  and uses the Spack CP2K package to generate a matching arch file and an environment loading script.

So, then we should move the arch file generation into a separate script, e.g. `configure.py`.

>  Pinning the libxc version would prevent that, but would also have required manual intervention to upgrade libxc to 4.0.4 and later 4.0.5.

I think, in this case there is little value in automatic upgrades. It hurts reproducibility and may even introduce regressions. So, having fewer and more thoroughly tested version combinations in the wild is a very good thing.

dev-zero:
> > 1. if you need an optimized binary with additional features, install Spack, then run `spack install cp2k@7.1`
> 
> Yes, it would be great if Spack became our main distribution channel for end-users.
> 
> > First bump the SIRIUS version in Spack upstream and introduce upper-version-bounds in the CP2K package for older releases to not break them. Then add a custom CP2K package with the updated SIRIUS dependency, together with the required patches. Afterwards submit the changes to the CP2K package to Spack and drop our copy of it again.
> 
> That's much more complicated that what we currently have. The problem is that you want to use the upstream package already during development, which leads to a circular dependency. To avoid this the toolchain should not rely on the upstream cp2k package at all. It should only use Spack to install dependencies - just like we do with Ubuntu.
> 
> > and uses the Spack CP2K package to generate a matching arch file and an environment loading script.
> 
> So, then we should move the arch file generation into a separate script, e.g. `configure.py`.

Essentially it boils down on how much we can or want to re-use from what is already in Spack. Spack's `package.py` for CP2K is a recipe for: mapping CP2K optional features (variants) to dependencies + arch file generation + building the package.
The current version of the script re-uses the first two completely.
By installing dependencies manually (which is what we have to do to pin versions, afaik) we partially replicate part 1. Continue using the `package.py` afterwards would give as a consistency check between our toolchain and the Spack package.
Hence I would rather start with pinning only versions in the environment and see that we can keep the arch-file generation (part 2) from Spack if possible.
 
> > Pinning the libxc version would prevent that, but would also have required manual intervention to upgrade libxc to 4.0.4 and later 4.0.5.
> 
> I think, in this case there is little value in automatic upgrades. It hurts reproducibility and may even introduce regressions. So, having fewer and more thoroughly tested version combinations in the wild is a very good thing.

For our development I agree it makes sense to control the build environment closer to identify regressions coming from dependencies closer and to avoid that developers get a slightly different toolchain all the time.

Anyway, thanks for the valuable discussion. Since there's no showstopper here and the rest can be improved on the current script I'm going to merge this (especially since it also contains some other minor fixes) and open an issue to track the version pinning.

PR #1110, opened 2020-10-06
task_list: Remove ALLOCATEs from loop in integrate_v_rspace

oschuett:


alazzaro:
For the record, this PR doesn't improve performance on the Daint perf tests... 
Waiting for the next PR...

Issue #1084, opened 2020-09-04
Performance degradation for non-orthorhombic cells


mkrack:
Running the 256 H2O benchmark input on 2 Cray XC40 nodes (72 CPU cores or MPI processes) using various CP2K versions for an orthorhombic and non-orthorhombic cell gave the following timings listed as
`Total time (Self time for calculate_rho_elec)`
in seconds (single run, no statistics):
| CP2K version | orthorhombic cell | non-orthorhombic cell |
| :-: | -: | -: |
| 2.1 | 371 (31) | 458 (96) |
| 2.4 | 388 (31) | 454 (98) |
| 3.0 | 367 (31) | 460 (97) |
| 4.1 | 367 (31) | 458 (97) |
| 5.1 | 362 (31) | 455 (98) |
| 6.1 | 359 (31) | 502 (97) |
| 7.1 | 343 (30) | 471 (96) |
| 8.0 (trunk) | 353 (44) | 641 (290) |

The current trunk version 8.0 shows a pronounced performance degradation for non-orthorhombic cells.

oschuett:
@mkrack, thanks for the detailed analysis! This is indeed much worse that what I (apparently inadequately) measured.

When I ported the general collocate to C, I simplified the implementation somewhat. The slowdown is probably because I replaced the lenghty recursion for the `exp` function with this [single line](https://github.com/cp2k/cp2k/blob/master/src/grid/ref/grid_ref_collocate.c#L708).

I think, the best way to fix this would be to merge the new cpu backend from @mtaillefumier.

mtaillefumier:
@mkrack, the current implementation of the dgemm backend is still slower than the original fotran code but faster than the reference one. It is not coming from the backend since the dgemm backend was faster than the cp2k original code back in May (22s for 6 iterations for dgemm while I get 34 s for the original Fortran code the test case is H2O-32 OMP_NUM_THREADS=1). Something else is slowing things done.

To be noticed, I did not change any of the fortran code (except adding one parameter to collocate function). I am not going to make any PR until the root cause is found.

oschuett:
I've replaced the exp function in the inner loop with multiplications (#1098). This should give a significant speedup.
@mkrack, can you confirm?

mkrack:
@oschuett Good job!
Here is the updated table with the total times in seconds (single run, no statistics):
| CP2K version | orthorhombic cell | non-orthorhombic cell |
| :-: | -: | -: |
[...]

PR #1114, opened 2020-10-08
grid: Use dgemm for contractions and allocate thread local memory in master

oschuett:
Following up on suggestions https://github.com/cp2k/cp2k/issues/1084#issuecomment-705606065 and https://github.com/cp2k/cp2k/issues/1084#issuecomment-705642538.

### DIFF
--- a/src/grid/Makefile
+++ b/src/grid/Makefile
@@ -24,9 +24,9 @@ ALL_OBJECTS := grid_collocate_replay.o \
 	cd $(dir $<); $(CC) -c $(CFLAGS) $(notdir $<)
 
 grid_collocate_miniapp.x: grid_collocate_miniapp.o $(ALL_OBJECTS)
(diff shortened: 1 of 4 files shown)


hfp:
I would probably call `dgemm` rather than `cblas_dgemm`, and twiddle the arguments of dgemm accordingly. This is preferable for two reasons: (1) Fortran code only relies on BLAS, (2) not every BLAS implementation provides CBLAS. Further, I would just provide the prototypes of, e.g., `dgemm` within the translation unit where it is called. This also prevents the problem of relying on a header file.

oschuett:
I'm now calling `dgemm_` directly. And I've moved the `pab` matrix onto the stack to save one allocation.

@hfp, do you think this is ready to be merged?

hfp:
Looks good to me and tests are apparently passing.

PR #1115, opened 2020-10-09
avoid false sharing

hfp:
Avoid false sharing of data between threads by having page-aligned TLS. Perform final reduction outside of the parallel region (no critical section needed).

### DIFF
--- a/src/grid/ref/grid_ref_task_list.c
+++ b/src/grid/ref/grid_ref_task_list.c
@@ -7,6 +7,7 @@
 
 #include <assert.h>
 #include <omp.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


hfp:
The test failures seem to be unrelated to the source code change: `fatal: Could not parse object 'e219b4c6c981bccc09945eefd375a3f9032ecbb9'`.

src/grid/ref/grid_ref_task_list.c
```diff
@@ -130,26 +131,29 @@ static void collocate_one_grid_level(
...
+  const size_t grid_size = (grid_size_unaligned + align1) & ~align1;
+  void *const grid_pool = malloc(grid_size * nthreads + align1);
+  const uintptr_t grid_pool_aligned = ((uintptr_t)grid_pool + align1) & ~align1;
 
 // Using default(shared) because with GCC 9 the behavior around const changed:
 // https://www.gnu.org/software/gcc/gcc-9/porting_to.html
-#pragma omp parallel default(shared)
+#pragma omp parallel default(shared) num_threads(nthreads)
   {
     // Clear thread local copy of the grid.
```
oschuett: Please move this comment down, or the memset up.

oschuett:
>  fatal: Could not parse object 'e219b4c6c981bccc09945eefd375a3f9032ecbb9

Yes, this sometimes happens when one pushes a new commit while the CI job is starting.

> Perform final reduction outside of the parallel region (no critical section needed).

I chose a critical region to allow threads that finish early to proceed with the reduction. What's the advantage of doing the reduction outside of the parallel region?

hfp:
> I chose a critical region to allow threads that finish early to proceed with the reduction.

Good point!

hfp:
I will let tests pass first and then bring back the critical section.

PR #1123, opened 2020-10-15
Update fypp

frobnitzem:
Explicit utf-8 encoding is needed on open() to work on all OS-s.


oschuett:
@frobnitzem, thanks for the PR!

However, we try to stay close to the [upstream version](https://github.com/aradi/fypp) of Fypp. Hence, I've opened #1125 instead.

PR #1133, opened 2020-10-19
Toolchain: Remove dependencies on locales

### LINKED ISSUE #1128: Make toolchain independend of locals
Recently, locals were added as a dependency to the toolchain: https://github.com/cp2k/cp2k/commit/ff9eaa847ab2b3f17211b2e7cfdd011a2be5ca20, https://github.com/cp2k/cp2k/commit/625daf76555288ddf0e688b5892b62916e3a7eee,  https://github.com/cp2k/cp2k/commit/61740544b7b575d0d5daefac40fdf8df18c7049b, https://github.com/cp2k/cp2k/commit/d158da078f5deb09adf89dddab4efed516268dcf.

This causes friction and warnings, e.g. with Perl. So, I think we should instead specify the encoding [explicitly](https://github.com/cp2k/cp2k/commit/0a944608872d1b4d31fbce5a5b3eca5eea08b63c).

oschuett:
Resolves #1128.

### DIFF
--- a/tools/regtesting/do_regtest
+++ b/tools/regtesting/do_regtest
@@ -10,7 +10,9 @@
 # command line argument passing
 #
 ###################################################################################
-export LC_ALL=en_US.UTF-8
+
+# ensure error messages are in English
+export LC_ALL=C
 
 # init
 dir_base=$PWD
--- a/tools/docker/Dockerfile.test_python
+++ b/tools/docker/Dockerfile.test_python
@@ -7,10 +7,6 @@ WORKDIR /workspace
 COPY ./scripts/install_basics.sh .
 RUN ./install_basics.sh
 
-# set a proper unicode-enabled locale, see https://hub.docker.com/_/ubuntu?tab=description
-RUN localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
-ENV LANG en_US.UTF-8
-
 COPY ./scripts/install_python.sh .
 RUN ./install_python.sh
 
--- a/tools/docker/scripts/install_basics.sh
+++ b/tools/docker/scripts/install_basics.sh
@@ -7,7 +7,6 @@ if grep -q "Ubuntu" /etc/os-release ; then
     apt-get update -qq
     apt-get install -qq --no-install-recommends \
         ca-certificates                         \
-        locales                                 \
         git                                     \
         less                                    \
         nano                                    \
@@ -15,7 +14,6 @@ if grep -q "Ubuntu" /etc/os-release ; then
         rsync                                   \
         wget
     rm -rf /var/lib/apt/lists/*
-    locale-gen "en_US.UTF-8"
     echo "done."
 
 elif grep -q "Fedora" /etc/os-release ; then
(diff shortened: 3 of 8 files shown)


tools/regtesting/do_regtest
```diff
@@ -10,7 +10,6 @@
 # command line argument passing
 #
 ###################################################################################
-export LC_ALL=en_US.UTF-8
```
dev-zero, line 13: The point of this one here is to make sure that when a user reports an error it will be in `en_US.UTF-8` rather than whatever she has on her system. For example messages coming from other (possibly localized) runtimes.

oschuett: This is one is tricky. The `en_US.UTF-8` is not installed by default on Ubuntu. Hence, many of our users won't have it either. The obvious choice would be `C.UTF-8`, but that seems to have [portability issues]( https://github.com/cp2k/cp2k/commit/61740544b7b575d0d5daefac40fdf8df18c7049b) too. Since it's only about the output, maybe we should just go back to `LC_ALL=C`?

dev-zero:
```console
$ wget http://releases.ubuntu.com/18.10/ubuntu-18.10-desktop-amd64.manifest -q -O- | grep -i locales
krb5-locales	1.16-2ubuntu1
locales	2.28-0ubuntu1
```

I doubt many users work inside a minimal Ubuntu docker container.

dev-zero: ... and since SSH usually has `AcceptEnv LANG LC_ALL` a users locale (which is in todays times usually a UTF-8 locale) is also set when logged in from remote. Hence a proper locale rather than `C` or `C.UTF-8` should actually be closer to what our users have.

mkrack: My experience (with e.g. raspbian) is that distributions tend to provide a minimal set of locales (C locale and selected system language) plus the locale generator. So, they leave it to the user to generate missing, actually desired additional locales.

oschuett: > I doubt many users work inside a minimal Ubuntu docker container.

Certainly, every desktop users will have some locales installed, but I don't think we can assume that `en_US.UTF-8` is always among them.

So, I think we should just not overwrite the LC settings. I'm not even sure what localized error messages do_regtest would produce and we respect the user's settings everywhere else anyways.

dev-zero: > My experience (with e.g. raspbian) is that distributions tend to provide a minimal set of locales (C locale and selected system language) plus the locale generator.

JeOS distros like raspbian definitely limit the number of locales installed due to space limitations on MicroSDs, I am not so sure about usual desktop distros.

> So, I think we should just not overwrite the LC settings. I'm not even sure what localized error messages do_regtest would produce and we respect the user's settings everywhere else anyways.

Error messages coming from a C++, Fortran or MPI runtime for example.

oschuett: > Error messages coming from a C++, Fortran or MPI runtime for example.

I can't recall ever seeing a localized error messages from any of those runtimes.

Nevertheless, `LC_ALL=C` ensures that messages are in English without making any further assumptions about the user.

Issue #1116, opened 2020-10-09
Wrong results for GAPW with mGGA functionals


mkrack:
Energy and potential are not consistent using GAPW with meta-GGA functionals. An OT CG SCF run for a water molecule using GAPW with TPSS does not converge. A debug run with diagonalization shows wrong analytical (atomic) forces as discussed [here](https://groups.google.com/g/cp2k/c/K4A-gWzZrMc/m/wRWjktVIBgAl)

abussy:
I am somewhat familiar with GAPW, I'll have a look.

mkrack:
The following H2O input still does not converge with GAPW for the mGGA functional TPSS using OT CG indicating that the issue has not been resolved fully. OT DIIS converges with the basis set DZVP-ALL instead of 6-31Gxx, but there is then still a force mismatch detected.
@abussy Please, could you check again.
```
@SET XCFUN     TPSS

@SET BASIS_SET 6-31Gxx
@SET POTENTIAL ALL

@SET GAPW      0
@IF ${POTENTIAL} == ALL
@SET GAPW      1
@ENDIF

&GLOBAL
 PROJECT H2O_${XCFUN}
 RUN_TYPE debug
 PRINT_LEVEL medium
 PREFERRED_DIAG_LIBRARY sl
&END GLOBAL

&DEBUG
 DEBUG_FORCES yes
 DEBUG_STRESS_TENSOR no
&END DEBUG

&FORCE_EVAL
 METHOD Quickstep
 &DFT
  BASIS_SET_FILE_NAME BASIS_SET
  BASIS_SET_FILE_NAME BASIS_MOLOPT
  BASIS_SET_FILE_NAME EMSL_BASIS_SETS
  BASIS_SET_FILE_NAME GTH_BASIS_SETS
  POTENTIAL_FILE_NAME POTENTIAL
  CHARGE 0
  MULTIPLICITY 1
  &MGRID
   CUTOFF 800
   NGRIDS 5
   REL_CUTOFF 60
  &END MGRID
  &QS
   EPS_DEFAULT 1.0E-14
@IF ${GAPW}
   METHOD GAPW
   ALPHA0_H 10.0
   EPSFIT 1.0E-8
   EPSISO 1.0E-12
   EPSSVD 0.0
   EPSRHO0 1.0E-8
   LMAXN0 4
   LMAXN1 6
   QUADRATURE GC_LOG
@ENDIF
  &END QS
  &SCF
[...]

PR #1140, opened 2020-10-23
Fixed GAPW meta-GGA bug

abussy:
An issue concerning GAPW with meta-GGA functionals was raised (#1116). Energies obtained with this combination were slightly different than those of different programs and DEBUG runs would fail due to wrong analytical forces.

meta-GGA functionals require the computation of the dot product of the gradients of Gaussian functions on the grid. The way it was done was wrong for atomic grids. The gradients were computed in spherical coordinates, but the dot product was made as in cartesian coordinates, leading to inconsistencies and wrong results. 

The fix involves computing the gradients in cartesian coordinates instead, such that the dot product can easily be computed.  This choice of coordinate is also more consistent with the rest of the program, where cartesian coordinates are always preferred for spherical harmonics expansions.

DEBUG runs for all-electron calculations now yield sub 1% errors on the forces (tested with water, 6-31G* and TPSS/M06 for both spin restricted and unrestricted cases). Two regtests files had to be modified.

### DIFF
--- a/src/qs_grid_atom.F
+++ b/src/qs_grid_atom.F
@@ -248,7 +248,6 @@ CONTAINS
                END IF
                grid_atom%cos_azi(ia) = COS(grid_atom%azi(ia))
                pol = ACOS(cosia)
-               IF (grid_atom%sin_pol(ia) < 0.0_dp) pol = -pol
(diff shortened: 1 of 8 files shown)


dev-zero:
ok, wow!

PR #1145, opened 2020-10-28
Change abort to warning

mkrack:


oschuett:
What's the motivation for these changes? Isn't forcing developers to notice and fix these bugs immediately a good thing?

dev-zero:
I am a bit puzzled about these changes: what good is the refcounting when we start to not treat it as an error anymore? During farming or when using CP2K as a library this would result in memleaks, wouldn't it?

mkrack:
There are sporadically program aborts using the Intel compiler in these release routines. It is not clear to me if that is a compiler or a program bug. The recent fix #1138 showed that the GNU compiler is not quite sensitive to the pointer status. CPASSERT triggers an immediate abort of all MPI processes without any useful backtrace information (at least in this case). It is not clear if all processes will show the same ref_count status. This is possibly not the best way to resolve this issue. Do you have any better suggestion?

oschuett:
This always happens with `QS/regtest-rel/Hg_rel.inp`, right? It's probably some uninitialized memory. I'll run valgrind.

mkrack:
> 
> 
> I am a bit puzzled about these changes: what good is the refcounting when we start to not treat it as an error anymore? During farming or when using CP2K as a library this would result in memleaks, wouldn't it?

Memleaks are not the issue here, since the CPASSERT is just a safeguard against "over-releasing" which should not happen in fact, if one can rely on an accurate pointer status. Once the reference counting reaches zero the pointer and its components are de-allocated and the pointer is nullified. Another call to the release routine for that pointer should never reach the CPASSERT.

mkrack:
> 
> 
> This always happens with `QS/regtest-rel/Hg_rel.inp`, right? It's probably some uninitialized memory. I'll run valgrind.

Yes.

oschuett:
Valgrind on `sopt` doesn't find anything. Maybe you could try to run valgrind on your binary?

I actually don't think this has anything to do with the input sections. They just happen to be the first allocations we do for which we have an integrity check. There is probably a buffer overflow that overwrite the sections' memory. You also have recurring crashes in MPI. Maybe those are just another symptom of the same problem?

mkrack:
I tried already valgrind without success. I agree that this issue is most likely a symptom of another problem. The aborts occur always at the very end when everything is released. It is indeed likely that something is overwritten. The test input employs a quite large ANO-type basis set with many primitive Gaussian functions and large l numbers.

oschuett:
> The test input employs a quite large ANO-type basis set with many primitive Gaussian functions and large l numbers.

Yes, my first suspicion was the parser code. I tried making the basis 1000x larger but valgrind still found nothing.

Looking at the [dashboard archive](https://dashboard.cp2k.org/archive/psi-intel-psmp/index_full.html) shows that `Hg_rel.inp` had problems for a [long time](https://dashboard.cp2k.org/archive/psi-intel-psmp/commit_34b43fc68aa664314ba075bd60a8f8a63716ec3a.txt). What's odd is that the failure appear to be clustered. This might hint at a hardware problem. Since `Hg_rel.inp` is one of our longest regtests, it might just get hit by chance.

PR #1148, opened 2020-10-29
grid: Remove legacy implementation

oschuett:


dev-zero:
so, what is the state of #1084? Unfortunately that one can be considered a showstopper for v8.1.

oschuett:
The performance is back at a reasonable level and it will likely continue to improve. At this point I would no longer consider #1084 a showstopper for v8.1.

Going back to the legacy implementation is also no longer trivial because I already removed much of the glue code.

Issue #1047, opened 2020-08-20
Intel compiler warns about argument mismatches in C routines


mkrack:
```
/data/user/krack/rt/intel-psmp/cp2k/src/grid/grid_task_list.c(370): warning #167: argument of type "double (*)[ncoa]" is incompatible with parameter of type "const double (*)[*]"
            /*pab=*/(double(*)[ncoa])pab,
                    ^

/data/user/krack/rt/intel-psmp/cp2k/src/grid/grid_collocate_replay.c(233): warning #167: argument of type "double (*)[size]" is incompatible with parameter of type "const double (*)[*]"
                          /*sphi=*/sphi,
                                   ^

/data/user/krack/rt/intel-psmp/cp2k/src/grid/grid_collocate_replay.c(234): warning #167: argument of type "double (*)[npgf]" is incompatible with parameter of type "const double (*)[*]"
                          /*zet=*/zet_array, basis_set);
                                  ^

/data/user/krack/rt/intel-psmp/cp2k/src/grid/grid_collocate_replay.c(293): warning #167: argument of type "double (*)[3]" is incompatible with parameter of type "const double (*)[3]"
        block_num_list, radius_list, rab_list, &blocks_buffer, task_list);
                                     ^

/data/user/krack/rt/intel-psmp/cp2k/src/grid/grid_collocate_replay.c(403): warning #167: argument of type "double (*)[n1]" is incompatible with parameter of type "const double (*)[*]"
                             n1, n2, o1, o2, la_max, lb_max, pab, cycles,
                                                             ^

/data/user/krack/rt/intel-psmp/cp2k/src/grid/grid_collocate_replay.c(424): warning #167: argument of type "double (*)[3]" is incompatible with parameter of type "const double (*)[3]"
            zetb, rscale, dh, dh_inv, ra, rab, npts_global, npts_local,
                          ^

/data/user/krack/rt/intel-psmp/cp2k/src/grid/grid_collocate_replay.c(424): warning #167: argument of type "double (*)[3]" is incompatible with parameter of type "const double (*)[3]"
            zetb, rscale, dh, dh_inv, ra, rab, npts_global, npts_local,
                              ^

/data/user/krack/rt/intel-psmp/cp2k/src/grid/grid_collocate_replay.c(425): warning #167: argument of type "double (*)[n1]" is incompatible with parameter of type "const double (*)[*]"
            shift_local, border_width, radius, o1, o2, n1, n2, pab, grid_test);
                                                               ^

[...]

PR #1156, opened 2020-11-06
Add dgemm and GPU backends for collocate

mtaillefumier:
This pull request adds two backends as a replacement of the collocate reference implementation. The first backend uses dgemm instead of custom written code, while the second backend is a custom implementation of collocate on GPU. 

- The dgemm backend (or CPU backend) is globally faster than the fortran implementation for the generic case. As for the orthorhombic case, the speedup depends on the basis set used for the calculation. The more complex the basis set is the better. They are two variants of this backend, one implementing the spherical cutoff the other one without. Applying the spherical cutoff has a performance penalty but is useful for running the regtests. It is **on** by default when the CPU backend is activated. 

- The GPU backend (on v100) is around 6 to 7 times faster than running the same input file on the same node with 8 OMP threads (and dgemm). The limiting factor of this backend is the atomicAdd present in the kernels not the arithmetic. I found a solution around it but is still finishing the second version of this kernel. 

Most importantly, these two backends are *not* selected automatically for various reasons. 
- The arithmetic is different which means that the regtests deviate for 1/3 of the time (too tight tolerance) from the reference value. 
- They are still experimental in the sense that improvement can be made. I doubt we can much further improve the dgemm variant without doing specific optimizations.
- the gpu backend is still far from its maximum capacity. 

To test these backends just add this in the global section

&GRID
    BACKEND CPU (GPU)
    APPLY_CUTOFF TRUE or FALSE
&end

*NOTE* you need to explicitly indicate -ftree-vectorize -ftree-loop-vectorize -march=native (or better) to get full vectorization.


oschuett:
Thanks for the PR!

Most formatting issues can resolved by running `make pretty`. The header guards should [match](https://github.com/cp2k/cp2k/blob/master/tools/conventions/analyze_src.py#L134) the file name.

hfp:
Very good work! It more than restores the previous performance. I played a bit with this PR and have the following comments or questions:

* To avoid unresolved symbols, I had to change `inline` to `static inline` in [grid/cpu/tensor_local.h](https://github.com/cp2k/cp2k/blob/c9c89898dc234849cce2c6fddc67b5c1361a872c/src/grid/cpu/tensor_local.h).
* For the gemm/cpu backend with spherical cutoff, the `exp` [based hotspot](https://github.com/cp2k/cp2k/issues/1084#issuecomment-718802092) remains (code reuse from reference backend).
* Is it possible to handle the cutoff automatically (without keyword)?  
This [test input](https://github.com/cp2k/cp2k/issues/1084#issuecomment-718802092) causes OOB accesses if the cutoff is disabled.

I guess `restrict` is a C99 keyword. However, it might be an issue with some older toolchain hence the `__restrict` intrinsic is the way to go. In any case, we can think about a macro layer to ease some of these portability issues. Is this something desired?  I can introduce some of that including macros to wrap-up `iv-dep` and friends (or vectorization helper macros in general).

mtaillefumier:
> Very good work! It more than restores the previous performance. I played a bit with this PR and have the following comments or questions:
> 
>     * To avoid unresolved symbols, I had to change `inline` to `static inline` in [grid/cpu/tensor_local.h](https://github.com/cp2k/cp2k/blob/c9c89898dc234849cce2c6fddc67b5c1361a872c/src/grid/cpu/tensor_local.h).

I am used to c++ where this kind of things rarely happen with inline functions. I had not this problem with gcc but true I did not test with icc. I will change the inline with static inline right away. correction I have -std=c99 in the compilation options
 
>     * For the gemm/cpu backend with spherical cutoff, the `exp` [based hotspot](https://github.com/cp2k/cp2k/issues/1084#issuecomment-718802092) remains (code reuse from reference backend).

I use the `integrate` function from the _reference backend_ that's why you still see the `exp`. I also have `integrate` with dgemm but I would have to break the interface apart to use it as I can not integrate this work on the Fortran side easily any more due to recent changes (computation of the forces were moved on the C side). 
 
>     * Is it possible to handle the cutoff automatically (without keyword)?

if the keyword is not present then the cutoff is applied. This was a design choice to help check the regtests without having to modify the input files. Doing so costs you some performance though. I vectorized the last stage but i is still a memory bounded problem. 

>       This [test input](https://github.com/cp2k/cp2k/issues/1084#issuecomment-718802092) causes OOB accesses if the cutoff is disabled.

Are you running with `mpirun -n`. If so I still need to figure out exactly how this border thing works (it seems to be bound dependent). You should not see any issue if the grid is not split over multiple mpi ranks. the problem is located in one function only. 

> I guess `restrict` is a C99 keyword. However, it might be an issue with some older toolchain hence the `__restrict` intrinsic is the way to go. In any case, we can think about a macro layer to ease some of these portability issues. Is this something desired? I can introduce some of that including macros to wrap-up `iv-dep` and friends (or vectorization helper macros in general).

`restrict` keyword in C99 but gcc, clang (probably icc) support __restrict and __restrict__ as well. these three keywords mean the same thing both in C and C++ but of course c++ standard uses `__restrict` (i believe) instead of `restrict`. I may have mixed the three words.

One way to fix this is to define the keyword for c90 as empty. But then optimization by the compiler might give a difference in performance.

inline is also c99 but it is easier to fix this issue
```

#if __STDC_VERSION__ >= 199901L
  /* "inline" is a keyword */
#else
# define inline static
#endif
```

As for the macros please do so since they are compiler dependent. maybe openmp-simd might do some of the trick. I have to put these macros otherwise gcc does not necessarily vectorize the loops and using libxsmm macros is not really portable. 

Please consider this as advanced/experimental since there are few issues still open. But basically, if I run the regtests and change only one line of the code, then the only problem I have is tolerance. Even the graphite cases passes with this code (it is a difficult one)

hfp:
> I had no problem with gcc, but true I did not test with icc

My test was with GNU GCC; perhaps it somehow depends on compiler flags. There are a lot of legacy ARCH files and people who get confused if we ask "for more" (like additional flags or a newer GCC covering C99 by default). The `inline` keyword makes a difference between C and C++ (among other things). I think we should stick with C99 if not C/earlier.

mtaillefumier:
On 11/11/20 7:57 AM, Hans Pabst wrote:
>
>     I had no problem with gcc, but true I did not test with icc
>
> My test was with GNU GCC; perhaps it somehow depends on compiler
> flags. There are a lot of legacy ARCH files and people who get
> confused if we ask "for more" (like additional flags or a newer GCC
> covering C99 by default). The |inline| keyword makes a difference
> between C and C++ (among other things). I think we should stick with
> C99 if not C/earlier.
>
C99 is a good compromise. C90 as well provided that we suppress al VLAs.
gcc supports c99 since at least 5.0.0 (we do not use complex arithmetic
so it is not an issue). Removing inlined functions is simple as well.

But most of these issues can be solved with a proper build system
(cmake, meson, whatever) because users are also getting confused with
this. For instance, my last PR does not compile with cuda because the
cuda header files are not found (it is the least I could expect, same
with cblas, blas, etc...). I do not want to update SIRIUS to the latest
version because we have a new hard dependency which in itself is not
complicated to compile (it use cmake) but integrating it in the
toolchain will be a nightmare. The same holds with some of the work done
by the unit I am part of, which implement different methods for
minimizing the KS equations but has one extra dependency.

src/grid/hybrid/grid_collocate_hybrid.c
```diff
@@ -0,0 +1,505 @@
...
+/*  CP2K: A general program to perform molecular dynamics simulations         */
+/*  Copyright 2000-2020 CP2K developers group <https://cp2k.org>              */
+/*                                                                            */
+/*  SPDX-License-Identifier: GPL-2.0-or-later                                 */
+/*----------------------------------------------------------------------------*/
+
+#ifdef __COLLOCATE_GPU
+
+#include <assert.h>
+#include <cuda.h>
```
oschuett, line 11: The Cuda headers can not be included into C file, because Cuda assumes C++. Hence, you can only use them in a `.cu` file.

mtaillefumier:
cuda.h is part of the public C and C++ interfaces. It has been like that for ages and I never had any issue with these header files both in C and C++ for as long as cuda is here. The compiler just has to know where they are located.

can be found here in the documentation https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html section 3

nvcc assumes c++ only for the cu files although you can write clean C like kernels. It is because they use a lot of templated code for many of the device functions.

INSTALL.md
```diff
@@ -189,6 +189,8 @@ version of CP2K (ssmp or psmp), the FFTW3 threading library libfftw3_threads
 - Use `-D__CUDA_PROFILING` to turn on Nvidia Tools Extensions.
   It requires to link `-lnvToolsExt`.
 - Link to a blas/scalapack library that accelerates large DGEMMs (e.g. libsci_acc)
+- Use the `-D__COLLOCATE_GPU` flag to compile the gpu backend for collocate. It
```
oschuett: Do you want to have separate flags for collocate and integrate? Otherwise, I'd suggest to use `__GRID_CUDA`, which matches the `__PW_CUDA` that we already have. 

If you want to have separate flags for each backend then I'd suggest something along the lines of `_GRID_HYBRID_BACKEND`.

mtaillefumier:
A common flags is better. I will change it to __GRID_CUDA. Thanks for
the suggestion

oschuett:
> A common flags is better. I will change it to __GRID_CUDA. Thanks for the suggestion

Then please also add it [here](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/generate_arch_files.sh#L104) to the toolchain.

oschuett:
> cuda.h is part of the public C and C++ interfaces.

You are right. It does however throw some warnings for `-Wpedantic` which confused me.

So, I guess we could add `-I${CUDA_PATH}/include` to the [CUDA_DFLAGS](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/generate_arch_files.sh#L104). However, confining Cuda calls to `.cu` files has the advantage that these files are automatically ignored when `nvcc` is not available. Furthermore, I'm a bit worried about the aforementioned `-Wpedantic` noise.

mtaillefumier:
> > cuda.h is part of the public C and C++ interfaces.
> 
> You are right. It does however throw some warnings for `-Wpedantic` which confused me.
> 
> So, I guess we could add `-I${CUDA_PATH}/include` to the [CUDA_DFLAGS](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/generate_arch_files.sh#L104). However, confining Cuda calls to `.cu` files has the advantage that these files are automatically ignored when `nvcc` is not available. Furthermore, I'm a bit worried about the aforementioned `-Wpedantic` noise.

-Wpedantic comes from thrust. I removed the include files since I am not using thrust yet.

mtaillefumier:
> > > cuda.h is part of the public C and C++ interfaces.
> > 
> > 
> > You are right. It does however throw some warnings for `-Wpedantic` which confused me.
> > So, I guess we could add `-I${CUDA_PATH}/include` to the [CUDA_DFLAGS](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/generate_arch_files.sh#L104). However, confining Cuda calls to `.cu` files has the advantage that these files are automatically ignored when `nvcc` is not available. Furthermore, I'm a bit worried about the aforementioned `-Wpedantic` noise.
> 
> -Wpedantic comes from thrust. I removed the include files since I am not using thrust yet.

mtaillefumier:
sorry. Got confused with the interface.

mtaillefumier:
> So, I guess we could add `-I${CUDA_PATH}/include` to the [CUDA_DFLAGS](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/generate_arch_files.sh#L104). However, confining Cuda calls to `.cu` files has the advantage that these files are automatically ignored when `nvcc` is not available. Furthermore, I'm a bit worried about the aforementioned `-Wpedantic` noise.

some of these warnings come from the fact that cuda header files use union without giving them a name. It is not c99 standard but I would simply ignore this. 
the warnings about enum with values set to 0xFFFFFFFF that are not in int range are really strange since int is 32 bits by default. 
The last one is funny. I pass dh[3][3] to a function of this type

`func(const double dh[3][3]) 
`
and it is complaining about it. Really do not understand why. I always put constant any argument that is not modified in the function.

oschuett:
> The last one is funny. I pass dh[3][3] to a function of this type

That is probably another occurrence of #1047. The intel compiler warns about those by default. This is the reason why I use `-Wpedantic` during [development](https://github.com/cp2k/cp2k/blob/master/src/grid/Makefile#L8).

mtaillefumier:
> > The last one is funny. I pass dh[3][3] to a function of this type
> 
> That is probably another occurrence of #1047. The intel compiler warns about those by default. This is the reason why I use `-Wpedantic` during [development](https://github.com/cp2k/cp2k/blob/master/src/grid/Makefile#L8).

the only option is then to pass a pointer such as const double *const dh_,  and then do some ugly magic code as

const double *const dh[3] = {dh_, dh_ + 3, dh_ + 6};

oschuett:
> The last one is funny. I pass dh[3][3] to a function of this type `func(const double dh[3][3])`

You can cast it like this:

`
 func( (const double(*)[3]) dh)
`

However, in the case of `dh` you should be able to just preserve the `const` as this is how it's [passed](https://github.com/cp2k/cp2k/blob/master/src/grid/grid_task_list.h#L103) in.

mtaillefumier:
On 11/12/20 9:38 AM, Ole Schütt wrote:
>
>     The last one is funny. I pass dh[3][3] to a function of this type
>     |func(const double dh[3][3])|
>
> You can cast it like this:
>
> |func( (const double(*)[3]) dh)|
>
> However, in the case of |dh| you should be able to just preserve the
> |const| as this is how it's passed
> <https://github.com/cp2k/cp2k/blob/master/src/grid/grid_task_list.h#L103>
> in.
>
That's a good point.

I have a local copy of dh to avoid passing this all the time to all
functions that need it. I encapsulate this in a struct to be more
pedantic. But it remains that const whatever[][] is somehow troublesome.
One way to do this properly (actually durty) might be to do this (not
even sure it compiles)

func(const void *const a)

{

    const double *const dh[3]= {(const double *const )a, (const double
*const )a + 3, (const double *const )a + 6};

}

or

simply replace dh[3][3] with a 1D array. But dh[3][3] is easier to work
with.

I might simply replace dh[3][3] (in the internals of the backend) with
dh[9] or better const double *const dh.

hfp:
As const-ness was discussed here quite a bit, I wanted to note that constant function arguments do not add to caller's safety if only the variable's value is used (somewhat bulky and blown interface). For example, `double *const pd`, or `const double val` do not improve call-side's safety or protection against modification of any call-side memory. In fact, wrt C++ such `const` does not contribute to the signature in the sense of separate overloads, e.g., `void fun(double *const pd)` and `void fun(double* pd)` are not distinct overloads of `fun`. They protect however the implementation-side by disallowing the use of function's arguments as ordinary variables inside of the function's body.

mtaillefumier:
> As const-ness was discussed here quite a bit, I wanted to note that constant function arguments do not add to caller's safety if only the variable's value is used (somewhat bulky and blown interface). For example, `double *const pd`, or `const double val` do not improve call-side's safety or protection against modification of any call-side memory. In fact, wrt C++ such `const` does not contribute to the signature in the sense of separate overloads, e.g., `void fun(double *const pd)` and `void fun(double* pd)` are not distinct overloads of `fun`. They protect however the implementation-side by disallowing the use of function's arguments as ordinary variables inside of the function's body.

Sure, I know that. The compiler still issues a warning if you want to modify the values of such argument. It is the important information I want to have since I can always make the mistake to modify the value of an argument while it is not intended. In C or C++ it is always possible to find tricks around this so safety is only relative.

mtaillefumier:
compilation is broken again because of last Ole PR. I will fix the error. Still trying to find a way to properly deal with the grid splitting over mpi ranks.

oschuett:
> compilation is broken again because of last Ole PR.

Oh, I really did not expect this to affect your PR. I'm sorry!

mtaillefumier:
Don't worry about it. I use grid_common.h and your task structure (i
find them convenient) and the change you did were actually not much of
an issue. I fixed it already. Please merge your next pull request so I
can update mine accordingly. As I said I just have trouble with the grid
splitting (both cpu and GPU but i find the solution for the cpu the gpu
side is simple). The rest works fine.

mtaillefumier:
I will close this PR. See PR #1178 for an updated version

hfp:
Since #1178 is merged, I will close this PR as mentioned above (please reopen this PR if it is still being worked on).

PR #1160, opened 2020-11-10
Feature/update fprettify

dev-zero:
@pseewald I merged the 3 `ASSOCIATE` stmts in `qs_tensor.F` into a single one


pseewald:
Do we still need python 2.7 support (pseewald/fprettify#91)

oschuett:
> Do we still need python 2.7 support ([pseewald/fprettify#91](https://github.com/pseewald/fprettify/issues/91))

Nope :-)

We've [dropped](https://github.com/cp2k/cp2k/pull/927) support for Python 2 and once https://github.com/cp2k/dbcsr/commit/97086fa19d40784b48a7580377b26bfd48ac552d lands we can remove it from the toolchain too.

dev-zero:
for the record: waiting for a new fprettify release, will update the PR afterwards

pseewald:
@dev-zero fprettify 0.3.7 has been released. Due to the newly added default of formatting e.g. `endif` --> `end if` I expect a rather large diff. I can change the defaults if needed.

edit: there is also fypp indentation by default. I think this is a good default but it could be disabled by `--disable-fypp`.

yff2015cn:
when I use fprettify in Ubuntu18.04 (fprettify 0.3.7 )， I see the following  error:
  File "/home/ff/.local/bin/fprettify", line 7, in <module>
    from fprettify import run
  File "/home/ff/.local/lib/python2.7/site-packages/fprettify/__init__.py", line 75, in <module>
    sys.stdin.detach(), encoding='UTF-8', line_buffering=True)
Could you show me how to fix it?

dev-zero:
> when I use fprettify in Ubuntu18.04 (fprettify 0.3.7 )， I see the following error:
> Traceback (most recent call last):
> File "/home/ff/.local/bin/fprettify", line 7, in
> from fprettify import run
> File "/home/ff/.local/lib/python2.7/site-packages/fprettify/**init**.py", line 75, in
> sys.stdin.detach(), encoding='UTF-8', line_buffering=True)
> Could you show me how to fix it?

Well, this is actually not the right place for it, but from what I can see from this traceback is that you're trying to run `fprettify` with Python 2, which is not supported anymore. You should install `fprettify` with the `pip3` command instead of just `pip`: `pip3 install fprettify`. If you don't have the `pip3` command, install it first with `sudo apt install python3-pip`.
For further requests, please use the `fprettify` project website at https://github.com/pseewald/fprettify

dev-zero:
@pseewald with the updated `fprettify` we're running into a problem here:

```
$ ./tools/prettify/prettify.py src/motion/cp_lbfgs.F 
ERROR src/motion/cp_lbfgs.F: processing file failed
  File "./tools/prettify/prettify.py", line 478, in main
    replace=args.replace,
  File "./tools/prettify/prettify.py", line 277, in prettifyInplace
    outfile = prettifyFile(infile=infile, filename=filename, **kwargs)
  File "./tools/prettify/prettify.py", line 261, in prettifyFile
    "Prettify did not converge in {} steps.".format(max_pretty_iter)
RuntimeError: Prettify did not converge in 5 steps.
0 file(s) prettified, 1 file(s) left unchanged.
```

pseewald:
I'll have a look

dev-zero:
Ok, it seems that `fprettify` and `prettify_cp2k` don't agree on how to indent line continuations. This is one hunk of the diff between iterations:

```
@@ -2589,7 +2589,7 @@
 
 1004  FORMAT(/, a4, 1p, 6(1x, d11.4), /, (4x, 1p, 6(1x, d11.4)))
 2001  FORMAT                                                            &
-          &  (/, 'At iterate', i5, 4x, 'f= ', 1p, d12.5, 4x, '|proj g|= ', 1p, d12.5)
+           &  (/, 'At iterate', i5, 4x, 'f= ', 1p, d12.5, 4x, '|proj g|= ', 1p, d12.5)
 3001  FORMAT(2(1x, i4), 2(1x, i5), 2x, a3, 1x, i4, 1p, 2(2x, d7.1), 1p, 2(1x, d10.3))
 
       RETURN
```

see now pseewald/fprettify#93

dev-zero:
while this is a bug in fprettify I wonder whether this could be a an opportunity to replace the labeled `FORMAT` statements with format strings instead and fix the `WRITE(*, ...)` to a proper iounit.

oschuett:
Obsolete due to #1735.

PR #1161, opened 2020-11-12
NNP: Neural Network Potential force environment

cschran:
Add implementation of Behler-Parrinello HD-NNPs
Contributions by F. Brieuc, O. Marsalek, H. Forbert are acknowledged.


cschran:
I'm right now a little surprised about the `-Werror=realloc-lhs` flag used in the CI. I'm using the toolchain for my development build and `local_warn.psmp` has this only as a warning. If I upgrade the flag to an error, I'm getting a lot of additional errors, for example in `ai_overlap.F`.

src/nnp_environment_types.F
```diff
@@ -0,0 +1,618 @@
...
+      TYPE(molecule_list_type), POINTER                  :: molecules
+      TYPE(particle_list_type), POINTER                  :: particles
+
+      CPASSERT(ASSOCIATED(nnp_env))
+      CPASSERT(nnp_env%ref_count > 0)
+
+      IF (PRESENT(nnp_potential_energy)) THEN
+         nnp_env%nnp_potential_energy = nnp_potential_energy
+      END IF
+      IF (PRESENT(nnp_forces)) nnp_env%nnp_forces = nnp_forces
```
oschuett: When assigning to an ALLOCATABLE please use `(:)` to prevent automatic reallocations, like this:

```
 IF (PRESENT(nnp_forces)) nnp_env%nnp_forces(:) = nnp_forces
```

cschran: Ohh, ok! Got it, thanks!

oschuett:
> I'm using the toolchain for my development build and local_warn.psmp has this only as a warning.

Please use `local.pdbg` for development.

The `local_warn.psmp` is only meant for our convention check to generate input for [analyze_gfortran_warnings.py](https://github.com/cp2k/cp2k/blob/master/tools/conventions/analyze_gfortran_warnings.py).

oschuett:
Can you please add your new test directory to [TEST_DIRS](https://github.com/cp2k/cp2k/blob/master/tests/TEST_DIRS).

cschran:
looks like this was lost in the process of transferring the branch from our local cp2k-repo to my personal fork...

cschran:
all right, the memory leaks are unfortunate. Didn't notice them with my usual valgrinding, but that has happened before. The gnu leak sanitizer seems to do a much better job. I'll deal with those tomorrow. Sorry for the many problems...

oschuett:
No need to apologize. I actually think you prepared this PR very well.

cschran:
Looks like after these initial hiccups, it's now passing the regtests.

src/nnp_force.F
```diff
@@ -0,0 +1,682 @@
...
+      INTEGER, INTENT(IN)                                :: unit_nr
+
+      CHARACTER(len=*), PARAMETER :: routineN = 'nnp_print_bias_forces', &
+         routineP = moduleN//':'//routineN
+
+      CHARACTER(len=default_path_length)                 :: fmt_string
+      INTEGER                                            :: i
+
+      ! TODO use write_particle_coordinates from particle_methods.F
+      IF (unit_nr > 0) THEN
```
dev-zero: since this is the only condition under which this function prints something and there's a single caller as far as I've seen, you could do `IF (unit_nr > 0) CALL nnp_print_bias_force(nnp)` instead?
Same goes for `nnp_print_sumforces`?

src/nnp_force.F
```diff
@@ -0,0 +1,682 @@
...
+!> \author Christoph Schran (christoph.schran@rub.de)
+! **************************************************************************************************
+   SUBROUTINE nnp_print_force_sigma(nnp, unit_nr)
+      TYPE(nnp_type), POINTER                            :: nnp
+      INTEGER, INTENT(IN)                                :: unit_nr
+
+      CHARACTER(len=*), PARAMETER :: routineN = 'nnp_print_force_sigma', &
+         routineP = moduleN//':'//routineN
+
+      CHARACTER(len=default_path_length)                 :: fmt_string
```
dev-zero: `PARAMETER`?

dev-zero (commented):
just some minor inconsistencies in subroutine declarations and other remarks, LGTM

src/nnp_force.F
cschran: That's actually valid for almost all the print functions. Will change that, thanks!

src/nnp_model.F
```diff
@@ -0,0 +1,275 @@
...
+         END DO
+      END DO
+
+      DO i = 1, arc%n_nodes(1)
+         denergydsym(i) = arc%layer(nnp%n_layer)%tmp_der(i, 1)
+      END DO
+
+      CALL timestop(handle)
+
+      RETURN
```
oschuett: `RETURN` is not needed at the end of a SUBROUTINE (see also other occurrences).

src/nnp_model.F
```diff
@@ -0,0 +1,275 @@
...
+                  nnp%arc(i)%n_nodes(j), "nodes in layer", j
+            END DO
+         END DO
+      END IF
+
+      RETURN
+
+   END SUBROUTINE nnp_write_arc
+
+! ***************************************************************************
```
oschuett: Some of the Doxygen boxes have too short top lines. I guess, we have a bug in our doxyfier script.

src/nnp_force.F
cschran: I assume you mean that `routineN` is obsolete, if I'm not using it for anything? I will remove those where ever this is the case.

dev-zero: no, sorry, I mean that `fmt_string` here could be made a `PARAMETER` (with `len=*`) since you write it only once

src/nnp_acsf.F
```diff
@@ -0,0 +1,1336 @@
...
+!> \author Christoph Schran (christoph.schran@rub.de)
+! **************************************************************************************************
+   SUBROUTINE nnp_sort_acsf(nnp)
+      TYPE(nnp_type), POINTER                            :: nnp
+
+      CHARACTER(len=*), PARAMETER :: routineN = 'nnp_sort_acsf', routineP = moduleN//':'//routineN
+
+      INTEGER                                            :: i, j, k, loc
+
+      ! First sort is according to symmetry function type
```
oschuett, line 714: We actually have a generic [sort implementation](https://github.com/cp2k/cp2k/blob/master/src/common/array_sort.fypp).

All you have to do is [include](https://github.com/cp2k/cp2k/blob/master/src/task_list_methods.F#L78) it and define a [less_than routine](https://github.com/cp2k/cp2k/blob/master/src/task_list_methods.F#L2882-L2909).

src/nnp_acsf.F
```diff
@@ -0,0 +1,1336 @@
...
+
+! **************************************************************************************************
+!> \brief Calculate atom centered symmetry functions for given atom i
+!> \param nnp ...
+!> \param i ...
+!> \param dsymdxyz ...
+!> \date   2020-10-10
+!> \author Christoph Schran (christoph.schran@rub.de)
+! **************************************************************************************************
+   SUBROUTINE nnp_calc_acsf(nnp, i, dsymdxyz)
```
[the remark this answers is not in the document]
cschran: you are right in that this function could/should potentially be split into two separate functions. However, I noticed very recently during some benchmarks on archer2 that this function (together with `nnp_calc_energy_forces`) is responsible for a substantial fraction of the runtime in the limit of large system sizes (~8000 atoms).

That means I'm planning to optimize this function and also `nnp_calc_energy_forces` for better performance at large system sizes anyway. That's why I would be inclined to leave it like it is in this first commit, if that is OK with the other developers?

src/nnp_force.F
oschuett: Since `fmt_string` is used only once, it could just be inlined.

src/nnp_acsf.F
cschran: This is the remaining TODO from the comments above. I'll push a first commit that mainly addresses the coding conventions and will look into the sorting afterwards.

oschuett: It seems this PR is essentially ready to be merged. Should I wait for the sorting or do you want to open a separate PR for that?

src/nnp_acsf.F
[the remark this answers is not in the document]
dev-zero: ok, np.

src/nnp_acsf.F
cschran: I think a second PR would actually be nicer. That gives me a little bit more time to focus on other things and I can bundle the sorting with additional improvements that will come up eventually (as mentioned above).

Thanks a lot for your feedback on the PR!

oschuett:
Thanks for contributing!

I've added NNP to the [release notes](https://www.cp2k.org/version_history). If you want to further promote it consider adding something to our [wiki](https://www.cp2k.org/howto) or [youtube channel](https://www.youtube.com/c/Cp2kOrgChannel).

cschran:
Thanks! I'll look into both.

PR #1168, opened 2020-11-16
grid: Add GPU backend

oschuett:


alazzaro:
Something wrong due to this PR for the Daint regtest:

https://www.cp2k.org/static/regtest/trunk/cscs-daint-xc50_gpu/CRAY_XC50-gfortran_gpu.psmp.out

Any clue?

oschuett:
I introduced the new flag `-D__GRID_CUDA`. Please try `make distclean`.

alazzaro:
It does `make distclean` for before each multiplication. Is this macro mandatory? I mean, can we avoid it and still compiling with CUDA?

oschuett:
> I mean, can we avoid it and still compiling with CUDA?

Probably no. ... I guess that's a bug.

alazzaro:
Well, the next question is: is there any reason to switch it off when CUDA is enabled? Can we have only the CUDA macro?

oschuett:
> is there any reason to switch it off when CUDA is enabled? 

Since this is brand new code, it's probably good to have a way to disable it.

Furthermore, we do not yet have a Cuda flag. We have `__PW_CUDA` and some more for DBCSR. Nevertheless, it might a good idea to introduce a single flag to enable all Cuda kernels in CP2K at some point.

alazzaro:
Wait, the -D_ACC is still there for CP2K in general (it is not used in DBCSR):

`-D__ACC` needed to enable accelerator support`

mtaillefumier:
I agree that we need to be able to disable it if needed, especially with
two different ways to do computation on GPU.

alazzaro:
Disabling is fine, but I would rather go to a switch a runtime...

mtaillefumier:
There is a switch at runtime.

section GRID, keyword BACKEND.

if you add this

section GRID

BACKEND REF

&end

in the global section of your input file you have what you want.

oschuett:
The `__ACC` flag is not used in CP2K either - apparently it was left behind during the DBCSR extraction.

I think, 2021 will be the year where we figure out how to manage multiple Cuda kernels running on multiple GPUs.
However, for the time being I'd rather keep separate kill switches for maximum control.

I've opened #1172 to fix your build issues.

alazzaro:
We are going in a loop here... I don't need to disable it... I'm asking if there is a need to avoid the compilation if CUDA is enabled in CP2K (via -D_ACC). You said you want to keep the macro (a specific macro for it) for some development reason, and now you are telling me that there is a flag at compile time, so I wonder what we happen if I have disabled at compile time and then I enabled at runtime...

mtaillefumier:
You have two options. Either you compile cp2k the -D__GRID_CUDA (it is a
compilation constant not a macro, collocate will run on GPU by default)
and you can run cp2k normally with an extra option in your input file
*or* you can compile cp2k all together *without* the -D__GRID_CUDA
compilation flags (collocate is done on cpu). It is up to you. The
current PR just fix the compilation issue nothing more.

oschuett:
> I wonder what we happen if I have disabled at compile time and then I enabled at runtime...

Then you'll end up [here](https://github.com/cp2k/cp2k/blob/master/src/grid/grid_task_list.c#L72) :-)

> You said you want to keep the macro 

Actually, I think we should remove `-D__ACC` for the time being because it's too generic.

> I'm asking if there is a need to avoid the compilation if CUDA is enabled in 

While there is a runtime flag, the compile time flag is more thorough.

alazzaro:
> The `__ACC` flag is not used in CP2K either - apparently it was left behind during the DBCSR extraction.
> 
> I think, 2021 will be the year where we figure out how to manage multiple Cuda kernels running on multiple GPUs.
> However, for the time being I'd rather keep separate kill switches for maximum control.

Well, it is a placeholder for future activity ;)

https://github.com/cp2k/cp2k/blob/8bb4190b3113b082ab19807aa877e0f86d433aeb/src/cp2k_info.F#L134-L136

mtaillefumier:
it might but I agree with @oschuett <https://github.com/oschuett>.
-D__ACC is way too generic. We need a different flag for this core part
of cp2k.

oschuett:
> Well, it is a placeholder for future activity ;)

Yes, once we have gathered some experience I'm happy to reconsider. However, for the next release (in about 6 weeks) I'd very much like to keep a separate flags so people have to opt in explicitly.

PR #1169, opened 2020-11-16
GAPW meta-GGA bug fixing

abussy:
This PR is an answer to issue #1116 and a follow-up to PR #1140, which respectively reported and tried to address a bug in GAPW with meta-GGA functionals. I was not able to find the exact issue in the code. Instead, I rewrote from scratch the 2 main steps involved, namely evaluating tau on the atomic grids and integrating the derived potential. The new code is slightly more brute force, as everything is explicitly put on the grid and no corner is cut.

Testing involved debug runs for the forces, on all combination of 6-311G** and def2-DZVP basis sets, TPSS and M06_L functionals and H2O, CO and C2H4 molecules. Most errors are at 0.01% or less (max error at 0.07%). Moreover, the OT method was used in combination with the CG minimzer and 3PNT LINESEARCH and the total energy was systematically decreasing at each CG step of the SCF procedure, which is sound behaviour.

### DIFF
--- a/src/qs_vxc_atom.F
+++ b/src/qs_vxc_atom.F
@@ -26,8 +26,9 @@ MODULE qs_vxc_atom
    USE memory_utilities,                ONLY: reallocate
    USE message_passing,                 ONLY: mp_sum
    USE orbital_pointers,                ONLY: indso,&
-                                              nsoset,&
-                                              soset
+                                              nsoset
+   USE paw_proj_set_types,              ONLY: get_paw_proj_set,&
+                                              paw_proj_set_type
    USE qs_energy_types,                 ONLY: qs_energy_type
    USE qs_environment_types,            ONLY: get_qs_env,&
                                               qs_environment_type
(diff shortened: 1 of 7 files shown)


src/xc/xc_atom.F
```diff
@@ -595,9 +595,9 @@ SUBROUTINE fill_rho_set(rho_set, lsd, nspins, needs, rho, drho, tau, na, ir)
       LOGICAL, INTENT(IN)                                :: lsd
       INTEGER, INTENT(IN)                                :: nspins
       TYPE(xc_rho_cflags_type), INTENT(in)               :: needs
-      REAL(dp), DIMENSION(:, :), POINTER                 :: rho
-      REAL(dp), DIMENSION(:, :, :, :), POINTER           :: drho
-      REAL(dp), DIMENSION(:, :), POINTER                 :: tau
+      REAL(dp), DIMENSION(:, :)                          :: rho
```
dev-zero: `INTENT` missing

src/qs_vxc_atom.F
```diff
@@ -307,24 +305,20 @@ SUBROUTINE calculate_vxc_atom(qs_env, energy_only, exc1, gradient_atom_set, &
...
-                                           trho_h, trho_s, tau_h, tau_s)
-                  END IF
                   IF (donlcc) THEN
                      CALL calc_rho_nlcc(grid_atom, nspins, gradient_f, &
                                         ir, rho_nlcc(:, 1), rho_h, rho_s, rho_nlcc(:, 2), drho_h, drho_s)
                   END IF
-                  CALL fill_rho_set(rho_set_h, lsd, nspins, needs, rho_h, drho_h, tau_h, na, ir)
-                  CALL fill_rho_set(rho_set_s, lsd, nspins, needs, rho_s, drho_s, tau_s, na, ir)
+                  CALL fill_rho_set(rho_set_h, lsd, nspins, needs, rho_h, drho_h, tau_h(:, ir, :), na, ir)
+                  CALL fill_rho_set(rho_set_s, lsd, nspins, needs, rho_s, drho_s, tau_s(:, ir, :), na, ir)
```
dev-zero, line 321: this will generate temporaries, right? Just to be aware of it.

abussy: I did not think about it when coding, but yes, I think it will generate temporaries. I don't see how to address it without changing the overall flow though.

alazzaro:
Wow, this PR triggered an internal "catastrophic" error for the Intel compiler:

https://dashboard.cp2k.org/archive/psi-intel-psmp/commit_e18b1c5ca44b5f3bf3bf309180ce69b3151bd695.txt

abussy:
I merged two PR simultaneously, and judging by the text, it seems that #1170 is responsible instead.

alazzaro:
You are too efficient 🥇 

I think the reason is that the OpenMP region is too big. Maybe @hfp can take a look?
For the time being, I agree with your PR to remove that region...

abussy:
@alazzaro @hfp I don't think it is necessary to have a look into it. This is not a critical part of my code in terms of performance.

PR #1171, opened 2020-11-16
common/rng: add a shuffle() function

dev-zero:


hfp:
Cool!  If you are interested in "shuffling without shuffling" you can have a look at [my idea of co-prime/2](https://github.com/hfp/libxsmm/blob/master/include/libxsmm_math.h#L63). I am just mentioning this as it impacts the interface (or the way the code is written on the client-side). In a narrow sense my shuffle is not "random" but a "Restklassenring". An example for using `libxsmm_shuffle` is [here](https://github.com/hfp/libxsmm/blob/master/samples/magazine/magazine_xsmm.c#L81).

dev-zero:
@hfp ah, cool, thanks! Well, I guess for many things that method would be enough, but then it is not necessarily something I would put in an RNG class (or maybe I misunderstood something). Reminds me of the first part of [this blog entry here](https://probablydance.com/2018/06/16/fibonacci-hashing-the-optimization-that-the-world-forgot-or-a-better-alternative-to-integer-modulo/) I read some time ago, although their focus is a different one.

hfp:
@dev-zero Thank you for the URL, I like such nuggets very much! You are right, this sort of shuffle is not so much about RNG. Although, if you aim for a repeatable stream of random numbers ("deterministic"), it becomes hard to judge one or the other.

abussy:
Shouldn't the reshuffling routine of  `hfx_load_balance_methods.F` be replaced by this as part of this PR?

https://github.com/cp2k/cp2k/blob/d8fd13adce305730108831e14eb3bc4dfdde74ff/src/hfx_load_balance_methods.F#L2515-L2540

dev-zero:
@abussy yes, will do that, but I didn't want to break your PR since yours was first

abussy:
@dev-zero dev Ah, thanks. I understood it the otherway around.

abussy:
The reason why the regtest is currently failing is on me. The shuffling of the array leads to the output being in a somewhat random order, messing up the regtester. I'll fix it.

abussy:
Fixed in #1176.

PR #1172, opened 2020-11-16
grid: Allow using Cuda without __GRID_CUDA

oschuett:


alazzaro:
OK, thanks. Once it is merged I will add the change on the Daint regtest, if it is OK for you...

mtaillefumier:
it means changing the toolchain as well if you want this to work the way
you intend.

PR #1178, opened 2020-11-17
Added the hybrid cpu/gpu and dgemm backends

mtaillefumier:
This pull request adds two backends as a replacement of the collocate reference implementation. The first backend uses dgemm instead of custom written code, while the second backend is a custom implementation of collocate on GPU.

The dgemm backend (or CPU backend) is globally faster than the fortran implementation for the generic case. As for the orthorhombic case, the speedup depends on the basis set used for the calculation. The more complex the basis set is the better. They are two variants of this backend, one implementing the spherical cutoff the other one without. Applying the spherical cutoff has a performance penalty but is useful for running the regtests. It is on by default when the CPU backend is activated.

The HYBRID (GPU/CPU) backend (on v100) is around 6 to 7 times faster than running the same input file on the same node with 8 OMP threads (and dgemm). It supports multiple GPU / rank natively (off right now, since cp2k does not support this mode). The limiting factor of this backend is the atomicAdd present in the kernels not the arithmetic and the decontraction on CPU (although this will be solved soon). I found a solution around it but I am still finishing the second version of this kernel. It is also slightly slower than the current GPU backend but these two backends will eventually be merged at some point. 

Running with split grid works now for both backends. 

Most importantly, these two backends are not selected automatically for various reasons.
- The arithmetic is different which means that the regtests deviate for 1/3 of the time (too tight tolerance) from the reference value. this affects teh dgemm for reasons that I still do not fully understand.
- They are still experimental in the sense that improvement can be made. I doubt we can much further improve the dgemm variant without doing specific optimizations.
- The gpu backend is still far from its maximum capacity.

To test these backends just add this in the global section

&GRID
BACKEND CPU (HYBRID)
APPLY_CUTOFF TRUE or FALSE
&end

They are minor modifications to the reference and GPU backends. Mostly use of enum for the different backends and function types. 

NOTE you need to explicitly indicate -ftree-vectorize -ftree-loop-vectorize -march=native (or better) to get full vectorization.


src/grid/grid_task_list.c
```diff
@@ -33,60 +41,119 @@ void grid_create_task_list(
...
+
+    /* does not allocate anything on the GPU. */
+    /* allocation only occurs when the collocate (or integrate) function is
+     * called. Resources are released before exiting the function */
+
+    /* I do *not* store the address of config.device_id */
+    initialize_grid_context_on_gpu(
+        (*task_list)->hybrid, 1 /* number of devices */, &config.device_id);
+    update_queue_length((*task_list)->hybrid, config.queue_length);
+    (*task_list)->backend = GRID_BACKEND_HYBRID;
```
oschuett: This is a lot of code that could be moved into your backends. Maybe you can follow the existing scheme and introduce `grid_hybrid_create_task_list` and  `grid_cpu_create_task_list` routines? These routines could also contain the calls to `apply_cutoff`.

src/grid/grid_task_list.c
```diff
@@ -121,46 +201,47 @@ void grid_free_task_list(grid_task_list *task_list) {
...
-  double *grid_before[nlevels]; // Only used for validation.
-  if (task_list->validate) {
-    for (int level = 0; level < nlevels; level++) {
-      const size_t sizeof_grid = sizeof(double) * npts_local[level][0] *
-                                 npts_local[level][1] * npts_local[level][2];
-      grid_before[level] = malloc(sizeof_grid);
-      memcpy(grid_before[level], grid[level], sizeof_grid);
-      memset(grid[level], 0, sizeof_grid);
-    }
-  }
```
oschuett, line 140: Why did you remove this part?

src/grid/grid_task_list.c
```diff
@@ -169,27 +250,32 @@ void grid_collocate_task_list(
...
+      break;
     }
     // Call reference implementation.
     grid_ref_collocate_task_list(task_list->ref, orthorhombic, func, nlevels,
                                  npts_global, npts_local, shift_local,
                                  border_width, dh, dh_inv, grid_ref);
 
     // Compare results.
-    const double tolerance = 1e-12;
+    const double tolerance = 1e-14; // TODO: tune to a reasonable value.
```
oschuett: This probably slipped in by accident.

src/grid/grid_task_list.c
```diff
@@ -200,21 +286,17 @@ void grid_collocate_task_list(
...
-              fprintf(stderr, "   value:    %le\n", ref_value);
-              fprintf(stderr, "   level:    %i\n", level);
-              fprintf(stderr, "   idx:      %i\n", idx);
-              fprintf(stderr, "   ijk:      %i  %i  %i\n", i, j, k);
+              printf("Error: Grid validation failure\n");
+              printf("   diff:     %le\n", diff);
+              printf("   rel_diff: %le\n", rel_diff);
+              printf("   value:    %le\n", ref_value);
+              printf("   level:    %i\n", level);
+              printf("   ijk:      %i  %i  %i\n", i, j, k);
```
oschuett: This probably slipped in by accident.

src/input_cp2k_global.F
```diff
@@ -730,6 +733,22 @@ SUBROUTINE create_grid_section(section)
...
+                          " apply a spherical cutoff on the top of the cube. "// &
+                          " There is a performance penalty using it in "// &
+                          "combination with the cpu backend but it is on by "// &
+                          "default for the regtests", default_l_val=.TRUE., &
+                          lone_keyword_l_val=.TRUE.)
+      CALL section_add_keyword(section, keyword)
+      CALL keyword_release(keyword)
+
+      CALL keyword_create(keyword, __LOCATION__, name="QUEUE_LENGTH", &
+                          description="Number of gaussians that the gpu backend treats in one round", &
```
oschuett: gpu -> hybrid

src/input_cp2k_global.F
```diff
@@ -730,6 +733,22 @@ SUBROUTINE create_grid_section(section)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
+      CALL keyword_create(keyword, __LOCATION__, name="APPLY_CUTOFF", &
+                          description="When enabled the cpu and gpu backends "// &
```
oschuett: gpu -> hybrid

src/grid/grid_task_list.c
mtaillefumier: I am afraid this will stay there. What is the point here. Unhappy with with the fact the name of function contains the functionality first and then on it applies to. I am not changing this. 

Also why should i remove the fact that I update a structure inside a function that creates it. Updating and creating are two different things that why I did it.

oschuett: My point is that this logic is an implementation detail of your backends that does not belong into this high level file.

While I really wish you'd just follow the scheme that I have laid out, I'm willing to merge this nevertheless as it does not inflict the other backends.

src/input_cp2k_global.F
mtaillefumier: corrected, thanks.

src/input_cp2k_global.F
mtaillefumier: thanks.

src/grid/grid_task_list.c
mtaillefumier: I am afraid so. I put it back in place.

src/grid/grid_task_list.c
mtaillefumier: because anytime the function containing this code is called the grid is always set to zero before (it is done on the fortran side). So the copy to temporary array, then the second zeroing of the grid is not needed. You can just apply the collocate function to the grid and then call the reference backend to the second temporary space you allocate after and do the comparison.

src/input_cp2k_global.F
oschuett: Are you sure? I still read "gpu backend".

src/grid/grid_task_list.c
```diff
@@ -169,27 +251,32 @@ void grid_collocate_task_list(
...
+      break;
     }
     // Call reference implementation.
     grid_ref_collocate_task_list(task_list->ref, orthorhombic, func, nlevels,
                                  npts_global, npts_local, shift_local,
                                  border_width, dh, dh_inv, grid_ref);
 
     // Compare results.
-    const double tolerance = 1e-12;
+    const double tolerance = 1e-12; // TODO: tune to a reasonable value.
```
oschuett: I actually think 1e-12 is a reasonable value.

mtaillefumier: I have no problem with this.

src/input_cp2k_global.F
mtaillefumier: Corrected. My apologies I thought I corrected it.

src/grid/grid_task_list.c
mtaillefumier: trailing comment. removed.

oschuett:
LGTM, shall I merge it?

PR #1180, opened 2020-11-20
changed "inline" to "static inline" for compilation prior to C99

hfp:


hfp:
Convention check error is unrelated.

```text
Found 1 unsuppressed issues:

  semi_empirical_int_utils.F: Routine EVAL called with an implicit interface.
```

PR #1182, opened 2020-11-20
optimized some grid reorder function

hfp:
This mainly improves code generation with GCC.

### DIFF
--- a/src/grid/ref/grid_ref_collint.h
+++ b/src/grid/ref/grid_ref_collint.h
@@ -404,25 +404,6 @@ static inline void general_ci_to_grid(
     const double gp[3], GRID_CONST_WHEN_COLLOCATE double *ci,
     GRID_CONST_WHEN_INTEGRATE double *grid) {
 
-  //--------------------------------------------------------------------
-  // Find bounds for the inner loop based on a quadratic equation in i.
-  //
-  // The real-space vector from the center of the gaussian to the
-  // grid point i,j,k is given by:
-  //   r = (i-gp[0])*dh[0,:] + (j-gp[1])*dh[1,:] + (k-gp[2])*dh[2,:]
-  //
-  // Separating the term that depends on i:
-  //   r = i*dh[0,:] - gp[0]*dh[0,:] + (j-gp[1])*dh[1,:] + (k-gp[2])*dh[2,:]
-  //     = i*dh[0,:] + v
-  //
-  // The squared distance works out to:
-  //   r**2 = dh[0,:]**2 * i**2  +  2 * v * dh[0,:] * i  +  v**2
-  //        = a * i**2           +  b * i                +  c
-  //
-  // Solving r**2==radius**2 for i yields:
-  //    d =  b**2  -  4 * a * (c - radius**2)
-  //    i = (-b \pm sqrt(d)) / (2*a)
-  //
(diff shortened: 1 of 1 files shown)


hfp:
Merging this PR shall wait until we have Dashboard MC results of previous changes.

alazzaro:
> Merging this PR shall wait until we have Dashboard MC results of previous changes.

Do you mean the MC tests on Daint?

hfp:
> Do you mean the MC tests on Daint?

Yes, unless we want to judge my performance claim differently. It should at least show no regression on top of Mathieu's changes. Although, now that I write this comment - ref-backend shall not be affected by M.'s change...

mtaillefumier:
Short question. The dashboard uses the regtests inputs files or not ? If
so none of my code should be called at all since it always use the
reference backend. Of course compilation issues are a separate issue.

alazzaro:
Just started the test ;)

mtaillefumier:
On 11/20/20 11:29 AM, Hans Pabst wrote:
>
>     Do you mean the MC tests on Daint?
>
> Yes, unless we want to judge my performance claim differently. It
> should at least show no regression on top of Mathieu's changes.
> Although, now that I write this comment - ref-backend shall not be
> affected by M.'s change...
>
No they should not. The only thing I modified is the use of enum instead
of #define. And the two changes of the new code should not either.

alazzaro:
The daint tests are using some of the tests from:

https://github.com/cp2k/cp2k/tree/master/benchmarks

So, we are not changing the input...

hfp:
> Just started the test ;)

> No they should not.

Ok, I merge this PR when green since Alfio triggered the latest MC perf test. With this PR merged, we should at least see no degradation.

mtaillefumier:
Thanks. Then they should pass. If they don't it means something else is
going on.

mtaillefumier:
If the code compiles. If libxsmm is not used then the compilation will
fail because of the unused parameter warning treated as error. I have
the workaround already. inlined functions as well but it was fixed already.

src/grid/ref/grid_ref_collint.h
```diff
@@ -396,13 +396,13 @@ static inline void general_ci_to_reg(const int lp, const double exp_aiibic,
...
-                   GRID_CONST_WHEN_INTEGRATE double *grid) {
+static inline void general_ci_to_grid(
+    const int lp, const int jg, const int kg, const int ismin, const int ismax,
+    const int npts_local[3], const int bounds_i[2], const int index_min[3],
+    const int map_i[], const double zetp, const double a, const double b,
+    const double c, const double exp_ab, const double exp_2a,
+    const double gp[3], GRID_CONST_WHEN_COLLOCATE double *ci,
+    GRID_CONST_WHEN_INTEGRATE double *grid) {
 
   //--------------------------------------------------------------------
```
oschuett: Maybe this comment should be moved alongside.

alazzaro:
Test completed: https://dashboard.cp2k.org/archive/cray-xc50-daint-mc-psmp_perf/index.html
Seems everything OK

src/grid/ref/grid_ref_collint.h
hfp: Yes, that was my concern too. I will put attach it to the call-side where this work now takes place.

Issue #911, opened 2020-05-12
Regtest G0W0_H2O_PBE0_30_pts.inp fails on PASCAL


alazzaro:
I've noticed that the test

QS/regtest-gw-cubic/G0W0_H2O_PBE0_30_pts.inp

is failing from time to time on the CUDA Pascal dashboard (check the latest commits). The message is:
```
 *** WARNING in rpa_util.F:919 :: The Cholesky decomposition before        ***
 *** inverting the RPA matrix / dielectric function failed. In case of     ***
 *** low-scaling RPA/GW, decreasing EPS_FILTER in the &LOW_SCALING section ***
 *** might increase the overall accuracy making the matrix positive        ***
 *** definite. Code will abort.                                            ***
```
It seems that the problem started somewhere after with commit 2b878fb

pseewald:
The failure of this regtest seems to be really specific to this system or configuration. I think Cholesky decomposition is using pdgemm so could it be an issue with COSMA?

Judging from the [dashboard](https://dashboard.cp2k.org/archive/cuda-pascal/index.html) this issue was introduced somewhere between 
c9a73d8af15f4c1569558323638c3ee4ad53d705 and b04255bb341f02ca9ce559e82f1d9644be817dd9

Was there any update in the libraries between those commits?

This commit range also includes a DBCSR update. I noticed that DBCSR behaves differently since rc-13: in this regtest we are filtering DBCSR matrices with `eps_filter` of `0.0` which previously did not remove blocks with exact zeroes but now it does. I see this from a diff on the RPA output:
```
<      Time:                                                              0.077661
<      Occupancy of D occ:                                      3.2E+03 / 100.000%
<      Occupancy of D virt:                                     3.2E+03 / 100.000%
---
>      Time:                                                              0.014462
>      Occupancy of D occ:                                      0.0E+00 /   0.000%
>      Occupancy of D virt:                                     0.0E+00 /   0.000%
```

This is probably not the reason for this issue, but I wonder whether that change in DBCSR is intended.

alazzaro:
> This commit range also includes a DBCSR update. I noticed that DBCSR behaves differently since rc-13: in this regtest we are filtering DBCSR matrices with `eps_filter` of `0.0` which previously did not remove blocks with exact zeroes but now it does. I see this from a diff on the RPA output:
> 
> ```
> <      Time:                                                              0.077661
> <      Occupancy of D occ:                                      3.2E+03 / 100.000%
[...]

PR #1187, opened 2020-11-23
Docker: Pin DBCSR version for pdbg test

oschuett:
See discussion in #911

### DIFF
--- a/tools/docker/scripts/test_regtest.sh
+++ b/tools/docker/scripts/test_regtest.sh
@@ -19,6 +19,17 @@ if command -v ompi_info &> /dev/null ; then
     export OMPI_MCA_plm_rsh_agent=/bin/false
 fi
 
+# Switch to stable DBCSR version if requested.
+if [ -n "${USE_STABLE_DBCSR}" ] ; then
+    echo "Switching to stable DBCSR version..."
+    if ! git -C cp2k/exts/dbcsr checkout -b v2.1.0-rc16 6c52382 ; then
+        echo -e "\nSummary: Could not checkout stable DBCSR version."
+        echo -e "Status: FAILED\n"
+        exit 0
+    fi
+fi
+
+# Compile cp2k.
 echo -en "\nCompiling cp2k... "
 cd /workspace/cp2k || exit 1
 if make -j ARCH="${ARCH}" VERSION="${VERSION}" &> make.out ; then
@@ -31,7 +42,7 @@ else
     exit 0
 fi
 
-
+# Check benchmark files for input errors.
 if [[ "${ARCH}" == "local" ]] ; then
     echo -en "\nChecking benchmarks... "
     if ! ./tools/regtesting/check_inputs.py "./exe/${ARCH}/cp2k.${VERSION}" "./benchmarks/" ; then
@@ -41,7 +52,7 @@ if [[ "${ARCH}" == "local" ]] ; then
     fi
 fi
 
-
+# Run regtests.
 echo -e "\n========== Running Regtests =========="
 make ARCH="${ARCH}" VERSION="${VERSION}" TESTOPTS="${TESTOPTS}" test
 
--- a/tools/docker/scripts/test_regtest.sh
+++ b/tools/docker/scripts/test_regtest.sh
@@ -22,7 +22,7 @@ fi
 # Switch to stable DBCSR version if requested.
 if [ -n "${USE_STABLE_DBCSR}" ] ; then
     echo "Switching to stable DBCSR version..."
-    if ! git -C cp2k/exts/dbcsr checkout -b v2.1.0-rc16 6c52382 ; then
+    if ! git -C cp2k/exts/dbcsr checkout v2.1.0-rc16 ; then
         echo -e "\nSummary: Could not checkout stable DBCSR version."
         echo -e "Status: FAILED\n"
         exit 0
--- a/tools/docker/Dockerfile.test_pdbg
+++ b/tools/docker/Dockerfile.test_pdbg
@@ -11,6 +11,7 @@ RUN ./install_basics.sh
 COPY ./scripts/install_regtest.sh .
 RUN ./install_regtest.sh local pdbg
 
+ENV USE_STABLE_DBCSR=1
 COPY ./scripts/ci_entrypoint.sh ./scripts/test_regtest.sh ./
 CMD ["./ci_entrypoint.sh", "./test_regtest.sh", "local", "pdbg"]
 


alazzaro:
I'm not sure, what's the reason of this PR?
CP2K always gets release with tag as submodule. We will never push the develop to CP2K....

tools/docker/scripts/test_regtest.sh
```diff
@@ -19,6 +19,17 @@ if command -v ompi_info &> /dev/null ; then
     export OMPI_MCA_plm_rsh_agent=/bin/false
 fi
 
+# Switch to stable DBCSR version if requested.
+if [ -n "${USE_STABLE_DBCSR}" ] ; then
+    echo "Switching to stable DBCSR version..."
+    if ! git -C cp2k/exts/dbcsr checkout -b v2.1.0-rc16 6c52382 ; then
```
dev-zero: you should be able to simply do `git checkout v2.1.0-rc16`, no?
And why is this not an `ARG`?

dev-zero:
Can you please first completely write down how this is supposed to work and what people have to do if they have to upgrade DBCSR due to changes they added and need for further CP2K development?

tools/docker/scripts/test_regtest.sh
oschuett: I prefer to the SHA, tags can change.

I agree it's a bit hackish. A cleaner way would be to move DBCSR into the toolchain and add a `--use-stable-dbcsr` flag.

alazzaro: DBCSR in the toolchain is expected for the next year... This is will be v3.x...

dev-zero: > I prefer to the SHA, tags can change.

Can, yes, will, no. With DBCSR packaged in Spack (and others) we would break things if we would be start changing tags. Please use tags instead to make it more maintainable.

oschuett:
> Can you please first completely write down how this is supposed to work ...

We can continue to upgrade DBCSR at a high pace. And DBCSR can continue to add new APIs.

The only difference is that CP2K will not rely on a new API and thereby lock itself into a new version until it had at least 6 months of field testing.

So, it's quite simple: Try to avoid API changes. And if unavoidable, please wait 6 months for them to become available in CP2K.

Compared to our other core dependencies that is still dangerously close to the edge. In the case of #911 it actually took us almost 7 months to figure that we [upgraded](https://github.com/cp2k/cp2k/pull/886) to a bad version.

tools/docker/scripts/test_regtest.sh
oschuett: Ok, I switch to using the tag.

dev-zero:
> > Can you please first completely write down how this is supposed to work ...
> 
> We can continue to upgrade DBCSR at a high pace.

What does that mean?

> And DBCSR can continue to add new APIs.
> 
> The only difference is that CP2K will not rely on new a API and thereby lock itself into a new version until it had at least 6 month of field testing.

Ok, but how do we get the field testing without being able to use it in CP2K?

> 
> So, it's quite simple: Try to avoid API changes.

This is clear.

> And if unavoidable, please wait 6 months for them to become available in CP2K.

The problem we have here is not the addition of new APIs, but the mix of functionality changes with addition of new APIs, no?

> Compared to our other core dependencies that is still dangerously close to the edge.

How so? With libint for example we were rather quick to pick it up. But also other dependencies (SIRIUS) we pick up right away. So shouldn't this also be applied to SIRIUS, right?

> In the case of #911 it actually took us almost 7 months to figure that we [upgraded](https://github.com/cp2k/cp2k/pull/886) to a bad version.

Which we figured out only by testing against it.

oschuett:
>> We can continue to upgrade DBCSR at a high pace.

> What does that mean?

It means that you can upgrade to a new version of DBCSR as often as you like.

> Ok, but how do we get the field testing without being able to use it in CP2K?

The field testing is needed to ensure that the existing functionality, e.g. `dbcsr_multiply`, continues to work.


> Which we figured out only by testing against it.

Right, and if the tensor API hadn't changed then we would just rollback now. Instead either the DBCSR project has to scramble and put out a hotfix release or we have to partially undo Patrick's changes and return to [v2.1.0-rc12](https://github.com/cp2k/dbcsr/releases/tag/v2.1.0-rc12). Both are bad options. Hence, in the future we should make sure that we give ourselves enough time to discover problems while they are still easily resolvable.

> The problem we have here is not the addition of new APIs, but the mix of functionality changes with addition of new APIs, no?

Of course the semantics of existing APIs should not change either. My point is that we must be able to rollback to an older version of DBCSR when we discover a problem down the line.

> With libint for example we were rather quick to pick it up.

On the contrary. We're currently holding back upgrades of both [libint](https://github.com/cp2k/cp2k/pull/660) and [libxc](https://github.com/cp2k/cp2k/pull/914).

> So shouldn't this also be applied to SIRIUS, right?

Maybe when it becomes more popular. 

We have to keep in mind that DBCSR is used in almost every CP2K run. Any bug in DBCSR can potentially affect our entire user base. It therefore deserves a very high level of diligence.

mtaillefumier:
On 11/23/20 1:45 PM, Tiziano Müller wrote:
>
>         Can you please first completely write down how this is
>         supposed to work ...
>
>     We can continue to upgrade DBCSR at a high pace.
>
> What does that mean?
>
>     And DBCSR can continue to add new APIs.
>
>     The only difference is that CP2K will not rely on new a API and
>     thereby lock itself into a new version until it had at least 6
>     month of field testing.
>
> Ok, but how do we get the field testing without being able to use it
> in CP2K?
>
>     So, it's quite simple: Try to avoid API changes.
>
> This is clear.
>
>     And if unavoidable, please wait 6 months for them to become
>     available in CP2K.
>
> The problem we have here is not the addition of new APIs, but the mix
> of functionality changes with addition of new APIs, no?
>
>     Compared to our other core dependencies that is still dangerously
>     close to the edge.
>
> How so? With libint for example we were rather quick to pick it up.
> But also other dependencies (SIRIUS) we pick up right away. So
> shouldn't this also be applied to SIRIUS, right?
>
The SIRIUS API is stable and meant to be stable because it is included
in some others software as well (two at least) so we can not afford to
change API all the time. Last time we changed it (following the
recommendations of C2PK devs), I made the changes directly in CP2K. The
only thing we do is eventually add new functions *without* breaking the
others. So CP2K can work with the dev version or the stable one (modulo
internal SIRIUS dependencies).

dev-zero:
> It means that you can upgrade to a new version of DBCSR as often as you like.

But will this then not break this test?

> > Ok, but how do we get the field testing without being able to use it in CP2K?
> 
> The field testing is needed to ensure that the existing functionality, e.g. `dbcsr_multiply`, continues to work.

But for a large part of DBCSR CP2K is still the field test. Because, as you noted, almost everything of CP2K requires DBCSR. Forcing devs to obey a waiting period of 6 months until they are able to update the dependency in CP2K will only get them less coverage until then and we will possibly detect such issues only after a release, which means that then we have to put out hotfixes.
And more: devs may have already moved on after 6 months, meaning that if they haven't seen the issue due to not enough test coverage there might be nobody around to fix it.

> > Which we figured out only by testing against it.
> 
> Right, and if the tensor API hadn't changed then we would just rollback now. Instead either the DBCSR project has to scramble and put out a hotfix release or we have to partially undo Patrick's changes and return to [v2.1.0-rc12](https://github.com/cp2k/dbcsr/releases/tag/v2.1.0-rc12). Both are bad options. Hence, in the future we should make sure that we give ourselves enough time to discover problems while they are still easily resolvable.

No, putting out a hotfix release is what likely should be done if there would be a release of CP2K with the issue. But this is still on the development branch, right?
So, we found an issue in a development branch and we fix it in the development branch. I don't get why we need to make it more complicated.

> > The problem we have here is not the addition of new APIs, but the mix of functionality changes with addition of new APIs, no?
> 
> Of course the semantics of existing APIs should not change either. My point is that we must be able to rollback to an older version of DBCSR when we discover a problem down the line.

Sorry, but this is wishful thinking. Requiring that we can never change anything in DBCSR just for CP2K to be able to rollback to an arbitrary version within a development branch is not going to fly.
We already have the requirement that a release version of CP2K has to depend on a release version of DBCSR (see the release plan). For this release version of DBCSR we can easily put out hotfixes if needed.

> > With libint for example we were rather quick to pick it up.
> 
> On the contrary. We're currently holding back upgrades of both [libint](https://github.com/cp2k/cp2k/pull/660) and [libxc](https://github.com/cp2k/cp2k/pull/914).

Yeah, because they broke the build system. And the reason why the libxc PR was not merged was peer-review and discussion.

So, this PR just adds another test which is going to fail and getting ignored like others on the dashboard until a release. So, I would rather prefer to do more releases and focus on fixing issues, and, if such issues arise, try to make a standalone test case for DBCSR out of it.

oschuett:
> But will this then not break this test?

You mean, will every upgrade restart the 6 month clock? Probably yes, unless there is consensus that a particular upgrade carries very low risk. However, DBCSR uses about 4 levels of parallelism and almost every change bears a high risk of introducing race conditions.

> Forcing devs to obey a waiting period of 6 months until they are able to update the dependency in CP2K ....

Again, this is only for new APIs. According to @alazzaro there will be very few of those in the future. And changing the API twice per year is still a lot of churn.

> But for a large part of DBCSR CP2K is still the field test. 

I'm sure the DBCSR project is resourceful enough to conduct thorough tests on their own.

> So, this PR just adds another test which is going to fail and getting ignored like others on the dashboard until a release. 

Nope because I'll be watching it :-)

> But this is still on the development branch, right?

True, it's good that we caught this before the release. Still, the CP2K head is broken and there is no easy way of fixing it.

I still think that this will benefit both projects. CP2K doesn't get locked into broken versions and DBCSR won't end up in the hot seat.

dev-zero:
> > But will this then not break this test?
> 
> You mean, will every upgrade restart the 6 month clock? Probably yes, unless there is consensus that a particular upgrade carries very low risk. However, DBCSR uses about 4 levels of parallelism and almost every change bears a high risk of introducing race conditions.

So, this prior statement is actually not true?

> It means that you can upgrade to a new version of DBCSR as often as you like.


> > Forcing devs to obey a waiting period of 6 months until they are able to update the dependency in CP2K ....
> 
> Again, this is only for new APIs. According to @alazzaro there will be very few of those in the future. And changing the API twice per year is still a lot of churn.
> 
> > But for a large part of DBCSR CP2K is still the field test.
> 
> I'm sure the DBCSR project is resourceful enough to conduct thorough tests on their own.

We have a code coverage of 61% (probably higher since the GPU tests are not yet added to the board), actual help is always appreciated.

> > So, this PR just adds another test which is going to fail and getting ignored like others on the dashboard until a release.
> 
> Nope because I'll be watching it :-)

Good, because I was wondering about the other 2 of this here:

20201123_16h30m34s_grim

> > But this is still on the development branch, right?
> 
> True, it's good that we caught this before the release. Still, the CP2K head is broken and there is no easy way of fixing it.

There are releases of CP2K with known defects and inaccuracies and easy backports, yet nobody bothered to make releases for that. But a break of the development branch is such a problem, I have problems following that logic.

> I still think that this will benefit both projects. CP2K doesn't get locked into broken versions and DBCSR won't end up in the hot seat.

Every time there is a new API (or a new dependency for that matter) there is a risk for a "lock-in" (where lock-in here means you can't downgrade without losing something) DBCSR could have done better, by making an intermediate release which would possibly allow for an easier backport now, sure. On the other hand, the likelihood for something like this not having been caught due to insufficient testing prior to, then merged shortly before a release only to end up in the field is much higher by introducing such artificial barriers. This problem already has a solution: an extensive CI and long release cycles.

Plus, you might want to wait with this change until v2.1.0. I wouldn't call a `-rc` a stable version (also something DBCSR can improve).

oschuett:
> So, this prior statement is actually not true?

You can upgrade as often as you like without changing the API - or change the API twice per year.

> Good, because I was wondering about the other 2 of this here:

Classic [whataboutism](https://en.wikipedia.org/wiki/Whataboutism) :-/

> But a break of the development branch is such a problem, I have problems following that logic.

The problem is not that our development branch broke. The problem is that our usual strategy of detecting and fixing bugs quickly has failed.

This particular bug was [introduced](https://github.com/cp2k/dbcsr/commit/0922e21dd3c027ac32ff3ff9e3659547e79f6706) to DBCSR over two years ago. It reached CP2K through an upgrade that contained also [75 other commits](https://github.com/cp2k/dbcsr/compare/e4579de89417c79d9d76be781c4018f5476e6d71...f385868eec7c89f40ac9d6d9783fc34a05d2f68e) and which was part of a [larger PR](https://github.com/cp2k/cp2k/pull/886/commits). This is why it took us 7 months to track it down and why we're now having difficulties to fix it.

Since it takes us longer to detect bugs in DBCSR we also have to retain the ability to rollback for longer.

dev-zero:
> > So, this prior statement is actually not true?
> 
> You can upgrade as often as you like without changing the API - or change the API twice per year.

No, also changes to the behaviour of DBCSR can trigger that (now one can start discussing what constitutes an API, of course).

> > Good, because I was wondering about the other 2 of this here:
> 
> Classic [whataboutism](https://en.wikipedia.org/wiki/Whataboutism) :-/

Demonstrating that things don't get fixed by just monitoring them.

> > But a break of the development branch is such a problem, I have problems following that logic.
> 
> The problem is not that our development branch broke. The problem is that our usual strategy of detecting and fixing bugs quickly has failed.
> 
> This particular bug was [introduced](https://github.com/cp2k/dbcsr/commit/0922e21dd3c027ac32ff3ff9e3659547e79f6706) to DBCSR over two years ago. It reached CP2K through an upgrade that contained also [75 other commits](https://github.com/cp2k/dbcsr/compare/e4579de89417c79d9d76be781c4018f5476e6d71...f385868eec7c89f40ac9d6d9783fc34a05d2f68e) and which was part of a [larger PR](https://github.com/cp2k/cp2k/pull/886/commits). This is why it took us 7 months to track it down and why we're now having difficulties to fix it.
> 
> Since it takes us longer to detect bugs in DBCSR we also have to retain the ability to rollback for longer.

At the same time one can argue that to detect bugs in DBCSR one has to test more/longer.

An important change merged half a year in the future is worse than changes merged now, since the people who have worked on those have possibly moved on by then and may not be around to help fixing the issues.

oschuett:
Ok, I think all the main arguments have been made. I'll merge this PR now. It will alert us when we're about to lock ourselves in again. At that point we'll have a concrete API change and can continue the discussion.

PR #1190, opened 2020-11-23
NEGF: add new print options

dryndyk:
New sections
NEGF%PRINT%PROGRAM_RUN_INFO
NEGF%PRINT%PROGRAM_RUN_DETAILS
for control of printing levels.
Now "silent" is silent, etc.

CPABORT => CPWARN in src/qs_moments.F:1347,
it blocked print level HIGH for some structures.

### DIFF
--- a/src/negf_methods.F
+++ b/src/negf_methods.F
@@ -216,7 +216,7 @@ CONTAINS
       CALL negf_sub_env_create(sub_env, negf_control, blacs_env, global_env%blacs_grid_layout, global_env%blacs_repeatable)
       CALL negf_env_create(negf_env, sub_env, negf_control, force_env, negf_mixing_section, log_unit)
 
-      IF (log_unit > 0 .AND. print2) THEN
+      IF (log_unit > 0 .AND. verbose_output) THEN
          DO icontact = 1, SIZE(negf_control%contacts)
             WRITE (log_unit, "(/,' NEGF| Atoms in the contact region',I2,':',I4)") &
                icontact, SIZE(negf_control%contacts(icontact)%atomlist_bulk)
--- a/src/negf_methods.F
+++ b/src/negf_methods.F
@@ -37,10 +37,12 @@ MODULE negf_methods
    USE cp_log_handling,                 ONLY: cp_get_default_logger,&
                                               cp_logger_get_default_io_unit,&
                                               cp_logger_type
-   USE cp_output_handling,              ONLY: &
-        cp_p_file, cp_print_key_finished_output, cp_print_key_should_output, cp_print_key_unit_nr, &
-        debug_print_level, high_print_level, low_print_level, medium_print_level, &
-        silent_print_level
+   USE cp_output_handling,              ONLY: cp_p_file,&
+                                              cp_print_key_finished_output,&
+                                              cp_print_key_should_output,&
+                                              cp_print_key_unit_nr,&
+                                              debug_print_level,&
+                                              high_print_level
    USE cp_para_types,                   ONLY: cp_para_env_type
    USE cp_subsys_types,                 ONLY: cp_subsys_type
    USE dbcsr_api,                       ONLY: dbcsr_copy,&
--- a/src/negf_methods.F
+++ b/src/negf_methods.F
@@ -37,10 +37,10 @@ MODULE negf_methods
    USE cp_log_handling,                 ONLY: cp_get_default_logger,&
                                               cp_logger_get_default_io_unit,&
                                               cp_logger_type
-   USE cp_output_handling,              ONLY: cp_p_file,&
-                                              cp_print_key_finished_output,&
-                                              cp_print_key_should_output,&
-                                              cp_print_key_unit_nr
+   USE cp_output_handling,              ONLY: &
+        cp_p_file, cp_print_key_finished_output, cp_print_key_should_output, cp_print_key_unit_nr, &
+        debug_print_level, high_print_level, low_print_level, medium_print_level, &
+        silent_print_level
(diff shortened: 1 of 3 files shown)


dev-zero:
I think we should rather control verbosity through the section parameter of a print section, rather than extra print section keywords, and keep the print keywords for controlling the printing of concrete additional properties.

Example:

https://github.com/cp2k/cp2k/blob/72339f3ab4a5af43150d59429dc2c871dcab1268/src/qs_active_space_methods.F#L967-L986

dryndyk:
Yes, I agree. It looks better. I will rewrite it. Thanks.

dryndyk:
I have a question. As far as I see, PRINT_LEVEL is used only twice in the code + GLOBAL%PRINT_LEVEL. 
PROGRAM_RUN_INFO is used everywhere.

dev-zero:
I would probably keep the `PROGRAM_RUN_INFO` and get the section parameter describing the level from there

dev-zero:
such that one can do:

```
&NEGF
  &PRINT
    &PROGRAM_RUN_INFO
    &END
  &END
&END
```

for the default, and

```
&NEGF
  &PRINT
    &PROGRAM_RUN_INFO HIGH
    &END
  &END
&END
```

for the more verbose one

dev-zero:
... on the other hand we also have similar flags like in this PR in xtb, dftb and tddpt2 (`DETAILED_ENERGY`), but also explicit `VERBOSITY` flags in transport and linear scaling. We should probably document a standard and harmonize this at some point.

dryndyk:
OK. I will make one iteration to improve and later we can see.

dryndyk:
> such that one can do:
> 
> ```
> &NEGF
>   &PRINT
>     &PROGRAM_RUN_INFO
>     &END
>   &END
> &END
> ```
> 
> for the default, and
> 
> ```
> &NEGF
>   &PRINT
>     &PROGRAM_RUN_INFO HIGH
>     &END
>   &END
> &END
> ```
> 
> for the more verbose one

No, the logic of PROGRAM_RUN_INFO is the opposite. If HIGH it will print something only when GLOBAL%PRINT_LEVEL is HIGH. And it is OK to be consistent with other output. But I want to have the possibility to print more only for NEGF. So, we need another keyword. I think PRINT_LEVEL inside  PROGRAM_RUN_INFO to overwrite  GLOBAL%PRINT_LEVEL with GLOBAL%PRINT_LEVEL as default.

dev-zero:
Yeah, you are right. Well, then let's leave it at this?

dryndyk:
Wait, I will try to add  PROGRAM_RUN_INFO%PRINT_LEVEL instead of PROGRAM_RUN_DETAILS.
Will be better.

dryndyk:
So, is it OK now?

dev-zero:
> So, is it OK now?

LGTM except that `debug1` is not used, right?

dryndyk:
What do you mean by debug1?

src/negf_methods.F
```diff
@@ -194,9 +195,51 @@ SUBROUTINE do_negf(force_env)
...
+            print2 = .FALSE.
+         CASE (high_print_level)
+            print1 = .TRUE.
+            print2 = .TRUE.
+         CASE (debug_print_level)
+            print1 = .TRUE.
+            print2 = .TRUE.
+         CASE DEFAULT
+            !do nothing
+         END SELECT
```
dev-zero:
```suggestion
         SELECT CASE (print_level)
         CASE (high_print_level, debug_print_level)
            verbose_output = .TRUE.
         CASE DEFAULT
            verbose_output = .FALSE.
         END SELECT
```

src/negf_methods.F
```diff
@@ -194,9 +195,51 @@ SUBROUTINE do_negf(force_env)
...
+      END IF
+
+      IF (log_unit > 0) THEN
+         WRITE (log_unit, '(/,T2,A,T62)') "COMPUTE THE RELEVANT HAMILTONIAN MATRICES"
+      END IF
+
       CALL negf_sub_env_create(sub_env, negf_control, blacs_env, global_env%blacs_grid_layout, global_env%blacs_repeatable)
       CALL negf_env_create(negf_env, sub_env, negf_control, force_env, negf_mixing_section, log_unit)
 
+      IF (log_unit > 0 .AND. print2) THEN
```
dev-zero:
```suggestion
      IF (log_unit > 0 .AND. verbose_output) THEN
```

src/negf_methods.F
```diff
@@ -37,10 +37,10 @@ MODULE negf_methods
...
                                               cp_logger_get_default_io_unit,&
                                               cp_logger_type
-   USE cp_output_handling,              ONLY: cp_p_file,&
-                                              cp_print_key_finished_output,&
-                                              cp_print_key_should_output,&
-                                              cp_print_key_unit_nr
+   USE cp_output_handling,              ONLY: &
+        cp_p_file, cp_print_key_finished_output, cp_print_key_should_output, cp_print_key_unit_nr, &
+        debug_print_level, high_print_level, low_print_level, medium_print_level, &
+        silent_print_level
```
dev-zero:
```suggestion
   USE cp_output_handling,              ONLY: cp_p_file,&
                                              cp_print_key_finished_output,&
                                              cp_print_key_should_output,&
                                              cp_print_key_unit_nr,&
                                              debug_print_level,&
                                              high_print_level
```

dev-zero (commented):
sorry, that's what I meant. You should be able to directly accept the changes if you agree.

dryndyk:
OK. Looks good!

Issue #12, opened 2018-10-03
CP2K manual: set default version to 6.1 and replace 'SVN Trunk' reference


dev-zero:
maybe something like `development version` instead?

oschuett:
`Development Version` is pretty long. How about just `Development` or `Latest` as a label?

dev-zero:
`Development` is fine I guess. `Latest` could also be *latest release*

oschuett:
Fixed via #17

Issue #10, opened 2018-10-03
Update cp2k.org Subversion references and instructions to Git


dev-zero:
- [x] https://www.cp2k.org/download
- [x] https://www.cp2k.org/dev:release_checklist ... might have to change considerably, depending on availability of CI infrastructure to automatically build tarballs
- [x] https://www.cp2k.org/dev:regtesting
- [x] https://www.cp2k.org/dev:starting
- [x] https://www.cp2k.org/performance
- [x] https://www.cp2k.org/dev
- [ ] https://www.cp2k.org/news
- [x] https://www.cp2k.org/howto:install_with_plumed
- [x] https://www.cp2k.org/howto:cp2k_omen
- [x] https://www.cp2k.org/tools:cubecruncher
- [x] https://www.cp2k.org/admin:kohn ... rename to `sham`?
- [x] https://www.cp2k.org/sidebar
- [x] https://www.cp2k.org/about
- [x] https://www.cp2k.org/dbcsr
- [x] https://www.cp2k.org/tools:emacs
- [ ] https://www.cp2k.org/version_history

dev-zero:
@oschuett can you please update the documentation on `admin:kohn`?

oschuett:
I created https://www.cp2k.org/static/download_stats.html.

oschuett:
I remove the https://www.cp2k.org/admin:kohn page. Since I switched to [Ansible](https://github.com/cp2k/cp2k-infra) it's not needed anymore.

PR #1195, opened 2020-11-27
Space group symmetry

### LINKED ISSUE #562: Verify recent versions of spglib with CP2K
Current master of CP2K has the following check:

https://github.com/cp2k/cp2k/blob/5566d17405ab6f351561817a83675b3344ac2a5a/src/cryssym.F#L168-L170

with spglib-1.9.9 being released in 2016 and the toolchain shipping 1.12.2.

Counterintuitively I get this warning even with `FORCE_EVAL/DFT/KPOINTS/SYMMETRY` set to `.FALSE.`

Do we have a test set for spglib-based functionality?

pcazade:
Dear CP2K Developers,

I introduced space group symmetry and its conservation during geometry and cell optimization. It works for all FORCE_EVAL methods (FIST and QS), as well as core/shell systems. I would like to make this feature to be available in standard CP2K.

When requested, the symmetry is automatically detected with the help of Spglib. Only the coordinates, the forces, and the stress tensor are symmetrized according to the space group symmetry operations. So far, this is not used to reduce the computational cost of the energy evaluation.

Two new files were introduced: src/motion/space_groups.F and src/motion/space_groups_types.F

The files listed below were modified:

-src/start/input_cp2k_motion.F
-src/motion/gopt_f_types.F
-src/motion/gopt_f77_methods.F
-src/motion/gopt_f77_methods.f90
-src/motion/bfgs_optimizer.F
-src/motion/cg_optimizer.F
-src/motion/cp_lbfgs.F
-src/motion/cp_lbfgs_geo.F
-src/motion/cp_lbfgs_optimizer_gopt.F

Regards,
Pierre


oschuett:
@pcazade thanks for this PR!

It seems you are using a pretty old branch. Can you please run the following commands and resolve any arising merge conflicts:
```
git remote add upstream https://github.com/cp2k/cp2k.git
git fetch upstream
git rebase upstream/master
```

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+      spgr%nparticle_sym = spgr%nparticle
+      spgr%n_atom_sym = spgr%n_atom
+      spgr%n_core_sym = spgr%n_core
+      spgr%n_shell_sym = spgr%n_shell
+
+      ! output unit
+      IF (PRESENT(iounit)) THEN
+         spgr%punit = iounit
+      ELSE
+         spgr%punit = 6
```
dev-zero: hardcoded print units are difficult to maintain, you can use `get_unit_number` or `open_file` from the `cp_files` module instead to get dynamic consistent unit numbers (or files)

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+         spgr%eps_geo = eps_geo
+      ELSE
+         spgr%eps_geo = 1.e-4_dp
+      END IF
+
+      ! accuracy for symmetry
+      IF (PRESENT(eps_sym)) THEN
+         spgr%eps_sym = eps_sym
+      ELSE
+         spgr%eps_sym = 1.e-4_dp
```
dev-zero: I wonder: since this is the creation method you could assume that the `spgr_type` is uninitialized and forgo the repetition of the defaults since you already have them specified for the type above (or the other way round: remove the defaults from the type and exclusively set them here, although I would prefer the defaults on the type)

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+      CASE (default_minimization_method_id, default_ts_method_id)
+         CALL init_cell(spgr%cell_ref, hmat=cell%hmat)
+      CASE (default_cell_method_id)
+         SELECT CASE (gopt_env%cell_method_id)
+         CASE (default_cell_direct_id)
+            CALL init_cell(spgr%cell_ref, hmat=gopt_env%h_ref)
+         CASE (default_cell_geo_opt_id, default_cell_md_id)
+            CPABORT("SPACE_GROUP_SYMMETRY SHOULD NOT BE INVOKED DURING THE CELL STEP.")
+         CASE DEFAULT
+            CPABORT("")
```
dev-zero: no blank `CPABORT` if possible, since they always provide some sort of documentation as well

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+      ALLOCATE (spgr%atype(spgr%nparticle))
+      spgr%atype(1:spgr%nparticle) = types(1:spgr%nparticle)
+
+      spgr%n_operations = 0
+
+      !..try spglib
+      major = spg_get_major_version()
+      minor = spg_get_minor_version()
+      micro = spg_get_micro_version()
+      IF (major == 0) THEN
```
dev-zero: since this can't be dynamic you might instead use `#ifdef __SPGLIB` instead

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+         CALL cp_warn(__LOCATION__, "Symmetry library SPGLIB not available")
```
dev-zero: maybe a `CPABORT` is better at this point: if I followed the flow correctly one should only end up here if the user specifically requested it, hence a `CPABORT` might be more inline with the rest of the code where we usually error out when the user requests an unavailable feature?

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+         spglib = .FALSE.
+      ELSE
+         !IF (major /= 1 .OR. minor /= 9 .OR. micro /= 9) THEN
+         !   CALL cp_warn(__LOCATION__, "Version of Symmetry Library SPGLIB not tested")
+         !END IF
```
dev-zero: please remove

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+         spglib = .TRUE.
+         CALL cite_reference(Togo2018)
+         spgr%international_symbol = '           '
+         ierr = spg_get_international(spgr%international_symbol, TRANSPOSE(cell%hmat), tmp_coor, tmp_types, &
+                                      spgr%n_atom_sym, eps_sym)
+         IF (ierr == 0) THEN
+            CALL cp_warn(__LOCATION__, "Symmetry Library SPGLIB failed")
```
dev-zero: can we tell the user more about what went wrong here?

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+!> \par History
+!>      01.2020 created [pcazade]
+!> \author Pierre-André Cazade (first version)
+! **************************************************************************************************
+   SUBROUTINE spgr_gen(spgr, scoor, types, cell, gopt_env, eps_geo, eps_sym, pol, ranges, &
+                       nparticle, n_atom, n_core, n_shell, iounit)
+      
+      IMPLICIT NONE
+      
+      TYPE(spgr_type)                                    :: spgr
```
dev-zero: `INTENT` missing for most dummy arguments (same for other places)

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+      nparticle = 0
+      n_atom = 0
+      n_core = 0
+      n_shell = 0
+
+      NULLIFY (particles)
+      NULLIFY (core_particles)
+      NULLIFY (shell_particles)
+
+      particles => subsys%particles
```
dev-zero: please use `cp_subsys_get(...)` instead of directly accessing the members

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+      ! Check if we have other kinds of particles in this subsystem
+      IF (ASSOCIATED(shell_particles)) THEN
+         n_shell = shell_particles%n_els
+         CPASSERT(ASSOCIATED(core_particles))
+         n_core = subsys%core_particles%n_els
+         ! The same number of shell and core particles is assumed
+         CPASSERT(n_core == n_shell)
+      ELSE IF (ASSOCIATED(core_particles)) THEN
+         ! This case should not occur at the moment
+         CPASSERT(ASSOCIATED(shell_particles))
```
dev-zero: does this assertion trigger at all? If `ASSOCIATED(shell_particles)` is `.TRUE.` we would be in the first `IF` branch, right?

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+
+      IF ( .NOT.( spgr%nparticle == (natom + nshell) ) ) THEN
+         CPABORT("spgr_find_equivalent_atoms: nparticle not equal to natom + nshell.")
+      END IF
+
+      DO ia = 1, spgr%nparticle
+         spgr%eqatom(:,ia) = ia
+      END DO
+
+      !$OMP PARALLEL DO PRIVATE (ia,ib,ir,ri,rb,ro,rot,tr,diff) SHARED (spgr,scoord,natom,nop)
```
dev-zero: please also set `DEFAULT(NONE)` for `PARALLEL DO`s

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+            DO ib = 1, natom
+               IF (.NOT.spgr%lat(ib)) CYCLE
+               rb(1:3) = scoord(1:3, ib)
+               ro(1) = REAL(rot(1, 1), dp)*rb(1)+REAL(rot(2, 1), dp)*rb(2)+REAL(rot(3, 1), dp)*rb(3)+tr(1)
+               ro(2) = REAL(rot(1, 2), dp)*rb(1)+REAL(rot(2, 2), dp)*rb(2)+REAL(rot(3, 2), dp)*rb(3)+tr(2)
+               ro(3) = REAL(rot(1, 3), dp)*rb(1)+REAL(rot(2, 3), dp)*rb(2)+REAL(rot(3, 3), dp)*rb(3)+tr(3)
+               ro(1) = ro(1) - REAL(NINT(ro(1)-ri(1)))
+               ro(2) = ro(2) - REAL(NINT(ro(2)-ri(2)))
+               ro(3) = ro(3) - REAL(NINT(ro(3)-ri(3)))
+               diff = SQRT(SUM((ri(:)-ro(:))**2))
```
dev-zero: `NORM2(ri - ro)` instead?

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+         routineP = moduleN//':'//routineN
+
+      INTEGER                                            :: ia, ib, ir, nparticle, nop, nops, handle, &
+                                                            ja, jb
+      REAL(KIND=dp), DIMENSION(3)                        :: ri, ro, tr, rf, rn
+      REAL(KIND=dp), DIMENSION(3, 3)                     :: rot
+      REAL(KIND=dp), DIMENSION(:)                        :: coord
+      REAL(KIND=dp), DIMENSION(:), ALLOCATABLE           :: cold
+
+      INTEGER, EXTERNAL :: OMP_GET_THREAD_NUM, OMP_GET_NUM_THREADS
```
dev-zero: leftovers?

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+      CHARACTER(LEN=*), PARAMETER :: routineN = 'print_spgr', &
+         routineP = moduleN//':'//routineN
+
+      INTEGER                                            :: i, iunit, j, plevel, handle
+
+      CALL timeset(routineN,handle)
+
+      iunit = spgr%punit
+      IF (iunit >= 0) THEN
+         plevel = spgr%plevel+5
```
dev-zero: maybe add a `print_key` instead?

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+     WRITE (UNIT=spgr%punit, FMT=fmtstr2) "X", stress_tensor(1, 1:3)
+     WRITE (UNIT=spgr%punit, FMT=fmtstr2) "Y", stress_tensor(2, 1:3)
+     WRITE (UNIT=spgr%punit, FMT=fmtstr2) "Z", stress_tensor(3, 1:3)
+
+     CALL timestop(handle)
+
+   END SUBROUTINE spgr_write_stress_tensor
+
+! **************************************************************************************************
+!> \brief routine prints cell details. Only for debugging purpose.
```
dev-zero: but it's still used in a non-debugging way?

src/motion/space_groups.F
```diff
@@ -0,0 +1,1229 @@
...
+! **************************************************************************************************
+!> \brief routine prints hessian matrix. Only for debugging purpose.
+!> \param hess ...
+!> \param iter ...
+!> \param io output unit
+!> \par History
+!>      01.2020 created [pcazade]
+!> \author Pierre-André Cazade (first version)
+! **************************************************************************************************
+   SUBROUTINE spgr_print_hess(hess,iter,io)
```
dev-zero: please remove if not used, or hide them (with proper format strings) behind a print key if you think they could be useful for others

dev-zero (changes requested):
Nice work! I looked through the code and found a couple of small things which may have to be changed besides the required `git rebase` and running the prettifier.
I haven't checked the branching point, but if you're running into a lot of conflicts due to whitespace changes or the dreaded `could not detach HEAD` error, the following post may be of use: https://github.com/orgs/cp2k/teams/cp2k-developers/discussions/8

pcazade:
Dear Ole and Tiziano,

Thanks for the comments. I will work on the modifications you suggested. I
will also check the discussion you linked as I have issues with the rebase
procedure.

I realized I left a lot of print/write that were for debugging purposes. I
don't think any of the print_cell or print_hess would be useful for a user,
only for a developer. Do you recommend removing these routines or to hide
them behind a keyword?

Regarding making spgr_type part of geo_f_type, I considered something like
this, though necessarily with the exact strcucture type, at some point of
the development but I gave up because of issues in accessing other modules
and/or structure types. I will check again in particular with  geo_f_type.

Regards,
Pierre

pcazade:
Dear Tiziano,

The link you shared does not work. Would mind sending another one?

In my last commit, I think I addressed most, if not all, of the issues you noticed. I only have to do the merge/rebase which fails because of a commit earlier this year. Would you be able to suggest a solution to this.

Regards,
Pierre

pcazade:
Dear CP2K developers,

I successfully applied git rebase master to my branch. I then used git push --force as described in the guidelines.

Regards,
Pierre

dev-zero:
@pcazade can you remove the commit dbc683e ? If it is the latest commit simply do `git reset --hard HEAD~1`, otherwise do an interactive rebase using `git rebase -i upstream/master` and then switch the `pick` to a `drop`.

The following errors must be fixed though I'm afraid:

```
-D__COMPILE_REVISION="\"git:93b5cbf\"" -D__DATA_DIR="\"/workspace/cp2k/data\"" -D__SHORT_FILE__="\"minbas_methods.F\"" -I'/workspace/cp2k/src/' -I'/workspace/cp2k/obj/local/sdbg/exts/dbcsr' minbas_methods.F90 
/workspace/cp2k/src/motion/space_groups.F:780:6:

  780 |       fold=force
      |      1
/workspace/cp2k/src/motion/space_groups.F:717:6:

  717 |       cold=coord
      |      1
/workspace/cp2k/src/motion/space_groups.F:597:6:

  597 |       xold=x
      |      1
```

If the LHS is already allocated and has the correct size you can use `xold(:) = x(:)` to prevent the compiler from generating reallocation code. If it needs to be reallocated do it explicitly by a `DEALLOCATE` beforehand.

Also, please run `make pretty` (or `make pre-commit`) and add the suggested changes as another commit (we will then squash all commits from this PR into one when merging).

src/motion/cg_optimizer.F
```diff
@@ -198,14 +290,33 @@ RECURSIVE SUBROUTINE cp_cg_main(force_env, x0, gopt_param, output_unit, globenv,
...
+         END IF
+
          ! Get Conjugate Directions:  updates the searching direction (h)
          wildcard = "   CG"
          CALL get_conjugate_direction(gopt_env, Fletcher_Reeves, g, xi, h)
 
+         ! Symmetrize coordinates and forces
+         IF (keep_sg) THEN
+            CALL spgr_apply_rotations_force(spgr, g)
+            CALL spgr_apply_rotations_force(spgr, xi)
```
dev-zero: Why is there another `CALL spgr_apply_rotations_force(spgr, xi)` needed here?

src/qs_efield_berry.F
```diff
@@ -248,9 +248,9 @@ SUBROUTINE qs_efield_derivatives(qs_env, just_energy, calculate_forces)
       use_virial = virial%pv_availability .AND. (.NOT. virial%pv_numer)
       use_virial = use_virial .AND. calculate_forces
       ! disable stress calculation
-      IF (use_virial) THEN
-         CPABORT("Stress tensor for periodic E-field not implemented")
-      END IF
+      !IF (use_virial) THEN ! [pcazade] enabled the stress tensor
+      !   CPABORT("Stress tensor for periodic E-field not implemented")
+      !END IF
```
dev-zero: ?

pcazade: Reading the code it seems that the stress tensor for periodic E-field is actually implemented. Juerg told me that it is disabled because not sufficiently tested. I needed this information. It has nothing to do with space group symmetry though. Should I put it back?

dev-zero: Either that, or switch the `CPABORT` to a `CPWARN` with an appropriate message, @juerghutter what do you think?

juerghutter:
I would suggest to leave the CPABORT. It should only be removed
after testing and adding regression tests.

Universität Zürich                   E-mail: hutter@chem.uzh.ch
Winterthurerstrasse 190
CH-8057 Zürich, Switzerland

-----"pcazade" <notifications@github.com> wrote: -----
To: "cp2k/cp2k" <cp2k@noreply.github.com>

pcazade:
I clearly did something wrong with exts/dbcsr. After I did the git rebase master, git status was telling me that exts/dbcsr was modified and I most likely did a git rm exts/dbcsr. What should I do to solve this issue? Simply add again exts/dbcsr, although it's empty or download dbcsr from the master, put its content in my own exts/dbcsr and then git add the whole directory.

alazzaro:
You can "reset" submodules via:

`git submodule update --init --recursive`

pcazade:
Thanks. Should I perform a git add exts/dbcsr and make it part of a new commit?

pcazade:
I performed the git submodule update --init --recursive. Now when I execute git status, I get this:

On branch space_group_symmetry
Your branch is up to date with 'origin/space_group_symmetry'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)
  (commit or discard the untracked or modified content in submodules)

	modified:   exts/dbcsr (modified content)

no changes added to commit (use "git add" and/or "git commit -a")

No git add nor git add --all seem to be able to address this issue. It looks to me as if the dbcsr directory belonged to a different git repository.

I am sorry if my questions are silly but I am not a git expert. I would like to know what are the different steps that I need to do to address the issue that create the file conflict in exts/dbcsr and prevents the merge to happen.

Thanks,
Pierre

dev-zero:
@pcazade I will follow up next week, sorry, busy atm.

pcazade:
@dev-zero no bother, we are all there.

dev-zero:
@pcazade ok, here are the instructions on how to rebase without the change to `exts/dbcsr`.
I assume that in your working copy of the repository you have a remote `origin` pointing to your repository and a remote `upstream` pointing to the cp2k repository. Please verify that it looks like this (or adapt instructions as needed, note: https:// URLs also work):

```console
$ git remote -v
origin	git@github.com:pcazade/cp2k.git (fetch)
origin	git@github.com:pcazade/cp2k.git (push)
upstream	git@github.com:cp2k/cp2k.git (fetch)
upstream	git@github.com:cp2k/cp2k.git (push)
```

then do:

```console
$ git fetch upstream
$ git rebase upstream/master
Failed to merge submodule exts/dbcsr (commits don't follow merge-base)
Auto-merging exts/dbcsr
CONFLICT (submodule): Merge conflict in exts/dbcsr
error: could not apply f67098838e... Modifications to allow compilation.
Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".
Could not apply f67098838e... Modifications to allow compilation.
$ git restore --staged exts/dbcsr
$ git rebase --continue
$ git push --force
```

should you be using an old version of git where the `git restore` command doesn't exist, the instructions are slightly different (starting from `git restore`):

```
$ git fetch upstream
$ git rebase upstream/master
First, rewinding head to replay your work on top of it...
Applying: add Space Group Symmetry
Applying: Extend SPGR to CG optimizer and all CELL_OPT.
Applying: Fix symmetrization of the positions and forces to be in agreement with the stress tensor.
Applying: Space Group Symmetry applied to core-shell systems
Applying: SPGR core/shell bug fixed
Applying: Implementations of required modifications for patch
Applying: Modifications to allow compilation.
Using index info to reconstruct a base tree...
M	exts/dbcsr
<stdin>:37: trailing whitespace.
      
<stdin>:39: trailing whitespace.
      
<stdin>:58: trailing whitespace.
      
warning: 3 lines add whitespace errors.
Falling back to patching base and 3-way merge...
warning: Failed to merge submodule exts/dbcsr (commits don't follow merge-base)
Auto-merging exts/dbcsr
CONFLICT (submodule): Merge conflict in exts/dbcsr
Failed to merge in the changes.
Patch failed at 0007 Modifications to allow compilation.
The copy of the patch that failed is found in:
   /data/tiziano/cp2k/.git/rebase-apply/patch

When you have resolved this problem, run "git rebase --continue".
If you prefer to skip this patch, run "git rebase --skip" instead.
To check out the original branch and stop rebasing, run "git rebase --abort".
$ git reset HEAD exts/dbcsr  # unstage
$ rm -rf exts/dbcsr/tools/fprettify # manually remove new submodule
$ git submodule update --init --recursive  # reset state for all submodules to original
$ git rebase --continue
$ git push --force
```

pcazade:
@dev-zero

Hi Tiziano,

The commands you shared worked. The next issue now if with the regtest sdbg. It gives the error below which I am confused by. I checked the files, in question in case a problem occurred during the merge/rebase but everything seems to be defined as expected.

D__SHORT_FILE__="\"motion/bfgs_optimizer.F\"" -I'/workspace/cp2k/src/motion/' -I'/workspace/cp2k/obj/local/sdbg/exts/dbcsr' bfgs_optimizer.F90 
/workspace/cp2k/src/motion/cg_optimizer.F:226:68:

  226 |                          .FALSE., gopt_env%force_env%para_env, spgr=spgr)
      |                                                                    1
Error: Keyword argument 'spgr' at (1) is not in the procedure
/workspace/cp2k/src/motion/cg_optimizer.F:295:71:

  295 |                             .FALSE., gopt_env%force_env%para_env, spgr=spgr)
      |                                                                       1
Error: Keyword argument 'spgr' at (1) is not in the procedure
/workspace/cp2k/src/motion/cp_lbfgs_optimizer_gopt.F:686:90:

  686 |                                master=optimizer%master, para_env=optimizer%para_env, spgr=spgr)
      |                                                                                          1
Error: Keyword argument 'spgr' at (1) is not in the procedure
make[3]: *** [/workspace/cp2k/Makefile:504: cg_optimizer.o] Error 1
make[3]: *** Waiting for unfinished jobs....
make[3]: *** [/workspace/cp2k/Makefile:504: cp_lbfgs_optimizer_gopt.o] Error 1
/workspace/cp2k/src/motion/bfgs_optimizer.F:292:68:

  292 |                          .FALSE., gopt_env%force_env%para_env, spgr=spgr)
      |                                                                    1
Error: Keyword argument 'spgr' at (1) is not in the procedure
/workspace/cp2k/src/motion/bfgs_optimizer.F:406:71:

  406 |                             .FALSE., gopt_env%force_env%para_env, spgr=spgr)
      |                                                                       1
Error: Keyword argument 'spgr' at (1) is not in the procedure
make[3]: *** [/workspace/cp2k/Makefile:504: bfgs_optimizer.o] Error 1
make[2]: *** [/workspace/cp2k/Makefile:144: all] Error 2
make[1]: *** [/workspace/cp2k/Makefile:126: sdbg] Error 2
make: *** [Makefile:121: all] Error 2

Summary: Compilation failed.
Status: FAILED

dev-zero:
The `cp_eval_at` are not fully resolved at compilation time since those are MODULE-less F77-compatible functions. But since `cp_eval_at` is an F77-function it means that `OPTIONAL` is not allowed in its declaration. So, we probably have to resort to passing along the `spgr_type` via a pointer of an already allocated structure (`gopt_f_type`, again) or pass it to `cp_eval_at` as non-optional pointer and then in `cp_eval_at` (or wherever it is used) check whether it is passed by checking the pointer state (`ASSOCIATED(spgr)`) instead of the argument being present.

pcazade:
Thanks Tiziano. What puzzles me is that It compiles for me without any problem. Furthermore, in cp_eval_at, it is not the only argument passed as OPTIONAL:

REAL(KIND=dp), DIMENSION(:), OPTIONAL, &
      POINTER                                :: gradient
   INTEGER, INTENT(IN)                      :: master
   LOGICAL, INTENT(IN), OPTIONAL            :: final_evaluation

I will check if it works by not making OPTIONAL and testing with ASSOCIATED instead, as you suggested.

dev-zero:
> Thanks Tiziano. What puzzles me is that It compiles for me without any problem. Furthermore, in cp_eval_at, it is not the only argument passed as OPTIONAL:
> 
> REAL(KIND=dp), DIMENSION(:), OPTIONAL, &
> POINTER :: gradient
> INTEGER, INTENT(IN) :: master
> LOGICAL, INTENT(IN), OPTIONAL :: final_evaluation
> 
> I will check if it works by not making OPTIONAL and testing with ASSOCIATED instead, as you suggested.

sorry, I found the issue: there is actually a forward declaration/an interface in `src/motion/gopt_f77_methods.f90` which needs to be augmented

dev-zero:
The following patch makes it build:

```
diff --git a/src/motion/gopt_f77_methods.f90 b/src/motion/gopt_f77_methods.f90
index 5f0a8c8c6..0ddb9197d 100644
--- a/src/motion/gopt_f77_methods.f90
+++ b/src/motion/gopt_f77_methods.f90
@@ -21,10 +21,11 @@
 INTERFACE
 
    RECURSIVE SUBROUTINE cp_eval_at(gopt_env, x, f, gradient, master, &
-                                   final_evaluation, para_env)
+                                   final_evaluation, para_env, spgr)
 
       USE cp_para_types, ONLY: cp_para_env_type
       USE gopt_f_types, ONLY: gopt_f_type
+      USE space_groups, ONLY: spgr_type
       USE kinds, ONLY: dp
 
       TYPE(gopt_f_type), POINTER               :: gopt_env
@@ -35,6 +36,7 @@ INTERFACE
       INTEGER, INTENT(IN)                      :: master
       LOGICAL, INTENT(IN), OPTIONAL            :: final_evaluation
       TYPE(cp_para_env_type), POINTER          :: para_env
+      TYPE(spgr_type), INTENT(IN), OPTIONAL    :: spgr
 
    END SUBROUTINE cp_eval_at
``

**Comment by dev-zero** [comment]
... and now that other PRs have been merged you have to rebase again, sorry!

**Comment by pcazade** [comment]
@dev-zero Thanks. I did not realise that "gopt_f77_methods.h" was replace by "gopt_f77_methods.f90". I patched the file, committed, rebased, and force pushed.

**Comment by oschuett** [comment]
The [gopt_f77_methods.f90](https://github.com/cp2k/cp2k/blob/master/src/motion/gopt_f77_methods.f90) is a piece of technical debt that we should clean up. I've opened #1233.

**Comment by dev-zero** [comment]
@pcazade it seems that the patch is somehow messing up the parsing of PCF files, but I don't yet see how.

**Comment by pcazade** [comment]
@dev-zero That's my fault. I modified the routine that reads PSF files as there was an issue with files for large systems generated by VMD. I revert back to the original format.

It seems to have fixed the issue. The remaining issues are with my regtests. They work fine for my machine and do not require the PSF parser. It seems that for some reason, keywords are not recognised. I guess these are the keywords I added for space group symmetry. I double checked the file in src/start/input_cp2k_motion.F after the rebase and the keywords are there. Would you have any idea regarding this issue?

Finally, silly question, is there a way for me to download the current commit of this branch on my laptop, so that I can keep working on it?

**Comment by dev-zero** [comment]
@pcazade wrt git: make a clone of your repo on your laptop. With the HTTPS URL:

```
git clone https://github.com/pcazade/cp2k.git
cd cp2k
git checkout space_group_symmetry  # this should automatically create a branch space_group_symmetry in the local git repo setup to track the remote branch origin/space_group_symmetry

# and then add the CP2K upstream as well
git remote add upstream https://github.com/cp2k/cp2k.git
git fetch upstream
```

while being at it: it seems you pulled (merged?) the commits of the current master into your branch instead of rebasing it. Running the following should resolve that:

```
git fetch upstream
git rebase upstream/master
git push --force
```

**Comment by ltalirz** [comment]
Thanks a lot for this work @pcazade , very much looking forward to getting this merged!

**Comment by pcazade** [comment]
Hi Leopold,

I have to get back to it. I have been busy with articles recently. I have
to admit that I also got frustrated by the only part that was failing are
the regtests that work fine on my workstation. I am clueless about the
actual issue.

Regards,
Pierre

**Comment by ltalirz** [comment]
I see, perhaps we can figure this one out together and get it across the finishing line.

Looking at the regtests, 1 of the tests you added (quartz) actually succeeds, while 3 (copper, zro and zro-filename) fail.
Does that provide any hints?

The stacktrace of the errors go through a couple of functions related to input parsing:
```
#0  0x7f971f8d720f in ???
#1  0x7f971f8d718b in ???
#2  0x7f971f8b6858 in ???
#3  0x2724150 in __machine_internal_MOD_m_abort
	at /workspace/cp2k/src/base/machine_posix.f90:37
#4  0x2720f24 in __message_passing_MOD_mp_abort
	at /workspace/cp2k/src/mpiwrap/message_passing.F:1076
#5  0x26d51d4 in __cp_error_handling_MOD_cp_abort_handler
	at /workspace/cp2k/src/common/cp_error_handling.F:73
#6  0x2721c13 in __base_hooks_MOD_cp_abort
	at /workspace/cp2k/src/base/base_hooks.F:76
#7  0x245c524 in __input_parsing_MOD_section_vals_parse
	at /workspace/cp2k/src/input/input_parsing.F:250
#8  0x245aa97 in __input_parsing_MOD_section_vals_parse
	at /workspace/cp2k/src/input/input_parsing.F:170
#9  0x245aa97 in __input_parsing_MOD_section_vals_parse
	at /workspace/cp2k/src/input/input_parsing.F:170
#10  0x245aa97 in __input_parsing_MOD_section_vals_parse
	at /workspace/cp2k/src/input/input_parsing.F:170
#11  0x1008265 in __input_cp2k_read_MOD_read_input
	at /workspace/cp2k/src/input_cp2k_read.F:73
#12  0x45d208 in cp2k_run
	at /workspace/cp2k/src/start/cp2k_runs.F:215
#13  0x461bf5 in __cp2k_runs_MOD_run_input
	at /workspace/cp2k/src/start/cp2k_runs.F:983
#14  0x44fab2 in cp2k
	at /workspace/cp2k/src/start/cp2k.F:337
#15  0x450189 in main
	at /workspace/cp2k/src/start/cp2k.F:44
```
In particular, `#7` is this:
https://github.com/cp2k/cp2k/blob/f858f91e370704abd981365a2a23b2b9d4ccc580/src/input/input_parsing.F#L248-L250
I.e. it seems the code aborted because it encountered an unknown keyword.


@dev-zero @oschuett I would be very useful to be able to look at the cp2k output in such cases.
Is there a way to get it?

**Comment by oschuett** [comment]
> I would be very useful to be able to look at the cp2k output in such cases. Is there a way to get it?

It's unfortunate that the report contains only part of the error message. I'm not sure where it [gets lost](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest#L424).

Furthermore, one could modify [test_regtest.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_regtest.sh) to copy _interesting files_ to `/workspace/artifacts` which the CI would then make available for download.

**Comment by pcazade** [comment]
@oschuett @ltalirz I checked again the regtests. ZrO tests are based on existing tests in FIST regtests. I simply added the keywords related to space group symmetry. If it works for quartz.inp, it should work for zro.inp and copper.inp. Without the output files, it is difficult to know what went wrong. There is a problem with quartz and copper, though. Both are QM calculations and should be moved to QS regtests. I will work on that along with addressing the main issue. Quartz checks compatibility with OT, while copper tests compatibility with standard diagonalization.

File: tests/Fist/regtest-spgr/TEST_FILES_RESET [hunk 19]
```diff
@@ -0,0 +1 @@
+##
```
oschuett: `TEST_FILES_RESET` files are no longer needed because every regtest now has a fixed reference value.

tests/Fist/regtest-spgr/TEST_FILES
```diff
@@ -0,0 +1,10 @@
+# runs are executed in the same order as in this file
+# the second field tells which test should be run in order to compare with the last available output
+# e.g. 0 means do not compare anything, running is enough
+#      1 compares with the space group name and number of operations provied in the SPGR section of the output file
+#      for details see cp2k/tools/do_regtest
+copper.inp                                   1      Fm-3m   192
+quartz.inp                                   1      P3_221    6
+zro.inp                                      1      P2_1/c    4
+zro-filename.inp                             1      P2_1/c    4
```
oschuett: The last two columns look strange. They should be the threshold, e.g. `1e-14`, and the reference result, e.g. `-0.211229963864E+00`.

ltalirz:
> one could modify [test_regtest.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_regtest.sh) to copy _interesting files_ to `/workspace/artifacts` which the CI would then make available for download.

So, should one copy the cp2k output for failing tests [here](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest#L424) or would that be too much?

As you mention, though, something weird seems to be going on - the output in the test log is not 100 lines long, yet it does not look like a complete cp2k output (no header, ...).

ltalirz:
Edit: I think the missing output has to do with the way cp2k is called [here](https://github.com/cp2k/cp2k/blob/b603aa781d6f0329b06a8535cbc63128475d2fe6/tools/regtesting/do_regtest#L395).

I think I can reproduce it locally as well (using some random messed-up input, not related to this PR)
```
$ cp2k.ssmp -i cp2k.inp

 Possible matches for unknown subsection

 FFGLOBAL

   subsection GLOBAL in section %__ROOT__ score:  19
   enum GLOBAL in section %__ROOT__%MOTION%MD%THERMOSTAT for keyword REGION score:  9
   enum GLOBAL in section %__ROOT__%MOTION%MD%SHELL%THERMOSTAT for keyword REGION score:  9
   enum GLOBAL in section %__ROOT__%MOTION%MD%ADIABATIC_DYNAMICS%THERMOSTAT_FAST for keyword REGION score:  9
   enum GLOBAL in section %__ROOT__%MOTION%MD%ADIABATIC_DYNAMICS%THERMOSTAT_SLOW for keyword REGION score:  9

 *  \___/             unknown subsection FFGLOBAL of section __ROOT__          *
 * / \                                               input/input_parsing.F:186 *


            2 section_vals_parse
            1 read_input

Program received signal SIGABRT: Process abort signal.

Backtrace for this error:
#0  0x1059b74dc
#1  0x1059b684d
#2  0x7fff702745fc
Abort trap: 6

$ cp2k.ssmp -i cp2k.inp &> abc.out
Abort trap: 6

leopold@B126028:~/scratch/cp2k-test
$ cat abc.out

Program received signal SIGABRT: Process abort signal.

Backtrace for this error:
#0  0x104d164dc
#1  0x104d1584d
#2  0x7fff702745fc
```

This is using the cp2k 7.1 executable compiled on conda-forge on my Mac.

I get the same result using cp2k.sopt 7.1 from conda-forge on linux.

When using an older version of cp2k (5.1) on linux, the error messages correctly end up in the output file.

oschuett:
> So, should one copy the cp2k output for failing tests here or would that be too much?

No, more than 100 lines per crash would make the report unreadable for humans and also the dashboard has a [size limit](https://github.com/cp2k/cp2k/blob/master/tools/dashboard/generate_dashboard.py#L646).

The `/workspace/artifacts` directory is a side channel for preserving arbitrary files from a CI run. We use e.g. to exfiltrate the [html manual](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_manual.sh#L25). So, the logic for copying the files would have to go into [test_regtest.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_regtest.sh).

tests/Fist/regtest-spgr/TEST_FILES
pcazade: These tests are to make sure space group symmetry works. The final energy does not really matter.

oschuett:
> Edit: I think the missing output has to do with the way cp2k is called [here](https://github.com/cp2k/cp2k/blob/b603aa781d6f0329b06a8535cbc63128475d2fe6/tools/regtesting/do_regtest#L395).

Nope, it's #1352.

tests/Fist/regtest-spgr/TEST_FILES
oschuett: If possible you should use the opportunity to also compare something from the output. It makes for a much better test.


If you really just want to check that it doesn't crash then you have to use test type zero.
So, your `TEST_FILES` should look like this:
```
# you can add comments like this
copper.inp                                   0
quartz.inp                                   0
# comment lines can be anywhere
zro.inp                                      0
zro-filename.inp                             0
```

oschuett:
Ok, now the full error message is in the [report](https://storage.googleapis.com/cp2k-ci/run-cp2k-sdbg-4c544ca5_report.txt). Turns out the `NDIGITS` keyword was removed in https://github.com/cp2k/cp2k/commit/f9fe71118952d3992826427bc3c40fea08bb6ccc. Furthermore, you have to update the copyright banner of `space_groups.F` to the year 2021.

pcazade:
@oschuett Thanks. I will fix both issues and try again to push.

ltalirz:
No failures anymore, only 2 wrong outputs... getting closer ;-)

P.S. Did you see ole's review comments above https://github.com/cp2k/cp2k/pull/1195#discussion_r571103953 ?

ltalirz:
@oschuett tests pass

pcazade:
@ltalirz @oschuett That's good news that all the tests were successful. I took Ole's suggestion in consideration and removed the unnecessary file. I also relocated the QM regtest files in to the appropriate directory.

pcazade:
@ltalirz @oschuett @dev-zero Is there anything else I should be doing?

ltalirz:
I think you're done for the time being :-) Now that the tests pass, they may want to do a final review of the code changes, which can need a bit of time. A bit more patience and we'll get it merged ;-)

src/motion/space_groups.F
```diff
@@ -0,0 +1,1029 @@
...
+      ! Check if we have other kinds of particles in this subsystem
+      IF (ASSOCIATED(shell_particles)) THEN
+         n_shell = shell_particles%n_els
+         CPASSERT(ASSOCIATED(core_particles))
+         n_core = subsys%core_particles%n_els
+         ! The same number of shell and core particles is assumed
+         CPASSERT(n_core == n_shell)
+      ELSE IF (ASSOCIATED(core_particles)) THEN
+         ! This case should not occur at the moment
+         CPASSERT(ASSOCIATED(shell_particles))
```
dev-zero:
```suggestion

```

src/motion/cg_optimizer.F
```diff
@@ -85,12 +101,51 @@ RECURSIVE SUBROUTINE geoopt_cg(force_env, gopt_param, globenv, geo_section, &
...
+               CALL identify_space_group(spgr, subsys, subsys%cell, geo_section, gopt_env, output_unit)
+               CALL spgr_apply_rotations_coord(spgr, x0)
+               CALL print_spgr(spgr)
+            CASE (default_cell_geo_opt_id, default_cell_md_id)
+               keep_sg = .FALSE.
+            CASE DEFAULT
+               keep_sg = .FALSE.
+            END SELECT
+         CASE DEFAULT
+            keep_sg = .FALSE.
```
oschuett: I think, here you should call `CPABORT("KEEP_SPACE_GROUP not implemented for method foo")` instead.

src/motion/cg_optimizer.F
```diff
@@ -198,14 +288,32 @@ RECURSIVE SUBROUTINE cp_cg_main(force_env, x0, gopt_param, output_unit, globenv,
...
-                         .FALSE., gopt_env%force_env%para_env)
+
+         ! Symmetrize coordinates and forces
+         IF (keep_sg) THEN
+            CALL cp_eval_at(gopt_env, x0, opt_energy, xi, gopt_env%force_env%para_env%mepos, &
+                            .FALSE., gopt_env%force_env%para_env, spgr=spgr)
+         ELSE
+            CALL cp_eval_at(gopt_env, x0, opt_energy, xi, gopt_env%force_env%para_env%mepos, &
+                            .FALSE., gopt_env%force_env%para_env)
+         END IF
```
oschuett: This branching is not needed. Just keep the `ELSE` block. In Fortran one can pass optional argument forward.

src/motion/cp_lbfgs.F
```diff
@@ -269,12 +277,22 @@ SUBROUTINE setulb(n, m, x, lower_bound, upper_bound, nbd, f, g, factr, pgtol, wa
...
       ENDIF
 
-      CALL mainlb(n, m, x, lower_bound, upper_bound, nbd, f, g, factr, pgtol, &
-                  wa(lws), wa(lwy), wa(lsy), wa(lss), wa(lwt), &
-                  wa(lwn), wa(lsnd), wa(lz), wa(lr), wa(ld), wa(lt), wa(lxp), &
-                  wa(lwa), &
-                  iwa(1), iwa(n + 1), iwa(2*n + 1), task, iprint, &
-                  csave, lsave, isave(22), dsave)
+      ! passes spgr to mainlb
+      IF (PRESENT(spgr)) THEN
```
oschuett: Same here. Just keep the `ELSE` block and pass the optional argument `spgr` forward.

src/motion/bfgs_optimizer.F
```diff
@@ -260,13 +273,34 @@ RECURSIVE SUBROUTINE geoopt_bfgs(force_env, gopt_param, globenv, geo_section, go
       ALLOCATE (dr(ndf))
       dr(:) = 0.0_dp
 
+      ! find space_group
+      CALL section_vals_val_get(geo_section, "KEEP_SPACE_GROUP", l_val=keep_sg)
```
oschuett: I think, it would be helpful to generally rename `keep_sg` into `keep_space_group`.

src/motion/cg_optimizer.F
pcazade: That's handy and good to know.

src/motion/cp_lbfgs_optimizer_gopt.F
```diff
@@ -520,23 +540,16 @@ RECURSIVE SUBROUTINE cp_opt_gopt_step(optimizer, n_iter, f, last_f, &
...
-         END IF
-      END IF
-
-      DO
-         ifMaster: IF (is_master) THEN
-            IF (optimizer%task(1:7) == 'RESTART') THEN
-               ! restart the optimizer
-               optimizer%status = 0
-               optimizer%task = 'START'
+            IF (keep_sg) THEN
```
oschuett: And also here. Please just pass the optional `spgr` argument.

src/motion/gopt_f77_methods.F
```diff
@@ -204,6 +218,14 @@ RECURSIVE SUBROUTINE cp_eval_at(gopt_env, x, f, gradient, master, &
                CPASSERT((SIZE(gradient) == nsize + 6))
                CALL pack_subsys_particles(subsys=subsys, f=gradient(1:nsize), fscale=-1.0_dp)
                CALL apply_cell_change(gopt_env, cell, gradient, update_forces=.TRUE.)
+               IF (keep_sg) THEN
+                  CALL spgr_apply_rotations_force(spgr, gradient)
+                  CALL spgr_apply_rotations_stress(spgr, cell, av_ptens)
+                  IF (para_env%ionode) THEN
+                     WRITE (6, *) master, para_env%mepos
```
oschuett: Left over debug code?

src/motion/space_groups.F
```diff
@@ -0,0 +1,1029 @@
...
+            tr(1:3) = spgr%translations(1:3, ir)
+            DO ib = 1, natom
+               IF (.NOT. spgr%lat(ib)) CYCLE
+               rb(1:3) = scoord(1:3, ib)
+               ro(1) = REAL(rot(1, 1), dp)*rb(1) + REAL(rot(2, 1), dp)*rb(2) + REAL(rot(3, 1), dp)*rb(3) + tr(1)
+               ro(2) = REAL(rot(1, 2), dp)*rb(1) + REAL(rot(2, 2), dp)*rb(2) + REAL(rot(3, 2), dp)*rb(3) + tr(2)
+               ro(3) = REAL(rot(1, 3), dp)*rb(1) + REAL(rot(2, 3), dp)*rb(2) + REAL(rot(3, 3), dp)*rb(3) + tr(3)
+               ro(1) = ro(1) - REAL(NINT(ro(1) - ri(1)))
+               ro(2) = ro(2) - REAL(NINT(ro(2) - ri(2)))
+               ro(3) = ro(3) - REAL(NINT(ro(3) - ri(3)))
```
oschuett: `dp` is missing

src/start/input_cp2k_motion.F
```diff
@@ -862,6 +866,68 @@ RECURSIVE SUBROUTINE create_geoopt_section(section, location, label, description
...
+         description="Accuracy required for symmetry detection", &
+         usage="EPS_GEO {REAL}", &
+         default_r_val=1.e-4_dp, repeats=.FALSE.)
+      CALL section_add_keyword(section, keyword)
+      CALL keyword_release(keyword)
+
+      ! collects keywords related to precision for finding the space group
+      CALL keyword_create( &
+         keyword, __LOCATION__, name="EPS_SPGR", &
+         description="Accuracy required for symmetry detection", &
```
oschuett: `EPS_GEO` and `EPS_SPGR` have the same descriptions. Maybe you need only one? Also these keyword names are not very self-explanatory. How about `EPS_SYMMETRY`?

tests/Fist/regtest-spgr/TEST_FILES
```diff
@@ -0,0 +1,8 @@
+# runs are executed in the same order as in this file
+# the second field tells which test should be run in order to compare with the last available output
+# e.g. 0 means do not compare anything, running is enough
+#      1 compares with the space group name and number of operations provied in the SPGR section of the output file
+#      for details see cp2k/tools/do_regtest
+zro.inp                                      0      P2_1/c    4
```
oschuett: If you want to keep the comment (`P2_1/c    4`) then please move it to a separate line:
```
# P2_1/c    4
zro.inp                                      0
zro-filename.inp                             0      
```

tests/QS/regtest-spgr/TEST_FILES
```diff
@@ -0,0 +1,8 @@
+# runs are executed in the same order as in this file
+# the second field tells which test should be run in order to compare with the last available output
+# e.g. 0 means do not compare anything, running is enough
+#      1 compares with the space group name and number of operations provied in the SPGR section of the output file
+#      for details see cp2k/tools/do_regtest
+copper.inp                                   0      Fm-3m   192
```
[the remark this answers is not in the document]
pcazade: Sure.

src/start/input_cp2k_motion.F
pcazade: One keyword is used by SPGLIB, the other in my routines. Originally, the idea was to be able to have different thresholds to make it easier to identify the space group of a set of coordinates with low accuracy (e.g. PDB files). It also helped me pass the spglib check to be able to test my routines. I agree that, at this stage, it does not make sense anymore to have both. I will take care of it.

tests/Fist/regtest-spgr/TEST_FILES
pcazade: Is there a way to compare strings? Actually, I would like the regtest to check if the space group is properly identified after compilation?

src/motion/space_groups.F
pcazade: Should I use CPABORT instead?

src/motion/cg_optimizer.F
pcazade: I am not sure CPABORT is what we need here. Correct me if I am wrong but, don't one end up in the CASE(default_cell_geo_opt_id) during the cell optimization step when one is not using the direct cell method but rather alternate cell opt and geo opt. If it is the case, the routines for space group symmetry have to be disabled for this particular step, and enabled for the geo_opt one.

src/motion/space_groups.F
dev-zero: No, nothing. You will never hit this assertion. If `ASSOCIATED(shell_particles)` is true you are in the first branch of this `IF/ELIF/ELSE`.

src/motion/cg_optimizer.F
pcazade: However, for default_cell_md_id, it makes sense to abort, though.

oschuett: You are right, that would probably break the legacy cell opt. However, I doubt anybody is still using it and we're planing to remove it (#1233).

Unfortunately, our motion code is in pretty bad shape. I guess, you could tell from the fact that you had to hook into each optimizer (CG, BFGS, LBFGS) separately.

So, I'd rather error on the side of caution. Are you sure it's ok to continue with `keep_sg = .FALSE.` in all those cases?

tests/Fist/regtest-spgr/TEST_FILES
oschuett: Unfortunately our `do_regtest` script can not readily compare strings.
However, the return value from [`spg_get_international`](https://spglib.github.io/spglib/api.html#spg-get-international) is actually the space group number. So, you could just output that one alongside the symbol.

src/motion/cg_optimizer.F
pcazade: As far as I can tell and test, it is. Furthermore, space_group_symmetry works for the legacy cell opt. I made sure of it. I see not harm keeping this until the feature is removed. I will split the CASE however to have a CPABORT for the md case.

oschuett: Sounds good! You certainly know the code better than me at this point.

tests/Fist/regtest-spgr/TEST_FILES
pcazade: Silly me, I should have thought about that earlier.

I have an issue with spgr%international_symbolol, spgr%pointgroup_symbol, and spgr%schoenflies. There is a hidden character at the end that is visible with vim (^@). It also introduces a shift by one character visible with any editor. Is there a way to delete the last character like in python. I could do a TRIM(ADJUSTL(spgr%international_symbolol)) first, then how do I remove the last character?

src/motion/gopt_f77_methods.F
pcazade: Yes, I am afraid. Or rather prints to make sure I understand the code correctly.

src/motion/cg_optimizer.F
pcazade: There are many more cases for which I do this branching than the two you commented about. Should I update all of them? For example, there is a similar case on line 135 of this file (src/motion/cg_optimizer.F):

      IF (keep_space_group) THEN
         CALL cp_cg_main(force_env, x0, gopt_param, output_unit, globenv, &
                         gopt_env, do_update=my_do_update, spgr=spgr)
      ELSE
         CALL cp_cg_main(force_env, x0, gopt_param, output_unit, globenv, &
                         gopt_env, do_update=my_do_update)
      ENDIF

tests/Fist/regtest-spgr/TEST_FILES
oschuett: This is probably the null byte that terminates strings in C. In Fortran this is not needed because strings have fixed length. The conversion can be done like [this](https://github.com/cp2k/cp2k/blob/master/src/grid/grid_api.F#L1347-#L1352). I guess, we should add a helper function for this to [string_utilities.F](https://github.com/cp2k/cp2k/blob/master/src/common/string_utilities.F) some day.

src/motion/cg_optimizer.F
oschuett: Yes, please try to remove as many of these branches as possible to keep code duplication low.

tests/Fist/regtest-spgr/TEST_FILES
dev-zero: here you go: #1358

pcazade: @dev-zero @oschuett Thanks for the string converter. I did a fetch upstream and rebase to get it into my branch, but it seems it is not there. When I try to compile, I get an error that the function "strlcpy_c2f" is not not string_utilities module. I guess I did not use the correct git function.

On a different topic, the space group number is not accessible in Fortran from Spglib. There is not function for that. Ideally, accessing the dataset would solve this issue and spare three calls to Spglib space group identifying routine (not an issue for small systems but definitely a bottleneck for tens thousand+ systems). The module spglib_f08 contains an interface to access the dataset and even a fortran definition of the dataset, yet, during compilation I get an error that it does not exist. I cannot find what the issue is. The dataset type and get function are public:  https://github.com/spglib/spglib/blob/master/example/spglib_f08.f9

dev-zero: @pcazade the link above was a PR, which I now just merged. If you fetch/rebase again now you should have the function. Usage of it can be seen in either `src/start/libcp2k.F` or `src/grid/grid_api.F`

pcazade: @dev-zero Thanks. Works like a charm.

src/motion/cg_optimizer.F
pcazade: @oschuett Alright, I removed all the branches I could find and hopefully all there are. A problem that appeared to me, though, is that I rely a lot on IF(PRESENT(spgr)) to set keep_space_group=.TRUE. Should I use the function NULL instead to check if spgr is actually defined? Is there a better option?

oschuett: Thanks. Simply write `keep_space_group = PRESENT(spgr)`. I'd consider that well-written Fortran.

The `NULL()` function only returns a null pointer in [Fortran](https://gcc.gnu.org/onlinedocs/gfortran/NULL.html).

pcazade: The thing is, without the branches, spgr is passed by default. For example:

IF (keep_space_group) THEN
   CALL cp_cg_main(force_env, x0, gopt_param, output_unit, globenv, &
                                 gopt_env, do_update=my_do_update, spgr=spgr)
ELSE
   CALL cp_cg_main(force_env, x0, gopt_param, output_unit, globenv, &
                                  gopt_env, do_update=my_do_update)
ENDIF

is now:

CALL cp_cg_main(force_env, x0, gopt_param, output_unit, globenv, &
                               gopt_env, do_update=my_do_update, spgr=spgr)

Did I misunderstand what you wanted me to do with these branches? If what I did is what you asked me to do, spgr being always there, the if(pesent(spgr)) test does not work anymore.

oschuett: No, it will still work. You can simply pass an `OPTIONAL` argument on to another function.

pcazade: @oschuett Sorry, I did not make my case clear. I have no doubts it works when I require the use of space group symmetry. I actually already checked it does when I tested that I did not delete to much or the wrong part of the branch.

My concern is for when someone is not using space group symmetry. The test present(spgr) will give true and keep_space_group will also be true. However, space group symmetry is not defined and this makes cp2k crash. I also tested that. Should I switch to if (spgr%n_operattions > 0) instead or make spgr an optional pointer rather than a variable? Would the latter affect the syntax of the functions?

pcazade: @oschuett removing the branches has unanticipated consequences. It creates many situation where spgr might be passed to routine by another routine which did not get it passed. I totally understand the need to avoid duplication of the code for the sake of maintenance, clarity, and readability, but could we leave it as it is for know and mark it as a later task. LBFGS and CP_EVAL_AT are particularly susceptible to this problem. Ideally, SPGR should be part of gopt_env. I did not do it because there are issues as well and gopt_env is not always available where I need it. I guess it all comes down to what you mention earlier: the motion code is in pretty bad shape.

oschuett: I see your point and I tend to agree that there is no simple and clean solution.

However, before we give up maybe you could, as a last attempt, try to move `spgr` into the `gopt_env`, ie. [`gopt_f_type`](https://github.com/cp2k/cp2k/blob/master/src/motion/gopt_f_types.F#L60) ?

pcazade: @oschuett I remember now one the reasons I did not do it in first place, space_group.F needs gopt_f_type. I can not call space_group module in gopt_f_type definition. An easy fix is to create a separate file where spgr type is defined: space_group_t.F. Of course, I have to check that an object with the type gopt_f_type is always passed in the routines where spgr is also passed. I will give it a try. It would certainly make the code cleaner. It would make it easier to access space group data for other use.

pcazade: Done. spgr structure is now found into gopt_f_type. I made a series of quick tests an everything seems fine to me. Hopefully, I did not forget any branches. Please let me know otherwise.

src/motion/bfgs_optimizer.F
pcazade: The renaming is done.

src/motion/space_groups.F
```diff
@@ -0,0 +1,995 @@
...
+      n_atom = particles%n_els
+      ! Check if we have other kinds of particles in this subsystem
+      IF (ASSOCIATED(shell_particles)) THEN
+         n_shell = shell_particles%n_els
+         CPASSERT(ASSOCIATED(core_particles))
+         n_core = subsys%core_particles%n_els
+         ! The same number of shell and core particles is assumed
+         CPASSERT(n_core == n_shell)
+      ELSE IF (ASSOCIATED(core_particles)) THEN
+         ! This case should not occur at the moment
```
oschuett, line 355: Then maybe put a `CPABORT` here just in case.

tests/Fist/regtest-spgr/TEST_FILES
```diff
@@ -0,0 +1,8 @@
+# runs are executed in the same order as in this file
+# the second field tells which test should be run in order to compare with the last available output
+# e.g. 0 means do not compare anything, running is enough
+#      1 compares with the space group name and number of operations provied in the SPGR section of the output file
+#      for details see cp2k/tools/do_regtest
+zro.inp                                      0
```
oschuett: Do you still want the test to check that the correct space group was recognized?

oschuett (commented):
Looks good. Just a few nits. Almost ready to merge :-)

tests/Fist/regtest-spgr/TEST_FILES
pcazade: @oschuett I would like to but there is no function in spglib_f08 module to get the space group number. The only way would be to get the dataset. However, despite both the dataset structure and dataset get function being defined in the module and being made public, compilation fails saying these cannot be found. So, I can only get the various space group names which are all strings. I don't know, considering the structure of the regtests, if it is possible to compare strings.

oschuett: According to the [docs](https://spglib.github.io/spglib/api.html#spg-get-international) the routine `spg-get-international` returns the space group number. So, you should already have it in `ierr` in your [code](https://github.com/cp2k/cp2k/blob/5095393704225a9e265ec3c0f80189f28470a976/src/motion/space_groups.F#L245).

src/motion/space_groups.F
pcazade: Done.

src/motion/space_groups.F
```diff
@@ -0,0 +1,995 @@
...
+            CALL init_cell(spgr%cell_ref, hmat=gopt_env%h_ref)
+         CASE (default_cell_geo_opt_id)
+            CPABORT("SPACE_GROUP_SYMMETRY SHOULD NOT BE INVOKED DURING THE CELL STEP.")
+         CASE (default_cell_md_id)
+            CPABORT("SPACE_GROUP_SYMMETRY IS NOT COMPATIBLE WITH MD.")
+         CASE DEFAULT
+            CPABORT("SPACE_GROUP_SYMMETRY INVOKED WITH AN UNKNOWN OPTIMIZATION METHOD.")
+         END SELECT
+      CASE DEFAULT
+         CPABORT("SPACE_GROUP_SYMMETRY IS NOT COMPATIBLE WITH MD.")
```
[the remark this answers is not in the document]
pcazade: Done

src/start/input_cp2k_motion.F
```diff
@@ -862,6 +866,59 @@ RECURSIVE SUBROUTINE create_geoopt_section(section, location, label, description
...
+         keyword, __LOCATION__, name="SYMM_EXCLUDE_RANGE", &
+         description="Range of atoms to exclude from symmetry. This keyword can be repeated.", &
+         repeats=.TRUE., usage="SYMM_EXCLUDE_RANGE {Int} {Int}", type_of_var=integer_t, n_var=2)
+      CALL section_add_keyword(section, keyword)
+      CALL keyword_release(keyword)
+
+      ! collects keywords related to print SPGR info
+      CALL keyword_create( &
+         keyword, __LOCATION__, name="SPGR_FILE_NAME", &
+         description="Name of output file for detailed Space Group Symmetry iformation.", &
```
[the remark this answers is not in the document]
pcazade: Done

src/motion/cp_lbfgs.F
```diff
@@ -425,6 +439,11 @@ SUBROUTINE mainlb(n, m, x, lower_bound, upper_bound, nbd, f, g, factr, pgtol, ws
 !
 !                           *  *  *
 
+      keep_space_group = .FALSE.
+      IF (PRESENT(spgr) .AND. ASSOCIATED(spgr)) THEN
+         IF (spgr%keep_space_group) keep_space_group = .TRUE.
```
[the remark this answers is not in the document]
pcazade: Done. I wonder why I did not that in the first place...

tests/Fist/regtest-spgr/TEST_FILES
pcazade: Brilliant. I did not realise. That's fixed too. Now, how do I make it so that the regtest compare this value? Which number should I put instead of 0 in the second column of the TEST_FILES?

pcazade:
@oschuett @dev-zero I am trying to run more extensive tests. When using dft, I encounter this error. I recompiled from scratch this morning (starting with the toolchain step) and still get this error:

 *  \___/         Cholesky decomposition failed. Matrix ill conditioned ?      *
 * / \                                                 cp_dbcsr_cholesky.F:120 *

There is a memory allocation issue with BLAS:

BLAS : Bad memory unallocation! :  256  0x7fc238000000

Any idea what is the issue?

pcazade:
The error in conventions was straightforward to address (a float remaining from copying a routine I initially wrote in python). However, the error in minimal arch is not that clear. I understand that the absence of spglib induces an allocation issue in L-BFGS in link with my changes, but without any file and line reference, it is hard to pinpoint what exactly is causing the issue. All these files don't request a calculation involving space group symmetry, therefore spgr%keep_space_group should be .false., and none of the spglib dependent routines should be called. The definition of spgr_type does not involve the module spglib_f08, this should be well defined.

I compiled the minimal version of cp2k and tested it on some of the files which failed in the Minimal arch file test. They work fine on my workstation. I don't have the segmentation fault that can be observed in the report file.

oschuett:
The problem with the minimal test is [here](https://github.com/cp2k/cp2k/blob/7e45325f2b82fbda804349c70699a00b0d68e421/src/motion/cp_lbfgs_optimizer_gopt.F#L527) and [here](https://github.com/cp2k/cp2k/blob/7e45325f2b82fbda804349c70699a00b0d68e421/src/motion/cp_lbfgs.F#L443) where you used `IF (PRESENT(spgr) .AND. ASSOCIATED(spgr)) `. This segfaults because `.AND.` is not short-circuiting in Fortran. It's a trap I also like to fall into myself.

pcazade:
@oschuett Thanks. Should I decompose this test then?

IF (PRESENT(spgr)) THEN
   IF (ASSOCIATED(spgr)) THEN

tests/Fist/regtest-spgr/TEST_FILES
oschuett: You need to add a new pattern to [TEST_TYPES](https://github.com/cp2k/cp2k/blob/master/tests/TEST_TYPES). The format is described at the end of that file. Don't forget to increment the counter at the top. Afterwards you can put `93` instead of `0` in the second column of your TEST_FILES.

oschuett:
> @oschuett Thanks. Should I decompose this test then?

Yes, unfortunately you have to. You can write it little bit more compact like this:
```
IF (PRESENT(spgr)) THEN
   IF (ASSOCIATED(spgr)) keep_space_group = spgr%keep_space_group
END IF

oschuett:
It seems this PR is ready to be merged. Do you agree @pcazade?

pcazade:
@oschuett I think so too.

oschuett:
Alright, it's done. Thanks for contributing. Great work!

pcazade:
Thanks. Delighted to be able to contribute to such a great project cp2k is.

ltalirz:
@pcazade I didn't want to drag out this PR any longer than necessary, so I did not comment on the implementation itself but now that it's merged, I'm wondering whether you can provide a bit more insight/documentation:

 * The spacegroup is autodetected using spglib, is this correct? Since there is no reference to the spacegroup anywhere else in the input, this might be good to mention in the help string of `KEEP_SPACEGROUP`
 * This is not for the help, but more general for users of the feature: perhaps a few comments on the implementation might be useful (you could edit the top comment on the PR). I.e. does the code take advantage of the symmetry to reduce computations or is symmetrization applied a posteriori, etc.

pcazade:
@ltalirz No problem. It makes sense. After I add the description you suggest around the KEEP_SPACE_GROUP keyword in input_cp2k_motion.F, should I make another PR to merge these changes too?

Issue #893, opened 2020-05-01
CP2K 7.1-Cuda Bandgap and HF energies different from previous versions


AndresOrtegaGuerrero:
Dear CP2K developers,

I have used CP2K 5.1 and 6.1 for PBE0-TC-LRC calculations in the past.
I have found that my Bandgap values are consistent in versions 5.1 and 6.1 in our local cluster and other cluster too.
I did calculations with the versions 6.1 and 7.1 with Cuda in Piz Daint, and I have found that the Bandgaps different and the HF energies (and total energies) are now different
There is no difference using a restart wfn file or optimizing the SCF cycle from the beginning. 
This is one example but I have noticed in different systems of periodic calcuations.
5.1 and 6.1 no cuda
(1) 
   Overlap energy of the core charge distribution:               0.00007052589464
  Self energy of the core charge distribution:              -4822.08795661199383
  Core Hamiltonian energy:                                   1456.10019696034419
  Hartree energy:                                            1951.78336227053728
  Exchange-correlation energy:                               -433.84963055662922
  Hartree-Fock Exchange energy:                              -127.87395151550244
  Dispersion energy:                                           -0.46932014114032
  Total energy  ENERGY| Total FORCE_EVAL ( QS ) energy (a.u.):            -1976.397228950271256
   HOMO - LUMO gap [eV] :    2.910486

6.1 and 7.1 with cuda 
 (2)
    Overlap energy of the core charge distribution:               0.00007052589464
  Self energy of the core charge distribution:              -4822.08795661199383
  Core Hamiltonian energy:                                   1456.47181019598679
  Hartree energy:                                            1951.50352600623592
  Exchange-correlation energy:                               -438.90435624956706
  Hartree-Fock Exchange energy:                              -121.46009108451653
  Dispersion energy:                                           -0.46932014114033
  Total energy ENERGY| Total FORCE_EVAL ( QS ) energy (a.u.):            -1974.946317359100703
  HOMO - LUMO gap [eV] :    2.747721

best, 

Andres Ortega
LSMO EPFL

ltalirz:
Is it possible to disable the GPU in the gpu-version of cp2k?
I.e. can Andres test the result on piz daint with the GPU disabled? 

@AndresOrtegaGuerrero it might be useful if you attach the input file and the submission script

pseewald:
Why was this issue closed? I can have a look at it, it would be helpful if you could provide an input file.

[...]

PR #1200, opened 2020-12-03
DBCSR: Add regtest for thread dist (issue #893)

oschuett:


alazzaro:
@oschuett I think you will wait for the new DBCSR to check with? I mean, the test is supposed to fail without the fix...

oschuett:
Actually, my plan was to submit this right away. Since you have a patch coming, the dashboard shouldn't be red for too long.
It will allow you to validate your patch and we document the fix in the dashboard history.

Wouldn't you agree?

alazzaro:
OK, then I understood your point and agree with you. 
Yes, this new DBCSR RC will go in today.
I'm still waiting for a reply to the #893 (we got one)... Probably there is another missing problem, people are reporting wrong results even with a single thread...

PR #1204, opened 2020-12-04
GW2X| Implementation of the GW2X method 

abussy:
The GW2X method (https://doi.org/10.1002/qua.1543) was implemented. Combined with the local RI used in XAS_TDP, the GW2X method allows for efficient calculation of the ionization potential of core states. This can be used as a correction scheme for XAS LR-TDDFT, or on its own for XPS calculations. The implementation uses DBCSR tensors extensively.


src/xas_tdp_correction.F
```diff
@@ -0,0 +1,1870 @@
...
+
+   CHARACTER(len=*), PARAMETER, PRIVATE :: moduleN = 'xas_tdp_correction'
+
+   PUBLIC :: gw2x_shift, get_soc_splitting
+
+CONTAINS
+
+! **************************************************************************************************
+!> \brief First implementation of the very legit GW2X correction of the Kohn-Sham IP from
+!>        https://onlinelibrary.wiley.com/doi/pdf/10.1002/qua.1543
```
dev-zero: maybe do a bibliography entry for this one?

src/xas_tdp_correction.F
```diff
@@ -0,0 +1,1870 @@
...
+      CHARACTER(len=*), PARAMETER :: routineN = 'GW2X_shift', routineP = moduleN//':'//routineN
+
+      INTEGER :: ex_idx, exat, first_domo(2), handle, i, ido_mo, iloc, ilocat, ispin, jspin, &
+         locat, nao, natom, ndo_mo, nhomo(2), nlumo(2), nonloc, nspins, start_sgf
+      INTEGER, DIMENSION(:), POINTER                     :: nsgf_blk
+      LOGICAL                                            :: pseudo_canonical
+      REAL(dp)                                           :: og_hfx_frac
+      REAL(dp), ALLOCATABLE, DIMENSION(:, :)             :: contract_coeffs_backup
+      TYPE(admm_type), POINTER                           :: admm_env
+      TYPE(cp_1d_r_p_type), DIMENSION(:), POINTER        :: homo_evals, lumo_evals
```
dev-zero: `ALLOCATABLE` instead and change the interface for `GW2X_os_iterations` to take it just as a normal array?

src/xas_tdp_correction.F
```diff
@@ -0,0 +1,1870 @@
...
+!> \param lumo_evals ...
+!> \param donor_state ...
+!> \param xas_tdp_control ...
+!> \param qs_env ...
+! **************************************************************************************************
+   SUBROUTINE GW2X_rcs_iterations(first_domo, ja_X, oI_Y, mo_template, homo_evals, lumo_evals, &
+                                  donor_state, xas_tdp_control, qs_env)
+
+      INTEGER                                            :: first_domo
+      TYPE(dbcsr_t_type)                                 :: ja_X
```
dev-zero: `INTENT` missing

src/xas_tdp_correction.F
```diff
@@ -0,0 +1,1870 @@
...
+
+            !compute the diff to the new step
+            diff = -g/dg
+
+            !and the new omega
+            omega_k = omega_k + diff
+            diff = diff*evolt
+
+            IF (output_unit > 0) THEN
+               WRITE (UNIT=output_unit, FMT="(T21,I18,F32.6)") &
```
dev-zero, line 627: is it possible that this is a debug leftover? Plus, why always the `flushes`?

src/xas_tdp_types.F
```diff
@@ -262,13 +275,14 @@ MODULE xas_tdp_types
       INTEGER                                 :: state_type
       INTEGER, DIMENSION(:), POINTER          :: blk_size
       REAL(dp), DIMENSION(:, :), POINTER      :: energy_evals
+      REAL(dp), DIMENSION(:, :), POINTER      :: gw2x_evals
```
dev-zero, line 278: couldn't this be an allocatable instead?

src/xas_tdp_correction.F
abussy: Yes you are right. Also, the text could be somewhat updated.

src/xas_tdp_correction.F
abussy: This would be a pain in open-shell calculations, I think. Carrying the eigenvalues this way allows to iterate over the spin and access the relevant eigenvalues when needed. The alternative would be to explicitelly have and alpha and beta version of both arrays, which I do not like.

src/xas_tdp_correction.F
abussy: It is not a debug leftover. For large systems, an iteration can be quite slow. This helps keeping track of the calculation status. The `flush` is also there for that reason. If the output is buffered, then printing this info is much less informative if it comes all at once.

src/xas_tdp_types.F
abussy: It might be. I am not sure how much the code would need to be adapted. I will look into it.

src/xas_tdp_correction.F
dev-zero: How so? You just need a dynamically allocated array of type `cp_1d_r_p_type`, correct? And you are using it only in that procedure without having to call another API which would require it to be a POINTER or TARGET.

src/xas_tdp_correction.F
dev-zero: ok, I see

src/xas_tdp_correction.F
abussy: I might have missunderstood. You mean `TYPE(cp_1d_r_p_type), DIMENSION(:), ALLOCATABLE  :: homo_evals, lumo_evals` ? In that case, yes, there is nothing preventing me from doing that and I will change it.

dev-zero: yes, that :)

alazzaro:
This PR introduced some errors in some of the dashboard regtests, for example:

https://dashboard.cp2k.org/archive/psi-intel-psmp/commit_0a2bdfe05cbcc354d398fd15d54bf44a67842ed1.txt

abussy:
Indeed. This seems to be limited to the intel compiler though. I'll investigate.

abussy:
#1212 Should fix it.

Issue #464, opened 2019-07-12
Explore GPU support in ELPA


oschuett:
Currently, the [cuda toolchain](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/Dockerfile.cuda_mkl) does not include ELPA. I briefly tried to build ELPA with GPU support and failed. Anecdotal evidence suggests that it's not quite ready yet. Nevertheless, we should give it a try every once in a while.

dev-zero:
updating ELPA may help #189, #190

ghost:
What version of gfortran is your toolchain using?  Our (FHI-aims/ELSI) experience is that ELPA should be buildable with GPU support, but it can be a bit finicky with compiler versioning.

oschuett:
The [test](https://dashboard.cp2k.org/archive/cuda-pascal/index.html) is using GCC 7.4.0 from Ubuntu 18.04 and Cuda tools 10.1, V10.1.168.

Since you manged to build ELPA with CUDA, what kind of speedups do you typically see?

oschuett:
There is now [elpa-2020.05.001.rc1](https://elpa.mpcdf.mpg.de/elpa-tar-archive) with lots of [GPU improvements](https://arxiv.org/pdf/2002.10991.pdf).

oschuett:
I made some progress.

* The Good: Since #1213 the toolchain can build ELPA with Cuda.
* The Bad: In most cases the GPU kernel is not used and instead ELPA falls back to the CPU:
  ```
  ELPA: Warning, CUDA kernel only works with block size 2^n (n = 1, 2, ..., 10)!
  The compute kernel will be executed on CPUs!
  ```
* And the Ugly: That fall back to the CPU often leads to [segfaults](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-pascal-aa124875_report.txt):
  ```
  Program received signal SIGSEGV: Segmentation fault - invalid memory reference.
  Backtrace for this error:
  ...
  #3  0x5634150d22ca in __elpa2_compute_MOD_bandred_real_double._omp_fn.1 at ../src/elpa2/elpa2_bandred_template.F90:954
  #4  0x7f30659da8e5 in ???
  #5  0x5634150ffa41 in __elpa2_compute_MOD_bandred_real_double at ../src/elpa2/elpa2_bandred_template.F90:947
  ```

oschuett:
[...]

PR #1213, opened 2020-12-09
Docker: Enable ELPA with Cuda

oschuett:
This is a followup to #1141.


dev-zero:
@oschuett it would be really useful to also get a regtest with an ELPA CUDA kernel, and one may also have to figure out whether the current wrapper to generate matching blocksizes also works for CUDA. The small testcase I had time to run reported a lot of fallbacks to CPU due to non-matching blocksizes, see also comment in 2fae766 . For debugging the QR is was thinking of using [serialbox](https://github.com/GridTools/serialbox) to capture the input passed along to ELPA to be able to isolate data for a unittest.

oschuett:
Yes, ELPA is really a low hanging fruit to improve our GPU support (#464).

Does [AUTO](https://github.com/cp2k/cp2k/blob/master/src/fm/cp_fm_elpa.F#L208) select the GPU kernel when available?

> For debugging the QR is was thinking of using serialbox to capture the input passed along to ELPA to be able to isolate data for a unittest.

Yes, we should definitely add more unit tests. However, I'm not sure we need to pull in another library for serialization. Can't we just use `WRITE` / `READ` or hdf5? Actually, with ELPA we could just generate random input and compare against LAPACK.

dev-zero:
> Yes, ELPA is really a low hanging fruit to improve our GPU support (#464).

Well, that's what needs to be proven and really depends on whether we succeed on supplying matrix blocks of the right size.

> Does [AUTO](https://github.com/cp2k/cp2k/blob/master/src/fm/cp_fm_elpa.F#L208) select the GPU kernel when available?

No, because https://github.com/cp2k/cp2k/blob/3567d8e35507119586737a68c7f87590c4e8640f/src/fm/cp_fm_elpa.F#L464-L465
Whether or not ELPA will use a CUDA kernel if that is set but kernel is left to AUTO has to be determined by reading ELPA's source or docs. As far as I remember: don't think so.

> > For debugging the QR is was thinking of using serialbox to capture the input passed along to ELPA to be able to isolate data for a unittest.
> 
> Yes, we should definitely add more unit tests. However, I'm not sure we need to pull in another library for serialization. 

https://github.com/fortesg/fortrantestgenerator would make it at lot easier to write such unittests based on serialbox.

> Can't we just use `WRITE` / `READ` or hdf5? Actually, with ELPA we could just generate random input and compare against LAPACK.

That's another option, yes.

oschuett:
The build fails because it can't link BLAS. Any ideas?

dev-zero:
> The build fails because it can't link BLAS. Any ideas?

```
[...]
checking for Fortran libraries of mpif90...  -L/usr/lib/x86_64-linux-gnu -L/opt/cp2k-toolchain/install/cosma-2.2.0/lib/../lib -L/opt/cp2k-toolchain/install/libxsmm-1.16.1/lib/../lib -L/opt/cp2k-toolchain/install/libxc-4.3.4/lib/../lib -L/opt/cp2k-toolchain/install/libint-v2.6.0-cp2k-lmax-4/lib/../lib -L/opt/cp2k-toolchain/install/SpFFT-0.9.13/lib/../lib -L/opt/cp2k-toolchain/install/fftw-3.3.8/lib/../lib -L/usr/lib/gcc/x86_64-linux-gnu/9 -L/usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu -L/usr/lib/gcc/x86_64-linux-gnu/9/../../../../lib -L/lib/x86_64-linux-gnu -L/lib/../lib -L/usr/lib/../lib -L/opt/cp2k-toolchain/install/cosma-2.2.0/lib -L/opt/cp2k-toolchain/install/libxsmm-1.16.1/lib -L/opt/cp2k-toolchain/install/libxc-4.3.4/lib -L/opt/cp2k-toolchain/install/libint-v2.6.0-cp2k-lmax-4/lib -L/opt/cp2k-toolchain/install/SpFFT-0.9.13/lib -L/opt/cp2k-toolchain/install/fftw-3.3.8/lib -L/usr/local/cuda/lib64/stubs -L/usr/lib/gcc/x86_64-linux-gnu/9/../../.. -lpthread -lgfortran -lm -ldl -lmpichfort -lmpich -lquadmath
[...]
```
... MKL seems to be missing? And most of the other deps build only a static library and/or no executables, so linking is deferred until CP2K.

on the other hand it also says:

```
checking whether we can link a program with a blas lib... yes
checking for library containing dlarrv... none required
checking whether we can link a program with a lapack lib... yes
```
maybe some of the options are not passed along to the makefile building the executables.

PR #1238, opened 2020-12-20
Enable regression testing of pexsi, quip, and libvori with arm

mkrack:


oschuett:
It's strange that PEXSI returns different integer types between ARM64 and x86-64. Should we maybe file a bug with [them](https://bitbucket.org/berkeleylab/pexsi/)?

mkrack:
I was also puzzled by that behaviour, since the iso_c_binding for the `plan` variable of type c_intptr should be consistent, i.e. that type should have a size of 8 bytes on 64 bit systems. The length of 4 bytes indicates that an armv7l system is considered as a 32 bit system which seems to be correct. Though the ARM Cortex-A72 processor of the RPi4 implements a 64 bit instruction set, the (Raspbian) armv7l is still running in 32 bit mode.
```
> lscpu
Architecture:        armv7l
Byte Order:          Little Endian
CPU(s):              4
On-line CPU(s) list: 0-3
Thread(s) per core:  1
Core(s) per socket:  4
Socket(s):           1
Vendor ID:           ARM
Model:               3
Model name:          Cortex-A72
Stepping:            r0p3
CPU max MHz:         1500.0000
CPU min MHz:         600.0000
BogoMIPS:            144.00
Flags:               half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm crc32
> file /bin/bash
/bin/bash: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for GNU/Linux 3.2.0, BuildID[sha1]=3e5e2847bbc51da2ab313bc53d4bdcff0faf2462, stripped

dev-zero:
I think the flag should probably be `__32BIT` (or something like that) rather than `__ARM` then, but the main problem is that we are storing an `INTEGER(c_intptr_t)` in an `INTEGER(KIND=int_8)` in the first place rather than keeping it a `c_intptr_t`:
https://pexsi.readthedocs.io/en/latest/corefunction.html#fortran-interface

The question is why: if it is an attempt to hide possibly architecture-specific (and the ISO_C_BINDINGS) from the rest of CP2K, then it might be better to always store the handle as an `int_8` and cast where needed rather than changing structures depending on architecture. Or be consequent and store it properly as `c_intptr_t`.

@pseewald your opinion?

mkrack:
@dev-zero Indeed, that flag should be named differently after realizing that it is not a ARM specific issue. The best would be, of course, if we could avoid that flag completely.

mkrack:
Resolved by #1242

PR #1254, opened 2020-12-27
grid: Add simple coalesced group reduction to improve performance for Cuda <11

oschuett:


oschuett:
Since Daint is currently unavailable, I used GCP to validate that this change does indeed fix the [performance issue](https://dashboard.cp2k.org/archive/cray-xc50-daint-gpu-psmp_perf/index.html).

The results for the [H2O-32.inp](https://github.com/cp2k/cp2k/blob/master/benchmarks/QS/H2O-32.inp) benchmark are:

```
===== CUDA 11 on V100 =====
  CP2K                                 1  1.0    0.016    0.016   93.776   93.776
  grid_integrate_task_list           116 12.3   57.895   57.895   57.895   57.895
  grid_collocate_task_list           116  9.6    5.381    5.381    5.381    5.381

===== CUDA 10 on V100 =====
  CP2K                                 1  1.0    0.015    0.015   94.691   94.691
  grid_integrate_task_list           116 12.3   57.887   57.887   57.887   57.887
  grid_collocate_task_list           116  9.6    5.386    5.386    5.386    5.386

===== No GPU / 12 OpenMP threads =====
  CP2K                                 1  1.0    0.022    0.022  124.879  124.879
  grid_integrate_task_list           116 12.3   53.387   53.387   53.387   53.387
  grid_collocate_task_list           116  9.6   45.453   45.453   45.453   45.453
```

The good news is that the difference between Cuda 10 and 11 is negligible. Furthermore, collocate gets a solid 8x speedup from the GPU. Unfortunately, the integrate kernel only matches the CPU's performance. While I'm optimistic that this can be improved, it will probably require a fundamental redesign of the integrate kernel.

Nevertheless, the Cuda kernels should already be fast enough to make CP2K eligible for new machines like LUMI or Leonardo.

PR #1257, opened 2020-12-28
Add build time error for missing OpenMP

oschuett:


mbanck:
Not sure it matters much, but when I try to compile CP2K without `-fopenmp` in `FCFLAGS`, it already bails out when linking `grid_miniapp`, it doesn't get till `cp2k_info.F`:

```
/usr/bin/ld: /<<PKGBUILDDIR>>/lib/Linux-x86_64-gfortran/psmp/libcp2kgridcpu.a(grid_integrate_dgemm.o): in function `integrate_one_grid_level_dgemm._omp_fn.0':
/<<PKGBUILDDIR>>/src/grid/cpu/grid_integrate_dgemm.c:761: undefined reference to `omp_get_num_threads'
/usr/bin/ld: /<<PKGBUILDDIR>>/src/grid/cpu/grid_integrate_dgemm.c:762: undefined reference to `omp_get_thread_num'
[...]
```

oschuett:
> it already bails out when linking grid_miniapp, it doesn't get till cp2k_info.F

While many people only build the cp2k main binary, we should try to catch the problem as early as possible - ideally before the compiler fails with cryptic message. Hence, I'll move the check into [base_uses.f90](https://github.com/cp2k/cp2k/blob/master/src/base/base_uses.f90).

src/base/base_uses.f90
```diff
@@ -12,6 +12,10 @@
                                              timeset,&
                                              timestop
 
+! Check for OpenMP early on - ideally before the compiler fails with a cryptic message.
+#if !defined(_OPENMP)
+   "OpenMP is required. Please add -fopenmp to your Fortran compiler flags."
```
mbanck, line 17: Is `#error` bad style in the Fortran code files in CP2K? The current error message gets the point across, but it IMO looks a bit nicer with `#error`:

```
   17 |    "OpenMP is required. Please add -fopenmp to your Fortran compiler flags."
      |     1
Error: Invalid character in name at (1)
```

vs.

```
   17 | #error  "OpenMP is required. Please add -fopenmp to your Fortran compiler flags."
      |  1~~~~
Error: #error "OpenMP is required. Please add -fopenmp to your Fortran compiler flags."
```

dev-zero: The preprocessor employed by Fortran is the C preprocessor in traditional mode, hence no `#error` available and some preprocessor will then choke on the unsupported preprocessor directive rather than on the error message itself (even when disabled). But this will be solved once #1259 lands.

PR #1258, opened 2020-12-28
Update to libvori 201228

mkrack:


oschuett:
After this last upgrade the tests on [i386](https://dashboard.cp2k.org/archive/i386/index.html) started failing:
```
Running tests...
Test project /opt/cp2k-toolchain/build/libvori-201228/build
    Start 1: test01
1/1 Test #1: test01 ...........................Child aborted***Exception:   0.25 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.25 sec

The following tests FAILED:
	  1 - test01 (Child aborted)
Errors while running CTest
```

I'm wondering what has changed?
@Brehministrator, maybe you could publish at least the release versions of libvori on GitHub?

mkrack:
I disabled libvori for the 32 bit armv7l regression test. I suggest to do the same for i386 for the time being.

martinbrehm:
> After this last upgrade the tests on [i386](https://dashboard.cp2k.org/archive/i386/index.html) started failing:
> 
> ```
> Running tests...
> Test project /opt/cp2k-toolchain/build/libvori-201228/build
>     Start 1: test01
> 1/1 Test #1: test01 ...........................Child aborted***Exception:   0.25 sec
> 
> 0% tests passed, 1 tests failed out of 1
> 
> Total Test time (real) =   0.25 sec
> 
> The following tests FAILED:
> 	  1 - test01 (Child aborted)
> Errors while running CTest
> ```
> 
> I'm wondering what has changed?
> @Brehministrator, maybe you could publish at least the release versions of libvori on GitHub?

@oschuett The changes were relatively minor. Maybe the largest change was that I added "-fsigned-char" to the compiler options - which should not have any effect on i386, because char is signed there by default. Unfortunately, I cannot access the log files, so I cannot find out what went wrong there. Can you send me the output of the failed test via email?

Yes, it would probably be a good idea to have the libvori on GitHub. It is currently hard-coded for CP2k and therefore not really suitable for use apart from CP2k (I will create a real "release" version of libvori during the next months). So I have to decide if I put it to GitHub already now, or if I wait until the library is ready for general use (and has at least some bits of documentation).

mkrack:
Indeed, it would be nice to have libvori on GitHub with a specific testing for various architectures and releases which could then be used by CP2K.

martinbrehm:
@oschuett @mkrack @dev-zero I was not happy with the fact that libvori would not work on 32 bit, so I took some more hours of time to track down this issue. And indeed I was successfull :-) It was really stupid (as all bugs, after you have found them): There is a data type "off_t" in C++ which stores file offsets and sizes. On 64 bit platforms, this type is always 64 bit wide. On 32 bit platforms, it is 32 bits wide - unless you define "#define _FILE_OFFSET_BITS 64", then it is 64 bits wide. I use this definition for being able to work with large files. However, by mistake, I did not include the header where I define this in all my source files. So the variable was 32 bits wide when the class constructor was called, but 64 bits wide when it shall be used... And this messed up the stack. I have fixed this issue now. The self test of the library now runs fine on my Raspberry Pi 2B (which I have now finally taken out of my cupboard). The behavior on 64 bit platforms should not be affected in any way by the changes.

You can find the new "libvori-201229.tar.gz" here:
  https://brehm-research.de/files/libvori-201229.tar.gz

It should work now on all platforms (i386, x86-64, Arm). Maybe you should give it a short try locally before updating it in the toolchain.

I also decided to put this preliminary version of the library to GitHub. I have to read a little bit how to do it :-) Should happen within the next few hours:

  https://github.com/Brehministrator

Sorry for stressing your patience so much with the libvori...

Best regards,
Martin

martinbrehm:
Libvori is already on Github now:
https://github.com/Brehministrator/libvori

martinbrehm:
Sh**t... Now there seem to be floating point exceptions in the libvori and libbqb regtests on i386... The algorithms in the library are completely built for non-signaling floating point operations (where instead #nan or #inf is stored in the variable), and checks at all relevant places if the values are valid or not. So it is clear that it will throw SIGFPE when signaling is enabled. But why is this only the case on i386? Is it by intent that CP2k enables floating point exceptions there? At startup of any POSIX process, these exceptions are masked and need to be manually switched on in the code...

See https://linux.die.net/man/3/feenableexcept

martinbrehm:
@oschuett Thanks for patching the FPE in the library. Unfortunately, there are *many* more places where this will happen in the BQB compression code (see my last comment). Why does this happen only at i386? :-/

oschuett:
We enable SIGFPE in our debug builds. So, the i386 test is now actually fine. Good job fixing that!

The reason the sdbg test is now failing is that libvori was previously [missing](https://github.com/cp2k/cp2k/pull/1263/commits/e1bcae4bc0bafe6b0588348ddbad846061c2e43b) from our Dockerfile. Hence, it wasn't actually tested by the CI until now. (I should really refactor the toolchain and remove this footgun).

> Unfortunately, there are many more places where this will happen in the BQB compression code

Ok, no problem. Then we'll simply disable the trapping before calling your code. We [already](https://github.com/cp2k/cp2k/blob/master/src/pexsi_interface.F#L382-L395) do this for other dependencies.

martinbrehm:
Thanks! Otherwise I could also change the behavior of the library similar to your patch in all the other places. But for the short term, it is safer to keep it as it is and simply disable the trapping. The risk of introducing subtle errors in the last minute is simply too high. I will take care of this within a few weeks, so that there will be no more FPEs in a future version of libvori.

oschuett:
> I will take care of this within a few weeks, so that there will be no more FPEs in a future version of libvori.

Awesome! Personally, I'm a big fan of FPEs. Once a code base is free of them they become a good debugging tool and safety net.

PR #1259, opened 2020-12-28
[WIP] CMake

dev-zero:
there are still several todos:

* [ ] decide which subdirectories will get a separate `CMakeLists.txt`. From DBCSR we see that it may be beneficial to have one `CMakeLists.txt` for a subdirectory if its features are encapsulated and it requires separate treatment with regard to dependencies. `grid/` might be a good candidate, but possibly also `common/`, etc.
* [ ] support all optional deps

Currently this PR requires a patched DBCSR which exports `dbcsr_config: get_accdrv_active_device_id` as `dbcsr_get_accdrv_active_device_id` in the API.


mtaillefumier:
grid is indeed a good candidate for a separate CmakeList (I am thinking about rocm/hip/cuda).

oschuett:
First of all, it's great that you're tackling the problem of dependency discovery. Writing ARCH files by hand has become unpractical and we can not expect everybody to use the toolchain.

I realize that this is a difficult and messy problem by nature. Hence, I'm not surprised that [FindBLAS.cmake](https://github.com/dev-zero/cp2k/blob/feature/cmake/cmake/FindBLAS.cmake) alone takes 1k lines. However, I'm a bit worried about where we are heading. CP2K has currently around 30 dependencies. To support all of those will take several thousand lines of code. So, we are setting ourselves up for a sizable and complex code base.

I think, we would be better off writing that code in Python instead of cmake because:
1. Python is much wider known, which will allow more people to help with the maintenance.
2. Python has much better tooling. Linters, unit tests, and static type analysis will greatly improve code quality.

I do realize that cmake has some convenient builtins. I'm sure we can replicate what we need using Python's large standard library.

mtaillefumier:
I would **not** DIY a build system these days. Using Python is no different from using bash to make a build system. Basically a nightmare.

CMake has its pro and con but it wildly supported with tons functions build in. 30 dependencies are nothing compared to other projects (think of X11, gnome and kde desktops, qt which can reach a few 100 when you want the full functionality). Moreover it is supported on all supercomputers or can be compiled easily.

oschuett:
> CMake has its pro and con but it wildly supported with tons functions build in.

Of course cmake has tons of functionality, but how much of it will we actually use? What we really need is missing, e.g. a discovery mechanism for BLAS.

> 30 dependencies are nothing compared to other projects (think of X11, gnome and kde desktops, qt which can reach a few 100 when you want the full functionality).

These project have other requirements (e.g. to support Windows) and a much larger developer community.

So, my worry remains: We're going to accumulate **alot** of cmake code and we will have a hard time maintaining it.

dev-zero:
roadmap here: https://github.com/dev-zero/cp2k/projects/2

dev-zero:
> > CMake has its pro and con but it wildly supported with tons functions build in.
> 
> Of course cmake has tons of functionality, but how much of it will we actually use? What we really need is missing, e.g. a discovery mechanism for BLAS.

No, it's [not](https://cmake.org/cmake/help/latest/module/FindBLAS.html), but we have a patched one for the OpenMP-enabled OpenBLAS. I've added https://github.com/dev-zero/cp2k/issues/40 as a reminder to polish the patch and send upstream.

haampie:
Wanting to roll your own dependency discovery tooling because an upstream Find*.cmake script doesn't work is a sentiment that goes counter to working towards having a unified build system for the C/C++/Fortran community.

@dev-zero is doing the right thing: keep a patched version of a Find*.cmake file around in the package repo, create an upstream PR to Kitware/CMake with the same patch, get feedback from people in the know about that PR, wait for a future CMake release which is typically the next regularly released patch version, bump the minimum required CMake version in the repo when it's out there and drop the local patch. 

That way not only CP2K users benefit, but the whole community. The community includes dependencies of CP2K itself that use CMake, so why not contribute back? And reversely you're also on the receiving end, cause CMake is rarely releasing breaking changes, and you get bug fixes to obscure problems with existing Find*.cmake on systems you never test on, instead of having people open issues here. Also do note that `spack` has excellent integration with CMake: after resolving all variants of all dependencies optimally, it will feed CMake a list of search paths of actual install prefixes it knows for sure to be correct, through `cmake -DCMAKE_PREFIX_PATH="list;of;paths"` so that dependency discovery is very unlikely to fail.

oschuett:
I understand that cmake became popular because of its strong community. However, I wonder how much CP2K will be able to profit from that community. Looking at their [list of modules](https://cmake.org/cmake/help/latest/manual/cmake-modules.7.html#find-modules) it seems we will only be using: FindMPI, FindBLAS/LAPACK, and FindCUDAToolkit. For the remaining ~26 dependencies we are on our own.

The fact that FindBLAS/LAPACK does not meet all our needs is probably the norm rather than the exception. At least I wouldn't be surprised if we have to fork more of the builtin modules. Of course, we can try to always upstream our changes, but that is just [additional work](https://github.com/Kitware/CMake/blob/master/Help/dev/review.rst) with very little incentives. Chances are we'll end up owning our forked modules rather sooner than later.

In any case, I think I made my concerns clear. You are obviously all very convinced of cmake. So, let's see how it plays out.

hfp:
Just to give my (irrelevant) comment - "it's not Python" might be the only remaining critique of CMake (look at SCons). Still, CMake is/became popular (perhaps because of being "practical" or being "multi-language" as opposed to "Boost Build" (and other build systems natively understanding header/include dependencies).

Though, GNU Make's grammar (which is not Python either) is for timeline/obvious reason (somewhat) screwed. However, the main concept of GNU Make (file time stamps) is easy enough and I never had a problem fixing things "with one more line of Makefile code".

I believe: absolute control is still best with GNU Make (own tests needed) whereas CMake starts to be more convenient if "things can be fixed" (enough customization possible despite of all the "CMake oddities"). The hardest part with CMake is to capture all of the (many) concepts (if any), i.e., even stuff "hacked into later on".

IMHO, CMake vs GNU Make comes down to "WYSIWYG" vs "Black Box" - but who likes Black Magic Boxes?

oschuett:
The cmake migration has been picked up by @mtaillefumier. He will open a new PR when ready.

Thanks a lot @dev-zero for laying the groundwork!

PR #1301, opened 2021-01-11
grid: use posix_memalign instead of deprecated memalign

dev-zero:
"modern" systems like macOS BigSur do not provide malloc.h anymore,
this is required for CP2K to build on macOS

This could potentially break the build on older systems and may require
additional #ifdef logic.

Besides, it seems there are some hardcoded alignments to 4096 which should
probably be replaced by `sysconf(_SC_PAGESIZE)` or `4*sysconf(_SC_PAGESIZE)`.
As an alternative one could use `aligned_alloc` (from C11), but that has
additional restrictions on the size.

### DIFF
--- a/src/grid/cpu/coefficients.c
+++ b/src/grid/cpu/coefficients.c
@@ -6,7 +6,6 @@
 /*----------------------------------------------------------------------------*/
 
 #include <assert.h>
-#include <malloc.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
--- a/src/grid/cpu/collocation_integration.c
+++ b/src/grid/cpu/collocation_integration.c
@@ -7,7 +7,9 @@
 
 #include <stdbool.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
 
 #ifdef __GRID_CUDA
 #include <cublas_v2.h>
@@ -41,7 +43,8 @@ struct collocation_integration_ *collocate_create_handle() {
   handle->coef_alloc_size = realloc_tensor(&handle->coef);
   handle->pol_alloc_size = realloc_tensor(&handle->pol);
 
-  handle->scratch = memalign(4096, sizeof(double) * 32768);
+  handle->scratch =
+      aligned_alloc(sysconf(_SC_PAGESIZE), sizeof(double) * 32768);
   handle->scratch_alloc_size = 32768;
   handle->T_alloc_size = 8192;
   handle->W_alloc_size = 2048;
@@ -106,7 +109,7 @@ void initialize_W_and_T(collocation_integration *const handler,
     if (handler->scratch)
       free(handler->scratch);
     handler->scratch =
-        memalign(64, sizeof(double) * handler->scratch_alloc_size);
+        aligned_alloc(64, sizeof(double) * handler->scratch_alloc_size);
     if (handler->scratch == NULL)
       abort();
   }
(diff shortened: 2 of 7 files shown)


dev-zero:
also, might be worth to check the error code, in case we provide an invalid alignment

oschuett:
LGTM from my side, but since this code is owned by @mtaillefumier we should wait for his ok.

The REF backend uses [this](https://github.com/cp2k/cp2k/blob/master/src/grid/ref/grid_ref_task_list.c#L181) to allocate aligned memory, which was introduced by @hfp.

mtaillefumier:
memalign is not deprecated but posix_memalign is probably better. I suspect they call the same code anyway. 

I checked the code. You should pass the address of the variable not the variable itself

void *p = memalign(64, sizeof(..)..) should be replaced by

posix_memalign(&p, 64, sizeof(...) * ...);

I have the tendency to use posix_memalign in my own code but I do not know why I used memalign instead. posix_memalign may require some compilation flags to be activated (-DPOSIX_....). 

Normally malloc and free should be present in stdlib.h so go ahead with these changes. I can also do them for you.

page alignment is strictly speaking not necessary. alignment to a power of 2 (64, 128, 256) is probably fine. I am not a big fan of manual alignment when std functions can do it for us. But I trust @hfp judgement on that. If  I were you I would just replace memalign by posix_memalign(&p, 64, sizeof(...)) and it is good to go.

dev-zero:
> memalign is not deprecated but posix_memalign is probably better. I suspect they call the same code anyway.

["The memalign function is obsolete and aligned_alloc or posix_memalign should be used instead. "](https://www.gnu.org/software/libc/manual/html_node/Aligned-Memory-Blocks.html)

> 
> I checked the code. You should pass the address of the variable not the variable itself

Right, forgot about that one, will fix.

> I have the tendency to use posix_memalign in my own code but I do not know why I used memalign instead. posix_memalign may require some compilation flags to be activated (-DPOSIX_....).

Doesn't seem like it, but the docs also state that if C11 is used, then one should better use `aligned_alloc` for better portability to non-POSIX systems.

> Normally malloc and free should be present in stdlib.h so go ahead with these changes. I can also do them for you.
> 
> page alignment is strictly speaking not necessary. alignment to a power of 2 (64, 128, 256) is probably fine. I am not a big fan of manual alignment when std functions can do it for us. But I trust @hfp judgement on that. If I were you I would just replace memalign by posix_memalign(&p, 64, sizeof(...)) and it is good to go.

Ok, will do that, together with replacing the `4096` with `sysconf(_SC_PAGESIZE)` calls.

mtaillefumier:
thanks for the modifications. There is a small mistake though. should be

posix_memalign(&p, alignment, size);

dev-zero:
> thanks for the modifications. There is a small mistake though. should be posix_memalign(&p, alignment, size);

yeah, I should review my own commits better, especially after a `sed` ;-)

mtaillefumier:
On 1/12/21 7:10 PM, Tiziano Müller wrote:
>
>     thanks for the modifications. There is a small mistake though.
>     should be posix_memalign(&p, alignment, size);
>
> yeah, I should review my own commits better, especially after a |sed| ;-)
>
lol happens to everyone. that's why we have PR reviews ;-)

dev-zero:
hmm, but this seems to most definitely break the mingw-build :(

dev-zero:
I don't remember anymore, can we assume C11 standard?

dev-zero:
since C99 was even retracted in favour of C11 I updated the requirements and use `aligned_alloc` now

mtaillefumier:
perfect it means I can use more c11 things in the code. Thanks.

oschuett:
The [Cuda tests](https://dashboard.cp2k.org/archive/cuda-pascal/index.html) fail to compile because there is yet another [memalign](https://github.com/cp2k/cp2k/blob/master/src/grid/hybrid/grid_collocate_hybrid.c#L469).

PR #1311, opened 2021-01-15
Darwin arch files work on Intel and ARM

fxcoudert:
Rename the arch files accordingly


fxcoudert:
The name `Darwin-IntelMacintosh-gfortran` is actually misleading, since there is no architecture-specific content in those files. CP2K compilers perfectly fine for Apple Silicon (we ship binaries for it in Homebrew already), the new ARM macOS platform.

dev-zero:
> The name `Darwin-IntelMacintosh-gfortran` is actually misleading, since there is no architecture-specific content in those files. CP2K compilers perfectly fine for Apple Silicon (we ship binaries for it in Homebrew already), the new ARM macOS platform.

Good to know, thanks!

dev-zero:
Maybe call them `Darwin-macOS-gfortran.*` to make it clearer for those less involved in software development?

fxcoudert:
`Darwin-macOS-gfortran` would be like gnu-linux (kernel and OS). So I'd rather keep `Darwin-gfortran`, or `macOS-gfortran` if you find that clearer, but not both.

dev-zero:
well, then let's leave it as this

PR #1322, opened 2021-01-21
toolchain: fix PLUMED installation

dev-zero:
... this way PLUMED picked up MKL, and GSL from the toolchain.
Still to do:

* use our OpenBLAS (will still pick up an external `libblas.so` or use an internal one)
* use fftw from MKL (but that does not seem to be supported anyway with the current toolchain)


oschuett (approved):

oschuett:
Btw, the CI still uses the old install order in [Dockerfile](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/Dockerfile). You can leave it as. I'm going to refactor it anyways to prevent this duplication in the future.

PR #1324, opened 2021-01-23
Fix CFLAGS on Darwin, misc arch changes

fxcoudert:
1. Pass `-fopenmp` in `CFLAGS` to fix the build on Darwin
2. Align Darwin arch files with other platforms:
  - I see the other `.sdbg` have been removed
  - Other gfortran-based arch files have `-g` as part of the build flags


dev-zero:
looks good, thanks a lot!

PR #1336, opened 2021-01-28
LibXC: Fix unused variable

fstein93:


fstein93:
Will be fixed in #1328 .

PR #1348, opened 2021-02-02
Add some remarks to INSTALL.md

### LINKED ISSUE #1269: INSTALL.md: remove conditionals for OpenMP, document -fallow-argument-mismatch when using MPICH and gcc-10
We are still referring to optional OpenMP support: `"If compiling with OpenMP support ..."` and possibly in other places.

To prevent issues being filed we should furthermore document that adding `-fallow-argument-mismatch` may be required when building CP2K with gcc-10 and MPICH (like on the new UZH/CSCS Eiger).

### LINKED ISSUE #1295: documentation: add minimal required versions for packages (like SIRIUS)
Hi

I tried to compile cp2k-8.1 with sirius 6.5.3 (with gfortran 10.2.0 and Cygwin on Windows 10) and got the following errors during compiling sirius_interface.f:

```
   61 |    USE sirius,                          ONLY: &
      |       1
......
  163 |       CALL sirius_create_context(sirius_mpi_comm, sctx)
      |                                                       2
Error: ‘sirius_create_context’ at (1) has a type, which is not consistent with the CALL at (2)
/home/user/cp2k-8.1/src/sirius_interface.F:275:57:

  275 |                                       spin_orbit=.FALSE.)
      |                                                         1
Error: Type mismatch in argument ‘spin_orbit’ at (1); passed LOGICAL(4) to LOGICAL(1)
/home/user/cp2k-8.1/src/sirius_interface.F:61:7:

   61 |    USE sirius,                          ONLY: &
      |       1
......
  423 |          CALL sirius_create_kset_from_grid(sctx, k_grid(1), k_shift(1), use_symmetry=.TRUE., kset_handler=ks_handler)
      |                                                                                                                     2
Error: ‘sirius_create_kset_from_grid’ at (1) has a type, which is not consistent with the CALL at (2)
/home/user/cp2k-8.1/src/sirius_interface.F:61:7:

   61 |    USE sirius,                          ONLY: &
      |       1
......
  425 |          CALL sirius_create_kset_from_grid(sctx, k_grid(1), k_shift(1), use_symmetry=.FALSE., kset_handler=ks_handler)
      |                                                                                                                      2
Error: ‘sirius_create_kset_from_grid’ at (1) has a type, which is not consistent with the CALL at (2)
/home/user/cp2k-8.1/src/sirius_interface.F:61:7:

   61 |    USE sirius,                          ONLY: &
      |       1
......
  428 |       CALL sirius_create_ground_state(ks_handler, gs_handler)
      |                                                             2
Error: ‘sirius_create_ground_state’ at (1) has a type, which is not consistent with the CALL at (2)
/home/user/cp2k-8.1/src/sirius_interface.F:109:53:

  109 |       CALL sirius_finalize(.FALSE., .FALSE., .FALSE.)
      |                                                     1
Error: Type mismatch in argument ‘call_mpi_fin’ at (1); passed LOGICAL(4) to LOGICAL(1)
/home/user/cp2k-8.1/src/sirius_interface.F:98:37:

   98 |       CALL sirius_initialize(.FALSE.)
      |                                     1
Error: Type mismatch in argument ‘call_mpi_init’ at (1); passed LOGICAL(4) to LOGICAL(1)
```

The "passed LOGICAL(4) to LOGICAL(1)" errors can be solved by replaceing ".FALSE./.TRUE." with "LOGICAL(L=.FALSE.(or .TRUE.), KIND=1)".

According to the function declaration on genetated.F90 (found in sirius/src/api. The latest file also can be seen in https://github.com/electronic-structure/SIRIUS/blob/master/src/api/generated.f90),

```
function sirius_create_context(fcomm) result(res)
implicit none
integer(C_INT), intent(in) :: fcomm
    :
END FUNCTION

function sirius_create_kset_from_grid(handler,k_grid,k_shift,use_symmetry) result(res)
implicit none
type(C_PTR), intent(in) :: handler
integer(C_INT), intent(in) :: k_grid
integer(C_INT), intent(in) :: k_shift
logical(C_BOOL), intent(in) :: use_symmetry
    :
END FUNCTION

function sirius_create_ground_state(ks_handler) result(res)
implicit none
type(C_PTR), intent(in) :: ks_handler
   :
END FUNCTION
```

the type of arguments of the functions "sirius_create_context, sirius_create_kset_from_grid and sirius_create_ground_state" are apparently
different between sirius_interface.f and declaration of sirius 6.5.3.

How could be solved the problem?

Thank you in advance.

oschuett:
Closes #1269 and closes #1295.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -37,6 +37,10 @@ cd tools/toolchain/
 Sub-points here discuss prerequisites needed to build CP2K. Copies of the
 recommended versions of 3rd party software can be downloaded from <https://www.cp2k.org/static/downloads/>.
 
+Generally, CP2K supports only one version for each of its dependencies.
+These are defined by the [toolchain scripts](./tools/toolchain/scripts/).
+Other versions might work too, but we don't test them. So, your mileage may vary.
+
 ### 2a. GNU make (required, build system)
 
 GNU make should be on your system (gmake or make on linux) and used for the build,
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -73,8 +73,10 @@ matching your compiler, and download all patches!
   - <http://math-atlas.sourceforge.net>
   - <https://www.tacc.utexas.edu/research-development/tacc-software/gotoblas2>
 
-If compiling with OpenMP support then it is recommended to use a non-threaded
-version of BLAS. In particular if compiling with MKL and using OpenMP you must
+Please note that the BLAS/LAPACK implementation used by CP2K needs to be
+OpenMP thread-safe. Examples are: the sequential variant of the Intel MKL,
+the Cray libsci, the OpenBLAS OpenMP variant and the reference BLAS/LAPACK packages.
+In particular if compiling with MKL you must
 define `-D__MKL` to ensure the code is thread-safe. MKL with multiple OpenMP
 threads in CP2K requires that CP2K was compiled with the Intel compiler.
 If the `cpp` precompiler is used in a separate precompilation step in combination
@@ -125,9 +127,8 @@ use gfortran).
 :warning: Note that on machines and compilers which support SSE you can configure
 FFTW3 with `--enable-sse2`. Compilers/systems that do not align memory (NAG f95,
 Intel IA32/gfortran) should either not use `--enable-sse2` or otherwise set the
-define `-D__FFTW3_UNALIGNED` in the arch file. When building an OpenMP parallel
-version of CP2K (ssmp or psmp), the FFTW3 threading library libfftw3_threads
-(or libfftw3_omp) is required.
+define `-D__FFTW3_UNALIGNED` in the arch file. Since CP2K is OpenMP parallelized,
+the FFTW3 threading library libfftw3_threads (or libfftw3_omp) is required.
 
 ### 2g. LIBINT (optional, enables methods including HF exchange)
 
@@ -204,8 +205,7 @@ Library ELPA for the solution of the eigenvalue problem
 
 - ELPA replaces the ScaLapack `SYEVD` to improve the performance of the diagonalization
 - A version of ELPA can to be downloaded from <http://elpa.rzg.mpg.de/software>.
-- During the installation the `libelpa.a`
-  (or `libelpa_openmp.a` if OpenMP is enabled) is created.
+- During the installation the `libelpa_openmp.a` is created.
 - Minimal supported version of ELPA is 2018.05.001.
 - Add `-D__ELPA` to `DFLAGS`
 - Add `-I$(ELPA_INCLUDE_DIR)/modules` to `FCFLAGS`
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -73,8 +77,8 @@ matching your compiler, and download all patches!
   - <http://math-atlas.sourceforge.net>
   - <https://www.tacc.utexas.edu/research-development/tacc-software/gotoblas2>
 
-If compiling with OpenMP support then it is recommended to use a non-threaded
-version of BLAS. In particular if compiling with MKL and using OpenMP you must
+Although CP2K is OpenMP parallelized it is recommended to use a non-threaded
+version of BLAS. In particular if compiling with MKL you must
```
dev-zero: Hmm, I think we should add add a note here stating that we also require the BLAS/LAPACK calls/library to be OpenMP-threadsafe (especially OpenBLAS).

INSTALL.md
```diff
@@ -37,6 +37,10 @@ cd tools/toolchain/
 Sub-points here discuss prerequisites needed to build CP2K. Copies of the
 recommended versions of 3rd party software can be downloaded from <https://www.cp2k.org/static/downloads/>.
 
+Generally, CP2K supports only one version for each of its dependencies.
+These are defined by the [toolchain scripts](./tools/toolchain/scripts/).
+Other versions might work too, but we don't test them. So, your mileage may vary.
+
```
dev-zero, line 43: I hoped we'd provide the information in a more user-friendly, especially when users want to reuse for example minor versions of libraries they have already installed. And for things like BLAS/LAPACK/ScaLAPACK this also does not generally apply, does it?

INSTALL.md
oschuett: Good point. How should we word this? Also, doesn't threadsafe usually imply that it's threaded?

INSTALL.md
oschuett: I think, it's transparent and honest. We only test one combination of versions and we shouldn't make any further promises.

> And for things like BLAS/LAPACK/ScaLAPACK this also does not generally apply, does it?

Same here. We only test one combination of versions. I do agree that these libraries are very stable and there is a high chance that other versions will work too, but we can't be certain.

INSTALL.md
dev-zero: > Good point. How should we word this?

Please note that the BLAS/LAPACK implementation used by CP2K needs to be OpenMP thread-safe. Examples are: the sequential variant of the Intel MKL, the Cray libsci, the OpenBLAS OpenMP variant and the reference BLAS/LAPACK packages.

> Also, doesn't threadsafe usually imply that it's threaded?

No, see the reference LAPACK/BLAS implementation (just hope someone verified that at some point), the sequential MKL.

oschuett: Sounds good. I've adopted your paragraph.

PR #1352, opened 2021-02-05
Machine: Enable flushing by default

oschuett:


ltalirz:
Thanks for the quick fix!

While I agree with the new default behavior, I wonder whether there were concrete use cases that led to the previous default for performance reasons (perhaps also not that easy to detect?)?
I also notice that [many regtests and benchmarks already use the keyword](https://github.com/cp2k/cp2k/search?p=19&q=flush_should_flush) - which could be a solution for the regtests and benchmarks, but would of course still result in users ending up with less helpful output files after a crash.

Would it be possible to trigger a flush when an abort is requested?

oschuett:
> While I agree with the new default behavior, I wonder whether there were concrete use cases that led to the previous default for performance reasons (perhaps also not that easy to detect?)?

I actually only changed the behavior for the short time during startup before the input is parsed. Afterwards, the value gets [copied](https://github.com/cp2k/cp2k/blob/7162243fa0d0286adeedbd8b229a8f6d0d3ac67d/src/environment.F#L543) from the [input](https://manual.cp2k.org/trunk/CP2K_INPUT/GLOBAL.html#FLUSH_SHOULD_FLUSH), which defaults to TRUE as well.

> Would it be possible to trigger a flush when an abort is requested?

Yes, that's a good idea. I included it in #1355.

ltalirz:
> I actually only changed the behavior for the short time during startup before the input is parsed.

Thanks for the clarification!

> Afterwards, the value gets copied from the input, which defaults to TRUE as well.

I see - looks like Joost introduced it in 2008 with default `FALSE` but then changed it to `TRUE` one year later ;-)
https://github.com/cp2k/cp2k/search?q=FLUSH_SHOULD_FLUSH&type=commits

> Yes, that's a good idea. I included it in #1355.

Great!

PR #1362, opened 2021-02-16
Updated SIRIUS to latest official version

mtaillefumier:
Updating SIRIUS to the latest version

PR #1365, opened 2021-02-17
Toolchain: Compile SPLA only when MPI is available

oschuett:
Followup to #1362.


dev-zero:
could you also fix the typo in the `spla` description?

PR #1366, opened 2021-02-17
precommit: Add shfmt to format shell scripts

oschuett:


dev-zero:
could you please add that to `.pre-commit-config.yml` as well?

PR #1368, opened 2021-02-18
Space group symmetry

pcazade:
@ltalirz @oschuett This is a follow up PR to implement Leopold's suggestions.  I added a more complete description of the space group feature in the KEEP_SPACE_GROUP information.

### DIFF
--- a/src/start/input_cp2k_motion.F
+++ b/src/start/input_cp2k_motion.F
@@ -869,18 +869,10 @@ CONTAINS
       ! collects keywords related to Space Group Symmetry
       CALL keyword_create( &
          keyword, __LOCATION__, name="KEEP_SPACE_GROUP", &
-         description="Enables keeping the space group symmetry during optimization."// &
-         "The space group is automatically detected and preserved."// &
-         "The ability to identify the space group depends on the quality"// &
-         "and accuracy of the input coordinates and of the cell parameters."// &
-         "PDB files usually are not accurate enough."// &
-         "EPS_SYMMETRY keyword can be used to adjust the detection threshold."// &
-         "Ranges of atoms can be excluded from it via the repeatable keyword SYMM_EXCLUDE_RANGE."// &
-         "SYMM_REDUCTION allows to take into account the effect of a polarization electric on the symmetry."// &
-         "It typically reduces it. The vector should have the same direction has the electric field."// &
-         "The amplitude does not matter."// &
-         "Finally, space group symmetry is only applied on the coordinates, the forces, and the stress tensor."// &
-         "It does not affect and/or reduce the computational cost of energy evaluation.", &
+         description="Detect space group of atoms and preserve during optimization."// &
+         "The space group symmetry is applied to coordinates, forces, and the stress tensor."// &
+         "It does not affect/reduce computational cost."//&
+         "Use EPS_SYMMETRY to adjust the detection threshold.", &
          usage="KEEP_SPACE_GROUP .TRUE.", &
          default_l_val=.FALSE., lone_keyword_l_val=.TRUE., repeats=.FALSE.)
       CALL section_add_keyword(section, keyword)
@@ -898,7 +890,9 @@ CONTAINS
       ! collects keywords related to reduction of symmetry due to an external field
       CALL keyword_create( &
          keyword, __LOCATION__, name="SYMM_REDUCTION", &
-         description="Polarization vector of the electric field for space group symmetry reduction", &
+         description="Direction of the external static electric field."\\ &
+         "Some symmetry operations are not compartible with the dircetion of an electric field."\\ &
+         "These operations are used when enforcing the space group.", &
          usage="SYMM_REDUCTION  0.0 0.0 0.0", &
          repeats=.FALSE., n_var=3, &
          type_of_var=real_t, default_r_vals=(/0.0_dp, 0.0_dp, 0.0_dp/))
@@ -908,7 +902,9 @@ CONTAINS
       ! collects keywords related to ranges of atoms to symmetrize
       CALL keyword_create( &
          keyword, __LOCATION__, name="SYMM_EXCLUDE_RANGE", &
-         description="Range of atoms to exclude from space group symmetry. This keyword can be repeated.", &
+         description="Range of atoms to exclude from space group symmetry."\\ &
+         "These atoms are excluded from both identification and enforcement."\\ &
+         "This keyword can be repeated.", &
          repeats=.TRUE., usage="SYMM_EXCLUDE_RANGE {Int} {Int}", type_of_var=integer_t, n_var=2)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
--- a/src/start/input_cp2k_motion.F
+++ b/src/start/input_cp2k_motion.F
@@ -869,8 +869,19 @@ CONTAINS
       ! collects keywords related to Space Group Symmetry
       CALL keyword_create( &
          keyword, __LOCATION__, name="KEEP_SPACE_GROUP", &
-         description="Keep the space group symmetry operation during optimization.", &
-         usage="KEEP_SPACE_GROUP yes", &
+         description="Enables keeping the space group symmetry during optimization."// &
+         "The space group is automatically detected and preserved."// &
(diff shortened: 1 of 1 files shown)


pcazade:
@oschuett I should have done a git rebase, I presume.

ltalirz:
just create a new branch from master and  `git cherry-pick` your latest two commits on top

then force-push to your fork, overwriting this branch

src/start/input_cp2k_motion.F
```diff
@@ -869,16 +869,27 @@ RECURSIVE SUBROUTINE create_geoopt_section(section, location, label, description
...
+         "and accuracy of the input coordinates and of the cell parameters."// &
+         "PDB files usually are not accurate enough."// &
+         "EPS_SYMMETRY keyword can be used to adjust the detection threshold."// &
+         "Ranges of atoms can be excluded from it via the repeatable keyword SYMM_EXCLUDE_RANGE."// &
+         "SYMM_REDUCTION allows to take into account the effect of a polarization electric on the symmetry."// &
+         "It typically reduces it. The vector should have the same direction has the electric field."// &
+         "The amplitude does not matter."// &
+         "Finally, space group symmetry is only applied on the coordinates, the forces, and the stress tensor."// &
+         "It does not affect and/or reduce the computational cost of energy evaluation.", &
+         usage="KEEP_SPACE_GROUP .TRUE.", &
```
ltalirz:
```suggestion
         description="Detect space group of atoms and preserve during optimization."// &
         "The space group symmetry is applied to coordinates, forces, and the stress tensor."// &
         "It does not affect/reduce computational cost."//&
          "Use EPS_SYMMETRY to adjust the detection threshold.", &
         usage="KEEP_SPACE_GROUP .TRUE.", &
```

ltalirz (commented):
thanks @pcazade !

a few questions/comments from my side

src/start/input_cp2k_motion.F
```diff
@@ -869,16 +869,27 @@ RECURSIVE SUBROUTINE create_geoopt_section(section, location, label, description
...
          default_l_val=.FALSE., lone_keyword_l_val=.TRUE., repeats=.FALSE.)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
       ! collects keywords related to precision for finding the space group
       CALL keyword_create( &
          keyword, __LOCATION__, name="EPS_SYMMETRY", &
-         description="Accuracy required for symmetry detection", &
+         description="Accuracy required for the identification of the space group.", &
```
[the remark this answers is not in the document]
pcazade: @ltalirz Yes, EPS_SYMMETRY is used for symprec in Spglib, but it is also used later on by my routine to build the array of equivalent atoms. Regarding PDB, basically EPS_SYMMETRY should be larger that 1e-3. However, I would argue against as making the threshold easier to pass may lead to identify symmetries that are not there or finding atoms as being equivalent, when they are not. Basically, PDB should not be used for symmetry. I know, I tried a lot. Coordinates with at least 5-6 digits are best.

src/start/input_cp2k_motion.F
```diff
@@ -887,7 +898,7 @@ RECURSIVE SUBROUTINE create_geoopt_section(section, location, label, description
       ! collects keywords related to reduction of symmetry due to an external field
       CALL keyword_create( &
          keyword, __LOCATION__, name="SYMM_REDUCTION", &
-         description="Polarisation vector of electric field for symmetry reduction", &
+         description="Polarization vector of the electric field for space group symmetry reduction", &
```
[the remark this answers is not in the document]
pcazade: @ltalirz This is indeed intended for calculations with an applied electric field. The former partially breaks the symmetry and this needs to be taken into account. It is not very elegant but I did not know how to easily access this information in FIST and QS. Furthermore, the electric field is defined differently in FIST and QS. In one, it is a fractional (scaled) vector, and in the other it is given directly in cartesian coordinates. Finally, having this option in SPGR, let users break/lower the symmetry if they want/need to.

src/start/input_cp2k_motion.F
```diff
@@ -897,23 +908,23 @@ RECURSIVE SUBROUTINE create_geoopt_section(section, location, label, description
       ! collects keywords related to ranges of atoms to symmetrize
       CALL keyword_create( &
          keyword, __LOCATION__, name="SYMM_EXCLUDE_RANGE", &
-         description="Range of atoms to exclude from symmetry. This keyword can be repeated.", &
+         description="Range of atoms to exclude from space group symmetry. This keyword can be repeated.", &
```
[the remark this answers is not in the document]
pcazade: @ltalirz These atoms are excluded form both identification and enforcement. This is something I needed to study protein crystals where the position of water molecules is not know and most likely not symmetric.

src/start/input_cp2k_motion.F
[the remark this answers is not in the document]
pcazade: @ltalirz Further clarifications here. The space group that is identified is the same with or without the electric field. However, the former forbids some symmetry operations of that space group. This keyword is used to determine which of the space group symmetry operations are forbidden to prevent them from being enforced.

src/start/input_cp2k_motion.F
```diff
@@ -869,16 +869,19 @@ RECURSIVE SUBROUTINE create_geoopt_section(section, location, label, description
...
+         usage="KEEP_SPACE_GROUP .TRUE.", &
          default_l_val=.FALSE., lone_keyword_l_val=.TRUE., repeats=.FALSE.)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
       ! collects keywords related to precision for finding the space group
       CALL keyword_create( &
          keyword, __LOCATION__, name="EPS_SYMMETRY", &
-         description="Accuracy required for symmetry detection", &
+         description="Accuracy required for the identification of the space group.", &
```
ltalirz: Thanks for the clarification!

Looking at the [definition of symprec](https://spglib.github.io/spglib/variable.html#variables-symprec), it seems to me that it actually has a unit - namely the unit of the atomic vectors you pass to spglib.

Are you passing them in in Angstroms?
If yes, then EPS_SYMMETRY should probably get a unit, with Angstrom per default.

Perhaps something like
```suggestion
         description="Accuracy for space group determination. Roughly speaking, two atomic positions v1, v2 are considered identical if |v1 - v2| < EPS_SYMMETRY. ", &
```

I think that should then be enough information for people to figure out by themselves how much precision they have in their structures (i.e. we may not need to mention PDB specifically).

pcazade: @ltalirz EPS_SYMMETRY is dimensionless. It is the difference between the scaled position vectors which is compared to it.

ltalirz: Thanks, that is good to know for users - it means that doubling a cell can make atoms equivalent that were not considered equivalent before (and in a very large cell, very far away atoms can be considered equivalent).

In any case, I think as long as it is documented that's fine.

ltalirz:
Thanks for the updates @pcazade - good to merge from my side!

src/start/input_cp2k_motion.F
pcazade: @ltalirz No problem with supercell. Spglib detects that very well. Basically, it adds more translation operations. Rotations are not affected. So atoms that are equivalent in the original cell will also be equivalent with their translation images. I tested this extensively as k-points don't work well (if at all, sorry but there's a problem there in cp2k), and supercells are required.

src/start/input_cp2k_motion.F
```diff
@@ -869,16 +869,20 @@ RECURSIVE SUBROUTINE create_geoopt_section(section, location, label, description
       ! collects keywords related to Space Group Symmetry
       CALL keyword_create( &
          keyword, __LOCATION__, name="KEEP_SPACE_GROUP", &
-         description="Keep the space group symmetry operation during optimization.", &
-         usage="KEEP_SPACE_GROUP yes", &
+         description="Detect space group of atoms and preserve during optimization."// &
```
oschuett: Please add a space at the end of each continued line. The `//` operator just concatenates strings.

pcazade: @oschuett Done.

PR #1373, opened 2021-02-22
Motion: Use existing printkeys for space group output

oschuett:
@pcazade, please take a look at this PR. It fixes an issue that our [farming test](https://dashboard.cp2k.org/archive/farming/index.html) found. The problem is that the space group output gets written to [default_output_unit](https://github.com/cp2k/cp2k/blob/master/src/motion/space_groups_types.F#L131) instead of using a [printkey](https://www.cp2k.org/dev:printkey). 

In this PR I'm changing the output to the exiting [PROGRAM_RUN_INFO](https://manual.cp2k.org/trunk/CP2K_INPUT/MOTION/GEO_OPT/PRINT/PROGRAM_RUN_INFO.html) and [STRESS](https://manual.cp2k.org/trunk/CP2K_INPUT/MOTION/PRINT/STRESS.html) printkeys. Furthermore, I'm removing the `SPGR_FILE_NAME` keyword and renaming `SPGR_PRINT_LEVEL` into `SPGR_PRINT_ATOMS`.

The output of [PROGRAM_RUN_INFO](https://manual.cp2k.org/trunk/CP2K_INPUT/MOTION/GEO_OPT/PRINT/PROGRAM_RUN_INFO.html) can still be redirected to a file. Please let me know if this works for you or if I should rather create a separate printkey?


pcazade:
@oschuett This looks fine to me. Regarding printing in a separate file, I introduced it as this list of operations and equivalent atoms can be big, for example when having a system with 100000 atoms... Is redirecting with PROGRAM_RUN_INFO enough?

pcazade:
@oschuett Another question. Would your changes fix the fact that sprg_print is done multiple times when using the psmp version? It did not happen in my initial version of the space_group feature, but it is something I noticed locally after all the changes we made.

oschuett:
> Is redirecting with PROGRAM_RUN_INFO enough?

Yes, every printkey can be redirected to a separate file:

```
&CELL_OPT
  SPGR_PRINT_ATOMS T
  &PRINT
    &PROGRAM_RUN_INFO
      FILENAME lengthy_output.txt
    &END
  &END
&END
```

> Would your changes fix the fact that sprg_print is done multiple times when using the psmp version? 

Yes, it does.

Btw, I ended up not using the  [STRESS](https://manual.cp2k.org/trunk/CP2K_INPUT/MOTION/PRINT/STRESS.html) printkey, because it's meant for a trajectory. Instead, I'm sending everything to [PROGRAM_RUN_INFO](https://manual.cp2k.org/trunk/CP2K_INPUT/MOTION/GEO_OPT/PRINT/PROGRAM_RUN_INFO.html).

pcazade:
@oschuett Thanks for your answers. Hopefully, your changes will also address the issues with python and AiiDA. I received email about errors with these. However, it is unclear to me, what is the error and therefore, how to fix it.

oschuett:
> Hopefully, your changes will also address the issues with python and AiiDA.

Those are not related to your PR. I'll take care of them.

PR #1375, opened 2021-02-23
Added: Quotations around directories

FMcil:
Added quotations to ${INSTALLDIR}/tsan.supp, ${INSTALLDIR}/lsan.supp and ${SETUPFILE} to allow pathnames with spaces.


oschuett:
Thanks for this PR.

However, are you sure that `install_gcc.sh` is the only affected file? A quick check suggests otherwise:

```
cd cp2k/tools/toolchain
shellcheck install_cp2k_toolchain.sh scripts/*.sh scripts/stage*/*.sh | grep SC2086 | wc -l
405
```

PR #1387, opened 2021-03-01
grid: Use __attribute__((always_inline))

oschuett:


oschuett:
I'm using `#pragma GCC unroll` and ` __attribute__((always_inline))` to force the compiler into generating specialized version of hotspot functions. This seems to be a viable alternative to C++ templates.

While these constructs are not part of the C standard, the Intel compiler apparently supports both.

@hfp, @alazzaro what do you think?

hfp:
> I'm using `#pragma GCC unroll` and ` __attribute__((always_inline))` to force the compiler into generating specialized version of hotspot functions. This seems to be a viable alternative to C++ templates.
> 
> While these constructs are not part of the C standard, the Intel compiler apparently supports both.
> 
> @hfp, @alazzaro what do you think?

I would wrap them both up into macros including the `__attribute__` "keyword". It does not have to be [sophisticated aka complicated](https://github.com/hfp/libxsmm/blob/master/include/libxsmm_macros.h#L127). Btw, applying/using such macros that expand into pragmas is without sequence point (semicolon).

Out of curiosity, can you measure an advantage using these constructs for the code in question?

alazzaro:
I agree with @hfp . We have CCE and NAG compiler, so wrapping is a safe solution.
I'm also interested in the performance benefit.

BTW, how does [libgrid](https://github.com/cp2k/cp2k/tree/master/tools/autotune_grid) work with the new code?

oschuett:
I already merged something [very similar](https://github.com/cp2k/cp2k/commit/ada15fe29628d70f225d509f6983088db0f685a9), which speed up collocate/integrate by 20% - 30%. The only problem is that it's brittle because it relies on the compiler to do the inlining voluntarily.

> We have CCE and NAG compiler, so wrapping is a safe solution.

Since we're talking about C code here, are those compilers still relevant? I'd expect people use one of gcc, icc, or clang.
I also couldn't find any documentation about which pragmas and attributes are supported by CCE and NAG.

So, I'd suggest that for now I only wrap the `unroll` pragma, because GCC has a different syntax, but keep the `always_inline` attribute until we know for a fact that a compiler doesn't support it.

> BTW, how does libgrid work with the new code?

It obsolete, I'll remove it. We could do auto tuning again in C, but I think our time is better spend adding `ivdep` pragmas manually.

alazzaro:
So, you are right about CCE, which is actually now based on clang... I don't know about nagfor, but I would assume that your assumption is fair enough.

Triggered by your latest changes, I did executed the Daint performance tests and indeed I see some performance benefit (especially on the MC partition) Great!

PR #1388, opened 2021-03-02
small bug fixes

mtaillefumier:
this PR contains some small fixes for spla (toolchain) and the dgemm backend.

### DIFF
--- a/tools/toolchain/install_cp2k_toolchain.sh
+++ b/tools/toolchain/install_cp2k_toolchain.sh
@@ -309,8 +309,8 @@ with_spglib="__INSTALL__"
 with_hdf5="__DONTUSE__"
 with_elpa="__INSTALL__"
 with_libvdwxc="__DONTUSE__"
-with_spfft="__INSTALL__"
-with_spla="__INSTALL__"
+with_spfft="__DONTUSE__"
+with_spla="__DONTUSE__"
 with_cosma="__INSTALL__"
 with_libvori="__INSTALL__"
 # for MPI, we try to detect system MPI variant
@@ -619,6 +619,9 @@ while [ $# -ge 1 ]; do
     --with-libvori*)
       with_libvori=$(read_with $1)
       ;;
+    --with-spla*)
+      with_spla=$(read_with $1)
+      ;;
     --help*)
       show_help
       exit 0
@@ -746,6 +749,7 @@ fi
 # SIRIUS dependencies. Remove the gsl library from the dependencies if SIRIUS is not activated
 if [ "$with_sirius" = "__INSTALL__" ]; then
   [ "$with_spfft" = "__DONTUSE__" ] && with_spfft="__INSTALL__"
+  [ "$with_spla" = "__DONTUSE__" ] && with_spla="__INSTALL__"
   [ "$with_gsl" = "__DONTUSE__" ] && with_gsl="__INSTALL__"
   [ "$with_libxc" = "__DONTUSE__" ] && with_libxc="__INSTALL__"
   [ "$with_fftw" = "__DONTUSE__" ] && with_fftw="__INSTALL__"
--- a/src/grid/cpu/grid_collocate_dgemm.c
+++ b/src/grid/cpu/grid_collocate_dgemm.c
@@ -245,7 +245,7 @@ void apply_sphere_cutoff_ortho(struct collocation_integration_ *const handler,
   int **map = handler->map;
   map[1] = map[0] + 2 * cmax + 1;
   map[2] = map[1] + 2 * cmax + 1;
-  memset(map[0], 0xff, sizeof(int) * 3 * (2 * cmax + 1));
+  // memset(map[0], 0xff, sizeof(int) * 3 * (2 * cmax + 1));
 
   for (int i = 0; i < 3; i++) {
     for (int ig = 0; ig < handler->cube.size[i]; ig++) {
@@ -336,7 +336,7 @@ void apply_spherical_cutoff_generic(
   int **map = handler->map;
   map[1] = map[0] + 2 * cmax + 1;
   map[2] = map[1] + 2 * cmax + 1;
-  memset(map[0], 0xff, sizeof(int) * 3 * (2 * cmax + 1));
+  //	memset(map[0], 0xff, sizeof(int) * 3 * (2 * cmax + 1));
 
   for (int i = 0; i < 3; i++) {
     for (int ig = 0; ig < handler->cube.size[i]; ig++) {
@@ -476,8 +476,8 @@ void collocate_l0(double *scratch, const double alpha, const bool orthogonal_xy,
 
 /* compute the following operation (variant) it is a tensor contraction
 
-   V_{kji} = \sum_{\alpha\beta\gamma} C_{\alpha\gamma\beta} T_{2,\alpha,i}
-   T_{1,\beta,j} T_{0,\gamma,k}
+         V_{kji} = \sum_{\alpha\beta\gamma} C_{\alpha\gamma\beta} T_{2,\alpha,i}
+         T_{1,\beta,j} T_{0,\gamma,k}
 
 */
 void tensor_reduction_for_collocate_integrate(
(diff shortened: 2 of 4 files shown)


tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -57,178 +57,178 @@ OPTIONS:
 
 -h, --help                Show this message.
 -j <n>                    Number of processors to use for compilation, if
-                          this option is not present, then the script
-                          automatically tries to determine the number of
-                          processors you have and try to use all of the
-                          processors.
+													this option is not present, then the script
```
oschuett: Looks like some unintended reformatting slipped in.

hfp:
Just a note: I am getting crashes ("invalid memory references") with the CPU backend for instance when running H2O-64 using an PSMP binary. the stacktrace looks like:

```text
#6  0x0000000004292bd4 in dgemm_simplified (m=0x2aaab1e73a50) at /nfs/home2/hpabst/cp2k/src/grid/cpu/utils.c:56
#7  0x0000000004295a68 in rotate_to_cartesian_harmonics (ibasis=0x1cd13710, jbasis=0x1cd13710, iatom=32, jatom=44, iset=0, jset=0, block=0x242e2ae0,
    work=0x2aaab1e73d50, pab=0x2aaab1e73c30) at /nfs/home2/hpabst/cp2k/src/grid/cpu/grid_collocate_dgemm.c:99
#8  0x00000000042990dd in extract_blocks (ctx=0x1cbc9a50, task=0x2aacb042f9e0, pab_blocks=0x1ce6cc50, work=0x2aaab1e73d50, pab=0x2aaab1e73c30)
    at /nfs/home2/hpabst/cp2k/src/grid/cpu/grid_collocate_dgemm.c:1002
#9  0x0000000004299139 in compute_coefficients (ctx=0x1cbc9a50, handler=0x208fc040, previous_task=0x0, task=0x2aacb042f9e0, pab_blocks=0x1ce6cc50,
    pab=0x2aaab1e73c30, work=0x2aaab1e73d50, pab_prep=0x2aaab1e73e70) at /nfs/home2/hpabst/cp2k/src/grid/cpu/grid_collocate_dgemm.c:1015
#10 0x0000000004299a79 in collocate_one_grid_level_dgemm._omp_fn.0 () at /nfs/home2/hpabst/cp2k/src/grid/cpu/grid_collocate_dgemm.c:1182
```

For the above stacktrace, I merged this PR (but it certainly happens also with CP2K 8.1 or master revision). The problem seems to be independent of whether BLAS or LIBXSMM is used (in both cases I am getting these invalid memory references).

Can you reproduce this as well, or maybe I am doing something wrong?

```text
&GLOBAL
  ...
  &GRID
    BACKEND CPU
  &END
  ...
&END GLOBAL
```

Above, the default CUTOFF setting (`FALSE`) is used but `APPLY_CUTOFF TRUE` leads to the same issue.

mtaillefumier:
@hfp : thanks for reporting this issue. I will look into it. I never had this issue before I also do not think this is coming from mkl or libxsmm so I suspect that the changes I made did not solve the initial problem. 
```

&GRID
    BACKEND CPU
  &END
```

means that the cut-off is applied. I choose to set it on default because I wanted to run the regtests. I am sorry to ask but could you rerun the example with APPLY_CUTOFF FALSE. I think I know where the initial issue is but I want to be sure it is actually there. 

The fact that it crashes exactly in `rotate_to_cartesian_harmonics` is really surprizing because this path is taken all the time independently on mpi or not. I think the memory is corrupted and the corruption propagates through the runs.

mtaillefumier:
I ran the the same test locally and I get a crash when I do **not** apply the cut-off. If I relax the constraint then it crash for some obscur reasons. It is mpi related but this code never call mpi which means the problem comes from the way I treat the grid splitting. I never encountered this problem before. 

Could you postpone the merge ?

mtaillefumier:
I found the issue and fixed it. I run the same example than for good measure with different numbers of MPI ranks (both ortho and non-ortho with and without cutoff) and all pass.

hfp:
> 
> 
> I found the issue and fixed it. I run the same example than for good measure with different numbers of MPI ranks (both ortho and non-ortho with and without cutoff) and all pass.

Awesome!  I will try it asap and leave a note here. Sorry, for not replying earlier (I noticed your questions now but believe there is no open question left ;-).

mtaillefumier:
No hurries though. I would still be interested in by a test on your side. You can safely ignore my questions :). Two pairs of eyes are better than one.

hfp:
Unfortunately, I am still getting these invalid memory references/crashes. I tested with multiple ranks/threads (PSMP debug variant, no libxsmm, no libxc, no libint, no elpa, etc.), and it is not related to the rank-count but rather fails with multiple threads. So, things seem to work correctly in a single-threaded fashion (no crash).

mtaillefumier:
Do you use mkl or openblas ?

mtaillefumier:
When I run it with libxsmm+mkl+mpi and none of the other dependencies, I have absolutely no issues. The multithreaded version is no different from the single threaded one. Each thread has its own scratch memory which is allocated locally, and all data that are shared between the threads (they could be independent processes like I would treat the multigpu case)  are considered read only. The only thing I could see is the call to cblas_functions routines. 

I guess valgrind might be able to catch where the mistake comes from but I do have any crash whatsoever with the minimal setup. Of course I use mkl + libxsmm which are known to be thread safe. original cblas implementation should also work.

Do you have any errors output of simply crashes ?

hfp:
I was linking against MKL. I also put unlimited stack size (just in case). Did you build your binary with OpenMP/flag? As I said (PSMP), this might be related.

mtaillefumier:
> I was linking against MKL. I also put unlimited stack size (just in case). Did you build your binary with OpenMP/flag? As I said (PSMP), this might be related.

That's interesting. something else must be wrong. stask size is not unlimited because I do not use the stack for tables only local variables.

I have this in CFLAGS

$(DFLAGS) -g -O2 \
	-march=skylake-avx512 -O2 -g -ftree-loop-vectorize -ftree-vectorize -fopenmp-simd -fopenmp

mtaillefumier:
I recompiled with mkl only and everything seems to work fine. I see no difference if I compile with libxsmm + mkl or mkl whatsoever. 

It also run on my local machine (Fedora 33) and centos 8. My arch file is the following. There is nothing really outstanding in this configuration.

```
CC          = gcc
CXX         = g++
AR          = ar -r
FC          = mpif90
LD          = mpif90
#
DFLAGS      = -D__LIBXSMM -D__MKL  -D__parallel  -D__MKL -D__SCALAPACK -D__FFTW3  -D__HDF5 -D__SPGLIB
#
WFLAGS      = -Werror=aliasing -Werror=ampersand -Werror=c-binding-type -Werror=intrinsic-shadow -Werror=intrinsics-std -Werror=line-truncation -Werror=tabs -Werror=target-lifetime -Werror=underflow -Werror=unused-but-set-variable -Werror=unused-variable -Werror=unused-dummy-argument -Werror=conversion -Werror=zerotrip -Wno-maybe-uninitialized -Wuninitialized -Wuse-without-only
#
FCDEBFLAGS  = -fbacktrace -ffree-form -fimplicit-none -std=f2008
CFLAGS      = -march=native -mtune=native \
	        -g  -O2 \
	        -funroll-loops $(PROFOPT) \
	        -fopenmp -fopenmp-simd \
	        -I/usr/include/openmpi-x86_64 \
	        -m64 \
	        -I/opt/intel/oneapi/mkl/2021.1.1/env/../include \
	        -I/usr/include/libxsmm -std=c11 -Wall -Wextra $(DFLAGS)
FCFLAGS     = -march=native -mtune=native -fno-omit-frame-pointer -g  -O2 -fopenmp -I/usr/include/openmpi-x86_64 -m64 -I'/usr/include/libxsmm' $(FCDEBFLAGS) $(WFLAGS) $(DFLAGS)
#
LDFLAGS     = $(FCFLAGS) -Wl,--enable-new-dtags -pthread -Wl,-rpath -Wl,/usr/lib64/openmpi/lib -Wl,--enable-new-dtags -L/usr/lib64/openmpi/lib    -L'/usr/lib64' -Wl,-rpath='/usr/lib64' -L'/home/mathieut/Projects/codes/cp2k/tools/toolchain/install/libxsmm-1.16.1/lib' -Wl,-rpath='/home/mathieut/Projects/codes/cp2k/tools/toolchain/install/libxsmm-1.16.1/lib' -L'/usr/lib64' -Wl,-rpath='/usr/lib64'
LIBS        = -lsymspg -lxsmmf -lxsmm -ldl -lpthread -lfftw3_mpi -lfftw3 -lfftw3_omp   -lmpi_cxx -lmpi  /opt/intel/oneapi/mkl/2021.1.1/env/../lib/intel64/libmkl_scalapack_lp64.so -Wl,--start-group /opt/intel/oneapi/mkl/2021.1.1/env/../lib/intel64/libmkl_gf_lp64.so /opt/intel/oneapi/mkl/2021.1.1/env/../lib/intel64/libmkl_sequential.so /opt/intel/oneapi/mkl/2021.1.1/env/../lib/intel64/libmkl_core.so /opt/intel/oneapi/mkl/2021.1.1/env/../lib/intel64/libmkl_blacs_openmpi_lp64.so -Wl,--end-group -lpthread -lm -ldl -lstdc++
#
FYPPFLAGS   = -n --line-marker-format=gfortran5

```

hfp:
Let's sort this out next week. I will go for your ARCH file and check.

mtaillefumier:
> Let's sort this out next week. I will go for your ARCH file and check.

perfect. I can not reproduce this error anymore.

PR #1389, opened 2021-03-03
LibXC 5.1.2 and add extra LibXC sections

fstein93:
This PR is a follow up to #914 . 

- bumps LibXC to 5.1.2 fixing a memory leak in the Fortran interface and allows to
- add one section to the XC_FUNCTIONALS section per LibXC functional
- LIBXC section is deprecated and will be removed later.

Can anyone of the admins add the new tarball to the static downloads, please? Then I will change the download link again.


oschuett:
> Can someone of the admins add the new tarball to the static downloads, please? 

Voilà: https://www.cp2k.org/static/downloads/libxc-5.1.2.tar.gz

fstein93:
Does anyone have an idea why the CI fails to build Sirius?

oschuett:
To debug this you could temporarly remove the output redirection in [install_sirius.sh:141](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/stage8/install_sirius.sh#L141).

My guess would be that SIRIUS is not compatible with libxc-5.1.2.

oschuett:
While cmake swallowed the actual compiler [output](https://storage.googleapis.com/cp2k-ci/run-cp2k-sdbg-6c92a13a_report.txt), one of the failing files is `potential/xc.cpp`. Hence, it seems that SIRIUS 7.0.2 is indeed incompatible with libxc 5.1.2. So much for semantic versioning :-/

fstein93:
I wonder what exactly makes it fail. I can build CP2K locally with SIRIUS. I am using the arch files of the toolchain. The tests don't work though.

fstein93:
I was comparing the two failing reports. It is probably something else because in the new report, potential/potential.cpp does not fail It is probably just by accident (now it is potential/xc.cpp). I am rebuilding the toolchain from scratch now.

oschuett:
> I wonder what exactly makes it fail.

The container runs out of memory. For some reason it launches up to 62 instances of `cc1plus` in parallel although the machine has only 32 vCPUs. Maybe cmake ignores the `-j $NPROC` argument?

oschuett:
Oh, I know what's going on. The `$NPROCS` variable is set in [install_cp2k_toolchain.sh](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/install_cp2k_toolchain.sh#L342) and then stored in the Docker cache. This CI run reuses layers that were created on a node with 64 vCPUs. I only recently added those for the new [performance test](https://dashboard.cp2k.org/archive/perf-openmp/index.html).

I'll take care of this.

oschuett:
There is a merge conflict now. Could you please remove your changes to `install_sirius.sh`?

fstein93:
Sure. I am just accelerating some of the regtests ( #174 )

PR #1391, opened 2021-03-03
expand macro arguments

hfp:
Fixed GRID_PRAGMA_UNROLL(N) macro which needs one more macro expansion to populate final N-expression (e.g., GRID_PRAGMA_UNROLL(MAX_LP_OPTIMIZED + 1)).

Other:

* Avoid warnings about unused variables (due to __SPGLIB conditional compilation).
* Avoid warning about OUT dummy arg without assigned value (unused variable).

### DIFF
--- a/src/grid/common/grid_common.h
+++ b/src/grid/common/grid_common.h
@@ -8,11 +8,12 @@
 #define GRID_COMMON_H
 
 #define GRID_STRINGIFY(SYMBOL) #SYMBOL
+#define GRID_EXPAND(ARG) ARG
 
 #if defined(__GNUC__)
-#define GRID_PRAGMA_UNROLL(N) _Pragma(GRID_STRINGIFY(GCC unroll N))
+#define GRID_PRAGMA_UNROLL(N) _Pragma(GRID_STRINGIFY(GRID_EXPAND(GCC unroll N)))
 #else
-#define GRID_PRAGMA_UNROLL(N) _Pragma(GRID_STRINGIFY(unroll(N)))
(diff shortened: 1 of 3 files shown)


oschuett:
I also just opened #1390 to fix the build problems.

I don't think another macro expansion is needed because my `GRID_PRAGMA_UNROLL` macro does not itself contain any macros - unlike your [LIBXSMM_PRAGMA_UNROLL_N](https://github.com/hfp/libxsmm/blob/master/include/libxsmm_macros.h#L405).

hfp:
> I also just opened #1390 to fix the build problems.
> 
> I don't think another macro expansion is needed because my `GRID_PRAGMA_UNROLL` macro does not itself contain any macros - unlike your [LIBXSMM_PRAGMA_UNROLL_N](https://github.com/hfp/libxsmm/blob/master/include/libxsmm_macros.h#L405).

I tested this with GCC 9.3 and it solved it. I think you need to expand `N` (unroll level) which is not done my the stringification (but needs one more expansion).

hfp:
However, CI failure tells the story is not over and probably #1390 is more appropriate. Let me take the macro-"fix" out (or close this PR), and resend the two minor issues here as separate PR.

oschuett:
> I tested this with GCC 9.3 and it solved it.

I think, the [minimal](https://dashboard.cp2k.org/archive/minimal-sdbg/index.html) test fails because `-O0` does not propagate the constants aggressively enough. Hence, I switched from `const int` to a macro.

Issue #1323, opened 2021-01-22
Slower performance in cp2k-8.1


JuntingYu1996:
I built the newest cp2k-8.1 and previous cp2k-7.1 on my cluster using intel-2018-update3 compiler (including icc, ifort, intel-mpi and MKL). Their porformances on a simple BOMD task are tested, and timing information are attached. It seems version 8.1 is slower than 7.1. I also noticed that in version 8.1, the grid part in src is rewritten in C, and the SUBROUTINEs **grid_integrate_task_list** and **grid_collocate_task_list** take up a lot of time, while in version 7.1 the SUBROUTINEs **integrate_v_rspace** and **calculate_rho_elec** are faster. 

Is there any suggestions that the performance of version 8.1 can be improved?

[8.1_timing.txt](https://github.com/cp2k/cp2k/files/5854234/8.1_timing.txt)

[7.1_timing.txt](https://github.com/cp2k/cp2k/files/5854233/7.1_timing.txt)

BTW, the arch files for compiling 8.1 and 7.1 are almost the same, except that in 8.1 arch file, the following line should be added accounting for grid part written in C.

CFLAGS += -std=c11 -qopenmp -I$(LIBXSMM)/include

oschuett:
The collocate and integrate routines were indeed re-implemented in C before GPU support was added. Generally, we believe that the performance of the new implementation is [comparable](https://github.com/cp2k/cp2k/issues/1084#issuecomment-718779047) to the old one.

Maybe you can tell us a bit more about your setup, e.g. which basis set and how many OpenMP threads are you using?

JuntingYu1996:
TZV2P-GTH basis sets with GTH pseudopotentials are used for all atoms. The xc functional is SCAN. 
In the SCF part, I used OT/IRAC algorithm with DIIS minimizer.
I used 112 MPI processes, with 1 OpenMP thread per MPI process for running cp2k.

hfp:
> I used 112 MPI processes, with 1 OpenMP thread per MPI process for running cp2k.

Is this running MPI ranks on HT threads or is this an AP-platform?

JuntingYu1996:
> > I used 112 MPI processes, with 1 OpenMP thread per MPI process for running cp2k.
> 
> Is this running MPI ranks on HT threads or is this an AP-platform?

[...]

PR #1398, opened 2021-03-05
grid: Vectorize inner loop of ortho kernel

oschuett:


alazzaro:
Would be possible to use the OpenMP simd directive?

oschuett:
Ideally, we shouldn't need any pragmas. Unfortunately, GCC insists on alias checks in collocate despite `restict` pointers:

```C
static inline void ortho_reg_to_grid(const int ig,
                                     const double * restrict reg,
                                     double * restrict grid_0,
                                     double * restrict grid_1) {
  grid_0[ig] += reg[0];
  grid_1[ig] += reg[1];  // suspected aliasing
}

...

ortho_reg_to_grid(ig, reg, &grid[foo], &grid[bar])

```

For integrate we accumulate into registers and the vectorization kicks in by itself.

> Would be possible to use the OpenMP simd directive?

Yes, I tried that it. For collocate the timings get worse, because the vectorization fails for some reason.
For integrate the timings improve. However, within integrate we actually do have a data dependency when all simd lanes accumulate into the same coefficients. Hence, the resulting assembly might actually be incorrect, although I haven't seen any wrong results yet.

To summaries, I currently get the following timings:

|                 | collocate | integrate |
|-----------------|-----------|-----------|
| current master  |    0.826  |    0.657  |
| without pragma  |    0.725  |    0.627  |
| pragma ivdep    |    0.668  |    0.627  |
| pragma omp simd |    0.820  |    0.586  |

oschuett:
I was tired of haggling with the compiler over auto-vectorization and decided to use [Intel Intrinsics](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#) instead. This feels like a much more robust and future-proof approach. For collocate I managed to match the previous performance, for integrate I made further progress thanks to [this](https://stackoverflow.com/a/10454420) nifty reduction scheme.

dev-zero:
> I was tired of haggling with the compiler over auto-vectorization and decided to use [Intel Intrinsics](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#) instead. This feels like a much more robust and future-proof approach.

Ok, so the "reference" implementation will not run on ARM?
What is the oldest CPU supported, resp. which CPU flags are needed?

oschuett:
> What is the oldest CPU supported, resp. which CPU flags are needed?

There is a fall back implementation in plain C. It's used for the tail iterations (because masked load/store is slow) and in case AVX2 is not available: 

https://github.com/cp2k/cp2k/blob/a1ba7540e90340c4c0e8c4fc83946c2304328f90/src/grid/ref/grid_ref_collint.h#L191

dev-zero:
But shouldn't the whole `ortho_cx_to_grid_avx2` be wrapped in a `#ifdef __AVX2__` to avoid compilation on non-AVX2 architectures?

hfp:
Quick note, you may also want `__FMA__`. Indeed, GNU Compiler does not cover FMA under "AVX2". In turn, the compiler needs `-mavx2 -mfma` in order to cover your intrinsic code (another option is to use "embrella" flags like `-march=core-avx2`).

alazzaro:
So, I would suggest keeping the optimized C code without intrinsic for any CPU (yes, for sure ARM, but also some old x86 CPUs). That means the version of the code which is currently in the master, which is already faster than v8.1 (at least, accordingly to the performance test on the [dashboard](https://dashboard.cp2k.org/archive/cray-xc50-daint-mc-psmp_perf/index_full.html)). Let's give the possibility to some compilers to auto-vectorize (Intel, CCE are better than GNU in this respect).

Then, you can add your intrinsics version as a specialization (just like ELPA does for some specific kernels).

Concerning `#omp simd`, just to be sure, have you tried all clauses, for example the reduction one?

https://www.openmp.org/spec-html/5.0/openmpsu42.html

Then, in the code above:

```
static inline void ortho_reg_to_grid(const int ig,
                                     const double * restrict reg,
                                     double * restrict grid_0,
                                     double * restrict grid_1) {
  grid_0[ig] += reg[0];
  grid_1[ig] += reg[1];  // suspected aliasing
}

...

ortho_reg_to_grid(ig, reg, &grid[foo], &grid[bar])
```

Well, `grid_0` and `grid_1` points to the same array, so the compiler is applying a conservative approach. Here `bar - foo > simdlen`.  Maybe a `safelen = bar - foo` can solve the problem? (From what I can see in the current master, the code is a bit more complicated...)

oschuett:
> So, I would suggest keeping the optimized C code without intrinsic for any CPU

The new C implementation is already an improvement, because unlike the current master it does allow for auto-vectorization. It's just that with intrinsics I could squeeze out even more.

> Well, grid_0 and grid_1 points to the same array, so the compiler is applying a conservative approach.

I eventually realized that the compiler was right and the vectors can alias in rare cases. Still, I think there is some kind of bug where gcc loses the `restrict` property during the inline. Fortunately, with intrinsics we are no longer at the mercy of the compiler.

alazzaro:
> > So, I would suggest keeping the optimized C code without intrinsic for any CPU
> 
> The new C implementation is already an improvement, because unlike the current master it does allow for auto-vectorization. It's just that with intrinsics I could squeeze out even more.

Well, then even better... Let's keep the C w/ and w/o intrinsics version then...

hfp:
> Concerning `#omp simd`, just to be sure, have you tried all clauses, for example the reduction one?

Good point!  One thing to note, OpenMP SIMD can be toggled separately from OpenMP Multicore. Unlike the latter (`_OPENMP`), OpenMP SIMD is not covered by a preprocessor flag and may even be disabled if `_OPENMP` is signaled. In LIBXSMM, I am applying `-fopenmp-simd` accompanied by some own preprocessor symbol. In addition, some compilers enable it by default (which can be carved out as well). Further, the OpenMP standard version and/or compiler version (because of the flag) must be checked. It's rather [convoluted](https://github.com/hfp/libxsmm/blob/master/include/libxsmm_macros.h#L364).

Another (sometimes) effective mechanism is to apply [omp (simd) collapse](https://github.com/hfp/libxsmm/blob/master/include/libxsmm_macros.h#L381). Intel Compiler for instance has only very few patterns that must be exactly matched to trigger outer loop vectorization (in contrast to CCE); however, it can be triggered manually with "(omp) simd collapse" (Intel Compiler has the prestandard/non-OpenMP/legacy simd pragma).

oschuett:
Thanks for your input. I think, for now I'll stick with the Intrinsics. Let's see how it goes.

oschuett:
I'm actually now seeing a 5-10% speedups for collocate/integrate compared to cp2k 7.1. The remaining performance gap is due to cp_fm_cholesky_invert.

The sudden increase in memory usage is due to https://github.com/cp2k/cp2k/commit/522da0eaa280a6dfc9d06c7610ea67380f06840e. However, the old Fortran implementation also used to hold on to its thread-local grids. I don't understand why we're getting a higher usage now.

dev-zero:
> The sudden increase in memory usage is due to [522da0e](https://github.com/cp2k/cp2k/commit/522da0eaa280a6dfc9d06c7610ea67380f06840e). However, the old Fortran implementation also used to hold on to its thread-local grids. I don't understand why we're getting a higher usage now.

Maybe related: try `calloc` instead of the combination of `malloc`+`memset`. Can't find the source anymore, but afaik this is what is recommended on Linux to avoid reinitializing already zeroed pages (as returned by the kernel), which in turn can delay actual allocation.

Ok, found it here: https://stackoverflow.com/a/2688522/1400465

oschuett:
>  Daint doesn't use COSMA, so I have a fair comparison with 7.1.

And indeed looking at the [data](https://dashboard.cp2k.org/archive/cray-xc50-daint-mc-psmp_perf/plot_data.txt) from the [Daint MC](https://dashboard.cp2k.org/archive/cray-xc50-daint-mc-psmp_perf/index_full.html) benchmarks shows that the timings have recovered:

| Column                         | CP2K 7.1                       | Current Master                 |
| ------------------------------ | ------------------------------ | ------------------------------ |
| commit                         | 5cf83a5d6180fc12a5949044cee04c | 41c7a35766121a1ffef85fc6116f50 |
| h2o_32_ri_mp2_rpa/10           | 125.069538                     | 131.280252                     |
| h2o_32_ri_mp2_rpa/13           | 152.280124                     | 149.493314                     |
| h2o_64_ls_nrep3/7              | 209.203333                     | 210.456667                     |
| h2o_64_md/1                    | 28.749667                      | 27.282333                      |
| h2o_64_md_threads/21           | 27.107000                      | 28.806000                      |
| h2o_64_md_threads/23           | 40.531000                      | 36.763000                      |
| h2o_64_md_threads/25           | 33.980000                      | 32.583000                      |
| h2o_64_md_threads/27           | 35.186000                      | 36.140000                      |
| h2o_64_md_threads/29           | 39.032000                      | 38.472000                      |
| h2o_64_md_threads/31           | 33.891000                      | 34.169000                      |
| h2o_64_md_threads/33           | 46.282000                      | 42.509000                      |
| h2o_64_md_threads/35           | 50.528000                      | 47.213000                      |
| h2o_64_md_threads/37           | 65.133000                      | 57.055000                      |

> I've dropped GRID_CUDA too, the execution was much much slower...

Yes, the cuda kernels are next on my list.

> try calloc instead of the combination of malloc+memset.

Thanks for the hint! I'll give it a try.

mkrack:
I obtained the following timings for the 256 H2O benchmark input using either two "daint-mc" or two "daint-gpu" nodes:
<table>
 <thead>
  <tr>
   <th rowspan="2">CP2K version</th>
   <th colspan="2">orthorhombic</th>
   <th colspan="2">non-orthorhombic</th>
  </tr>
  <tr>
   <td>mc</td>
   <td>gpu</td>
   <td>mc</td>
   <td>gpu</td>
  </tr>
 </thead>
 <tbody>
  <tr>
   <td>7.1</td>
   <td>343</td>
   <td>500</td>
   <td>471</td>
   <td>803</td>
  </tr>
  <tr>
   <td>8.1</td>
   <td>358</td>
   <td>556</td>
   <td>476</td>
   <td>845</td>
  </tr>
  <tr>
   <td>9.0 (Mar 11)</td>
   <td>340</td>
   <td>471</td>
   <td>469</td>
   <td>845</td>
  </tr>
 </tbody>
</table>

CP2K 9.0 was compiled with the current arch files CRAY-X40_gnu.psmp and CRAY-XC50_gnu.psmp (no cosma, no openMP for ELPA).
The current CP2K trunk version recovers or even improves performance-wise w.r.t. CP2K v7.1 except for non-orthorhombic grids using the daint-gpu binary.

alazzaro:
Assuming that the code doesn't use the GPU for the GRID (at least I don't see the flag), so we are using the same code for MC and GPU for non-orthorhombic,  and looking at the numbers, I would say that the result on MC is the same as of 8.1 (7 seconds can be considered a fluctuation), so the slowness doesn't depend on the GPU or MC...

hfp:
Regarding "9.0 (Mar 11)", do we plan to have an 8.2 or jump into 9.1 at some point? Or better asked - would it be EOY or earlier?

( Note, above question does not suggest one or the other solution at this point; it is just asking if we want to plan for a release sooner than end of 2021 )

juerghutter:
I would suggest to release a version 8.2 in the near future (March/April).

Universität Zürich                   E-mail: ***@***.***
Winterthurerstrasse 190
CH-8057 Zürich, Switzerland

-----"Hans Pabst" ***@***.***> wrote: -----
To: "cp2k/cp2k" ***@***.***>
From: "Hans Pabst" ***@***.***>
Date: 03/11/2021 01:09PM
Cc: "Subscribed" ***@***.***>
Subject: Re: [cp2k/cp2k] grid: Vectorize inner loop of ortho kernel (#1398)


 Regarding "9.0 (Mar 11)", do we plan to have an 8.2 or jump into 9.1 at some point? Or better asked - would it be EOY or earlier?
 ( Note, above question does not suggest one or the other solution at this point; it is just asking if we want to plan for a release sooner than end of 2021 )

mkrack:
I would also vote for a 8.2 release end of April or in early summer when @oschuett has finished his work on the collocate/integrate part.

oschuett:
Yes, end of April sounds good. I still would like to implement [Mathieu's trick]( https://github.com/cp2k/cp2k/issues/1084#issuecomment-718835986) for the non-ortho kernel. Furthermore, I now have access to contemporary GPUs, which hopefully allows me to speedup the cuda integrate kernel.

hfp:
Thank you all for the response!

I believe the "grid exercise" shows several interesting points:

1. Fortran is able to deliver reasonable performance purely because of the language (given the journey of the grid code adopting C even if the code generation/backend is the same as for C/C++). C appears to be more verbose/explicit but also more sensitive to the formulation of the code. We adopted C because wanted it explicit and we finally got what we asked for.
2. My 2nd objection is somewhat emphasized, but we were able to deliver value (supposedly valued by users) due to improved performance. We ended up with [Intrinsics](https://software.intel.com/sites/landingpage/IntrinsicsGuide/), but the equation was "more effort delivered some (more) return". I think this is somewhat of an incremental/feasible approach at least for the CPU side.

I am in tension if the last point turns true for the GPU as well (which is by no means emphasized). I think this case can guide future "exercises". Even a mixed approach like adopting C/Intrinsics (ISO_C_BINDING) at kernel-level can be an interesting direction if users get faster time to solution.

oschuett:
I'm also rather satisfied with the migration. 

Long term the biggest win is probably the size reduction. The [old implementation](https://github.com/cp2k/cp2k/commit/fc2c8f188986c03016895bd90c3e2d68160acd1c) clocked in at 44k LoC, while the [new implementation](https://github.com/cp2k/cp2k/blob/master/src/grid/ref/grid_ref_collint.h) has only 1k LoC. This means the code can be adopted to future architecture with reasonable effort.

mkrack:
@alazzaro Right, the timings for the non-orthorhombic case have not changed from 8.1 to 9.0 for daint-mc and daint-gpu:
<table>
<thead>
<tr>
<th rowspan="3">Routine</th>
<th colspan="6">orthorhombic (256 H<sub>2</sub>O)</th>
<th colspan="6">non-orthorhombic (256 H<sub>2</sub>O)</th>
</tr>
<tr>
<th colspan="3">daint-mc</th>
<th colspan="3">daint-gpu</th>
<th colspan="3">daint-mc</th>
<th colspan="3">daint-gpu</th>
</tr>
<tr>
<th >7.1</th>
<th >8.1</th>
<th >9.0</th>
<th >7.1</th>
<th >8.1</th>
<th >9.0</th>
<th >7.1</th>
<th >8.1</th>
<th >9.0</th>
<th >7.1</th>
<th >8.1</th>
<th >9.0</th>
</tr>
</thead>
<tbody>
<tr>
<td>CP2K</td>
<td >343</td>
<td >358</td>
<td >340</td>
<td >500</td>
<td >556</td>
<td >471</td>
<td >471</td>
<td >476</td>
<td >469</td>
<td >803</td>
<td >845</td>
<td >845</td>
</tr>
<tr>
<td>collocate</td>
<td >29</td>
<td >36</td>
<td >24</td>
<td >77</td>
<td >88</td>
<td >61</td>
<td >94</td>
<td >99</td>
<td >99</td>
<td >245</td>
<td >256</td>
<td >257</td>
</tr>
<tr>
<td>integrate</td>
<td >26</td>
<td >41</td>
<td >25</td>
<td >67</td>
<td >103</td>
<td >63</td>
<td >88</td>
<td >109</td>
<td >108</td>
<td >237</td>
<td >284</td>
<td >281</td>
</tr>
</tbody>
</table>

mtaillefumier:
> @alazzaro Right, the timings for the non-orthorhombic case have not changed from 8.1 to 9.0 for daint-mc and daint-gpu:
> Routine 	orthorhombic (256 H2O) 	non-orthorhombic (256 H2O)
> daint-mc 	daint-gpu 	daint-mc 	daint-gpu
> 7.1 	8.1 	9.0 	7.1 	8.1 	9.0 	7.1 	8.1 	9.0 	7.1 	8.1 	9.0
> CP2K 	343 	358 	340 	500 	556 	471 	471 	476 	469 	803 	845 	845
> collocate 	29 	36 	24 	77 	88 	61 	94 	99 	99 	245 	256 	257
> integrate 	26 	41 	25 	67 	103 	63 	88 	109 	108 	237 	284 	281

Did anyone test the dgemm backend. It is far more generic that combo intrinsic / manual loop unrolling.

mkrack:
@mtaillefumier How do you test the dgemm backend?

alazzaro:
> @mtaillefumier How do you test the dgemm backend?

@mkrack It is something in my to-do list:

https://github.com/cp2k/cp2k/issues/1323#issuecomment-781117672

mtaillefumier:
> @mtaillefumier How do you test the dgemm backend?

You have to modify the input file and add this in the **global** section (unfortunately)

&GRID
BACKEND CPU
APPLY_CUTOFF FALSE
&END

The backend name is misleading actually we should probably call it dgemm (because it uses dgemm underneath). I rely heavily on optimized small dgemm (although I could not see much difference between mkl and mkl + libxsmm).

The CPU backend is usually faster than the reference backend but thanks to Ole work they are now close to each other for the orthorhombic case. 

I also have a completely different implementation of the same thing which shows serious improvements but it is unfinished business.

mkrack:
@mtaillefumier Here are the results for the `BACKEND cpu` compared to the default, i.e. without `&GRID` section:
<table>
<thead>
<tr>
<th>System</th>
<th colspan="4">orthorhombic (256 H<sub>2</sub>O)</th>
<th colspan="4">non-orthorhombic (256 H<sub>2</sub>O)</th>
</tr>
<tr>
<th>Partition</th>
<th colspan="2">daint-mc</th>
<th colspan="2">daint-gpu</th>
<th colspan="2">daint-mc</th>
<th colspan="2">daint-gpu</th>
</tr>
<tr>
<th>Backend</th>
<th >default</th>
<th >cpu</th>
<th >default</th>
<th >cpu</th>
<th >default</th>
<th >cpu</th>
<th >default</th>
<th >cpu</th>
</tr>
</thead>
<tbody>
<tr>
<td>CP2K 9.0 (Mar 11)</td>
<td >340</td>
<td >926</td>
<td >471</td>
<td >1452</td>
<td >469</td>
<td >Abort</td>
<td >845</td>
<td >Abort</td>
</tr>
</tbody>
</table>

The runs are either much slower or even abort with `KS energy is an abnormal value (NaN/Inf)` if the `BACKEND cpu` is selected.
I guess that my CP2K trunk binaries are not properly compiled for that backend choice.

mtaillefumier:
thanks @mkrack. 

That's very interesting actually. I never had such bad timings. I usually use these options to compile cp2k (note that I do not compile everything)

```
CC          = gcc
CXX         = g++
AR          = ar -r
FC          = mpif90
LD          = mpif90
#
DFLAGS      = -D__LIBXSMM -D__MKL  -D__parallel  -D__MKL -D__SCALAPACK -D__FFTW3  -D__HDF5 -D__SPGLIB
#
WFLAGS      = -Werror=aliasing -Werror=ampersand -Werror=c-binding-type -Werror=intrinsic-shadow -Werror=intrinsics-std -Werror=line-truncation -Werror=tabs -Werror=target-lifetime -Werror=underflow -Werror=unused-but-set-variable -Werror=unused-variable -Werror=unused-dummy-argument -Werror=conversion -Werror=zerotrip -Wno-maybe-uninitialized -Wuninitialized -Wuse-without-only
#
FCDEBFLAGS  = -fbacktrace -ffree-form -fimplicit-none -std=f2008
CFLAGS      = -DNDEBUG -march=native -mtune=native \
	        -g  -O2 \
	        -funroll-loops $(PROFOPT) \
	        -fopenmp -fopenmp-simd \
	        -I/usr/include/openmpi-x86_64 \
	        -m64 \
	        -I/opt/intel/oneapi/mkl/2021.1.1/env/../include \
	        -I/usr/include/libxsmm -std=c11 -Wall -Wextra $(DFLAGS)
FCFLAGS     = -march=native -mtune=native -fno-omit-frame-pointer -g  -O2 -fopenmp -I/usr/include/openmpi-x86_64 -m64 -I'/usr/include/libxsmm' $(FCDEBFLAGS) $(WFLAGS) $(DFLAGS)
#
LDFLAGS     = $(FCFLAGS) -Wl,--enable-new-dtags -pthread -Wl,-rpath -Wl,/usr/lib64/openmpi/lib -Wl,--enable-new-dtags -L/usr/lib64/openmpi/lib    -L'/usr/lib64' -Wl,-rpath='/usr/lib64' -L'/home/mathieut/Projects/codes/cp2k/tools/toolchain/install/libxsmm-1.16.1/lib' -Wl,-rpath='/home/mathieut/Projects/codes/cp2k/tools/toolchain/install/libxsmm-1.16.1/lib' -L'/usr/lib64' -Wl,-rpath='/usr/lib64'
LIBS        = -lsymspg -lxsmmf -lxsmm -ldl -lpthread -lfftw3_mpi -lfftw3 -lfftw3_omp   -lmpi_cxx -lmpi  /opt/intel/oneapi/mkl/2021.1.1/env/../lib/intel64/libmkl_scalapack_lp64.so -Wl,--start-group /opt/intel/oneapi/mkl/2021.1.1/env/../lib/intel64/libmkl_gf_lp64.so /opt/intel/oneapi/mkl/2021.1.1/env/../lib/intel64/libmkl_sequential.so /opt/intel/oneapi/mkl/2021.1.1/env/../lib/intel64/libmkl_core.so /opt/intel/oneapi/mkl/2021.1.1/env/../lib/intel64/libmkl_blacs_openmpi_lp64.so -Wl,--end-group -lpthread -lm -ldl -lstdc++
```
So it means that my last PR was not ready to be merged apparently. There is still a problem in the code.

hfp:
Just a side note, linking FFTW (`-lfftw3_mpi -lfftw3 -lfftw3_omp`) can badly interact with MKL's link line since the FFTW interface is part of MKL. We currently discuss a feature for MKL to offer a flavor without FFTW interface (which is not contrary to including it by default). I think I cleaned installer/toolchain a while ago to disable linking FFTW when MKL is linked, but I am not sure.

mtaillefumier:
thanks @hfp. It is good to know. the arch file was generated by the toolchain.

hfp:
( I reported issue #1413 )

mkrack:
@mtaillefumier Thanks for sharing your compilation flags.
The regression testing of your grid implementation needs to be improved obviously, since your last PR passed. To this end, I added a new regression test folder [QS/regtest-grid](https://github.com/mkrack/cp2k/tree/master/tests/QS/regtest-grid).  My [related PR](https://github.com/cp2k/cp2k/pull/1414) is, however, expected to fail, because one test will show a memory leak and the other will fail with NaN.

mkrack:
@mtaillefumier Is there any information printed about the &GRID section settings in the CP2K output?

oschuett:
There is a grid statistics printed at the end:
```
 -                                GRID STATISTICS                              -
 LP    KERNEL             BACKEND                              COUNT     PERCENT
 0     collocate general  REF                                      4       8.33%
 3     integrate general  REF                                      4       8.33%
 0     collocate general  GPU                                      4       8.33%
 3     integrate general  GPU                                      4       8.33%
 . . .
```
While the dgemm and hybrid backends are not yet hooked up they would be rather easy to [add](https://github.com/cp2k/cp2k/blob/master/src/grid/ref/grid_ref_collint.h#L922).

mkrack:
@oschuett Yes, I know that GRID output. It is very nice and useful, but it comes only at the end of the run. I like to have a very short echoing of the input settings at the start of the run, e.g. in the part labelled with `GLOBAL|`, just as an early confirmation that the code recognized the user requests.

mtaillefumier:
> @mtaillefumier Is there any information printed about the &GRID section settings in the CP2K output?

@mkrack no I never thought about it. By the way, libxsmm is not enough if you want to measure performances, you also need openblas or mkl. 
@oschuett I will drop the hybrid backend (preparing a PR just for that). I also have a couple of changes for the gpu backend that remove atomicAdd partially.

mkrack:
> @mkrack no I never thought about it. By the way, libxsmm is not enough if you want to measure performances, you also need openblas or mkl.

We need first a correctly working code without leaks. Then we can bother about performance tuning.

oschuett:
> I like to have a very short echoing of the input settings at the start of the run, 

@mkrack Right, that would be useful. I'll add it.

>I also have a couple of changes for the gpu backend that remove atomicAdd partially.

@mtaillefumier Yes, let's join forces. I'm looking forward to your improvements.

mkrack:
After #1415, #1416, #1417, and #1418 fixed the runtime failure of the new `QS/regtest-grid` [tests](https://github.com/cp2k/cp2k/tree/master/tests/QS/regtest-grid), the timings for the 256 H2O benchmark case on [Piz Daint](https://www.cscs.ch/computers/piz-daint) could also be collected for the `&GRID BACKEND cpu` (dgemm):
<table>
<thead>
<tr>
<th>System</th>
<th colspan="4">orthorhombic (256 H<sub>2</sub>O)</th>
<th colspan="4">non-orthorhombic (256 H<sub>2</sub>O)</th>
</tr>
<tr>
<th>Partition</th>
<th colspan="2">daint-mc</th>
<th colspan="2">daint-gpu</th>
<th colspan="2">daint-mc</th>
<th colspan="2">daint-gpu</th>
</tr>
<tr>
<th>Backend</th>
<th >default</th>
<th >cpu</th>
<th >default</th>
<th >cpu</th>
<th >default</th>
<th >cpu</th>
<th >default</th>
<th >cpu</th>
</tr>
</thead>
<tbody>
<tr>
<td>CP2K 7.1</td>
<td >343</td>
<td >-</td>
<td >500</td>
<td >-</td>
<td >471</td>
<td >-</td>
<td >803</td>
<td >-</td>
</tr>
<tr>
<td>CP2K 8.1</td>
<td >358</td>
<td >-</td>
<td >556</td>
<td >-</td>
<td >476</td>
<td >-</td>
<td >845</td>
<td >-</td>
</tr>
<tr>
<td><a href="https://github.com/cp2k/cp2k/commit/dc07142f76999ba1529714096e2a5a20386f381c">CP2K 9.0 (Mar 13)</a></td>
<td >325</td>
<td >388</td>
<td >463</td>
<td >627</td>
<td >451</td>
<td >391</td>
<td >847</td>
<td >686</td>
</tr>
</tbody>
</table>

The `BACKEND cpu` (with APPLY_CUTOFF no) shows similar timings for the orthorhombic and non-orthorhombic case and thus is much faster than the default case for non-orthorhombic cells which makes it the better choice for such systems.

mtaillefumier:
Thanks. That's a first step.

Vectorization with -march=native or at least -mavx2 is really needed
because the bottleneck functions are memory bound. Moreover using
-DNDEBUG will also have an effect. I will try by myself on daint but
thanks again for running these tests.

mkrack:
@mtaillefumier The CP2K binaries, which I used to generate the timings, were built using `cc` and `ftn` and thus optimization for the respective hardware should already be included, i.e. `-march=broadwell -mtune=broadwell` for daint-mc and `-march=core-avx2 -mtune=core-avx2` for daint-gpu. I doubt that `-DNDEBUG` will have a significant impact on the timings.

mtaillefumier:
thanks for the information. It is appreciated. I forgot the wrappers
most probably turn on these options by default.

PR #1406, opened 2021-03-09
fixed error in install_libvdwxc.sh

hfp:


oschuett:
How did `shfmt` report the problem?

When I run `shfmt_v3.2.2_linux_amd64  -ci -w -sr -i=2 install_libvdwxc.sh` it just silently exits with code zero.

hfp:
I will paste the output asap.

Meanwhile a note: the pre-commit hook does not seem to achieve a reproducible environment as it requires go-lang (like yet another Python hell). For instance, I am running Debian which defaults to Go 1.11 as installed with `apt` (Go 1.16 is the current release), and the pre-commit hook fails to setup/download components because of Go (it's not worth to mention the "go.mod" complaints and other non-sense about GOROOT).

oschuett:
> the pre-commit hook does not seem to achieve a reproducible environment

That is the reason why I built [this](https://github.com/cp2k/cp2k/tree/master/tools/precommit#install-git-hook) alternative precommit system that runs all those exotic tools in the cloud (or locally in a Docker container if you prefer).

hfp:
I have collected the output of `pre-commit run --all-files 2>&1` on current/clean master (without this PR). There is a lot of [output](https://github.com/cp2k/cp2k/files/6114223/precommit.log) (warnings), but the error is shown below.

```text
Check Yaml................................................................Passed
Check for added large files...............................................Passed
Check python ast..........................................................Passed
Reformat Python files with the black code formatter.......................Passed
markdownlint..............................................................Passed
shellcheck................................................................Failed
- hook id: shellcheck
- exit code: 1
[...]
In tools/toolchain/scripts/stage7/install_libvdwxc.sh line 15:
if [ "$MPI_MODE" = "no" ] && [ $with_sirius ="__FALSE__" ]; then
                               ^----------^ SC2154: with_sirius is referenced but not assigned.
                               ^----------^ SC2086: Double quote to prevent globbing and word splitting.
                                             ^-- SC1035: You are missing a required space here.
[...]
Run doxify................................................................Passed
Run prettify..............................................................Passed
Run analyze_src...........................................................Passed
Regenerate checked-in data files..........................................Passed
clang-format..............................................................Passed
Reformat shell scripts with shfmt.........................................Passed
```

I am using shellcheck 0.5.0.

hfp:
> How did `shfmt` report the problem?

I misspelled this by saying `shfmt` while the error is triggered by `shellcheck`.

dev-zero:
ah, I see it: the exclusion regex in `.pre-commit-config.yml` does not match the new subfolder structure of the the toolchain scripts. Since I didn't have time to fix all the issues in all the scripts I enabled shellcheck only for new files.

oschuett:
I'll take care of shellcheck and the tab detection.

hfp:
@oschuett Thank you!

PR #1414, opened 2021-03-12
Add regression test inputs for the &GRID section

mkrack:


hfp:
I am still looking into these errors about [invalid memory references](https://github.com/cp2k/cp2k/pull/1388#issuecomment-791417435) (even with H2O-64 input). Very nice to have some tests now!

oschuett:
Thanks for adding more grid regtests. Note, that we already have a few [here](https://github.com/cp2k/cp2k/blob/master/tests/LIBTEST/TEST_FILES#L36-L39).

mkrack:
@oschuett Obviously, SZV and DZVP basis sets do not show the problem. It is important to test also larger basis sets.

oschuett:
> It is important to test also larger basis sets.

I totally agree. I was thinking that it might be better to move the existing tests into your new directory to have them all in one place.

mkrack:
@oschuett Feel free to move the tests. Maybe, it is better to have a separate test folder for these tests to distribute the workload, since I expected that we will need more and longer running test cases.

hfp:
> 
> 
> I am still looking into these errors about [invalid memory references](https://github.com/cp2k/cp2k/pull/1388#issuecomment-791417435) (even with H2O-64 input). Very nice to have some tests now!

Sorry, I hooked this PR up a bit... I found some potential problem: @mtaillefumier [this code](https://github.com/cp2k/cp2k/blob/master/src/grid/cpu/grid_collocate_dgemm.c#L1207) seems to be wrong (similar for integrate) since a barrier must be discovered by all threads. The cpu backend consistently crashes for me if multiple threads are involved. One reason for not being able to reproduce such problem can be related to the GNU OpenMP runtime revision which depends on the compiler version (I am using GCC 9.3.0).

mtaillefumier:
> > I am still looking into these errors about [invalid memory references](https://github.com/cp2k/cp2k/pull/1388#issuecomment-791417435) (even with H2O-64 input). Very nice to have some tests now!
> 
> Sorry, I hooked this PR up a bit... I found some potential problem: @mtaillefumier [this code](https://github.com/cp2k/cp2k/blob/master/src/grid/cpu/grid_collocate_dgemm.c#L1207) seems to be wrong (similar for integrate) since a barrier must be discovered by all threads. The cpu backend consistently crashes for me if multiple threads are involved. One reason for not being able to reproduce such problem can be related to the GNU OpenMP runtime revision which depends on the compiler version (I am using GCC 9.3.0).

Hmm. The barrier is correct here. it is a reduction over all threads but instead of having a critical section I split the table in pieces such that each thread can add data at different locations in parallel. The omp barrier call just say wait for all threads to finish their work before starting the next block.  

I using GCC 10.2.1 on my desktop and gcc 8.3.0 on our test cluster. If you think there is an issue with openmp then I will try with gcc 9 as well. 

Out of curiosity how many threads do you use exactly ?

alazzaro:
the CI test is complaining that there is a memory alignment problem... This is SDBG, so a single thread I assume...

mkrack:
The test case H2O-1_non-ortho_grid_dgemm.inp even aborts with my gfortran regtest when I use 1 MPI task with 1 OpenMP thread.
2 MPI tasks and 1 OpenMP thread run, but might give wrong numbers, same for 1 MPI task and 2 OpenMP threads. All other combinations of MPI tasks and OpenMP threads which I tried abort.

mtaillefumier:
> the CI test is complaining that there is a memory alignment problem... This is SDBG, so a single thread I assume...

I saw it. the sanitizer complains that the size of the allocated memory is not a multiple of the alignment. I will submit a patch to fix this.

mtaillefumier:
I fixed the issue with this PR https://github.com/cp2k/cp2k/pull/1415.

hfp:
> Hmm. The barrier is correct here. it is a reduction over all threads but instead of having a critical section I split the table in pieces such that each thread can add data at different locations in parallel. The omp barrier call just say wait for all threads to finish their work before starting the next block.

Sorry, I was misreading the condition `if (num_threads > 1)` like `if (thread_id > 1)`.

mtaillefumier:
> > Hmm. The barrier is correct here. it is a reduction over all threads but instead of having a critical section I split the table in pieces such that each thread can add data at different locations in parallel. The omp barrier call just say wait for all threads to finish their work before starting the next block.
> 
> Sorry, I was misreading the condition `if (num_threads > 1)` like `if (thread_id > 1)`.

@hfp  It happens. Still I can not figure out why you have a crash while it is working fine on my side. I ran different benchmarks with different values for OMP and MPI and it always works. All tests have the same iterations, etc...

mtaillefumier:
@mkrack Sorry, i missed them. Memory leaks are fixed with PR https://github.com/cp2k/cp2k/pull/1416

mkrack:
@mtaillefumier Thanks for fixing the memory leaks. The results for the H2O-1_*_grid_dgemm.inp tests depend on the number of threads.

mtaillefumier:
> @mtaillefumier Thanks for fixing the memory leaks. The results for the H2O-1_*_grid_dgemm.inp tests depend on the number of threads.

Unless blas and lapack are not thread safe they should not depend on the number of threads. the only explanation for this error is the cblas functions I use since I only work with mkl and/or openblas. Using fortran functions in c code is not portable and prone to errors. lapack has the cblas implementation since 2013 thanks Intel. 

I run the tests locally and I got this for the total energy. (OMP=4)

H2O-1_non-ortho_grid_dgemm.inp : -17.17857472591210
H2O-1_non-ortho_grid_auto.inp : -17.17857472391006
H2O-1_ortho_grid_auto.inp : -17.17857476450136
H2O-1_ortho_grid_dgemm.inp : -17.17857476455123

mkrack:
@mtaillefumier Could you also try the same dgemm test with a different number of OMP threads?

mtaillefumier:
> @mtaillefumier Could you also try the same dgemm test with a different number of OMP threads?

I will do it right away

mtaillefumier:
with OMP = 2

H2O-1_non-ortho_grid_dgemm.inp : -17.17857472591210
H2O-1_non-ortho_grid_auto.inp : -17.17857472391005
H2O-1_ortho_grid_auto.inp : -17.17857476450136
H2O-1_ortho_grid_dgemm.inp : -17.17857476455124

with OMP = 1

H2O-1_non-ortho_grid_dgemm.inp : -17.17856354040013
H2O-1_non-ortho_grid_auto.inp : -17.17857472391006
H2O-1_ortho_grid_auto.inp : -17.17857476450136
H2O-1_ortho_grid_dgemm.inp : -17.17856369567907

there is a small difference between OMP = 1 and OMP = 2. I need to investigate this

mkrack:
I will take your OMP=1 energies as references, since my runs do not provide a stable reference energy. I also still get the abort with `KS energy is an abnormal value (NaN/Inf)`.

mtaillefumier:
> I will take your OMP=1 energies as references, since my runs do not provide a stable reference energy. I also still get the abort with `KS energy is an abnormal value (NaN/Inf)`.

Ok. That not a problem. I will relink against lapack/blas (just need to add liblapacke.a to the libs).

mtaillefumier:
@mkrack thanks for these updates. I would take the values for OMP>1 because they match the values obtained with the reference backend. I also tried with the cut-off on and I get this difference between OMP=1 and OMP > 1. I find the error (most probably accumulation error).

hfp:
If MKL is used, the environment variable `MKL_CBWR=AUTO` or even more STRICT ([reference](https://software.intel.com/content/www/us/en/develop/documentation/onemkl-linux-developer-guide/top/obtaining-numerically-reproducible-results/setting-the-environment-variable-for-conditional-numerical-reproducibility.html)) can be helpful for our regression tests. The given reference is part of the general topic about [numerical reproducible results](https://software.intel.com/content/www/us/en/develop/documentation/onemkl-linux-developer-guide/top/obtaining-numerically-reproducible-results.html) and the necessary (pre-)conditions. The term "CBWR" is MKL's legacy term for **C**onditional **B**it-**W**ise **R**eproducibility, also/now known as/called CNR (**C**onditional **N**umerical **R**eproducibility).

mtaillefumier:
I found the issue and this PR https://github.com/cp2k/cp2k/pull/1417 should fix it.

I also run the tests with OMP_NUM_THREADS=1 and OMP_NUM_THREADS=2

results are 

OMP=2
H2O-1_ortho_grid_auto.inp : -17.17857475634941
H2O-1_ortho_grid_dgemm.inp : -17.17857476455124

OMP=1
H2O-1_ortho_grid_dgemm.inp : -17.17857476455108
H2O-1_ortho_grid_auto.inp : -17.17857475634940

and for the non-ortho case

H2O-1_non-ortho_grid_auto.inp : -17.17857472462476
H2O-1_non-ortho_grid_dgemm.inp : -17.17857472591210 (OMP = 1)
H2O-1_non-ortho_grid_dgemm.inp : -17.17857472591210 (OMP = 2)

and with @hfp information I get this (MKL_CBWR="AVX")
H2O-1_non-ortho_grid_dgemm.inp : -17.17857472591211

mkrack:
@mtaillefumier Thanks for the new fixes. Unfortunately, the two H2O-1_*_dgemm.inp tests are still failing:
```
/data/user/krack/github/mkrack/TEST-Linux-x86-64-gfortran-regtest-psmp-2021-03-13_01-29-26/QS/regtest-grid/H2O-1_non-ortho_grid_dgemm.inp.out
                        eps_eigval:                                     1.00E-05
                        --------------------------------------------------------
                        level_shift [a.u.]:                                 0.00
                        --------------------------------------------------------
                        Mixing method:                           DIRECT_P_MIXING
                        --------------------------------------------------------
                        No outer SCF

 MD_PAR| Molecular dynamics protocol (MD input parameters)
 MD_PAR| Ensemble type                                                     NPT_I
 MD_PAR| Number of time steps                                                  3
 MD_PAR| Time step [fs]                                                 0.100000
 MD_PAR| Temperature [K]                                              300.000000
 MD_PAR| Temperature tolerance [K]                                      0.000000
 MD_PAR| Pressure [bar]                                                 0.000000
 MD_PAR| Barostat time constant [fs]                                 1000.000000
 MD_PAR| Print MD information every                                    1 step(s)
 MD_PAR| File type   Print frequency [steps]                          File names
 MD_PAR| Coordinates          1             H2O-1_non-ortho_grid_dgemm-pos-1.xyz
 MD_PAR| Cell                 1                H2O-1_non-ortho_grid_dgemm-1.cell
 MD_PAR| Velocities           1             H2O-1_non-ortho_grid_dgemm-vel-1.xyz
 MD_PAR| Energies             1                H2O-1_non-ortho_grid_dgemm-1.ener
 MD_PAR| Dump                20             H2O-1_non-ortho_grid_dgemm-1.restart

 ROT| Rotational analysis information
 ROT| Principal axes and moments of inertia [a.u.]
 ROT|                           1                   2                   3
 ROT| Eigenvalues      4.00383397110E+03   7.52264050485E+03   1.15264744759E+04
 ROT|      x             -0.000000000000      0.000000000000      1.000000000000
 ROT|      y              1.000000000000      0.000000000000      0.000000000000
 ROT|      z              0.000000000000      1.000000000000      0.000000000000
 ROT| Number of rotovibrational vectors                                        6

 DOF| Calculation of degrees of freedom
 DOF| Number of atoms                                                          3
 DOF| Number of intramolecular constraints                                     0
 DOF| Number of intermolecular constraints                                     0
 DOF| Invariants (translations + rotations)                                    0
 DOF| Degrees of freedom                                                       9

 DOF| Restraints information
 DOF| Number of intramolecular restraints                                      0
 DOF| Number of intermolecular restraints                                      0

 THERMOSTAT| Thermostat information for PARTICLES
 THERMOSTAT| Type of thermostat            Canonical Sampling/Velocity Rescaling
 THERMOSTAT| CSVR time constant [fs]                                   10.000000
 THERMOSTAT| Initial kinetic energy                           0.000000000000E+00
 THERMOSTAT| End of thermostat information for PARTICLES

 THERMOSTAT| Thermostat information for BAROSTAT
 THERMOSTAT| Type of thermostat            Canonical Sampling/Velocity Rescaling
 THERMOSTAT| CSVR time constant [fs]                                   10.000000
 THERMOSTAT| Initial kinetic energy                           0.000000000000E+00
 THERMOSTAT| End of thermostat information for BAROSTAT

 MD_VEL| Velocities initialization
 MD_VEL| Initial temperature [K]                                      300.000000
 MD_VEL| COM velocity             0.0000000000     0.0000000000    -0.0000000000

 Number of electrons:                                                          8
 Number of occupied orbitals:                                                  4
 Number of molecular orbitals:                                                 4

 Number of orbital functions:                                                 40
 Number of independent orbital functions:                                     40

 Extrapolation method: initial_guess


 SCF WAVEFUNCTION OPTIMIZATION

  Step     Update method      Time    Convergence         Total energy    Change
     1 P_Mix/Diag. 0.40E+00    0.1    96.67291391       -16.9780522186 -1.70E+01

 *  \___/                KS energy is an abnormal value (NaN/Inf).             *
 * / \                                                     qs_ks_methods.F:890 *


            9 qs_ks_build_kohn_sham_matrix
            8 rebuild_ks_matrix
            7 qs_ks_update_qs_env
            6 scf_env_do_scf_inner_loop
            5 scf_env_do_scf
            4 qs_energies
            3 qs_forces
            2 qs_mol_dyn_low
            1 CP2K
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
EXIT CODE:  1  MEANING:  RUNTIME FAIL
/data/user/krack/github/mkrack/TEST-Linux-x86-64-gfortran-regtest-psmp-2021-03-13_01-29-26/QS/regtest-grid/H2O-1_ortho_grid_dgemm.inp.out : 
 Total energy: : ref = -17.178574725912100 new = -17.22722889546401  
 relative error :   2.82425977e-03 >  numerical tolerance = 1.0E-12  
```

PR #1419, opened 2021-03-13
grid: Move tests from LIBTEST into QS/regtest-grid

oschuett:


mkrack:
Why do the test file and project names start with `Ar_` after renaming though the system is graphite?

oschuett:
Oh, my bad. While the ortho files are indeed derived from [Ar-2.inp](https://github.com/cp2k/cp2k/blob/master/tests/QS/regtest-gpw-1/Ar-2.inp), the non-ortho tests are based on [graphite3.inp](https://github.com/cp2k/cp2k/blob/master/tests/QS/regtest-nonortho/graphite3.inp).
I'll rename them.

PR #1421, opened 2021-03-15
Cleanup and a small improvements for the gpu backend

mtaillefumier:
- remove the hybrid cpu/gpu backend
- fix the warnings when -pedantic is on (there are still some warmings occuring when -DBDEBUG is given to the compiler)
- add output information in the global section that indicates which grid backend is used for the calculations
- small performance improvements for the integrate kernel. Reduction to shared memory for the coefficients is removed from the zyx loop for  l = 0 and l = 1.
- shift some atomic operations outside loops when possible.

PR #1422, opened 2021-03-16
the number of backends is hardcoded everywhere

mtaillefumier:
Signed-off-by: Dr. Mathieu Taillefumier <mathieu.taillefumier@free.fr>

The number of backends is directly hard coded instead of being defined as a preprocessor constant. Surprisingly, I can not reproduce the issue at all (PR https://github.com/cp2k/cp2k/pull/1421).

### DIFF
--- a/src/grid/common/grid_library.c
+++ b/src/grid/common/grid_library.c
@@ -105,7 +105,8 @@ void grid_library_counter_add(const int lp, const enum grid_backend backend,
                               const enum grid_library_kernel kernel,
                               const int increment) {
   const int back = backend - GRID_BACKEND_REF;
-  const int idx = back * 4 * 20 + kernel * 20 + imin(lp, 19);
+  const int idx =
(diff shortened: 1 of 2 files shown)


mtaillefumier:
I am waiting for the regtests to finish before fixing indentation

hfp:
For our curiosity, shall I launch the P100 test on this PR?

mtaillefumier:
That would be very helpful. I can not reproduce this issue at all.

oschuett:
Maybe wait until the tests on #1423 are done. I think this problem is on the host side.

hfp:
> Maybe wait until the tests on #1423 are done. I think this problem is on the host side.

ACK

oschuett:
I [realized](https://github.com/cp2k/cp2k/commit/6d312f7784da8491784779286455b654a9254ba4) that I could actually move the `struct grid_library_globals` into `grid_library.c` and thereby keep the macros out of the public header file. Thanks for bringing this issue up!

PR #1423, opened 2021-03-17
grid: Assert that max number of threads

oschuett:


hfp:
Such code like `max_nthreads * sizeof(something)` is sort of custom memory management, specifically thread-local storage. It depends on the actual case whether such code is more desirable/flexible than `omp threadprivate`. This is just a reminder that `omp threadprivate` exists and can be useful. An example can be found [here](https://github.com/cp2k/dbcsr/blob/develop/src/acc/opencl/acc_opencl.h#L197) (after declaration); please note the location (header) and the corresponding [definition](https://github.com/cp2k/dbcsr/blob/develop/src/acc/opencl/acc_opencl.c#L43) of this variable. This was for global variables, but is considerably easier for variables local to a function, etc.

One last note, manually managing TLS sometimes comes with a wrong assumption such as storage slots being repeatedly assigned to a specific thread, which is not the case for non-static schedules and perhaps different OpenMP regions.

oschuett:
Ok, I think I figured it out. ~~I forgot to add `imax` [here](https://github.com/cp2k/cp2k/blob/master/src/grid/gpu/grid_gpu_task_list.cu#L322) and [here](https://github.com/cp2k/cp2k/blob/master/src/grid/gpu/grid_gpu_task_list.cu#L325).~~ A variable was sometimes uninitialized.

PR #1427, opened 2021-03-18
Add three Intel regression tests

mkrack:


oschuett:
Wohoo, this improves our test coverage for intel compiler versions significantly!

mkrack:
I hope so, but it creates also new issues like the test case `QS/regtest-hfx-wfn-fitting/CH4-rsLDA.inp`. This test just runs for a fixed number of OT (10) steps and the final energy fluctuates depending on the compiler version. I am not sure if this is a compiler issue or just a numerical issue, since that test does not seem to converge even if the number of OT steps is increased.

alazzaro:
Thanks @mkrack !

The Intel 19.1.3 is reporting 3 failed tests in the grid part. It seems part of the OMP region... Would it be possible to run with a single thread?

I don't have any clue about the `QS/regtest-hfx-wfn-fitting/CH4-rsLDA.inp`. It is interesting that the only the 3 new Intel compilers are reporting this problem. Would it be possible also here to run with a single thread and some lower optimization (-01)?

alazzaro:
One more thing: I found that the flag `-fp-model precise` is not needed anymore since v18...

mkrack:
@alazzaro There is no impact from the number of threads, but the error disappears when the optimization level is reduced from `-O2` to `-O1`. The problem persists however with `-O2 -no-vec`. I tried to compile selected source files with `-O1`, but without success so far.

> One more thing: I found that the flag -fp-model precise is not needed anymore since v18...

Where did you find that information? The [latest Intel compiler documentation](https://software.intel.com/content/www/us/en/develop/documentation/fortran-compiler-oneapi-dev-guide-and-reference/top/compiler-reference/compiler-options/compiler-option-details/floating-point-options/fp-model-fp.html) still lists `-fp-model fast=1` as default.

alazzaro:
> @alazzaro There is no impact from the number of threads, but the error disappears when the optimization level is reduced from `-O2` to `-O1`. The problem persists however with `-O2 -no-vec`. I tried to compile selected source files with `-O1`, but without success so far.
> 
> > One more thing: I found that the flag -fp-model precise is not needed anymore since v18...
> 
> Where did you find that information? The [latest Intel compiler documentation](https://software.intel.com/content/www/us/en/develop/documentation/fortran-compiler-oneapi-dev-guide-and-reference/top/compiler-reference/compiler-options/compiler-option-details/floating-point-options/fp-model-fp.html) still lists `-fp-model fast=1` as default.

Sorry, I was not clear: what I mean is that I can run the entire regtests without specifying the `-fp-model`. Of course, I agree with you, it is safe to keep it.

PR #1428, opened 2021-03-18
grid: Add optimized branches for lp <= 2 to GPU kernels

oschuett:


oschuett:
The GPU integrate kernel should now be 4x faster.
@alazzaro, could you please re-enable `__GRID_CUDA` on the Daint performance tests?

alazzaro:
Done. The H2O-32 RI-MP2/RI-RPA goes up by a lot. Could you please check?

For the other tests the difference is smaller (some go up, some go down), need more tests to understand.

oschuett:
Thanks @alazzaro! At least for the H2O benchmarks the GPU performance now seems to be comparable to the CPU.
It would be good to understand why the RI tests are not yet on par. Could you please post their grid statistics?

I'm currently working on reducing the register usage of the Cuda kernels to improve their occupancy.

alazzaro:
So, this is the comparison of the timers:

```
mp_min_d                           4.370    11.346  160
mp_sum_d                           4.803     4.784   -0
mp_sync                            4.809     6.338   32
pw_cuda_c1dr3d_3d                  4.888    63.560 1200
ao_to_mo_and_store_B_mult_alph     4.899    28.234  476
pw_cuda_r3dc1d_3d                  6.497    76.603 1079
mp2_ri_gpw_compute_en_comm         9.554     9.526   -0
mp_sendrecv_dm3                   10.382    11.002    6
grid_integrate_task_list          49.136    31.740  -35
integrate_four_center_bin         55.722    57.157    3
mp2_ri_gpw_compute_en_expansio    66.697    66.551   -0
CP2K_Total                       227.121   327.243   44
```

What explodes is the PW_CUDA part. My understanding is that PW_CUDA is running in parallel with the GRID part, so the two kernel conflicts. I can try to run with PW_CUDA disabled...

These are the grid statistics:

```
 LP    KERNEL             BACKEND                              COUNT     PERCENT
 1     integrate ortho    GPU                             1370627904      34.79%
 2     integrate ortho    GPU                             1059063772      26.88%
 0     integrate ortho    GPU                              778592040      19.76%
 3     integrate ortho    GPU                              492370447      12.50%
 4     integrate ortho    GPU                              146980134       3.73%
 5     integrate ortho    GPU                               25820978       0.66%
 0     collocate ortho    REF                               11698176       0.30%
 1     collocate ortho    REF                                9191424       0.23%
 2     collocate ortho    REF                                7520256       0.19%
 0     integrate ortho    REF                                5849088       0.15%
 1     integrate ortho    REF                                4595712       0.12%
 1     collocate ortho    GPU                                4489536       0.11%
 3     collocate ortho    REF                                4177920       0.11%
 2     integrate ortho    REF                                3760128       0.10%
 2     collocate ortho    GPU                                3560380       0.09%
 6     integrate ortho    GPU                                2892471       0.07%
 0     collocate ortho    GPU                                2477928       0.06%
 3     integrate ortho    REF                                2088960       0.05%
 3     collocate ortho    GPU                                1692487       0.04%
 4     collocate ortho    REF                                 835584       0.02%
 4     collocate ortho    GPU                                 522198       0.01%
 4     integrate ortho    REF                                 417792       0.01%
 5     collocate ortho    GPU                                  95810       0.00%
 0     collocate general  REF                                  92160       0.00%
 0     integrate general  REF                                  46080       0.00%
 6     collocate ortho    GPU                                  11055       0.00%
```

BTW, I get a lot of (Warnings):

```
 Cannot destroy the ELPA object!
 This is a critical error!
 This might lead to a memory leak in your application!
 But you do not check the error codes!
 Cannot destroy the ELPA object!
 This is a critical error!
 This might lead to a memory leak in your application!
 But you do not check the error codes!
```

Is this a known problem?

dev-zero:
> ```
>  Cannot destroy the ELPA object!
>  This is a critical error!
>  This might lead to a memory leak in your application!
>  But you do not check the error codes!
>  Cannot destroy the ELPA object!
>  This is a critical error!
>  This might lead to a memory leak in your application!
>  But you do not check the error codes!
> ```
> 
> Is this a known problem?

Yes. We currently see the following issues on our machines with ELPA:

* memory leaks (4TB RES+)
* hangs in MPI calls issued by ELPA (discovered by attaching gdb to running process and triggering a backtrace)
* sudden crashes of CP2K without any message (my guess is that MPI crashes, hence no processing of any I/O)
* wrong results in diagonalization

It is not related to CUDA or OpenMP (`OMP_NUM_THREADS=1` triggers it nonetheless), but most likely related to small matrices since it can be more easily triggered by small systems and when I looked into some of the hanging processes by attaching gdb and going up the stack matrix dim was usually < 20.

Also, while it got more problematic with the latest ELPA and CP2K releases, I observed similar problems with small matrix sizes with CP2K 7.1 (and the ELPA we used back then) but didn't have enough time to come up with a minimal reproducer to report to upstream.

So, for now disable ELPA (either at compile or at runtime).

mkrack:
I also observed that PW_CUDA causes a significant performance penalty and that is why `-D__PW_CUDA` is currently disabled in the [CRAY_XC50_gnu.psmp](https://github.com/cp2k/cp2k/blob/fd98e12cbc78dc2153d2b88ab1bcd231031b7f2e/arch/CRAY-XC50_gnu.psmp#L64) arch file.

I also get the ELPA warnings. The results are fine, but I do not check for potential memory leaks on CRAY. I tried to compile ELPA with and without OpenMP, but this has no impact on the appearance of that warning message.

Furthermore concerning ELPA: I get crashes with ELPA on CRAY especially when `ELPA_QR on` is selected. That is why I do not use ELPA currently on CRAY for production, since the performance gained with ELPA compared to ScaLAPACK is minor (at least for medium-sized systems)

oschuett:
> These are the grid statistics:

They look pretty similar to the H2O benchmark. It might be worth to add a specialized branch for lp=3 too. For lp>4 the Cuda kernels are currently falling off a cliff because we can not store [more coefficients](https://github.com/cp2k/cp2k/blob/master/src/grid/gpu/grid_gpu_collint.h#L38) in the L1 cache.

In any case, in order to optimize this I'll need a smaller benchmark that I can run at home on a single node.

> I can try to run with PW_CUDA disabled...

Yes, please give that a try. AFAIK, PW_CUDA provides very little speedup even under ideal conditions. Furthermore, it does not play well with multi-GPU and will probably get replaced by SpFFT soon. So, I'd be ok with disabling it for the time being.

> `Cannot destroy the ELPA object!`

This is news to me. I guess, we should ask for the error code when calling [elpa_deallocate](https://github.com/cp2k/cp2k/blob/master/src/fm/cp_fm_elpa.F#L506).

mkrack:
@oschuett I could run the 256 H2O benchmark case with and without `-D__GRID_CUDA` (and `-D__PW_CUDA` disabled) on daint-gpu. Would that help you?

alazzaro:
OK, I have the results without PW_CUDA:

(CPU, w/ PW_CUDA, w/o PW_CUDA)

```
mp_min_d                           4.370    11.346  160    10.305  136
mp_sum_d                           4.803     4.784   -0     3.847  -20
mp_sync                            4.809     6.338   32     5.639   17
pw_cuda_c1dr3d_3d                  4.888    63.560 1200     ----------
ao_to_mo_and_store_B_mult_alph     4.899    28.234  476   129.807 2550
pw_cuda_r3dc1d_3d                  6.497    76.603 1079     ----------
mp2_ri_gpw_compute_en_comm         9.554     9.526   -0     9.645    1
mp_sendrecv_dm3                   10.382    11.002    6    10.372   -0
grid_integrate_task_list          49.136    31.740  -35   138.760  182
integrate_four_center_bin         55.722    57.157    3    56.335    1
mp2_ri_gpw_compute_en_expansio    66.697    66.551   -0    66.157   -1
CP2K_Total                       227.121   327.243   44   334.258   47
```

So, it is even slower... We have to look at the GPU profiler data to see what's going on...

oschuett:
Thanks @mkrack, but I can run the 64 H2O locally, which I believe is a good enough proxy for the 256 H2O.

I now realized that after my recent optimizations I can also run `H2O-32-RI-MP2-TZ.inp` locally. So, I'll have a look.

PR #1435, opened 2021-03-23
add missing -fopenmp-simd flag

mtaillefumier:


mkrack:
Just adding the flag `-fopenmp-simd` does not resolve the compilation error caused by the new pragmas. 

```
/data/user/krack/rt/gnu-psmp/cp2k/src/grid/cpu/utils.c: In function 'extract_sub_grid':
/data/user/krack/rt/gnu-psmp/cp2k/src/grid/cpu/utils.c:206:46: error: expected '#pragma omp' clause before 'nontemporal'
#pragma omp simd linear(dst, src) simdlen(8) nontemporal(src)
                                              ^~~~~~~~~~~
```

mtaillefumier:
I will suppress the statement. It is easier that way.

PR #1436, opened 2021-03-23
Enable check_diag for all eigensolver routines

mkrack:


src/grid/common/grid_common.h
```diff
@@ -10,7 +10,9 @@
 #define GRID_STRINGIFY(SYMBOL) #SYMBOL
 
 // GCC introduced the unroll pragma with version 8 using a custom syntax.
-#if defined(__GNUC__) && __GNUC__ >= 8
+#if defined(__INTEL_COMPILER)
+#define GRID_PRAGMA_UNROLL(N) _Pragma(GRID_STRINGIFY(unroll(N)))
```
oschuett, line 14: Isn't this already covered by the `#else` branch? I believe all compilers except for gcc use the same syntax.

mkrack: The Intel compilers have `-gcc` defined as the default and thus the GNU macros like `__GNUC__` are also defined with Intel.

PR #1445, opened 2021-03-30
Enable toolchain built with arm64 (aarch64)

mkrack:


dev-zero:
What was the problem with the `RelWithDebug` for libvori?

dev-zero:
> libvori gives wrong results with `-O2`. Switching to `-O1` resolves the issue. It is certainly possible to adapt the flags for `RelWithDebug`, but for this the `CMakeLists.txt` file in libvori.tar.gz has to be updated.

Well, if `-O2` gives wrong results with a somewhat recent gcc there is most likely a problem in the code itself, especially with just AArch64 (without SVE) which should be stable enough.

tools/toolchain/scripts/stage6/install_quip.sh
```diff
@@ -103,6 +103,8 @@ case "$with_quip" in
       export QUIP_ARCH=linux_${quip_arch}_gfortran
       # hit enter a few times to accept defaults
       echo -e "${MATH_LDFLAGS} $(resolve_string "${MATH_LIBS}") \n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n" | make config > configure.log
+      # Download up-to-date config.guess
+      wget http://savannah.gnu.org/cgi-bin/viewcvs/*checkout*/config/config/config.guess -O src/fox/config/config.guess -o wget.log
```
oschuett: Maybe this should better be a patch? Otherwise, please add a checksum.

mkrack: I think it would be the best to replace the `config.guess` file from 2006 in the current QUIP archive file with the most recent version from 2021.

oschuett: I would prefer to keep the upstream archives unmodified. Fortunately, they just upgraded their fox submodule: https://github.com/libAtoms/QUIP/commit/b4336484fb65b0e73211a8f920ae4361c7c353fd. So, we can just create a new archive, ie. upgrade to the latest QUIP version.

PR #1457, opened 2021-04-09
Dashboard: Add Python type annotations

oschuett:


tools/dashboard/generate_dashboard.py
```diff
@@ -19,71 +21,157 @@
...
 mpl.use("Agg")  # change backend, to run without X11
-import matplotlib.pyplot as plt
-from matplotlib.ticker import AutoMinorLocator
+import matplotlib.pyplot as plt  # type: ignore
+from matplotlib.ticker import AutoMinorLocator  # type: ignore
 
 # ======================================================================================
-class GitLog(list):
-    def __init__(self):
+class GitLog:
```
dev-zero, line 42: Most of those classes look like they could profit from becoming `dataclasses`: https://docs.python.org/3/library/dataclasses.html, might need the backport from here https://pypi.org/project/dataclasses/ if only Python 3.6 is available

tools/dashboard/generate_dashboard.py
```diff
@@ -19,71 +21,157 @@
...
+def gen_frontpage(
+    config: ConfigParser, log: GitLog, status_fn: str, outdir: str, send_emails: bool
+) -> None:
+
+    status: Dict[str, Status] = {}
     if path.exists(status_fn):
-        status = eval(open(status_fn, encoding="utf8").read())
-    else:
-        status = dict()
+        status = pickle.load(open(status_fn, "rb"))
```
dev-zero:
```python
with open(status_fn, "rb") as fhandle:
    status = pickle.load(fhandle)
```

tools/dashboard/generate_dashboard.py
```diff
@@ -109,46 +194,47 @@ def get_sortkey(s):
 
         # find latest commit that should have been tested by now
         threshold = now - timedelta(hours=timeout)
-        ages_beyond_threshold = [i for i, c in enumerate(log) if c["date"] < threshold]
-        threshold_age = ages_beyond_threshold[0]
+        beyond_threshold = [i for i, c in enumerate(log.commits) if c.date < threshold]
+        threshold_age = beyond_threshold[0]
```
dev-zero:
```python
threshold_age =  next(i for i, c in enumerate(log.commits) if c.date < threshold)
```
... unless I overlooked that the full `beyond_threshold` is also needed somewhere.

tools/dashboard/generate_dashboard.py
```diff
@@ -19,71 +21,157 @@
...
-def main():
+def main() -> None:
     if len(sys.argv) < 4:
         usage = "<config-file> <status-file> <output-dir> [--send-emails]"
         print(f"Usage: generate_dashboard.py {usage}")
         sys.exit(1)
 
     config_fn, status_fn, outdir = sys.argv[1:4]
-    outdir = outdir.strip("/")
+    outdir = outdir.rstrip("/")
```
dev-zero: using the builtin `argparse` instead of handrolling and then using `type=pathlib.Path` might avoid some headaches when it comes to raw path manipulation, see also https://docs.python.org/3/library/argparse.html#type

tools/dashboard/generate_dashboard.py
```diff
@@ -183,19 +269,21 @@ def gen_archive(config, log, outdir):
...
                 report = parse_report(report_txt, log)
-                report["url"] = path.basename(fn)[:-3]
+                report.archive_path = path.basename(fn)[:-3]
                 reports_cache[fn] = report
-            sha = report["git-sha"]
-            assert sha not in archive_reports
-            archive_reports[sha] = report
+            if not report.sha:
+                print(report)
+            assert report.sha and report.sha not in archive_reports
```
dev-zero, line 262: using `assert ..., "some message here"` helps to document the assertion and giving additional feedback in case it fails

tools/dashboard/generate_dashboard.py
```diff
@@ -204,21 +292,23 @@ def gen_archive(config, log, outdir):
         # loop over all relevant commits
         all_url_rows = []
         all_html_rows = []
-        max_age = 1 + max([-1] + [log.index[sha] for sha in archive_reports.keys()])
-        for commit in log[:max_age]:
-            sha = commit["git-sha"]
+        max_age = 1 + max([-1] + [log.index_by_sha[sha] for sha in archive_reports])
```
dev-zero, line 273: `max(-1, *(log.index_by_sha[sha] for sha in archive_reports))`

tools/dashboard/generate_dashboard.py
```diff
@@ -381,15 +466,23 @@ def make_plot_image(fn, title, ylabel, curves, max_age, full_archive):
...
+    idx_end = log.index_by_sha[report.sha] if report.sha else 0
     if not last_ok:
         return  # we don't know when this started
-    idx_last_ok = log.index[last_ok]
+    idx_last_ok = log.index_by_sha[last_ok]
     if idx_end == idx_last_ok:
         return  # probably a flapping tester
-    emails = set([log[i]["author-email"] for i in range(idx_end, idx_last_ok)])
-    emails = [e for e in emails if "noreply" not in e]
+    emails_raw = set([log.commits[i].author_email for i in range(idx_end, idx_last_ok)])
```
dev-zero: no need to create a list first and you can also drop the explicit index:
`set(c.author_email for c in log.commits[idx_end:idx_last_ok])`

tools/dashboard/generate_dashboard.py
```diff
@@ -543,35 +635,38 @@ def write_file(fn, content, gz=False):
         if old_content == content:
             print("File did not change: " + fn)
             return
-    f = gzip.open(fn, "wb") if (gz) else open(fn, "wb")
-    f.write(content.encode("utf-8"))
-    f.close()
+
+    encoded_content = content.encode("utf-8")
+    if gz:
+        gzip.open(fn, "wb").write(encoded_content)
```
dev-zero: use context managers

tools/dashboard/generate_dashboard.py
oschuett: Oh, I didn't know about `dataclasses`. Very nice.

tools/dashboard/generate_dashboard.py
oschuett: Neat :-)

tools/dashboard/generate_dashboard.py
oschuett: Unfortunately, this does not work when the list is empty: 

```
    max_age = 1 + max(-1, *(log.index_by_sha[sha] for sha in archive_reports))
TypeError: 'int' object is not iterable

```

tools/dashboard/generate_dashboard.py
oschuett: Good idea. Thanks for the thorough review! I learned a lot :-)

PR #1464, opened 2021-04-13
Update qs_scf_post_gpw.F

martinbrehm:
Added APPEND flag for &PRINT / &VORONOI.


dev-zero:
@Brehministrator you can leave the PRs open. A PR is always tied to the branch you used, hence when you push to the branch (doesn't matter whether additional commits or changed commits via a force-push) the PR gets updated automatically.
And when we see that additional commits are iterations on the first one (like a prettification commit) we usually squash all the commits into one when merging.

PR #1481, opened 2021-04-23
grid: Allocate rs_grid%r from pinned memory

oschuett:


src/offload/offload_api.F
```diff
@@ -0,0 +1,131 @@
...
+      END INTERFACE
+
+      CALL timeset(routineN, handle)
+
+      CALL offload_create_buffer_c(length=length, buffer=buffer%c_ptr)
+      CPASSERT(C_ASSOCIATED(buffer%c_ptr))
+
+      host_buffer_c = offload_get_buffer_host_pointer_c(buffer%c_ptr)
+      CPASSERT(C_ASSOCIATED(host_buffer_c) .OR. length == 0)
+      CALL C_F_POINTER(host_buffer_c, buffer%host_buffer, shape=(/length/))
```
dev-zero: as per the standard wrt to `host_buffer_c`:

shall be a scalar of type C_PTR. It is an INTENT (IN) argument. Its value shall be
• the C address of an interoperable data entity,
• the result of a reference to C_LOC with a noninteroperable argument, or
• the C address of a storage sequence that is not in use by any other Fortran entity.

hence I'd consider a call to `C_F_POINTER` with `host_buffer_c` being a `NULL`  pointer may result in undefined behaviour independently of `length`.

oschuett: I think, this is legit because the C standard explicitly allows for `malloc(0)` to return `NULL`.

dev-zero: But Fortran wants the first argument pointer in `C_F_POINTER` to be an address, hence not NULL, independently of the argument of `shape`.

dev-zero: Well, begs the question whether `NULL` is also in Fortran standard considered the C address of an interoperable data entity.

oschuett: Yes, my line of thinking was that `NULL` is indeed a valid C address for a zero-size array. Furthermore, the code has been working for [a while](https://github.com/cp2k/cp2k/blob/5b32b9ab39c2be5751b193892c89f6f335af6109/src/grid/grid_api.F#L590).  However, Steve Lionel [agrees](https://narkive.com/BbeqaaSL.13) with you.

The only alternate solution I can think of is the following. However, then I'd have to add special cases for deallocating the zero-sized Fortran array or rely on the implementation detail that it does not actually allocate anything.

```
IF (length == 0) THEN
   ALLOCATE(buffer%host_buffer(0))
ELSE
   CPASSERT(C_ASSOCIATED(host_buffer_c))
   CALL C_F_POINTER(host_buffer_c, buffer%host_buffer, shape=(/length/))
ENDIF
```

So, in the end I believe the current solution is still preferable.

dev-zero: Given that it is already difficult to get CP2K to build correctly on flang and fujitsu I'd prefer standards-compliant code.
Even if it requires an extra branch at deallocation.

oschuett: Fair point. I've changed it

PR #1486, opened 2021-04-24
printing of real space KS and S matrices

ducryf:
Allow the printing of the real space KS and overlap matrices.

Subroutine write_matrix_csr is growing rather large, should I split it up?

### DIFF
--- a/src/input_cp2k_dft.F
+++ b/src/input_cp2k_dft.F
@@ -2255,6 +2255,14 @@ CONTAINS
                           lone_keyword_l_val=.TRUE.)
       CALL section_add_keyword(print_key, keyword)
       CALL keyword_release(keyword)
+      CALL keyword_create(keyword, __LOCATION__, name="Real_space", &
+                          description="Print the KS matrix in real-space instead of k-space.. ", &
+                          usage="REAL_SPACE {logical}", &
+                          repeats=.FALSE., &
+                          default_l_val=.FALSE., &
+                          lone_keyword_l_val=.TRUE.)
+      CALL section_add_keyword(print_key, keyword)
+      CALL keyword_release(keyword)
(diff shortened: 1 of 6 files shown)


ducryf:
Printing of the real space representation of the KS and S matrix was brought up twice on the forum recently:
https://groups.google.com/g/cp2k/c/x186nkdw7nc/m/FTraZJroAQAJ
https://groups.google.com/g/cp2k/c/gpvLxmSz8N8

This pr will also print the index-to-cell mapping to the output file.

oschuett:
Thanks for your PR!

> Subroutine write_matrix_csr is growing rather large, should I split it up?

Yes, good idea! It seems the routine now handles four cases (symmetric/non-symmetric) x (realspace/k-space). So, breaking it up along these lines seems natural. Maybe you can even reduce the code size a little bit by re-using some parts.

ducryf:
I've separated the desymmetrization of the real-space matrices into a subroutine. The printing of the real-space and k-space is also split into two separate subroutines. I hope this makes the code a bit simpler.

The reg test simply prints the rs representation of the KS matrices. Is there a possibility of testing a number from a file, e.g. the first entry of the ks.csr file?

dev-zero:
@ducryf afaik testing of values in anything else than the CP2K output is not possible in the regtesting script at the moment.
To add a new "test type" you'd have to add a new entry to `tests/TEST_TYPES` (consisting of an expression and a column number) and the line number of that new entry will be the test ID to be used in the `tests/QS/regtest-kp-2/TEST_FILES` (second column).
Does that help?

ducryf:
@dev-zero Thanks for the explanation. I've added a regtest to check the number of images that are printed. At least this makes sure that the desymmetrization works correctly.

PR #1489, opened 2021-04-26
Fix and change directive error if OpenMP flag is missing

alazzaro:
CCE has `-homp` flag to enable OpenMP, need to be more general in the error message.

### DIFF
--- a/src/base/base_uses.f90
+++ b/src/base/base_uses.f90
@@ -14,7 +14,7 @@
 
 ! Check for OpenMP early on - ideally before the compiler fails with a cryptic message.
 #if !defined(_OPENMP)
-#error "OpenMP is required. Please add the corresponding flag (eg. -fopenmp for GFortran) to your Fortran compiler flags."
+   "OpenMP is required. Please add the corresponding flag (eg. -fopenmp for GFortran) to your Fortran compiler flags."
 #endif
 
 ! Dangerous: Full path can be arbitrarily long and might overflow Fortran line.
--- a/src/base/base_uses.f90
+++ b/src/base/base_uses.f90
@@ -14,7 +14,7 @@
 
 ! Check for OpenMP early on - ideally before the compiler fails with a cryptic message.
 #if !defined(_OPENMP)
-   "OpenMP is required. Please add -fopenmp to your Fortran compiler flags."
+#error "OpenMP is required. Please add the corresponding flag (eg. -fopenmp for GFortran) to your Fortran compiler flags."
 #endif
 
 ! Dangerous: Full path can be arbitrarily long and might overflow Fortran line.


src/base/base_uses.f90
```diff
@@ -14,7 +14,7 @@
 
 ! Check for OpenMP early on - ideally before the compiler fails with a cryptic message.
 #if !defined(_OPENMP)
-   "OpenMP is required. Please add -fopenmp to your Fortran compiler flags."
+#error "OpenMP is required. Please add the corresponding flag (eg. -fopenmp for GFortran) to your Fortran compiler flags."
```
oschuett: Unfortunately, `#error` is not support by the traditional preprocessor and therefore not available in Fortran (according to the standard).

alazzaro: We use it already [here](https://github.com/cp2k/cp2k/blob/ec1cf9238dadae8274d74a47306a84345b4bbf05/src/environment.F#L231) (and in pw_cuda). Concerning the standard, CCE wants it. Most of the FPP in compilers support it (expect [NAG](https://stackoverflow.com/questions/29609305/user-defined-errors-with-nagfor-preprocessor)?)

dev-zero: We need something which breaks compilation, but there's no standard way of doing it. Usually this would be the job of the build system (once switched to CMake this will be solved in a portable way).
The only case we could hit are broken preprocessors, namely those which first parse for invalid preprocessor directives before checking whether the specific line is enabled or not, right?

alazzaro: not the vanilla FPP for sure... CCE, GNU and Intel recognize `#error`. I just want to have a meaningful message, I can even avoid it...

oschuett: Maybe we could move the error to a C file? For example there was a plan to move some of the code in [machine.F](https://github.com/cp2k/cp2k/blob/master/src/base/machine.F) to C.

alazzaro: A C file? I think here we want to check that the Fortran compiler has the OpenMP part.
Currently, CCE gives the following error:
```
ftn-855 crayftn: ERROR MEMORY_UTILITIES, File = ../../../../../home/users/alazzaro/cp2k/CP2K_AUTO_REGTEST/CRAY-XC40-cce/popt/cp2k/src/common/memory_utilities.F, Line = 14, Column = 8 
  The compiler has detected errors in module "MEMORY_UTILITIES".  No module information file will be created for this module.


ftn-100 crayftn: ERROR MEMORY_UTILITIES, File = ../base/base_uses.f90, Line = 17, Column = 4 
  This statement must begin with a label, a keyword, or an identifier.

```

Then, I have to look at the specific line of the code to check the error...
Let's leave it without the directive, I will just change the message...

oschuett: You're right. I didn't think this through. I actually already added a check on the C [side](https://github.com/cp2k/cp2k/blob/master/src/grid/common/grid_library.c#L36).

PR #1491, opened 2021-04-26
Misc fixes

alazzaro:


src/mpiwrap/message_passing.F
```diff
@@ -2108,6 +2109,44 @@ SUBROUTINE mp_comm_split(comm, sub_comm, ngroups, group_distribution, &
 
    END SUBROUTINE mp_comm_split
 
+! **************************************************************************************************
+!> \brief Get the local rank on the node according to the global communicator
+!> \return Node Rank id
+!> \author Alfio Lazzaro
+! **************************************************************************************************
+   FUNCTION mp_get_node_global_rank() &
```
oschuett, line 2117: Out of curiosity, what are planing to use this for and how is it different from `omp_get_thread_num()`?

oschuett: Got it, thanks for the explanation!

> It turns out that the best performance comes in DBCSR with a compact mapping

Oh, that's surprising. AFAIK we are not yet doing any GPU-to-GPU communication, are we?
I can definitely see how it would be beneficial once we start using CUDA-aware MPI.

alazzaro: Well, I should give a bit more context, you are right... 
The problem of the rank assignment is because GPUs in multi-socket systems are organized in a compact way.
So, again the 4 GPUs and 8 ranks example: GPU 0 and 1 are on socket 0, GPU 2 and 3 are on socket 1; ranks 0-3 are on socket 0, ranks 4-7 are on socket 1.
If we do the round-robin assignment as it is now, then (for example) rank 4 sits on socket 1, but it has to connect to GPU 0 on socket 0... 
G2G is on the plan for DBCSR, we need it for sure....

oschuett: Ok, that makes sense. 

> GPU 0 and 1 are on socket 0, GPU 2 and 3 are on socket 1; ranks 0-3 are on socket 0, ranks 4-7 are on socket 1.

I'm wondering if this is really always the case or if we should consider using [nvmlDeviceGetCpuAffinity](https://docs.nvidia.com/deploy/nvml-api/group__nvmlAffinity.html#group__nvmlAffinity). Still, as a heuristic it's most certainly better than our current round-robin.

alazzaro: Well, the round-robin is not good for sure and I will replace it with the compact as part of this PR (I'm testing it, takes a while).
Then, of course you get a system with funny GPU placement over sockets/numa domains, but this is something a rank remapping can fix. As far as I know, the multi-gpu systems I'm aware of are using a compact assignment. For instance, this is the [Summit design](https://www.olcf.ornl.gov/wp-content/uploads/2018/10/Summit_Node.jpg).

oschuett: I agree. Thanks for the link. It's great that we're is now getting serious multi-GPU support!

oschuett:
Should this be merged before the 8.2 release?

mkrack:
In my opinion, yes, if the dashboard stays green, of course.

Issue #1490, opened 2021-04-26
Warning and issues reported by CCE


alazzaro:
CCE compiler reports some warnings, potentially errors:

```
ftn-7276 crayftn: WARNING MATRIX_SIGN_SUBMATRIX_MU_ADJUST, File = ../../../../../home/users/alazzaro/cp2k/CP2K_AUTO_REGTEST/CRAY-XC40-cce/popt/cp2k/src/iterate_matrix.F, Line = 1602 
  OpenMP reduction result, trace, does not have a subsequent use.

ftn-7212 crayftn: WARNING NNP_GRADIENTS, File = ../../../../../home/users/alazzaro/cp2k/CP2K_AUTO_REGTEST/CRAY-XC40-cce/popt/cp2k/src/nnp_model.F, Line = 187 
  Variable "j" is used before it is defined.

ftn-7276 crayftn: WARNING TPSS_LDA_EVAL, File = ../../../../../home/users/alazzaro/cp2k/CP2K_AUTO_REGTEST/CRAY-XC40-cce/popt/cp2k/src/xc/xc_tpss.F, Line = 168 
  OpenMP reduction result, non_coer, does not have a subsequent use.

ftn-7254 crayftn: WARNING SAP_SORT, File = ../../../../../home/users/alazzaro/cp2k/CP2K_AUTO_REGTEST/CRAY-XC40-cce/popt/cp2k/src/sap_kind_types.F, Line = 243 
   A loop starting at line 243 which was marked for worksharing will not be workshared.

ftn-7208 crayftn: WARNING BUILD_CORE_PPNL, File = ../../../../../home/users/alazzaro/cp2k/CP2K_AUTO_REGTEST/CRAY-XC40-cce/popt/cp2k/src/core_ppnl.F, Line = 520 
  Privatized version of variable "natom" is used before it is defined.

ftn-7208 crayftn: WARNING BUILD_COM_MOM_NL, File = ../../../../../home/users/alazzaro/cp2k/CP2K_AUTO_REGTEST/CRAY-XC40-cce/popt/cp2k/src/commutator_rpnl.F, Line = 646 
  Privatized version of variable "natom" is used before it is defined.

[...]

PR #1492, opened 2021-04-27
Warning and issues reported by CCE #1490

juerghutter:
Correct for CCE Compiler Warnings:
iterate_matrix.F, Line = 1602 OpenMP reduction result, trace, does not have a subsequent use.
/xc/xc_tpss.F, Line = 168 OpenMP reduction result, non_coer, does not have a subsequent use.
core_ppnl.F, Line = 520 Privatized version of variable "natom" is used before it is defined.
commutator_rpnl.F, Line = 646 Privatized version of variable "natom" is used before it is defined.
xtb_matrices.F, Line = 1467 Variable "iw" is used before it is defined.

Still open:
nnp_model.F, Line = 187 Variable "j" is used before it is defined.
sap_kind_types.F, Line = 243 A loop starting at line 243 which was marked for worksharing will not be workshared.

### DIFF
--- a/src/commutator_rpnl.F
+++ b/src/commutator_rpnl.F
@@ -528,12 +528,12 @@ CONTAINS
 !$OMP PARALLEL &
 !$OMP DEFAULT (NONE) &
 !$OMP SHARED  (basis_set, matrix_rv, matrix_rxrv, matrix_rrv, &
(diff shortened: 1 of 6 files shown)


alazzaro:
@juerghutter Thanks for the fixes!
I have a fix for the `sap_kind_types.F` in my PR.

PR #1499, opened 2021-04-30
Parallelize the calculation of 2c integrals for RI with GPW using OpenMP

gsitaram:


src/mp2_eri_gpw.F
```diff
@@ -243,8 +245,35 @@ SUBROUTINE mp2_eri_2c_integrate_gpw(qs_env, para_env_sub, dimen_RI, mo_coeff, my
 
          CALL timestop(handle2)
 
-         ! integrate the little bastards
          offset = 0
+         ! Prepare offset ahead of OMP parallel loop
+         max_nseta = 0
+         DO iatom = 1, natom
```
oschuett: Instead of this loop you could call [get_qs_kind_set(maxnset=maxnset)](https://github.com/cp2k/cp2k/blob/master/src/qs_kind_types.F#L816).

src/mp2_eri_gpw.F
```diff
@@ -243,8 +245,35 @@ SUBROUTINE mp2_eri_2c_integrate_gpw(qs_env, para_env_sub, dimen_RI, mo_coeff, my
...
+            nseta = basis_set_a%nset
+            nsgfa => basis_set_a%nsgf_set
+            DO iset = 1, nseta
+               offset = offset + nsgfa(iset)
+               offset_2d(iset, iatom) = offset
+            END DO
+         END DO
+
+         ! integrate the little bastards
+         !$OMP PARALLEL DO DEFAULT(SHARED) &
```
oschuett: As a precaution we require that every OpenMP parallel region uses `DEFAULT(NONE)`.

src/mp2_eri_gpw.F
```diff
@@ -271,6 +300,8 @@ SUBROUTINE mp2_eri_2c_integrate_gpw(qs_env, para_env_sub, dimen_RI, mo_coeff, my
                ALLOCATE (I_ab(nsgfa(iset), 1))
                I_ab = 0.0_dp
 
+               offset = offset_2d(iset, iatom)
```
oschuett, line 300: This assignment is not needed because the `offset` variable is used only once. Simply plug it in directly:

`MODULO(offset_2d(iset, iatom), para_env_sub%num_pe)`

oschuett (changes requested):
Thanks for your PR. Looks good generally. Please run `make pretty` to fix the formatting issue.

PS: If you plan to add GPU support for this, please let me know.

gsitaram:
> Thanks for your PR. Looks good generally. Please run `make pretty` to fix the formatting issue.
> 
> PS: If you plan to add GPU support for this, please let me know.

Thanks for reviewing my changes. I have run `make pretty`, added `DEFAULT(NONE)` to the OpenMP pragma and used `qs_get_kind_set()` instead of the explicit loop. As for your comment about avoiding the use of the `offset` variable, I saw that it is used two more times in the following line of code:
`                  L_local_col(offset - nsgfa(iset) + 1:offset, i_counter) = I_ab(1:nsgfa(iset), 1)`
So, I left it as it is. I hope this PR may now be merged.

Issue #1444, opened 2021-03-29
ELPA: memory leaks, wrong results and crashes when using small matrices


dev-zero:
as mentioned in #1428 

We currently see the following issues on our machines with ELPA:

* memory leaks (4TB RES+)
* hangs in MPI calls issued by ELPA (discovered by attaching gdb to running process and triggering a backtrace, maybe waiting for memory allocation?)
* sudden crashes of CP2K without any message (my guess is that MPI crashes, hence no processing of any I/O)
* wrong results in diagonalization

It is not related to CUDA or OpenMP (`OMP_NUM_THREADS=1` triggers it nonetheless), but most likely related to small matrices since it can be more easily triggered by small systems and when I looked into some of the hanging processes by attaching gdb and going up the stack matrix dim was < 100.

Also, while it got more problematic with the latest ELPA and CP2K releases, I observed similar problems with small matrix sizes with CP2K 7.1 (and the ELPA we used back then) but didn't have enough time to come up with a minimal reproducer to report to upstream.

dev-zero:
```
#0  0x00007f486a761914 in uct_mm_iface_progress () from /usr/lib64/libuct.so.0
#1  0x00007f486add91f2 in ucp_worker_progress () from /usr/lib64/libucp.so.0
#2  0x00007f486b020527 in mca_pml_ucx_progress () from /usr/lib64/mpi/gcc/openmpi3/lib64/openmpi/mca_pml_ucx.so
#3  0x00007f487e73d67c in opal_progress () from /usr/lib64/mpi/gcc/openmpi3/lib64/libopen-pal.so.40
#4  0x00007f486b020725 in mca_pml_ucx_recv () from /usr/lib64/mpi/gcc/openmpi3/lib64/openmpi/mca_pml_ucx.so
#5  0x00007f4881b203a5 in PMPI_Recv () from /usr/lib64/mpi/gcc/openmpi3/lib64/libmpi.so.40
#6  0x00007f48802a1b70 in pmpi_recv__ () from /usr/lib64/mpi/gcc/openmpi3/lib64/libmpi_mpifh.so.40
[...]

PR #1501, opened 2021-05-02
ELPA: Re-enable by default, tighten CHECK_DIAG threshold, and check elpa_deallocate

oschuett:
See #1444 for discussion.


mkrack:
This PR has become obsolete because of #1505

PR #1515, opened 2021-05-10
Use ELPA as default only for large matrices

mkrack:
The eigensolver from the ELPA library will only be used for problems with more than 16 eigenvectors, even if ELPA is the default (PREFERRED_DIAG_LIBRARY) library. The ScaLAPACK library is used as fallback. This will avoid potential instabilities for small systems.

### DIFF
--- a/src/fm/cp_fm_diag.F
+++ b/src/fm/cp_fm_diag.F
@@ -63,8 +63,14 @@ MODULE cp_fm_diag
 
    REAL(KIND=dp), PARAMETER, PUBLIC :: eps_check_diag_default = 5.0E-14_dp
 
-   ! These saved variables are diagonalization global
+   ! The following saved variables are diagonalization global
+   ! Stores the default library for diagonalization
    INTEGER, SAVE       :: diag_type = 0
+   ! Minimum number of eigenvectors for the use of the ELPA eigensolver.
+   ! The ScaLAPACK eigensolver is used as fallback for all smaller cases.
+   INTEGER, SAVE       :: elpa_neigvec_min = 0
+   ! Threshold value for the orthonormality check of the eigenvectors obtained
+   ! after a diagonalization. A negative value disables the check.
    REAL(KIND=dp), SAVE :: eps_check_diag = -1.0_dp
 
    ! Constants for the diag_type above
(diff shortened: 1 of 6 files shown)


src/fm/cp_fm_diag.F
```diff
@@ -173,14 +175,14 @@ SUBROUTINE choose_eigv_solver(matrix, eigenvectors, eigenvalues, info)
       ! Sample peak memory
       CALL m_memory()
 
-      IF (diag_type == DIAG_TYPE_ELPA) THEN
-         CALL cp_fm_diag_elpa(matrix, eigenvectors, eigenvalues)
-      ELSE IF (diag_type == DIAG_TYPE_SCALAPACK) THEN
+      IF ((nmo < 16) .OR. (diag_type == DIAG_TYPE_SCALAPACK)) THEN
```
dev-zero: Maybe add thie constant as a `PARAMETER` and put the documentation next to that constant rather than in the `\History` block to make it a bit clearer?

mkrack:
- Add an ELPA_NEIGVEC_MIN keyword in the &GLOBAL section which allows for a restriction of the ELPA usage to larger matrices. The current default is 16 eigenvectors.
- The default for ELPA_FORCE_REDISTRIBUTE has been changed from .FALSE. to .TRUE.
- Printout of ELPA information revised

CP2K performs a redistribution of BLACS matrices in order to fulfill the requirements of ELPA which is especially needed for small matrices and a large number of MPI ranks. It seems that the heuristics in CP2K for an automatic redistribution worked fine for older ELPA releases, but not for the most recent ELPA version. Forcing a redistribution for matrices with the keyword ELPA_FORCE_REDISTRIBUTE resolves the problem with the [test case](https://github.com/cp2k/cp2k/issues/1444#issuecomment-836541273) of @juerghutter (#1444) like using 16 for ELPA_NEIGVEC_MIN which uses ScaLAPACK as a fallback for the small matrices.
Nevertheless, the ELPA_NEIGVEC_MIN keyword is still useful to restrict the usage of the ELPA eigensolver to large matrices, since
- small problems should not be sent to the GPU
- a redistribution requires an unneeded overhead and further checks for small systems

PR #1518, opened 2021-05-12
arch: Use -fp-model strict for xc_sr_lda.F with Intel compiler

### LINKED ISSUE #1434: Recent Intel compilers give wrong results and refuse loop unrolling
The test input [CH4-rsLDA.inp](https://github.com/cp2k/cp2k/blob/master/tests/QS/regtest-hfx-wfn-fitting/CH4-rsLDA.inp) returns wrong results using Intel compiler versions newer than 18.0.5.

oschuett:
Fixes #1434.

### DIFF
--- a/arch/Linux-x86-64-intel-minimal.psmp
+++ b/arch/Linux-x86-64-intel-minimal.psmp
@@ -11,7 +11,7 @@ LD        = mpiifort
 AR        = ar -r
 
 # Only one test input (CH4-rsLDA.inp) gives wrong results with inlining enabled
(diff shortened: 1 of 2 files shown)


mkrack:
@oschuett I can confirm that this fix resolves #1434

fstein93:
I am even thinking of removing the code because we have it available with LiBXC 5 plus higher derivatives.

oschuett:
I'm going to merge this PR to unblock the release. @fstein93, feel free to remove the code if you like.

PR #1520, opened 2021-05-13
Toolchain: Upgrade to COSMA 2.5.0

### LINKED ISSUE #1303: Performance regression in cp_fm_cholesky_invert due to COSMA
Some MD benchmarks indicate that depending on the system and distribution, using COSMA as a drop-in replacement for ScaLAPACK's `p?gemm` may have a detrimental effect on the performance (2-10x slowdown). Being able to runtime-enable/disable COSMA might be useful to use the same binary (like in an HPC environment).

Results from single runs of the QS H2O-64 benchmark on Alps/Eiger for different combinations of nodes/ranks-per-node/threads-per-rank:

```
      ndsXrkXthr  w COSMA  w/o COSMA
H2O-64 001x16x001  113.325    91.972
H2O-64 001x16x008   52.169    58.794
H2O-64 001x16x016   52.282    57.014
H2O-64 001x32x001   92.991    70.400
H2O-64 001x32x004   52.922    55.827
H2O-64 001x32x008   47.206    52.074
H2O-64 001x64x001   82.751    59.122
H2O-64 002x16x001   86.097    60.065
H2O-64 002x32x001   49.011    44.661
H2O-64 002x32x004   38.814    41.799
H2O-64 002x32x008   36.725    34.418
H2O-64 002x64x001   44.567    42.155
H2O-64 002x64x002   43.750    40.968
H2O-64 004x16x001   42.407    40.835
H2O-64 004x16x008   23.166    29.085
H2O-64 004x16x016   28.860    38.558
H2O-64 004x32x004   27.440    34.376
H2O-64 004x32x008   33.463    33.509
H2O-64 004x64x001   38.561    31.303
H2O-64 004x64x004   40.259    34.583
H2O-64 008x16x001   26.896    34.986
H2O-64 008x16x016   27.898    34.209
H2O-64 008x32x001   22.859    29.652
H2O-64 008x32x004   21.519    30.548
H2O-64 008x32x008   32.602    28.152
H2O-64 008x64x001   97.097    57.198
H2O-64 008x64x002   56.865    30.906
H2O-64 008x64x004   59.107    23.895
H2O-64 016x16x001   29.727    22.195
H2O-64 016x16x008   24.580    30.904
H2O-64 016x32x001   57.219    28.038
H2O-64 016x32x008   66.675    34.022
H2O-64 016x64x001   57.809    23.472
H2O-64 016x64x002   59.376    23.472
H2O-64 016x64x004   45.507    25.930
H2O-64 032x16x001   46.417    14.291
H2O-64 032x16x008   40.335    23.079
H2O-64 032x16x016   51.210    31.355
H2O-64 032x32x001   26.210    19.150
H2O-64 032x32x004   34.538    22.598
H2O-64 032x32x008   45.671    33.595
H2O-64 032x64x001   70.165    34.154
H2O-64 032x64x002   68.457    34.545
H2O-64 032x64x004   76.184    65.280
H2O-64 064x16x001   26.098    16.464
H2O-64 064x16x008  114.806   118.204
H2O-64 064x16x016   54.157    59.624
H2O-64 064x32x001   42.855    30.415
H2O-64 064x32x004   53.473    37.060
H2O-64 064x32x008   63.877    51.211
H2O-64 064x64x001  221.895    60.353
H2O-64 064x64x002   79.736   203.070
H2O-64 064x64x004   91.319    68.089
```

oschuett:
I took the liberty to grab the COSMA upgrade from #1517 to decouple it from the SIRIUS upgrade and unblock our release.

Fixes #1303.


oschuett:
I just realized that [COSMA v2.3.0](https://github.com/eth-cscs/COSMA/releases/tag/v2.3.0) is from September 2020. So, it's not yet the new release that will fix #1303.

tools/toolchain/scripts/stage4/install_cosma.sh
```diff
@@ -144,7 +141,7 @@ case "$with_cosma" in
     ;;
 esac
 if [ "$with_cosma" != "__DONTUSE__" ]; then
-  COSMA_LIBS="-lcosma_pxgemm -lcosma -lgrid2grid IF_CUDA(-lTiled-MM|)"
+  COSMA_LIBS="-lcosma_pxgemm -lcosma"
```
kabicm: I think this line would have to be: 
```
COSMA_LIBS="-lcosma_pxgemm -lcosma -lcosta_scalapack -lcosta IF_CUDA(-lTiled-MM|)"
```

Now that COSMA is a shared library, I think scalapack would also have to be built as a shared library, but I guess this is already the case?

dev-zero: Not in the toolchain, no.

dev-zero: @kabicm do you know whether the linker strictly adheres to the order when resolving symbols provided by both a shared and a static library or whether it moves static libraries to the top? If the library being static or dynamic changes preference we must ensure both are either static or shared I guess.

oschuett: Thanks for the hint. I've tried building scalapack as a shared library but then ELPA complained. 

I think, the best solution would be to link COSMA statically and then call `cosma::pxgemm<double>` directly in 
[cp_fm_gemm](https://github.com/cp2k/cp2k/blob/master/src/fm/cp_fm_basic_linalg.F#L362). This would keep the toolchain complexity low and also satisfy our long-standing desire to have an off-switch.

kabicm: @dev-zero this is a very good question and I am really not sure. Since i used function interpose, it requires that both scalapack and cosma are built as shared libraries.

@oschuett this is fine with me. In fact, instead of `cosma_pxgemm`, we also provide `cosma_prefixed_pxgemm` library which actually provides `cosma_pdgemm, cosma_psgemm, cosma_pcgemm` and `cosma_pzgemm` functions.

oschuett:
I've managed to build CP2K with COSMA 2.4.0. In the process I've removed Valgrind and reflapack from the toolchain, because they got in the way and were (somewhat) broken and (AFAIK) unused anyways. I'll try to bring back Valgrind when time permits.

Unfortunately, we're now getting wrong results with Cuda. I've seen the problem on Pascal and Volta.

So, here is the complete list of open problems:

- [ ] `-DCOSMA_SCALAPACK` and `-DBUILD_SHARED_LIBS=NO` are [incompatible](    
 https://github.com/cp2k/cp2k/blob/2968e9409a77bc8538f3f7bccbe6415ddca9c4f3/tools/toolchain/scripts/stage4/install_cosma.sh#L51).
- [ ]  `$OPENBLASROOT` is no longer recognized by cmake - I had to export `$OPENBLAS_ROOT`.
- [ ] Despite `-DBUILD_SHARED_LIBS=NO` the shared OpenBLAS is still needed to make FindBLAS happy. Could be [this issue](https://gitlab.kitware.com/cmake/cmake/-/issues/16221).
- [ ]  Add check for `$COSMA_DIM_THRESHOLD` to `cosma_prefixed_pxgemm`.
- [ ]  Remove [the call](https://github.com/kabicm/Tiled-MM/blob/8a7e5a7db06be0689a7edbfab497ca577bf9eee6/src/Tiled-MM/gpu_blas_handle.hpp#L12) to `set_device` from Tiled-MM to enable multi-GPU. See discussion in [#1530](https://github.com/cp2k/cp2k/issues/1530#issuecomment-845505470).
- [x] Call `cudaSetDevice` on the CP2K side before invoking `cosma_pdgemm`.
- [ ] Fix [wrong results](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-pascal-5835a37f_report.txt) with Cuda (probably an OpenMP race condition).

oschuett:
The wrong results with Cuda seems to be an OpenMP race condition. With a single thread it works - with multiple threads the results become random.

kabicm:
This is very unusual, since the GPU backend is not changed and it doesn't use any OpenMP. But there are other parts using OpenMP, so I will investigate.

kabicm:
The changes that we have just made in the current COSMA master branch should fix both issues: https://github.com/cp2k/cp2k/issues/1525 and https://github.com/cp2k/cp2k/issues/1530, as well as the problems mentioned in: https://github.com/cp2k/cp2k/pull/1520#issuecomment-846287877.

Should we release a new COSMA version so that you can test it, or can you test the current master before we release it?

oschuett:
Thanks a lot @kabicm! Everything looks good. Please go ahead and make a release.

kabicm:
Thanks a lot @oschuett! Do you think we still need those "do_cosma" and "-D_COSMA" flags? Since now basically we can disable COSMA just by setting `COSMA_DIM_THRESHOLD` high enough. We can also define a special value for this variable in COSMA, like `inf` or even easier `-1` that would mean -> always use scalapack. What do you think?

oschuett:
> Do you think we still need those "do_cosma" and "-D_COSMA" flags? 

Yes, those are still needed in case somebody wants to build CP2K without COSMA.

kabicm:
Ok, in that case there is no need for those special values.

kabicm:
It's released now. Since it fixes some important issues, it's promoted to `v2.5.0`.

PR #1523, opened 2021-05-13
Remove srlda and bump LibXC to 5.1.4

fstein93:
Can someone add the new tarball please?


oschuett:
I've uploaded the tarball and restarted the test. Could you please also revert my https://github.com/cp2k/cp2k/commit/cfbb1e63b8611ad04b32317830b90d5e804692cb as part of this PR.

I noticed that the regtest results changed significantly. Is this because you now picked a different functional or was our implementation flawed?

fstein93:
I have removed the respective lines from the arch files.

I don't know what exactly happened there. I assume it's numerical noise when the longrange LDA DFT energy has the same order of magnitude as the standard DFT energy. That might still happen with LibXC because I have to call all functionals separately (I have four DFT sections) plus I do not know how exactly the different parts are actually calculated within LibXC. Plus, LibXC does not provide a short cut for shortrange DFT. But at least, LibXC provides higher derivatives.

oschuett:
Thanks for the detailed explanation!

PR #1527, opened 2021-05-16
Fix sap_sort() and disable fft-fpga-emulation dashboard tester

oschuett:
The first commit is followup to #705.

@robertschade, please feel free to re-enable the FPGA tester once it's working again.

### DIFF
--- a/src/sap_kind_types.F
+++ b/src/sap_kind_types.F
@@ -234,20 +234,20 @@ CONTAINS
    SUBROUTINE sap_sort(sap_int)
       TYPE(sap_int_type), DIMENSION(:), POINTER          :: sap_int
 
-      INTEGER                                            :: iac, na, sap_int_size
+      INTEGER                                            :: iac, na
 
 ! *** Set up a sorting index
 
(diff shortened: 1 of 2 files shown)


alazzaro:
It turns out that the problem with the parallel region is that CCE wants to optimize the internal sort function, and it cannot get OpenMP to parallelize for that... I'm going to downgrade the error to be a warning...

oschuett:
> CCE wants to optimize the internal sort function, and it cannot get OpenMP to parallelize for that... 

That is really strange given that our [sort function](https://github.com/cp2k/cp2k/blob/master/src/common/array_sort.fypp) does not contain any OpenMP directives.

alazzaro:
No, the problem is not the OpenMP region inside, rather the fact that it doesn't know if we are changing the data structure inside... My hope was to use an intermediate variable for the DO loop range, but it seems it is not enough...
Anyway, now there are tons of other problems, but that's OK (previously CCE was POPT, I would expect some problems with PSMP)...

PR #1533, opened 2021-05-21
Add arch file for GNU with MKL

mkrack:


arch/CRAY-XC40_gnu-mkl.psmp
```diff
@@ -0,0 +1,92 @@
...
+LIBS        = $(PLUMED_DEPENDENCIES) -lz
+LIBS       += $(ELPA_LIB)/libelpa.a
+LIBS       += $(LIBVORI_LIB)/libvori.a
+LIBS       += $(LIBXC_LIB)/libxcf03.a
+LIBS       += $(LIBXC_LIB)/libxc.a
+LIBS       += $(LIBINT_LIB)/libint2.a
+LIBS       += $(SPGLIB_LIB)/libsymspg.a
+LIBS       += $(MKL_LIB)/libmkl_scalapack_lp64.a
+LIBS       += -Wl,--start-group
+LIBS       += $(MKL_LIB)/libmkl_intel_lp64.a
```
hfp, line 87: I think `libmkl_gf_lp64.a` should be linked rather than `libmkl_intel_lp64.a`.

hfp: The same is true for `CRAY-XC40_gnu-mkl.psmp` and `Linux-x86-64-gfortran_opencl_nvidia.sopt`.

mkrack: True, but the binaries work correctly linked in both ways. MKL seems to be robust in that respect.

PR #1548, opened 2021-05-31
Speedup Cuda regtests via new do_regtest.py script.

oschuett:


tools/regtesting/do_regtest.py
```diff
@@ -0,0 +1,437 @@
...
+    with open(cfg.error_summary, "wt", encoding="utf8", errors="replace") as err_fh:
+        for num_done, task in enumerate(asyncio.as_completed(tasks)):
+            batch_result = await task
+            all_results += batch_result.results
+            print(f">>> {batch_result.batch.workdir}")
+            print("\n".join([str(r) for r in batch_result.results]))
+            print(f"<<< {batch_result.batch.workdir} ({num_done + 1}", end="")
+            print(f" of {len(tasks)}) done in {batch_result.duration:.2f} sec")
+            sys.stdout.flush()
+            err_fh.write("\n".join([r.error for r in batch_result.results if r.error]))
```
dev-zero: there should be no need to for the explicit list comprehension here

tools/regtesting/do_regtest.py
```diff
@@ -0,0 +1,437 @@
...
+    print('Plot: name="timings", title="Timing Distribution", ylabel="time [s]"')
+    for p in (100, 99, 98, 95, 90, 80):
+        v = percentile(timings, p / 100.0)
+        print(f'PlotPoint: name="{p}th_percentile", plot="timings", ', end="")
+        print(f'label="{p}th %ile", y={v:.2f}, yerr=0.0')
+
+    print("\n-------------------------------- Summary -------------------------------")
+    total_duration = time.perf_counter() - start_time
+    num_tests = len(all_results)
+    num_failed = sum([r.status in ("TIMEOUT", "RUNTIME FAIL") for r in all_results])
```
dev-zero: dito

tools/regtesting/do_regtest.py
```diff
@@ -0,0 +1,437 @@
...
+    num_wrong = sum([r.status == "WRONG RESULT" for r in all_results])
+    num_ok = sum([r.status == "OK" for r in all_results])
+    print(f"Number of FAILED  tests {num_failed}")
+    print(f"Number of WRONG   tests {num_wrong}")
+    print(f"Number of CORRECT tests {num_ok}")
+    print(f"Total number of   tests {num_tests}")
+    summary = f"\nSummary: correct: {num_ok} / {num_tests}"
+    summary += f"; wrong: {num_wrong}" if num_wrong > 0 else ""
```
dev-zero, line 171:
```suggestion
    summary += f"; wrong: {num_wrong if num_wrong > 0 else ''}"
```

... and then thoe whole thing can be one string:

```
summary = (
    f"\nSummary: correct: {num_ok} / {num_tests}"
    f"; wrong: {num_wrong if num_wrong > 0 else ''}"
    f"; failed: {num_failed  if num_failed > 0 else ''}"
    f"; {total_duration/60.0:.0f}min"
    )

File: tools/regtesting/do_regtest.py [hunk 4]
```diff
@@ -0,0 +1,437 @@
...
+    summary += f"; {total_duration/60.0:.0f}min"
+    print(summary)
+    print("Status: " + ("OK" if num_ok == num_tests else "FAILED") + "\n")
+
+    print("*************************** testing ended ******************************")
+    sys.exit(num_tests - num_ok)
+
+
+# ======================================================================================
+class Config:
```
dev-zero, line 182: this looks like a dataclass

tools/regtesting/do_regtest.py
```diff
@@ -0,0 +1,437 @@
...
+        self.debug = args.debug
+        self.restrictdirs = args.restrictdir if args.restrictdir else [".*"]
+        self.gpu_round_robin = 0
+        self.num_gpu_devices = 2  # TODO detect number of GPUs
+
+    def cp2k_command(self) -> List[str]:
+        if self.mpiranks == 1:  # Speedup non-mpi test by using all available GPUs.
+            self.gpu_round_robin = (self.gpu_round_robin + 1) % self.num_gpu_devices
+            os.environ["CUDA_VISIBLE_DEVICES"] = str(self.gpu_round_robin)
+        os.environ["OMP_NUM_THREADS"] = str(self.ompthreads)
```
dev-zero: I'd advise against manipulating the current environ (especially when you're delaying the actual process making use of that environ) and rather pass an explicit dict with the `env` keyword to wherever you're spawning the process where the dict passing is a merge of `os.environ` and what `cp2k_command` should return. So the signature would then rather be:

```
     def cp2k_command(self) -> Tuple(List[str], Dict[str,str]):
```

dev-zero:
Please also keep https://www.cp2k.org/dev:regtesting in mind.

tools/regtesting/do_regtest.py
oschuett: I don't think this will work because I also want `"; failed: "` to disappear when there are no failures.

tools/regtesting/do_regtest.py
oschuett: Not really, because there is quite a lot going in the constructor.

tools/regtesting/do_regtest.py
oschuett: Good point. I refactor the launching of binaries.

tools/regtesting/do_regtest.py
oschuett: Indeed. Thanks!

oschuett:
@dev-zero, thanks a lot for the review!

> Please also keep https://www.cp2k.org/dev:regtesting in mind.

I intent to roll out the new `do_regtest.py` script rather slowly. My main focus for now is to speedup the Cuda run so that we can afford a second GPU run for HIP.

Of course, feel free to take it for a spin on Daint. I'd be very curious. The new script is essentially an improved version of the farming mode.

tools/regtesting/do_regtest.py
dev-zero: `dataclass` doesn't mean you can't have a custom constructor

tools/regtesting/do_regtest.py
dev-zero: ok, I see, but then I'd probably use  a prefixed `if` for clearness rather than appending an empty string:

```python
if num_failed:
    summary += f"..."
```

oschuett: If this were some serious business logic then I would agree. But just for assembling this string I'd rather not spend the extra lines.

tools/regtesting/do_regtest.py
oschuett: Well, but with a custom constructor one loses most of the advantages of a data class. It would certainly require more lines of code than the current solution.

tools/regtesting/do_regtest.py
dev-zero: it makes code more readable

PR #1557, opened 2021-06-14
Toolchain: Add HIP support

oschuett:


alazzaro:
Nice that you are preparing for that...
I have some questions:
* Do we have a way to test the HIP stuff, i.e. an AMD GPU somewhere?
* I like the idea to have a single `OFFLOAD_CC` variable to set the OFFLOAD compilation (maybe CC is a bit misleading since HIP is C++...) But, why we want to keep `NVCC` and `HIPCC`?

oschuett:
> Do we have a way to test the HIP stuff, i.e. an AMD GPU somewhere?

Unfortunately, I do not have an AMD GPU. My plan was to use HIP on top of Cuda. However, I already ran into some bumps:
 - Cosma uses rocBLAS instead of hipBLAS (https://github.com/eth-cscs/COSMA/issues/92).
 - DBCSR requires runtime compilation, which is not yet support for Cuda (https://github.com/ROCm-Developer-Tools/HIP/issues/268)

> But, why we want to keep NVCC and HIPCC?

Good question. I'll remove them.

alazzaro:
> > Do we have a way to test the HIP stuff, i.e. an AMD GPU somewhere?
> 
> Unfortunately, I do not have an AMD GPU. My plan was to use HIP on top of Cuda. However, I already ran into some bumps:
> 
> * Cosma uses rocBLAS instead of hipBLAS ([eth-cscs/COSMA#92](https://github.com/eth-cscs/COSMA/issues/92)).
> * DBCSR requires runtime compilation, which is not yet support for Cuda ([ROCm-Developer-Tools/HIP#268](https://github.com/ROCm-Developer-Tools/HIP/issues/268))

Maybe we can get some help by AMD on those issues @gsitaram?


> 
> > But, why we want to keep NVCC and HIPCC?
> 
> Good question. I'll remove them.

Well, in DBCSR we have a way to check if we have HIP or CUDA:

https://github.com/cp2k/dbcsr/blob/b0e4849409e8f8a01411fe5009b978a6bc01bee0/.cp2k/Makefile#L134

The drawback is that people have to change their arch files then...

gsitaram:
> > > Do we have a way to test the HIP stuff, i.e. an AMD GPU somewhere?
> > 
> > 
> > Unfortunately, I do not have an AMD GPU. My plan was to use HIP on top of Cuda. However, I already ran into some bumps:
> > 
> > * Cosma uses rocBLAS instead of hipBLAS ([eth-cscs/COSMA#92](https://github.com/eth-cscs/COSMA/issues/92)).
> > * DBCSR requires runtime compilation, which is not yet support for Cuda ([ROCm-Developer-Tools/HIP#268](https://github.com/ROCm-Developer-Tools/HIP/issues/268))
> 
> Maybe we can get some help by AMD on those issues @gsitaram?
> 
> > > But, why we want to keep NVCC and HIPCC?
> > 
> > 
> > Good question. I'll remove them.
> 
> Well, in DBCSR we have a way to check if we have HIP or CUDA:
> 
> https://github.com/cp2k/dbcsr/blob/b0e4849409e8f8a01411fe5009b978a6bc01bee0/.cp2k/Makefile#L134
> 
> The drawback is that people have to change their arch files then...

rocBLAS is recommended for performance reasons although hipBLAS offers portability. If you are able to secure access to an AMD GPU, then I would recommend leaving rocBLAS in.
I will check the status of the hiprtc report internally and send an update.

oschuett:
@gsitaram, thanks for chiming in. Unfortunately, I don't have access to the usual HPC facilities because I lack an academic affiliation. Nor could I find a cloud provider that offers AMD Instinct devices. Hence, I'm afraid for the time being we'll have to use Cuda for integration testing.

> rocBLAS is recommended for performance reasons

Oh, that's good to know. Does this apply to all calls or just some corner cases? I was hoping that the overhead would generally be negligible given that the APIs are very similar.

oschuett:
Now I'm getting a [build error](https://storage.googleapis.com/cp2k-ci/run-cp2k-hip-pascal-8c76c2ce_report.txt) in DBCSR:
```
make[3]: *** No rule to make target '/workspace/cp2k/exts/dbcsr/src/acc/libsmm_acc/../hipblaswrap/hipblas.cpp', needed by 'makedep'.  Stop.
```

It seems the entire `hipblaswrap` directory does not exist anymore.

I think, this is a good opportunity for switching to cmake (#351).

alazzaro:
> Now I'm getting a [build error](https://storage.googleapis.com/cp2k-ci/run-cp2k-hip-pascal-8c76c2ce_report.txt) in DBCSR:
> 
> ```
> make[3]: *** No rule to make target '/workspace/cp2k/exts/dbcsr/src/acc/libsmm_acc/../hipblaswrap/hipblas.cpp', needed by 'makedep'.  Stop.
> ```
> 
> It seems the entire `hipblaswrap` directory does not exist anymore.
> 
> I think, this is a good opportunity for switching to cmake (#351).

This is expected, unfortunately. I will take a look in the next days. I still think we should keep v2 of DBCSR with the makefile for CP2K....

oschuett:
> I still think we should keep v2 of DBCSR with the makefile for CP2K....

I guess, until we find a solution for the runtime compilation it doesn't matter because we can't enable `DBCSR_ACC` for HIP anyways.

Issue #1211, opened 2020-12-09
Spurious runtime failures with Farming and SIRIUS


dev-zero:
The farming regtester shows spurious runtime failures at least since commit 5662ec2 or even earlier:

```
Program received signal SIGABRT: Process abort signal.

Backtrace for this error:
 Job finished:           11
 Job:           12  Dir: /workspace/cp2k/regtesting/local/psmp/TEST-local-psmp-2020-10-07_19-13-44/tests/QS/regtest-admm-gapw assigned to group            0
#0  0x7f85c2d8120f in ???
#1  0x7f85c2d8118b in ???
#2  0x7f85c2d60858 in ???
#3  0x7f85c46277fb in _ZN9__gnu_cxx27__verbose_terminate_handlerEv
	at /opt/cp2k-toolchain/build/gcc-10.2.0/libstdc++-v3/libsupc++/vterminate.cc:95
#4  0x7f85c4632895 in _ZN10__cxxabiv111__terminateEPFvvE
	at /opt/cp2k-toolchain/build/gcc-10.2.0/libstdc++-v3/libsupc++/eh_terminate.cc:48
#5  0x7f85c4632900 in _ZSt9terminatev
	at /opt/cp2k-toolchain/build/gcc-10.2.0/libstdc++-v3/libsupc++/eh_terminate.cc:58
#6  0x7f85c4632b93 in __cxa_throw
	at /opt/cp2k-toolchain/build/gcc-10.2.0/libstdc++-v3/libsupc++/eh_throw.cc:95
 Job finished:           93
 Job:           94  Dir: /workspace/cp2k/regtesting/local/psmp/TEST-local-psmp-2020-10-07_19-13-44/tests/QS/regtest-debug-2 assigned to group            2
#7  0x2f553b7 in _ZN5utils9terminateEPKciRKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
	at /opt/cp2k-toolchain/build/SIRIUS-7.0.0/src/utils/utils.hpp:53
#8  0x2f55541 in _ZN5utils9terminateEPKciRKNSt7__cxx1118basic_stringstreamIcSt11char_traitsIcESaIcEEE
	at /opt/cp2k-toolchain/build/SIRIUS-7.0.0/src/utils/utils.hpp:59
#9  0x2ffbd13 in _ZNK6sirius4Band44diag_full_potential_first_variation_davidsonERNS_13Hamiltonian_kE
[...]

PR #1558, opened 2021-06-15
[grid] Add hip backend

mtaillefumier:
- Add hip backend for AMD GPU support. The same code can also be used for NVIDIA
- simple template programming used for the hip kernels. The interface is agnostic (use of template functions for cuda/hipMemcpy etc...) and the task list data structure can be reused across backend.
- cross platform compatible since there are no hardware specific (such as inline asm) code.
- algorithm are tailored for minimal use of atomic operations (not possible to avoid them for collocate) at the prize of more memory use. It means in particular that calculating the hab or the coefficients for collocate is split from the main collocate and integrate kernels. It improves occupancy and has no drawbacks in performance.
- does not assume specific device_id. If cp2k provides one then it will take it otherwise defaulting to 0 (since the backend is not mpi aware).
- Minimal changes to the Makefile (it is hacky) -> cmake
- single kernel for orthorhombic and non orthorhombic cases. Orthorhombic case consider the generic case of a orthogonal basis for the grid lattice vectors (the dh is not necessarily diagonal).  
- Performances : H2O-32.inp (benchmarks/QS) in seconds. Results taken from cp2k timings

| V100                            |                  |           |           |             |              |             |
|------------------------------ | -------------- | ------- | -------- | --------- | ---------- | --------- |
| grid_integrate_task_list |           116 | 12.3  |  4.437 |   4.437 |   4.437 |   4.437 |
| grid_collocate_task_list |           116 | 9.6  |  2.614  |  2.614  |  2.614  |  2.614 |

| Mi100                         |                   |           |             |         |            |        |
| ----------------------------- | -------------- | -------- | --------- | ------ | -------- | ------ |
|grid_integrate_task_list     |      116| 12.3 |   4.140  |  4.140  |  4.140 |   4.140|
|grid_collocate_task_list     |      116|  9.6  |  4.459  |  4.459   | 4.459  |  4.459 |

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -319,6 +319,26 @@ SIRIUS is a domain specific library for electronic structure calculations.
   please see <https://brehm-research.de/bqb> for more information as well as
   the `bqbtool` to inspect BQB files.
 
+### 2u. ROCM/HIP (Support for AMD GPU)
+
+The code for the grid backend was developed and tested on Mi100 but should work
+out of the box on Nvidia hardware as well.
+
+- USE `-D__GRID_HIP` to enable AMD GPU support for collocate and integrate
+  runtines.
+- Add `GPUVER=Mi50, Mi60, Mi100`
+- Add 'OFFLOAD_CC = hipcc'
+- Add  `-lamdhip64` to the `LIBS` variable
+- Add `OFFLOAD_FLAGS ='-fopenmp -m64 -pthread -fPIC -D__GRID_HIP -O2
+  --offload-arch=gfx908 --rocm-path=$(ROCM_PATH)`' where 'ROCM_PATH' is the path
+  where the rocm dsk resides. Architectures Mi100 (gfx908), Mi50 (gfx906)
+- the hip backend for the grid library supports nvidia hardware as well. It uses
+  the same code and can be used to validate the backend in case of access to
+  Nvidia hardware only. To get the compilation working, follow the steps above
+  and set the `OFFLOAD_FLAGS` with right `nvcc` parameters (see the cuda section
+  of this document). The environment variable `HIP_PLATFORM` should be set to
+  `HIP_PLATFORM=nvidia` to indicate to hipcc to use the nvcc compiler instead.
+
 <!---
 ### 2u. LibMaxwell (External Maxwell Solver)
 
--- a/Makefile
+++ b/Makefile
@@ -68,15 +68,15 @@ endif
 
 # Declare PHONY targets =====================================================
 .PHONY : $(VERSION) $(EXE_NAMES) \
-         dirs makedep default_target all \
-         toolversions extversions extclean libcp2k cp2k_shell exts python-bindings \
-         pre-commit pre-commit-clean \
-         pretty precommit precommitclean doxygenclean doxygen \
-         fpretty fprettyclean \
-         doxify doxifyclean \
-         install clean realclean distclean mrproper help \
-         test testbg testclean testrealclean \
-         data \
+				 dirs makedep default_target all \
+				 toolversions extversions extclean libcp2k cp2k_shell exts python-bindings \
+				 pre-commit pre-commit-clean \
+				 pretty precommit precommitclean doxygenclean doxygen \
+				 fpretty fprettyclean \
+				 doxify doxifyclean \
+				 install clean realclean distclean mrproper help \
+				 test testbg testclean testrealclean \
+				 data \
 	 $(EXTSPACKAGES)
 
 # Discover files and directories ============================================
@@ -88,10 +88,21 @@ OBJ_SRC_FILES  = $(shell cd $(SRCDIR); find . ! -path "*/preprettify/*" -name "*
 OBJ_SRC_FILES += $(shell cd $(SRCDIR); find . ! -path "*/preprettify/*" ! -path "*/python*" -name "*.c")
 OBJ_SRC_FILES += $(shell cd $(SRCDIR); find . ! -path "*/preprettify/*" -name "*.cpp")
 OBJ_SRC_FILES += $(shell cd $(SRCDIR); find . ! -path "*/preprettify/*" -name "*.cxx")
-OBJ_SRC_FILES += $(shell cd $(SRCDIR); find . ! -path "*/preprettify/*" -name "*.cc")
-ifneq ($(OFFLOAD_CC),)
+OBJ_SRC_FILES += $(shell cd $(SRCDIR); find . ! -path "*/preprettify/*" -path "./grid/hip" -prune -type f -name "*.cc")
+
+ifneq (,$(findstring hipcc,$(OFFLOAD_CC)))
+OBJ_SRC_FILES += ./grid/hip/grid_hip_integrate.cc
+OBJ_SRC_FILES += ./grid/hip/grid_hip_collocate.cc
+OBJ_SRC_FILES += ./grid/hip/grid_hip_context.cc
+#exclude the src/grid/gpu directory from the list to avoid poluting the C namespace with empty functions.
+OBJ_SRC_FILES += $(shell cd $(SRCDIR); find . ! -path "*/preprettify/*" -path "./grid/gpu" -prune -type f -name "*.cu")
+endif
+
+# if we compile for cuda by directly calling nvcc then include all cuda files.
+ifneq (,$(findstring nvcc,$(OFFLOAD_CC)))
 OBJ_SRC_FILES += $(shell cd $(SRCDIR); find . ! -path "*/preprettify/*" -name "*.cu")
 endif
+
 
 # Included files used by Fypp preprocessor and standard includes
 INCLUDED_SRC_FILES = $(filter-out base_uses.f90, $(notdir $(shell find $(SRCDIR) ! -path "*/preprettify/*" -name "*.f90")))
@@ -490,10 +501,10 @@ cp2k_info.o: $(GIT_REF)
 
 # Add some practical metadata about the build.
 FCFLAGS += -D__COMPILE_ARCH="\"$(ARCH)\""\
-           -D__COMPILE_DATE="\"$(shell date)\""\
-           -D__COMPILE_HOST="\"$(shell hostname)\""\
-           -D__COMPILE_REVISION="\"$(strip $(REVISION))\""\
-           -D__DATA_DIR="\"$(DATA_DIR)\""
+					 -D__COMPILE_DATE="\"$(shell date)\""\
+					 -D__COMPILE_HOST="\"$(shell hostname)\""\
+					 -D__COMPILE_REVISION="\"$(strip $(REVISION))\""\
+					 -D__DATA_DIR="\"$(DATA_DIR)\""
 
 # $(FCLOGPIPE) can be used to store compiler output, e.g. warnings, for each F-file separately.
 # This is used e.g. by the convention checker.
@@ -527,6 +538,21 @@ endif
 %.o: %.cu
 	$(OFFLOAD_CC) -c $(OFFLOAD_FLAGS) $<
 
+
+ifneq (,$(findstring hipcc,$(OFFLOAD_CC)))
+
+#define specific rules for the hip backend since they need to be propcessed by hipcc
+
+grid_hip_collocate.o: grid_hip_collocate.cc
+	$(OFFLOAD_CC) -c $(OFFLOAD_FLAGS) $<
+grid_hip_integrate.o: grid_hip_integrate.cc
+	$(OFFLOAD_CC) -c $(OFFLOAD_FLAGS) $<
+grid_hip_context.o: grid_hip_context.cc
+	$(OFFLOAD_CC) -c $(OFFLOAD_FLAGS) $<
+
+#%.o: %.cc
+#	$(OFFLOAD_CC) -c $(OFFLOAD_FLAGS) $<
+endif
 
 # module compiler magic =====================================================
 ifeq ($(MC),)
--- a/src/cp2k_info.F
+++ b/src/cp2k_info.F
@@ -289,6 +289,9 @@ CONTAINS
 #if defined(__GRID_CUDA)
       flags = TRIM(flags)//" grid_cuda"
 #endif
+#if defined(__GRID_HIP)
+      flags = TRIM(flags)//" grid_hip"
+#endif
 
    END FUNCTION cp2k_flags
 
--- a/src/f77_interface.F
+++ b/src/f77_interface.F
@@ -112,7 +112,6 @@ MODULE f77_interface
                                               nnp_env_release,&
                                               nnp_type
    USE offload_api,                     ONLY: offload_get_device_count,&
-                                              offload_get_device_id,&
                                               offload_set_device_id
    USE periodic_table,                  ONLY: init_periodic_table
    USE pw_cuda,                         ONLY: pw_cuda_finalize,&
@@ -277,6 +276,7 @@ CONTAINS
 
          ! Initialize the DBCSR configuration
          ! Attach the time handler hooks to DBCSR
+#if defined __DBCSR_ACC
          IF (offload_get_device_count() > 0) THEN
             CALL dbcsr_init_lib(default_para_env%group, timeset_hook, timestop_hook, &
                                 cp_abort_hook, cp_warn_hook, io_unit=unit_nr, &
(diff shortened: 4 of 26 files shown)


oschuett:
Thanks for the HIP backend! Overall looks good :-)

Let me merge my #1557, then we can start testing it.

mtaillefumier:
No problems. I may have to adjust few things but nothing critical. I have a python error although I never change this python script.

oschuett:
> I have a python error although I never change this python script.

Yeah, that one is on me - I didn't freeze the mypy version. It'll get fixed by #1557 as well.

mtaillefumier:
I did not test building the hip backend for supporting cuda but I can already say that a -I$(ROCM_PREFIX)/include is missing. Give me a little bit of time to test this because it was not in my immediate todo list.

mtaillefumier:
This won't work at all. the toolchain needs debugging.

mtaillefumier:
- [x] Fix compilation hipcc / CUDA
- [x] decide what to to with the failing tests
- [x] split the three cc/hip from the others files (need hipcc for them, the others don't)
- [x] perf regression on Mi100.
- [x] numerical difference between nvidia/amd GPU

oschuett:
> This won't work at all. the toolchain needs debugging.

Yes, your PR arrived a bit too early ;-)

As a stop-gap I'd propose we hard-code `HIP_PLATFORM=nvidia` via #1559.

In the long run we hopefully figure out a way to abstract away the rocm/nvidia differences.

mtaillefumier:
No Worries. The soonish the better so I can clear out the code. I am testing the compilation on hip/cuda right now. Hopefully I will find all errors. It just take a little bit of time since CI testing will require hip/cuda instead of hip/rocm until you have access to hardware. And hipcc documentation is not really explicit about the cuda side of things. I think it is mostly figuring out the parameters to give to the compiler.

mtaillefumier:
Ok, we can generate arch files for Mi50 and Mi100 GPUs as well. The HIP_PLATFORM still need manual intervention. hipcc does not seem to have an command line option to indicate that we want to call nvcc. 
Compilation using the arch files for both AMD and Nvidia pass on Fedora (compiler issue with the stock one, cuda does not support it).  
to trigger one or the other just add --enable-hip --gpu-ver=Mi50 (Mi100) for AMD GPU or --gpu-ver=V100, etc... for Nvidia GPUs

NB :  
the patch '**Remove HIPCC specific variable. Now NVCC=hipcc can be used**' is actually not relevant anymore. I use the OFFLOAD_CC and OFFLOAD_FLAGS everywhere.

oschuett:
Thanks for adding the AMD GPU models.

Now the compilation [fails](https://storage.googleapis.com/cp2k-ci/run-cp2k-hip-pascal-f90c66f6_report.txt) in a different place. You can get the full message by changing [this line](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_regtest.sh#L40).

mtaillefumier:
I will look at it ASAP.

mtaillefumier:
HIP/NVCC does not like assert statements for some reasons.

mtaillefumier:
I have to close it temporarily because it does not work on nvidia hardware. No point of triggering the CI all the time I push something.

mtaillefumier:
even if everything is green wait before merging the PR. Seems that one benchmark is off

oschuett:
The DBCSR version that we're currently using in CP2K has a broken Makefile wrt. HIP. It was [recently fixed](https://github.com/cp2k/dbcsr/pull/452/commits/c57cee07937b191de51b2d061f1651885b91e64d). Hence, we'd need a new DBCSR release and then upgrade to that. However, HIP on cuda will still not work because of https://github.com/ROCm-Developer-Tools/HIP/issues/268.

alazzaro:
> The DBCSR version that we're currently using in CP2K has a broken Makefile wrt. HIP. It was [recently fixed](https://github.com/cp2k/dbcsr/pull/452/commits/c57cee07937b191de51b2d061f1651885b91e64d). Hence, we'd need a new DBCSR release and then upgrade to that. However, HIP on cuda will still not work because of [ROCm-Developer-Tools/HIP#268](https://github.com/ROCm-Developer-Tools/HIP/issues/268).

I can make a new DBCSR RC.
My understanding is that @oschuett requested access to the [AMD cloud machines](https://forms.office.com/Pages/ResponsePage.aspx?id=H5bYPYjkYE6OEagtmU4YPe0_mupu_3hBiIEF5n5QuJlUOE80REFYRTRHWldBSE5OTDBJRlBGR1ZWSC4u&wdLOR=cF88686DF-EA0B-46C7-A39A-05ED61E4648D), so I suggest to test HIP only on AMD machines.

oschuett:
I did indeed apply for an AMD cloud account yesterday. Hopefully, there will be enough resources to run a nightly regressions test - maybe even the single node benchmarks.

However, I think we should still try to test as much as possible before merging a PR. Since the public clouds do not yet offer AMD GPUs we'll have to use HIP on Cuda for the time being.

mtaillefumier:
- I think it should not be a problem to run the single node benchmarks. H2O-32.inp for instance is under 1 min and anything below an hour should not cause too much trouble.
- I finally found the missing code. All regtests pass on K80 (in validation mode on) except the one in the directory **regtest-grid**. They will pass if the validation mode is off but the difference seems to trigger the abort for these 4 tests.
- I will revert the few regtests tolerance to their original value.
- It is definitely a good idea to test this code on NVIDIA and AMD hardware. They should give the same answer....
- I introduced a workaround to be able to test the backend with the CI.

mtaillefumier:
I squashed everything in one commit only.

regtests pass on both NVIDIA and AMD GPUs (with the validation mode off). With validation mode on 4 of the regtest-grid fails.

oschuett:
We're still [getting](https://storage.googleapis.com/cp2k-ci/run-cp2k-hip-pascal-bc722bfe_report.txt) `ABORT in dbcsr_config.F:490 Invalid accelerator device ID`. However, for some reason this happens only in about 50% of cases. Could this somehow be related to the fact that we're running with 2 MPI ranks?

hfp:
> 
> 
> We're still [getting](https://storage.googleapis.com/cp2k-ci/run-cp2k-hip-pascal-bc722bfe_report.txt) `ABORT in dbcsr_config.F:490 Invalid accelerator device ID`. However, for some reason this happens only in about 50% of cases. Could this somehow be related to the fact that we're running with 2 MPI ranks?

If the (Nvidia-)device is in exclusive mode, either MPS must be running or only a single process/rank can access the device. The MPS/daemon is apparently only compatible with CUDA (e.g., does not work with Nvidia's OpenCL implementation). One can capture the output of `nvidia-smi` even as an ordinary user, which will list the compute mode. However, putting the device into non-exclusive mode requires permissions (`nvidia-smi --compute-mode=DEFAULT`).

alazzaro:
> We're still [getting](https://storage.googleapis.com/cp2k-ci/run-cp2k-hip-pascal-bc722bfe_report.txt) `ABORT in dbcsr_config.F:490 Invalid accelerator device ID`. However, for some reason this happens only in about 50% of cases. Could this somehow be related to the fact that we're running with 2 MPI ranks?

This problem was fixed in a recent [DBCSR commit](https://github.com/cp2k/dbcsr/commit/b1ec95400b2858c008d282e69e45e484c6a7d4b1). Give me a day and I will release a new DBCSR RC candidate...

oschuett:
I think, every other run fails. Since I'm using my new [do_regtest.py](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest.py), the cp2k process is re-used unless it crashes. Apparently, for the second run it calls `dbcsr_init_lib` from [cp2k_runs.F](https://github.com/cp2k/cp2k/blob/master/src/start/cp2k_runs.F#L200). So, we'd need the same workaround that Mathieu already added to [f77_interface.F](https://github.com/cp2k/cp2k/pull/1558/files#diff-1a6d61eba8779718232bca5cdb1f1390567d368e60f95e0db93f46bbb344739b) in `cp2k_runs.F` as well.

For the record, the CI uses the [nvidia persistence daemon](https://github.com/cp2k/cp2k-ci/blob/master/toolbox/start_stuff.sh#L11).

mtaillefumier:
Let me implement the same work around. I want to see this regtests results ;).

mtaillefumier:
fingers crossed. I included the same workaround. curiously I never add this problem on my machine (i have 3 gpus on it).

oschuett:
Thanks for adding that workaround so quickly!

> curiously I never add this problem on my machine

The `do_regtest.py` is very new and currently [only used](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_regtest.sh#L61) for the GPU tests in the CI. The idea is to speedup the regtests by re-using the cp2k process and in particular the JIT cache from DBCSR. Otherwise a sizable amount of time is spend on just compiling the DBCSR Cuda kernels.

oschuett:
We're getting close :-) Only [two tests](https://storage.googleapis.com/cp2k-ci/run-cp2k-hip-pascal-0191bd94_report.txt) are giving wrong results. Since those tests have very tight thresholds and the deviation is quite large we should take them serious though.

mtaillefumier:
I could not detect the failure because I compiled cp2k without libint :/. Still searching what's wrong with these two.

Timstone813:
> * Add hip backend for AMD GPU support. The same code can also be used for NVIDIA
> * simple template programming used for the hip kernels. The interface is agnostic (use of template functions for cuda/hipMemcpy etc...) and the task list data structure can be reused across backend.
> * cross platform compatible since there are no hardware specific (such as inline asm) code.
> * algorithm are tailored for minimal use of atomic operations (not possible to avoid them for collocate) at the prize of more memory use. It means in particular that calculating the hab or the coefficients for collocate is split from the main collocate and integrate kernels. It improves occupancy and has no drawbacks in performance.
> * does not assume specific device_id. If cp2k provides one then it will take it otherwise defaulting to 0 (since the backend is not mpi aware).
> * Minimal changes to the Makefile (it is hacky) -> cmake
> * single kernel for orthorhombic and non orthorhombic cases. Orthorhombic case consider the generic case of a orthogonal basis for the grid lattice vectors (the dh is not necessarily diagonal).
> * Performances : H2O-32.inp (benchmarks/QS) in seconds. Results taken from cp2k timings
> 
> V100						
> grid_integrate_task_list	116	12.3	4.437	4.437	4.437	4.437
> grid_collocate_task_list	116	9.6	2.614	2.614	2.614	2.614
> Mi100						
> grid_integrate_task_list	116	12.3	4.140	4.140	4.140	4.140
> grid_collocate_task_list	116	9.6	4.459	4.459	4.459	4.459

Please give the version of rocm, bdcsr, cmake, and the cmake parameters at toolchain. I compile and linker success, But when I run     error: segmentation error 
I use 
`cmake .. -DUSE_MPI=ON -DUSE_HIP=ON -DUSE_SMM=blas  -DUSE_OPENMP=OFF -DCMAKE_Fortran_COMPILER=mpifort -DCMAKE_CXX_COMPILER=hipcc -DUSE_OPENMP=OFF -DUSE_SMM=libxsmm -DUSE_ACCEL=hip -DWITH_CUDA_PROFILING=OFF -DWITH_C_API=ON -DWITH_EXAMPLES=ON -DWITH_GPU=Mi50 -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=ON -DTEST_MPI_RANKS=1 -DTEST_OMP_THREADS=1 -DCMAKE_INSTALL_PREFIX=~/cp2k-hip_backend/exts/dbcsr-install`
And the rocm-3.9.1

mtaillefumier:
> > * Add hip backend for AMD GPU support. The same code can also be used for NVIDIA
> > * simple template programming used for the hip kernels. The interface is agnostic (use of template functions for cuda/hipMemcpy etc...) and the task list data structure can be reused across backend.
> > * cross platform compatible since there are no hardware specific (such as inline asm) code.
> > * algorithm are tailored for minimal use of atomic operations (not possible to avoid them for collocate) at the prize of more memory use. It means in particular that calculating the hab or the coefficients for collocate is split from the main collocate and integrate kernels. It improves occupancy and has no drawbacks in performance.
> > * does not assume specific device_id. If cp2k provides one then it will take it otherwise defaulting to 0 (since the backend is not mpi aware).
> > * Minimal changes to the Makefile (it is hacky) -> cmake
> > * single kernel for orthorhombic and non orthorhombic cases. Orthorhombic case consider the generic case of a orthogonal basis for the grid lattice vectors (the dh is not necessarily diagonal).
> > * Performances : H2O-32.inp (benchmarks/QS) in seconds. Results taken from cp2k timings
> > 
> > V100
> > grid_integrate_task_list	116	12.3	4.437	4.437	4.437	4.437
> > grid_collocate_task_list	116	9.6	2.614	2.614	2.614	2.614
> > Mi100
> > grid_integrate_task_list	116	12.3	4.140	4.140	4.140	4.140
> > grid_collocate_task_list	116	9.6	4.459	4.459	4.459	4.459
> 
> Please give the version of rocm, bdcsr, cmake, and the cmake parameters at toolchain. I compile and linker success, But when I run error: segmentation error
> I use
> `cmake .. -DUSE_MPI=ON -DUSE_HIP=ON -DUSE_SMM=blas -DUSE_OPENMP=OFF -DCMAKE_Fortran_COMPILER=mpifort -DCMAKE_CXX_COMPILER=hipcc -DUSE_OPENMP=OFF -DUSE_SMM=libxsmm -DUSE_ACCEL=hip -DWITH_CUDA_PROFILING=OFF -DWITH_C_API=ON -DWITH_EXAMPLES=ON -DWITH_GPU=Mi50 -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=ON -DTEST_MPI_RANKS=1 -DTEST_OMP_THREADS=1 -DCMAKE_INSTALL_PREFIX=~/cp2k-hip_backend/exts/dbcsr-install`
> And the rocm-3.9.1

1. Compilation might fail with rocm-3.9.1 because of interface changes in hip. I would suggest updating to rocm 4.0 at the very least (the latest version 4.2 is even a better choice)
2. compiling dbcsr with rocm support is highly experimental at that stage and I would suggest you not to activate for now.
3. II would suggest you to use the toolchain or to compile dbcsr without GPU support. The toolchain might be the fastest way to get things working. Note that you need to define **ROCM_PATH** and **LD_LIBRARY_PATH**  accordingly if you use the toolchain. 

An example of toolchain parameters that work (it won't build the cp2k full support of all functionalities though)

`./install_cp2k_toolchain.sh --with-hdf5=system --with-gsl=system --with-cosma=no --with-libvori=no --with-elpa=no --with-libxsmm --with-mkl=system --with-fftw=system --with-libint --with-cmake=system --with-libvdwxc=no --enable-cuda=no --enable-hip=yes --gpu-ver=K80 --with-sirius=no --with-spfft=no --with-spla=no`

if you want all functionalities then

`./install_cp2k_toolchain.sh  --enable-hip=yes --gpu-ver=Mi50 
`

mtaillefumier:
Would it be possible to squash these commits during merge.

hfp:
> Would it be possible to squash these commits during merge.

Are you saying it's ready for merge?  If so, I can merge-squash this PR...

mtaillefumier:
It is ready on my side. But better wait for the conventional cuda backend to pass the regtests. Although I tested the compilation already.

hfp:
> It is ready on my side. But better wait for the conventional cuda backend to pass the regtests. Although I tested the compilation already.

```text
In file included from /workspace/cp2k/src/grid/grid_task_list.h:16,
                 from /workspace/cp2k/src/grid/grid_replay.c:26:
/workspace/cp2k/src/grid/gpu/grid_gpu_task_list.h:19:10: fatal error: cuda_runtime.h: No such file or directory
   19 | #include <cuda_runtime.h>
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
make[3]: *** [/workspace/cp2k/Makefile:519: grid_replay.o] Error 1
/workspace/cp2k/src/grid/cpu/collocation_integration.c:15:10: fatal error: cublas_v2.h: No such file or directory
   15 | #include <cublas_v2.h>
      |          ^~~~~~~~~~~~~
compilation terminated.
```

mtaillefumier:
Well, works with the local_cuda.psmp I have. I modified the arch generation script to include the cuda include directory.

oschuett:
Merged :-) Good work!

mtaillefumier:
cuda pascal fails because of SIRIUS but I touch absolutely nothing related to this.

mtaillefumier:
Cool thanks. I could not understand why the PR was merged. Thought i triggered it somehow.

oschuett:
> cuda pascal fails because of SIRIUS but I touch absolutely nothing related to this.

Yes, that's a known issue: https://github.com/cp2k/cp2k/issues/1211 and https://github.com/electronic-structure/SIRIUS/issues/520.

Sometimes SIRIUS gives wrong results after being re-initialized. In the past this only happened with [farming](https://dashboard.cp2k.org/archive/farming/index.html). Since I've switch the GPU tests to the new [do_regtest.py](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest.py) it also started happening [there](https://dashboard.cp2k.org/archive/cuda-pascal/index.html).

mtaillefumier:
I need to update sirius then. Thanks for the info.

PR #1559, opened 2021-06-16
Toolchain: Hardcode HIP_PLATFORM=nvidia for now

oschuett:


alazzaro:
So, if I understand correctly, we are providing HIP as a way to run on nvidia gpus?

mtaillefumier:
It is kind of mandatory right now since the ci does not have access to amd gpus yet. I am debugging the toolchain so that we may not need it after all. hip supports both platforms so my guess is that it is only activated when hip support is on.

oschuett:
Obviously, we're adding HIP to support AMD GPUs. However, without a CI we'd be flying blind. So, this hip-on-cuda setup provides us with a smoke test. Of course, it would be nice to also test the real thing. Ideally, somebody with access to these devices contributes a nightly dashboard tests.

mtaillefumier:
Go ahead with the merge if it passes the regtests. I can continue fixing the toolchain afterwards.

PR #1572, opened 2021-07-02
[toolchain] Add missing dflags when compiling grid_hip on AMD GPU

mtaillefumier:
I forgot to add the -D__GRID_HIP to the mix

### DIFF
--- a/arch/Linux-x86-64-gfortran-regtest.psmp
+++ b/arch/Linux-x86-64-gfortran-regtest.psmp
@@ -93,7 +93,7 @@ LIBS       += $(LIBINT_LIB)/libint2.a
 LIBS       += $(SPGLIB_LIB)/libsymspg.a
 # Only needed for SIRIUS
 LIBS       += ${SIRIUS_LIB}/libsirius.a
-LIBS       += $(GNU_PATH)/SpFFT/1.0.3/lib/libspfft.a
+LIBS       += $(GNU_PATH)/SpFFT/1.0.4/lib/libspfft.a
 LIBS       += $(GNU_PATH)/SpLA/1.4.0/lib/libspla.a
 LIBS       += $(GNU_PATH)/hdf5/1.12.0/lib/libhdf5.a
 # Only needed for COSMA
--- a/src/offload/offload_buffer.c
+++ b/src/offload/offload_buffer.c
@@ -42,8 +42,8 @@ void offload_create_buffer(const int length, offload_buffer **buffer) {
   // With size 0 cudaMallocHost doesn't null the pointer and cudaFreeHost fails.
   (*buffer)->host_buffer = NULL;
   offload_set_device();
-  OFFLOAD_CHECK(
-      hipMallocHost((void **)&(*buffer)->host_buffer, requested_size));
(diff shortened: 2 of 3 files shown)


oschuett:
For some reason the regtests are now taking a lot more time. It might be a load-balancing issue with the GPUs. I'm looking into it.

mtaillefumier:
I ran the regtests again on Mi100 with a minimal configuration of cp2k. the origin might also come from elsewhere but thanks for looking at this. 

--------------------------------- Summary --------------------------------
Number of FAILED  tests 0
Number of WRONG   tests 0
Number of CORRECT tests 2866
Total number of   tests 2866
GREPME 0 0 2866 0 2866 X

Summary: correct: 2866 / 2866; 44min
Status: OK

--------------------------------------------------------------------------
Regtest took 2618.00 seconds.
--------------------------------------------------------------------------
Mon Jul  5 03:23:38 PDT 2021

oschuett:
The regtests now passed twice ([here](https://storage.googleapis.com/cp2k-ci/run-cp2k-hip-mi100-ccc58bd4_report.txt) and [here](https://storage.googleapis.com/cp2k-ci/run-cp2k-hip-mi100-8d5bb98d_report.txt)). Each time they took 7 minutes. Without the `GRID_HIP` they took only 3 minutes. The four GPUs are fully utilized:
```
======================= ROCm System Management Interface =======================
================================= Concise Info =================================
GPU  Temp   AvgPwr  SCLK     MCLK     Fan   Perf  PwrCap  VRAM%  GPU%  
0    37.0c  94.0W   1502Mhz  1200Mhz  255%  auto  290.0W    7%   100%  
1    37.0c  80.0W   1502Mhz  1200Mhz  255%  auto  290.0W    7%   100%  
2    39.0c  87.0W   1502Mhz  1200Mhz  255%  auto  290.0W    8%   100%  
3    38.0c  101.0W  1502Mhz  1200Mhz  255%  auto  290.0W    7%   100%  
================================================================================
```

For some reason it first re-compiles everything from scratch, which takes another 7 minutes. I thought, I had [fixed that](https://github.com/cp2k/cp2k/commit/a05a6033f539c5a614246f4b274dc60aad943108).

Together this make 14 minutes, which cuts it rather close. Hence, I've now increased the job's wall time again from 15 to 20 minutes. Hopefully, we can figure out both issues and put the limit back to 10 minutes.

mtaillefumier:
That's good. Most of these tests are too small to run full speed on GPUs. It might take longer just to allocate memory on GPU than doing the calculations. Rocm also has fairly long initialization time as well. CPU pinning is also very important for performance oriented runs. I would not be worried if all regtests pass in 15 min or 7 min.

mtaillefumier:
I checked some of the tests that are taking longer than say 30s. **Ar-1.inp** is one of them. If I run it alone the runtime is 14s instead of 50s. They are few of these tests that take longer than expected but it will be difficult to pinpoint the origin of the problem.

oschuett:
It might be that the GPUs are just overloaded. Since these nodes have 128 cores and we're running with two OpenMP threads, each of the four GPUs is shared among 16 processes.

mtaillefumier:
that's very true. What's is the command you are using for running the regtests ? I will try as well.

PR #1574, opened 2021-07-02
regtesting: Fix visible_gpu_devices

oschuett:


tools/dashboard/generate_dashboard.py
```diff
@@ -699,13 +699,13 @@ def parse_report(report_txt: Optional[str], log: GitLog) -> Report:
     if report_txt is None:
         return Report(status="UNKNOWN", summary="Error while retrieving report.")
     try:
-        m = re.search("(^|\n)CommitSHA: (\w{40})\n", report_txt)
+        m = re.search("(^|\n|\r)CommitSHA: (\w{40})[\n\r]", report_txt)
         sha = GitSha(m.group(2)) if m else None
-        summary = re.findall("(^|[\n\r])Summary: (.+)\n", report_txt)[-1][1]
-        status = re.findall("(^|\n)Status: (.+)\n", report_txt)[-1][1]
-        plot_matches = re.findall("(^|\n)Plot: (.+)(?=\n)", report_txt)
+        summary = re.findall("(^|\n|\r)Summary: (.+?)[\n\r]", report_txt)[-1][1]
```
dev-zero: Some notes:

* Regex-strings should be prefixed with `r`
* Rather than making the regex aware of DOS line endings it might be a better idea to fix `retrieve_report` to always return string with `\n` line endings, which it currently does only when returning from cache since Python transparently converts DOS line endings when reading from a file:

```
In [19]: with open("test", "w") as fhandle:
    ...:     fhandle.write("foo\r\nbar")
    ...: 

In [20]: from pathlib import Path

In [21]: Path("test").read_text()
Out[21]: 'foo\nbar'
```
* `.` doesn't match newline characters (unless `re.DOTALL`) is given, hence once `retrieve_report` consistently returns strings with Unix-style line endings the non-eager modifier `?` can be dropped, along together with the `[\n\r]`.
* the final regex would then be: `re.findall(r"^Summary: (.+)", report_txt, re.MULTILINE)`

oschuett: Thanks a lot for your comments. The regexs look indeed much cleaner now.

PR #1575, opened 2021-07-05
Dashboard: Add HIP tests

oschuett:


alazzaro:
I have a naive question: on the dashboard I see that HIP on MI100 tasks 7 minutes, while it is 30 minutes on P100. Is this expected?

oschuett:
The HIP Mi100 test runs on four GPUs in addition to 128 CPU cores, while the HIP Pascal test runs on a single P4 GPU and only 12 CPU cores.

PR #1582, opened 2021-07-09
Fix DBCSR HIP build to use CC and not OFFLOAD_CC

gsitaram:
When linking with OpenBLAS that is built with g++ with OpenMP support and enabling DBCSR's HIP backend, if hipcc is used to build DBCSR's cpp files, then in the final CP2K executable, we see two OpenMP runtimes linked. One (libgomp.so) that OpenBLAS pulled in, and another (libomp.so) that DBCSR pulled in. To avoid this, it is best to build DBCSR using the same host compiler as CP2K and add a flag to indicate that the HIP_PLATFORM is AMD. This commit and an equivalent commit to DBCSR will rely on the setting of USE_ACCEL to determine which compiler and flags to use for compiling DBCSR's cpp files for AMD devices.

I tested on AMD and NVIDIA devices, but not with the OpenCL backend. Also, I am not sure if other parts of CP2K need to be modified, I only fixed INSTALL.md and the generate_arch_files.sh script. Please test and let me know if more modifications are necessary.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -179,7 +179,8 @@ the FFTW3 threading library libfftw3_threads (or libfftw3_omp) is required.
 
 - Specify OFFLOAD_CC (e.g. `OFFLOAD_CC = nvcc`) and
   OFFLOAD_FLAGS (e.g. `OFFLOAD_FLAGS = -O3 -g -w --std=c++11`) variables.
-- Use the `-D__DBCSR_ACC` and `USE_ACCEL=cuda` to enable accelerator support for matrix multiplications.
+- Use the `-D__DBCSR_ACC` and `USE_ACCEL=cuda` to enable accelerator support for 
+  matrix multiplications.
 - Add `-lstdc++ -lcudart -lnvrtc -lcuda -lcublas` to LIBS.
 - Specify the GPU type (e.g. `GPUVER   = P100`),
   possible values are K20X, K40, K80, P100, V100.
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -179,7 +179,7 @@ the FFTW3 threading library libfftw3_threads (or libfftw3_omp) is required.
 
 - Specify OFFLOAD_CC (e.g. `OFFLOAD_CC = nvcc`) and
   OFFLOAD_FLAGS (e.g. `OFFLOAD_FLAGS = -O3 -g -w --std=c++11`) variables.
-- Use the `-D__DBCSR_ACC` and `USE_ACCEL=cuda` to enable accelerator support for 
+- Use the `-D__DBCSR_ACC` and `USE_ACCEL=cuda` to enable accelerator support for
   matrix multiplications.
 - Add `-lstdc++ -lcudart -lnvrtc -lcuda -lcublas` to LIBS.
 - Specify the GPU type (e.g. `GPUVER   = P100`),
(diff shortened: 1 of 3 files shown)


INSTALL.md
```diff
@@ -179,7 +179,7 @@ the FFTW3 threading library libfftw3_threads (or libfftw3_omp) is required.
 
 - Specify OFFLOAD_CC (e.g. `OFFLOAD_CC = nvcc`) and
   OFFLOAD_FLAGS (e.g. `OFFLOAD_FLAGS = -O3 -g -w --std=c++11`) variables.
-- Use the `-D__DBCSR_ACC` to enable accelerator support for matrix multiplications.
+- Use the `-D__DBCSR_ACC` and `USE_ACCEL=cuda` to enable accelerator support for matrix multiplications.
```
mtaillefumier: Formatting : need to the line after for. That's why the CI fails.

mtaillefumier (approved):
I have no objection to merge this. The formatting in the INSTALL.md needs minor changes though

INSTALL.md
gsitaram: Added a newline after for to break up the sentence into two.

oschuett: Unfortunately, [markdownlint](https://github.com/DavidAnson/markdownlint) is still [complaining](https://storage.googleapis.com/cp2k-ci/run-cp2k-precommit-ed724f01_report.txt) about line length and trailing spaces. You can run this checks yourself via `make pretty`.

gsitaram: make pretty was useful. Removed all those trailing whitespaces.

oschuett:
The build with HIP [fails](https://storage.googleapis.com/cp2k-ci/run-cp2k-hip-mi100-0784d801_report.txt). Does this change require an upgrade of DBCSR?

mtaillefumier:
Hmm. I did not activate the hip backend by default for some reasons (i do not recall which one) so when we built things before this PR dbcsr was compiled for cpu only. The main reason for this PR is to fix the build system when dbcsr is compiled with hip support but something I did not think off when I reviewed the PR is that we work with a recent version of dbcsr. @gsitaram  might be able to confirm this.

mtaillefumier:
The solution right now is to comment USE_ACCEL=hip when we generate the arch files.

gsitaram:
> The solution right now is to comment USE_ACCEL=hip when we generate the arch files.

Yes, using `USE_ACCEL=cuda/hip/opencl` seemed cleaner than relying on whether `OFFLOAD_CC` is `nvcc/hipcc`. USE_ACCEL is used only for DBCSR and `__DBCSR_ACC` will still control whether the GPU backend is used or not.

We want to keep OFFLOAD_CC and OFFLOAD_FLAGS as is because CP2K's grid GPU backend uses them.

oschuett:
> We want to keep OFFLOAD_CC and OFFLOAD_FLAGS as is because CP2K's grid GPU backend uses them.

Since the [acc](https://github.com/cp2k/dbcsr/tree/develop/src/acc) package was moved out of CP2K together with DBCSR, we now have to re-implement that functionality. I chose the new name  [offload](https://github.com/cp2k/cp2k/tree/master/src/offload) to avoid confusion.

So, if this new flag is supposed to be also used within CP2K then we should call it something like `OFFLOAD_VENDOR`.

oschuett:
Alternatives that come to mind are `OFFLOAD_TARGET`,  `OFFLOAD_PLATFORM`, or simply `OFFLOAD=<vendor>`.

Btw, I'm planing to refactor those flags: Currently, each hotspot has to be enabled individually, e.g. `-D__PW_CUDA  -D__GRID_CUDA`. In the future I'd like to have a single centralized flag to enable all GPU code. For debugging we should still have individual disable flags, e.g. `-D__OFFLOAD=hiprocm -D__NO_OFFLOAD_PW`.

alazzaro:
@gsitaram I've introduced your changes as part of my [PR](#1640 ). 
I've included @oschuett suggestion to use `OFFLOAD_TARGET` for cuda or hip within CP2K toolchain.

PR #1586, opened 2021-07-12
Offload: Add timing ranges and mem_info

oschuett:


oschuett:
This PR makes the Nvidia Tools Extensions (`libnvToolsExt`) a hard dependency. The cost of calling that API without a profiler is negligible. So, there is no real downside to always including it.

gsitaram:
@oschuett, I am not sure it will work with roctx ranges because I don't see the include file (`roctracer/roctx.h`) or the libraries that it needs to link with (`-L${ROCM_PATH}/lib -lroctx64 -lroctracer64`). I will have to test. 
Moreover, in some cases, we don't enable any OFFLOAD backend in CP2K, but only in COSMA. As a novice CP2K user, I have found roctx ranges in the timeline extremely useful to understand which function calls which others by just adding these `roctxRangePush()` and `roctxRangePop()` calls. In your new offload package proposal, it seems like `__OFFLOAD_HIP` will be set only if `__GRID_HIP` is enabled. Do you have any thoughts about making it work for both cases (with and without `__GRID_HIP`)?

oschuett:
It should possible to just set ` __OFFLOAD_HIP` without `__GRID_HIP`. However, I'm not sure it will be of great value. Essentially, you'll only get a single [timing range](https://github.com/cp2k/cp2k/blob/master/src/cp_gemm_interface.F#L163) around our call into `cosma_pdgemm`. If you want to get more visibility into cosma then you have to ask its authors to add profiling too.

> I don't see the include file (roctracer/roctx.h) or the libraries that it needs to link with (`-L${ROCM_PATH}/lib -lroctx64 -lroctracer64`).

Right, there are still some technical kinks to be ironed out. I'm on it.

oschuett:
Actually, CP2K also has a builtin feature for writing a callgraph: https://manual.cp2k.org/trunk/CP2K_INPUT/GLOBAL.html#CALLGRAPH

gsitaram:
> Actually, CP2K also has a builtin feature for writing a callgraph: https://manual.cp2k.org/trunk/CP2K_INPUT/GLOBAL.html#CALLGRAPH

In that case, let's not overcomplicate things. We can enable roctx support by default if one of the HIP backends are enabled.

alazzaro:
> This PR makes the Nvidia Tools Extensions (`libnvToolsExt`) a hard dependency. The cost of calling that API without a profiler is negligible. So, there is no real downside to always including it.

@oschuett Time ago (~2y) we tested the profiling enabled by default in DBCSR and the performance penalty was not negligible at all + some memory footprint increase. Are you sure about the performance penalty?

oschuett:
I ran some benchmarks and there indeed seems to be a significant performance penalty for `libnvToolsExt`. Hence, I'll introduce a  `__OFFLOAD_PROFILING` flag once I actually get it to work. Good catch @alazzaro!

alazzaro:
> I ran some benchmarks and there indeed seems to be a significant performance penalty for `libnvToolsExt`. Hence, I'll introduce a `__OFFLOAD_PROFILING` flag once I actually get it to work. Good catch @alazzaro!

Let me elaborate on what I found time ago (in DBCSR), for the record. Apart from what [NVIDIA says](https://docs.nvidia.com/gameworks/content/developertools/mobile/tegra_system_profiler/3.6-default/05-nvtx-trace.htm), the calls to the profiler API have minimal performance effect only if there is a microsecond (I would say more) between _any_ two calls, including two consecutive pop's or push's. I mean, it doesn't need to be >1 microsecond spent in a range. It turns out that this is a common pattern in CP2K (and DBCSR), for example:

timeset full function
timeset part 1
....
timestop part1
timeset part 2
...
timestop part2
timestop full function

Those 3 sets of API calls are consecutive and this is where we see the overhead... 
At that time, my conclusion was that we need to lower the granularity of the timers for the GPU (i.e. changing timeset API to reduce the number of GPU ranges), but a change in the timers API was somehow discussed in the past and rejected...

mkrack:
This commit seems to break the compilation on 32 bit architectures ([armv7l](https://www.cp2k.org/static/regtest/trunk/Linux-armv7l-gnu-regtest/psmp/regtest-0) and [i386](https://storage.googleapis.com/cp2k-ci/dashboard_i386_report.txt)).

oschuett:
@alazzaro, thanks for the details.

I did a bit of experimenting myself: I emptying the `timeset_handler` and `timestop_handler` routines in [timings.F](https://github.com/cp2k/cp2k/blob/master/src/common/timings.F) and then ran the `H2O-32.inp` benchmark. Without the timing regions the benchmark ran 3% faster. So, I guess we could try to remove some of the smaller regions - especially in `mpiwrap` and `dbcsr`. However, we also have bigger fish to fry.

PR #1589, opened 2021-07-14
Bump to LibXC 5.1.5

fstein93:
Can any of the admins add the new tarball please?

### DIFF
--- a/tools/toolchain/scripts/stage3/install_libxc.sh
+++ b/tools/toolchain/scripts/stage3/install_libxc.sh
@@ -10,7 +10,7 @@
 SCRIPT_DIR="$(cd "$(dirname "$SCRIPT_NAME")/.." && pwd -P)"
 
 libxc_ver="5.1.5"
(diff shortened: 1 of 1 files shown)


oschuett:
Their [homepage](https://www.tddft.org/programs/libxc/download/) seems to be down. Therefore, I've [download](https://www.cp2k.org/static/downloads/libxc-5.1.5.tar.gz) the tarball from https://gitlab.com/libxc/libxc/-/releases/5.1.5. However, it has a different sha256: `101d6ea9e013006deae074843f0d02ab2813e16734e47ff7b0551babc4497163`.

fstein93:
I think the git tarball does not work because it misses some files created by autoconf.

oschuett:
The tddft.org server is back up. However, I've got yet another sha256: `02e4615a22dc3ec87a23efbd3d9be5bfad2445337140bad1720699571c45c3f9` ¯\_(ツ)_/¯

fstein93:
Thanks

PR #1592, opened 2021-07-15
Do not compile fourth derivatives in LibXC

fstein93:
We do not use fourth derivatives in our code.


dev-zero:
Since our toolchain scripts double as documentation for dependencies atm, please add a comment stating that this is not strictly necessary and that a "vanilla" libxc (e.g. one where fourth derivatives are compiled in) works as well.

PR #1594, opened 2021-07-16
Try to implement a version of npt ensemble for absolute frozen coordinates

pclabaut:


ducryf:
Would it not be more appropriate to add a keyword to  MOTION / CONSTRAINT / FIXED_ATOMS instead of adding a new ensemble type to MOTION / MD? Then a similar feature could be added for CELL_OPT too which has the same "issue" as NPT.

pclabaut:
I agree, it would probably make more sense.
I chose this option for now because it felt quicker to implement (theoretically, just duplicate everything related to npt_f) and I am severely running out of time.
However, if you are willing to help me, I would gladly change this.

ducryf:
I am not referring to the implementation, only to the keyword. You could still have two ensembles internally and select npt_f_ensemble or npt_fa_ensemble depending on whether MOTION / CONSTRAINT / FIXED_ATOMS / FIX_CARTESIAN (or something like that) is present or not.

pclabaut:
Yes, this would probably be the way to go. But with the way the ensemble is handled in input_cp2k_md, using "input_constants", I am not quite sure how to do that...

PR #1579, opened 2021-07-08
Add the GAL21 forcefield to CP2K, Fix a barostat issue with fixed atoms

pclabaut:

PR #1595, opened 2021-07-20
Barostat npt for frozen atoms in absolute coordinate

### LINKED ISSUE #1591: Barostat issues with frozen atoms
Hello,

Trying some MD runs, I have found that the coordinate scaling induced by NPT ensemble usage does affect the position of frozen atoms, and therefore, inter-atomic distances that should be fixed as well. It does not seem to be the desired behaviour in my opinion. 

I have worked on a fix for my own runs but it seems to provoke numerous failures in the test cases. My question is the following: should the npt_i and npt_f be properly fixed (which would probably imply more changes than what I did, as well as modifying some test cases), should I just implement my current version as a new keyword, or should I simply keep my version for myself ?

In any case, I will be happy to provide my working materials but my development time is limited since my PhD is about to come to an end.

Let me know which path you prefer !

Cheers,

Paul Clabaut

pclabaut:
Sorry, I made a mess with the previous pull request. This one should be cleaner...


pclabaut:
Hello,

Now, the patch is clean and works as intended. It implements only the version of the npt barostat fixing atoms at their absolute coordinates for the isotropic version of the npt barostat. 
I know it could be improved by using an alternative keyword to enable this change of fixation mode and extending it to the npt_f barostat. 
However, I will not have further development time and will therefore leave to you any further improvements.
I hope this will be useful though.

Best regards

src/motion/integrator.F
```diff
@@ -1148,6 +1151,338 @@ SUBROUTINE nvt(md_env, globenv)
...
+! **************************************************************************************************
+!> \brief npt_ia integrator for particle positions & momenta
+!>      isotropic box changes
+!> \param md_env ...
+!> \param globenv ...
+!> \par History
+!>      none
+!> \author CJM
+! **************************************************************************************************
+   SUBROUTINE npt_ia(md_env, globenv)
```
oschuett, line 1163: Maybe instead of duplicating the code you could just add a flag to the `npt_i` routine below?

src/motion/integrator_utils.F
```diff
@@ -1302,6 +1305,223 @@ SUBROUTINE vv_first(tmp, atomic_kind_set, local_particles, particle_set, &
...
+!> \param nparticle_kind ...
+!> \param shell_adiabatic ...
+!> \param dt ...
+!> \param lfixd_list ...
+!> \param u ...
+!> \par History
+!>      none
+!> \author Paul Clabaut (Jully 2021)
+! **************************************************************************************************
+   SUBROUTINE vv_first_frozen(tmp, atomic_kind_set, local_particles, particle_set, &
```
oschuett, line 1326: Maybe instead of duplicating the code you could just add a flag to the `vv_first` routine above?

oschuett (commented):
Thanks for this PR and sorry for the late reply.

juerghutter:
What is the status of this PR? Ready to merge?

oschuett:
My two comments have not yet been addressed. As far as I can tell, this PR unnecessarily duplicates >500 lines of code.

alazzaro:
And since you are there, please fix the error I've reported here: https://github.com/cp2k/cp2k/pull/1579#issuecomment-915921963

ducryf:
Judging from his comment in #1594 ([...] and I am severely running out of time.) I'd guess that pclabaut is not working on this anymore.

oschuett:
Since this PR was almost ready I picked it up as #2000.

PR #1597, opened 2021-07-20
Use COSMA as fallback, if available

alazzaro:


alazzaro:
@oschuett @kabicm
it turns out that COSMA was not used in some of the tests, with a fallback to SCALAPACK (thanks to @gsitaram who have found this problem). This PR fixed the problem, however now we get numerical problems... 
Any suggestion? I can debug...

oschuett:
Yes, we're currently not calling into COSMA when any of the `[a,b,c]_first_[col,row]` arguments are present. AFAIK, this still covers all performance critical use cases. Hence, I didn't see the need to take that risk.


If you always want to call COSMA then you can simply remove [these lines](https://github.com/cp2k/cp2k/blob/8b101df83da9a3e0cee04cfa2f0f558843e15864/src/cp_gemm_interface.F#L95-L100). I don't think we need a second fall back mechanism, the one in [input_cp2k_global.F](https://github.com/cp2k/cp2k/blob/8b101df83da9a3e0cee04cfa2f0f558843e15864/src/input_cp2k_global.F#L629-L633) should be sufficient.

Generally, our test coverage of COSMA is pretty low and only non-zero [because](https://github.com/cp2k/cp2k/blob/8b101df83da9a3e0cee04cfa2f0f558843e15864/tools/docker/scripts/test_regtest.sh#L57)  we set `COSMA_DIM_THRESHOLD=0`.

alazzaro:
@oschuett We have an evidence that COSMA is not used for the LUMI acceptance test with the MP2-RPA benchmark, which is performance critical. Changing the lines you are suggesting is not enough. The problem is in the following IF statements.

alazzaro:
BTW, my changes are to call COSMA to replace pdgemm in all possible scenarios within cp_gemm_interface, which was the original wrapping case....

oschuett:
I see what you mean. Still, I don't like this double fall back. It's just too complicated. I guess, the simplest solution would be to remove all those IF statements together with the DBCSR option. I'm pretty sure COSMA will always be faster anyways.

Alternatively, we could rephrase those IF statements like this:
```
IF (my_multi == do_dbcsr) THEN
  IF (not_supported_by_dbcsr()) my_multi = do_scalapack
END IF
```

alazzaro:
Yes, I think the initial idea was to keep DBCSR as a special case. which it was reasonable in pre-cosma era (I agree with you). I like your idea on the IF statements, however didn't fix the problem. Somehow the problem is on the COSMA side... I see only a test where we use DBCSR_MM as backend (`QS/regtest-rtp-3/H2O_rtp_dbcsr_gemm.inp`). Now I'm removing the DBCSR backend at all...

src/cp_gemm_interface.F
```diff
@@ -81,23 +74,16 @@ SUBROUTINE cp_gemm(transa, transb, m, n, k, alpha, matrix_a, matrix_b, beta, &
...
 
       my_multi = cp_fm_get_mm_type()
 
       ! catch the special case that matrices have different blocking
-      ! SCALAPACK can deal with it but dbcsr doesn't like it
+      ! SCALAPACK/COSMA can deal with it but dbcsr doesn't like it
       CALL cp_fm_get_info(matrix_a, nrow_locals=a_row_loc, ncol_locals=a_col_loc)
       CALL cp_fm_get_info(matrix_b, nrow_locals=b_row_loc, ncol_locals=b_col_loc)
       CALL cp_fm_get_info(matrix_c, nrow_locals=c_row_loc, ncol_locals=c_col_loc)
-      IF (PRESENT(a_first_row)) my_multi = do_scalapack
```
oschuett, line 95: Don't forget to pass all those `[a,b,c]_first_[row,col]` arguments to `cosma_pdgemm()` now.

alazzaro: Yes, working in progress (and good catch) ;)

PR #1605, opened 2021-07-28
Arch file building/using shared libraries only

mkrack:


dev-zero:
@mkrack about the linked FFTW: shouldn't we rather link against `ffw3_omp` instead of `fftw3_threads`? According to the FFTW3 docs they seem to have the same API https://www.fftw.org/fftw3_doc/Usage-of-Multi_002dthreaded-FFTW.html and we probably do not have nested threading calls to FFTW, but still...

mkrack:
@dev-zero Thanks for the hint. I will update the arch files according to the FFTW3 docs.

Issue #1617, opened 2021-08-03
toolchain: fail to compile quip in cp2k8.1: undefined reference to MPI library files


Lucien1234:
Hi all, I am currently attempting to compile cp2k8.1 by toolchain. I used gcc-9.3.1 and intel mkl.
The command I used is:
**_./install_cp2k_toolchain.sh --with-openmpi=install --math-mode=mkl --with-scalapack=no --with-ptscotch=install --with-superlu=install --with-pexsi=install --with-quip=install --with-plumed=install_**
Everything goes well until quip.

The information shown during compiling process is limited：
**_...
==================== Installing QUIP ====================
QUIP-1ff93b3400b83e804f0f2e857e70c5e4133d9705.tar.gz is found
Installing from scratch into /home/bin/cp2k_pool/cp2k-8.1/tools/toolchain/install/quip-1ff93b3400b83e804f0f2e857e70c5e4133d9705
ERROR: (./scripts/install_quip.sh, line 100) Non-zero exit code detected._**

Thus I check the make.log file, finding that the reason is loss of some library files.
**_...
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: /opt/intel/compilers_and_libraries_2020.4.304/linux/mkl/lib/intel64/libmkl_blacs_openmpi_lp64.so: undefined reference to `ompi_mpi_dblprec'
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: /opt/intel/compilers_and_libraries_2020.4.304/linux/mkl/lib/intel64/libmkl_blacs_openmpi_lp64.so: undefined reference to `MPI_Isend'
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: /opt/intel/compilers_and_libraries_2020.4.304/linux/mkl/lib/intel64/libmkl_blacs_openmpi_lp64.so: undefined reference to `MPI_Type_f2c'
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: /opt/intel/compilers_and_libraries_2020.4.304/linux/mkl/lib/intel64/libmkl_blacs_openmpi_lp64.so: undefined reference to `MPI_Op_f2c'
[...]

PR #1621, opened 2021-08-10
Add -lfftw3_mpi conditionally to the arch files

ducryf:
Only link to libfftw3_mpi in the parallel arch files created by the toolchain. Fixes compiling issues with the ssmp and sdbg builds, e.g. here: https://groups.google.com/g/cp2k/c/Z3-D8zw5hJM

### DIFF
--- a/tools/toolchain/scripts/stage2/install_mkl.sh
+++ b/tools/toolchain/scripts/stage2/install_mkl.sh
@@ -82,12 +82,12 @@ if [ "$with_mkl" != "__DONTUSE__" ]; then
 
   case $MPI_MODE in
     intelmpi | mpich)
-      mkl_scalapack_lib="-lmkl_scalapack_lp64"
-      mkl_blacs_lib="-lmkl_blacs_intelmpi_lp64"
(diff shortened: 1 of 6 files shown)


ducryf:
I have also added conditional linking to scalapack and blacs_openmpi_lp64 from MKL. This fixes #1617 and does not link libraries that need mpi in the ssmp and sdbg arch files generated by the toolchain.

Is there a check that compiles the toolchain with MKL? Might be worth running it.

oschuett:
Unfortunately, we don't test with MKL at the moment. We used to via [Dockerfile.cuda_mkl](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/Dockerfile.cuda_mkl), but when HIP arrived I switched to the simpler [Dockerfile.cuda](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/Dockerfile.cuda). My plan was to setup a dedicated `Dockerfile.intel` with the entire Intel stack, which is now available via OneAPI. However, I got stuck somewhere and didn't get back to it yet.

PR #1635, opened 2021-09-08
Bump to LibXC 5.1.6

fstein93:
Can of the admins add the new tarball (http://www.tddft.org/programs/libxc/down.php?file=5.1.6/libxc-5.1.6.tar.gz)?


oschuett:
Done: https://www.cp2k.org/static/downloads/libxc-5.1.6.tar.gz

PR #1640, opened 2021-09-11
Update DBCSR

alazzaro:
1. Move DBCSR Makefile in CP2K
2. Update DBCSR


alazzaro:
@oschuett I'm trying to enable DBCSR HIP support for the Mi100 test. However, the test is failing. Any clue why?

oschuett:
The scratch space is full. Maybe @gsitaram could have a look?

gsitaram:
> The scratch space is full. Maybe @gsitaram could have a look?

Hi Ole,
I have asked the people in charge of the cloud resource to take a look.

gsitaram:
> The scratch space is full. Maybe @gsitaram could have a look?

Ole, they have cleared some space. How much space do you typically use?

oschuett:
Thanks a lot @gsitaram!

The docker image has about 11GB. I'll need that space in `/global/scratch` and `/home` where docker stages images during export. For the moment it's working again, but both devices are pretty close to that limit.

gsitaram:
@oschuett, it seems like this restriction is only on the login node. Is it possible for you to allocate a compute node via slurm and then test there?

alazzaro:
OK, I think I was able to get the toolchain to produce the rights flags to compile the DBCSR HIP backend. I've also updated the INSTALL.md accordingly. However, the CI suddenly crashed, I assume because of the disk space....

oschuett:
> However, the CI suddenly crashed, I assume because of the disk space....

Actually, it timeouts after 20min. For some reason the tests are running much slower.

However, disk space also remains a risk because the home directory is shared across all nodes.

alazzaro:
@oschuett do you have an output I can compare? This is the first time we run HIP with DBCSR, I can expect some problems...

oschuett:
I'll try to take a look later today. Maybe you can also get access to the AMD cluster directly?

alazzaro:
> I'll try to take a look later today. Maybe you can also get access to the AMD cluster directly?

OK, I see a factor 2x (or even more) slowdown in some tests, I will investigate more...

oschuett:
I think, it's the hardware. I've added three more nodes to my growing exclusion list. Maybe there is an underlying cooling issue?

Unfortunately, in the meantime the root filesystem has filled up to a 100%. So, now we can only wait.

hfp:
Would it be feasible to proceed with this PR without HIP tests, i.e., decouple related issues? It would be nice to test CP2K/master and DBCSR/develop. Though, a new RC of DBCSR can help drawing a new baseline for multi-GPU status.

alazzaro:
I've just tagged a [new DBCSR](https://github.com/cp2k/dbcsr/releases/tag/untagged-0e06c872d5a4ab87a912) release :) 
We should keep in the HIP tests, the problem on the CI is unrelated to DBCSR itself.

oschuett:
There was apparently some network issue last night, which caused a lot of dashboard tests to fail. I'm rerunning those tests right now.

Issue #1302, opened 2021-01-12
No DBCSR submodule available


Mert-csrg:
Hello,

Trying to install cp2k on a HPC unit that i have no sudo permissions. Installed libraries and created local arch files via install_cp2k_toolchain.sh.

Then i try to "make" cp2k

> Discovering programs ...
> /truba/home/merts/source/cp2k-7.1.0/exts/Makefile.inc:2: *** "No DBCSR submodule available, please run 'git submodule update --init --recursive'".  Stop.
> make[1]: *** [sopt] Error 2
> make: *** [extversions] Error 2

when i try to use "git submodule update --init --recursive"

> fatal: Not a git repository (or any parent up to mount point /truba)
> Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

i face with this issue.

I tried to remove cp2k/exts/dbcsr file and redownload it from this repository but problem still insists. How can i solve the problem ? I'm not very experienced with installing programs and using linux. I didn't have that issue on my personal computer.

I even changed the default Makefile.inc with the Makefile.inc in this repository. Then i got the 

> Discovering programs ...
> make -C /truba/home/merts/source/cp2k-7.1.0/exts/dbcsr -f .cp2k/Makefile \
>    INCLUDEMAKE=/truba/home/merts/source/cp2k-7.1.0/arch/local.popt \
>    LIBDIR=/truba/home/merts/source/cp2k-7.1.0/lib/local/popt/exts/dbcsr \
>    OBJDIR=/truba/home/merts/source/cp2k-7.1.0/obj/local/popt/exts/dbcsr \
>    ACC="" \
>    ACCFLAGS=""
> Removing stale archives for popt ...
> Resolving dependencies for popt ...
> Removing stale archives ...
> Resolving dependencies ...
> /truba/home/merts/source/cp2k-7.1.0/exts/dbcsr/.cp2k/Makefile:220: *** "No FYPP submodule available, please read README.md on how to properly download DBCSR".  Stop.
> make[3]: *** [libdbcsr] Error 2
> make[2]: *** [dbcsr] Error 2
> make[2]: *** Waiting for unfinished jobs....
> make[1]: *** [popt] Error 2
> make: *** [all] Error 2

dev-zero:
You most likely used the `Source Code` link from the Github releases instead of the versioned tarball (`cp2k-X.Y.tar.bz2`).
[...]

PR #1643, opened 2021-09-13
Specify that versioned tarballs should be used instead of source code.

rashidrafeek:
This is mentioned in the [download page](https://www.cp2k.org/download) but not in `INSTALL.md`. Might be better to mention here also? This is the second time I faced #1302.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -2,7 +2,7 @@
 
 ## 1. Acquire the code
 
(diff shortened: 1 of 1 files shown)


oschuett:
Good idea! Could you please add line breaks to stay below the 80 chars limit.

rashidrafeek:
Done.

dev-zero:
Thanks, merged.

PR #1662, opened 2021-09-30
ocl: support OpenCL devices in DBCSR

hfp:
* Slightly revised CP2K's installation manual and added section about OpenCL Devices.
* Support OFFLOAD_TARGET=opencl and added ARCH-file for GNU Fortran and OpenCL.
* Map GPUVER to a file with tuned parameters (to be embedded into the build).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -162,8 +162,9 @@ the FFTW3 threading library libfftw3_threads (or libfftw3_omp) is required.
 
 - A library for small matrix multiplies can be built from the included source
   (see exts/dbcsr/tools/build_libsmm/README).  Usually only the double precision
-  real and perhaps complex types are needed.  Link to the generated libraries. For a
-  couple of architectures, prebuilt LIBSMMs are available at <https://www.cp2k.org/static/downloads/libsmm/>.
+  real and perhaps complex types are needed.  Link to the generated libraries.
+  For a couple of architectures, prebuilt LIBSMMs are available at
+  <https://www.cp2k.org/static/downloads/libsmm/>.
 - Add `-D__HAS_smm_dnn` to the defines to enable using the double precision
   real library.  Similarly use `-D__HAS_smm_snn` for single precision real and
   `-D__HAS_smm_znn` / `-D__HAS_smm_cnn` for double / single precision complex.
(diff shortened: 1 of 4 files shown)


alazzaro:
LGTM

hfp:
@alazzaro Thank you, Alfio!

Though, there is no test yet. I can probably go ahead and adjust some CUDA based test/image (or whatever). Anyhow, perhaps you or @oschuett want to try/setup a test. Before we rediscover it: Daint-GPUs are in "exclusive mode" which means only the CUDA runtime can put more than one rank onto a device ("exclusive mode" is different from "persistent mode"; just saying since it was discussed at some point and both modes were confused).

oschuett:
Yes, if you prepare a `Dockerfile.opencl` similar to [Dockerfile.cuda](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/Dockerfile.cuda), then I'd be happy to add a dashboard test for this.

Do I assume correctly, that the OpenCL backend also runs on a CPU?

hfp:
> Do I assume correctly, that the OpenCL backend also runs on a CPU?

In principle yes, it's just that an ICD must exist for the CPU. For Intel, the product package called Intel OpenCL SDK for Applications (if not aged/replaced meanwhile) contains such an ICD. However, latest Intel Compute Runtime for Linux does not bring an ICD for the CPU by default. On the other hand, OneAPI brings an ICD for the CPU which seems to be meant for debug support rather than performance/production (at least that's my assumption). I will try/testing this road in my CI setup and install OneAPI for the sake of testing the CPU road.

Speaking more of Intel, GPUs just work fine including laptop/integrated graphics. I also tested under macOS using a Mac Pro with an AMD Vega56, which worked fine too at least using my stand-alone reproducer in DBCSR (though SP only with this hardware/software combination; macOS CPU ICD is somewhat broken btw). I had no chance to try any other AMD GPU. Speaking of Nvidia GPUs - they all work fine (I tested P100, V100, and A100). In fact, I implemented inline-PTX to get FP-atomic-add into the kernel (unlike with CUDA, FP-atomics do not seems to be available otherwise with OpenCL on Nvidia).

oschuett:
I'm a bit hesitant to add another GPU test because they are kinda costly. So, if you can get this to run on CPU that would be great. Another benefit of using OneAPI would be that we also get to test its other components, ie. MKL, Intel MPI + Compilers, etc.

hfp:
Let me merge since this way curious people can already try the code. I hope to send the followup PR this week (Docker based tests).

> GPU test because they are kinda costly

ACK.

PR #1665, opened 2021-10-05
Add a key to print the atom name instead of the element symbol

ducryf:


ducryf:
This pr adds a keyword (PRINT_ATOM_NAME) to TRAJECTORY, FORCES, and VELOCITY in MOTION%PRINT. This will cause the atom name to be written to the xyz file instead of the element symbol.

ducryf:
I changed the keyword to PRINT_ATOM_KIND and improved its description. If it is activated the labels in the output file match the ones from the input, i.e. the label from &KIND is printed. With this switch different kinds of the same element can directly be distinguished in the trajectory files.

mkrack:
Does that key work in general or only with QMMM?
`A4` is a quite short format to fit generic atomic kind names. Afaik, the XMOL format has no column restrictions and requires just spaces between the columns `atom_name, x, y, z`. That would allow dumping also longer names without string truncations.

ducryf:
This works in general. I simply chose the QMMM regtest because it already uses kind names that differ from the element.
I have removed the restriction of 4 characters for the kind name.

mkrack:
Thanks for changing the format.
It seems that the variable `dummy` (just the line above your format definition) could be removed everywhere in that module, right?

mkrack:
The `LEN=4` for the variable `name` should then also be increased to allow for longer atom kind names.

ducryf:
The function qmmm_ff_precond_only_qm potentially modifies atm_name. So I'm not sure that this function call should be omitted. 

I kept the `LEN=4` for `name`  because I did not want to change anything with the pdb format. `name` is not used for the xmol format anymore, so there should not be a limitation.

mkrack:
You are right.

ducryf:
I have nothing more to add to this pr. It can be merged if you have no further requests.

PR #1627, opened 2021-08-19
Update sirius version and sirius input files parameters

mtaillefumier:
- SIRIUS is updated to the latest version : it fix some input parameters issues that were reported to us.
- Adjust the regtests because SIRIUS uses a different stopping convention during the SCF loop (the old one is still present and can be activated)

PR #1674, opened 2021-10-13
Fix initialization of string variables

mtaillefumier:
- string variable are C initialized which make FORTRAN unhappy
- explicitly set pseudo_potential or full_potential

### DIFF
--- a/src/input_cp2k_pwdft.F
+++ b/src/input_cp2k_pwdft.F
@@ -148,7 +148,7 @@ CONTAINS
       CHARACTER(len=*), INTENT(in)                       :: section_name
 
       CHARACTER(len=128)                                 :: name, name1, possible_values(1:16)
-      CHARACTER(len=512)                                 :: default_string_val, description, usage
+      CHARACTER(len=4096)                                :: default_string_val, description, usage
       INTEGER                                            :: ctype, dummy_i, enum_i_val(1:16), i, j, &
                                                             length, num_possible_values, vec_length
       INTEGER, ALLOCATABLE, DIMENSION(:)                 :: ivec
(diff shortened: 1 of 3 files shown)


oschuett:
Thanks a lot for following up on #1627!

Unfortunately, there seems to still be an issue. Have look at line 403064 in `cp2k_input.xml` from the [artifacts](https://storage.googleapis.com/cp2k-ci/run-cp2k-manual-981a2723_artifacts.tgz):
```
    <KEYWORD repeats="no" removed="no">
     <NAME type="default">NUM_BANDS_TO_PRINT</NAME>
     <DATA_TYPE kind="integer">
      <N_VAR>1</N_VAR>
     </DATA_TYPE>
     <USAGE></USAGE>
     <DESCRIPTION>d
</DESCRIPTION>
```

mtaillefumier:
No worries. It seems the problem might be linked to strings that are not long enough. So I changed the length to something like 4096 for 3 of the variables. I checked the generated xml files locally and the characters are not present anymore.
I might also need to update sirius so that the string lengths are passed properly.

mkrack:
@mtaillefumier When you are going to update SIRIUS anyway, please, could you also add a flag to sirius_init which allows keeping SIRIUS silent, i.e. no write to stdout, to get rid of the lines
```
SIRIUS 7.0.2, git hash: https://api.github.com/repos/electronic-structure/SIRIUS/git/ref/tags/v7.0.2
Warning! Compiled in 'debug' mode with assert statements enabled!
```
which are always printed, even if SIRIUS is not used.

mtaillefumier:
The only option is to only call **sirius_init** when needed, not all the time. These lines are there a for good reason, indicate which version of SIRIUS is used. Suppressing this message would probably imply breaking the api which I am not keen to do without good reason. 
If you absolutely want to get these two lines out, then open an issue on the SIRIUS github page.

mkrack:
@mtaillefumier You are right, it is probably better to handle this on the CP2K side (#1680).

PR #1694, opened 2021-10-25
Adjust Tolerance and fix regtest

fbelle:


mkrack:
It seems that the tolerance of just one test needs further adjustment.

fbelle:
Indeed, my initial adjustment was insufficient.

mkrack:
Thanks, now, it looks fine.

PR #1698, opened 2021-10-26
Bump LibXC to 5.1.7

fstein93:
Can any of the admins add the new tarball, please?


mkrack:
[libxc 5.1.7](https://www.cp2k.org/static/downloads/) should be available now.

fstein93:
Thank you!

PR #1703, opened 2021-10-27
RI-HFX| fixing regtests

abussy:
Previous PR on the subject broke the dashboard. This should fix it.


oschuett:
That failure of the i386 test is due to my #1702. I'll take care of it.

PR #1706, opened 2021-10-28
Analytical derivatives of the MO coefficients wrt nuclear coordinates and the evaluation of Atomic Polar Tensors

edditler:
Dear CP2K developers,

We used the code in this pull request for the calculations in this publication: [doi.org/10.1063/5.0041056](doi.org/10.1063/5.0041056).
We added another response calculation called `dcdr_linres` to `qs_linres_module`. The purpose is the analytical evaluation of the derivatives of the MO coefficients with respect to nuclear coordinates. The routine `dcdr_response_dR` calculates `$dCR = dC^0/dR^lambda_beta$`, where `lambda` enumerates the atom indices and `beta` the cartesian coordinates. The coefficient derivatives are then used to calculate the Atomic Polar Tensors (APTs) `$P^lambda_alpha,beta = d<mu_alpha>/dR^lambda_beta$`, where `mu_alpha` are the components of the electric dipole operator. The routine `apt_dR_localization` calculates the APTs using the localized dipole operator and localized orbitals.

This pull request touches quite a few files in the core of CP2K to evaluate integral derivatives with respect to nuclear coordinates - instead of electronic coordinates. I hope that I didn't incur significant performance issues in the standard ground-state calculations.

This input: https://gist.github.com/edditler/97d8a3d2b8f23eb2bb0bcbee3bacfcfc results in this output: https://gist.github.com/edditler/0b517b67bd01346f5c6ad7674c5dc21c on my machine, using the sdbg, ssmp, and psmp binaries.

I ran the regression tests using the arch files generated by the toolchain with the command `tools/regtesting/do_regtest -arch local -version psmp -maxtasks 10 -mpiranks 2 -ompthreads 2`

One can also use the implementation to evaluate the analytical Hessian matrix and improve geometry optimizations, but we didn't consider the needed interface.

Best regards,

Edward


src/grid/grid_api.F
```diff
@@ -400,12 +412,19 @@ END SUBROUTINE grid_ref_integrate_pgf_product_c
       ENDIF
 
       IF (calculate_forces .AND. PRESENT(hdab)) THEN
-         hdab(:, :, :) = 0.0_dp
+         ! hdab(:, :, :) = 0.0_dp
```
oschuett: Why do you not want to zero `hdab` anymore (same for `hadb` below)?

src/grid/ref/grid_ref_integrate.c
```diff
@@ -105,6 +106,13 @@ void grid_ref_integrate_pgf_product(
                       get_force_a(a, b, i, zeta, zetb, m1, cab, false);
                 }
               }
+              if (hadb != NULL) {
```
oschuett, line 109: Please also add `hadb` to the comment  in line 101.

oschuett (commented):
Thanks for contributing Edward!

The changes to the grid code look good to me. 

Please consider adding a few regtests.

src/grid/grid_api.F
edditler: I think the reason is that the line of code would overwrite the results before they are accumulated in `vhxc_block` here: https://github.com/cp2k/cp2k/blob/f1afc0d91c75b6fa9467fda469365eda1d4dfd1f/src/qs_integrate_potential_product.F#L502-L505

The only other call with `hdab` that I can find is here https://github.com/cp2k/cp2k/blob/master/src/qs_rho0_ggrid.F#L558-L566, where it's superfluous to zero the array again.

edditler:
Regarding the regression tests: Do I understand right that I need to move the output from a separate file https://github.com/cp2k/cp2k/blob/c623102701aa8ab42a710cb570884d21d85820ca/src/qs_dcdr_utils.F#L707-L709 into the main CP2K output to be able to compare the result of the sum rules/APTs in the regtests?

mkrack:
Yes. You need only to print a key result in the CP2K output file which you want to check, preferentially tagged e.g. with `APT| `.

src/grid/grid_api.F
oschuett: I believe for `forces` and `virials` some callers actually rely on the zeroing. While I would prefer the API to be more consistent, I noticed that we already don't zero `a_hdab`. So, I guess it's ok to also not do it for `hdab` and `hadb`. 

If you decide to keep it this way then please remove those lines entirely instead of just commenting them out.

oschuett:
Since you are already using a printkey, it should be sufficient to set it to `__STD_OUT__` in the input file.

src/grid/grid_api.F
edditler: I removed both comments and added some regtests for our implementation. The other regtests also pass without the zeroing in `grid_api`. That's why I assumed that it's probably not necessary.

oschuett:
Unfortunately, there is still a [merge conflict](https://github.com/cp2k/cp2k/pull/1706/conflicts) because our formatting changed in the meantime (`ENDIF` -> `END IF`). I'd suggest you run `make pretty` again and then squash all your changes into a single commit.

edditler:
> Unfortunately, there is still a [merge conflict](https://github.com/cp2k/cp2k/pull/1706/conflicts) because our formatting changed in the meantime (`ENDIF` -> `END IF`). I'd suggest you run `make pretty` again and then squash all your changes into a single commit.

This morning, I ran `make pretty` and observed that the number of merge conflicts increased. The force push reversed that commit. I wanted to avoid resolving merge conflicts too often. Is this PR ready for merging otherwise? I'll get on it first thing tomorrow.

src/qs_dcdr_utils.F
```diff
@@ -0,0 +1,1079 @@
...
+      CALL dbcsr_deallocate_matrix_set(dcdr_env%matrix_t)
+      CALL dbcsr_deallocate_matrix_set(dcdr_env%matrix_s1)
+      CALL dbcsr_deallocate_matrix_set(dcdr_env%matrix_t1)
+      CALL dbcsr_deallocate_matrix_set(dcdr_env%matrix_apply_op_constant)
+      CALL dbcsr_deallocate_matrix_set(dcdr_env%matrix_d_vhxc_dR)
+      CALL dbcsr_deallocate_matrix_set(dcdr_env%matrix_vhxc_perturbed_basis)
+      CALL dbcsr_deallocate_matrix_set(dcdr_env%matrix_hc)
+      CALL dbcsr_deallocate_matrix_set(dcdr_env%matrix_ppnl_1)
+
+      CLOSE (11111)
```
oschuett: Left over?

src/qs_dcdr_utils.F
```diff
@@ -0,0 +1,1079 @@
...
+      CALL cp_print_key_finished_output(output_unit, logger, lr_section, &
+                                        "PRINT%PROGRAM_RUN_INFO")
+
+      IF (dcdr_env%output_unit > 0) &
+         WRITE (dcdr_env%output_unit, *) 'List of atoms: ', dcdr_env%list_of_atoms
+
+      CALL timestop(handle)
+   END SUBROUTINE dcdr_env_init
+
+! *****************************************************************************
```
oschuett: Lopsided box. There are multiple occurrences, just run grep.

oschuett (commented):
I found two nits, otherwise I think this is ready to be merged.

src/qs_dcdr_utils.F
edditler: Missed that

src/qs_dcdr_utils.F
edditler: Fixed

oschuett:
I'm going to squash merge this, but I feel that _"Response for dC/dR_lambda^beta"_ might not be a very useful commit message.
How about _"Add analytical derivatives of MO coefficients wrt nuclear coordinates"_ ?

edditler:
Seems good! Can you change the commit message or should I push once more?

oschuett:
It's done. Thanks for upstreaming your changes!

oschuett:
The [dashboard](https://dashboard.cp2k.org) found a couple of [convention issues](https://dashboard.cp2k.org/archive/conventions/commit_9cea0ffd4a856bab1aac22ccbb9cd5115206d1f1.txt) and what appears to be a [double free](https://dashboard.cp2k.org/archive/gcc7/commit_9cea0ffd4a856bab1aac22ccbb9cd5115206d1f1.txt).

Instead of `cp_fm_gemm` please use `cp_gemm` from [cp_gemm_interface.F](https://github.com/cp2k/cp2k/blob/master/src/cp_gemm_interface.F).

edditler:
Should I create a new PR for the fixes?

oschuett:
Yes, it would be great if you could fix these issue with a follow up PR.

mkrack:
There are also issues with the tests using the Intel compiler.

edditler:
> There are also issues with the tests using the Intel compiler.

The assertion fails in https://github.com/cp2k/cp2k/blob/master/src/qs_dcdr_ao.F#L769-L771. I'm not sure what the fix is. Maybe replacing by `CPASSERT(ASSOCIATED(k_block))`? Code identical to these three lines seems to be used at other places in the codebase, or will the issue be somewhere else?

Two more failing tests, where I don't understand what's going wrong, are

- https://storage.googleapis.com/cp2k-ci/dashboard_i386_report.txt
- https://storage.googleapis.com/cp2k-ci/dashboard_minimal_report.txt

mkrack:
The i386 test was failing already before your PR. The minimal, GCC7, and the Intel tests might have failed for the same reason, which is possibly fixed by your latest fix. We will see.

PR #1716, opened 2021-10-29
gpu target for elpa and cray math-mode for cosma

ducryf:
Add the option to compile cosma with the cray libsci backend when math-mode=cray is selected.

Set the compute capability when compiling elpa.


ducryf:
This pr only changes the tool chain which compiled successfully. Apparently an error occurred when fetching the code to compile cp2k:
========== Fetching Git Commit ==========
fatal: reference is not a tree: 2d82637715b7c25df1a364ae80de6c7fd872aa05

How can I deal with this issue?

PR #1721, opened 2021-11-02
Change from accurate sum to standard sum in pw_integral_ab

holly-t:
Hello,

Putting in this PR to see the effect on the CI tests from changing the accurate sum to a standard sum in pw_integrate_ab. I have run the regression tests and there appears to be a few wrong results as a result of the change so this may not be able to merge. Or we could consider doing a non-accurate sum version of pw_integral_ab when called from the qmmm_gpw routines.

Thanks,
Holly

### DIFF
--- a/src/pw/pw_methods.F
+++ b/src/pw/pw_methods.F
@@ -31,8 +31,7 @@ MODULE pw_methods
    USE fft_tools,                       ONLY: BWFFT,&
                                               FWFFT,&
                                               fft3d
-   USE kahan_sum,                       ONLY: accurate_dot_product,&
-                                              accurate_sum
+   USE kahan_sum,                       ONLY: accurate_sum
    USE kinds,                           ONLY: dp
    USE machine,                         ONLY: m_memory
    USE message_passing,                 ONLY: mp_sum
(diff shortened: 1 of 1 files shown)


fstein93:
Two comments:
1) In order to merge it, you should check whether you obtain the same results for larger systems where the effect of the accurate summation should be more pronounced, not just for the regtests.
2) With respect to performance, you should exploit the Fortran intrinsic DOT_PRODUCT function instead.

juerghutter:
An optimal solution for me would be to add an option with
- accurate sum (default)
- normal sum
- dynamic, depending on global number of entries

hfp:
> depending on global number of entries

If any of the sums is in an OpenMP reduction, "dynamic" can also consider the number of threads. Funny enough, parallelized reductions may be more accurate even if they have other problems like non-deterministic order. I think some OpenMP implementations allow deterministic reductions (e.g., `KMP_DETERMINISTIC_REDUCTION=TRUE` with Intel OpenMP).

holly-t:
Happy to make changes to use the DOT_PRODUCT where it is possible and check the results for larger systems.

@juerghutter would this be an input option? How should this be passed down to pw_integral_ab?

juerghutter:
This would be a call option to pw_integral_ab. The programmer can then choose to use accurate/standard/dynamic. I wouldn't go as far as make it a user option.

oschuett:
Superseded by #1779.

PR #1729, opened 2021-11-05
Doc Update for DFT Poisson Solvers

patrickmelix:
Adds descriptions to `%FORCE_EVAL%DFT%POISSON_SOLVER` keywords. Mainly stating the supported periodicity of the keywords. Found these through looking at the code and trial-error runs of the latest code. Therefore, please, someone who knows this part of the code better *double check the info* added. The respective page in the Wiki is extremely outdated.

Moved the description of the MT keyword from the subsection header to the keyword itself because it is here that the user would expect it in my opinion.

### DIFF
--- a/src/input_cp2k_poisson.F
+++ b/src/input_cp2k_poisson.F
@@ -542,7 +542,7 @@ CONTAINS
       CPASSERT(.NOT. ASSOCIATED(section))
       CALL section_create( &
          section, __LOCATION__, name="wavelet", &
-         description="Sets up parameters of  wavelet based poisson solver.", &
+         description="Sets up parameters of wavelet based poisson solver.", &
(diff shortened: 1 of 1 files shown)


oschuett:
Please run `make pretty` to fix the formatting.

patrickmelix:
Sorry for these newbie problems... Is there an overview of all these little helpers? Doing it now...

oschuett:
> Is there an overview of all these little helpers?

There is https://www.cp2k.org/dev:starting, but generally our documentation could probably use some more curation.
We try to make up for it by responding quickly to PRs :-)

patrickmelix:
Ok, so learning by doing :+1:  Working on the documentation as you might have noticed :wink: 
Indeed, never seen such quick replies anywhere else! Thanks for the effort and your support.

PR #1735, opened 2021-11-09
Upgrade to fpretty 0.3.7

oschuett:
This is a pretty large (whitespace) change because fpretty's default style has changed, e.g. `ENDIF` -> `END IF`.

While we could change the settings I'd prefer to stick with the defaults. Any other opinions?


mkrack:
I appreciate the new defaults and vote for `END IF` and `END DO`.

PR #1745, opened 2021-11-18
Fixes in dcdr

edditler:
Follow up PR to https://github.com/cp2k/cp2k/pull/1706

- Remove a `STAT=` and a `print` statement;
- replace calls to `cp_fm_gemm` by `cp_gemm`
- nullify DBCSR matrix before initializing


edditler:
I hope I wasn't too quick. My understanding is that `ASSOCIATED(matrix)` is undefined when `matrix` wasn't nullified before and can return true, deallocating a non-existent matrix.

PR #1753, opened 2021-11-23
python: change shebang to /usr/bin/env python3

ducryf:
Most python files use /usr/bin/env python3. A few however point to /usr/bin/python3.
This pr changes the few /usr/bin/python3 to /usr/bin/env python3.

### DIFF
--- a/tools/minimax_tools/minimax_to_fortran_source.py
+++ b/tools/minimax_tools/minimax_to_fortran_source.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 
 import re
 import argparse
--- a/tools/precommit/precommit.py
+++ b/tools/precommit/precommit.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 
 # author: Ole Schuett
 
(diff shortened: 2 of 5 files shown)


oschuett:
Out of curiosity, did you actually encounter a system where `/usr/bin/env` existed but `/usr/bin/python3` did not?

To prevent  backsliding we could add this to [check_file_properties.py](https://github.com/cp2k/cp2k/blob/master/tools/precommit/check_file_properties.py).

dev-zero:
In any virtualenv or a setup with pyenv, conda, etc. you want Python scripts to use python3 from $PATH rather than the hardcoded system one.

ducryf:
Yes, that's it. I'm using a conda env to compile cp2k. Because of that I could not run make test without changing the shebang.

hfp:
> Out of curiosity, did you actually encounter a system where `/usr/bin/env` existed but `/usr/bin/python3` did not?
> 
> To prevent backsliding we could add this to [check_file_properties.py](https://github.com/cp2k/cp2k/blob/master/tools/precommit/check_file_properties.py).

Any stock distribution may not break your assertion. However, due to the lengthy and somewhat painful Python transition, I found systems (aka custom) with quite some unexpected "solutions", e.g., Shell alias introducing "python3".

PR #1769, opened 2021-12-03
LIBXSMM 1.17

hfp:
* Removed build options MALLOC=0 (new default), and INTRINSICS=1 (no necessity to control).

### DIFF
--- a/tools/toolchain/scripts/stage4/install_libxsmm.sh
+++ b/tools/toolchain/scripts/stage4/install_libxsmm.sh
@@ -9,8 +9,8 @@
 [ "${BASH_SOURCE[0]}" ] && SCRIPT_NAME="${BASH_SOURCE[0]}" || SCRIPT_NAME=$0
 SCRIPT_DIR="$(cd "$(dirname "$SCRIPT_NAME")/.." && pwd -P)"
 
-libxsmm_ver="1.16.2"
-libxsmm_sha256="bdc7554b56b9e0a380fc9c7b4f4394b41be863344858bc633bc9c25835c4c64e"
+libxsmm_ver="1.17"
+libxsmm_sha256="8b642127880e92e8a75400125307724635ecdf4020ca4481e5efe7640451bb92"
 source "${SCRIPT_DIR}"/common_vars.sh
 source "${SCRIPT_DIR}"/tool_kit.sh
 source "${SCRIPT_DIR}"/signal_trap.sh
@@ -66,16 +66,12 @@ EOF
       # library
       cd libxsmm-${libxsmm_ver}
       make -j $(get_nprocs) \
-        MALLOC=0 \
-        INTRINSICS=1 \
         CXX=$CXX \
         CC=$CC \
         FC=$FC \
         PREFIX=${pkg_install_dir} \
         > make.log 2>&1
       make -j $(get_nprocs) \
-        MALLOC=0 \
-        INTRINSICS=1 \
         CXX=$CXX \
         CC=$CC \
         FC=$FC \


hfp:
The new release may [downloaded](https://github.com/hfp/libxsmm/archive/refs/tags/1.17.tar.gz) and copied to https://www.cp2k.org/static/downloads/.

mkrack:
Done.

mkrack:
@hfp The build of libxsmm 1.17 works, but I get the following traceback (does not occur with 1.16.2)
```
================================================================================
LIBXSMM release-1.17 (Linux@merlin-l-001.psi.ch)
--------------------------------------------------------------------------------
GNU Compiler Collection: gcc 10.3.0, g++ 10.3.0, and gfortran 10.3.0
C / C++ target: -mfma -mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl
Fortran target: -mfma -mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl
--------------------------------------------------------------------------------
Environment: CC CXX FC FORTRAN
--------------------------------------------------------------------------------
--- LIBXSMM build log
/bin/sh: 0+(/opt/psi/Programming/gcc/10.3.0/bin/gfortran<<2): syntax error: operand expected (error token is "/opt/psi/Programming/gcc/10.3.0/bin/gfortran<<2)")
/bin/sh: 0+(/opt/psi/Programming/gcc/10.3.0/bin/gfortran<<2): syntax error: operand expected (error token is "/opt/psi/Programming/gcc/10.3.0/bin/gfortran<<2)")
  File "./scripts/libxsmm_interface.py", line 81, in <module>
    raise ValueError("Fortran interface level is inconsistent!")
ValueError: Fortran interface level is inconsistent!
```
Do I have to worry about it?

hfp:
> Done.

Thank you!

( Btw, LIBXSMM/master should work with ARM64 targets now. I tried in conjunction with CP2K already, but I did not run the full set of regtests yet. I am saying this just because you maintain the Dashboard entries targeting ARM, and may want to try. Our master/main branch is meant to be stable, but you can also pick a specific revision. )

hfp:
> @hfp The build of libxsmm 1.17 works, but I get the following traceback (does not occur with 1.16.2)
> 
> ```
> ================================================================================
> LIBXSMM release-1.17 (Linux@merlin-l-001.psi.ch)
> --------------------------------------------------------------------------------
> GNU Compiler Collection: gcc 10.3.0, g++ 10.3.0, and gfortran 10.3.0
> C / C++ target: -mfma -mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl
> Fortran target: -mfma -mavx512f -mavx512cd -mavx512dq -mavx512bw -mavx512vl
> --------------------------------------------------------------------------------
> Environment: CC CXX FC FORTRAN
> --------------------------------------------------------------------------------
> --- LIBXSMM build log
> /bin/sh: 0+(/opt/psi/Programming/gcc/10.3.0/bin/gfortran<<2): syntax error: operand expected (error token is "/opt/psi/Programming/gcc/10.3.0/bin/gfortran<<2)")
> /bin/sh: 0+(/opt/psi/Programming/gcc/10.3.0/bin/gfortran<<2): syntax error: operand expected (error token is "/opt/psi/Programming/gcc/10.3.0/bin/gfortran<<2)")
> Traceback (most recent call last):
>   File "./scripts/libxsmm_interface.py", line 81, in <module>
>     raise ValueError("Fortran interface level is inconsistent!")
> ValueError: Fortran interface level is inconsistent!
> ```
> 
> Do I have to worry about it?

Hmm, I fully validated and ran regtests with the new release, but as always environments are different - I will investigate. This is not expected nor wanted just to say it...

mkrack:
> 

Thanks, I will give a try with ARM64.

hfp:
> [...] Do I have to worry about it?

The issue stems from an environment setting or from the Makefile (ARCH-file) using a variable called `FORTRAN`. LIBXSMM's build system uses such a variable to optionally disable support for Fortran (e.g., no Fortran interface would be generated and no libxsmmf library is created). Anyway, the problem is the variable's content is treated as number whereas your `FORTRAN` contains the path to the compiler. I can fix this for master but prefer to keep v1.17 as is.

mkrack:
Yes, when I load the module for the GNU compiler, the variable `FORTRAN` is set to the full path of `gfortran`. Why do you check just for a numerical content of that variable? What's about just checking if the that variable is defined or not in the environment?

hfp:
> Yes, when I load the module for the GNU compiler, the variable `FORTRAN` is set to the full path of `gfortran`. Why do you check just for a numerical content of that variable? What's about just checking if the that variable is defined or not in the environment?

I suggest to use `FC` for your case, which is one of the standard/basic variables referring to the plain Fortran compiler (it's a bit outgrown with Fortran but most common are `F77`, `F90`, or a the neutral `FC` plus `MPI` prefixes). I have not seen `FORTRAN` used in Makefiles (or other build systems), hence I used it to control something else (in this case it does not refer to the compiler but an "interface level").

hfp:
> Why do you check just for a numerical content of that variable? What's about just checking if the that variable is defined or not in the environment?

I can certainly do this going forward, but I think it's too late for v1.17.

mkrack:
I do not set `FORTRAN`. It is done by the system/admin side during the `module load gcc`. I will (re)set the variable `FORTRAN` to `1` before running the build script.

hfp:
> I will (re)set the variable FORTRAN to 1 before running the build script.

Perhaps use `FORTRAN=2` or execute `unset FORTRAN`.

PR #1777, opened 2021-12-06
EC| Add print section for EC properties, and Voronoi dipoles

fbelle:


src/energy_corrections.F
```diff
@@ -2122,13 +2170,88 @@ SUBROUTINE ec_properties(qs_env, ec_env)
...
+
+            ! add Pout and Pz
+            CALL pw_copy(ec_env%rhoout_r(1)%pw, rho_elec_rspace%pw)
+            CALL pw_axpy(ec_env%rhoout_r(2)%pw, rho_elec_rspace%pw)
+
+            CALL pw_axpy(ec_env%rhoz_r(1)%pw, rho_elec_rspace%pw)
+            CALL pw_axpy(ec_env%rhoz_r(2)%pw, rho_elec_rspace%pw)
+
+            mb_rho => rho_elec_rspace%pw
+            !CALL voronoi_analysis( qs_env, rho_elec_rspace%pw, print_key, unit_nr )
```
oschuett: Why are these calls to `voronoi_analysis` commented out?

fbelle: The original code in qs_scf_post_gpw.F was refactured at some point. I kept the artifact around for no good reason.

PR #1784, opened 2021-12-08
New version of SIRIUS

mtaillefumier:
It may seems not a lot, but we redesigned the interface to better handle error and runtime options. Now each function can return and error code (I did not use it in cp2k since the error code is optional). Hopefully this will fix some of the issues that several people reported to us (thanks).
- Also fixed few warnings that popped up here and there
- next thing will be Hubbard U + V


dev-zero:
@mtaillefumier thanks a lot!

PR #1798, opened 2021-12-10
Fix toolchain for cuda+gcc11 and fix build without libint

oschuett:


dev-zero:
@oschuett <s>are you sure about those changes to the libint wrapper? I just put them in to fix building without libint. See https://github.com/cp2k/cp2k/commit/cbcfb96792964a6fdb47be18e84db1ab4314cba7#diff-fe2db683caafd6ce4f4236149923334b2f2437d0f3889cb2300acfe5764e9b38</s>
... ok, I see, the var declaration was moved somewhere in between the version I've been working with and the latest version.

PR #1829, opened 2021-12-23
Update DBCSR (v2.2)

alazzaro:


hfp:
@alazzaro what's the timeline for v2.2 or the RC?   I still have https://github.com/cp2k/dbcsr/pull/534 running...

alazzaro:
> @alazzaro what's the timeline for v2.2 or the RC? I still have [cp2k/dbcsr#534](https://github.com/cp2k/dbcsr/pull/534) running...

At this point there are several issues on the table, so I went for the conservative solution to make a release up to commit https://github.com/cp2k/dbcsr/commit/0271399eda4fa6694ff2774a2aa4e1afc8a5b16c.

Please note that I have explicitly marked "Experimental" HIP and OpenCL in the Install.md.

We agreed with @oschuett that there will be time for a new CP2K release sometime in 2022 (May-June).

oschuett:
Yes, once LUMI and Leonardo become available we can focus on fixing any multi-GPU and HIP issues.

hfp:
@alazzaro Thank you!

> I went for the conservative solution to make a release up to commit cp2k/dbcsr@0271399.

Makes sense. This is the last release without this (late) change about synchronization. I agree, this needs to be tested more and we can potentially reintroduce the established method as compile-time toggle. Indeed, the new synchronization was fixing issues as a side-effect (potential data races).

> Please note that I have explicitly marked "Experimental" HIP and OpenCL in the Install.md.

ACK, makes sense.

> We agreed with @oschuett that there will be time for a new CP2K release sometime in 2022 (May-June).

Perfect!

PR #1851, opened 2022-01-01
HFX: Do not rely on block order of tensor iterator

oschuett:


oschuett:
@JWilhelm and @abussy FYI: The new sparse matrix library will no longer iterate over the blocks in sorted order.

JWilhelm:
> @JWilhelm and @abussy FYI: The new sparse matrix library will no longer iterate over the blocks in sorted order.

Thanks for letting us know, Ole. Do you have a feeling whether this change will have an impact on performance or memory consumption?

oschuett:
This PR should not affect memory or performance, because the HFX code already kept a copy of the indices.

And so far I'm not aware of any other places that depend on the iteration order.

Dropping the block sorting will allow the new matrix library to have better OpenMP performance.

JWilhelm:
> Dropping the block sorting will allow the new matrix library to have better OpenMP performance.

Sounds good!

abussy:
> And so far I'm not aware of any other places that depend on the iteration order.

Neither do I. To be honest, I did not even realize the iterator followed a specific block order.

PR #1854, opened 2022-01-02
Enable toolchain build for a generic target

mkrack:
- Merge GNU arch files for generic and static serial binaries
- Make arch file sourceable
- Sync compiler flags with toolchain


alazzaro:
@mkrack This PR is very very clever. I see you did for the ARM system (?)
I had a similar problem when building a container for a cross-platform (effectively a cross-compile).
In the past I tried to make a toolchain version within the container to be portable, but I found several places where we take current system flags to compile. For instance, in FFTW we do the following:

https://github.com/cp2k/cp2k/blob/83d825b20d1a102d9e3a9b3f3a47c5ac9eb2f74c/tools/toolchain/scripts/stage3/install_fftw.sh#L54-L56

This is something we should not use, suppose I want to build a container on my laptop (AVX2) to run on a different system (AVX512)... I will try to extend your "generic" flag idea.

mkrack:
> 

Indeed, a (better) cross-compilation support of the CP2K toolchain is missing. The `--generic` flag could be extended to a `--target=<platform>` flag including for instance `--target=x86-64-generic` as generalization.

Issue #1830, opened 2021-12-24
Memory leaks with Intel MPI


gsamsonidze:
There are memory leaks in Intel MPI 2019.x and 2021.3 which cause OOM errors in long MD simulations with CP2K. It was fixed in Intel MPI 2021.4, see this ticket on Intel forums:

https://community.intel.com/t5/Intel-oneAPI-HPC-Toolkit/Memory-leak-in-D-amp-C-eigensolver-parallel/m-p/1337226#M32303

You may want to include this information on [Compiler support](https://www.cp2k.org/dev:compiler_support) page.

dev-zero:
@gsamsonidze thanks a lot for this!
The thread also mentions you seeing freezes with 2021.4. The last known "good" version is 2018.4 then. Since last time I tried it CP2K does not build with the LLVM-based Intel Fortran Compiler I assume this is the classic tools.
Wondering whether the reported hang involved ELPA, could you check?

gsamsonidze:
Yes the last stable version is 2018.4. The oneAPI builds are the classic tools (ifort). The reported hang involved ELPA, I will rerun with SCALAPACK to check if it solves the issue.

gsamsonidze:
| Intel  | Diag      | Status                                                |
|--------|-----------|-------------------------------------------------------|
| 2018.4 | ELPA      | Runs fine for 5000 MD steps, no errors                |
| 2021.4 | ELPA      | Freezes after ~300 MD steps                           |
| 2018.4 | SCALAPACK | Signal 11 (segmentation violation) after ~50 MD steps |
| 2021.4 | SCALAPACK | Assertion failed [1] during the 1st MD step (2nd SCF) |

[1] Here is the full error message: `Assertion failed in file ../../src/mpid/ch4/src/intel/ch4_shm_coll.c at line 2266: comm->shm_numa_layout[my_numa_node].base_addr`
A Google search gives the following result: https://community.intel.com/t5/Intel-oneAPI-HPC-Toolkit/Assertion-failed-in-ch4-shm-coll-c-at-line-2147/m-p/1186157

The system size is ~600 atoms (organometallic molecule solvated in liquid water at the metal interface), PBE+D3 xc-functional, GTH pseudopotentials, DZVP-/TZVP-MOLOPT-SR basis sets. The version of CP2K is 9.1.

boegel:
[...]

Issue #1788, opened 2021-12-09
SIRIUS: elpa2 solver gets stuck


dev-zero:
Detailed configuration is attached as [Br-elpa-stuck.zip](https://github.com/cp2k/cp2k/files/7682787/Br-elpa-stuck.zip), but essentially when setting:

```
         GEN_EVP_SOLVER_NAME elpa2
         STD_EVP_SOLVER_NAME elpa2
         MPI_GRID_DIMS 2 2
```

the calculation gets reproducibly stuck at iteration 4, with the main processes spinning at 100%.

This is with CP2K 9f7a817, SIRIUS v7.2.8 and ELPA 2021.05.002_bugfix.

Also notable is that `MPI_GRID_DIMS` gets ignored completely.

oschuett:
This was likely fixed by the upgrade to ELPA 2021.11.002. See also #1956.

PR #1857, opened 2022-01-05
Update Cray arch files

mkrack:
- Sync target names for ELPA with toolchain
- Adapt path names for ELPA correspondingly
- Update usage description
- Revise compiler flags

### DIFF
--- a/arch/CRAY-XC40-gnu.psmp
+++ b/arch/CRAY-XC40-gnu.psmp
@@ -7,11 +7,14 @@
 #              ELPA 2021.11.001, PLUMED 2.7.2, SPGLIB 1.16.2, GSL 2.7,
 #              COSMA 2.5.1, SIRIUS 7.3.0
 #
-# Usage: Source this arch file and then run make as instructed.
-#        The compiler version can be specified as argument.
-#        Otherwise a toolchain installation is expected as default.
+# Usage: Source this arch file with the GNU compiler version specified as argument:
+#        > source arch/CRAY-XC40-gnu.psmp 10.3.0
+#        Then run make as instructed after a CP2K toolchain installation has been
+#        performed successfully with
+#        > cd tools/toolchain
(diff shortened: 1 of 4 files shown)


oschuett:
Oh, won't disabling OpenMP severely hurt ELPA's performance?

We actually just upgraded to `MPI_THREAD_SERIALIZED` to keep it working (#1564).

mkrack:
Right, I forgot to restrict the change in install_elpa.sh to Cray toolchain builds where OpenMP is effectively disabled already.

oschuett:
> Cray toolchain builds where OpenMP is effectively disabled already.

I wasn't aware of that. Can you elaborate? This would still affect a large part of our user base.

mkrack:
These [build flags](https://github.com/cp2k/cp2k/blob/5da5e8ed3126c5a52c1b70c7996a60254e540919/tools/toolchain/scripts/stage5/install_elpa.sh#L80) already allow ELPA to set the number of OMP threads to one at runtime which happens on Cray XC40/50.

oschuett:
> These build flags already allow ELPA to set the number of OMP threads to one at runtime which happens on Cray XC40/50.

But that's pretty bad, right? Does it even make sense keep using ELPA in that case?

mkrack:
Of course, the CPU version of ELPA without OpenMP is much faster than ScaLAPACK (56s vs 135s per SCF iteration step involving a 29k by 29k matrix diagonalization using 288 CRAY-XC50 cores). I run most of my production cases with OMP_NUM_THREADS=1 anyway unless the memory demand per MPI task does not allow this. The GPU kernel is slightly faster (10%) than the fastest CPU kernel for that case.

oschuett:
Thanks a lot for clarifying!

Did you also observe deadlocks like #1830 and #1788?

I'm fine with disabling it for now, but we'll eventually need OpenMP support for AMD EPYC and ARM based machines.

mkrack:
ELPA with OpenMP is only disabled for Cray systems in the moment. We can further fine-tune the flags based on the actual target system, such that ARM and AMD EPYC are not affected.

PR #1874, opened 2022-01-17
Update toolchain build for Intel/MKL

mkrack:


oschuett:
It's great that you are improving on the toolchain!

Maybe this is a good time to pitch the idea of (gradually) converting it to Python?

Python's simplicity combined with its novel [static typing](http://mypy-lang.org/) should make it more robust and easier to maintain.

mkrack:
> Maybe this is a good time to pitch the idea of (gradually) converting it to Python?

That is certainly on the to-do list, but first we should get things work for GNU, Intel, and Cray using the current scheme.
> 
> Python's simplicity combined with its novel [static typing](http://mypy-lang.org/) should make it more robust and easier to maintain.

Thanks for the hint.

PR #1882, opened 2022-01-18
Fix compilation issue on summit

mtaillefumier:
- the initial code will fail on systems where libxsmm is not installed
  because the cpp test assumes the macro LIBXSMM_VERSION4 exists. The
  macro is present in global namespace only when libxsmm.h is included


hfp:
@mtaillefumier Thank you very much!

( This PR looks good to me )

PR #1901, opened 2022-01-24
GPU options Cleanup

mtaillefumier:
- use `-D__FUNCTIONALITY_GPU` flags instead of `-D__FUNCTIONALITY_CUDA` or `-D__FUNCTIONALITY_HIP`
- specialization to dedicated hardware can be done with `-D__OFFLOAD_CUDA` or `-D__OFFLOAD_HIP `(or others). It is mandatory to specify which platform is needed. if not and one of the `-D__FUNCTIONALITY_GPU` is specified then compilation error should occur.
- dbm supports hip. Right now it is just cut and paste of cuda backend. Code **sharing** is exploited as much as possible using the offload api.
- Updated the INSTALL.md to reflect these changes.
- Updated the arch file for daint as well
- removed some compilation options that date back from cuda 7.0 (or so). 
- modified the toolchain accordingly


oschuett:
Unfortunately, this PR conflicts with something I'm currently working on. I'm trying to extend the name convention used by the build system to also support HIP, which would spare us from having to list individual files in the Makefile:
```
#   .F    ->  Fortran
#   .c    ->  C
#   .cu   ->  CUDA
#   .cc   ->  HIP
#   .h    ->  C header
#   .hpp  ->  CUDA or HIP header
```


As for the flags, I totally agree that they need a cleanup. However, I had a slightly different scheme in mind:

 - Replace `OFFLOAD_TARGET` with `__OFFLOAD_CUDA` and `__OFFLOAD_HIP`, which enable _everything_.
 - Add new flags `__NO_OFFLOAD_PW`, `__NO_OFFLOAD_GRID`, and `__NO_OFFLOAD_DBM` for debugging.

Ideally, `__PW_FPGA` should also be folded into that scheme via `__OFFLOAD_FPGA`. However, we currently have no way of testing it. So, let's keep it around until it start blocking us.

The main part of this PR is to leverage your new [offload_operations.h](https://github.com/cp2k/cp2k/blob/master/src/offload/offload_operations.h) for code sharing in grid and dbm. This is great! My only comment here would be that for consistency I'd like to always use _"offload"_  instead of _"gpu"_ for naming.


So, I'd propose you reduce this PR to the code sharing part, while I take care of the flags. What do you think?

mtaillefumier:
> 

indeed `__PW_FPGA` should also follow the same logic but I also have no way to test it (well actually I do but I do not know if the code is compatible with the fpga we have inhouse).

I share your view about how to treat the files compared to the extensions. My only concern is that `.hpp` is a c++ header file  and is clearly not associated to `hip` or `cuda` in general. `.cc` files follow the same logic but i think it is a good compromise to keep the schema you have and leave the .cpp files as c++ files eventually. All hip code is in *.cc files so excluding them when hip is off is not an issue.

Only files that need to be included during compilation time should be included. Doing a general find won't cut it and the cmake build won't rely on it either (I still need to finish the integration with the CI). 

The convention of turning everything on by default is also reasonable. Give me 1/2 day to make the changes and push them I still need to modify the makefile but the modifications should not be problematic.

I always use offload keyword in the offload module. But it make little sense to prefix every function that are not running on cpu offload_*****. having `gpu_function_name` or `fpga_function_name` is clearer. We could also use `device` instead of offload but it is equally vague. Anyway, we can always come back and modify this.

mtaillefumier:
Same work but including @oschuett suggestion for compilation flags and files convention. The name of the compilation flags can be changed with a single sed.

INSTALL.md
```diff
@@ -179,11 +179,24 @@ the FFTW3 threading library libfftw3_threads (or libfftw3_omp) is required.
   and `LIBS += -L${LIBXSMM_DIR}/lib -lxsmmf -lxsmm -ldl`
 - LIBSMM is not used if LIBXSMM is enabled.
 
-### 2j. CUDA (optional, improved performance on GPU systems)
+### 2j. GPU support (optional, improved performance on GPU systems)
 
-- Specify OFFLOAD_CC (e.g. `OFFLOAD_CC = nvcc`) and
-  OFFLOAD_FLAGS (e.g. `OFFLOAD_FLAGS = -O3 -g -w --std=c++11`) variables.
-  Remember to include the support for the C++11 standard.
+By default CUDA or HIP support are enabled, imply that all cp2k code that
```
oschuett, line 187: I think, we should keep CPU-only as the default.
To enable GPU support the user should have to set either `__OFFLOAD_CUDA` or `__OFFLOAD_HIP`.

src/grid/Makefile
```diff
@@ -6,40 +6,40 @@ clean:
...
-               cpu/coefficients.o \
-               cpu/grid_collocate_dgemm.o \
-               cpu/grid_integrate_dgemm.o \
-               cpu/non_orthorombic_corrections.o \
-               cpu/utils.o \
-               cpu/collocation_integration.o \
-               cpu/grid_context_cpu.o \
-               cpu/grid_prepare_pab_dgemm.o \
-               cpu/tensor_local.o
+															../offload/offload_library.o \
```
oschuett: This indentation is pretty wild. Unfortunately, there seems to be no tool for formatting Makefiles.

oschuett (changes requested):
Sorry for the late review. I created a bit of a kerfuffle in the CI.

INSTALL.md
mtaillefumier: I also thought it would be better.

src/grid/Makefile
mtaillefumier: emacs is the culprit. space / tab... I will fix it.

oschuett: Thanks! In the meantime I've committed #1930. So, in the future `make pretty` should take care of it.

mtaillefumier: Thanks. I just saw your message. I need to do a make pretty again. Had issues with rebasing.

oschuett:
I'm going to respectfully close this PR now and instead submit a series of PRs myself - starting with #1985.

PR #1905, opened 2022-01-25
Protect against when GPU compilation is requested, but no PW

alazzaro:


mtaillefumier:
That's not the solution, just a work around. 

My point is that the build system should take care of this (the makefile in our case) and add the file if needed. Compiling empty files is bad practice and many compiler will complain about it if -pedantic or -ansi is on.

alazzaro:
I agree with you, I was simply lazy on implementing a better solution (and in any case, I agree with your comment on getting the build system on fixing it). BTW, I didn't make anything new, it is already there:

https://github.com/cp2k/cp2k/blob/6ed042cac1087e89817c5d0b274f19b5d88e6367/src/pw/pw_gpu.F#L26

oschuett:
I think, the best solution would be to ifdef the body of each subroutine, like e.g. in [pw_fpga.F](https://github.com/cp2k/cp2k/blob/master/src/pw/pw_fpga.F).
Then one also does not have to check the flag at [every call site](https://github.com/cp2k/cp2k/blob/6ed042cac1087e89817c5d0b274f19b5d88e6367/src/start/cp2k_runs.F#L214).

PR #1909, opened 2022-01-27
Fix for fft performance regression

mtaillefumier:
This PR contains hopefully some improvements that fix the performance regressions observed after my last PR.
Changes are numerous but not difficult

- fft plans are reused when possible. The default number of saved planes is fixed to 8.
    device allocations / opposite are only made when necessary.
- use _OFFLOAD{CUDA,HIP} instead of _PW{CUDA,HIP}. We still need -D__PW{CUDA,HIP} for the offload module until the build system is cleaned up.
- some minor cleanups as well


mtaillefumier:
@oschuett thanks for launching the pascal tests. It seems that the tests run faster now.  There is still one possibility, the number of fft planes that are kept. We can increase the number at the prize of space buffers are allocated. 
There is still one thing that can be done. create a fft plan independently of the direction we want to go.

oschuett:
> We can increase the number at the prize of space buffers are allocated.

IIRC, the original code held on to many of those Cuda objects via the [pw pools](https://github.com/cp2k/cp2k/blob/master/src/pw/pw_pool_types.F). Are you not using that mechanism anymore?

mtaillefumier:
This pw_pool is for cpu only. There is no gpu memory pool allocator in this. so if it was used before in anything else than the cuda explicit part then it is still in use.
The original code was using cudaMalloc/cudaFree for each and every call of the gpu driver. The only thing that it was storing are planes.
- You suggested a memory pool for cp2k and I also think it becomes needed as well. It would remove many of the doggy things I do and reduce memory footprint a little.
- I wonder if I can use the memory pool you integrated in dbm (if it supports gpus)

Further optimization can be done but at the prize of doing more invasive changes.

oschuett:
> I wonder if I can use the memory pool you integrated in dbm (if it supports gpus)

Yes, memory allocation comes up each time we look at performance - especially with GPUs.

I'm a bit undecided about the topic though: On one hand having a single memory pool shared by all subsystems would reduce the total consumption. On the other hand it would introduce an implicit coupling that could lead to some really weird and hard to debug problems.

I'd therefore suggest we err on the side of caution and keep the pools separate until we actually run out of memory.

PR #1911, opened 2022-01-28
ShellCheck fixes

e-kwsm:
- fix [SC2006](https://github.com/koalaman/shellcheck/wiki/SC2006)
- fix [SC2039](https://github.com/koalaman/shellcheck/wiki/SC2039)
- fix [SC2071](https://github.com/koalaman/shellcheck/wiki/SC2071)
- fix [SC2173](https://github.com/koalaman/shellcheck/wiki/SC2173)

### DIFF
--- a/benchmarks/QS_single_node/do_benchmark
+++ b/benchmarks/QS_single_node/do_benchmark
@@ -7,7 +7,7 @@
 # optional output postfix, useful to separate results of different benchmark runs
 postfix=""
 # used ranks, can be any value, by default the number of physical cores on the node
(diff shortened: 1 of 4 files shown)


oschuett:
That a lot for those fixes!

Please note that the [do_regtest](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest) script is deprecated as we now use mostly [do_regtest.py](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest.py).

PR #1926, opened 2022-02-02
PEP8 fixes

e-kwsm:
- fix [E401](https://github.com/PyCQA/pycodestyle/blob/2.8.0/docs/intro.rst#error-codes)
- fix [E711](https://github.com/PyCQA/pycodestyle/blob/2.8.0/docs/intro.rst#error-codes)
- fix [E713](https://github.com/PyCQA/pycodestyle/blob/2.8.0/docs/intro.rst#error-codes)
- fix [E722](https://github.com/PyCQA/pycodestyle/blob/2.8.0/docs/intro.rst#error-codes)
- fix [W601](https://github.com/PyCQA/pycodestyle/blob/2.8.0/docs/intro.rst#error-codes)
- fix [W605](https://github.com/PyCQA/pycodestyle/blob/2.8.0/docs/intro.rst#error-codes)

### DIFF
--- a/tools/build_utils/check_archives.py
+++ b/tools/build_utils/check_archives.py
@@ -2,7 +2,8 @@
 
 # author: Ole Schuett
 
-import sys, os
+import os
+import sys
 from subprocess import check_output
 from os import path
 from glob import glob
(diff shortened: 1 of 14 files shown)


oschuett:
We always run [black](https://github.com/psf/black) on all our Python files, but apparently it does not guarantee full pep8 compliance. Thanks for fixing those.

PR #1932, opened 2022-02-07
Shell: Bump version to signify changed LIBXC section

oschuett:
As a follow up to #1921 the [ASE calculator](https://gitlab.com/ase/ase/-/blob/master/ase/calculators/cp2k.py) needs to be patched.


oschuett:
I patched the ASE calculator via https://gitlab.com/ase/ase/-/merge_requests/2676.

Issue #1912, opened 2022-01-28
Empty enum label in SIRIUS input section


e-kwsm:
- [syntax/cp2k.vim#L64](https://github.com/cp2k/vim-cp2k/blob/247a31ea174bd4db2f7f6042708b6db0a648fc4a/syntax/cp2k.vim#L64) is invalid, which implies `cp2k_input.xml` has something like `<NAME/>`
- `keyword` is inappropriate for words that contain non-[`iskeyword`](https://vimhelp.org/options.txt.html#%27iskeyword%27) characters, e.g., [syntax/cp2k.vim#L65-L67](https://github.com/cp2k/vim-cp2k/blob/247a31ea174bd4db2f7f6042708b6db0a648fc4a/syntax/cp2k.vim#L65-L71); use `syn match cp2kConstant /\<-X\>/`
- it is preferable to put the following part on ftplugin/cp2k.vim rather than syntax/:
  https://github.com/cp2k/cp2k/blob/4308454eda8033a1de5b0dc6ccb64ad28edb1814/tools/input_editing/vim/vim.xsl#L103-L153

cp2k/vim-cp2k#3

oschuett:
I believe the culprit is the first option of the keyword [FORCE_EVAL / PW_DFT / PARAMETERS / PRECISION_GS](https://manual.cp2k.org/cp2k-9_1-branch/CP2K_INPUT/FORCE_EVAL/PW_DFT/PARAMETERS.html#PRECISION_GS), which was introduced in #1784.

@mtaillefumier, could you please take a look.

mtaillefumier:
I fixed it in SIRIUS code. Will be fixed next time SIRIUS is updated.

PR #1937, opened 2022-02-10
Catch NAME entries of zero string length (#1912)

mkrack:
- Detect NAME entries of zero string length during the cp2k.vim file generation
- Stop file generation and write error message pointing to the CP2K source code line


oschuett:
I was planing to add a `CPASSERT` to `keyword_create()` after we upgrade to the next SIRIUS release. See also #1912.

mkrack:
@oschuett A `CPASSERT` might be a better solution to avoid such issues in the future. Should I cancel this PR.

oschuett:
Yes, I'd suggest to cancel this PR. In any case we have to wait for that SIRIUS release.

PR #1951, opened 2022-02-17
Use SPLA for offloading dgemm on GPUs in the mp2 module

mtaillefumier:
dgemm in the mp2 module are offloaded to GPU using SPLA which already has this feature implemented. It is similar to what libsci_acc would do but it is now vendor agnostic and will support any blas vendor implementation.

- SPLA update to the latest.
- SPLA will offload the dgemm when the product 2 m n k is above a given threshold. The value is hardcoded but we can add an extra input option to set it up manually. I Fixed it to 2 * (128 ^ 3) but it can be increased further. 
- SPLA is not a hard dependency.

They are several direction for improvement. 
- Some of the matrices can be loaded only one time for instance on GPU, etc... 
- Dgemms can also be batched. But the prize to pay is code cluttering in absence of a generic interface.
- maybe use some of spla functionalities to handle mpi communication.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -399,12 +399,12 @@ The SPLA library is a hard dependency of SIRIUS but can also be used as a
 standalone library. It provides a generic interface to the blas gemm family with
 offloading on GPU. Offloading supports both CUDA and ROCM.
 
-To make the functionality available, add the flag `-D__SPLA` to the `DFLAGS`
-variable and compile SPLA with Fortran interface and GPU support. Please note
-that only the functions replacing the dgemm calls with `offload_dgemm` will
-eventually be offloaded to the GPU. The SPLA library has internal criteria to
-decide if it is worth to do the operation on GPU or not. Calls to `offload_dgemm`
-also accept pointers on GPU or a combination of them.
+To make the functionality available, add the flag `-D__SPLA -D__OFFLOAD_GEMM` to
+the `DFLAGS` variable and compile SPLA with Fortran interface and GPU support.
+Please note that only the functions replacing the dgemm calls with
+`offload_dgemm` will eventually be offloaded to the GPU. The SPLA library has
+internal criteria to decide if it is worth to do the operation on GPU or not.
+Calls to `offload_dgemm` also accept pointers on GPU or a combination of them.
 
 <!---
 ### 2w. LibMaxwell (External Maxwell Solver)
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -396,14 +396,14 @@ CP2K's grid backend does not yet support OpenCL devices.
 ### 2w. matrix-matrix multiplication offloading on GPU using SPLA
 
 The SPLA library is a hard dependency of SIRIUS but can also be used as a
-standalone library. It provide a generic interface to the blas gemm family with
+standalone library. It provides a generic interface to the blas gemm family with
 offloading on GPU. Offloading supports both CUDA and ROCM.
 
-To make the functionality availablem, add the flag `-D__SPLA` to the `DFLAGS`
-variable and compile SPLA with fortran interface and GPU support. Please note
+To make the functionality available, add the flag `-D__SPLA` to the `DFLAGS`
+variable and compile SPLA with Fortran interface and GPU support. Please note
 that only the functions replacing the dgemm calls with `offload_dgemm` will
-eventually be offloaded to the GPU. The spla library has a internal criteria to
-decide if it is worth to do the operation on GPU or not. Calling offload_dgemm
+eventually be offloaded to the GPU. The SPLA library has internal criteria to
+decide if it is worth to do the operation on GPU or not. Calls to `offload_dgemm`
 also accept pointers on GPU or a combination of them.
 
 <!---
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -393,6 +393,19 @@ CP2K's grid backend does not yet support OpenCL devices.
 - Refer to <https://cp2k.github.io/dbcsr/> for more information, e.g.,
   environment variables or how to tune kernels (auto tuned parameters).
 
+### 2w. matrix-matrix multiplication offloading on GPU using SPLA
+
+The SPLA library is a hard dependency of SIRIUS but can also be used as a
+standalone library. It provide a generic interface to the blas gemm family with
+offloading on GPU. Offloading supports both CUDA and ROCM.
+
+To make the functionality availablem, add the flag `-D__SPLA` to the `DFLAGS`
+variable and compile SPLA with fortran interface and GPU support. Please note
+that only the functions replacing the dgemm calls with `offload_dgemm` will
+eventually be offloaded to the GPU. The spla library has a internal criteria to
+decide if it is worth to do the operation on GPU or not. Calling offload_dgemm
+also accept pointers on GPU or a combination of them.
+
 <!---
 ### 2w. LibMaxwell (External Maxwell Solver)
 
--- a/src/cp2k_info.F
+++ b/src/cp2k_info.F
@@ -304,8 +304,8 @@ CONTAINS
       flags = TRIM(flags)//" offload_profiling"
 #endif
 
-#if defined(__SPLA)
-      flags = TRIM(flags)//" slpa_gemm_offloading"
+#if defined(__SPLA) && defined(__OFFLOAD_GEMM)
+      flags = TRIM(flags)//" spla_gemm_offloading"
 #endif
 
    END FUNCTION cp2k_flags
--- a/src/cp2k_info.F
+++ b/src/cp2k_info.F
@@ -304,6 +304,10 @@ CONTAINS
       flags = TRIM(flags)//" offload_profiling"
 #endif
 
+#if defined(__SPLA)
+      flags = TRIM(flags)//" slpa_gemm_offloading"
+#endif
+
    END FUNCTION cp2k_flags
 
 ! **************************************************************************************************
--- a/src/mp2_ri_gpw.F
+++ b/src/mp2_ri_gpw.F
@@ -386,11 +386,6 @@ CONTAINS
                            local_ba = 0.0_dp
                         END IF
                         CALL dgemm_counter_start(dgemm_counter)
-#if __GNUC__ >= 9
-                        CPASSERT(IS_CONTIGUOUS(my_local_i_aL))
-                        CPASSERT(IS_CONTIGUOUS(my_local_j_aL))
-                        CPASSERT(IS_CONTIGUOUS(local_ab))
-#endif
                         CALL offload_dgemm('T', 'N', my_B_size(ispin), my_B_size(jspin), dimen_RI, 1.0_dp, &
                                            my_local_i_aL, dimen_RI, my_local_j_aL, dimen_RI, &
                                          0.0_dp, local_ab(my_B_virtual_start(ispin):my_B_virtual_end(ispin), :), my_B_size(ispin), &
--- a/src/mp2_ri_gpw.F
+++ b/src/mp2_ri_gpw.F
@@ -41,6 +41,7 @@ MODULE mp2_ri_gpw
                                               mp2_type,&
                                               three_dim_real_array,&
                                               two_dim_int_array
+   USE offload_gemm,                    ONLY: offload_dgemm
 
 !$ USE OMP_LIB, ONLY: omp_get_max_threads, omp_get_thread_num
 #include "./base/base_uses.f90"
@@ -114,9 +115,11 @@ CONTAINS
                                                             my_beta_beta_case, my_open_shell_SS
       REAL(KIND=dp) :: amp_fac, mem_for_aK, mem_for_comm, mem_for_iaK, mem_for_rep, mem_min, &
          mem_per_group, mem_real, my_Emp2_Cou, my_Emp2_EX, sym_fac, t_new, t_start
-      REAL(KIND=dp), ALLOCATABLE, DIMENSION(:, :)        :: external_ab, external_i_aL, local_ab, &
+      REAL(KIND=dp), ALLOCATABLE, DIMENSION(:, :), &
+         TARGET                                          :: external_ab, external_i_aL, local_ab, &
                                                             local_ba, t_ab
-      REAL(KIND=dp), ALLOCATABLE, DIMENSION(:, :, :)     :: BI_C_rec, local_i_aL, local_j_aL, &
+      REAL(KIND=dp), ALLOCATABLE, DIMENSION(:, :, :), &
+         TARGET                                          :: BI_C_rec, local_i_aL, local_j_aL, &
                                                             Y_i_aP, Y_j_aP
       TYPE(cp_para_env_type), POINTER                    :: para_env_exchange, para_env_P, &
                                                             para_env_rep
@@ -375,7 +378,6 @@ CONTAINS
                DO iiB = 1, my_block_size
                   DO jjB = 1, my_block_size
                      CALL timeset(routineN//"_expansion", handle3)
-
                      ASSOCIATE (my_local_i_aL => local_i_aL(:, :, iiB), my_local_j_aL => local_j_aL(:, :, jjB))
 
                         ! calculate the integrals (ia|jb) strating from my local data ...
@@ -384,9 +386,15 @@ CONTAINS
                            local_ba = 0.0_dp
                         END IF
                         CALL dgemm_counter_start(dgemm_counter)
-                        CALL dgemm('T', 'N', my_B_size(ispin), my_B_size(jspin), dimen_RI, 1.0_dp, &
-                                   my_local_i_aL, dimen_RI, my_local_j_aL, dimen_RI, &
-                                   0.0_dp, local_ab(my_B_virtual_start(ispin):my_B_virtual_end(ispin), :), my_B_size(ispin))
+#if __GNUC__ >= 9
+                        CPASSERT(IS_CONTIGUOUS(my_local_i_aL))
+                        CPASSERT(IS_CONTIGUOUS(my_local_j_aL))
+                        CPASSERT(IS_CONTIGUOUS(local_ab))
+#endif
+                        CALL offload_dgemm('T', 'N', my_B_size(ispin), my_B_size(jspin), dimen_RI, 1.0_dp, &
+                                           my_local_i_aL, dimen_RI, my_local_j_aL, dimen_RI, &
+                                         0.0_dp, local_ab(my_B_virtual_start(ispin):my_B_virtual_end(ispin), :), my_B_size(ispin), &
+                                           mp2_env%offload_gemm_ctx)
                         ! Additional integrals only for alpha_beta case and forces
                         IF (my_alpha_beta_case .AND. calc_forces) THEN
                            local_ba(my_B_virtual_start(jspin):my_B_virtual_end(jspin), :) = &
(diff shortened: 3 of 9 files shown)


src/mp2_types.F
```diff
@@ -331,6 +361,7 @@ END SUBROUTINE mp2_env_create
 
 ! **************************************************************************************************
 !> \brief Initialize a dgemm_counter
+!>     > \param dgemm_counter ...
```
fstein93: Was this line added by accident?

src/mp2_ri_gpw.F
```diff
@@ -1469,32 +1534,65 @@ SUBROUTINE mp2_update_P_gamma(mp2_env, para_env_sub, gd_B_virtual, &
 
       CALL dgemm_counter_start(dgemm_counter)
       IF (.NOT. (alpha_beta)) THEN
+#if defined(__SPLA)
+         spla_error_code = spla_dgemm(SPLA_OP_TRANSPOSE, SPLA_OP_NONE, my_B_size(ispin), my_B_size(ispin), virtual(ispin), 1.0_dp, &
+                                      c_loc(t_ab), virtual(ispin), c_loc(local_ab), virtual(ispin), &
+                                      1.0_dp, c_loc(mp2_env%ri_grad%P_ab(ispin)%array(:, &
+                                                                              my_B_virtual_start(ispin):my_B_virtual_end(ispin))), &
+                                      my_B_size(ispin), mp2_env%spla_ctx)
+#else
```
fstein93: Is it possible to shift this dependence on SPLA to an external module which makes it more reusable to other parts of the code?

src/mp2_types.F
mtaillefumier: make pretty does that.

src/mp2_ri_gpw.F
mtaillefumier: And would be the point of adding yet another interface to an interface. The only thing needed would be to move the spla context one step up if you want to use it elsewhere. Right now the context is located in the mp2 structure. 

Moreover I do not want to overload the dgemm function because offloading is not always needed.

fstein93: If you are allowed to have only one SPLA structure, I prefer to have it at a higher level then such that a developer who wants to use SPLA elsewhere does not run into conflicts with the MP2 code.

fstein93:
I have just rechecked the installation instructions. Could you also add a paragraph to SPLA in our [Installation instructions](https://github.com/cp2k/cp2k/blob/master/INSTALL.md)?

mtaillefumier:
I will do it for sure. I am trying to find why I do not get the same answer when I run with 2 or more mpi ranks. I will also clean up the generic interface and see if I can move the context one level above the mp2 type so we can also use spla for pdgemm families.

INSTALL.md
```diff
@@ -390,6 +390,19 @@ CP2K's grid backend does not yet support OpenCL devices.
 - Refer to <https://cp2k.github.io/dbcsr/> for more information, e.g.,
   environment variables or how to tune kernels (auto tuned parameters).
 
+### 2w. matrix-matrix multiplication offloading on GPU using SPLA
+
+The SPLA library is a hard dependency of SIRIUS but can also be used as a
+standalone library. It provide a generic interface to the blas gemm family with
```
fstein93: It provides

INSTALL.md
```diff
@@ -390,6 +390,19 @@ CP2K's grid backend does not yet support OpenCL devices.
...
+offloading on GPU. Offloading supports both CUDA and ROCM.
+
+To make the functionality availablem, add the flag `-D__SPLA` to the `DFLAGS`
```
fstein93: available

INSTALL.md
```diff
@@ -390,6 +390,19 @@ CP2K's grid backend does not yet support OpenCL devices.
...
+variable and compile SPLA with fortran interface and GPU support. Please note
```
fstein93: Fortran (capital letter)

INSTALL.md
```diff
@@ -390,6 +390,19 @@ CP2K's grid backend does not yet support OpenCL devices.
...
+that only the functions replacing the dgemm calls with `offload_dgemm` will
+eventually be offloaded to the GPU. The spla library has a internal criteria to
```
fstein93: SPLA (for consistency)
has internal criteria

INSTALL.md
```diff
@@ -390,6 +390,19 @@ CP2K's grid backend does not yet support OpenCL devices.
...
+decide if it is worth to do the operation on GPU or not. Calling offload_dgemm
+also accept pointers on GPU or a combination of them.
```
fstein93: The call to offload_dgemm also accepts

oschuett:
Since this is a performance optimization, I'm wondering if we should add a daily benchmark to catch regressions.
Do you happen to have an input file at hand that can run on a single GPU?

mtaillefumier:
> Since this is a performance optimization, I'm wondering if we should add a daily benchmark to catch regressions. Do you happen to have an input file at hand that can run on a single GPU?

that's a very good point. The only issue to have a benchmark that does not take too long to run. I rely on the regtests for validation and I put the threshold to 2 n m m, m = n = k =128 but it might be too low. Maybe a pure dgemm benchmark would be enough ?

mtaillefumier:
It raises another question, shall we think about making the dgemm calls a bit more consistent across the code. Many things are called dgemm but they do not necessarily means exactly the same thing. I am thinking for instance of cp_dgemm which is I think a pdgemm.

fstein93:
I am also more in favor of a DGEMM-only benchmark as MP2-based benchmarks are too expensive to test just the calls to dgemm. With 32 water molecules, the DGEMM part takes at most 30 % of the total computation time on Daint (ca. 1 min on 8 nodes, see regular tests on Daint). With smaller systems, the ratio is even lower.

oschuett:
If possible I'd prefer to run a full MP2 calculation, e.g. by reducing system size and loosening parameters. Usually, the top 5 routines in the timings will remain the same - albeit with different factions. A more realistic benchmark will catch a much broader range of problems.

mtaillefumier:
I probably forgot to build the fortran module for SPLA...

fstein93:
I am compiling a benchmark which should run on a single node (ammonia, 16 atoms per unit cell). @mtaillefumier : How does your GPU relate to the Daint GPU section or do you test directly on Daint?

mtaillefumier:
I am not sure to understand the question actually. I can test on pretty much anything from K80 to A100, Mi50 ~ mi200 (if I ask nicely). SPLA is vendor agnostic so it will run on pretty anything that NVIDIA/AMD have available.

The thing is that I will need to compile cp2k with and without SPLA for the time being because I did not add any optional input parameter to turn it on or off.

fstein93:
@mtaillefumier : I have sent you a message with a simple benchmark.

src/mp2_ri_gpw.F
```diff
@@ -384,9 +386,15 @@ SUBROUTINE mp2_ri_gpw_compute_en(Emp2_Cou, Emp2_EX, Emp2_S, Emp2_T, BIb_C, mp2_e
                            local_ba = 0.0_dp
                         END IF
                         CALL dgemm_counter_start(dgemm_counter)
-                        CALL dgemm('T', 'N', my_B_size(ispin), my_B_size(jspin), dimen_RI, 1.0_dp, &
-                                   my_local_i_aL, dimen_RI, my_local_j_aL, dimen_RI, &
-                                   0.0_dp, local_ab(my_B_virtual_start(ispin):my_B_virtual_end(ispin), :), my_B_size(ispin))
+#if __GNUC__ >= 9
```
oschuett: I'd move those checks into `offload_dgemm`.

src/mp2_types.F
```diff
@@ -323,6 +333,15 @@ SUBROUTINE mp2_env_create(mp2_env)
 
       ALLOCATE (mp2_env)
 
+      CALL offload_set_device()
+
+#if defined(__OFFLOAD_CUDA) || defined(__OFFLOAD_HIP)
```
oschuett: This logic should go into `offload_gemm_create`.

src/mp2_types.F
```diff
@@ -323,6 +333,15 @@ SUBROUTINE mp2_env_create(mp2_env)
 
       ALLOCATE (mp2_env)
 
+      CALL offload_set_device()
```
oschuett: This should be called  inside `offload_dgemm` before every call to SPLA.

src/offload_gemm.F
```diff
@@ -0,0 +1,395 @@
...
+!> \param A ...
+!> \param lda ...
+!> \param B ...
+!> \param ldb ...
+!> \param beta ...
+!> \param C ...
+!> \param ldc ...
+!> \param ctx ...
+! **************************************************************************************************
+   SUBROUTINE offload_dgemm(opA, opB, m, n, k, &
```
oschuett, line 67: This routine should get a timer region.

src/offload_gemm.F
```diff
@@ -0,0 +1,395 @@
...
+!> \param A ...
+!> \param lda ...
+!> \param B ...
+!> \param ldb ...
+!> \param beta ...
+!> \param C ...
+!> \param ldc ...
+!> \param ctx ...
+! **************************************************************************************************
+   SUBROUTINE offload_sgemm(opA, opB, m, n, k, &
```
oschuett: These other `gemm` routines are not being used and probably never will. So, I'd rather remove them.

src/cp2k_info.F
```diff
@@ -304,6 +304,10 @@ FUNCTION cp2k_flags() RESULT(flags)
       flags = TRIM(flags)//" offload_profiling"
 #endif
 
+#if defined(__SPLA)
+      flags = TRIM(flags)//" slpa_gemm_offloading"
```
oschuett: For consistency please just use `" spla"`.

src/offload_gemm.F
```diff
@@ -0,0 +1,395 @@
+!--------------------------------------------------------------------------------------------------!
+!   CP2K: A general program to perform molecular dynamics simulations                              !
+!   Copyright 2000-2022 CP2K developers group <https://cp2k.org>                                   !
+!                                                                                                  !
+!   SPDX-License-Identifier: GPL-2.0-or-later                                                      !
+!--------------------------------------------------------------------------------------------------!
+
+MODULE offload_gemm
```
oschuett, line 8: I'm a bit unhappy about the file name, because usually the prefix indicates the sub-directory. However, as you pointed out in https://github.com/cp2k/cp2k/pull/1951#issuecomment-1049811610, our `pdgemm` / `dgemm` code needs a bit of reorganization anyways. So, we can merge it like this and then I'll move things around in the coming weeks.

oschuett:
For some reason the new [RI-MP2_ammonia.inp](https://github.com/cp2k/cp2k/blob/master/benchmarks/QS_single_node/RI-MP2_ammonia.inp) benchmark is slower [after](https://storage.googleapis.com/cp2k-ci/run-cp2k-perf-openmp-a55b4568_report.txt) this change than [before](https://dashboard.cp2k.org/archive/perf-openmp/commit_2e6367024a297e833823c1085e0d4b7f48dd4c0d.txt).

mtaillefumier:
> For some reason the new [RI-MP2_ammonia.inp](https://github.com/cp2k/cp2k/blob/master/benchmarks/QS_single_node/RI-MP2_ammonia.inp) benchmark is slower [after](https://storage.googleapis.com/cp2k-ci/run-cp2k-perf-openmp-a55b4568_report.txt) this change than [before](https://dashboard.cp2k.org/archive/perf-openmp/commit_2e6367024a297e833823c1085e0d4b7f48dd4c0d.txt).

It may have something to do with this `REAL(8), DIMENSION(*), INTENT(in), TARGET :: B`. 

spla does not support non contiguous tables. I implemented this `REAL(8), DIMENSION(:, :), INTENT(in), TARGET ::B` and it did not work as soon as I run the code with more than one mpi rank.

src/mp2_types.F
mtaillefumier: Actually it is even wrong with the current state of cp2k build system. This pass will never be taken unless the PR #1901 is merged somehow. One way to solve this is to add yet another `-D__OFFLOAD_DGEMM` type of compilation flag. There is also no point of using SPLA on cpu only runs so I might make the interface more dependent of the offload flags. Suggestions are welcone to try to make this cleaner.

src/mp2_ri_gpw.F
mtaillefumier: Indeed. I left it but I check this internally. I will remove these.

src/mp2_types.F
oschuett: Yeah, #1901 ended up in a weird place. I'll look into it later this week (probably by merging some of the changes I had already prepared). For the meantime, I'd suggest we decouple these two PRs by introducing `-D__OFFLOAD_DGEMM`.

oschuett:
> spla does not support non contiguous tables. 

If you are referring to Fortran90 pointers, then neither does BLAS. It does of courses support the usual leading dimension arguments. However, even an additional copy would IMHO not explain the massive slowdown:


|        | 32 MPI ranks | 32 OpenMP threads |
| ------ | ------------ | ----------------- |
| Before |  35.943      | 150.173           |
| After  |  52.257      | 433.789           |

mtaillefumier:
Actually the problem lies inside SPLA. It has an internal way to handle gemm that are not OMP parallel and I think the problem comes from this. Fortunately it is possible to turn it off during compilation time. Now some numbers about the ammonia test.
Running the code OMP = 32 is not particularly wise because OMP does not really scale that well above 8 threads (of course it is code dependent). Nevertheless it is a good exercise so I ran the same test on a AMD node and 4 A100. Results are 

MPI 32 OMP 1 spla
CP2K                                14.125
offload_gemm                   0.282

MPI 32 OMP 1 no spla
CP2K   13.975 
offload_gemm     0.376 

MPI 1 OMP 32 spla
CP2K   94.374
offload_dgemm NA

MPI 1 OMP 32 no spla
CP2K   103.763
offload_gemm   9.978

The only difference is SPLA (SPLA internal OMP off) **on** or **off** between the two executables. All the others parameters are identical. Runs with MKL for blas and lapack.

this benchmarks is still too small to say anything about the offloading benefit. 

Now with a different set of parameters OMP = 4 MPI 4

No spla :  
- CP2K: 20.698 s
- offload_dgemm : 2.530 s
SPLA : 
- CP2K : 18.929
- offload_dgemm : 0.390 s

We have a factor 8 difference for the offloading of dgemm but overall it presents relatively little compared to the rest.

fstein93:
@mtaillefumier : What dominates the costs? If it is Hartree-Fock (Integrate_four_center_integrals or similar), remove the HF section (like in the benchmark finally published by @oschuett). (Hartree-Fock is not at all GPU accelerated. For larger systems, Hartree-Fock becomes irrelevant or is reduced with RI or ADMM.)

oschuett:
> this benchmarks is still too small to say anything about the offloading benefit.

Yes, I sized the benchmark using this PR with the old threading. I've opened #1971 to fix it. Afterwards 30% will be spent in `mp2_ri_gpw_compute_en_expansio`.

mtaillefumier:
thanks. The important was to identify why it was slowing down and resolve the issue. I will try the newest version of this benchmark and report the results back.

mtaillefumier:
@fstein93 : yes time is spent in other lapack functions. The story is different with the test case you sent me by email. In that case the four center-integrals dominate all the rest.

oschuett:
Performance looks ok. Unfortunately, there is now a conflict - otherwise this seems ready to be merged.

mtaillefumier:
I will fix it tomorrow. I had some technical difficulties to tests the last few days.

mtaillefumier:
I managed to run cp2k on the A100 system we have in house. Runs on AMD cpu with mkl so timing reports for cpu dgemms are a bit off because mkl is slower on amd systems.

| run | offloading | cpu dgemm  |
| ---- | -------------- | ------------------ |
| OMP 1 MPI 32 | 3.746 | 10.416 |
| OMP 2 MPI 16 | 4.325 |  19.039 |
| OMP 4 MPI 8 | 3.644 | 37.803 |
| OMP 8 MPI 4 | 4.325 | 75.547 |

mtaillefumier:
Squashing this PR is probably better.

mtaillefumier:
It is failing because the mp2_type initialization occurs somewhere else and not in the module where the mp2_type is defined.
@fstein93 why did you suppress the `mp2_env_create` function but left the 'mp2_env_release' function. Either suppress both or leave both of them. 

SPLA is initialized in the `mp2_env_create` function which is present in my branch but not in the cp2k branch.

fstein93:
You are right, I will remove the release function. I have removed it because after some more cleanup, there was nothing left for initialization.

mtaillefumier:
Actually I *do* need the create function to initialize the spla context otherwise I have to potentially initialize a spla context everywhere a mp2_type variable is. It is cumbersome and error prone as we can see with the regtests.

fstein93:
Then, I will reintroduce an empty creation function.

fstein93:
The mp2_env_create routine must be called instead of ALLOCATE(qs_env%mp2_env) in qs_environment.F (around line 890).

mtaillefumier:
I will make that change.

oschuett:
Gentle ping. This PR broke a couple of our dashboard tests.

fstein93:
To give some hint on the broken tests:
All test cases have the variable GROUP_SIZE 2 in common (in the WF_CORRELATION section). This keyword sets the size of the processes communicating via the communicator of para_env_sub. This affects the distribution of the virtual orbitals (start and end values are given by my_B_start and my_B_end) as this index is not distributed among these subgroups.
Can it be that different compilers treat the ranges of the arrays slightly differently when you use just certain stripes of them?

alazzaro:
As far I can see, the only tests with problems are the Intel tests at PSI. If so, I don't see how they are correlated to SPLA. @mtaillefumier confirmed that SPLA is not used on the CPU tests, so I suggest to open a new issue to track those problems...

mkrack:
The problem is triggered by the keyword `GROUP_SIZE 2` in the WF_CORRELATION section as @fstein93 guessed already. The test `H2O_block_mp2_grad.inp` does not fail for `GROUP_SIZE 1`.

alazzaro:
Thanks @mkrack for the confirmation. What it is not clear to me (sorry for my ignorance) if this is SPLA related. If not, then it should deserve its own issue...

fstein93:
Is it possible that the different compilers make different assumptions on the contiguity of the passed arrays or pass them differently? If we set `GROUP_SIZE 1`, the passed arrays will always be contiguous in memory whereas if we set `GROUP_SIZE 2`, they are not; then the lower and upper bounds of the first (contiguous) index do not match the asked bounds of the blocks. As a consequence, we will have the same issues with larger group sizes. This keyword is useful to reduce the memory needs of the integration steps (with GPW) and the communication costs of the contraction step. Apparently, it also helps with MME, but I lack familiarity with that method. As such, I don't want to remove this feature.

mkrack:
> Thanks @mkrack for the confirmation. What it is not clear to me (sorry for my ignorance) if this is SPLA related. If not, then it should deserve its own issue...

Currently, I do not see any link to SPLA.

mtaillefumier:
my two cents. @fstein93 is most probably right. The compilers may treat differently `a(c:c+4,:)`  for some unknown reason (to me) making the array not continuous in memory. This behavior was the reason why I had to change the offload gemm interface from `dimension(:,:) a` to `dimension(*) a` and the first option did not work when I ran the tests on more than one rank.

mtaillefumier:
A quick way to test this is to replace `dimension(:,:)` with `dimension(*)`

mkrack:
It seems that #2019 fixed the issue with the Intel compiler. Obviously, that compiler is affected by a wrong choice for an INTENT attribute.

PR #1970, opened 2022-02-28
Facilitate build of a (generic) statically linked CP2k binary

mkrack:


oschuett:
Do you think this could be integrated into the toolchain, e.g. `./install_cp2k_toolchain.sh --static` ?

mkrack:
The idea is that you run just
```
source arch/Linux-gnu-x64_64-static.ssmp generic
make -j ARCH=Linux-gnu-x64_64-static VERSION=ssmp GENERIC=yes test
```
I hope that can be somehow automatised, e.g. for the release process to create and upload the generic statically linked CP2K binaries.

oschuett:
The problem I see with this approach is that we're getting a lot of duplication between the handwritten arch files and [generate_arch_files.sh](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/generate_arch_files.sh).

I think, it would be desirable to some day have only generated arch files. I do realize that this is a tall order. One would probably have to start by rewriting `generate_arch_files.sh` as a Python script before adding more features.

mkrack:
Indeed, `generate_arch_files.sh` should create better structured arch files. At least, we have some handwritten templates now.

oschuett:
> At least, we have some handwritten templates now.

Yes, gathering the necessary know-how is definitely equally important. Thanks for doing this!

PR #1972, opened 2022-03-02
Add LibXC equivalents to ADMM exchange potentials

### LINKED ISSUE #1960: Suitable warning message for ADMM forces
Hi, 
I noticed that if user sets `ADMM_PURIFICATION_METHOD` to values except `MO_DIAG` and `NONE`, geometry optimisation will fail but the log file does not provide an error message. 
image

Inspecting the `admm_methods.F` at line 2033 shows the source of error as forces are not implemented otherwise.
https://github.com/cp2k/cp2k/blob/d40d240e9e932ae2deae73189bc2060db0eb963a/src/admm_methods.F#L2026-L2034
but the `CPABORT` does not contain error message.
If it helps I can open a PR adding these error messages.
Best regards,
Pezhman

fstein93:
This commit adds the LibXC equivalents to the already available exchange correction functionals for ADMM. Because LibXC does not have a truncated-Coulomb-based functional (see [here](https://gitlab.com/libxc/libxc/-/issues/324)), the corresponding options still use the CP2K versions.
The main use of the new functionality is the availability of analytical 2nd derivatives (Double Hybrids, Harris functional, TDDF(P)T) with ADMM for most operators (apart from truncated Coulomb and mixed truncated Coulomb/longrange Coulomb).

Fixes #1960


oschuett:
I always disliked the way `create_admm_xc_section` meddled with the input section.

I think, it would be much cleaner if the input was first parsed into a `TYPE`. Then one could do the ADMM modifications on that type variable in a generic fashion before passing it on for evaluation.

Maybe this is a good opportunity for a refactoring?

juerghutter:
Unfortunately, xc_section is an argument in approx. 20 routines. This would be a hell of a refactoring.

PR #1985, opened 2022-03-08
Offload: Simplify offload_operations.h and add offload_fft.h

oschuett:


mtaillefumier:
What the all purpose of redoing what I have done and is already working. I used c++ for a very good reason here and you are redoing it.....

oschuett:
> I used c++ for a very good reason here and you are redoing it.....

As I've mentioned on many occasions, CP2K is not a C++ code and I would like to keep it that way. While CUDA and HIP do rely on C++, this should not be taken as a free pass.

However, my main intention was to simplify things by introducing `offload_fft.h`. Since this required a major refactoring, I eventually decided to start over from Andreas' original C code.

Please don't take this the wrong way. Thanks to your early work we already got two months worth of experience with `hipFFT` under our belts!

mtaillefumier:
For your information, the fft work is currently used by AMD / Cray for lumi and ORNL systems for benchmarking and testing so it  affect them as more than it affects me. The code was as generic as it could be (without using too much templating) with one single class that encapsulate everything that's why I do not see really see this PR as useful. Renaming functions is not an issue just matter of convention.

While I could understand the will of simplifying things (which i did by the way 5 files in total) the argument that it is c++ so it is a **no go** is quite arguable. I could even use hipfft to encapsulate both AMD and NVidia calls which would reduce things by one header file.
With this logic we would have to rewrite the hip backend for the grid module, dbcsr,  spla, spfft, DBCSR, or  SIRIUS just because they do not suit your view on things.

alazzaro:
My 2c. The current accelerated FFT code is working very well on several of my tests. I would not change it, unless there is a good reason for doing that (performance, portability... ).

oschuett:
> While I could understand the will of simplifying things (which i did by the way 5 files in total)...

Yes, you did a great job simplifying it by introducing `offload_operations.h` and now I'm simplifying it some more by introducing `offload_fft.h`.

> the fft work is currently used by AMD / Cray for lumi and ORNL systems 

Quite frankly, I don't think we should hold our development because of that. The two possible solutions I see are either for Cray to stick to release versions or to automatize their tests and contribute them to the dashboard. Of course, I'd very much prefer the later, because for HIP performance we are flying blind at the moment.

alazzaro:
> > While I could understand the will of simplifying things (which i did by the way 5 files in total)...
> 
> Yes, you did a great job simplifying it by introducing `offload_operations.h` and now I'm simplifying it some more by introducing `offload_fft.h`.
> 
> > the fft work is currently used by AMD / Cray for lumi and ORNL systems
> 
> Quite frankly, I don't think we should hold our development because of that. The two possible solutions I see are either for Cray to stick to release versions or to automatize their tests and contribute them to the dashboard. Of course, I'd very much prefer the later, because for HIP performance we are flying blind at the moment.

This is what I'm trying to do, just need to wait when LUMI will become available... However, this is a chicken-egg problem: we have make HIP code working and then performance. I don't understand the point of the performance right now...

oschuett:
> This is what I'm trying to do, just need to wait when LUMI will become available...

That's awesome!

> we have make HIP code working and then performance. I don't understand the point of the performance right now...

Of course, we also have to check correctness, but there we have at least the [HIP Pascal regtest](https://dashboard.cp2k.org/archive/hip-pascal/index.html). Nevertheless, I'd also very much prefer to run the regtests on actual AMD hardware.

PR #1995, opened 2022-03-11
Fix HIP/ROCm compilation some more

oschuett:


alazzaro:
Is now the Mi100 test available?
Likely you need a new DBCSR. Could you test without it if the tests pass?

oschuett:
> Is now the Mi100 test available?

Sorta... there are only two usable compute nodes left.

>  Could you test without it if the tests pass?

Ok, it's running.

The [error](https://storage.googleapis.com/cp2k-ci/run-cp2k-hip-mi100-e9100f0a_report.txt) looks like https://github.com/RadeonOpenCompute/ROCm/issues/1611. So we probably need to upgrade to ROCm 5.0.

oschuett:
Ok, it's running, but very slowly. This is an old problem, which I haven't been able to solve yet. Nevertheless, I'm going to merge this PR as it at least fixes the build.

alazzaro:
The error should be already fixed in the new DBCSR, so no need to upgrade to rocm 5. Note that all internal tests we did are based on rocm 4.5.2, so I will stick with this version to set a baseline in cp2k and update afterward.
I will take care of it.

Last request: could you add the test to the dashboard?

oschuett:
> Note that all internal tests we did are based on rocm 4.5.2

I'm also not eager to upgrade because it's [very tedious](https://github.com/cp2k/cp2k/blob/master/tools/docker/generate_dockerfiles.py#L426-L475) on the HIP/CUDA side.

> The error should be already fixed in the new DBCSR

Great! Then you can simply revert https://github.com/cp2k/cp2k/pull/1995/commits/adae71f251538d43cb4889657680539349b50188 as part of your #1983.

> Last request: could you add the test to the dashboard?

I don't think adding it to the dashboard makes sense before it's actually passing. I did look into those mysterious slowdowns when they appeared. They are flaky. My best theory so far is thermal throttling.

Furthermore, we are actually down to a single suitable node. I hope we get access to new hardware within the next week. Otherwise, I'll setup a build-only test for HIP/ROCm and add it to the dashboard.

PR #2004, opened 2022-03-14
dbm: minor correction related to miniapp

hfp:
* Account for renamed translation unit/symlink (Makefile).
* Print performance in MFLOP/s (1e-9 would be GFLOP/s).


hfp:
> Print performance in MFLOP/s (1e-9 would be GFLOP/s).

If GFLOP/s was desired, the code may be changed again.

PR #2006, opened 2022-03-14
RI_HFX| much better sparsity management in RHO flavor forces

abussy:
This re-implementation of the forces in the RHO flavor of RI-HFX manages sparsity in a better way than before. In particular, the order of the many 3-center tensor contractions is changed such that densification happens at the last possible step, thus preserving sparsity for longer. This is partly made possible thanks to #1984.

### DIFF
--- a/src/hfx_ri.F
+++ b/src/hfx_ri.F
@@ -47,11 +47,7 @@ MODULE hfx_ri
         dbt_iterator_type, dbt_mp_environ_pgrid, dbt_nd_mp_comm, dbt_pgrid_create, &
         dbt_pgrid_destroy, dbt_pgrid_type, dbt_type
    USE distribution_2d_types,           ONLY: distribution_2d_type
-   USE hfx_types,                       ONLY: alloc_containers,&
-                                              block_ind_type,&
-                                              dealloc_containers,&
-                                              hfx_compression_type,&
-                                              hfx_ri_init,&
+   USE hfx_types,                       ONLY: hfx_ri_init,&
                                               hfx_ri_release,&
                                               hfx_ri_type
    USE input_constants,                 ONLY: hfx_ri_do_2c_cholesky,&
@@ -107,7 +103,7 @@ MODULE hfx_ri
    IMPLICIT NONE
    PRIVATE
 
-   PUBLIC :: hfx_ri_update_ks, hfx_ri_update_forces
+   PUBLIC :: hfx_ri_update_ks, hfx_ri_update_forces, get_force_from_3c_trace, get_2c_der_force, get_idx_to_atom
 
    CHARACTER(len=*), PARAMETER, PRIVATE :: moduleN = 'hfx_ri'
 CONTAINS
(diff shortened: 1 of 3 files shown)


oschuett:
Performance-wise this PR was somewhat of a mixed bag. We really need to better understand what's happening in dbt_contract.

image

abussy:
From my own tests (on rather larger and sparser systems, without GPUs), I have observed a major performance benefit when it comes to the force evaluation. The overall time spent in the routine is at least halved. I didn't spent much time looking at more specific timings though. Having a closer look, I have for 64 water on 16 nodes (96 MPI ranks, 2 OMP threads per rank):

|Time (s)    | Old      | New    |
|--------------|-----------|----------|
| SCF         | 680      | 680     |
| Forces     | 575      | 141     |
| dbt_contract | 1041 |  656 |
| dbm_multiply | 772 | 461 |

So here I have a clear improvement, both int `dbt_contract` and `dbm_multiply`. I suspect that the 32 water system might be too small and not sparse enough for us to see the benefit, and that we have some overhead associated to `retain_sparsity`. I'll do an extra run with GPU to compare. I also agree that better understanding `dbt_contract` cannot hurt.

oschuett:
Thanks for double-checking! Yes, these small single-node benchmarks always have to be taken with a grain of salt. Comparing the changes for the [MPI vs OpenMP](https://dashboard.cp2k.org/archive/perf-openmp/index.html) run is also interesting. I believe what we're seeing for `dbt_contract` is the lack of OpenMP parallelism combined with load-imbalance. That's good news because these problems should be fairly straightforward to fix.

hfp:
> single-node benchmarks always have to be taken with a grain of salt.

Indeed, but still useful to set expectations for strong and weak scaling performance. Is there a chance to extend the miniapp to have (larger/different) inputs reproducing "real" performance like the task files for the grid backend?

oschuett:
> Is there a chance to extend the miniapp to have (larger/different) inputs reproducing "real" performance like the task files for the grid backend?

Yes, this is definitely high up on my list.

abussy:
> Performance-wise this PR was somewhat of a mixed bag. We really need to better understand what's happening in dbt_contract.
> 
> image

@oschuett I am aware that this is old history now, but I might know what was going on. In an effort to unify the convention of RI-HFX and low-scaling post-SCF methods (RPA and Laplace SOS-MP2), I changed the meaning of the HF%RI%MEMORY_CUT keyword. 

In all these methods, there are two nested DO loops when contracting 3D tensors. To save memory (in particular with GPU), these loops are split into smaller chunks corresponding to MEMORY_CUT. In the post-SCF methods, each loop is split by a factor of MEMORY_CUT for a total of MEMORY_CUT^2 chunks. In the RI-HFX code, each loop was split by a SQRT(MEMORY_CUT) factor for a total of MEMORY_CUT chunks.

With this PR I unified the conventions to that of the post-SCF methods, but forgot to change the `RI-HFX_H2O-32.inp` performance test accordingly. For consistency, the MEMORY_CUT keyword in that input should be changed to 3. With the current value of 9, there are 81 tiny chunks of tensors being contracted, which probably leads to large overheads in `dbt_contract`.

oschuett:
Good find! Yes, there are certainly a lot of moving parts in DBT - thanks for digging into it.
Since future GPUs will have significantly more memory, it's probably worthwhile to re-tune these heuristics.

PR #2011, opened 2022-03-17
Update DBCSR v2.3.0-rc0

alazzaro:
Including the first release with tested support of the DBCSR HIP backend.
It passes HIP on NVIDIA test (part of the CI) and HIP on AMD Mi100 (internally tested at HPE), with ROCM v4.5.2.
Thanks to @gsitaram for the improved HIP code and @mtaillefumier for the Mi100 optimized kernel parameters!


oschuett:
Oops, my #2027 caused a merge conflict, which I didn't expect. I'm sorry! Fortunately, it's easy to resolve.

Otherwise this PR looks good to me (I only checked the changes in cp2k itself).

Issue #1826, opened 2021-12-21
cp_gemm default COSMA ?


juerghutter:
Do we really want to set COSMA as default for cp_gemm?

On my AMD EPYC 7401 with 36 cores I see:

do_scalapack
 cp_gemm                            809  6.9    0.002    0.003   94.027   94.945
 cp_gemm_fm_gemm          809  7.9    0.002    0.002   94.025   94.943
 cp_fm_gemm                      809  8.9   94.023   94.941   94.023   94.941

do_cosma
 cp_gemm                            809  6.9    0.003    0.003 1016.553 1017.590
 cp_gemm_cosma               809  7.9 1016.551 1017.587 1016.551 1017.587

oschuett:
Did you use many OpenMP threads?

Our [performance test](https://dashboard.cp2k.org/archive/perf-openmp/index.html) show that COSMA does not really take  advantage of threads (which is unfortunate).

alazzaro:
Just to clarify, it is BLAS which should be threaded (which is not our case). COSMA works well (on the CPU only case) when multiple ranks are involved. However, the same should affect SCALAPACK too (at the end it is the same BLAS). The difference is really large...

oschuett:
Fair point, we should probably switch to threaded BLAS.

alazzaro:
Yes, that's what I do with OpenBLAS. Then you should make sure that you set
```
OMP_MAX_ACTIVE_LEVELS=1
```
to avoid nested parallelism.

hfp:
I am using threaded BLAS by default, it was an argument in the past to keep "POPT" pure (no threads). POPT configuration is now gone by default, and we should assume BLAS/OpenMP to consider a threshold before using threads (aka `OMP_DYNAMIC=true`).

juerghutter:
Just some more info:
This is a popt executable with 36 tasks, so no threading. There was also no accidental threading.
When running with 16 tasks, CPU time improves from 435 sec to 309 sec per iteration!

mkrack:
@juerghutter Which input do you run? I observe only minor differences between COSMA and ScaLAPACK on daint-mc for the H2O-256 benchmark using 2 nodes (72 MPI tasks, 1 OMP thread per MPI task).

juerghutter:
Here is a simple Libtest input. On my computer using popt with 36 tasks I get:

PDGEMM:   1.725 sec
COSMA :   32.719 sec

&GLOBAL
  PROGRAM_NAME TEST
[...]

PR #2018, opened 2022-03-23
OpenBLAS 0.3.19 -> 0.3.20

mkrack:
- Update arch files

### DIFF
--- a/arch/Linux-arm64-gnu-static.ssmp
+++ b/arch/Linux-arm64-gnu-static.ssmp
@@ -3,14 +3,11 @@
 # CP2K (GNU) arch file for a serial static ARM64 binary
 #
 # Tested with: GNU 11.2.0, FFTW 3.3.10, LIBINT 2.6.0, LIBBVORI 210412,
-#              LIBXC 5.1.7, OpenBLAS 0.3.19, SPGLIB 1.16.2
+#              LIBXC 5.2.2, OpenBLAS 0.3.20, SPGLIB 1.16.2
 #
-# Usage: Source this arch file and then run make as instructed after a
-#        CP2K toolchain installation has been performed successfully with
-#        > cd tools/toolchain
-#        > ./install_cp2k_toolchain.sh --mpi-mode=no --with-gcc=system --with-cmake=system
+# Usage: Source this arch file and then run make as instructed.
 #
-# Author: Matthias Krack (matthias.krack@psi.ch, PSI, December 2021)
+# Author: Matthias Krack (23.03.2022)
 #
 # \
 	if [[ "${0}" == "${BASH_SOURCE}" ]]; then \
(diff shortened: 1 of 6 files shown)


oschuett:
After this innocent update the timing of `dbm_multiply` in the [RI-HFX_32OMP](https://dashboard.cp2k.org/archive/perf-openmp/index.html) benchmark increased by 50% from [170s to 252s](https://dashboard.cp2k.org/archive/perf-openmp/RI-HFX_H2O-32_timings_32omp.txt).

This might be due to https://github.com/xianyi/OpenBLAS/pull/2775, which is currently under debate for a rollback https://github.com/xianyi/OpenBLAS/pull/3546.

I guess, we are affected because DBM calls [dgemm](https://github.com/cp2k/cp2k/blob/master/src/dbm/dbm_multiply_cpu.c#L111) (through libxsmm) within an OpenMP parallel region.

@hfp, what do you think? Should we maybe add our 2ct to that debate?

hfp:
@oschuett I am confused about what this patch actually achieves and the statements sounding somewhat obscure (as always when things get optimized ;-). What I captured is (not saying I confirm/agree) is that changing the OpenMP teamsize from parallel region to another can cause overhead. It seems OpenBLAS uses OpenMP parallelism by controlling the teamsize from region to region depending on problem size. That latter sounds somewhat wrong to me since OMP_DYNAMIC exist and one can set chunk sizes. Anyway, it's hard to oversee since this may also mix with numerical reproducible results under multiple threads. Obviously, if we got hit by a slowdown we can make a neutral statement and what for resolution (might be some revert). One issue is `dgemm` does not exist in a single-thread flavor wrt to API like "dgemm_sequential" but rather controls things under the hood.

I can see that OpenBLAS is quite important and perhaps there is a way to get this issue under control. I wonder why this happens, do we have OpenMP Nested turned on?

hfp:
Maybe [this is relevant](https://github.com/cp2k/cp2k/issues/1826#issuecomment-998838763) and if so, we can attempt to disable nested parallelism programmatically.

oschuett:
> we can attempt to disable nested parallelism programmatically.

Yes, this seems like the more robust solution.

Do you think we can just disable nested parallelism globally, ie. when cp2k starts up? Or should we be more selective and only changed the value during `dbm_multiply`?

hfp:
I checked OpenBLAS source and it seems it's not calling `omp_set_nested` (which would make things more complicated). Regarding you question, I think disabling it globally is appropriate. If we want to handle it very nicely, we can check if any environment variable related to OpenMP nested parallelism is present and then just omitting to change things... though this is a bit intriguing...  First step is to confirm nested parallelism is indeed enabled. I always thought nesting is disabled by default...

oschuett:
Ok, it wasn't the nesting. I also can't reproduce the slowdown on my AMD CPU. Furthermore, this [OpenBLAS release](https://github.com/xianyi/OpenBLAS/compare/v0.3.19...v0.3.20) had a number of fixes around AVX512. So my new theory is that we're suffering from [dynamic frequency scaling](https://en.wikipedia.org/wiki/Advanced_Vector_Extensions#Downclocking).

It seems only `dbm_multiply` with many threads is affected, which kinda makes sense because it still contains some serial code paths that eventually start to dominate. Hence, I will let this slowdown go and instead focus on fixing the threading issues in DBM.

PR #2025, opened 2022-03-26
Disable nested OpenMP parallelism

oschuett:
See #2018 for the discussion.


oschuett:
This has no affect on the performance. Turns out nested OpenMP parallelism is disabled by default.

PR #2030, opened 2022-03-28
Add wb97x2 (Anna Hehn)

fstein93:
This PR adds XC sections of the wB97X-2 functional. A few parameters can be set externally as documented by the new regtests.


fstein93:
@oschuett : I don't know why but the CI restarts the sdbg test over and over again. I have already seen that the compilation process was initiated. Do you have any idea why that happens?

oschuett:
> Do you have any idea why that happens?

Probably because of bad weather in the cloud (we're using [preemptible instances](https://cloud.google.com/compute/docs/instances/preemptible)).
I actually want to try out GCP's new spot instances, which promise significantly lower preemption rates at slightly higher prices.

PR #2031, opened 2022-03-28
Implementation of low-scaling SOS-Laplace MP2 forces

abussy:
This PR brings low-scaling SOS-Laplace MP2 gradients into CP2K. The new code is an extension of the energy code for low-scaling RPA and exploits the same machinery: sparse tensor contractions. The implementation covers the following areas:
- Double hybrid (with same-spin MP2 only)
- ADMM
- Meta (tau) functionals
- Stress tensor
- Spin polarized systems
- GPW and MME for the 2-center periodic integrals and derivatives
- Libint for all other integrals
- Option to keep the Laplace quadrature constant over a MD

The code was carefully tested with many debug runs. New regtests covering all the above have been added.


fstein93:
I have two suggestions:
1) Move the new tests into a new directory QS/regtest-ri-sos-laplace-mp2 like it is done with other methods.
2) Can you try to reduce the runtime of the regtests to under 30 s (see issue #174) if possible?

abussy:
I followed your suggestions. I did my best to reduce the execution time of the new regtest, but it might not be under 30 seconds in the end

fstein93:
Thank you. LGTM

Issue #2044, opened 2022-03-31
Wrong program name in `tools/toolchain/scripts/stage4/install_libxsmm.sh`


zc62:
https://github.com/cp2k/cp2k/blob/aa9c8d9962e810d991202283608a233847e566fc/tools/toolchain/scripts/stage4/install_libxsmm.sh#L92

The actual name should be `libxsmm_gemm_generator`, not `libxsmm_generator` as in the script.

hfp:
We should remove references of the stand-alone generator in the install script. It is legacy and it is unsupported going forward. Practically nobody uses it as all kernels are JIT-generated.

oschuett:
While we're at it we could also remove [install_libsmm.sh](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/stage4/install_libsmm.sh).

hfp:
> While we're at it we could also remove [install_libsmm.sh](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/stage4/install_libsmm.sh).

On x86-64, libsmm finally carried almost 90% kernels taken from libxsmm (if I remember correctly). Btw, our mighty LIBXSMM 2.0 will come up with ARM64 kernels incl. SVE.

Issue #1999, opened 2022-03-13
fortran_example_error about Init() when compling libint-v2.6.0-cp2k-lmax-5


alas-go:
Dear Sir/Madam。
Master branch was compiled wiht toolchain method on the OS of Ubuntu 20.4 with gcc 9.4，intel classic 2021.4.
When compiled to Libint v2.6, it report some errors about init_( ).
[libint-v2.6-compiling erro wiht intel classic 2021.4.txt](https://github.com/cp2k/cp2k/files/8239789/libint-v2.6-compiling.erro.wiht.intel.classic.2021.4.txt)

zc62:
I noticed the same error. There is a trick in the script: https://github.com/cp2k/cp2k/blob/aa9c8d9962e810d991202283608a233847e566fc/tools/toolchain/scripts/stage3/install_libint.sh#L90
It looks like you chose intelmpi and the trick was applied, and ifort was used for the linking and thus unable to find c++ libraries. But without the trick, the compilation fails too. Using c++ compiler to do the linking for the fortran example also fails.

In the end I chose to compile v2.7.0 beta 5 (https://github.com/cp2k/libint-cp2k/releases), though a pre-release, worth a try.
The cmake build method works, here is the command that I used:
```bash
mkdir build
cd build
cmake -DCMAKE_INSTALL_PREFIX=$HOME/cp2k_install -DCMAKE_C_COMPILER=icc -DCMAKE_C_FLAGS="-qopenmp -O2 -xCORE-AVX512" -DCMAKE_CXX_COMPILER=icpc -DCMAKE_CXX_FLAGS="-qopenmp -O2 -xCORE-AVX512" -DCMAKE_Fortran_COMPILER=ifort -DCMAKE_Fortran_FLAGS="-qopenmp -O2 -xCORE-AVX512" -DLIBINT2_INSTALL_LIBDIR=$HOME/cp2k_install/lib -DREQUIRE_CXX_API=OFF -DENABLE_FORTRAN=ON ..
# note that the toolchain script incorrectly uses -DCXXFLAGS=, and did not specify Fortran compilers.
CMAKE_BUILD_PARALLEL_LEVEL=8 cmake --build .
cmake --build . --target install
```
Please modify according to your needs. Then in the end choose `--with-libint=system`
(You can read my original [google groups post](https://groups.google.com/g/cp2k/c/YAXAK-RDPxo))

alas-go:
[...]

Issue #1095, opened 2020-09-17
toolchain: Move arch file generation into a separate configure.py


oschuett:
Our [arch files](https://github.com/cp2k/cp2k/tree/master/arch) files have become quite complex and it's no longer practical to write them manually.

In the toolchain we use [generate_arch_files.sh](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/generate_arch_files.sh) to generate arch files and in Spack the better part [package.py](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/cp2k/package.py) does the same.

To avoid this duplication we should have a `configure.py` script that works in both environments, e.g. by using [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/).

oschuett:
Instead of this proposal we've decided to adopt CMake.

PR #2049, opened 2022-04-01
Fix two toolchain issues and remove libsmm.

### LINKED ISSUE #2043: Typo in `tools/toolchain/scripts/stage5/install_elpa.sh`
https://github.com/cp2k/cp2k/blob/aa9c8d9962e810d991202283608a233847e566fc/tools/toolchain/scripts/stage5/install_elpa.sh#L106

Instead of
```bash
grep '\bavx512v1\b' /proc/cpuinfo 1> /dev/null && AVX512_flags+=" -mavx512v1"
```
it should be
```bash
grep '\bavx512vl\b' /proc/cpuinfo 1> /dev/null && AVX512_flags+=" -mavx512vl"
```
(Two letter `l`s were typed as number `1`s)

oschuett:
Fixes #2043 and #2044. Thanks a lot @zc62!

The code for libsmm is still available in the [dbcsr repository](https://github.com/cp2k/dbcsr/tree/develop/tools/build_libsmm).


zc62:
@oschuett Thanks for fixing these so quickly!

I have some questions regarding the toolchain. Do most developers use it to compile the libraries? Or should we compile the libraries ourselves and only use it to generate arch files, or even write our own arch files? Because I feel that it is very hard for a toolchain script to succeed in one run, and there are many possible pitfalls that the person who wrote the toolchain script could not expect because of difference in local enviroments. Also, libraries are changing so rapidly, and it is overwhelming to keep the script up-to-date with the changes in libraries. (In this case, I didn't even realize libsmm is deprecated, I chose libxsmm simply it says it is for x86-64.)

However, it appears that the toolchain is suggested to first-time users, but as a first-time user a week ago, I did not have success with it. The problem I had was mostly with Intel compilers (maybe because most developers are using gfortran? but to users, especially those who only want to run calculations, Intel compilers are common). The particular problems were, `libxc`'s configure does not add `-std=gnu99` to `icc`, thus breaking the compilation, while the toolchain script still thinks the installation of libxc is successful; second, in the arch file, `icc` was given `-std=c11`, which breaks the compilation whenever `time.h` is used, because it actually needs `-std=gnu11`.

Then I tried to build the libraries by myself (so that I can tell libxc I need `-std=gnu99`) and use the toolchain script as a guide. I only selectively compiled a few that I feel necessary, and I already found some small problems (like these two issues that I raised). I also encountered other problems that I think using the toolchain script will fail too, such as #1999 raised by another person. I learned how to compile libint 2.7 by the comments in the script
https://github.com/cp2k/cp2k/blob/9d664f589f2974eb2abbd05dbca882b4de6af279/tools/toolchain/scripts/stage3/install_libint.sh#L73-L78
but I found it is problematic too, as not specifying the fortran compiler to `ifort` will make compilation in cp2k fail, and `-DCXXFLAGS` is not really valid for cmake builds. There are numerous other problems such as `-lmkl_gf_lp64` was used regardless no matter using `gfortran` or `ifort`, no `-qopenmp` and `-nofor-main` in compiler flags in the arch files (as the generated arch files seem to have a template only for gfortran), and lacking AVX* optimization flags when building libraries (might be minor). I linked to my very naive google-groups post in #1999 to share my experience, but I believe average first-time users will encounter similar problems, and I guess there might still be a lot of tiny issues in the scripts.

oschuett:
The toolchain was originally only intended for developers and did not place great emphasis on portability. However, as more dependencies were added to CP2K, users - for lack of better alternatives - started using the toolchain as well. This has now grown into a pressing problem and we actually had a [long discussion](https://github.com/cp2k/cp2k/discussions/1989) about it just two weeks ago.

To address the problem, we're considering to replace our homegrown solution with Spack. As a first step, we need to separate the build configuration from the toolchain. To that end, @mtaillefumier is currently working on migrating our build system to cmake. If that doesn't pan out, the backup plan is #1095.

PR #2055, opened 2022-04-06
toolchain: support Intel MPI with GNU Compiler Collection

hfp:
* Note: fallback for, e.g., "mpiicpc" was wrong (using GCC already)


hfp:
I think "psmp" test would not check this case hence it is not triggered. However, the Dashboard and PSI's tests in particular can pickup this case.

PR #2064, opened 2022-04-13
Related to #2055: omit include path which can cause wrong MPI.mod with GNU Fortran

hfp:


hfp:
Note: Intel MPI is not tested by CI but per Dashboard (PSI). Intel MPI with GNU Fortran is not tested at all.

oschuett:
> Intel MPI is not tested by CI 

Yes, this is still on my todo list. In theory it should be rather straightforward thanks to the new [OneAPI containers](https://hub.docker.com/r/intel/oneapi-hpckit).

hfp:
> > Intel MPI is not tested by CI
> 
> Yes, this is still on my todo list. In theory it should be rather straightforward thanks to the new [OneAPI containers](https://hub.docker.com/r/intel/oneapi-hpckit).

Thanks!  I also made the comment to have an excuse for not triggering any parallel tests (will see what Dashboard says to this change).

PR #2065, opened 2022-04-13
Generate ARCH-files with OpenCL enabled

hfp:
* Can be combined with CUDA (and HIP/ROCm also shipping with OpenCL).
* Note: OpenCL support is currently limited to DBCSR component
* Slightly simplified build_dbcsr/Makefile.
* exts/Makefile.inc: strip flags line.


hfp:
Something is wrong with "HIP MI100 Regtest" (unrelated to the PR).

alazzaro:
The test on Mi100 doesn't work.... You can use HIP Pascal.

hfp:
> The test on Mi100 doesn't work.... You can use HIP Pascal.

Thanks!  Since I requested it once it will now run for every change to the PR, i.e., I will trigger "HIP Pascal" and merge even if "HIP MI100 Regtest" failed.

hfp:
This PR should be pretty cool as for instance ARCH-files with CUDA can now rely on OpenCL for DBCSR, i.e., `--enable-cuda` and `--enable-opencl` can be used together. Same is possible for HIP, but I did not implement it. Also, I won't donate A100 tuned parameters. However, I have included tuned parameters for P100 and V100. In addition, any GPU should work even without tuned parameters. In any case, the OpenCL BE embeds all found tuned parameters into the binary even for different GPUs.

hfp:
This also fixes, e.g., `local.psmp` (no CUDA!) which previously contained references to CUDA.

hfp:
Unfortunately, "HIP Pascal Regtest" does not pass although "HIP ROCm Build" passes. I may need help to get this fixed.

alazzaro:
> Unfortunately, "HIP Pascal Regtest" does not pass although "HIP ROCm Build" passes. I may need help to get this fixed.

Well, the error message says that it cannot find hipblas:

```
path to hipcc is /opt/rocm-4.5.2/hip/bin/hipcc
[91mERROR: (./scripts/generate_arch_files.sh) ld cannot find -lhipblas, please check if hip is installed or in system search path
ERROR: (./scripts/generate_arch_files.sh, line 156) Non-zero exit code detected.
```

So something is missing in the installed directory...

oschuett:
> Unfortunately, "HIP Pascal Regtest" does not pass although "HIP ROCm Build" passes. I may need help to get this fixed.

I took a brief look yesterday. The installers of hipBLAS and hipFFT leave behind broken symlinks for `libhipblas.so` and `libhipfft.so`.

Neither the CUDA base image nor the HIP tar-balls have changed, which suggests that those symlinks have been broken for a while. However, then I don't understand why our test kept passing until #2064. While I don't think #2064 is the culprit, someone should double-check by building [Dockerfile.test_hip_cuda_P100](https://github.com/cp2k/cp2k/blob/master/tools/docker/Dockerfile.test_hip_cuda_P100) from an earlier revision.

Currently, my best theory is that the HIP installers are not self-contained, but load some additional code from the internet.

hfp:
> > Unfortunately, "HIP Pascal Regtest" does not pass although "HIP ROCm Build" passes. I may need help to get this fixed.
> 
> I took a brief look yesterday. The installers of hipBLAS and hipFFT leave behind broken symlinks for `libhipblas.so` and `libhipfft.so`.
> 
> Neither the CUDA base image nor the HIP tar-balls have changed, which suggests that those symlinks have been broken for a while. However, then I don't understand why our test kept passing until #2064. While I don't think #2064 is the culprit, someone should double-check by building [Dockerfile.test_hip_cuda_P100](https://github.com/cp2k/cp2k/blob/master/tools/docker/Dockerfile.test_hip_cuda_P100) from an earlier revision.
> 
> Currently, my best theory is that the HIP installers are not self-contained, but load some additional code from the internet.

I first reproduced the issue locally using the recommended Dockerfile (on a system with P100). I then checked with cc2ee1bba734d0e82720bdaeedba90332d361c12, and it ran into exactly the same broken symlinks. Btw, I tried something like `check_lib -l:libhipblas.so.0.1 "hip"` and it passed that line (`generate_arch_files.sh`). However, the hipFFT installation is broken even more like the so-file is not even copied.

I believe this PR can be merged (HIP issues are unrelated).

hfp:
Generating Dockerfile(s) will be a follow-up PR. However, existing CUDA Dockerfile(s) may can simply use `--enable-opencl` (toolchain script) to use CUDA where supported except for DBCSR (OpenCL backend).

PR #2066, opened 2022-04-13
c11 vs gnu11 drops certain definitions (e.g., CLOCK_REALTIME) and function prototypes

hfp:
Affected DBM stand-alone and toolchain based builds (GCC 9.4.0).

### DIFF
--- a/src/dbm/Makefile
+++ b/src/dbm/Makefile
@@ -49,7 +49,7 @@ ALL_OBJECTS += dbm_multiply_gpu.o
 endif
 
 %.o: %.c $(ALL_HEADERS)
-	cd $(dir $<); $(CC) -c -std=gnu11 $(CFLAGS) $(notdir $<)
+	cd $(dir $<); $(CC) -c -std=c11 $(CFLAGS) $(notdir $<)
 
 dbm_miniapp.x: dbm_miniapp.o $(ALL_OBJECTS)
 	$(CC) $(CFLAGS) -o $@ $^ $(LIBS)
--- a/src/dbm/Makefile
+++ b/src/dbm/Makefile
@@ -49,7 +49,7 @@ ALL_OBJECTS += dbm_multiply_gpu.o
 endif
 
 %.o: %.c $(ALL_HEADERS)
-	cd $(dir $<); $(CC) -c -std=c11 $(CFLAGS) $(notdir $<)
+	cd $(dir $<); $(CC) -c -std=gnu11 $(CFLAGS) $(notdir $<)
 
 dbm_miniapp.x: dbm_miniapp.o $(ALL_OBJECTS)
 	$(CC) $(CFLAGS) -o $@ $^ $(LIBS)
--- a/src/dbm/dbm_miniapp.c
+++ b/src/dbm/dbm_miniapp.c
@@ -5,6 +5,9 @@
 /*  SPDX-License-Identifier: BSD-3-Clause                                     */
 /*----------------------------------------------------------------------------*/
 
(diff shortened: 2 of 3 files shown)


oschuett:
I think, instead of changing the compiler settings we should add `#define _POSIX_C_SOURCE 200809L`

oschuett:
Also, I'm wondering why this was not caught by any of our tests?

hfp:
> Also, I'm wondering why this was not caught by any of our tests?

I do not know why. I was not building GCC as part of the toolchain but rather used GCC 9.4.0 available on my system. Even without toolchain complexity, changing into `src/dbm` and typing `make` caused this particular problem.

> I think, instead of changing the compiler settings we should add `#define _POSIX_C_SOURCE 200809L`

I can investigate this and refine this PR. Another option is relying on `-D_POSIX_C_SOURCE=200809L` (rather than `#define` in some header), and to avoid issues with include order like `_POSIX_C_SOURCE` defined or not depending on a top-level header file.

hfp:
> I can investigate this and refine this PR. Another option is relying on `-D_POSIX_C_SOURCE=200809L` (rather than `#define` in some header), and to avoid issues with include order like `_POSIX_C_SOURCE` defined or not depending on a top-level header file.

On the other hand, placing this into `src/offload/offload_library.h` allows more sophisticated application like checking if `_POSIX_C_SOURCE` is already defined and a smaller version, etc.

oschuett:
> used GCC 9.4.0 available on my system.

Strange, because we are also running [tests](https://dashboard.cp2k.org/archive/gcc9/index.html) with the stock GCC 9.4.0 from Ubuntu.

> rather than #define in some header)

We should not define it in any header file, but only in the few `.c` files that actually need it. Currently, there are [two](https://github.com/cp2k/cp2k/search?q=_POSIX_C_SOURCE&type=code).

hfp:
> We should not define it in any header file, but only in the few `.c` files that actually need it. Currently, there are [two](https://github.com/cp2k/cp2k/search?q=_POSIX_C_SOURCE&type=code).

Make sense!

hfp:
> Strange, because we are also running [tests](https://dashboard.cp2k.org/archive/gcc9/index.html) with the stock GCC 9.4.0 from Ubuntu.

There are potentially more issues (which others ran into as well). For instance, building with ELPA (toolchain) causes unknown include path warning and subsequent error building with ELPA. I think structure of the ELPA-build somehow changed. I will investigate!

oschuett:
> I think structure of the ELPA-build somehow changed. I will investigate!

Thanks a lot!

Meanwhile, I'm also working on calling nvcc and hipcc with `-Wall -Wextra -Werror`.

oschuett:
Interestingly, another way to make the error disappear is to use `-fopenmp`.

hfp:
> Interestingly, another way to make the error disappear is to use `-fopenmp`.

Maybe the OpenMP header is included upfront and brings this definition?

hfp:
> > Strange, because we are also running [tests](https://dashboard.cp2k.org/archive/gcc9/index.html) with the stock GCC 9.4.0 from Ubuntu.
> 
> There are potentially more issues (which others ran into as well). For instance, building with ELPA (toolchain) causes unknown include path warning and subsequent error building with ELPA. I think structure of the ELPA-build somehow changed. I will investigate!

I think it was a spurious issue stemming from an unclean build. Note, toolchain cannot properly enable CUDA if ELPA support is enabled (and a previous toolchain-build exists with CUDA disabled).

PR #2073, opened 2022-04-15
k-points for the GW self-energy

JWilhelm:
Enables computation of band structures from GW; yet critical to do: stabilize calculations for large RI basis sets


src/hfx_ri.F
```diff
@@ -1560,6 +1560,8 @@ SUBROUTINE hfx_ri_update_ks_Pmat(qs_env, ri_data, ks_matrix, rho_ao, &
          CALL dbt_batched_contract_finalize(ri_data%t_3c_int_ctr_1(1, 1))
          CALL dbt_batched_contract_finalize(t_3c_3)
 
+         PRINT *, 'n_mem =', n_mem, 'n_mem_RI =', n_mem_RI
```
fstein93: Please remove this line.

src/rpa_gw.F
```diff
@@ -1451,6 +1498,38 @@ SUBROUTINE compute_QP_energies(vec_Sigma_c_gw, count_ev_sc_GW, gw_corr_lev_occ,
 
    END SUBROUTINE compute_QP_energies
 
+! **************************************************************************************************
+!> \brief ...
+!> \param vec_gw_energ ...
+!> \param z_value ...
+!> \param m_value ...
+! **************************************************************************************************
+   SUBROUTINE check_NaN(vec_gw_energ, z_value, m_value)
```
fstein93: What do you think about the following suggestion:
The routine checks only a single array (ALLOCATABLE attribute is not necessary here) and sets NaNs to a user-given value.

src/hfx_ri.F
JWilhelm: Thanks for spotting this, it's gone now.

src/rpa_gw.F
JWilhelm: Yes, good point, we can certainly do, code is updated.

JWilhelm:
In case additional code modifications are benefitial to get the PR accepted, I will be happy to make them.

PR #2082, opened 2022-04-26
Remove arch file

mkrack:


oschuett:
FYI: The [performance test](https://storage.googleapis.com/cp2k-ci/dashboard_perf-openmp_report.txt) failed after this PR because [I've moved it]( https://github.com/cp2k/cp2k-ci/commit/3064601ac97c89751fa40c03745b17d43e776d59)  to the T2D machine type in the hope of getting more consistent results. Presumably, it ran out of memory.

PR #2083, opened 2022-04-27
Fix HIP on CUDA build

oschuett:
A follow up to #2080.


exts/Makefile.inc
```diff
@@ -20,7 +20,7 @@ DBCSR_ACC      := $(OFFLOAD_CC)
 DBCSR_ACCFLAGS := $(OFFLOAD_FLAGS) -Xcompiler='-Wno-error'
 else ifeq (hip,$(DBCSR_USE_ACCEL))
 DBCSR_ACC      := $(CXX)
-DBCSR_ACCFLAGS := $(CXXFLAGS)
+DBCSR_ACCFLAGS := $(CXXFLAGS) -Wno-error
```
alazzaro: Maybe `-Wno-error` is too much... As far I can see, we need
```
-Wno-error=implicit-int -Wno-error=return-type
```

Issue #2084, opened 2022-04-27
Segfaults in xtb_hab_force


oschuett:
Since #2081 some regtests are [segfaulting](https://dashboard.cp2k.org/archive/minimal-sdbg/commit_9eaad38cde1a648051550b6b6c138c9243117927.txt) in the `xtb_hab_force` routine:
https://github.com/cp2k/cp2k/blob/9eaad38cde1a648051550b6b6c138c9243117927/src/xtb_matrices.F#L1232

It seems the array `neighbor_atoms` is never allocated within that routine. The code was only recently added via #1670.

juerghutter:
Fixed with #2092

PR #2092, opened 2022-05-02
Remove unused code in xTB response force

juerghutter:


juerghutter:
Fixes segfaults in xtb_hab_force (Issue #2084)

Issue #2095, opened 2022-05-03
Offset grid error for RPA computations


annahehn:
When increasing the cutoff of the WF grid for RPA computations, CP2K crashes with the error message "ortho_cxyz_to_grid: Assertion `offset + ub_cube[i] < npts_local[i]' failed." for large memory computations.
Attached are input and output files as well as corresponding slurm files from the computations performed on Daint.
[TMP_issue_cp2k.zip](https://github.com/cp2k/cp2k/files/8611320/TMP_issue_cp2k.zip)

alazzaro:
Maybe related to https://github.com/cp2k/cp2k/issues/1785 ?

mtaillefumier:
The error comes from this line
https://github.com/cp2k/cp2k/blob/701e9bc06a44956b6bbff732944475289bb57cef/src/grid/ref/grid_ref_collint.h#L1033

The problem is the following.  `collocate_pgf_product` (might also be valid for the collocate function) will branch to the orthorhombic case even if the grid is distributed and the border_width is set to zero. This scenario happens for instance here 

https://github.com/cp2k/cp2k/blob/701e9bc06a44956b6bbff732944475289bb57cef/src/mp2_eri_gpw.F#L348

for the mp2 simulations. I suspect the RPA code does something similar as well.

One way to temporeraly solve this issue is to force the orthorhombic parameter to false when the grid is distributed. But a proper fix would be to have an implementation of the orthorhombic case when the grid is distributed.

oschuett:
The grid code assumes that the region within the cutoff fits into the local part of the grid. If this is not the case then `use_subpatch=.TRUE.` has to be passed. This distinction allows for a significant speedup of the common case without `use_subpatch`. For GPW we runs [generate_qs_task_list](https://github.com/cp2k/cp2k/blob/532c618f7484ce7740d11b370a85beb5e63a9c9a/src/task_list_methods.F#L121) upfront to figure out when `use_subpatch` is needed.

So, the fix should be to add ` use_subpatch=.TRUE., subpatch_pattern=0` to those `collocate_pgf_product` and `integrate_pgf_product` calls in the MP2 code.

> But a proper fix would be to have an implementation of the orthorhombic case when the grid is distributed.

Yes, this might be useful - one should do an analysis of the expected performance gains.

There might also be some synergies with the GEEP code, where we essentially need the same functionality.

annahehn:
[...]

PR #2097, opened 2022-05-04
various fixes

mtaillefumier:
- explicitly check if the grid is distributed or not  before calling pgf_integrate / collocate
- Release memory taken by spla

### DIFF
--- a/src/grid/grid_api.F
+++ b/src/grid/grid_api.F
@@ -420,7 +420,7 @@ CONTAINS
                                  shift_local=shift_local, &
                                  border_width=border_width)
 
-      grid(1:, 1:, 1:) => rsgrid%r(:, :, :)  ! pointer assignment
+      grid(1:, 1:, 1:) => rsgrid%r(:, :, :) ! pointer assignment
 
 #if __GNUC__ >= 9
       CPASSERT(IS_CONTIGUOUS(rsgrid%desc%dh))
--- a/src/mp2_eri_gpw.F
+++ b/src/mp2_eri_gpw.F
@@ -191,7 +191,7 @@ CONTAINS
       INTEGER, ALLOCATABLE, DIMENSION(:, :)              :: offset_2d
       INTEGER, DIMENSION(:), POINTER                     :: la_max, la_min, npgfa, nsgfa
       INTEGER, DIMENSION(:, :), POINTER                  :: first_sgfa
-      LOGICAL                                            :: map_it_here
+      LOGICAL                                            :: map_it_here, use_subpatch
       REAL(KIND=dp)                                      :: cutoff_old, radius, relative_cutoff_old
       REAL(KIND=dp), ALLOCATABLE, DIMENSION(:)           :: e_cutoff_old, wf_vector
       REAL(KIND=dp), ALLOCATABLE, DIMENSION(:, :)        :: I_ab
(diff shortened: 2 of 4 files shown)


oschuett:
Thanks for picking this up! While your patch would most certainly work, it might have performance implications elsewhere. Hence, I'd prefer if you instead added `use_subpatch=.TRUE., subpatch_pattern=0` to those `collocate_pgf_product` and `integrate_pgf_product` calls in the MP2 code.

See also https://github.com/cp2k/cp2k/issues/2095#issuecomment-1117105588.

mtaillefumier:
- The mp2 code could probably benefit from tasklist if the number of integrals is large. 
- When the grid is orthorhombic and distributed, we can still use the generic implementation but skip the calculation of the exponential exp_ij, exp_ik, exp_jk as they always will give 1.

oschuett:
> * The mp2 code could probably benefit from tasklist if the number of integrals is large.

I agree, in particular because otherwise it won't run on the GPU.

> * When the grid is orthorhombic and distributed, we can still use the generic implementation but skip the calculation of the exponential exp_ij, exp_ik, exp_jk as they always will give 1.

Yes, `rab(:) = 0` is indeed a common case, which is e.g. also needed for GEEP. However, the orthorhombic code is also very efficient in computing the loop bounds from the given cutoff radius, which again is not needed by every use-case.

While it seems tempting to specialize the grid code, this would leave us potentially with 
(collocate/integrate) x (ortho/generic) x (spherical bounds) x (rab=0) x (cpu/cuda/hip) = 48 different code paths.

So, we should first do an in-depth analysis of the expected performance gains before adding this much complexity.

mtaillefumier:
Actually the cpu/hip backends support both ortho/non-ortho distributed grids and will use this to speed up computations. Even the ref implementation can use this information without modifying much of the code. 

a simple 
```

static inline void __attribute__((always_inline))
general_ci_to_grid(const int lp, const int jg, const int kg, const int ismin,
                   const int ismax, const int npts_local[3],
                   const int index_min[3], const int index_max[3],
                   const int map_i[], const int sections_i[],
                   const double gp[3], const int k, const int j,
                   const double exp_ij[], const double exp_jk[],
                   const double exp_ki[], GRID_CONST_WHEN_COLLOCATE double *ci,
                   GRID_CONST_WHEN_INTEGRATE double *grid) {

  const int base = kg * npts_local[1] * npts_local[0] + jg * npts_local[0];

  // AVX instructions can only load/store from evenly spaced memory locations.
  // Since the cube can wrap around due to the grid's periodicity,
  // the inner loop runs over sections with homogeneous cube to grid mapping.
  for (int istart = ismin; istart <= ismax; istart++) {
    const int istop = imin(ismax, istart + sections_i[istart - index_min[0]]);
    if (map_i[istart - index_min[0]] < 0) {
      istart = istop; // skip over out-of-bounds indicies
      continue;
    }

    const int cube2grid = map_i[istart - index_min[0]] - istart;
    for (int i = istart; i <= istop; i++) {
      const int ig = i + cube2grid;
      const double di = i - gp[0];

      const int stride_i = index_max[0] - index_min[0] + 1;
      const int stride_j = index_max[1] - index_min[1] + 1;
      const int stride_k = index_max[2] - index_min[2] + 1;
      const int idx_ij = (j - index_min[1]) * stride_i + i - index_min[0];
      const int idx_jk = (k - index_min[2]) * stride_j + j - index_min[1];
      const int idx_ki = (i - index_min[0]) * stride_k + k - index_min[2];

      // Mathieu's trick: Calculate 3D Gaussian from three precomputed 2D tables
      //
      // r   =  (i-gp[0])*dh[0,:] + (j-gp[1])*dh[1,:] + (k-gp[2])*dh[2,:]
      //     =  a                 + b                 + c
      //
      // r**2  =  (a + b + c)**2  =  a**2 + b**2 + c**2 + 2ab + 2bc + 2ca
      //
      // exp(-r**2)  =  exp(-a(a+2b)) * exp(-b*(b+2c)) * exp(-c*(c+2a))
      //
      const double gaussian = ortho ? (1):(exp_ij[idx_ij] * exp_jk[idx_jk] * exp_ki[idx_ki]);
```
and 

```
static inline void
precompute_exp_table(const int ortho, const int idir, const int jdir, const int index_min[3],
                     const int index_max[3], const double zetp,
                     const double dh[3][3], const double gp[3],
                     double exp_table[]) {

  if (ortho)
    return;

```
would speed up things a lot. Of course it is possible to refine it a bit depending on which direction is orthogonal to the other.... I am not suggesting it should be done just that it is not as complex as it seems.

mtaillefumier:
The two commits should be squashed

mtaillefumier:
thanks

Issue #2062, opened 2022-04-12
Should we remove GCC from the Toolchain?


oschuett:
I'm contemplating to remove GCC from our Toolchain. It takes 10 minutes to build in the CI and adds 655MB to our Docker images.

When GCC was originally added to the Toolchain we were still suffering from numerous compiler bugs around modern Fortran. Furthermore, we were experimenting with Thread Sanitizer. Both is not the case anymore.

So, it seems nowadays we could just always use the GCC provided by the Linux distribution.

mkrack:
That could a premature action. The generic version on our cluster is still gcc 4.8.5 (RedHat 7.9). Due to my experience, the toolchain build of gcc (if requested) does not trouble too much. Most resources are consumed by the libint build.

hfp:
I believe it is a good idea in particular since 4.8.x phases out by already lost support for RHEL/CentOS 7 bits. We should instead check if the toolchain picks up CXX, CC, and FC (or maybe related MPI flavors) correctly.

Though, we never completed the "latest GCC" effort by building GNU Binutils as well. I especially vote for it since we seem to build GCC in two phases ("bootstrap"), which is the default. The latter is expected to likely double the compile-time for GCC, or in turn we could bring it down to 5 minutes if we wanted to keep this option.

oschuett:
> The generic version on our cluster is still gcc 4.8.5 (RedHat 7.9). 

RedHat 7.9 will be end of life in June 2024, which suggests we should wait two more years.
On the other hand, GCC 7 came out 5 years ago and most clusters should have it by now.

So it's a trade-off between taking the hit on the CI time and Docker images or degrading support for the long tail.

> I especially vote for it since we seem to build GCC in two phases ("bootstrap")

No, we actually [disable bootstraping](https://github.com/cp2k/cp2k/blob/cc2ee1bba734d0e82720bdaeedba90332d361c12/tools/toolchain/scripts/stage0/install_gcc.sh#L72).

> Most resources are consumed by the libint build.

Yes, you are [right](https://storage.googleapis.com/cp2k-ci/run-cp2k-generic-psmp-2bf80dfb_report.txt). Libint takes 12.5 minutes while GCC takes only 8.4 minutes. Unfortunately, at the moment there is nothing we can do about libint.

hfp:
[...]

PR #2107, opened 2022-05-11
Docker: Upgrade to Ubuntu 22.04

oschuett:
I've generated new Dockerfiles for GCC 11 and GCC 12.

However, I don't think we need to add them to the Dashboard yet, because GCC 11 will be the used by most other tests and GCC 12 will be tested via [Dockerfile.test_openmpi-psmp](https://github.com/cp2k/cp2k/blob/13fcfd89a6293a40c706155cf1bed319c31105ac/tools/docker/Dockerfile.test_openmpi-psmp#L26) once #2106 is merged.


alazzaro:
@oschuett are you saying that we will have GCC 11 test on ubuntu around in the dashboard? which test is currently using GCC 11?

oschuett:
GCC 11 is now our "default" compiler and will be used by all the `Current Toolchain` tests.

Before today only the [OpenMPI test](https://dashboard.cp2k.org/archive/openmpi-psmp/index.html) was using GCC 11. 

I'd like to remove `--with-gcc=install` in the not-so-distant future (#2062). Until then we're using the OpenMPI test to [cover it](https://github.com/cp2k/cp2k/blob/895ec07cf52c4a7d7647ca49e6642962e81fa4b6/tools/docker/generate_dockerfiles.py#L30).

PR #2118, opened 2022-05-18
Docker: Fix Gromacs and AiiDA tests

oschuett:


oschuett:
The AiiDA tests are still failing because of https://github.com/aiidateam/aiida-cp2k/issues/158.

Issue #2125, opened 2022-05-22
make psmp test and fail to the tests because of QS/regtest-rma


scott-5:
At the beginning of compiling cp2k **v9.1** (**gcc: 9.3.0**/system: **CentOS7**), I come across testing slowly, finally it has been solved by `make -j1 ARCH=local VERSION="psmp" test TESTOPTS+="--maxtasks 4"`  #2074 without `warning: jobserver unavailable`, and all tests pass. 
But I found three tests are skipped (TMC/regtest_ana_on_the_fly, **QS/regtest-rma** and **QS/regtest-rma-3D**) like `Skipping QS/regtest-rma-3D because its requirements are not satisfied`, so I add `TESTOPTS+="--mpiranks 4"` and `--mpiranks 8` (as `make -j1 ARCH=local VERSION="psmp" test TESTOPTS+="--mpiranks 4 --maxtasks 8 --restrictdir QS/regtest-rma"`) to test **QS/regtest-rma** and **QS/regtest-rma-3D**. Then two both FAILED (TMC/regtest_ana_on_the_fly passed).
There are reports:
```
----------------------------- Settings ---------------------------------
MPI ranks:      4
OpenMP threads: 2
GPU devices:    0
Workers:        1
Timeout [s]:    400
Work base dir:  ~/cp2k-9.1/regtesting/TEST-local-psmp-2022-05-22_14-58-21
MPI exec:       ['mpiexec']
Keepalive:      False
Debug:          False
ARCH:           local
VERSION:        psmp
Flags:          omp,libint,fftw3,libxc,pexsi,elpa,parallel,mpi3,scalapack,cosma,quip,xsmm,plumed2,spglib,mkl,sirius,libvori,libbqb
------------------------------------------------------------------------
    H2O-32-dftb-ls-2_mult.inp                                               TIMED OUT ( 406.38 sec)
    H2O-32-dftb-ls-2.inp                                                 RUNTIME FAIL (   3.60 sec)
    H2O-OT-ASPC-1.inp                                                    RUNTIME FAIL (   3.62 sec)
    H2O-OT-ASPC-1_clusters.inp                                           WRONG RESULT (   1.72 sec)
    H2-big-nimages.inp                                                   RUNTIME FAIL (   3.62 sec)
    H2-big-nimages_clusters.inp                                          RUNTIME FAIL (   3.64 sec)
    H2O_grad_gpw.inp                                                     RUNTIME FAIL (   3.36 sec)
[...]

PR #2129, opened 2022-05-24
Enable RMA tests for any number of ranks

alazzaro:


alazzaro:
@oschuett Triggered by the discussion at https://github.com/cp2k/cp2k/issues/2125, I'm now enabling the RMA tests for any number of ranks. It passes the MPICH, but I would like to try the OpenMPI too (historically OpenMPI is not good with RMA calls). Could you tell me why the OpenMPI stops with:
```
<<< /opt/cp2k/regtesting/TEST-local-psmp-2022-05-24_08-38-36/QS/regtest-stda-force (53 of 313) done in 57.67 sec
make[3]: *** [/opt/cp2k/Makefile:253: test] Killed
make[2]: *** [/opt/cp2k/Makefile:148: test] Error 2
make[1]: *** [/opt/cp2k/Makefile:125: psmp] Error 2
make: *** [Makefile:120: test] Error 2
Removing intermediate container 1dcfa48a7f43
```

?

oschuett:
It probably ran out of memory or `/deb/shm`. I'll take a closer look this evening.

oschuett:
I ran the OpenMPI test locally via:
```
cd tools/docker; docker build -f Dockerfile.test_openmpi-psmp --shm-size=1g  ../../
```
It seems the RMA tests get stuck and just gobble up memory until they timeout. In my case the container used up to 70GB while the CI's limit is 22GB and normal consumption is around 12GB. Hence, there seems to be something seriously wrong with OpenMPI's RMA implementation.

AFAIK, RMA never really caught on because of poor vendor support. So, I'm wondering if we should maybe simply drop it?

alazzaro:
> 
> AFAIK, RMA never really caught on because of poor vendor support. So, I'm wondering if we should maybe simply drop it?

No, this is not an option. RMA algo runs quite well on Cray-mpi (MPICH based) with quite good speed-ups and in general with MPICH (and Intel MPI). OpenMPI needs to be properly compiled and requires hardware support from the network. 
What I will do is to [avoid RMA](https://github.com/cp2k/dbcsr/commit/029cf0e6a0e3b7f4e6ed335482725a658f5d83ed) in DBCSR when OpenMPI is used.

PR #2131, opened 2022-05-25
RPA| Low-scaling gradients

abussy:
Forces and stress tensors were implemented for the low-scaling RPA method. Most of the routines written for the low-scaling Laplace-SOS-MP2 gradients (PR #2031) were reused, as the equation differ only marginally. This was thoroughly tested with debug runs, and regtests were added. As for SOS-MP2, most (all ?) necessary features (ADMM, double hybrids, tau functionals, etc.) are covered.


fstein93:
I have found that the directory QS/regtest-rpa-cubic-scaling was accidentally removed in #1670 . Can you check that, please?

abussy:
Indeed. The k-point and RI-HFX tests are currently failing. I know why the latter does not work, but I am not sure for the former. I'll figure it out and add the directory back as part of this PR.

abussy:
The missing cubic-scaling RPA regtests are back and have been fixed. Moreover, the previously missing RPA@PBE and RPA@PBE0 (and RPA@ any kin of GGA, meta-GGA or hybrid for that matter) were implemented.

Issue #2130, opened 2022-05-25
Getting Hirshfeld / Mulliken Charges from the LIBCP2K Interface


Pibemanden:
As the title suggests I would like to be able to get the Hirshfeld or Mulliken charges using the general get results interface provided in `src/start/libcp2k.F`.
As far as I can tell implementing something that would enable one to be able to use the following `call get_result_r1(env, '[HIRSHFELD-CHARGE]', natoms, charges, ierr)` would be rather straight forward using the methods provided in `src/common/cp_result_methods.F`.

oschuett:
Looks like a good plan. Feel free to submit a pull request.

PR #2134, opened 2022-05-27
Added Mulliken and Hirshfeld charges to the results framework

Pibemanden:
As mentioned in #2130 I have added the` [MULLIKEN-CHARGES]` and `[HIRSHFELD-CHARGES]` to the fortran/C++ interface for CP2K.

### DIFF
--- a/src/hirshfeld_methods.F
+++ b/src/hirshfeld_methods.F
@@ -20,6 +20,9 @@ MODULE hirshfeld_methods
                                               pbc
    USE cp_control_types,                ONLY: dft_control_type
    USE cp_para_types,                   ONLY: cp_para_env_type
+   USE cp_result_methods,               ONLY: cp_results_erase,&
+                                              put_results
+   USE cp_result_types,                 ONLY: cp_result_type
    USE cp_units,                        ONLY: cp_unit_to_cp2k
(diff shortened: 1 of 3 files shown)


oschuett:
Thanks for contributing!

Could you please once run `make pretty` to fix the formatting. Furthermore, you could extend [libcp2k_unittest.c](https://github.com/cp2k/cp2k/blob/master/src/start/libcp2k_unittest.c) to cover your use case - if you like.

Pibemanden:
I tried running `make pretty` but it didn't change anything in the files I have altered. It only detected some errors in files which I haven't touched, but not changes there either according to git. So I am wondering what is causing the errors being reported by your build system.

oschuett:
> So I am wondering what is causing the errors being reported by your build system.

Can you share what kind of errors it reports? Generally, we test quite rigorously that everything is prettified. One thing you could try is updating your git submodules: ```git submodule update --recursive```.

Pibemanden:
Sorry I forgot to clone with the recursive tag before running `make pretty`, seems like everything might be working now though.

oschuett:
Yeah, these sub-modules are annoying. Hopefully, we can ditch them someday. Glad it worked out after all.

PR #2143, opened 2022-06-02
WFC: Some refactorings

fstein93:


oschuett:
This PR removed `QS/regtest-ri-rpa` from `TEST_DIRS`. While it was previously erroneously listed twice it's now missing entirely.

PR #2146, opened 2022-06-05
Printout for INITCOND for NEWTONX

annahehn:
Writing of vibrational modes for NewtonX interface


oschuett:
It would be great if we could setup a test for Newton-X, like we have to ASE, AiiDA, i-PI, and Gromacs. Unfortunately, there seems to be no way to download Newton-X without filling [this form](https://newtonx.org/download/).

annahehn:
Thank you Ole, I agree. I will work on this.
________________________________

PR #2160, opened 2022-06-08
Update DBCSR

alazzaro:


hfp:
Re:updated DBCSR, what's needed/prerequisite to leverage https://github.com/cp2k/dbcsr/commit/445fbfcfbdc4b3d41ea9bcd285209bf460f3d40f? I tried for instance [bench_dftb.inp](https://github.com/cp2k/cp2k/blob/master/benchmarks/QS_single_node/bench_dftb.inp) and it is exactly performing the same number of GPU up-/downloads as before.

alazzaro:
> Re:updated DBCSR, what's needed/prerequisite to leverage [cp2k/dbcsr@445fbfc](https://github.com/cp2k/dbcsr/commit/445fbfcfbdc4b3d41ea9bcd285209bf460f3d40f)? I tried for instance [bench_dftb.inp](https://github.com/cp2k/cp2k/blob/master/benchmarks/QS_single_node/bench_dftb.inp) and it is exactly performing the same number of GPU up-/downloads as before.

The condition is in [this line](https://github.com/cp2k/dbcsr/blob/939f6e13044fc0b2e9fba2ce81e4692091e34235/src/mm/dbcsr_mm_accdrv.F#L352). So, basically, if you do C+=AB (i.e. `keep_product_data==.TRUE.`) nothing will change. The optimization is only for C=AB, which is the case of the H2O-dft-ls. Could you confirm you are in such a case?

alazzaro:
> Re:updated DBCSR, what's needed/prerequisite to leverage [cp2k/dbcsr@445fbfc](https://github.com/cp2k/dbcsr/commit/445fbfcfbdc4b3d41ea9bcd285209bf460f3d40f)? I tried for instance [bench_dftb.inp](https://github.com/cp2k/cp2k/blob/master/benchmarks/QS_single_node/bench_dftb.inp) and it is exactly performing the same number of GPU up-/downloads as before.

The condition is in [this line](https://github.com/cp2k/dbcsr/blob/939f6e13044fc0b2e9fba2ce81e4692091e34235/src/mm/dbcsr_mm_accdrv.F#L352). So, basically, if you do C+=A*B (i.e. `keep_product_data==.TRUE.`) nothing will change. The optimzation is only for C=A*B, which is the case of the H2O-dft-ls.

alazzaro:
@hfp I've just realized I gave the answer to a different question in my previous reply... 
No, the optimization doesn't change the number of CPU-GPU data movement, it only avoids the final C-reduction on the CPU. It turns out that when keep_product_data is .FALSE., we are basically summing with a buffer full of zeros...

hfp:
> @hfp I've just realized I gave the answer to a different question in my previous reply... No, the optimization doesn't change the number of CPU-GPU data movement, it only avoids the final C-reduction on the CPU. It turns out that when keep_product_data is .FALSE., we are basically summing with a buffer full of zeros...

@alazzaro Thank you, that's good to know!

Issue #1379, opened 2021-02-26
Rewrite m_memory with POSIX getrusage instead of reading /proc


dev-zero:
`m_memory` is currently not thread-safe (for example when sampling memory inside a bigger OMP loop) and does not support macOS. While we could make it threadsafe, it might be a good opportunity to switch over to the POSIX API `getrusage` (same for the more extended one).

This can happen in 2 ways:

* remodel the `timeval` and `rusage` structs in Fortran and call the function directly. Advantage: no additional wrapper. Disadvantage: might break if the structs don't match what the system provides (unsure up to which part POSIX defines it) or if the datatypes don't match exactly.
* write a small wrapper in C which extracts the necessary part and returns them in a struct we define. Advantage: the C side will transparently handle order of fields, type conversion, verifies that we have a proper definition. Disadvantage: additional complexity and possible source of error (on the other hand, breakages are less likely to go unnoticed).

This is somewhat orthogonal to what #630 proposes

dev-zero:
> I agree that we should stick to the POSIX standard whenever possible. I'm a bit hesitant to give up on #630, because knowing the true peak memory could be quite useful. 

They are not mutually exclusive (as I wrote, these 2 issues are somewhat orthogonal).
In fact, the `VmHWM` is part of the set of data returned by `getrusage`:

```
               long   ru_maxrss;        /* maximum resident set size */
               long   ru_ixrss;         /* integral shared memory size */
               long   ru_idrss;         /* integral unshared data size */
               long   ru_isrss;         /* integral unshared stack size */```
```
 
> For the record, it would be rather easy to make the current implementation thread-safe: If we use C then we don't have to worry about hard coded [unit numbers](https://github.com/cp2k/cp2k/blob/master/src/base/machine.F#L359). And [m_memory_max](https://github.com/cp2k/cp2k/blob/master/src/base/machine.F#L383) won't be needed anymore if we do #630.

Hardcoded units is not an issue anymore with F2008 since we could use `newunit`.
The other way would be to make it an OMP critical zone.

> In any case, I'd prefer to have a C wrapper. The added complexity is a small price to pay for the gained robustness.

Ok.

oschuett:
> In fact, the VmHWM is part of the set of data returned by getrusage.

Sorry, I missed that. In this case there seems to be no downside to `getrusage`.

> Hardcoded units is not an issue anymore with F2008 since we could use newunit.

Neat! I didn't know about `newunit` yet.

Issue #630, opened 2019-11-16
Use VmHWM from /proc/self/status to measure max_memory.


oschuett:
Currently, [m_max_memory](https://github.com/cp2k/cp2k/blob/master/src/base/machine.F#L93) is determined by taking the maximum across all calls to `m_memory()`. Therefore, it's noisy and we've got several calls to `m_memory()` [scatter](https://github.com/cp2k/cp2k/blob/20fe0099bce4444784cb4990292e1d9e993a51dd/src/mp2_ri_gpw.F#L417-L418) throughout the code base.

Instead we should read `VmHWM` from [/proc/self/status](http://man7.org/linux/man-pages/man5/proc.5.html), which directly gives us the _Peak resident set size ("high water mark")_.

PR #2185, opened 2022-07-05
Optionally print peak memory usage after each MD/Opt step

abussy:
A new keyword, `MEMORY_INFO`,  was introduced in `&MOTION%PRINT`. If set to .TRUE. (its default value), peak memory usage is sampled and printed at each MD/Optimization step. This should help identify the presence of memory leaks when jobs crash randomly.


oschuett:
Btw, if you want to to know the exact peak memory usage then checkout #630 and #1379.

abussy:
These are not concepts I am familiar with. But if any of those end up being implemented, I'd be happy to use them here too.

PR #2189, opened 2022-07-08
Fix 2nd order GAPW properties

fstein93:
This PR fixes incorrect 2nd order properties of GAPW calculations introduced in #1974 .

### DIFF
--- a/src/xc/xc.F
+++ b/src/xc/xc.F
@@ -2732,6 +2732,16 @@ CONTAINS
          END IF
       END IF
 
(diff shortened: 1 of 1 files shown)


mkrack:
Is this fix relevant for the current release?

fstein93:
I think yes because it fixes the code in the case when 2nd order properties are required from GAPW calculations like TDDFT.

fstein93:
If there are no complaints, I will merge this PR at the end of the day.

PR #2198, opened 2022-07-15
Bump COSMA to version 2.6.0

fstein93:
Can one of the admins add the new tarball (https://github.com/eth-cscs/COSMA/releases/download/v2.6.0/COSMA-v2.6.0.tar.gz) ?

Questions/Comments for the GPU-specialists:
1. The new release enables GPU2GPU communication which requires to set a new variable -DCOSMA_WITH_GPU_AWARE_MPI for COSMA. As far as I know, some people are working on this kind of transfers in other parts of the code. Would it be better to have an additional flag for the whole toolchain or for COSMA only? Probably, we cannot hope that all MPI implementations are GPU aware, can we?
2. Related to question 1, COSMA can use the NCCL and RCCL libraries for enhanced communication.
3. There is now support for AMD GPUs.

### DIFF
--- a/tools/toolchain/scripts/stage4/install_cosma.sh
+++ b/tools/toolchain/scripts/stage4/install_cosma.sh
@@ -161,6 +161,13 @@ export COSMA_VERSION=${cosma_ver}
 export CP_LIBS="IF_MPI(${COSMA_LIBS}|) \${CP_LIBS}"
 EOF
   cat "${BUILDDIR}/setup_cosma" >> $SETUPFILE
+
+  cat << EOF >> ${INSTALLDIR}/lsan.supp
+# leaks related to COSMA (probably, only the last one is actually needed)
+leak:cosma::communicator::communicator
+leak:cosma::cosma_context<double>::register_state
+leak:cosma::pxgemm<double>
+EOF
 fi
 
 load "${BUILDDIR}/setup_cosma"
--- a/tools/toolchain/scripts/stage4/install_cosma.sh
+++ b/tools/toolchain/scripts/stage4/install_cosma.sh
@@ -6,8 +6,8 @@
 [ "${BASH_SOURCE[0]}" ] && SCRIPT_NAME="${BASH_SOURCE[0]}" || SCRIPT_NAME=$0
 SCRIPT_DIR="$(cd "$(dirname "$SCRIPT_NAME")/.." && pwd -P)"
 
-cosma_ver="2.5.1"
-cosma_sha256="085b7787597374244bbb1eb89bc69bf58c35f6c85be805e881e1c0b25166c3ce"
+cosma_ver="2.6.0"
+cosma_sha256="88405e382b37d247b67403dc3386758b1d2f133a46a902b4be08ffbc43397539"
 source "${SCRIPT_DIR}"/common_vars.sh
 source "${SCRIPT_DIR}"/tool_kit.sh
 source "${SCRIPT_DIR}"/signal_trap.sh


oschuett:
Since the tar-ball is hosted on Github there is no need to mirror it. We only do this for other sites. If Github goes down our CI is dead in the water anyways.

Your question about using GPU-aware-MPI vs NCCL/RCCL directly is something I'm also currently pondering.
From what I can tell there is no benefit of using GPU-aware-MPI but the big downside that the installation becomes more complex.

Hence, I'd suggest we stick to NCCL/RCCL and enable it whenever a library supports it.

That being said, I've actually zero experience with this. So, maybe @kabicm or @alazzaro could chime in?

fstein93:
Apparently, there are some leaks. But, it is impossible to say where exactly.

oschuett:
It's the `full_comm_` that gets created at [communicator.cpp:44](https://github.com/eth-cscs/COSMA/blob/922e3002e255d6116e3ced59bfbceaa753eda9cd/src/cosma/communicator.cpp#L44).

We essentially have three options:
1. Patch it ourselves
2. Suppressed it ([example](https://github.com/cp2k/cp2k/blob/c2e362352c151e91c01e25a203cc54b90ddc4f3e/tools/toolchain/scripts/stage5/install_elpa.sh#L206-L209))
3. Wait for a new release

kabicm:
@oschuett Regarding the leaks -- one big difference in this version is that all MPI comms are now cached in the global context and reused in the subsequent runs. However, this global context is destroyed when main goes out of scope. By then, MPI has already been finalized, which is why we skip freeing communicators in that case, because they will anyway be freed at that point:
https://github.com/eth-cscs/COSMA/blob/922e3002e255d6116e3ced59bfbceaa753eda9cd/src/cosma/communicator.cpp#L189

Could this be the reason for current leaks?

oschuett:
Yes, this sounds like a perfect explanation. Apparently, `MPI_Finalize` does not release *all* MPI resources. Maybe you could explicitly free the cached communicators before calling `MPI_Finalize`?

fstein93:
If that is the case, I am in favor of suppressing it.

kabicm:
Registering a callback function that would be executed before MPI_Finalize is a bit tricky in MPI. MPI objects from the global context would have to be registered as attributes to MPI_COMM_SELF and also in the global context destructor. Then, each of these destructors would have to check if these objects have already been freed by the other destructor. We had this before, but it complicates the code logic too much.

For this reason, I am also in favor of suppressing it.

oschuett:
> `leak:MPI_Comm_create_group`

Please use `leak:cosma::communicator::communicator` instead, which is more specific.

kabicm:
Btw, on Piz Daint, for NCCL to work, I had to set: `export MPICH_RDMA_ENABLED_CUDA=1`. Not sure if it's necessary on your ci/cd infrastructure.

oschuett:
>  I had to set: export MPICH_RDMA_ENABLED_CUDA=1

Was this for using GPU-aware-MPI (which uses NCCL under the hood) or also for using NCCL directly (which claims to work independently of MPI)?

kabicm:
Let me check if there is a leak in the `register_state` function!

kabicm:
I first thought that this line with `unique_ptr` reassignment was causing a leak:
https://github.com/eth-cscs/COSMA/blob/922e3002e255d6116e3ced59bfbceaa753eda9cd/src/cosma/context.cpp#L99

However, as discussed with @simonpintarelli, this should still be fine.

@oschuett I needed this variable defined on daint even when just using NCCL and not the GPU-aware MPI.

fstein93:
@kabicm We have still found another spot at https://github.com/eth-cscs/COSMA/blob/922e3002e255d6116e3ced59bfbceaa753eda9cd/src/cosma/cosma_pxgemm.cpp#L274 . Supposedly, this communicator will be duplicated again and deleted at the very end?

kabicm:
The reordered communicator is freed here: https://github.com/eth-cscs/COSMA/blob/922e3002e255d6116e3ced59bfbceaa753eda9cd/src/cosma/cosma_pxgemm.cpp#L390

It will be duplicated as full_comm here https://github.com/eth-cscs/COSMA/blob/922e3002e255d6116e3ced59bfbceaa753eda9cd/src/cosma/communicator.cpp#L53

The duplicated comm (i.e. full comm) is cached and will be freed in one of the two cases:
- in the next pxgemm invocation, it will be reused if the pxgemm parameters are the same, or destroyed if pxgemm params are different, as part of the context destructor: https://github.com/eth-cscs/COSMA/blob/922e3002e255d6116e3ced59bfbceaa753eda9cd/src/cosma/context.cpp#L99
- the main goes out of scope, the context, which is a global singleton class, gets destroyed. In this case however, MPI_Finalize will be called before the context goes out of the scope and this communicator will not be destroyed by the context destructor: https://github.com/eth-cscs/COSMA/blob/922e3002e255d6116e3ced59bfbceaa753eda9cd/src/cosma/communicator.cpp#L189
But this is already the end of the execution and it will be cleaned up by the operating system.

kabicm:
Let me check one more thing!

kabicm:
I still can't fine any issues.

fstein93:
@kabicm Thank you for your extensive help!

I will open issues adressing the different points from my entry post for the GPU specialists to improve the GPU support of COSMA in CP2K.

kabicm:
Thank you @fstein93 and @oschuett for working on integrating the latest COSMA version. I don't think I was too helpful this time, but let me know if there are any problems with the gpu version, especially with NCCL :)

alazzaro:
> Since the tar-ball is hosted on Github there is no need to mirror it. We only do this for other sites. If Github goes down our CI is dead in the water anyways.
> 
> Your question about using GPU-aware-MPI vs NCCL/RCCL directly is something I'm also currently pondering. From what I can tell there is no benefit of using GPU-aware-MPI but the big downside that the installation becomes more complex.
> 
> Hence, I'd suggest we stick to NCCL/RCCL and enable it whenever a library supports it.
> 
> That being said, I've actually zero experience with this. So, maybe @kabicm or @alazzaro could chime in?

So, GPU-aware-MPI is something more flexible and standard than NCCL/RCCL. However, I agree, it requires support by the hardware. Both MPICH and OpenMPI enable it by default whenever there is a CUDA installation and then there are some environment variables to enable/disable it. 
In any case, GPU-aware MPI does a good job only when there is hardware support for it. For instance, LUMI GPUs are directly attached to the network cards, so any MPI communication does not involve any CPU...

PR #2216, opened 2022-07-21
RI-RPA| Allow for XC correction in ADMM RI-RPA

abussy:
This PR is a general upgrade of RPA when ADMM is used. There are 3 main aspects:

1) Previously, any usage of ADMM when doing RPA calculations (switched on with `RI_RPA%ADMM`) ignored the GGA exchange correction. This simply meant doing HFX in a smaller basis, which is a rather strong approximation. This is now fixed, also for GW.

2) Because ADMM operations in RPA rely on the `qs_env%admm_env`, it was not possible to do ADMM RPA@PBE, which is arguably the most efficient way of doing RPA. This is now possible (with XC correction) thanks to a small refactoring of general ADMM. This also means that RPA specific `xc_section_aux` and `xc_section_primary` sections are created and stored.

3) The  `RI_RPA%HF` input section governing EXX calculations is, in most cases, identical to the `DFT%XC%HF` section with the exception of the FRACTION keyword. Therefore, efficiency and memory usage can be improved by reusing the HFX data from the `qs_env` and rescaling. This is now automatically done if the two HF section only differ by the FRACTION keyword (this is also true without ADMM).

New regtests for ADMM RPA@PBE were added and results for ADMM RPA@PBE0 were updated.


oschuett:
After this PR the [CH3_md_periodic_admm.inp](https://github.com/cp2k/cp2k/blob/master/tests/QS/regtest-rpa-cubic-scaling-2/CH3_md_periodic_admm.inp) test went from 49 to 288 seconds. Ideally, regtests shouldn't run longer than 30 seconds.

abussy:
#2227 this should fix it

oschuett:
Thanks @abussy !

PR #2223, opened 2022-07-25
Enable sirius support of dft+U

mtaillefumier:
- enable SIRIUS LDA+U support (+V term needs modifications in the cp2k input system)
- updated sirius

### DIFF
--- a/src/sirius_interface.F
+++ b/src/sirius_interface.F
@@ -130,17 +130,18 @@ CONTAINS
       CHARACTER(len=default_string_length)               :: label
       INTEGER                                            :: i, iatom, ibeta, ifun, ikind, iwf, j, l, &
                                                             n, natom, nbeta, nkind, nmesh, &
-                                                            num_mag_dims, sirius_mpi_comm, vdw_func
+                                                            num_mag_dims, sirius_mpi_comm, vdw_func, nu, lu
       INTEGER, DIMENSION(:), POINTER                     :: mpi_grid_dims
       INTEGER(KIND=C_INT), DIMENSION(3)                  :: k_grid, k_shift
       INTEGER, DIMENSION(:), POINTER                     :: kk
       LOGICAL                                            :: up, use_ref_cell
-      LOGICAL(4)                                         :: use_so, use_symmetry
+      LOGICAL(4)                                         :: use_so, use_symmetry, dft_plus_u_atom
       REAL(KIND=C_DOUBLE), ALLOCATABLE, DIMENSION(:)     :: fun
       REAL(KIND=C_DOUBLE), ALLOCATABLE, DIMENSION(:, :)  :: dion
       REAL(KIND=C_DOUBLE), DIMENSION(3)                  :: a1, a2, a3, v1, v2
       REAL(KIND=dp)                                      :: al, angle1, angle2, cval, focc, &
-                                                            magnetization, mass, pf, rl, zeff
+                                                            magnetization, mass, pf, rl, zeff, alpha_u, beta_u, &
+                                                            J0_u, J_u, U_u, occ_u, u_minus_J
       REAL(KIND=dp), ALLOCATABLE, DIMENSION(:)           :: beta, corden, ef, fe, locpot, rc, rp
       REAL(KIND=dp), DIMENSION(3)                        :: vr, vs
       REAL(KIND=dp), DIMENSION(:), POINTER               :: density
@@ -285,10 +286,12 @@ CONTAINS
          CALL get_qs_kind(qs_kind_set(ikind), zeff=zeff)
          NULLIFY (upf_pot, gth_potential)
          CALL get_qs_kind(qs_kind_set(ikind), upf_potential=upf_pot, gth_potential=gth_potential)
+
          IF (ASSOCIATED(upf_pot)) THEN
             CALL sirius_add_atom_type(sctx, label, fname=upf_pot%filename, &
                                       symbol=element_symbol, &
                                       mass=REAL(mass/massunit, KIND=C_DOUBLE))
+
          ELSEIF (ASSOCIATED(gth_potential)) THEN
 !
             NULLIFY (atom_grid)
@@ -393,6 +396,7 @@ CONTAINS
             DO iwf = 1, SIZE(wavefunction, 2)
                focc = wfninfo(1, iwf)
                l = NINT(wfninfo(2, iwf))
+               nu = NINT(wfninfo(3, iwf))
                IF (up) THEN
                   fun(1:nmesh) = wavefunction(1:nmesh, iwf)*rp(i)
                ELSE
@@ -402,7 +406,7 @@ CONTAINS
                END IF
                CALL sirius_add_atom_type_radial_function(sctx, &
                                                          label, "ps_atomic_wf", &
-                                                         fun(1), nmesh, l=l, occ=REAL(focc, KIND=C_DOUBLE), n=-1)
+                                                         fun(1), nmesh, l=l, occ=REAL(focc, KIND=C_DOUBLE), n=nu)
             END DO
 
 ! set total charge density of a free atom (to compute initial rho(r))
@@ -416,6 +420,39 @@ CONTAINS
             CALL sirius_add_atom_type_radial_function(sctx, label, "ps_rho_total", &
                                                       fun(1), nmesh)
 
+            CALL get_qs_kind(qs_kind_set(ikind), &
+                             dft_plus_u_atom=dft_plus_u_atom, &
+                             l_of_dft_plus_u=lu, &
+                             n_of_dft_plus_u=nu, &
+                             u_minus_j_target=u_minus_j, &
+                             U_of_dft_plus_u=U_u, &
+                             J_of_dft_plus_u=J_u, &
+                             alpha_of_dft_plus_u=alpha_u, &
+                             beta_of_dft_plus_u=beta_u, &
+                             J0_of_dft_plus_u=J0_u, &
+                             occupation_of_dft_plus_u=occ_u)
+
+            IF (dft_plus_u_atom) THEN
+               IF (nu < 1) THEN
+                  CALL cpabort("CP2K/SIRIUS (hubbard): principal quantum number not specified")
+               END IF
+
+               IF (lu < 0) THEN
+                  CALL cpabort("CP2K/SIRIUS (hubbard): l can not be negative.")
+               END IF
+
+               IF (occ_u < 0.0) THEN
+                  CALL cpabort("CP2K/SIRIUS (hubbard): the occupation number can not be negative.")
+               END IF
+
+               IF (ABS(u_minus_j) < 1e-8) THEN
+                  CALL sirius_set_atom_type_hubbard(sctx, label, lu, nu, &
+                                                    NINT(occ_u), U_u, J_u, alpha_u, beta_u, J0_u)
+               ELSE
+                  CALL sirius_set_atom_type_hubbard(sctx, label, lu, nu, &
+                                                    NINT(occ_u), u_minus_j, 0.0_dp, alpha_u, beta_u, J0_u)
+               END IF
+            END IF
             IF (ASSOCIATED(density)) DEALLOCATE (density)
             IF (ASSOCIATED(wavefunction)) DEALLOCATE (wavefunction)
             IF (ASSOCIATED(wfninfo)) DEALLOCATE (wfninfo)
@@ -426,7 +463,7 @@ CONTAINS
 !
          ELSE
             CALL cp_abort(__LOCATION__, &
-                          'CP2K/SIRIUS: atomic kind needs UPF or GTH potential definition')
+                          "CP2K/SIRIUS: atomic kind needs UPF or GTH potential definition")
          END IF
       END DO
 
--- a/src/sirius_interface.F
+++ b/src/sirius_interface.F
@@ -73,8 +73,9 @@ MODULE sirius_interface
         sirius_get_stress_tensor, sirius_ground_state_handler, sirius_import_parameters, &
         sirius_initialize, sirius_initialize_context, sirius_kpoint_set_handler, &
         sirius_option_get_info, sirius_option_get_section_length, sirius_option_set, &
-        sirius_set_atom_position, sirius_set_atom_type_dion, sirius_set_atom_type_radial_grid, &
-        sirius_set_lattice_vectors, sirius_set_mpi_grid_dims, sirius_update_ground_state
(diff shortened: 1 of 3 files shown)


mkrack (changes requested):
I think that the lines [2286-2298 in input_cp2k_subsys.F](https://github.com/cp2k/cp2k/pull/2223/commits/2093fd738e9de103d5797e9e906ef57ddc13a90d#diff-553109d2ebaa8b09a5c386555b10579fefc6f3022004729e2bcc88528e7c8d1fR2286-R2298) should not be in the scope of #if defined(__SIRIUS)

mtaillefumier:
you are right. it was actually wrong and I suppressed the compilation dependent part. Thanks.

mkrack (approved):

src/sirius_interface.F
```diff
@@ -396,6 +400,39 @@ SUBROUTINE cp_sirius_create_env(pwdft_env)
...
+                             U_of_dft_plus_u=U_u, &
+                             J_of_dft_plus_u=J_u, &
+                             alpha_of_dft_plus_u=alpha_u, &
+                             beta_of_dft_plus_u=beta_u, &
+                             J0_of_dft_plus_u=J0_u, &
+                             occupation_of_dft_plus_u=occ_u)
+
+            IF (dft_plus_u_atom) THEN
+               IF (nu < 1) THEN
+                  CALL cpabort("CP2K/SIRIUS (hubbard): principal quantum number not specified")
```
oschuett: It's either 
```
CPABORT("A short error message")
```
 or 
```
  CALL cp_abort(__LOCATION__,"A long error message "//&
                 "with even more valuable advises.")  
```


See also: https://www.cp2k.org/dev:error_handling.

oschuett:
I'm a bit supprised that you need all those new keywords. Can't you just re-use the exiting ones in the [DFT_PLUS_U](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/SUBSYS/KIND/DFT_PLUS_U.html) section?

mkrack (commented):
Why is this check unnecessary?

mtaillefumier:
cp2k implementation of dft+U is incomplete and takes for granted that the hubbard correction should be applied to the last (partially) quantum atomic level of each relevant atomic species. 

SIRIUS can deal with multi-orbitals hubbard corrections within a given atom or between neighbors (the later is *not* supported by cp2k). We impose these constraints to run with hubbard correction
- the user should indicate which atomic level ((n,l)-pair) should be considered for each relevant atomic type 
- the level occupation
- and of course either U, Ueff, J, etc.... 

Without this sirius will simply not run when hubbard corrections are requested. SIRIUS internally checks that the n and l values are sane and these checks should also be present in the interface. If not I need to add them.

I realized early on that cp2k implementation of dft+u is fairly incomplete (it is the simplified version) that's why I had to add extra parameters to the dft+U section.

mkrack:
I suggest to create a separate DFT+U section for SIRIUS and leave the CP2K section untouched. The implementations are too different.

mkrack:
> cp2k implementation of dft+U is incomplete and takes for granted that the hubbard correction should be applied to the last (partially) quantum atomic level of each relevant atomic species.

That is not true.

mtaillefumier:
> 

the cp2k implementation of DFT+U is a simplified version of the entire theory. 
I could not find where n and l information are given in the input file that's why I assumed the defaults like QE was doing one year ago.

mkrack:
> I realized early on that cp2k implementation of dft+u is fairly incomplete (it is the simplified version) that's why I had to add extra parameters to the dft+U section.

CP2K supports only Dudarev's approach. That is why mixing the section will confuse users.

mkrack:
> the cp2k implementation of DFT+U is a simplified version of the entire theory.
> I could not find where n and l information are given in the input file that's why I assumed the defaults like QE was doing one year ago.

Right, there is no input option for n yet, e.g. to exclude semi-core states of same l exclusively, because there was no need/request for it so far.

mtaillefumier:
> > the cp2k implementation of DFT+U is a simplified version of the entire theory.
> > I could not find where n and l information are given in the input file that's why I assumed the defaults like QE was doing one year ago.
> 
> Right, there is no input option for n yet, e.g. to exclude semi-core states of same l exclusively, because there was no need/request for it so far.

I understand your point perfectly that's why it made sense to exclude this. However SIRIUS needs this information otherwise it won't run any simulation involving the Hubbard correction. n, l are as important as U_eff or U and J even more so when we want to do multi-orbitals calculations (it is not activated in cp2k) or we want to use lda+U+V. 

I am not sure that creating a new section containing the same information is actually the best option as these information are tied to the atomic specie which already has its own section in cp2k input.

mkrack:
If we want to keep one DFT+U section, we have two options:
- we generalise to U and J and drop U_MINUS_J as keyword (which breaks compatibility with old inputs)
- we check if U or J are specified explicitly together with U_MINUS_J in the input file and stop the run if that is the case with an error message and ask the user to decide.

I can add n to the CP2K implementation to allow for a selection among the l shells of an atomic kind. Currently, all shells are selected which we could keep as the default.

mtaillefumier:
> If we want to keep one DFT+U section, we have two options:
> 
>     * we generalise to U and J and drop U_MINUS_J as keyword (which breaks compatibility with old inputs)
> 
>     * we check if U or J are specified explicitly together with U_MINUS_J in the input file and stop the run if that is the case with an error message and ask the user to decide.

I think the second option is better as it is not difficult to check if all three parameters are present. I can add it and say explicitly that it is either U_eff or J and/or U. 

> I can add n to the CP2K implementation to allow for a selection among the l shells of an atomic kind. Currently, all shells are selected which we could keep as the default.

l should be added as well. By default if n and l are not present and SIRIUS is not used then the parameters are ignored completely. It means in practice that the default behavior is cp2k. If SIRIUS is used then the parameters should be present otherwise cp2k aborts with an error message. 

It might be worth making the error message more explicit. 

How do we want to proceed. I will add the check for sure.

mkrack:
> l should be added as well. By default if n and l are not present and SIRIUS is not used then the parameters are ignored completely. It means in practice that the default behavior is cp2k. If SIRIUS is used then the parameters should be present otherwise cp2k aborts with an error message.

L is required with CP2K. N could be optional with the default all. N is rather the number/index of the n-th shell with the angular momentum number L than the principal quantum number. The latter is only well-defined for minimal basis sets, because you can have multiple L shells for a principal quantum number N. The N value in the CP2K basis set definition has no impact on the calculation and it is used at best for printout purposes (if it is used at all). Would that be compatible with SIRIUS?

Given your aim of generality, there could be, in principle, multiple L values with different U,J values for an atomic kind, e.g. acting on specific p and d states separately. Do you plan to allow for that?

Do you plan to add also DFT+U+V or did you implement it already?

mtaillefumier:
> > l should be added as well. By default if n and l are not present and SIRIUS is not used then the parameters are ignored completely. It means in practice that the default behavior is cp2k. If SIRIUS is used then the parameters should be present otherwise cp2k aborts with an error message.
> 
> L is required with CP2K. N could be optional with the default all. N is rather the number/index of the n-th shell with the angular momentum number L than the principal quantum number. The latter is only well-defined for minimal basis sets, because you can have multiple L shells for a principal quantum number N. The N value in the CP2K basis set definition has no impact on the calculation and it is used at best for printout purposes (if it is used at all). Would that be compatible with SIRIUS?

Yes it is. l is mandatory which is also needed by sirius. So we could simply ignore N when SIRIUS is not used explicitly.

> Given your aim of generality, there could be, in principle, multiple L values with different U,J values for an atomic kind, e.g. acting on specific p and d states separately. Do you plan to allow for that?
I would support the idea actually. SIRIUS apply the hubbard correction to all relevant atomic orbitals (n,l) and with possibly different U,J,U_eff parameters. Most of the work would be to design the input scheme in cp2k that supports it but does not break the old behavior. 


> Do you plan to add also DFT+U+V or did you implement it already?

DFT+U+V is present in SIRIUS as the EPFL group we work with, asked us to implement it. It is working for non magnetic and colinear magnetic systems and there is no barrier (at least in principle) to use it in full non-colinear simulations.

We could also turn on this functionality in cp2k but it will require the creation of a full section (or an input file of some sort) to pass the neighbors lists of each atom, their corresponding orbitals (2p-3d for instance) and their effective V.

mkrack:
> Yes it is. l is mandatory which is also needed by sirius. So we could simply ignore N when SIRIUS is not used explicitly.

@mtaillefumier Add N for SIRIUS and I will later consider it also for CP2K. Likewise, add U and J together with the discussed checks concerning U_MINUS_J/U_EFF.

> > Given your aim of generality, there could be, in principle, multiple L values with different U,J values for an atomic kind, e.g. acting on specific p and d states separately. Do you plan to allow for that?

Do you have any answer to that question?

> DFT+U+V is present in SIRIUS as the EPFL group we work with, asked us to implement it. It is working for non magnetic and colinear magnetic systems and there is no barrier (at least in principle) to use it in full non-colinear simulations.
> 
> We could also turn on this functionality in cp2k but it will require the creation of a full section (or an input file of some sort) to pass the neighbors lists of each atom, their corresponding orbitals (2p-3d for instance) and their effective V.

Indeed, DFT+U+V would require a separate input section, e.g. PLUS_V, within the DFT section analogous to the PLUS_U section there. The +V information could also be read from an external like for basis sets or force fields, but it also would be fine to have that information explicit in the input file in a first step and to use the INCLUDE mechanism if it becomes too large.

mtaillefumier:
> > Yes it is. l is mandatory which is also needed by sirius. So we could simply ignore N when SIRIUS is not used explicitly.
> 
> @mtaillefumier Add N for SIRIUS and I will later consider it also for CP2K. Likewise, add U and J together with the discussed checks concerning U_MINUS_J/U_EFF.
> 
> > > Given your aim of generality, there could be, in principle, multiple L values with different U,J values for an atomic kind, e.g. acting on specific p and d states separately. Do you plan to allow for that?
> 
> Do you have any answer to that question?

SIRIUS supports multi-orbitals on a single atomic type (and multi-orbital between atom pairs as well). We put this in place when we developed LDA+U+V. Supporting this in cp2k is a matter of agreeing how we represent this information in cp2k input files.

> 
> > DFT+U+V is present in SIRIUS as the EPFL group we work with, asked us to implement it. It is working for non magnetic and colinear magnetic systems and there is no barrier (at least in principle) to use it in full non-colinear simulations.
> > We could also turn on this functionality in cp2k but it will require the creation of a full section (or an input file of some sort) to pass the neighbors lists of each atom, their corresponding orbitals (2p-3d for instance) and their effective V.
> 
> Indeed, DFT+U+V would require a separate input section, e.g. PLUS_V, within the DFT section analogous to the PLUS_U section there. The +V information could also be read from an external like for basis sets or force fields, but it also would be fine to have that information explicit in the input file in a first step and to use the INCLUDE mechanism if it becomes too large.

That's a good option. I do not know how long it will take me to get this in cp2k as most of the work is in cp2k alone. We have two options, merging this PR and then creating a PR for the new functionality or add an extra commit on the top of this one. I let you decide what you think is the best course of action. 

Few comments : 

- I would just refrain reading this information from a basis set as cp2k pps are not particularly tailored for computations in the plane wave domain. My colleagues at ORNL had all the pain in the world to get computations converge to realistic values with cp2k/sirius and cp2k pp. 
- U,V parameters etc... should be given explicitly as they depend on the unit cell. They can only be deduced from linear response or other equivalent methods. 

To summarize what sirius supports for LDA+Hubbard
- U term (dudarev, etc...), colinear and non-colinear magnetism
- U+V term, colinear and non colinear magnetism (non colinear magnetism was not tested due to lack of interest)
- multi-shells are also supported in both U and U+V.

mkrack:
> SIRIUS supports multi-orbitals on a single atomic type (and multi-orbital between atom pairs as well). We put this in place when we developed LDA+U+V. Supporting this in cp2k is a matter of agreeing how we represent this information in cp2k input files.

Currently, that is not possible with CP2K, because the DFT_PLUS_U section in &KIND is not repeatable. We will have to change that to allow for multiple L and U,J per atomic kind. I can put that on my todo list.
 
> > Indeed, DFT+U+V would require a separate input section, e.g. PLUS_V, within the DFT section analogous to the PLUS_U section there. The +V information could also be read from an external like for basis sets or force fields, but it also would be fine to have that information explicit in the input file in a first step and to use the INCLUDE mechanism if it becomes too large. 
> That's a good option. I do not know how long it will take me to get this in cp2k as most of the work is in cp2k alone. We have two options, merging this PR and then creating a PR for the new functionality or add an extra commit on the top of this one. I let you decide what you think is the best course of action.
> 
> Few comments :
> 
> * I would just refrain reading this information from a basis set as cp2k pps are not particularly tailored for computations in the plane wave domain. My colleagues at ORNL had all the pain in the world to get computations converge to realistic values with cp2k/sirius and cp2k pp.

Here we misunderstood each other. I do not propose to read +V information from a basis set file. I gave basis set and force field parameters only as an example for input information often read from an external file in CP2K.

> * U,V parameters etc... should be given explicitly as they depend on the unit cell. They can only be deduced from linear response or other equivalent methods.
> 
> To summarize what sirius supports for LDA+Hubbard
> 
> * U term (dudarev, etc...), colinear and non-colinear magnetism
> * U+V term, colinear and non colinear magnetism (non colinear magnetism was not tested due to lack of interest)
> * multi-shells are also supported in both U and U+V.

DFT+U+V (only Dudarev) and multiple L shells could be implemented in CP2K. On the other hand, I think that it is rather unlikely that non-collinear calculations will become available in CP2K in the near future unless via CP2K/SIRIUS.

First, we should enable DFT+U with CP2K/SIRIUS with this and if needed some follow-up commits.

mtaillefumier:
> Here we misunderstood each other. I do not propose to read +V information from a basis set file. I gave basis set and force field parameters only as an example for input information often read from an external file in CP2K.

Thanks for the clarification. For sure this option should be supported as it becomes tedious to add this manually.

> DFT+U+V (only Dudarev) and multiple L shells could be implemented in CP2K. On the other hand, I think that it is rather unlikely that non-collinear calculations will become available in CP2K in the near future unless via CP2K/SIRIUS.
> 
> First, we should enable DFT+U with CP2K/SIRIUS with this and if needed some follow-up commits.

Let me add an other commit that actually clearly state that U,J or U_eff can be given but not both (although they are related in Dudarev version). I should be able to add this tomorrow.

mkrack:
@mtaillefumier Are you planning to commit this PR soon? Conflicts start to accumulate and I would like to continue revising the DFT+U implementation.

mtaillefumier:
I do not have commit rights actually. I can fix the conflicts though

mkrack:
> I do not have commit rights actually. I can fix the conflicts though

I will do the commit when you have resolved the conflicts.

PR #2229, opened 2022-08-01
automatically extrapolate k-point mesh size for periodic GW calculations

JWilhelm:
k-point mesh extrapolation for k-point integration of W(k). Bandgap of MoS2 monolayer converges now with the cell size, see figure below. 

Open issue: GW bandgap is not yet converged with RI basis set size. 

Possible solution: Use truncated Coulomb metric (tCm); probably tCm needs removal of G=0 component of RI s-functions. 

image

### DIFF
--- a/src/hfx_types.F
+++ b/src/hfx_types.F
@@ -2591,10 +2591,10 @@ CONTAINS
                   "HFX_RI_INFO| Omega:     ", ri_metric%omega
             CASE (do_potential_id)
                WRITE (UNIT=iw, FMT="(T3,A,T73,A)") &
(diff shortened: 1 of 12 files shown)


fstein93:
A question out of interest: Is it somehow possible to reduce the runtime of the RPA kpoint tests while still being reasonable? Especially the batched version takes often more than two minutes to complete on our machines, on the CI, it is a minute.

JWilhelm:
Changing periodicity from XYZ to Z reduced the computation time for me by a factor 10. Is this sufficient?

PR #2242, opened 2022-08-12
DBM/benchmark: minor enhancements

hfp:
* Initialize matrix operands for benchmark (to avoid potential NaNs in the FP-pipeline).
* Use LIBXSMM if possible (via LIBXSMMROOT environment variable).
* Typo: renamed bechmark_multiply to benchmark_multiply.

### DIFF
--- a/src/dbm/dbm_miniapp.c
+++ b/src/dbm/dbm_miniapp.c
@@ -72,6 +72,8 @@ static dbm_matrix_t *create_some_matrix(const int row_size, const int col_size,
   }
   dbm_matrix_t *matrix = NULL;
   dbm_create(&matrix, dist, "some name", nrow, ncol, row_sizes, col_sizes);
+  // Ensure non-NaN values to avoid degenerated performance
+  dbm_zero(matrix);
   dbm_distribution_release(dist);
   free(row_sizes);
   free(col_sizes);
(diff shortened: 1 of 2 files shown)


src/dbm/dbm_miniapp.c
```diff
@@ -72,6 +72,8 @@ static dbm_matrix_t *create_some_matrix(const int row_size, const int col_size,
   }
   dbm_matrix_t *matrix = NULL;
   dbm_create(&matrix, dist, "some name", nrow, ncol, row_sizes, col_sizes);
+  // Ensure non-NaN values to avoid degenerated performance
+  dbm_zero(matrix);
```
oschuett: This shouldn't have any effect, because `dbm_zero` only zeros existing blocks and newly created matrices are empty.

Furthermore, when a new block is added, e.g. via `dbm_reserve_blocks`, then it's [always zeroed](https://github.com/cp2k/cp2k/blob/52ce71a09dcc858b1f1c7a89a11e8623f0e6a91b/src/dbm/dbm_shard.c#L222).

hfp: > Furthermore, when a new block is added, e.g. via dbm_reserve_blocks, then it's [always zeroed](https://github.com/cp2k/cp2k/blob/52ce71a09dcc858b1f1c7a89a11e8623f0e6a91b/src/dbm/dbm_shard.c#L222).

Ok, good to know! I did not find this in the code (my fault). I will revert this part.

PR #2255, opened 2022-08-29
Add Mi250 GPU support

alazzaro:
This PR is tested on LUMI-G


oschuett:
The CI is not using the [pre-commit-config.yaml](https://github.com/cp2k/cp2k/blob/master/.pre-commit-config.yaml) file anymore. Since it became very cumbersome to install all those tools locally, I've built a cloud-based solution in [tools/precommit](https://github.com/cp2k/cp2k/tree/master/tools/precommit).  You can run the checks simply via `make pretty`. We should probably remove the old yaml config to avoid more confusion.

PR #2258, opened 2022-08-30
Upgrade to SIRIUS 7.3.2

### LINKED ISSUE #1912: Empty enum label in SIRIUS input section

oschuett:
Fixes #1912.

### DIFF
--- a/src/input/input_enumeration_types.F
+++ b/src/input/input_enumeration_types.F
@@ -80,6 +80,7 @@ CONTAINS
       enum%ref_count = 1
       ALLOCATE (enum%c_vals(SIZE(c_vals)))
       DO i = 1, SIZE(enum%c_vals)
+         CPASSERT(LEN_TRIM(c_vals(i)) > 0)
          enum%c_vals(i) = c_vals(i)
          CALL uppercase(enum%c_vals(i))
       END DO
--- a/tests/SIRIUS/regtest-1/He-full-potential.inp
+++ b/tests/SIRIUS/regtest-1/He-full-potential.inp
@@ -13,7 +13,7 @@
        USE_SYMMETRY true
        VALENCE_RELATIVITY none
        CORE_RELATIVITY none
-       NUM_FV_STATES 40
+       NUM_FV_STATES 10
        AW_CUTOFF 8.0
        PW_CUTOFF 20.00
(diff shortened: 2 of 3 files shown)


oschuett:
The `He-full-potential.inp` test [fails](https://storage.googleapis.com/cp2k-ci/run-cp2k-pdbg-ae6dad3a_report.txt) with the following message:
```
[solve] iterative solver tolerance (occupied, empty) : 2.5000e-05, 2.5000e-05
[get_singular_components] smallest eigen-value of the singular components:   0.0013624697271140
SIRIUS: run-time error

[orthogonalize] /opt/cp2k-toolchain/build/SIRIUS-7.3.2/src/SDDK/wf_ortho.cpp:199
[orthogonalize] error in factorization, info = 59
[orthogonalize] number of existing states: 0
[orthogonalize] number of new states: 61
[orthogonalize] number of wave_functions: 3
[orthogonalize] idx_bra: 0 idx_ket:2
```
@mtaillefumier, any idea what this might be?

mtaillefumier:
one parameter should be adjusted in the input file `NUM_FV_STATES 10`. I do not recall why I set this parameter to 40 but it is set to 10 in our tests which cp2k just borrowed.

PR #2270, opened 2022-09-06
[SIRIUS] libvdwxc functionality

mtaillefumier:
Add an extra option to support libvdwxc/sirius functionality in cp2k.


src/input_cp2k_pwdft.F
```diff
@@ -53,6 +54,13 @@ MODULE input_cp2k_pwdft
 
    PUBLIC :: create_pwdft_section
 
+#if defined(__LIBVDWXC)
```
oschuett: I think, these constants should be defined regardless of whether libvdwxc is present. We currently don't have a CI test for this, but I believe right now one can not build with SIRIUS and without libwdvxc.

mtaillefumier: it is possible to compile sirius without libvdwxc even with the toolchain. The reason is it is not a hard dependency so it is not enforced.

oschuett: Right, the problem is that you are importing these constants in `sirius_interface.F` regardless of whether libvdwxc is present or not. The simplest solution would be to have them always defined.

oschuett:
LGTM.
Do you want to add a test for the new feature? It would require a new [TEST_DIR](https://github.com/cp2k/cp2k/blob/master/tests/TEST_DIRS) conditioned on the `libvdwxc` flag.
Otherwise, I can also merge this PR as is.

mtaillefumier:
ideally we need tests cases for this but I prefer to see this validated with a real calculation than just a dummy test showing convergence (Opening an issue will remind me to do it). 
One user had trouble with vdw that what triggered this PR. So it is probably better to merge the PR so he can use vdw with SIRIUS and at the same time helps validating it.

I ran his input file and SIRIUS converged without any problem . Of course gold is easy.

mtaillefumier:
thanks. I can now ask the user to give it a spin.

PR #2271, opened 2022-09-06
Add RPA gradients

fstein93:
This PR adds gradients of the quartically scaling RPA kernel. The code was tested locally and on Piz Daint but not yet on LUMI-G. The code was optimized to reduce the memory footprint on CP2K side. It reuses as much of the SOS-MP2 and MP2-gradient code as possible.

The code employs COSMA but might require significantly more memory than before. Local DGEMM calls make use of SpLA. 

To my experience, the memory requirements increase significantly with COSMA (64 water molecules run well, larger systems don't because of COSMA). I do not know how SpLA increases the requirements. Thus, some more tuning might be required. Probably, machines with more memory like LUMI-G might not be affected.

This PR replaces the SIZE_INTEG_FREQ_GROUP keywords by the NUM_INTEG_GROUP keywords because of easier use.


alazzaro:
> 
> To my experience, the memory requirements increase significantly with COSMA (64 water molecules run well, larger systems don't because of COSMA). I do not know how SpLA increases the requirements. Thus, some more tuning might be required. Probably, machines with more memory like LUMI-G might not be affected.
> 

To be clear, are you talking about GPU or CPU memory? 
In any case, in both cases LUMI has much more memory...

fstein93:
Honestly speaking, both. On Daint, I got OOM events within COSMA, either because of missing memory on CPU or from CUDA. I tried to reduce the footprint from SpLA by allocating the SpLA object right before its use.

alazzaro:
OK, got it. LUMI will be different (512 GB on the node, 128 GB per each GPU)...

PR #2279, opened 2022-09-12
Add switch for local DGEMM calls

fstein93:
Useful for debugging if there is not enough memory.

### DIFF
--- a/src/admm_methods.F
+++ b/src/admm_methods.F
@@ -43,7 +43,6 @@ MODULE admm_methods
                                               cp_fm_set_element,&
                                               cp_fm_to_fm,&
                                               cp_fm_type
-   USE cp_gemm_interface,               ONLY: cp_gemm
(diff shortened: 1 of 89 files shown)


oschuett:
Thanks for picking this up! I also had it on my todo list to refactor this.

I think, it would be best to have a generic interface similar what we already have for [parallel gemm](https://github.com/cp2k/cp2k/blob/a400cd20616092f954660d2222435cb173b5ff77/src/cp_gemm_interface.F).

So, my plan would have been to:

- Rename `cp_gemm_interface.F` to `parallel_gemm_api.F`
- Rename `offload_gemm.F` to `local_gemm_api.F`
- Rename routines within these files accordingly.

fstein93:
Okay, I am renaming the calls to cp_gemm to parallel_gemm and calls to offload_dgemm to local_gemm. I will keep the preprocessor flag OFFLOAD_GEMM because of the common backbone.

PR #2294, opened 2022-09-19
Further refactoring of the fm package

fstein93:
This PR removes the reference counting of the cp_(c)fm_type and the POINTER attributes of these types within the fm package. This allows to write much cleaner code. It will require more effort to remove POINTER attributes within the calling routines and the different derived types as well as converting (arrays of) cp_(c)fm_p_type variables to arrays of cp_(c)fm_type.


oschuett:
Thanks a lot for modernizing the fm code!

The memory leaks that you're [getting now](https://storage.googleapis.com/cp2k-ci/run-cp2k-pdbg-010ca531_report.txt) are probably because of `ref_count = 0`, try initializing it to `1`.

Alternatively, you could just remove the `ref_count` entirely. While reference counting is widely-used within CP2K, it's not actually needed anywhere (AFAICT).

fstein93:
The failures had a different origin, namely the call of a copying routine. I solved it with a wrapper of the pointer-free version to be called if it the caller cannot ensure that source or the destination is associated at all. The workaround should not affect the performance too much. I hope I haven't overseen any case to be treated this way because of unsufficient case-coverage of the regtests.

oschuett:
>  I solved it with a wrapper of the pointer-free version to be called if it the caller cannot ensure that source or the destination is associated at all....  I hope I haven't overseen any case....

Hmm, this sounds like a new way to shoot oneself in the foot. Isn't there a way to handle this without making the public API more complicated?

fstein93:
A possible alternative is to use a dummy matrix if such an array is not associated. If you do not agree to it, I have to rewind my change and return to the old API with dummy pointers and reintroduce pointers in some cases.

oschuett:
> A possible alternative is to use a dummy matrix if such an array is not associated.

Yes, I think this is the way to go. A new matrix of `TYPE(cp_fm_type)` is empty at first, ie. its internal pointers are all null.

With this semantic you also don't need those new routines `cp_fm_create_p`, `cp_fm_release_p`, and `cp_fm_soft_copy`.

fstein93:
Ok, I removed the reference counting entirely. I kept the routine cp_fm_soft copy as a helper routine but moved it to the respective module as a private routine. I updated the initial comment. Potentially, it requires far more effort to subsequently remove more and more POINTER attributes and convert constructs using cp_(c)fm_p_types to cp_(c)fm_type. Because of the many additional (DE)ALLOCATE calls, the code blew a bit up but subsequent PRs can reduce it.

oschuett:
Thanks for removing the reference counting. This is now looking really nice :-)

>  I kept the routine cp_fm_soft copy as a helper routine...

I think you can just use intrinsic assignment of derived types, ie. simply write `matrix_b = matrix_a`.


> Potentially, it requires far more effort to subsequently remove more and more POINTER attributes and convert constructs using cp_(c)fm_p_types to cp_(c)fm_type.

I agree, this is the natural next step. The good news is that it can be done piecewise, ie. one array at a time.

alazzaro:
Cray compiler is giving an error as result of this PR:
```
ftn-855 ftn: ERROR CP_CFM_TYPES, File = ../../../pfs/lustrep4/projappl/project_462000031/lazzaroa/cp2k_cce/cp2k/src/fm/cp_cfm_types.F, Line = 12, Column = 8 
  The compiler has detected errors in module "CP_CFM_TYPES".  No module information file will be created for this module.


ftn-1707 ftn: ERROR CP_CFM_GET_INFO, File = ../../../pfs/lustrep4/projappl/project_462000031/lazzaroa/cp2k_cce/cp2k/src/fm/cp_cfm_types.F, Line = 660, Column = 23 
  Pointer or allocatable dummy argument "MATRIX" must not appear in a ALLOCATE statement as it has the intent(in) attribute.


ftn-1707 ftn: ERROR CP_CFM_GET_INFO, File = ../../../pfs/lustrep4/projappl/project_462000031/lazzaroa/cp2k_cce/cp2k/src/fm/cp_cfm_types.F, Line = 680, Column = 23 
  Pointer or allocatable dummy argument "MATRIX" must not appear in a ALLOCATE statement as it has the intent(in) attribute.
```

fstein93:
Does my new PR solve this issue?

alazzaro:
Definitely... Thanks!

PR #2297, opened 2022-09-21
Update dashboard (CRAY performance tests)

mkrack:


alazzaro:
Thanks @mkrack !
Consequently we can move the old tests to https://dashboard.cp2k.org/discontinued_tests.html ? @oschuett 
BTW, I see the top table links in that page are all broken...

mkrack:
Let's see first if everything works fine. If I am not mistaken, tests deleted from the dashboard are automatically moved to the discontinued tests.

oschuett:
> If I am not mistaken, tests deleted from the dashboard are automatically moved to the discontinued tests.

Actually, I'm maintaining that file by hand 🙃

> BTW, I see the top table links in that page are all broken...

Yeah, I believe they got lost during the migration from subversion to git. I've remove the broken entries. Furthermore, I've reversed the order of the table. Now the youngest entry is at the top.

PR #2309, opened 2022-09-29
Update arch file for leak check

mkrack:


oschuett:
Note that Addresse Sanitizer found mostly problems in the C code.

The Fortran code was surprisingly clean presumably because `-fcheck=bounds` already found most issues.

And as a heads up `-fsanitize=address` comes with a 3x slowdown.

mkrack:
@oschuett Thanks for your comments. That is good news concerning the Fortran code part.
The flag `-fsanitize=address` is only activated, if the flag `DO_CHECKS=yes` is defined which is used for this [regression test](https://dashboard.cp2k.org/archive/psi-gnu-mpich-psmp/index.html) to check for memory leaks. In this way, the arch file can still be used to build production binaries.

oschuett:
Sounds good! What I meant was that it would make sense to also add `-fsanitize=address` to the `CFLAGS`.

mkrack:
> Sounds good! What I meant was that it would make sense to also add `-fsanitize=address` to the `CFLAGS`.

Of course, I will do that.

PR #2311, opened 2022-09-30
Enable toolchain build for target CPU

mkrack:
- toolchain flag `--target-cpu=<cpu>` added to facilitate cross-compilation
- toolchain flag `--generic` removed, `--target-cpu=generic` can be used instead
- the default toolchain build is `--target-cpu=native` as before (host CPU)
- Docker and arch files updated correspondingly

### DIFF
--- a/arch/Linux-gnu-aarch64-static.ssmp
+++ b/arch/Linux-gnu-aarch64-static.ssmp
@@ -16,7 +16,9 @@
       exit 1; \
    fi; \
    this_file=${BASH_SOURCE##*/}; \
-   ./install_cp2k_toolchain.sh --mpi-mode=no --no-arch-files --with-gcc=system --with-cmake=system; \
+   cd tools/toolchain; \
+   [[ -z "${target_cpu}" ]] && target_cpu="native"; \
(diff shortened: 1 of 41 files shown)


alazzaro:
Are you considering the ELPA installer, namely the part starting with this line https://github.com/cp2k/cp2k/blob/a1070fec19854ac908ca2aa7523f81003e7653b0/tools/toolchain/scripts/stage5/install_elpa.sh#L91 ?

(same problem for FFTW)

oschuett:
Is [that part](https://github.com/cp2k/cp2k/blob/a1070fec19854ac908ca2aa7523f81003e7653b0/tools/toolchain/scripts/stage5/install_elpa.sh#L85-L105) actually still needed? According to ELPA's changelog there was [a bit of cleanup](https://gitlab.mpcdf.mpg.de/elpa/elpa/-/blob/master/Changelog#L93) in the meantime.

mkrack:
@alazzaro Thank you for rising that issue. Ideally, the toolchain should not bother with hardware specific configuration flags for building the libraries. I will restrict the tunings based on `/proc/cpuinfo` to the `--target-cpu=native`.

Issue #2262, opened 2022-09-01
Fails to build due to `spglib` tagged version no longer existing on the expected repo


zacharyrs:
The tagged version (`v1.16.2`) isn't available at [atztogo/spglib](https://github.com/atztogo/spglib), so the install process crashes when attempting to download the source.

https://github.com/cp2k/cp2k/blob/350aa13796e298b7c753ce0ae324e06efd7409d9/tools/toolchain/scripts/stage7/install_spglib.sh#L34

The tagged version is still available on the main repository [here](https://github.com/spglib/spglib/releases/tag/v1.16.2).

PR #2314, opened 2022-10-02
Toolchain: Download all package from cp2k.org

oschuett:
... to prevent problems like https://github.com/cp2k/cp2k/issues/2262 from reoccurring.


oschuett:
Since users are [affected](https://groups.google.com/g/cp2k/c/IgfAnOKAqnE/m/hYiaFi_BBQAJ) and our next regular release is still ~3 months out I'm planing a minor release with [this backport](https://github.com/cp2k/cp2k/commit/a95ec4033747cff22669506ef08e06d7d5f32936).

PR #2315, opened 2022-10-03
data: add pob-rev2 basis sets and ECP

dev-zero:


juerghutter:
Looks good to me. The ECPs might not work/need adjustments in CP2K.

PR #2327, opened 2022-10-05
Disable libvori and libxsmm explicitly for Darwin-aarch64

mkrack:


hfp:
Re: libxsmm on Darwin - this will come soon with v2.0. LIBXSMM's main version already supports AArch64 in various flavors.

PR #2332, opened 2022-10-07
Remove null pointers if they are passed as optional arguments

fstein93:
Passing NULL() as an optional argument is standard conform to indicate an absent actual argument but does not work with some compilers under special circumstances. My last PR #2294 has triggered them (not on Dashboard) in the context of the eigensolvers. This PR removes the null pointers such that the optional arguments are omitted makes the calls to the eigensolver routines more explicit.

@juerghutter Does this PR fix your issue?

### DIFF
--- a/src/qs_initial_guess.F
+++ b/src/qs_initial_guess.F
@@ -585,7 +585,6 @@ CONTAINS
                              cholesky_method=scf_env%cholesky_method, &
                              do_level_shift=.FALSE., &
                              level_shift=0.0_dp, &
-                             matrix_u_fm=null(), &
                              use_jacobi=.FALSE.)
          END IF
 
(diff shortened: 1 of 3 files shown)


alazzaro:
Actually, this is F2008 standard, but most compilers are still not supporting it...
(see https://community.intel.com/t5/Intel-Fortran-Compiler/Optional-Arguments-and-Their-Uses/td-p/1008607 for some discussions)

fstein93:
Thank you for the clarification. I was not aware of it being a part of the F2008 standard. Unfortunately, the missing bits cause issues with other configurations than those tested on our dashboard. I have updated the initial comment accordingly.

alazzaro:
> Thank you for the clarification. I was not aware of it being a part of the F2008 standard. Unfortunately, the missing bits cause issues with other configurations than those tested on our dashboard. I have updated the initial comment accordingly.

We had the same problem in DBCSR and we ended up by using fypp to check all parameters...  See here https://github.com/cp2k/dbcsr/blob/2ab4080758361499b1761f7e49a074956b1cdedd/src/dbcsr_api_c.F#L579 for an example

fstein93:
Here, the DBCSR approach might a bit overkill which is why I keep it simple.

PR #2337, opened 2022-10-09
Docker: Upgrade to nvidia/cuda:11.8.0-devel-ubuntu20.04

oschuett:


oschuett:
Some [tests](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-pascal-b412455f_report.txt) segfault in `dbcsr/src/acc/libsmm_acc/libsmm_acc.cpp:362`.

@alazzaro, any idea what this might be?

Note that, this PR upgrades CUDA (11.3.1 -> 11.8.0) and Ubuntu (20.04 -> 22.04), which implies GCC (9.4 -> 11.2).

alazzaro:
Interesting... Do you have a way to test the new GCC alone? (with the old CUDA)

hfp:
Beside of GCC 12.x with CUDA 11.8 (with the latter able to leverage that {so far still] GCC broken version), it looks like something broke in the kernel's source code likely preprocessor decisions since `ACC_RTC(Result) compileResult = ACC_RTC(CompileProgram)(kernel_program, nOptions, compileOptions)` seems to bail out.

hfp:
This might be just CUDA bailing on current kernel sources...

alazzaro:
But we [dump the error and source code of the jitted kernel](https://github.com/cp2k/dbcsr/blob/2ab4080758361499b1761f7e49a074956b1cdedd/src/acc/libsmm_acc/libsmm_acc.cpp#L363). The fact that we force to use any GCC with CUDA can be a problem here... (via `-allow-unsupported-compiler`)

oschuett:
We get the [same result](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-pascal-2505e005_report.txt) with `nvidia/cuda:11.8.0-devel-ubuntu20.04`. So, it's the CUDA version. Would it help to bisect the CUDA version to know exactly when it broke?

alazzaro:
@oschuett I tried some of the failing tests on a system with CUDA 11.6 (the newest I can find). Note that the hardware is Ampere (the only I have access to). The tests do not crash... So the question is: could you try [11.6.0-devel-ubuntu20.04](https://hub.docker.com/layers/nvidia/cuda/11.6.0-devel-ubuntu20.04/images/sha256-dc0dd351f30911abd8ae3c39d140ca1f548f95cd4cb2d67816daa6eabec6f958?context=explore), so we get a common baseline?

alazzaro:
Uhm, still crashing with the same error... The error is always during the JIT phase, or in general when we call the host CUDA driver. Are we are sure that the CUDA libraries of the container is compatible with the host one?

oschuett:
I was able to reproduce this on my local machine using CUDA 11.8.0 both within the container and on the underlying host.

Without Docker it works fine. So, there seems to be indeed an issue between Docker and NVRTC.

To reproduce this locally I ran the following commands:
```
cd tools/docker
docker build -f Dockerfile.test_cuda_V100 --shm-size=1g ../../ 
```

alazzaro:
> I was able to reproduce this on my local machine using CUDA 11.8.0 both within the container and on the underlying host.
> 
> Without Docker it works fine. So, there seems to be indeed an issue between Docker and NVRTC.
> 
> To reproduce this locally I ran the following commands:
> 
> ```
> cd tools/docker
> docker build -f Dockerfile.test_cuda_V100 --shm-size=1g ../../ 
> ```

V100? I thought we had the problem on P100... Anyway, your testing is quite useful, it means it is something we can report to NVIDIA... One more request: could you test on your local machine incrementally 11.4, 11.5,...? Maybe we can get a minimal working version better than the current 11.3.1...

alazzaro:
OK, so 11.4.2 is the latest CUDA version we can use the container... interesting... I will investigate more in the future...

oschuett:
> V100? I thought we had the problem on P100...

Yes, at home I have GV100s, but in the cloud I'm using a P4 for the regtests to save some money.

> OK, so 11.4.2 is the latest CUDA version we can use the container... 

Note that with 11.4.3 there was only [a single failure](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-pascal-e96472ca_report.txt), while later version had ~35. Furthermore, on my local machine CUDA 11.4.1 also failed.

So, it all seems kinda fuzzy, which would suggest a race condition. Perhaps it only shows up within containers because they have a slower filesystem?

alazzaro:
Well, the code is out since a while (>4 years), people never reported any issue (neither we found on our regtests). Note that the error is:

```
CUDA DRIVER API ERROR: ModuleLoadDataEx failed with error CUDA_ERROR_INVALID_PTX (/opt/cp2k/exts/dbcsr/src/acc/libsmm_acc/libsmm_acc.cpp::180)
```

which means that the generated PTX for the CUDA RT is incompatible with the CUDA version used by the application. So, once more, you have to check that the cuda driver installed on your machine is compatible with the CUDA version (basically, the file `/usr/lib64/libcuda.so` symbolic link):

https://docs.nvidia.com/deploy/cuda-compatibility/index.html#deployment-consideration-forward

hfp:
> `CUDA DRIVER API ERROR: ModuleLoadDataEx failed with error CUDA_ERROR_INVALID_PTX (/opt/cp2k/exts/dbcsr/src/acc/libsmm_acc/libsmm_acc.cpp::180)`

It could be also something related to cached kernel code which usually resides under `~/.nv`. Perhaps worth deleting this directory just in case some HASH clashed.

oschuett:
> you have to check that the cuda driver installed on your machine is compatible with the CUDA version 

I've added a call to nvidia-smi. I think, the driver is recent enough:
```
| NVIDIA-SMI 470.82.01    Driver Version: 470.82.01    CUDA Version: 11.4     |
```

I also went back to `cuda:11.4.3`. The tests failed again, but in a [different spot](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-pascal-29e0d818_report.txt) than [last time](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-pascal-e96472ca_report.txt). So, this reinforces my feeling that we're dealing with a race condition.


What's also interesting is that the report contains errors from two locations: `libsmm_acc.cpp:165` and `libsmm_acc.cpp::180`. Since these two branches are mutually exclusive and within a [critical region](https://github.com/cp2k/dbcsr/blob/e330c82961a1e49f858aa566a81a8257e9ee1265/src/acc/libsmm_acc/libsmm_acc.cpp#L259), it seems the messages are from different MPI processes.

My guess would be that something goes wrong in the [JIT cache](
https://developer.nvidia.com/blog/cuda-pro-tip-understand-fat-binaries-jit-caching/). In the [past](https://forums.docker.com/t/docker-volume-file-permission-and-locking/24335) there was an issue with [file locking in overlayfs](https://lore.kernel.org/lkml/1474028371-21288-6-git-send-email-mszeredi@redhat.com).

So, I think the best workaround for now is to simply set `CUDA_CACHE_DISABLE=1`.

alazzaro:
> So, I think the best workaround for now is to simply set `CUDA_CACHE_DISABLE=1`.

And it solves the problem? This is quite interesting...

oschuett:
> > So, I think the best workaround for now is to simply set CUDA_CACHE_DISABLE=1.

> And it solves the problem?

Yes :-)

Unfortunately, with Ubuntu 2022.04 the build time of ELPA increases by 4x. Let's see if [upgrading](https://github.com/cp2k/cp2k/pull/2370) to ELPA 2022.05.001 helps.

oschuett:
Unfortunately, building ELPA 2022.05.001 with Ubuntu 2022.04 also [takes 40 minutes](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-pascal-a393b434_report.txt), which is too much.

So, I'll only upgrade CUDA and keep the old Ubunutu for now.

alazzaro:
I agree with your conclusion.
I assume the longer time on the new ubuntu is due to the new compiler, likely lowering the  optimization will speed up the compilation.

oschuett:
> I assume the longer time on the new ubuntu is due to the new compiler, likely lowering the optimization will speed up the compilation.

Yes, that's probably it. I'll merge this PR now and put the Ubunutu upgrade on my todo list.

PR #2341, opened 2022-10-10
RI-HFX| more consistent GPU memory usage SCF/force

abussy:
The 3-center tensors involved in the force calculations with RI-HFX are typically denser than their SCF counterpart. To insure that the memory usage is consistent between SCF and forces (especially for GPU), an additional layer of batching was added to the latter. This is already done in low-scaling post-HF methods.

### DIFF
--- a/src/hfx_ri.F
+++ b/src/hfx_ri.F
@@ -2414,13 +2414,13 @@ CONTAINS
       CHARACTER(LEN=*), PARAMETER :: routineN = 'hfx_ri_forces_Pmat'
 
       INTEGER                                            :: dummy_int, handle, i_mem, i_spin, i_xyz, &
-                                                            j_mem, j_xyz, k_xyz, n_mem, natom, &
-                                                            unit_nr_dbcsr
+                                                            j_mem, j_xyz, k_mem, k_xyz, n_mem, &
+                                                            n_mem_RI, natom, unit_nr_dbcsr
       INTEGER(int_8)                                     :: nflop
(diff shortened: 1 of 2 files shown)


oschuett:
This change almost doubled the runtime of the [RI-HFX_H2O-32](https://github.com/cp2k/cp2k/blob/master/benchmarks/QS_single_node/RI-HFX_H2O-32.inp) benchmark on [GPU](https://dashboard.cp2k.org/archive/perf-cuda-volta/index.html):

image

@abussy, could you please take a look. Perhaps, this optimization is only advantageous for larger systems?

abussy:
This goes back to a previous comment that I made for PR #2006. To unify the input convention of RI-HFX and low-scaling post-HF methods, I changed the meaning of the MEMORY_CUT keyword for RI-HFX. I also forgot to change the value of the MEMORY_CUT keyword in the input to get consistent behavior (it should have been set to 3 at the time, instead of 9).

At the time, we already noticed a drop in performance because we went from a batching scheme that went from MEMORY_CUT individual contractions to one with MEMORY_CUT^2 individual contractions. So changing the value of the keyword from 9 to 3 would have fixed that.

Now that I noticed that forces require even more batching (because tensors are denser), we are facing some cases where there are MEMORY_CUT^3 individual contractions. With MEMORY_CUT set to 9, that's 729  contractions of tiny tensors.

Essentially, that would be fixed by using a reasonable and realistic value for MEMORY_CUT (default is 3). Here timings blow up because for a system of that size, using such a large memory factor introduces way too much overhead. Of course, this goes back to me forgetting to adapt the input of the benchmark.

PR #2348, opened 2022-10-13
Fix conventions

fstein93:
@mattiatj Is this what you intended? (If you declare a variable in Fortran and assign it a value in the declaration block, the content of the variable will be saved for the next call. Thus, it behaves like a static variable in C/C++.)


mattiatj:
Yes, thanks for fixing!

PR #2352, opened 2022-10-14
Fix undefined reference to pw_release

mkrack:


oschuett:
I really wonder why this wasn't caught by the `sdbg` test?

mkrack:
I asked myself the same. The gcc based arch files `minimal.sdbg` and `local_coverage.sdbg` as well as the Intel arch file detected the issue, the others not. That's indeed strange, no clue.

PR #2356, opened 2022-10-17
Fix missing library with HIP

fstein93:
This PR fixes the missing TiledMM package required if COSMA is compiled for AMD GPUs using HIP but not with CUDA.

### DIFF
--- a/tools/toolchain/scripts/generate_arch_files.sh
+++ b/tools/toolchain/scripts/generate_arch_files.sh
@@ -219,7 +219,7 @@ if [ "${ENABLE_HIP}" = __TRUE__ ] && [ "${GPUVER}" != no ]; then
       check_lib -lcufft "cuda"
       check_lib -lcublas "cuda"
       DFLAGS+=" IF_HIP(-D__HIP_PLATFORM_NVIDIA__ -D__HIP_PLATFORM_NVCC__ -D__OFFLOAD_HIP |) -D__DBCSR_ACC"
-      HIP_FLAGS=" -g -arch sm_${ARCH_NUM} -O3 -Xcompiler='-fopenmp -Wall -Wextra -Werror' --std=c++11 \$(DFLAGS)"
+      HIP_FLAGS+=" -g -arch sm_${ARCH_NUM} -O3 -Xcompiler='-fopenmp -Wall -Wextra -Werror' --std=c++11 \$(DFLAGS)"
       add_include_from_paths CUDA_CFLAGS "cuda.h" $INCLUDE_PATHS
       HIP_INCLUDES+=" -I${CUDA_PATH:-${CUDA_HOME:-/CUDA_HOME-notset}}/include"
       # GCC issues lots of warnings for hip/nvidia_detail/hip_runtime_api.h
--- a/tools/toolchain/scripts/stage4/install_cosma.sh
+++ b/tools/toolchain/scripts/stage4/install_cosma.sh
@@ -98,7 +98,7 @@ case "$with_cosma" in
       fi
 
       # Build HIP version.
-      if [ "$ENABLE_HIP" = "__TRUE__" ]; then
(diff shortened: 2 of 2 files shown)


alazzaro:
Since you are there, you need to adapt the cosma script for HIP. Currently it only supports [CUDA](https://github.com/cp2k/cp2k/blob/8d96446bea90f37d3e25c2492dcc34fe4f653cd4/tools/toolchain/scripts/stage4/install_cosma.sh#L79)

fstein93:
I am a bit confused. I can compile it that way on LUMI and the Rocm tester passes. Somehow, the Tiled-MM library went missing in the HIP Pascal Regtest. I am currently cross-checking the flags with the tester on the Dashboard.

alazzaro:
The CP2K HIP version gets compiled in COSMA with rocblas, which is not installed if you don't have ROCM. Indeed, this is the case of the HIP on NVIDIA device test that it is failing. In that particular case, we cannot build cosma with ROCM.

```
COSMA-v2.6.2.tar.gz: OK
Checksum of COSMA-v2.6.2.tar.gz Ok
Installing from scratch into /opt/cp2k-toolchain/install/COSMA-2.6.2
-- Selected BLAS backend for COSMA: ROCM
-- Selected SCALAPACK backend for COSMA: CUSTOM
-- The BLAS backend that SCALAPACK is linked to: CUSTOM
-- The CXX compiler identification is GNU 9.4.0
-- The C compiler identification is GNU 9.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found MPI_C: /usr/lib/x86_64-linux-gnu/libmpich.so (found version "3.1") 
-- Found MPI_CXX: /usr/lib/x86_64-linux-gnu/libmpichcxx.so (found version "3.1") 
-- Found MPI: TRUE (found version "3.1")  
-- Selected TILEDMM_GPU_BACKEND: ROCM
CMake Error at /opt/cp2k-toolchain/install/cmake-3.24.2/share/cmake-3.24/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find HIPLIBS (missing: HIPLIBS_HIP_LIBRARY HIPLIBS_HSA_LIBRARY
  HIPLIBS_THUNK_LIBRARY HIPLIBS_HSA_INCLUDE_DIRS)
Call Stack (most recent call first):
  /opt/cp2k-toolchain/install/cmake-3.24.2/share/cmake-3.24/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  libs/Tiled-MM/cmake/FindHIPLIBS.cmake:90 (find_package_handle_standard_args)
  libs/Tiled-MM/CMakeLists.txt:45 (find_package)


-- Configuring incomplete, errors occurred!
See also "/opt/cp2k-toolchain/build/cosma/build-cuda/CMakeFiles/CMakeOutput.log".
make: *** No targets specified and no makefile found.  Stop.
make: *** No rule to make target 'install'.  Stop.
Step cosma took 25.00 seconds.
```

alazzaro:
Note that @oschuett asked to support [hipblas in COSMA](https://github.com/eth-cscs/COSMA/issues/92).
At this point we have to introduce a case where we want to use HIP with ROCM (instead of HIP with CUDA) to get COSMA (and possibly other libraries) to compile...

alazzaro:
I think we can add a check if the library rocblas exists via

```
check_lib -lrocblas "rocm"
```

in the if, i.e. something like

```
if [ "$ENABLE_HIP" = "__TRUE__" ] && `check_lib -lrocblas "rocm" &> /dev/null` ; then 

```

PR #2364, opened 2022-10-19
Build cp2k with cmake

mtaillefumier:
Note 1 : can we avoid running the full CI for the time being ?
Note 2: the two build systems will coexist for at least one or two versions.

This PR is long overdue but this work reaches the point where it become usable to build cp2k with cmake. I kept the building process as simple as possible so that other devs can look around and tinker with it a little bit. 

This build system does only one task, configuring and compiling cp2k. The dependencies should be compiled separately as a first step. 

I am also fully aware of that it will not necessarily work for everyone the first time that's why I am asking devs, experienced users and package managers to test it and report issues as much as they can. I will open an issue on the cp2k github repo to regroup all of them in one place.

so far all dependencies should be detected with fluctuating success. The mandatory dependencies are 
- DBCSR
- BLAS / LAPACK / SCALAPACK implementation
- openmp
- c, c++, fortran compilers 

libxc, libint, cosma, libxsmm, fftw are turned on by default. All the other dependencies are off
All options are prefixed with CP2K_. This rule is enforced for the build system because it can potentially interfere with others options and variables names defined by the dependencies cmake config files (BLAS in particular). The -D__OPTIONS are unchanged. 

the build system supports CUDA and HIP (as much as HIP goes expect some issues)

how to compile cp2k : 

- 'MKL' with 'CUDA' (fftw, mpi, libxc, cosma, libint2, libxsmm should be installed somewhere)

`cmake -DCMAKE_INSTALL_PREFIX=/myprefix -DCP2K_BLAS_VENDOR=MKL -DCP2K_SCALAPACK_VENDOR=MKL -DCP2K_USE_ACCEL=CUDA -DCP2K_WITH_GPU=V100 ..
`
- MKL without CUDA : remove the -DCP2K_USE_ACCEL 

- MKL with HIP replace CUDA with HIP and set -DCP2K_WITH_GPU=Mi250 (of instance). Note that these options have default values

- Cray environments

 - with libsci and cuda support : 'MPICC=cc MPICXX=CC MPIFORT=ftn cmake -DCMAKE_INSTALL_PREFIX=/myprefix -DCP2K_BLAS_VENDOR=SCI -DCP2K_SCALAPACK_VENDOR=SCI -DCP2K_USE_ACCEL=CUDA -DCP2K_WITH_GPU=V100 ..`
 - with MKL see the mkl section because it is exactly the same command line.

list of known potential issues

- Configuring cp2k with ROCM 5.0.x is likely to fail. There is a bug in the rocm cmake infrastructure that is hard to go around
- Compiling cp2k woth ROCM 5.2.x will work but DBCSR won't work properly because of a bug in the rocm jiting functionality. It should be fixed in ROCM 5.3.x (so use it if you can)
- The current DBCSR build system is missing one important line that makes the cp2k build system fails during link time if DBCSR is compiled with libxsmm support. I opened a PR that fix this issue.
- gfortran does not seem to have a default include directory but it is only an issue if cp2k is installed in /usr (on unix)
- All dependencies should be compiled with fortran support (if they generate the .mod files). I will add a check for this peculiar point eventually.
- CUDA : the Nvidia hpc sdk install the cuda math libraries in a different directory that cuda. This can be source of errors when cmake is searching for cublas or cufft. 
- The generic detection of blas and lapack only cover the most important implementations of blas and lapack. It might still fail though because naming convention for these libraries is fluctuating.
- Cray libsci : dependencies might compile and link with libsci_mp.so vairants while cp2k will link with libsci.so by default. It will generate a warning at the end of cp2k execution. The solution is to pass -fopenmp during link time.
- some of the FindPackage.cmake files are rudimentary.
- Summit cluster. We need a custom command line to make it work. The only issue is really blas and lapack on this system.
 
missing things and improvements

- [ ] Add comments in the cmake build scripts
- [x] test the cp2k installation itself (follows the default unix layout, data are set to /myprefix/share/cp2k/data)
- [ ] Include the possibility to run the regtests with a simple make tests
- [x] Only include units files that need to be compiled when the option is on. It is partially done at that stage (we will have to delay this otherwise the current build system will break, I do not want that)
- [x] Fix the VERSION file
- [x] rename README.cmake to something more convenient (prettify fails)
- [x] review compilations options (maybe add some options for devs turn on/off some warnings)
- [x] add check for the module files. (for instance libint2.mod is missing on Fedora)


oschuett:
This is awesome! Adding a new build system to a grown code base is no easy feat. Great work!

> Note 1 : can we avoid running the full CI for the time being ?

Yes, we can merge this PR without tests. Over the coming weeks we can then migrate one CI test at a time.

I noticed that many of the files have a BSD license from ETH. Could you reach out to the original authors and ask if they'd be ok with releasing these files directly under GPL?

mtaillefumier:
> This is awesome! Adding a new build system to a grown code base is no easy feat. Great work!
> 
> > Note 1 : can we avoid running the full CI for the time being ?
> 
> Yes, we can merge this PR without tests. Over the coming weeks we can then migrate one CI test at a time.

I want to wait a little so that devs (mostly) try it and tell me what is failing. I will work on the hip side further (the compilation is not the issue, nor is cmake but rocm is). 

Additional note : For internal reason, we are using spack + cp2k + cmake to actually build cp2k (with the default dependencies). 
 
> I noticed that many of the files have a BSD license from ETH. Could you reach out to the original authors and ask if they'd be ok with releasing these files directly under GPL?

I blindly copy pasted the license agreement we use almost everywhere but i do not think GPL is an issue at least not on my side.
Except FindMKL.cmake where I modified our internal version to better suit findBLAS.cmake all the others are my own writing. Tiziano should also agree to this request but we can switch to GPL in my point of view (FindMKL.cmake might still stay BSD I will ask the author).

CMakeLists.txt
```diff
@@ -0,0 +1,822 @@
...
+    ${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}${VERSION_GIT})
+set(cp2k_APIVERSION ${VERSION_MAJOR}.${VERSION_MINOR})
+
+if(NOT DEFINED CMAKE_CUDA_STANDARD)
+  set(CMAKE_CUDA_STANDARD 11)
+  set(CMAKE_CUDA_STANDARD_REQUIRED ON)
+endif()
+
+# set language and standard
+set(CMAKE_CXX_STANDARD 11)
```
msimberg, line 68: I would recommend using `target_compile_features(... cxx_std_11)` instead (https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#id5).

CMakeLists.txt
```diff
@@ -0,0 +1,822 @@
...
+# =================================================================================================
+# OPTIONS
+include(CMakeDependentOption)
+
+# make MPI optional as well. More work but might be useful set(CP2K_USE_MPI ON
+# CACHE BOOL "Enable MPI support")
+
+set(CP2K_SET_POSITION_INDEPENDENT_CODE
+    ON
+    CACHE BOOL "Enable position independent code")
```
msimberg: I would recommend using `option` instead for all the `BOOL` options here (https://cmake.org/cmake/help/latest/command/option.html?highlight=option)
```suggestion
option(CP2K_SET_POSITION_INDEPENDENT_CODE
    ON
    "Enable position independent code")
```
It basically lets you omit the `CACHE` and `BOOL` (it's otherwise equivalent).

CMakeLists.txt
```diff
@@ -0,0 +1,822 @@
...
+    "GENERIC"
+    CACHE STRING "scalapack vendor/generic backend")
+set_property(CACHE CP2K_SCALAPACK_VENDOR PROPERTY STRINGS
+                                                  ${CP2K_SCALAPACK_VENDOR_LIST})
+
+if(NOT ${CP2K_SCALAPACK_VENDOR} IN_LIST CP2K_SCALAPACK_VENDOR_LIST)
+  message(FATAL_ERROR "Invalid scalapack vendor backend")
+endif()
+
+set(CP2K_SUPPORTED_ACCELEARTION_TARGETS CUDA HIP NONE)
```
msimberg:
```suggestion
set(CP2K_SUPPORTED_ACCELERATION_TARGETS CUDA HIP NONE)
```

CMakeLists.txt
```diff
@@ -0,0 +1,822 @@
...
+        CACHE BOOL "Use DBCSR cpu only version")
+    # set(CP2K_USE_HIP_CUDA OFF CACHE BOOL "Compile hip code for Nvidia gpus.
+    # Can be used for debugging hip code when hardware is not available")
+  endif()
+endif()
+
+set(CP2K_USE_HIP OFF)
+set(CP2K_USE_CUDA OFF)
+
+set(CMAKE_POSITION_INDEPENDENT_CODE ${CP2K_SET_POSITION_INDEPENDENT_CODE})
```
msimberg, line 209: Is there a benefit to having `CP2K_SET_POSITION_INDEPENDENT_CODE` instead of just setting `CMAKE_POSITION_INDEPENDENT_CODE` directly as a user?

CMakeLists.txt
```diff
@@ -0,0 +1,822 @@
...
+
+# =================================================================================================
+# LANGUAGES AND TESTING
+
+enable_language(CXX)
+enable_language(C)
+enable_language(Fortran)
```
msimberg: These are already enabled in the `project` call at the beginning, or is this required because of some CMake quirk?

CMakeLists.txt
```diff
@@ -0,0 +1,822 @@
...
+
+# get the git hash Get the latest abbreviated commit hash of the working branch
+execute_process(
+  COMMAND git log -1 --format=%h
+  WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
+  OUTPUT_VARIABLE CP2K_GIT_HASH
+  OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+execute_process(
+  COMMAND hostnamectl --transient
```
msimberg, line 236: Does this imply `systemd` and does `hostname` return the same?

CMakeLists.txt
```diff
@@ -0,0 +1,845 @@
...
+    list(APPEND CMAKE_PREFIX_PATH $ENV{__var})
+    set(ROCM_PATH
+        $ENV{__var}
+        CACHE PATH "Path to ROCm installation")
+  endif()
+endforeach()
+
+# =================================================================================================
+# OPTIONS
+include(CMakeDependentOption)
```
msimberg: Are you using this?

CMakeLists.txt
```diff
@@ -0,0 +1,845 @@
...
+
+# make MPI optional as well. More work but might be useful set(CP2K_USE_MPI ON
+# CACHE BOOL "Enable MPI support")
```
msimberg: I guess this should be uncommented or removed?

CMakeLists.txt
```diff
@@ -0,0 +1,845 @@
...
+  set(CMAKE_HIP_ARCHITECTURES gfx801 gfx900 gfx90a)
+  if(NOT CMAKE_BUILD_TYPE)
+    set(HIP_RELEASE_OPTIONS "-O3 -g -DNDEBUG")
+  elseif(${CMAKE_BUILD_TYPE} STREQUAL "RelWithDebInfo")
+    set(HIP_RELEASE_OPTIONS "-O3 -g -DNDEBUG")
+  elseif(${CMAKE_BUILD_TYPE} STREQUAL "Release")
+    set(HIP_RELEASE_OPTIONS "-O3 -DNDEBUG")
+  elseif(${CMAKE_BUILD_TYPE} STREQUAL "Debug")
+    set(HIP_RELEASE_OPTIONS "-O0 -g")
+  endif()
```
msimberg, line 370: These should probably be consistent with the available `CMAKE_BUILD_TYPES` defined starting on line 562?

CMakeLists.txt
```diff
@@ -0,0 +1,845 @@
...
+find_package(DBCSR 2.2.0 REQUIRED)
+
+# =================================================================================================
+# PACKAGE DISCOVERY (compiler configuration can impact package discovery)
+
+find_package(OpenMP REQUIRED COMPONENTS Fortran C CXX)
+
+# ==================================
+if(CP2K_USE_ELPA)
+  set(CP2K_ENABLE_ELPA_OPENMP_SUPPORT
```
msimberg: I guess you could use `CmakeDependentOption` for this?

CMakeLists.txt
```diff
@@ -0,0 +1,845 @@
...
+# cmake build -") message(STATUS "- -") message(STATUS
+# "--------------------------------------------------------------------")
+
+# message(STATUS "- Use CP2K_PREFIX/share for storing cp2k data") message(STATUS
+# "- Integrate the regtests") message(STATUS "- Use -DCP2K_OPTION_NANE instead
+# of -D__OPTION_NAME as it is reserved to compilers") message(STATUS "- create
+# the package file") message(STATUS "- clean up the find_packages and make them
+# more uniform") message(STATUS "- restore hip/cuda functionality")
+# message(STATUS "- fix findblas and findlapack for openblas/arm....")
+# message(STATUS "    ")
```
msimberg: Remove?

CMakeLists.txt
```diff
@@ -0,0 +1,845 @@
...
+  if(CP2K_USE_VDWXC)
+    message(STATUS "     - VDWXC")
+  endif()
+endif()
+
+if(CP2K_USE_COSMA)
+  message(STATUS " ")
+  message(STATUS "- COSMA")
+  message(STATUS "   - include directories : " ${COSMA_INCLUDE_DIRS})
+  # message(STATUS "  - libraries           : " ${COSMA_LIBRARIES})
```
msimberg: Remove or uncomment?

CMakeLists.txt
```diff
@@ -0,0 +1,845 @@
...
+
+if(NOT CP2K_USE_SIRIUS)
+  message(STATUS "   - SIRIUS")
+endif()
+
+if(NOT CP2K_USE_SPGLIB)
+  message(STATUS "   - SPGLIB")
+endif()
+
+# if (NOT CP2K_USE_MPI) message(STATUS "   - MPI") endif()
```
msimberg: Remove or uncomment?

cmake/CompilerConfiguration.cmake
```diff
@@ -0,0 +1,123 @@
+if(CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
+  set(CMAKE_Fortran_FLAGS
```
msimberg, line 2: I _think_ all of these `CMAKE_*_FLAGS` definitions will override any user-defined `CMAKE_*_FLAGS`. I would recommend either appending to the cache variables or again using targets and `target_compile_flags` (together with generator expressions for the various conditionals).

src/CMakeLists.txt
```diff
@@ -0,0 +1,1613 @@
...
+
+#
+# apps included in dbm
+#
+
+add_executable(dbm_miniapp dbm/dbm_miniapp.c ${DBM_SRCS_C} ${OFFLOAD_SRCS_C}
+                           ${DBM_SRCS_GPU})
+
+set_target_properties(dbm_miniapp PROPERTIES LINKER_LANGUAGE C)
+# target_compile_options(dbm_miniapp PUBLIC "${OpenMP_C_FLAGS}")
```
msimberg: Remove or uncomment? I think the `OpenMP::OpenMP_<lang>` target should anyway be enough for this.

msimberg (commented):
Very nice to see this! I've added a few comments that would make this a bit more idiomatic "modern" CMake, but none of the comments are blockers.

msimberg:
Summarizing my review comments:
- [ ] Remove or uncomment commented out code
- [ ] Change `CMAKE_*_FLAGS` to `target_compile_flags`
- [ ] Consistent `CMAKE_BUILD_TYPE`s
- [ ] Use `option` for boolean options
- [ ] Use `cmake_dependent_option` for dependent options
- [ ] systemd dependency for `hostnamectl` needed?
- [ ] Use `target_compile_features` where possible (`cxx_std_11`)
- [ ] Unnecessary `enable_language` (since set in `project`)?
- [ ] Necessary to have separate `POSITION_INDEPENDENT_CODE` variable instead of using the standard CMake variable?
- [ ] Fix typo in `CP2K_SUPPORTED_ACCELEARTION_TARGETS`

CMakeLists.txt
mtaillefumier: removed

CMakeLists.txt
mtaillefumier: sovled

CMakeLists.txt
mtaillefumier: applied to all languages that supports it.

CMakeLists.txt
mtaillefumier: removed

CMakeLists.txt
mtaillefumier: I do now.

CMakeLists.txt
mtaillefumier: removed

CMakeLists.txt
mtaillefumier: I need to review this in more details before marking it as resolved. the compilation options need cleanup

CMakeLists.txt
mtaillefumier: solved. Extended the use of cmake_dependent_options to all packages that needed it,

cmake/CompilerConfiguration.cmake
mtaillefumier: I kept the initial version for the time being but I will work on this to make simpler.

mtaillefumier:
It is time to get this PR in the wild. Please quash all commits when the PR is merged as it is not worth keeping my internal logic.

- the build works with CUDA, ROCM 5.3.x, MKL, SCI, OpenBLAS, cray-mpi, openmpi (and probably with mpich). The only dependencies that need more testing are QUIP, PEXSI, SuperLU. It is mostly a matter of detecting the libraries properly.
- All others dependencies, libint2, libxsmm, libxc, cosma, elpa, dbcsr, SIRIUS, libvori (a missing function is showing up at link time) work fine provided that cmake can find them. 
- I added a debug mode with the possibility to turn off some of the GPU features. It is turned off by default
- The VERSION is generated automatically but it will need some more work to distinguish between dev and official versions 
- More importantly there is **NO** modification of the source code
- CP2K installation follows the default CP2K_PREFIX/{bin,lib,share}.
- Data for pseudo potentials are stored in CP2K_PREFIX/share/data by default.
- the compilation options still need some work.

Happy hacking.

Issue #1860, opened 2022-01-06
Re-enable OpenMP for ELPA on Cray


oschuett:
The toolchain currently builds [ELPA without OpenMP on Cray](https://github.com/cp2k/cp2k/blob/cdcb2793be4c6302478245ba22b1e8c038bd7787/tools/toolchain/scripts/stage5/install_elpa.sh#L84). While this is ok for the current XC40/50 generation, we'll eventually need OpenMP support for EPYC and ARM based machines.

See also #1857.

oschuett:
There have been a number of improvements to ELPA's OpenMP code. @mkrack, perhaps we should revisit this ?

PR #2370, opened 2022-10-25
Toolchain: Upgrade to ELPA 2022.05.001

oschuett:


mtaillefumier:
you save me quite a lot of time with this PR. i cross my finger that this single update will solve the issues I have on cray systems at the moment.

oschuett:
Fingers crossed :-)

oschuett:
BTW, we're currently disabling OpenMP for ELPA on Cray machines (#1860). Perhaps, this upgrade also resolves that issue?

mtaillefumier:
Unfortunately this update does not solve the issue I am observing on cray systems. I have a user input that triggers a series of mpi failure deep inside ELPA and cray-mpi. So it is difficult for me to test if the issue reported in #1860 is fixed until the cray mpi stack is working.

PR #2382, opened 2022-11-04
Refactoring

fstein93:
This PR
- removes a lot of id_nr and ref_count component, related retain routines and removes POINTER attributes from their basic routines. A bunch of types are still missing.
- Switches pw_p_type to pw_types in case of derivatives of the Hartree potential.
- removes a lot of unused routineP parameters.

Some observations/hints for others:
1. If a retain routine is never used, remove the respective ref_count.
2. Many setters make use of retain+release. Then, the call to retain can be removed and the release can be wrapped in a double branch like. Do not forget to remove the corresponding release of the variable of the callee of the setter. The general structure
`call obj_retain(new_ptr)`
`call obj_release(obj%ptr)`
`obj%ptr => new_tr`
is replaced by something like
`if (associated(obj%ptr)) then`
`  if (.not. associated(obj%ptr, new_ptr)) then`
`    call obj_release(obj%ptr)`
`  end if`
`end if`
`obj%ptr => new_tr`
The employed two-argument version of the associated function is part of the Fortran 90 standard and already occassionally used in the code. (Until then, I had not known about it either.)
3. I agree with @oschuett that it is probably possible to drop reference counting entirely, but I would like to keep it in certain cases: Some derived types contain meta data (like cp_fm_struct or pw_grid_type) which is the same for many containers (cp_(c)fm_type, pw_type). Unfortunately, this meta data can be more memory consuming than the actual data (especially problematic with pw grids). Plus, memory pools require data chunks of the same size (cp_fm_pool_type, pw_pool_type).
4. If a routine starts with CPASSERT(ASSOCIATED(some_pointer)), check whether you actually need a pointer.


oschuett:
Oh wow - great work!

>  but I would like to keep it in certain cases: Some derived types contain meta data (like cp_fm_struct or pw_grid_type)

Yes, removing the reference counting from those broadly shared objects is more difficult. We'll have to define an owner for each of them. This in turn requires a consistent startup / shutdown sequence (which we currently lack). The first step would be to disentangle [f77_interface.F](https://github.com/cp2k/cp2k/blob/master/src/f77_interface.F). I'm confident that we'll get there eventually though.

In any case, this PR is already a huge leap!

Issue #1855, opened 2022-01-04
Some regtests fail on ARM (aarch64)


Tom-python0121:
Hello, I meet a problem:cp2k@9.1  test failed on centos8_aarch64.
### Steps to reproduce the issue

```console
[root@localhost ~]git clone --recursive https://github.com/cp2k/cp2k.git cp2k
[root@localhost ~]cd cp2k
[root@localhost cp2k]#cd tools/toolchain
[root@localhost cp2k]#./install_cp2k_toolchain.sh --with-libxsmm=no  --with-libvori=no
[root@localhost cp2k]#cd ../..
[root@localhost cp2k]#cp tools/toolchain/install/arch/* arch/
[root@localhost cp2k]#source /tools/toolchain/install/setup
[root@localhost cp2k]#yum install liblsan.aarch64 gcc-toolset-10-liblsan-devel.aarch64
[root@localhost cp2k]#make -j 128 ARCH=local VERSION="ssmp sdbg"
[root@localhost cp2k]#make -j 128 ARCH=local VERSION="ssmp sdbg" test
……
    cc2_lri.inp                                                                    OK (  11.08 sec)
<<< /home/stage/root/spack-stage-cp2k-9.1-d2yemk6wqui237fwkyhdwezq6yl3jt5c/spack-src/regtesting/TEST-local-ssmp-2022-01-04_17-02-45/QS/regtest-kp-2 (286 of 286) done in 382.49 sec
------------------------------- Errors ---------------------------------
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
/home/stage/root/spack-stage-cp2k-9.1-d2yemk6wqui237fwkyhdwezq6yl3jt5c/spack-src/regtesting/TEST-local-ssmp-2022-01-04_17-02-45/QS/regtest-gw-cubic/G0W0_kpoints_from_Gamma.inp.out
Difference too large: 7.56e-02 > 1e-05, ref_value: 14.55, value: 13.45.
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
[...]

PR #2393, opened 2022-11-13
Docker: Add test for ARM64

oschuett:


hfp (commented):
It would be nice to start testing LIBXSMM on AArch64 at some point. We have not released LIBXSMM 2.0, but support as needed by CP2K is ready since quite some time. Though, it requires to rely on our main branch. To decouple from our daily development, I recommend for instance `6f883620f58afdeebab28039fc9cf580e76a5ec6`.

oschuett:
Thanks for the info @hfp! Do you have a ETA for libxsmm 2.0? We still have a number of other issues to work out, e.g. #1855. So, perhaps the simplest solution is to just wait for the official release?

mkrack:
Thanks @oschuett. It is great that we have now an ARM64 tester which can reproduce the issues with the test inputs  in `QS/regtest-gw-kpoints` faster detected by [this tester](https://dashboard.cp2k.org/archive/gnu-aarch64-psmp/index.html) for quite a while already. I will retire that slow regression tester.

hfp:
> Thanks for the info @hfp! Do you have a ETA for libxsmm 2.0? We still have a number of other issues to work out, e.g. #1855. So, perhaps the simplest solution is to just wait for the official release?

We plan 2.0 for next year. Though, it slipped since we ever said "2.0" (1+ years). This is because of our research goals are somewhat different from just delivering predictable release cycles. Anyway, our cleanup work made good progress and I can see doing a release next year regardless of some pending items (if not closed over the year's end). Functionality (or even correctness) as of current releases (CP2K, DBM, DBCSR, etc.) is not expected to be related to any remaining 2.0 goals (libxsmm), hence one can just take for instance the earlier mentioned SHA.

hfp:
LIBXSMM's internal dependencies changed and using the Fortran interface now requires to link `xsmm`, `xsmmext`, and `xsmmf` (this order). I hope to get back to the previous dependencies (not "ext"), but even if previous deps are back, linking all of them should not harm.

oschuett:
Thanks for the link order @hfp. Unfortunately, I'm getting segfaults, e.g. when running `dbt_unittest`. So, I'll leave libxsmm out of this PR.

oschuett:
> It is great that we have now an ARM64 tester which can reproduce the issues with the test inputs in QS/regtest-gw-kpoints

Yes, it's good that the problem is so well reproducible.

> detected by [this tester](https://dashboard.cp2k.org/archive/gnu-aarch64-psmp/index.html) for quite a while already.

Indeed! Thanks for doing all the hard ground work. I was very impressed when the toolchain simply worked on ARM.

> I will retire that slow regression tester.

Go ahead. The new tester is up and running. For now ARM instances are still somewhat scarce on GCP (max 16 core per region), but this will certainly improve. Then we can also start to monitor performance.

PR #2404, opened 2022-11-20
Add Apple M1 (macOS) regression tester

mkrack:


tools/dashboard/dashboard.conf
```diff
@@ -247,13 +249,21 @@ host:        GCP
...
 sortkey:     702
 name:        Linux-gnu-aarch64.psmp (11.2.0)
 host:        Raspberry Pi 4
 report_url:  http://www.cp2k.org/static/regtest/trunk/Linux-gnu-aarch64/latest-mpich-psmp/regtest-0
 info_url:    http://www.cp2k.org/static/regtest/trunk/Linux-gnu-aarch64/latest-mpich-psmp/index.html
 
+[darwin-gnu-arm64-psmp]
+sortkey:     703
+name:        Darwin-gnu-arm64.psmp
+host:        Apple M1 (macOS)
```
oschuett: I think we should mention `MacStadium` here since they are sponsoring us.

mkrack: Right, will add that.

PR #2407, opened 2022-11-21
Fix blocksizes in ELPA

fstein93:
This PR enforces the block size to be a power of 2 (also on CPU) if an eigenvalue decomposition with ELPA is requested.

### DIFF
--- a/src/fm/cp_fm_diag_utils.F
+++ b/src/fm/cp_fm_diag_utils.F
@@ -297,7 +297,7 @@ CONTAINS
 #if defined(__SCALAPACK)
       REAL(KIND=dp)                            :: fake_local_data(1, 1)
       INTEGER                                  :: fake_descriptor(9), mepos_old, &
-                                                  io_unit, ngroups, ncol_block
+                                                  io_unit, ngroups, ncol_block, blksize, nrow_block
       TYPE(cp_fm_struct_type), POINTER         :: fm_struct_new
       TYPE(cp_para_env_type), POINTER          :: para_env
       TYPE(cp_logger_type), POINTER            :: logger
@@ -326,6 +326,8 @@ CONTAINS
       para_env => matrix%matrix_struct%para_env
       mepos_old = para_env%mepos
       ncol_block = -1 ! normally we also want to adjust the block size according to the optimal # of CPUs
+      nrow_block = -1
+      blksize = -1
 
       rdinfo%matrix_order = matrix%matrix_struct%nrow_global
       rdinfo%num_pe_old = para_env%num_pe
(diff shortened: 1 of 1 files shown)


src/fm/cp_fm_elpa.F
```diff
@@ -75,33 +81,50 @@ MODULE cp_fm_elpa
...
+   INTEGER, DIMENSION(18 &
+#if defined (__ELPA_NVIDIA_GPU)
+                      + 1 &
+#endif
+#if defined (__ELPA_AMD_GPU)
+                      + 1 &
+#endif
+#if defined (__ELPA_INTEL_GPU)
+                      + 1 &
+#endif
```
oschuett: Why is this needed? AFAIK, those constants are always available - even when the GPU kernels are not.

oschuett:
It would be awesome if this fixes #464!

src/fm/cp_fm_elpa.F
fstein93: On our cluster, ELPA_2STAGE_REAL_INTEL_GPU_SYCL is not available. Or, at least, I receive compiler errors due the missing symbol ELPA_2STAGE_REAL_INTEL_GPU_SYCL in elpa_constants.  The other two are explicitly checked later in the code. That's why I added the respective checks, too.
The issue is that you have to provide the length of fixed-size array explicitly. In that fashion, I can circumvent compilation issues because the compiler can obviously determine this size at compile-time.

oschuett: The `ELPA_2STAGE_REAL_INTEL_GPU_SYCL` kernel was added only recently. So, you are probably using an older version of ELPA. Given that ELPA had a number of severe issues in the past, I'm not sure we should actually support backwards compatibility.

fstein93: Then, we can actually remove these kind of checks.

fstein93: I removed the respective commit.

fstein93:
I have rerun some of the failing calculations on Daint and obtained the correct results. Do you think it is reasonable to employ Scalapack instead of ELPA in all failing tests? I know for sure that some passing tests (mostly from RPA/MP2) definitely make use of eigenvalue decompositions.

oschuett:
>  Do you think it is reasonable to employ Scalapack instead of ELPA in all failing tests?

No, we should really get to the bottom of this. Otherwise I'd have very little trust in the results - despite some tests occasionally passing. Fortunately, most of the tests failed because ELPA itself returned an error code. So, those should be rather straightforward to investigate.

oschuett:
The timeouts are probably due to https://github.com/marekandreas/elpa/issues/17.
I guess, until this point we never actually used ELPA's GPU kernels.

fstein93:
Apart from two special cases which require results by other regtests and the timeouts, there is only one kind of error in the bandred code of ELPA. I will consult the ELPA developers for that. It might take some time.

fstein93:
Apparently, the error is related to the 2stage solver because we observe timeouts if we employ the 1stage solver instead.

oschuett:
It seems #2447 fixed the timeouts. @fstein93, could you please revert back to the two-stage solver?

fstein93:
Apparently, everything works now. If there are no further issues raised, I will merge this PR.

oschuett:
I'm a bit disappointed that we don't see any improvements in our [performance test](https://dashboard.cp2k.org/archive/perf-cuda-volta/index.html).
Shouldn't this show up in the `diag_cu144_broy.inp` benchmark?

fstein93:
Might be but I don't know what kernel we had used before.

Issue #2411, opened 2022-11-24
Failing tests with the Intel compiler


fstein93:
Since my PR last week, the regtests have been failing with the Intel compiler at [rt_propagation_types.F](https://github.com/cp2k/cp2k/blob/63f81a75a05e81cf79788031d5dd532ff2fbf5d1/src/rt_propagation_types.F#L215). I can reproduce the error locally but I still have no idea, I know that commit [5c33d94](https://github.com/cp2k/cp2k/pull/2400/commits/5c33d94c092bcb25b1d5c4247019979f4c83b3a9) caused the crash but I could not tell so far how the crash is caused. I checked the association state of the pointers (apart from rtp%mos%new etc. as they are allocated right before), all of them are found to be associated. I run the calculations with gdb-ia provided by our the intel parallel studio distribution on our cluster without obtaining any further additional information. The Intel inspector even returns an issue on line 141 (right at the function entrance).

Does anyone of you have an idea?

mkrack:
I am just looking into this issue. The Intel compiler detects `mos_aux` as not associated on [entry](https://github.com/cp2k/cp2k/blob/master/src/rt_propagation_types.F#L140) to `rt_prop_create` which causes a SIGSEGV in this [call](https://github.com/cp2k/cp2k/blob/master/src/rt_propagation_types.F#L222) to `cp_fm_create`, finally. Tracing back leads to [line 465](https://github.com/cp2k/cp2k/blob/master/src/motion/rt_propagation.F#L465) in `rt_propagation.F` in which `qs_env%admm_env` and thus `mos_aux_fit` is not associated.

fstein93:
Thank you for the hint, I have found the culprit. The error is related to the INTENT of mos_aux in rt_prop_create. In rt_propagation.F, it provided as a component of admm_env which is not associated in the given tests. GCC apparently ignores this issue because it never accesses mos_aux whereas Intel does (at least for copying the data). Removal of the intent should work. The intent of the argument mos is correct because it is accessed directly at the begin of the routine.
Apparently, my employed Intel compiler+utilitiess is a bit outdated to catch this.

Thank you for your effort and please excuse the inconvenience.

mkrack:
Removing the INTENT(IN) attribute for `mos_aux` does not help, because its parent `admm_env` is also not associated. The argument has to become OPTIONAL in addition.

mkrack:
[...]

PR #2412, opened 2022-11-24
Add POINTER attribute for admm_env

mkrack:


mkrack:
This PR does not fix #2411.

src/motion/rt_propagation.F
```diff
@@ -463,7 +463,7 @@ SUBROUTINE rt_initial_guess(qs_env, force_env, rtp_control)
          ALLOCATE (qs_env%rtp)
          CALL get_qs_env(qs_env, matrix_s=matrix_s)
          CALL rt_prop_create(qs_env%rtp, qs_env%mos, qs_env%mpools, dft_control, matrix_s(1)%matrix, &
-                             rtp_control%linear_scaling, qs_env%admm_env%mos_aux_fit)
+                             rtp_control%linear_scaling)
```
fstein93: I think you have to add a branch here, one with ADMM, one without ADMM because made mos_aux became an optional parameter

mkrack: @fstein93 Thanks for the hint.

PR #2427, opened 2022-12-02
Add libtorch support to cmake build system

mtaillefumier:


oschuett:
Thanks for adding support for libtorch!
What's the issue with the licencing of `GetGitRevisionDescription.cmake`?

mtaillefumier:
> Thanks for adding support for libtorch! 

no problem. Better to react quickly.

> What's the issue with the licencing of `GetGitRevisionDescription.cmake`?

Yes we did not write this code actually. It is licensed under BSD and freely available but we can not change the licensing terms as the  copyright indicates both the author and the university he was working for. 
It is not a big issue just that we can not dual license the file without explicitly asking for permission from the authors. 
We can solve this either by rewriting the functionality or remove the banner check for this file only.

oschuett:
> We can solve this either by rewriting the functionality or remove the banner check for this file only.

I'd propose we simply remove that file together with most of these lines [CMakeLists.txt:29-67](https://github.com/cp2k/cp2k/blob/c025c99d5d860bcf6d6a049a55c865b3089da860/CMakeLists.txt#L29-L67).

From what I can tell, it's not consistent with the way we currently tag releases, which is documented [here](https://www.cp2k.org/dev:release_checklist).

mtaillefumier:
I will remove the incriminated file, that's simpler. 

I do not want however to have to modify the CMakelists.txt every time we make a release. This should be taken automatically from the tag if it exists and just using git should be enough (or we have this info from a file). The git:hash is taken care off automatically. 

The only important information is the commit hash anyway.

mtaillefumier:
this commit should fix the licensing issue altogether. The versioning is still unsolved to some extent. the git hash is always given and the code is compiled with this information so any output will indicate it the usual way.

PR #2432, opened 2022-12-05
add missing xyz output format for helium and fixes of corner cases

hforbert:


oschuett:
You can use:
```
 CALL cp_warn(__LOCATION__,"your lengthy message, "//&
              "which can even span multiple lines")
```

See also: https://www.cp2k.org/dev:error_handling

hforbert:
Thanks, fixed it the way elsewhere in that file.

PR #2437, opened 2022-12-06
Various fixes for the cmake build system

mtaillefumier:
- SPLA offloading could not be turned on because I used the wrong variable name
- added missing plumed dependency


oschuett:
There seems to be a merge conflict.

mtaillefumier:
the RPA gradients commit is showing up for no apparent reasons. It is not even in my git history.

mtaillefumier:
should be fine now.

oschuett:
> the RPA gradients commit is showing up for no apparent reasons. It is not even in my git history.

I looks to me like you rebased the upstream branch onto your own instead of the other way around. This is why in your branch the latest upstream commits appear after your own commit.

PR #2440, opened 2022-12-08
Improve numerical stability and improve performance

fstein93:
This PR improves the numerical stability of the calculation of the RPA densities. Thereby, the performance of the code is improved by at least a factor of two because one call to local_gemm is not needed anymore. Because the numerical accuracy was improved significantly, I changed the default block size which further accelerated RPA gradient calculations.

### DIFF
--- a/src/input_cp2k_mp2.F
+++ b/src/input_cp2k_mp2.F
@@ -471,7 +471,7 @@ CONTAINS
          "are calculated in batches of the size given by this keyword. Larger block sizes "// &
          "improve the performance but reduce the numerical accuracy. Recommended block sizes are multiples of the number of "// &
          "doubles per cache line (usually 8). Ignored with MP2 gradients. Set it to -1 to prevent blocking.", &
-         default_i_val=32)
+         default_i_val=-1)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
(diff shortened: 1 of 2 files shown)


oschuett:
>  Thereby, the performance of the code is improved by at least a factor of two ...

Oh wow, 2x is awesome!

However, for some reason this does not show up in our [performance tests](https://dashboard.cp2k.org/archive/perf-cray-xc40-gnu-psmp/index.html). Is this because they don't compute RPA forces?

fstein93:
Yes, we don't test RPA gradients. The factor of two stems from the removal of one call to local_gemm (before, there were 2). Thus, it is only related to the given part of the code, but in my tests it can mean 20 % of the total run time. Unfortunately, I couldn't test the code on LUMI yet because of currently running large pilot calculations of other groups.

oschuett:
> Yes, we don't test RPA gradients. 

Do you think this would be worth doing?
I guess, we could just create copies of our [existing benchmarks](https://github.com/cp2k/cp2k/tree/master/benchmarks/QS_mp2_rpa) with `RUN_TYPE ENERGY_FORCE`, right?

Perhaps, we could also create versions with just 16 or 8 water molecules that could be included in our [single node runs](https://dashboard.cp2k.org/archive/perf-openmp/index.html)?

PR #2441, opened 2022-12-08
Fix issues with the NVIDIA compiler

fstein93:


oschuett:
Out of curiosity, which NVIDIA compiler did you use and what was the problem?

fstein93:
I did not use any NVIDIA compiler, we received a mail from its developer. They are currently rechecking the code.

oschuett:
I also recently submitted two PR to address some issues with nvfortran:
 - https://github.com/cp2k/cp2k/pull/2441
 - https://github.com/cp2k/dbcsr/pull/635

I guess, we both received the same email :-)

I also added [this ARCH file](https://github.com/cp2k/cp2k/blob/master/arch/Linux-x86-64-nvhpc.ssmp) and  [this Dockerfile](https://github.com/cp2k/cp2k/blob/master/tools/docker/Dockerfile.test_nvhpc) which allow us to tests the build with NVIDIA's HPC compiler suite:

```
cd cp2k/tools/docker
docker build -f Dockerfile.test_nvhpc ../../
```

PR #2444, opened 2022-12-09
Real Time Print

glb96:
Dear CP2K community,

I am a new postdoc at the Zurich University with Marcella Iannuzzi and Jürg Hutter. My work is related to real-time propagation, and I will do some code development. 

This is my first merge request so I will be very glad to have any help/tips or to know if I am doing something wrong! 

The goal of these modifications are: 

- Create a bit of ''space'' in the code for further Real Time prints for MO-related properties (therefore some part of the code have been delocalized from the _output.F file to the _utils.F) 
- Print the MO_CUBE files in real time. This version is incomplete regarding all the possible input of the MO_CUBE section, but since some people need to have it done as soon as possible, I push this first version now and I will cover the other possibility later on.
- A small change in the MO print: by default the first time step is printed. This is an important feature for us right now, if required I can add a key in the MO section to make this print for the first time step optional. 
- I allow myself to set a function  in private mode since it is used only in the same file. 

Best,
Guillaume


oschuett:
Hi Guillaume, welcome to the team :-)

Your changes look generally good to me. However, I'd ask you to remove that merge commit from your branch because we want to keep our [git history linear](https://github.com/cp2k/cp2k/wiki/Git-Tips-&-Tricks#basics).

PS: No need to open a new PR. Just push your updated branch and the CI tests will automatically restart.

Issue #1030, opened 2020-08-14
Migrate mpiwrap to mpi_f08


oschuett:
To comply with the Fortran 2008 standard we should upgrade [mpiwrap](https://github.com/cp2k/cp2k/tree/master/src/mpiwrap) to the new [mpi_f80](https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node408.htm) API.

fstein93:
Then we can remove the checks for MPI version 2.0, too. The module mpi_f08 is part of the MPI 3.0 standard. But that shouldn't be a problem because it has been published in 2012 and probably every MPI library in use supports the 3.0 standard.

fstein93:
While grasping through the different documentations of MPI and the actual libraries (mostly OpenMPI), I have found that the mpif90 compiler is deprecated since OpenMPI version 1.7 (!) in favor of mpifort. There are also a few workarounds we can revise or remove in favor of clearer code.

fstein93:
While configuring OpenMPI, we have to [keep](https://github.com/cp2k/cp2k/blob/399e6db9ee1a21058cc4319e934d6c8fef1bc356/tools/toolchain/scripts/stage1/install_openmpi.sh#L68) the mpi1-compatibility because the currently employed versions of the libraries [PEXSI](https://pexsi.readthedocs.io/en/latest/download.html) and [PTScotch](https://www.labri.fr/perso/pelegrin/scotch/) depend on these symbols. The latest releases could solve this issue.

fstein93:
If you are interested in my current plans in this direction, check [this](https://github.com/fstein93/cp2k/projects/2). In general, I would like to introduce OOP because most MPI calls are bound to a single communicators/windows/files/requests/... (exceptions: mpi_init, mpi_finalize, mpi_waitall, ...), but could also be adopted to cartesian communicators, blacs contexts etc.

The syntax would then be something like
CALL communicator%send(sendbuffer, destination[, tag])
or
request = communicator%isend(sendbuffer, destination[, tag])

oschuett:
+1 for upgrading PEXSI and PTScotch.

> I would like to introduce OOP 

Yes, we should really adopt more OOP. It will allow us to replace many of those dreadful if-else-spaghetti-regions with polymorphism. See also #356 for inspiration.

oschuett:
[...]

PR #2449, opened 2022-12-12
Refactor MPI handles

fstein93:
This PR encapsulates all MPI handles used outside of the MPI wrapper module, i.e. communicators, requests, files, windows (the last one is not used in CP2K). File handles are only encapsulated as far as they are connected to the MPI module, otherwise, the ordinary Fortran integer handle will be used as in most parts of the code, the output is written to the regular output file. the status handles are not encapsulated as they are never used outside of the wrapper.

This PR prepares the introduction of the MPI-Fortran-2008 module.

I am open for comments.

### DIFF
--- a/src/almo_scf.F
+++ b/src/almo_scf.F
@@ -81,7 +81,8 @@ MODULE almo_scf
    USE kinds,                           ONLY: default_path_length,&
                                               dp
    USE mathlib,                         ONLY: binomial
-   USE message_passing,                 ONLY: mp_sum
+   USE message_passing,                 ONLY: mp_comm_type,&
+                                              mp_sum
    USE molecule_types,                  ONLY: get_molecule_set_info,&
                                               molecule_type
    USE mscfg_types,                     ONLY: get_matrix_from_submatrices,&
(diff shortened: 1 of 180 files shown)


oschuett:
Awesome!
Referencing #1030 for posterity.

fstein93:
Now, the handles are stored in separate module-private variables. Then, we define the CP2K wrappers and use the handles of the null objects as defaults for our handles and finally, we can define the CP2K-wrapped MPI constants.

oschuett:
Looks very good!

I was wondering what is your plan for switching over to the MPI-Fortran-2008 module? Will there be a flag or do we need to make a hard cut over? In the later case we should probably wait until after the release.

fstein93:
I think we should wait for the next release to include the Fortran-2008 module but we can already merge this PR before the next release as it just provides a plain refactoring without any differences in the actual content/performance.

alazzaro:
Checking the code, I found a number of mpi2 and mpi3 mentions still present in the code:

https://github.com/cp2k/cp2k/search?q=mpi2
https://github.com/cp2k/cp2k/search?q=mpi3

PR #2451, opened 2022-12-13
Filter empty keywords

mkrack:


oschuett:
Curious, why/where do we have empty keywords?

mkrack:
The first line of the CP2K keywords section in the current `vim` plugin file [cp2k.vim](https://manual.cp2k.org/trunk/cp2k.vim) is empty which causes an error message each time a CP2K input file is opened with vim using that plugin file. The `cp2k_input.xml` contains lines with
```
<NAME type="default"></NAME>
```
in some of the `libxc` parameters sections, at least in [MGGA_XC_B97M_V](https://manual.cp2k.org/cp2k-2022_1-branch/CP2K_INPUT/FORCE_EVAL/DFT/XC/XC_FUNCTIONAL/MGGA_XC_B97M_V.html). It seems that a parameter name of zero-string length is returned for this LIBXC functional. I will commit a check which will print a warning in `xc_libxc.F` to catch such cases, because the vim plugin file generation should be resilient.

PR #2452, opened 2022-12-13
Check for LIBXC parameter names of zero-string length

mkrack:


oschuett:
Maybe we should just add asserts to  `section_create` and `keyword_create` just like we did for `enum_create` via #2258?

mkrack:
There is a comma missing at the end of line 20 in [mgga_xc_b97mv.c](https://gitlab.com/libxc/libxc/-/blob/master/src/mgga_xc_b97mv.c#L20) which causes the zero-length string for the keyword name. This issue has been reported to the [libxc developers](https://gitlab.com/libxc/libxc/-/issues/437).

Issue #2379, opened 2022-10-27
building cp2k with cmake


mtaillefumier:
Please add a comment to this issue to report any build that do not work properly. To ease debugging, would it be possible to also copy/paste the cmake output.

mtaillefumier:
- need to test the build on mac os.

oschuett:
The CMake build system is coming along nicely. So, below I've tried to list all the configurations that we'll need to migrate:

### Prerequisites
- [ ] #351.
- [ ] Add `generate_cmake_commands.sh` as equivalent to [generate_arch_files.sh](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/generate_arch_files.sh)
- [ ] Adopt [do_regtest.py](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest.py) to support new binary paths.

### Libraries
- [ ] parallel (make MPI optional)
- [ ] elpa
- [ ] libvdwxc
- [ ] sirius
- [ ] plumed2
- [ ] pexsi
- [ ] quip
- [x] libint
- [x] fftw3
- [x] libxc
- [x] scalapack
- [x] cosma
- [x] xsmm
- [x] spglib
- [x] libvori
- [x] libtorch

### Architectures
- [ ] offload_cuda
- [ ] offload_hip
- [ ] dbcsr_acc
- [ ] spla_gemm_offloading
- [ ] elpa_nvidia_gpu
- [ ] ARM64
- [ ] [macOS](https://www.cp2k.org/howto:compile_on_macos)
- [ ] [i386](https://github.com/cp2k/cp2k/blob/master/tools/docker/Dockerfile.test_i386)
- [ ] [FPGAs](https://github.com/cp2k/cp2k/blob/master/INSTALL.md#2q-fpga-optional-plane-wave-fft-calculations)
- [ ] [Windows via Cygwin](https://www.cp2k.org/howto:compile_on_windows_with_cygwin)

### Compilers
- [ ] Intel Classic
- [ ] [Intel OneAPI](https://github.com/cp2k/cp2k/blob/master/tools/docker/Dockerfile.test_intel-psmp)
- [ ] [Nvidia HPC](https://github.com/cp2k/cp2k/blob/master/tools/docker/Dockerfile.test_nvhpc)
- [x] GCC

### Exotic Builds
[...]

PR #2462, opened 2022-12-20
[maint] CMake fixes and improvements

RMeli:
This PR provides some fixes and improvements for the new CMake build system, in particular when the dependencies are installed manually while the `PKG_CONFIG_PATH` environment variable is not set.

This PR only concerns required dependencies (`dbcsr`) and optional dependencies that are turned `ON` by default. (Other optional dependencies will be tested in due course.)

This PR also improves the `README_cmake.md` file, especially the formatting issues introduced with 786bc82ff9ded3e1f761cba6d8e25c3c9fe19bb1.

Related to #2379.


abussy:
Hi @RMeli! Before you merge your changes, may I suggest that you squash all your commits into a single one? This makes the navigation of CP2K history easier in the long run.

RMeli:
Hi @abussy , sure no problem. But isn't it what GitHub squash-merge does anyways?

abussy:
It may very well be. In that case, please ignore my previous comment

oschuett:
Hi Rocco, welcome to CP2K :-)

Your changes look good to me. I'd only ask you to please remove the merge commits to keep the [history linear](https://github.com/cp2k/cp2k/wiki/Git-Tips-&-Tricks#basics).


On a related note, I'm wondering why none of these issues showed up in our [testing](https://github.com/cp2k/cp2k/blob/0389982c0eb64f75500e146fe18c7d760ead9eeb/tools/docker/Dockerfile.test_cmake#L84)?

RMeli:
I'll clean up the history and start using `git rebase`.

> On a related note, I'm wondering why none of these issues showed up in our [testing](https://github.com/cp2k/cp2k/blob/0389982c0eb64f75500e146fe18c7d760ead9eeb/tools/docker/Dockerfile.test_cmake#L84)?

In your testing, you properly [setup `PKG_CONFIG_PATH`](https://github.com/cp2k/cp2k/blob/0389982c0eb64f75500e146fe18c7d760ead9eeb/tools/docker/Dockerfile.test_cmake#L83) (I added a comment about doing this in the `README_cmake.md` file), therefore everything is found by `pkg_check_modules()` and the paths where I found some issues are never triggered.

RMeli:
@oschuett BTW, are `-DCP2K_USE_LIBXSMM=NO` and `-DCP2K_USE_LIBINT2=NO` actually supposed to be `NO`/`OFF`?

https://github.com/cp2k/cp2k/blob/0389982c0eb64f75500e146fe18c7d760ead9eeb/tools/docker/Dockerfile.test_cmake#L84

oschuett:
> @oschuett BTW, are -DCP2K_USE_LIBXSMM=NO and -DCP2K_USE_LIBINT2=NO actually supposed to be NO/OFF?

Yes, we had to disable them because they didn't work at the time. Please enable then once they start working.

Generally, the plan was to setup a minimal cmake test and then expand its coverage over time.


> In your testing, you properly setup PKG_CONFIG_PATH, therefore everything is found by pkg_check_modules() and the paths where I found some issues are never triggered.

Yes, it's an inherent problem of build configuration that there are so many different code paths. At least with cmake the config step is now well isolated. So perhaps in the future we could try to sample this space e.g. with fuzzy testing.

mtaillefumier:
> Yes, we had to disable them because they didn't work at the time. Please enable then once they start working.
> 
> Generally, the plan was to setup a minimal cmake test and then expand its coverage over time.

it is a problem coming from the docker environment. The libint package for instance compiles without issues but building cp2k with cmake and libint within the container fails all the time. If i compile libint + cp2k + cmake on any cray system or fedora then everything is alright (even with cp2k-cmake spack version). 

> Yes, it's an inherent problem of build configuration that there are so many different code paths. At least with cmake the > config step is now well isolated. So perhaps in the future we could try to sample this space e.g. with fuzzy testing.

Every time a failure occurs almost always imply a non standard installation of the dependency. What makes the command line in the docker file so complex is the fact that every dependency is installed in a separate directory but environment variables such as `PKG_CONFIG_PATH` or some common cmake variables are not always set up properly. 

One option at least for the docker env is to install all dependencies in one directory or use spack when we create the container. I need to finish updating the cp2k package and open a PR upstream (or add the package file directly in the cp2k repo). 

 '-DCP2K_USE_LIBXSMM=YES' would work out of the box if libxsmm pc files are in the right directory. It is not the case by default and all package managers have to manually move the pc to `prefix/lib/pkgconfig` 

NB : it is always possible to indicate the libraries manually. `cmake` supports this option as well and it is practical for instance on power9 systems (mixing essl and openblas). 

- fuzzy testing :+1: we will always find cases where it fails.
- libvori is also problematic for some curious reason (linking issues).
- openblas (blas in general) will always be source of trouble because distributions do not follow any naming convention

Basically if a package has a pc file then cmake will use it. There was a bug in my cmake scripts that was triggered only in specific cases (which @RMeli found out after testing in a completely different environment than mine). 

@RMeli : libxsmm and libint are on by default but off in the docker file because of compilation failure.

oschuett:
> it is a problem coming from the docker environment.

There is nothing special about the Docker environment. It's a plain Ubuntu installation. So, any error found by the CI will also affect a majority of our users.

> What makes the command line in the docker file so complex is the fact that every dependency is installed in a separate directory but environment variables such as PKG_CONFIG_PATH or some common cmake variables are not always set up properly.

I agree. We should make sure all install dirs are added to `PKG_CONFIG_PATH`. Some script like [install_cosma.sh](https://github.com/cp2k/cp2k/blob/523d85c309ba47d334e69c8546862257e289ce38/tools/toolchain/scripts/stage4/install_cosma.sh#L185) already do it.


> ... or use spack when we create the container. I need to finish updating the cp2k package and open a PR upstream (or add the package file directly in the cp2k repo).

Improving our Spack support is certainly desirable. Have a look at [Tiziano's work](https://github.com/cp2k/cp2k/tree/master/tools/toolchain-spack). Once you got it to work we could setup a new dashboard test for Spack+CMake. 

Nevertheless, we shouldn't lose sight of the fact that the normal toolchain is still what's used by the majority of our users and what powers all of our testing. The good news is that most of its complexity stems from its entanglement with the Make build system. Hence, I expect it will collapse into something rather benign once we remove the Makefile.

mtaillefumier:
> There is nothing special about the Docker environment. It's a plain Ubuntu installation. So, any error found by the CI will also affect a majority of our users.

Ubuntu is not guilty. Just linking Libint and libvori cause issues. I do not have these problems when I build the dependencies by hand.
> 
> > What makes the command line in the docker file so complex is the fact that every dependency is installed in a separate directory but environment variables such as PKG_CONFIG_PATH or some common cmake variables are not always set up properly.
> 
> I agree. We should make sure all install dirs are added to `PKG_CONFIG_PATH`. Some script like [install_cosma.sh](https://github.com/cp2k/cp2k/blob/523d85c309ba47d334e69c8546862257e289ce38/tools/toolchain/scripts/stage4/install_cosma.sh#L185) already do it.
> 
> > ... or use spack when we create the container. I need to finish updating the cp2k package and open a PR upstream (or add the package file directly in the cp2k repo).
> 
> Improving our Spack support is certainly desirable. Have a look at [Tiziano's work](https://github.com/cp2k/cp2k/tree/master/tools/toolchain-spack). Once you got it to work we could setup a new dashboard test for Spack+CMake.

I am using Tiziano work as the package is already in spack repository. I even wonder if I should keep anything below 2023.1 and just use cmake with spack.
> 
> Nevertheless, we shouldn't lose sight of the fact that the normal toolchain is still what's used by the majority of our users and what powers all of our testing. The good news is that most of its complexity stems from its entanglement with the Make build system. Hence, I expect it will collapse into something rather benign once we remove the Makefile.

let's get the cmake build system and spack package more robust as spack is used by many major hpc centers for automatic deployment. PKG_CONFIG_PATH, etc... can be added to the toolchain without too much difficulties and without breaking the current make build system.

oschuett:
> I even wonder if I should keep anything below 2023.1 and just use cmake with spack.

Indeed, dropping support for pre-2023.1 versions should significantly simplify the Spack package.

> Just linking Libint and libvori cause issues. I do not have these problems when I build the dependencies by hand.

Yeah, libint can be tricky. The toolchain is currently using tar-balls that were pre-built [here](https://github.com/cp2k/libint-cp2k). Unfortunately, I don't know much about it. We could try to ask Alfio or Tiziano.

For the libvori we can reach out to Martin Brehm (user: Brehministrator), who is always very responsive and helpful.

mtaillefumier:
> Yeah, libint can be tricky. The toolchain is currently using tar-balls that were pre-built [here](https://github.com/cp2k/libint-cp2k). Unfortunately, I don't know much about it. We could try to ask Alfio or Tiziano.
> 

I get this kind of errors when i build cp2k with docker

```
/usr/include/c++/11/iostream:74: undefined reference to `std::ios_base::Init::~Init()'
/usr/bin/ld: /opt/cp2k-toolchain/install/libint-v2.6.0-cp2k-lmax-5/lib/libint2.a(CR_DerivGaussP0InBra_aB_g010__0__m__1___TwoPRep_unit__0__s__1___Ab__up_0.o): in function `_GLOBAL__sub_I_CR_DerivGaussP0InBra_aB_g010__0__m__1___TwoPRep_unit__0__s__1___Ab__up_0.cc':

```
> For the libvori we can reach out to Martin Brehm (user: Brehministrator), who is always very responsive and helpful.

the libvori build system needs some small fixes (libvori uses cmake so it is simple). I will contact him with a patch

PR #2463, opened 2022-12-20
Fix OpenMPI toolchain

abussy:
The toolchain for the installation of OpenMPI suffered from a typo introduced in #2438. This also illustrates that the default toolchain is not tested on the Dashboard.


fstein93:
As mentioned in my issue, your PR solves the issue on the dashboard since [1ed20a1](https://github.com/cp2k/cp2k/commit/1ed20a10a625a7316c33ffd111a2e2b815707aa1) .

PR #2482, opened 2022-12-27
Update DBCSR v2.5.0...

alazzaro:
and remove some old dashboard tests.


alazzaro:
@oschuett I don't see anything wrong with the new DBCSR (and definitely nothing that can affect wrong results). Please merge if you think it is ok.

PR #2495, opened 2023-01-11
Bump LibXC to version 6.1.0

fstein93:
We drop the patch because it was fixed in the given release.

Can anyone add the [new tarball](http://www.tddft.org/programs/libxc/down.php?file=6.1.0/libxc-6.1.0.tar.gz), please?


oschuett:
Voilà: https://www.cp2k.org/static/downloads/libxc-6.1.0.tar.gz
However, I'm getting a different sha256 checksum than you. Please double check.

fstein93:
We had that issue several times. I always used the hash of the tar of the LibXC github repository. Who knows...

oschuett:
This PR [broke](https://dashboard.cp2k.org/archive/cmake/commit_0e722df91b89847b24d3e2fa39485a0bdfbfba12.txt) the CMake build. None of the [commits](https://github.com/ElectronicStructureLibrary/libxc/compare/6.0.0...6.1.0) look suspicious. @mtaillefumier  any ideas?

mtaillefumier:
I will have a look later today as I can not access my password database 
at the moment. Probably something wrong  with the version number in the 
cmake build system. Thanks for reporting this

mtaillefumier:
I found the problem it is here 
https://github.com/cp2k/cp2k/blob/0e722df91b89847b24d3e2fa39485a0bdfbfba12/tools/docker/Dockerfile.test_cmake#L84

-DLibXC_ROOT=/opt/cp2k-toolchain/install/libxc-6.0.0

we were planning to open a PR regrouping all the work related to cmake and the toolchain but it might be better to have PR fixing all paths in the toolchain first.

PR #2499, opened 2023-01-13
Fix degenerate orbitals in MP2

fstein93:
There were some uninitialized variables and a wrong buffer which probably cause the segfaults on Mac.

### DIFF
--- a/src/mp2_ri_gpw.F
+++ b/src/mp2_ri_gpw.F
@@ -2270,6 +2270,7 @@ CONTAINS
 
                   do_send_i = .FALSE.
                   do_send_j = .FALSE.
+                  do_send_k = .FALSE.
                   IF (ijk_index <= send_ijk_index) THEN
                      ! something to send
                      ijk_counter_send = (ijk_index - MIN(1, integ_group_pos2color_sub(proc_send)))* &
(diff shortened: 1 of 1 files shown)


oschuett:
Oh wow. It's frightening how these things can slip through the cracks, i.e. go unnoticed by Valgrind and Address Sanitizer.

fstein93:
In this case, it is difficult to catch with just the regtests because you should not have enough memory to raise this error (like on our Mac tester).
On our machine, I had to adjust the code to prevent the exchange of data (in the output file: "Group size for integral replication" which is 2 on Mac but 1 in all other testers). In real-world calculations, it is only triggered if you have an open-shell system (I rarely do that) and the existence of nearly-degenerate occupied orbital pairs (requires special systems). Apparently, I was very lucky while testing the code such that I did not capture that one although I tried larger systems.

oschuett:
The macOS test turned green for the first time! Great catch!

> In this case, it is difficult to catch with just the regtests because you should not have enough memory to raise this error 

Should we then add more regtests to ensure these branches are always covered or would a more general low-memory test make more sense?

fstein93:
Well, in general, it is difficult to tell which part of the code is called with what amount of memory. In the given case, I can add the option to set the number of integration groups externally and then just add more tests for coverage.

PR #2504, opened 2023-01-16
Add option to set number of MP2 integration groups externally

fstein93:
Allows testing of low-memory regime. (see #2499)


src/input_cp2k_mp2.F
```diff
@@ -274,6 +274,16 @@ SUBROUTINE create_ri_mp2(section)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
+      CALL keyword_create(keyword, __LOCATION__, name="NUMBER_INTEGRATION_GROUPS", &
+                          description="Sets the number of integration groups of the communication scheme in RI-MP2."// &
```
oschuett, line 278: Maybe mention that for a value<=0 the group size will be automatically computed from the available memory.

fstein93: Ah right, I have accidentally removed that hint.

PR #2508, opened 2023-01-17
Reduce optimisation level for paw_basis_types.F

mkrack:
It seems that segfaults occur using the Intel compiler if paw_basis_types.F is compiled with an optimisation level higher than -O1.

### DIFF
--- a/arch/Linux-intel-x86_64.psmp
+++ b/arch/Linux-intel-x86_64.psmp
@@ -13,7 +13,7 @@
 #        Optionally, the Intel compiler version can be specified as argument.
 #        Replace or adapt the "module add" commands below if needed.
 #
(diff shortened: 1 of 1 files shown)


oschuett:
I'm wondering why the [Intel oneAPI test](https://dashboard.cp2k.org/archive/intel-psmp/index.html) did not find this? It also uses version `2021.6.0 20220226`.

mkrack:
> I'm wondering why the [Intel oneAPI test](https://dashboard.cp2k.org/archive/intel-psmp/index.html) did not find this? It also uses version `2021.6.0 20220226`.

@oschuett No clue. I will run a regression test from scratch using the Intel Classic compiler 22.2 which returns the same version string.

PR #2539, opened 2023-01-30
Apply object orientation to the MPI wrapper

fstein93:
This is a first approach to apply object orientation to the MPI wrapper. I applied the following changes:
- Type-bound procedures: The routines of the former interfaces were bound to the respective MPI types. I simplified the generic names a bit. In general, I switched from `CALL mp_<routine_name>(arguments, object)` to `CALL object%<routine_name>(arguments)`. All routines were made non-overridable to reduce the OOP overhead.
- The proper release of communicators is now checked in serial mode
- Requests: The functions mp_test and mp_wait are bound directly to the request handles whereas mp_[wait|test][any|all] are not because I cannot bind functions to arrays of requests.
- Finalization: Not implemented because of compiler bugs. A trial implementation fails with the Intel compiler for no obvious reasons in TMC and DBT/TAS.

As a follow-up, I will take care of the cp_para_env_type, cp_para_cart_type and cp_blacs_env_type. This will not happen here.


oschuett:
Maybe we should add a macro like `OMP_DEFAULT_NONE_WITH_OOP`  to [base_uses.f90](https://github.com/cp2k/cp2k/blob/master/src/base/base_uses.f90) to standardize [these workarounds](https://github.com/cp2k/cp2k/pull/2539/commits/d2f71b24f72311a481dea2ed49e41217da5f24b0).
This would also make it easier to clean up when we drop GCC 7.

fstein93:
I will apply the same change to the USE_FINALIZATION flag.

Due to Intel, we will apparently need a similar solution for the ASYNCHRONOUS attribute. I will check whether I can confirm the issue with our locally installed Intel compiler (20.4). A simple solution would be a CP_ASYNCHRONOUS macro to switch between ASYNCHRONOUS and VOLATILE depending on the compiler. (might reduce the performance, was on my TODO list anyways)

alazzaro:
Interestingly, there is a recent discussion on the finalization on the fortran forum: https://groups.google.com/g/comp.lang.fortran/c/PEyO40KT4B4

fstein93:
So, it could be a compiler bug then. In this case, I will drop finalization then.

fstein93:
The failing tests with GCC 7..10 are already known from the dashboard.

PR #2542, opened 2023-02-01
ignore user manual and vscode settings

rangsimanketkaew:
Ignore manual and vscode settings generated locally by users.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,9 +10,6 @@
 # ignore local arch files not meant to be shared
 /arch/local*
 
-# ignore manual generated locally
-/tools/manual/*
-
 # ensure people are not accidentally committing an old .svnignore when copying their stuff
 .svnignore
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -10,6 +10,9 @@
 # ignore local arch files not meant to be shared
 /arch/local*
 
+# ignore manual generated locally
+/tools/manual/*
```
oschuett: Unfortunately, this is too general because it exclucdes the entire [tools/manual](https://github.com/cp2k/cp2k/tree/master/tools/manual) directory. AFAIK, we don't have a default output dir. So, there is probably no sensible rule we could add here.

rangsimanketkaew: Thanks for the comment.

rangsimanketkaew: I have removed the local manual. Please check.

PR #2545, opened 2023-02-02
Add EVERY_N_STEP for LINRES and regtest

rangsimanketkaew:
Add a keyword "EVERY_N_STEP" for computing linear response every N-th step in dynamic simulation.


src/qs_linres_module.F
```diff
@@ -240,7 +240,10 @@ SUBROUTINE linres_calculation_low(qs_env)
       lr_section => section_vals_get_subs_vals(qs_env%input, "PROPERTIES%LINRES")
       CALL section_vals_get(lr_section, explicit=lr_calculation)
 
-      IF (lr_calculation) THEN
+      ALLOCATE (linres_control)
```
oschuett: The `linres_control` is not getting deallocated, which leads to a [memory leak](https://storage.googleapis.com/cp2k-ci/run-cp2k-sdbg-cda7b2e3_report.txt).

rangsimanketkaew: Thanks. Fixed.

src/qs_linres_types.F
```diff
@@ -58,6 +58,7 @@ MODULE qs_linres_types
       INTEGER                                   :: restart_every
       REAL(KIND=dp)                             :: energy_gap
       INTEGER                                   :: max_iter
+      INTEGER                                   :: every_n_step
```
oschuett: It seems you only need the value right after you read it from the input. Hence, there is no need to store it in `qs_linres_types`. Please remove it.

src/qs_linres_module.F
```diff
@@ -240,7 +240,12 @@ SUBROUTINE linres_calculation_low(qs_env)
       lr_section => section_vals_get_subs_vals(qs_env%input, "PROPERTIES%LINRES")
       CALL section_vals_get(lr_section, explicit=lr_calculation)
 
-      IF (lr_calculation) THEN
+      ALLOCATE (linres_control) 
+      CALL section_vals_val_get(lr_section, "EVERY_N_STEP", i_val=linres_control%every_n_step)
```
oschuett: You can simply read the value into a local variable because only need it in the next line.
Please remove the ALLOCATE/DEALLOCATE of `linres_control`.

oschuett (changes requested):
Sorry, I didn't take a closer look earlier. I think there is a much simpler way.

src/qs_linres_types.F
rangsimanketkaew: Added it in the derived type in case it could be useful for other tasks in the future.

src/qs_linres_module.F
rangsimanketkaew: Thanks. Done.

PR #2546, opened 2023-02-02
Fix typo in arch/Linux-intel-x86_64.psmp

bjoe2k4:
Shared part is not really tested afaict, but for the sake of completeness let's fix this typo.


mkrack:
@bjoe2k4 Thanks for fixing that typo. You are right, shared is not tested, because the toolchain does not build shared libs for all dependencies.

PR #2555, opened 2023-02-06
Replace arrays of cp_cfm_p_type with arrays of cp_cfm_type

fstein93:


fstein93:
I will ignore the slow tests of the pdbg tester and will not add them to the exceptions. I suppose that these are just accidental hits.

src/negf_methods.F
```diff
@@ -1734,19 +1697,35 @@ SUBROUTINE negf_retarded_green_function_batch(omega, v_shift, ignore_bias, negf_
...
-                        CALL cp_cfm_start_copy_general(gret_gamma_gadv_group(icontact, ipoint)%matrix, &
-                                                       gret_gamma_gadv(icontact, ipoint)%matrix, &
-                                                       para_env, info2(icontact, ipoint))
+                     IF (ASSOCIATED(gret_gamma_gadv(icontact, ipoint)%matrix_struct)) THEN
+                        IF (ASSOCIATED(gret_gamma_gadv_group(icontact, ipoint)%matrix_struct)) THEN
+                           CALL cp_cfm_start_copy_general(gret_gamma_gadv_group(icontact, ipoint), &
+                                                          gret_gamma_gadv(icontact, ipoint), &
+                                                          para_env, info2(icontact, ipoint))
+                        ELSE
+                           CALL cp_cfm_start_copy_general(cfm_dummy, &
```
oschuett, line 1706: The conventions check is [confused](https://dashboard.cp2k.org/archive/conventions/commit_9bd9310ac0e38a12f392c5c3c711eacc29e6a8cd.txt) by these calls that involve `cfm_dummy`.

While the check might just be too sensitive, I'm wondering why these additional calls to `cp_cfm_start_copy_general` had to be introduced?

fstein93: You are right. Originally, I had issues to replace the cp_cfm_p_type here because of its pointer character. In principal, we can remove all branches but the first. Nevertheless, I wonder about the NULL() pointers which are obviously absent in the entire file.

oschuett: > Nevertheless, I wonder about the NULL() pointers which are obviously absent in the entire file.

Yeah, I checked the AST. The `NULL()` are originating from the default intializer of `cp_cfm_type`.
So, you are not actually passing a null pointer, which is what that check was originally designed to prevent.

fstein93: I also double-checked the other occurrences of NULL() pointers. They also have their origin in the default initializer of `cp_fm_type`, but I cannot circumvent them because these matrices are used before and thus, can considered to be valid matrices. So, I will add the respective exceptions.

fstein93: I suspect that as soon as we will have fixed the missing default initializers raised by PR #2562 , we might see these false-positives more often. OOP could help, then.

oschuett: Yes, I'll have to improve the convention checker to distinguishes the two cases. For today the suppression will have to do.

PR #2565, opened 2023-02-08
Fix conventions in file src/negf_methods.F

fstein93:
This is a follow-up to #2555 . I introduced the removed branches to simplify the refactoring but I forgot to remove them afterwards. The left-over issues are added to the suppression file.


oschuett:
Note I've just merged https://github.com/cp2k/cp2k/commit/d99951c4234346617e7eaee907cea28e08bfea46, which changed the format of most entries in `conventions.supp`.

Sorry, for getting in your way.

fstein93:
Don't worry.

PR #2569, opened 2023-02-09
 SIGN option in DISTANCE CV

marci73:
The distance colvar can be used with sign if defined for only one component.

 This is useful to work with vector products

### DIFF
--- a/src/input_cp2k_colvar.F
+++ b/src/input_cp2k_colvar.F
@@ -810,6 +810,16 @@ CONTAINS
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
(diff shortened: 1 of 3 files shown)


src/input_cp2k_colvar.F
```diff
@@ -810,6 +810,16 @@ SUBROUTINE create_colvar_dist_section(section)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
+      CALL keyword_create(keyword, __LOCATION__, name="SIGN", &
+                          description="Whether the distance along one Cartesian axiz has to be considered with sign. "// &
+                          " Yhis option is valid only if only one dimension is selected.", &
```
mkrack, line 815: Typos: "axiz", "Yhis"; two times "only"

PR #2570, opened 2023-02-09
More cmake updates

mtaillefumier:
this PR mostly contains fixes to problem we encountered with building cp2k with spack and cmake.

most noticeable points are
- MPI can be turned off if needed.
- use c++ like namespace convention for each cmake module (CP2K::BLAS::blas etc....)
- cosma detection was not always working properly. Unfortunately this PR solves the problem partially as the issue is  cosma related. We have solved the problem on cosma side but need to go through PR -> release (it will take some time)
- spack + cp2k + cmake works with this PR (and before this PR) as well (with cosma fix). 
- toolchain is untouched


fstein93:
Just out of curiosity: Do you intend to add the support for the mpi_f08 module which is unfortunately not always available?

mtaillefumier:
you mean in cmake. Sure. Is there a dedicated flags for it ?

fstein93:
Yes, it is '-D__MPI_F08'. It is regularly tested with Intel OneAPI and OpenMPI. If you intend to turn it on using some test, then it is NOT sufficient to just check whether the code compiles but you also have to check if you can actually start any kind of communication. We have found this behavior by testing MPICH with GCC 11. (see #2486)

mtaillefumier:
cmake already supports the detection of the mpi interface to fortran 2008. I added an option that you can use to turn the interface on and there is a check if the actual implementation supports it or not. Note that I did not implement the actual tests you suggested because it would really be far more complicated to do so than just letting the user decide if the option is needed or not.

The option is off by default and should be enforced if mpi supports Fortran 2008

mtaillefumier:
we can squash all these commits in one.

PR #2580, opened 2023-02-16
Fix coding conventions introduced in qs_vcd module

edditler:


fstein93:
There are still some calls to cp_fm_gemm. Please replace them with calls to parallel_gemm (of the parallel_gemm_api module). It includes a switch between the Scalapack and the COSMA libraries.

PR #2583, opened 2023-02-16
Refactor para_env and blacs context

fstein93:
This PR refactors the cp_para_env_type, cp_para_cart_type and the cp_blacs_env_type.

1. Introduces the cp_blacs_type as the low-level wrapper type of the blacs context
2. Extends the mp_comm_type to cp_para_env_type, mp_cart_type to cp_para_cart_type, cp_blacs_type to cp_blacs_env_type. The first ones form the layer to MPI and BLACS, adding the most relevant metadata and bind the respective library function wrappers. The high-level types are meant to be used with POINTERs and provide reference counting and an ownership attribute.
3. Introduction of some convenience functions which allow to make components not meant to be used by developers private. The metadata is directly accessible to keep the code simpler and allow the use of the associate construct. At the end, these objects are rather low-level.
4. Making the root arguments of gather, scatter and bcast operations optional, using the respective source component by default. Applying some suitable simplifications of the main code. (To be continued)


src/fm/cp_context_types.F
```diff
@@ -15,43 +15,65 @@
 !>      12.2003 created [Joost]
 !> \author Joost VandeVondele
 ! **************************************************************************************************
-MODULE cp_blacs_calls
+MODULE cp_context_types
```
oschuett: I'd prefer if we could keep `blacs`  in the name. I think, just `context` is too generic.
How about `cp_blacs_context_type` or simply `cp_blacs_type` ?

src/common/cp_para_types.F
```diff
@@ -42,15 +39,18 @@ MODULE cp_para_types
 !>      08.2002 created [fawzi]
 !> \author Fawzi Mohamed
 ! **************************************************************************************************
-   TYPE cp_para_env_type
-      LOGICAL :: owns_group, ionode
-      INTEGER :: mepos, source, num_pe, ref_count
-      TYPE(mp_comm_type) :: group = mp_comm_null
+   TYPE, EXTENDS(mp_comm_type) :: cp_para_env_type
```
oschuett, line 38: Is there a reason we have to keep the `cp_para_env_type`?
Couldn't we just add those few fields to the `mp_comm_type` and then use it everywhere?

src/fm/cp_context_types.F
fstein93: I will change the name.

src/common/cp_para_types.F
fstein93: The cp_para_env_type implements reference counting and the owns_group variable to consider shared ownership, the mp_comm_type doesn't. The cp_para_env_type is meant to be used for pointers whereas the mp_comm_type is employed for local operations. Then, reference counting is not necessary.
In this spirit, I will keep the cp_para_env module which implements routines to deal with POINTERs including some sanity checks. Because they rely on CPABORT, I cannot merge them with cp_para_types due to potential circular dependencies.
Nevertheless, I will shift the meta-data to the mp_comm_type. This allows some adjustments to calls to the MPI_Bcast because their source process is often equivalent to the source component of the para_env.

fstein93:
I was trying to investigate a deviation with the Intel compiler introduced by [8cdf8ea](https://github.com/cp2k/cp2k/pull/2583/commits/8cdf8ea5f4bcfeab79a9fb79e463062d3abb7e71). I decided to increase the threshold slightly because after double-checking, I did not find any issues within the code.

The left-over wrong or failing tests are already known from the Dashboard.

PR #2620, opened 2023-02-22
cmake updates again

mtaillefumier:
- partially revert previous cmake PR.
- The list of files to be compiled is generated on the fly by cmake during the configuration stage
- create cp2kConfig.cmake and cp2kTargets.cmake to ease integration of cp2k in other cmake projects. `gromacs` will benefit from this
- grid, dbm, offload modules have their own CMakeLists.txt. Only C files are compiled and included in each `libcp2k_myname.a` library. Including Fortran files in the archives imply circular dependencies for the fortran api is compiled at the same time than `libcp2k.a`. **Do not create** a single archive from the created one. cp2k.pc and cp2kConfig.cmake will add the necessary information of us
- temporarily ignore `SIRIUS` version as `SIRIUS` was not exporting its version number to cmake. It will be changed again after updating it to the latest version. 
- Installation of cp2k (make install, ninja install) also include the modules files.


oschuett:
While I will merge this PR if you insist, I was actually hoping to generate even more of the cmake code.

For example, the three new `CMakeLists.txt` files that you are adding are very similar. The code related to flags is also quite repetitive. So I think using code generation for those parts would really make our lives easier.

mtaillefumier:
Do not generate the CMakeLists.txt automatically. 
I split them because it was suggested it in the first place otherwise I would have kept one CMakeLists.txt in src. I can remove them

Adding one line in a text file is not difficult that's why I find this search of source files a bit odd. On the top of it it imposes things like that

```
#ifdef HAVE_BLA
#endif 
```
at the begin and the end of source files that are option dependent. Better not include them at all in the build that's also partially why the list of sources files to be compiled is hard coded most of the time.

cmake generates this list automatically so leave this as it is please.

PR #2621, opened 2023-02-22
Fix CMake

mtaillefumier:
- Fix docker recipe for cmake
- add nvhpc support (module provided by nvidia)
- fix a spelling mistake in FindPlumed.cmake
- added an interface target for libraries that are common to all modules.
- use lists of target whenever cmake functions supports them otherwise use for loops for properties common to all targets
- use CMAKE_HIP_ARCHITECTURES and CMAKE_CUDA_ARCHITECTURES instead of custom variables


src/CMakeLists.txt
```diff
@@ -10,11 +10,104 @@ include(GNUInstallDirs) # required to get a proper LIBDIR variable
...
+set(CP2K_DBM_DFLAGS
+    $<$<BOOL:${CP2K_DISABLE_DBM_GPU}>:__NO_OFFLOAD_DBM>
+    $<$<NOT:$<BOOL:${CP2K_DISABLE_DBM_GPU}>>:${CP2K_GPU_DFLAGS}>
+    $<$<BOOL:${CP2K_USE_LIBXSMM}>:__LIBXSMM>)
+
+set(CP2K_GRID_DFLAGS
+    $<$<BOOL:${CP2K_DISABLE_GRID_GPU}>:__NO_OFFLOAD_GRID>
+    $<$<NOT:$<BOOL:${CP2K_DISABLE_GRID_GPU}>>:${CP2K_GPU_DFLAGS}>
+    $<$<BOOL:${CP2K_USE_LIBXSMM}>:__LIBXSMM>
+    $<$<STREQUAL:${CP2K_BLAS_VENDOR},"MKL">:__MKL>)
```
oschuett: Do we actually need this special treatment? Can't we just include `__NO_OFFLOAD_DBM` and `__NO_OFFLOAD_GRID` in the target_compile_definitions for all targets?

src/grid/CMakeLists.txt
```diff
@@ -4,8 +4,9 @@
 #!                                                                                                 !
 #!   SPDX-License-Identifier: BSD-3-Clause                                                         !
 #!-------------------------------------------------------------------------------------------------!
+cmake_minimum_required(VERSION 3.22)
 
-set(GRID_SRCS_C
```
oschuett, line 8: I know I said that eventually there should be one CMakeLists.txt per directory. However, this was meant in the context of generating all those files.

Since we're now back to handwriting, we should try to keep the cmake code to a minimum by making it as generic as possible. So, I think there should be no special treatment of any directory or binary target. Everything should be handled by the generic `src/CMakeLists.txt`.

src/CMakeLists.txt
```diff
@@ -113,20 +208,12 @@ target_link_libraries(
...
-    $<$<BOOL:${CP2K_USE_HIP}>:roc::hipblas>
-    $<$<BOOL:${CP2K_USE_HIP}>:hip::host>
     $<$<BOOL:${CP2K_USE_MPI}>:CP2K::SCALAPACK::scalapack>
-    $<$<BOOL:${CP2K_USE_LIBXSMM}>:CP2K::LibXSMM::libxsmmf>
-    $<$<BOOL:${CP2K_USE_LIBXSMM}>:CP2K::LibXSMM::libxsmmext>
     CP2K::LAPACK::lapack
-    CP2K::BLAS::blas
+    cp2k_grid
+    cp2k_dbm
+    $<$<BOOL:${CP2K_USE_CUDA}>:cp2k_cuda_libs>
```
oschuett, line 1494: Do we actually need [generator expressions](https://cmake.org/cmake/help/v3.0/manual/cmake-generator-expressions.7.html)? It seems they could all be replaced by much more readable if-constructs like this:


```
if(CP2K_USE_CUDA)
  target_link_libraries(cp2k_link_libs INTERFACE cp2k_cuda_libs)
endif()
```

mtaillefumier: Generator expressions will stay.

oschuett: Can you maybe elaborate a bit on why? I assume we both agree that they are taking quite a toll on the readability?

src/grid/CMakeLists.txt
mtaillefumier: I repeat do not generate these files all together. CMakeLists.txt modifications should be the responsibility of the developer who want to add a functionality to it. And if PR review is enforced (which is not the case), then mistakes can be corrected during the review process. 

There is no point for generating a CMakeLists.txt with a python script as CMakeLists.txt are not that difficult to read and modify. The difficulty is in finding dependencies and compilers which cmake does most of the work for us already

- We are handwriting one script of the other. I really does not see the rational behind using a script for generating CMakeLists.txt.

- If editing a CMakeLists.txt with a text editor is tedious then editing a python script (the argument cuts both ways) that generates these files is equally tedious if not more because we have to work with string and we loose the benefit of the cmake mode provided by the editor. The amount of work is the same if not more with the python solution.

- Actually This PR already reduces the amount of code.

oschuett: Yes, complicated changes, like adding a new dependency, should be done directly in the CMake files. However, complicated changes are rare.

The much more common case is adding or renaming a file. Our old build system handled those changes completely automatically. So, I'd like the new build system to be as easy to use as the old one.

Ideally we can achieve this automatization within CMake, but if that's not possible then the next best thing is a Python script.

src/CMakeLists.txt
mtaillefumier: This case is clear. Combination of them might be a bit more complicated to understand at first but it is compact and actually a very useful feature of the language.
Basically, I can not use language features I see fit for the job. I get that but these generator will stay as having a sequence of if() endif() is equally tedious to write and read. The generators expression were probably partially introduced to include such things in a unique function call.

I have multiple of these expression in the cmakelists.txt and I really do not see the point of translating them into if statements unless the expression is really complicated `$<$<AND:$<>,$<>>:>` type for instance.

src/grid/CMakeLists.txt
mtaillefumier: It is possible to generate the list with cmake (`file(GLOB ...)` syntax) I did that before reintroducing the explicit list (three lines of code for the most naive version for generating the list) but then I had to make thing complicated again because of the difference between executable/library (and it is more than three lines).

Renaming/deleting/adding a file happens once in a while and it is not complicated to search/replace a string so I really do not understand why the list should be generated automatically. 

- Compactness and readability well yes at first then things get a bit more complicated afterwards

This poses more issues than this functionality solves

- adding files automatically is dangerous when working with a dirty source tree (yes it happens often). Doing this imposes to developers to be cautious which is good but then what's the difference between this and editing a list. Cautious equally apply.
- the current tree structure imposes removing the executable source code from the list because you have to call _add_library_ for libraries and `add_executable` for executable. They are not generated the same way and do not necessarily have the same properties. There are both binary, all filles are binary it does not mean they represent the same thing and can be treated the same way.  cmake, the linker etc do not exactly behave the same way when they generate them. 
- it imposes thing like `#ifdef __FUNCTIONALITY put all the code we want #endif` because all files are compiled even when the content of the file is 100 % depending on a optional library that is turned off at compilation time (i think even fortran generates a warning or an error when the file is empty). What is in `src/grid/{gpu,hip)` with `__NO_OFFLOAD_GRID` is an example.  What about using macros in C or the imbrication of multiple interfaces before actually calling the dependency function (although I understand why it is needed sometimes). I have seen better question readability.
- we have no clue what files are actually compiled either. We might gain in compactness but we loose the information what source code is compiled and when the code is on or off. It is actually much easier to find the information "oh gpu support is on + option 1,2,3; the functionality depends on file x,y,z" in a text file than searching through the tree to find which file  is actually containing the code. Yes grep is useful but you still need to know what to search for etc.... 
- Also note that the list generation is not done in the Makefile contained in `src/grid` and `src/dbm` directories so if consistency is requested then apply the rules to everything. 
- The current build system is doing it but we do not have to always follow what was done in the past. (look at the toolchain for instance)

oschuett: > It is possible to generate the list with cmake (file(GLOB ...) syntax) ... but then I had to make thing complicated again because of the difference between executable/library (and it is more than three lines).

Yes, it's indeed complicated. The rest of your post is essentially a description of an algorithm that produces a CMakefile for a given set of source files. Currently, you are running this algorithm in your head and only checking in the result. I think, instead you should spell out this algorithm explicitly as code - using either CMake's language or Python. This would have two major advantages:
1. Anybody will be able to run the same algorithm on new files.
2. We can review and gradually improve the algorithm over time.

src/CMakeLists.txt
oschuett: I'm sorry, but I did not find `$<$<BOOL:${CP2K_USE_CUDA}>:cp2k_cuda_libs>` clear. The syntax does not resemble any other language I'm aware of. 

Besides the syntax the larger problem with this compact style is that it's hard to reason about.
The code would be much easier to understand if it were organized like this:
```
if(FOO)
  # All the effects of FOO.
endif()

if(BAR)
  # All the effects of BAR.
endif()

if(FOO AND BAR)
 # All the effects of FOO and BAR interacting.
endif()
```

HamzaDerim:
CMake Error at CMakeLists.txt:3 (project):
  Generator

    Visual Studio 17 2022

  could not find specified instance of Visual Studio:

    D:/visual studio


Exception: Unable to generate build files

I am using windows 10 operating system
I got this error in my be flutter project, I couldn't find the solution.
visual studio c++ is installed and flutter doctor is running smoothly, i don't get any errors.
please help.

mtaillefumier:
> CMake Error at CMakeLists.txt:3 (project): Generator
> 
> ```
> Visual Studio 17 2022
> ```
> 
> could not find specified instance of Visual Studio:
> 
> ```
> D:/visual studio
> ```
> 
> Exception: Unable to generate build files
> 
> I am using windows 10 operating system I got this error in my be flutter project, I couldn't find the solution. visual studio c++ is installed and flutter doctor is running smoothly, i don't get any errors. please help.

it is a separate issue. Would it be possible to translate this as an issue so i can track it. Note I do not use Windows for development at all and according to the message I suspect something is wrong with visual studio or cmake installation. The cmake build system does not encode anything specific to the generator.

src/CMakeLists.txt
mtaillefumier: > I'm sorry, but I did not find `$<$<BOOL:${CP2K_USE_CUDA}>:cp2k_cuda_libs>` clear. The syntax does not resemble any other language I'm aware of.
> 
> Besides the syntax the larger problem with this compact style is that it's hard to reason about. The code would be much easier to understand if it were organized like this:
> 
> ```
> if(FOO)
>   # All the effects of FOO.
> endif()
> 
> if(BAR)
>   # All the effects of BAR.
> endif()
> 
> if(FOO AND BAR)
>  # All the effects of FOO and BAR interacting.
> endif()
> ```


$<$<BOOL:${CP2K_USE_CUDA}>:cp2k_cuda_libs> is clear. It just mean check the value of CP2K_USE_CUDA (BOOL is to indicate the type) and if true add the target otherwise return an empty string. Really I have seen far more complicated syntax than this. the reverse polish notation been one.

The cmake community uses generator expressions when they can. Generators expressions have many advantages compared to the conventional if - then -else statement. it is possible to create a list of them and just pass the list to any functions that supports this functionality. The target functions are a typical use case.

src/grid/CMakeLists.txt
mtaillefumier: I wrote what you suggest me to do before reverting to a list. It is simply not manageable to take all files and then remove the files that contains main functions for instance of should be included in libraries only. 
We still have to do `git add/mv/rm` and it is really easy to search the same string in the list and remove or rename the item.

src/CMakeLists.txt
dev-zero: For the generator expressions: When in Rome, Do as the Romans Do. They are everywhere in modern CMake and some things (like object lib dependencies and the like) are likely not possible anymore without them. For better readability it makes sense to [distribute them over multiple lines](https://github.com/cp2k/dbcsr/blob/0a78582d8aed3f53915c101773e5a535e16770d8/src/CMakeLists.txt#L231-L254) and use indentation accordingly.

And when looking at Bash variable substitution or Ansible (Jinja2) filter expressions you will see that besides the different bracketing style they are not that uncommon.

src/grid/CMakeLists.txt
dev-zero: Also, [GLOB is discouraged](https://cmake.org/cmake/help/latest/command/file.html#filesystem) by CMake as it may break generators and adds cost to each run. Comparing the number of times people will call CMake to simply build, to the number of times new files are added, the case is quiet clear.

src/CMakeLists.txt
mtaillefumier: yes indentation is important (from a visual point of view). I will check again that I did not miss one or two.

oschuett: > some things (like object lib dependencies and the like) are likely not possible anymore without them. 

Ok, if they are actually unavoidable then so be it.

src/grid/CMakeLists.txt
oschuett: The fundamental problem is that we have over 1k source files, 50 preprocessor flags, and 30 dependencies. So we need a way to fix entire classes of bugs and not just specific combinations of the above. Otherwise we're setting ourselves up for an endless Whac-A-Mole of silly bugs. We need to raise the level of abstraction.

> It is simply not manageable to take all files and then remove the files that contains main functions ...

> Also, GLOB is discouraged...

So then let's resurrect the [generate_cmakelists.py](https://github.com/cp2k/cp2k/blob/74355ed86a302d19cf0c2ebc9855f0ecbd61f87e/cmake/generate_cmakelists.py). It will allow us to fully express the underlying algorithm while also following CMake customs.

And you're right, simple changes can be made directly in the CMakefile. Hence, you could just consider it a Linter ;-)

mtaillefumier: Whatever way the problem is "solved" the information needs to be encoded. I reverted the PR introducing what you want to reintroduce because what was done made no sense in practice. The 1k files / 30 dependencies argument falls short as well because this information needs to be encoded somewhere. I prefer the explicit list because it is clear what's going on. Then including or excluding dependencies can be handled by adding files to the list (not removing them actually) with if statements and compilation flags which we do already.   

Only the developer creating the PR knows about this at instant t. This should be reflected both in the build system and git. No automatic generation of the cmakelists.txt should be allowed or modified without express consent of the developer for traceability reason. It means the CI/CD or the merge process should not modify any files (even formatting should be an argument for rejection which we have) outside fixing merge conflicts (which we do not have in practice). 

To take an example that will speak to many, the version between a submitted paper and the officially published one is very often different. However the soon to be published version is always sent to the authors for proof readings before publication (it is the validation process on the authors side). It is almost never done automatically in high impact journals (i speak about nature, science, APS papers here).
 
I was not really describing an algorithm to generate this list but exposing arguments against the idea. It was interpreted as an algorithm which was not the purpose of the comment. @dev-zero was pointing out that GLOB should be avoided whenever possible he was not suggesting anything else.

oschuett: > The 1k files / 30 dependencies argument falls short as well because this information needs to be encoded somewhere.

The information is already encoded in the source files. The old build system relies on this. So whatever you write into the CMakefiles is redundant and therefore susceptible to inconsistencies.

>  I prefer the explicit list because it is clear what's going on. 

I realize that writing the high-level algorithm is more work and I'm happy to do it myself. I'm certain it will pay off as we tweak and refine it over the coming months/years. Having a process for systematic improvements is the essence of software engineering.

> No automatic generation of the cmakelists.txt should be allowed or modified without express consent of the developer for traceability reason. It means the CI/CD or the merge process should not modify any files....

I never suggested that it should happen automatically. The CI would [merely check](https://github.com/cp2k/cp2k/blob/74355ed86a302d19cf0c2ebc9855f0ecbd61f87e/tools/docker/scripts/test_python.sh#L31) that the committed CMakefile is consistent with what the generator would produce, i.e. it's essentially a very strict linter.

So, I would suggest we merge this PR and then I'll add a matching `cmakelists_linter.py`. What do you think?

mtaillefumier: > 



> The information is already encoded in the source files. The old build system relies on this. So whatever you write into the CMakefiles is redundant and therefore susceptible to inconsistencies.

- well the argument is based on the implicit vs explicit idea. coding this information directly in the source file is implicit and still error prone as it does not forbid compiling files that should not be compiled in the first place. The explicit argument (the CMakeLists.txt is the only source of trust thanks @dev-zero for pointing that out) remove this. 

- we put pressure on the preprocessor which will eventually generate empty files (which breaks C and C++ standards in theory, fortran compilers often expect some end statement at the end as well) for no real reasons. 

This is really not the topic of this PR, which was initially a bug fix and bacame a bit more than that (I had to fix typos etc...).

> I realize that writing the high-level algorithm is more work and I'm happy to do it myself. I'm certain it will pay off as we tweak and refine it over the coming months/years. Having a process for systematic improvements is the essence of software engineering.

you are more skilled in python than I do. if there is a way to do this with us agreeing about it I am sure you can get this working.

> I never suggested that it should happen automatically. The CI would [merely check](https://github.com/cp2k/cp2k/blob/74355ed86a302d19cf0c2ebc9855f0ecbd61f87e/tools/docker/scripts/test_python.sh#L31) that the committed CMakefile is consistent with what the generator would produce, i.e. it's essentially a very strict linter.
> 
> So, I would suggest we merge this PR and then I'll add a matching `cmakelists_linter.py`. What do you think?

I would support a consistency check by all means. But I will only consider the `CMakeLists.txt` as the only explicit source of truth not the source tree nor the content of the source files. People can commit file accidentally and this can result in catastrophic error along the way even in the implicit case. 

- The list should remain explicit if we want the system to be reliable. Modifying the list explicitly imply a commit (separated or not does not matter) etc... Which means we can tract it and validate it (the validation are the regtests etc...). 

- the difference between applications and libraries should also be considered. 

- if the list should be separated from the CMakeLists.txt and stored in a separate file for convenience with parsing then so be it (we can call it compiled_files_list.txt for instance and located in src). This file become defacto the **only** source of truth for the list of compiled files but no automatic generation of this file either (not even offer the possibility to do so) because it will encourage bad behavior. 

-The CMakeLists.txt should contain all cmake commands for targets etc....

oschuett: > I would support a consistency check by all means. 

Awesome :-)

> if the list should be separated from the CMakeLists.txt and stored in a separate file for convenience with parsing then so be it (we can call it compiled_files_list.txt for instance and located in src).

Yes, moving the source of truth to a config file would make the consistency check a lot easier!

> This file become defacto the only source of truth for the list of compiled files but no automatic generation of this file either (not even offer the possibility to do so) because it will encourage bad behavior.

Got it! Works for me.

> if there is a way to do this with us agreeing about it I am sure you can get this working.

I'm sure we'll find a way. I'm pretty flexible about the details. I mostly care about the ability to eliminate entire bug-classes at once.

PR #2627, opened 2023-02-24
Activate force/stress debug printing from input 

juerghutter:
Activate force/stress debug printing from input for TDDFT and energy corrections


oschuett:
Maybe you can also replace [this STOP](https://github.com/cp2k/cp2k/blob/master/src/xc/xc_fxc_kernel.F#L132) with a `CPABORT` as part of this PR?

juerghutter:
I will remove the 'stop' in the next change to the fxc potential method.

PR #2628, opened 2023-02-24
fix logicals for rtp/ehrenfest

mattiatj:
Make the names more intuitive to catch what is actually going on.

The fixed ion case should be handled differently than the case where the ks matrix is complex, as it also marks a different runtype (if .true. : rt_propagation if .false. : ehrenfest_dyn)

The reason is that this PR
https://github.com/cp2k/cp2k/pull/2524
broke one use case where the velocity representations of the expectation values where used.

Example input:
[H2O-delta-velrepr.txt](https://github.com/cp2k/cp2k/files/10824801/H2O-delta-velrepr.txt)


oschuett:
Since you already have an example input ([H2O-delta-velrepr.txt](https://github.com/cp2k/cp2k/files/10824801/H2O-delta-velrepr.txt)), how about turning that into a regression test?

mattiatj:
> Since you already have an example input ([H2O-delta-velrepr.txt](https://github.com/cp2k/cp2k/files/10824801/H2O-delta-velrepr.txt)), how about turning that into a regression test?

Very good idea :)

PR #2631, opened 2023-02-25
open shell in periodic GW, printing of space-local gap

JWilhelm:


JWilhelm:
The sdbg regtest is failing: 2 tests show "Runtime failure with code 23." (for other regtests, my local cp2k run also exits with "Runtime failure with code 23." but the regtester does not fail for these other regtests). Unfortunately, I cannot see in the detailed log what the issue is. Does maybe anyone know why the regtester fails?

src/rpa_gw.F
```diff
@@ -1539,6 +1596,366 @@ SUBROUTINE check_NaN(array, real_value)
...
+            DO i_img = 1, nimg
+               CALL dbcsr_set(rho_ao_weighted(i_spin, i_img)%matrix, 0.0_dp)
+            END DO
+         END DO
+
+      END DO
+
+      DO ikp = 1, nkp
+
+      END DO
```
oschuett: This looks funny.

oschuett (commented):
The two regtest fail because of a memory leaked DBCSR matrix. Because that matrix references all kinds of other objects (distribution, etc) the trace shows leaks in several places. Since the report includes only the last 100 lines it does not seem to include the root cause. So, you'll have to check your local logs.

JWilhelm:
Thanks, Ole, for having a look. After having a closer look, the failed regtests use a method that is no longer supported. I therefore removed the regtests and put a CPASSERT in the code.

JWilhelm:
not sure how to resolve the git history... hmm...

oschuett:
You can try something along these lines:
```shell
git branch backup
git fetch origin
git reset origin/master

# Your changes should still be in the working dir.
git status

# Add and commit your changes again.
git add -u
git commit -m "open shell in periodic GW, printing of local gap"
```

PR #2632, opened 2023-02-26
ci: use `GITHUB_ENV`

e-kwsm:
`set-env` is deprecated

- https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
- https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

### DIFF
--- a/.github/workflows/release-tarball.yml
+++ b/.github/workflows/release-tarball.yml
@@ -16,7 +16,7 @@ jobs:
         with:
           submodules: recursive
       - name: Set release tarball name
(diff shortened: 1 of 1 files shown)


oschuett:
Thanks for looking into this. However, we have not used those workflows for years. I think, it's best to simply delete them.

oschuett:
Closing in favor of #2635.
I'm sorry this took your time. I really appreciate the effort.

PR #2633, opened 2023-02-27
Update docker files

mkrack:


oschuett:
I think it would be helpful to see the entire error message. Maybe you could temporarily modify [install_sirius.sh](https://github.com/cp2k/cp2k/blob/bbf16fba41ecaeab08bff5fdc2964a932742a5fd/tools/toolchain/scripts/stage8/install_sirius.sh#L153)?

PR #2639, opened 2023-02-28
open-shell in periodic GW and local gap printing

JWilhelm:


oschuett:
The [G0W0_kpoints_in_self_energy_gap_printing.inp](https://github.com/cp2k/cp2k/blob/master/tests/QS/regtest-gw-kpoints/G0W0_kpoints_in_self_energy_gap_printing.inp) test keeps failing under [i386](https://dashboard.cp2k.org/archive/i386/index.html). It could just be numerical noise.

What would be the best way to tighten the GW band gap a little?

JWilhelm:
Unfortunately, there are many parameters that influence the GW energy levels. For me, it would be perfectly fine to relax the threshold of this regtest to 0.02.

PR #2640, opened 2023-02-28
Update docker files

mkrack:
- Avoid multiple slashes in path names
- Add links also for tool programs within the container
- Revise docker file for Intel oneAPI to include more functionalities


oschuett:
I realized that in the CI the disk IO was limited to 12Mbit/s. I've now upgraded it to SSD, which should speed things up by 4x.

mkrack:
@oschuett Great!

oschuett:
The Intel build got stuck again. There were 31 `mcpcom` processes that wouldn't finish. So, I just logged in and killed them manually.

oschuett:
The output from before everything got stuck suggests that the compiler got into an infinite inclusion loop. Maybe there is a `#pragma once` missing somewhere?

mkrack:
> The output from before everything got stuck suggests that the compiler got into an infinite inclusion loop. Maybe there is a `#pragma once` missing somewhere?

@oschuett Thanks for looking into that. I did not check the SIRIUS code, yet. The question is why does the SIRIUS build fail in the GC container while it works otherwise?

oschuett:
I don't know why SIRIUS doesn't work with the Intel compiler. In [the past](https://github.com/cp2k/cp2k/blob/4e8b7c4a5ed53c7d43bc840588464933da05e6c1/tools/docker/generate_dockerfiles.py#L493-L494) I had similar problems with SpFFT and therefore never even tried SIRIUS.

mkrack:
> I don't know why SIRIUS doesn't work with the Intel compiler. In [the past](https://github.com/cp2k/cp2k/blob/4e8b7c4a5ed53c7d43bc840588464933da05e6c1/tools/docker/generate_dockerfiles.py#L493-L494) I had similar problems with SpFFT and therefore never even tried SIRIUS.

SpFFT is now fine after fixing the issue with slashes.
Unfortunately, it seems that the CMakeLists.txt file of SIRIUS does not allow to adapt the compiler options in a flexible manner. An Intel compiler detection is present, but without effect.

mkrack:
@mtaillefumier Do you have any hint how to resolve this issue?

oschuett:
It got stuck again :-(
I've clicked "chancel" because otherwise it would just run for 3 hours and waste money.

oschuett:
Perhabs we first switch to the new ICX compiler before spending more time on this:
```
micpc: remark #10441: The Intel(R) C++ Compiler Classic (ICC) is deprecated and will be removed from product release in the second half of 2023. The Intel(R) oneAPI DPC++/C++ Compiler (ICX) is the recommended compiler moving forward. Please transition to use this compiler. Use '-diag-disable=10441' to disable this message.
```

oschuett:
I've [moved](https://github.com/cp2k/cp2k-ci/commit/95342dfe137613a145dbb8e82da7fcd7fcb90b9d) the Intel test to an [Intel machine](https://cloud.google.com/compute/docs/compute-optimized-machines#c2_machine_types). For posterity, [this](https://storage.googleapis.com/cp2k-ci/run-cp2k-intel-psmp-04f8a370_report.txt) was the report on AMD EPYC.

mkrack:
> I've [moved](https://github.com/cp2k/cp2k-ci/commit/95342dfe137613a145dbb8e82da7fcd7fcb90b9d) the Intel test to an [Intel machine](https://cloud.google.com/compute/docs/compute-optimized-machines#c2_machine_types). For posterity, [this](https://storage.googleapis.com/cp2k-ci/run-cp2k-intel-psmp-04f8a370_report.txt) was the report on AMD EPYC.

@oschuett Thank you. Is there a way to wipe the docker cache completely?

oschuett:
> Is there a way to wipe the docker cache completely?

Done.

oschuett:
The Intel tester still runs out of memory while building SIRIUS - although the machine has 120GiB!

Most of the time while building SIRIUS it "only" uses 60GB, but then very suddenly it seems to request a lot more. Maybe this is the link time optimization?

mkrack:
The SIRIUS build worked with the AMD EPYC (Milan). Do the Intel Xeon have less memory?

oschuett:
> The SIRIUS build worked with the AMD EPYC (Milan). Do the Intel Xeon have less memory?

Yes, the EPYC machines had 8GiB more, which appearently did the trick.... or the Intel compiler behaves differently on AMD hardware. In any case, I think we should find a way to get the memory consumption down.

mkrack:
> Yes, the EPYC machines had 8GiB more, which appearently did the trick.... or the Intel compiler behaves differently on AMD hardware. In any case, I think we should find a way to get the memory consumption down.

The only way I see then at the moment is that we drop SIRIUS. I tried already to reduce the opt level for the SIRIUS build to to `-O0` which keeps the memory demands during its compilation within quite moderate margins, but the regression tests become super slow and one test will run much longer than the current limit of 400s. The memory demand with opt level `-O1` is similar to `-O2` on my box, but maybe there are slight savings which let it pass.

oschuett:
As a last attempt we could try `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=OFF`.

If that doesn't work either then we should just call it a day and disable SIRIUS for now.

mkrack:
> As a last attempt we could try `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=OFF`.

I doubt that this flag will have any effect given the current SIRIUS CMakeLists.txt file.

PR #2649, opened 2023-03-03
Farming: rename of variables and introduction of external commands

juerghutter:


src/start/cp2k_runs.F
```diff
@@ -798,12 +805,19 @@ RECURSIVE SUBROUTINE farming_run(input_declaration, root_section, para_env, init
...
 
       ! and message passing deallocate structures
       DEALLOCATE (group_distribution)
-      DEALLOCATE (slave_distribution)
+      DEALLOCATE (minion_distribution)
+
+      ! External commands (global)
+      IF (para_env%is_source()) THEN
+         DO icmd = 1, farming_env%ext_scripts%n_global_final
+            CALL EXECUTE_COMMAND_LINE(farming_env%ext_scripts%ext_global_final(icmd))
```
oschuett: Can you elaborate a bit on your use case. I'm a bit worried about the security implications. Maybe there is a safer solution.

juerghutter: The actual use case would be pre/post-processing of inputs/outputs for farming jobs in a parameter optimization. This would simplify/generalize the interface to the optimizer. However, this can also be achieved with an external script calling CP2K.
The other possible application would be more complex workflows within the farming environment.

oschuett: >  However, this can also be achieved with an external script calling CP2K.

I'd very much prefer this route.

> The other possible application would be more complex workflows within the farming environment.

I don't think we should turn our input files into a scripting language. In my opinion the farming feature was already a step in the wrong direction.

I do agree that there is a lack of a **simple** workflow manager. So, maybe we should once sit together to gather all the requirements and then write a separate Python script that does it cleanly.

juerghutter: Okay, I removed the external command calls.

PR #2654, opened 2023-03-06
[cmake] Add file list consistency tests and others updates

mtaillefumier:
- Add file list consistency checks (directly in CMakelists.txt). The tests can turned on with the option CP2K_ENABLE_CONSISTENCY_CHECKS=ON. CMake will return any error if it find any difference between the src directory and the list. The test is a standalone test and does *not* require anything else than cmake. If ON cmake will only run the tests
- dbcsr can be built at the same time than cp2k (turned OFF by default)
- updated the src/CMakeLists.txt source list
- fix typo in multiple findPackage.cmake modules
- Turned off all options by default (exception mpi and fftw)
- Updated cosma (separate PR)
- Updated the Dockerfile.test_cmake to build dbcsr within cp2k.
- Regtests can be run with the executable generated by the cmake build. (we need to set CP2K_DATA_DIR !!!!!!)
- Updated the README_cmake.md
- initial install script for dbcsr. It is not used by the toolchain yet.

Authors : @mtaillefumier 
Authors : @RMeli


CMakeLists.txt
```diff
@@ -818,20 +904,15 @@ if(NOT CP2K_USE_LIBTORCH)
...
-  STATUS
-    "usual. It is a temporary fix until the regtests properly integrate cmake build systems types."
-)
+if(CP2K_ENABLE_REGTESTS)
+  message(
+    STATUS
+      "\n\n"
+      "To run the regtests you need to run the following commands\n"
+      "\n\n cd ..\n"
+      "tools/regtesting/do_regtest -cp2kdir . -nobuild -version ${__cp2k_cmake_exts} -arch ${__cp2k_cmake_name}\n\n"
```
oschuett: The old `do_regtest` bash script has been replaced by [do_regtest.py](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest.py). The new script uses slightly different arguments.

CMakeLists.txt
```diff
@@ -74,10 +72,14 @@ endforeach()
 # =================================================================================================
 # OPTIONS
 
+option(CP2K_BUILD_DBCSR "Build dbcsr before building cp2k" OFF)
```
oschuett: I see that this is convenient, but I don't think we should go down this road. CMake should only be responsible for configuring and building CP2K. If we let it also install the dependencies then we'll end up again with a hairball like the Toolchain.

CMakeLists.txt
```diff
@@ -74,10 +72,14 @@ endforeach()
...
 option(CP2K_ENABLE_REGTESTS
        "Enable installation of the binaries for running regtests afterwards"
        OFF)
 option(CMAKE_POSITION_INDEPENDENT_CODE "Enable position independent code" ON)
+option(CP2K_ENABLE_CONSISTENCY_CHECKS
```
oschuett, line 82: This is a really nice solution. Thank you for adding it.

CMakeLists.txt
```diff
@@ -192,7 +208,47 @@ cmake_dependent_option(
...
+  set(__cp2k_cmake_name "cmake_build_cuda")
+endif()
+
+if(CP2K_USE_ACCEL MATCHES "HIP")
+  set(__cp2k_cmake_name "cmake_build_hip")
+endif()
+
+# it is really ugly but we have to accomade the regtest script for the time
+# being without having to change it. Instead of trying to run a command after
+# each target been created I change the bin directory.
```
oschuett: I'm happy to adopt the regtest script. What would an ideal interface look like?

oschuett (commented):
Nice progress!
Once this PR is merged I'll enable CMake as a required check in the CI.  I should have done this last week, sorry.

CMakeLists.txt
mtaillefumier: oups. I copied the tests from the python script. I will correct it.

CMakeLists.txt
mtaillefumier: I totally agree with you but I was convinced by some users that we may support the dbcsr build. It is the only dependency I consider like this. It is off by default. spack does not care at all either and it is simpler with the docker images.

CMakeLists.txt
mtaillefumier: no worries

CMakeLists.txt
mtaillefumier: please review it as I may have forgotten something. I basically have it on all the time when I test cp2k. Never encountered an error so far. CMake will fail if there is inconsistency. 

> Nice progress! Once this PR is merged I'll enable CMake as a required check in the CI. I should have done this last week, sorry.

The solution is not perfect. I could also modify cmake such that it only run this test (but no configuration) when the option is turned on (we could run it as a script). So we do not have to create the docker image if the test fails. Basically I would be following the same idea than python but with cmake (that has all information). Drop a word if you consider the solution viable and I can do the necessary adjustments in the PR.

CMakeLists.txt
mtaillefumier: Actually I should revise the comment (there is a typo). I adapted the cmake build to accommodate the regtests as it is simpler that way. 
If the regtests are activated then all binaries will be located in `exe/cmake-build-{cuda,hip,cpu}` (and psmp or ssmp).  
the other reason is that people can run the regtests the conventional way. Of course make install still works as intended.

mtaillefumier:
we may need to update cosma. I will do these changes tomorrow (typo etc...).

CMakeLists.txt
mtaillefumier: replies to comments got mixed up a bit :p

CMakeLists.txt
oschuett: Ok, that's fair, the old build system does it too. I still would like to move DBCSR to the toolchain eventually, but we don't need to conflate this with the CMake migration.

CMakeLists.txt
oschuett: > I could also modify cmake such that it only run this test (but no configuration) when the option is turned on (we could run it as a script).

Yes, that would be great! Then we could add it to [test_python.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_python.sh). We should afterwards rename it to something more generic like `test_misc.sh`. The important point is that these tests do not require a toolchain build.

The reason why I hesitated to make the current CMake test obligatory is that it would drive up our cloud costs because it does require the toolchain.

CMakeLists.txt
oschuett: Ok sounds good.

CMakeLists.txt
mtaillefumier: it makes perfect sense. I need to add one more test to this as I realized I do half of the work. It does not change the logic though.

CMakeLists.txt
mtaillefumier: Agreed. I found an issue that needs investigating though. I will try to solve this problem by the end of the week.

I also have a install_dbcsr.sh template already which I could add to the PR if you want.

oschuett: > I also have a install_dbcsr.sh template already which I could add to the PR if you want.

Yes, if you already have a suitable `install_dbcsr.sh` then please include it. I'll take it from there... in a couple of weeks.

mtaillefumier:
I had to update cosma again because there was a bug in 2.6.4. We need to upload these tree files for cosma

https://github.com/eth-cscs/Tiled-MM/archive/refs/tags/v2.2.tar.gz -> Tiled-MM-v2.2.tar.gz
https://github.com/eth-cscs/COSMA/archive/refs/tags/v2.6.5.tar.gz -> COSMA-v2.6.5.tar.gz
https://github.com/eth-cscs/COSTA/archive/refs/tags/v2.2.tar.gz -> COSTA-v2.2.tar.gz

the CI will fail if these files are not present. I will include the dbcsr install file as well

oschuett:
Voilà:

https://www.cp2k.org/static/downloads/Tiled-MM-v2.2.tar.gz
https://www.cp2k.org/static/downloads/COSMA-v2.6.5.tar.gz
https://www.cp2k.org/static/downloads/COSTA-v2.2.tar.gz

mtaillefumier:
Merci. Sorry it took so long to get this going.

tools/toolchain/scripts/stage4/install_cosma.sh
```diff
@@ -35,16 +40,17 @@ case "$with_cosma" in
         echo "COSMA-v${cosma_ver}.tar.gz is found"
       else
         download_pkg_from_cp2k_org "${cosma_sha256}" "COSMA-v${cosma_ver}.tar.gz"
+        download_pkg_from_cp2k_org "${costa_sha256}" "COSTA-v${costa_ver}.tar.gz"
+        download_pkg_from_cp2k_org "${tiled_mm_sha256}" "Tiled_MM-v${tiled_mm_ver}.tar.gz"
```
oschuett: I think this should be `Tiled-MM`, ie. with `-` instead of `_`.

mtaillefumier: thanks.

mtaillefumier:
Thanks for the cmake trigger.

oschuett:
> Thanks for the cmake trigger.

You're welcome :-)

Unfortunately, the branch now contains a merge commit. I usually use `rebase` instead of `pull` because it does not include foreign changes in the PR, which also makes it easier to review:

```
git fetch upstream
git rebase upstream/master
```

mtaillefumier:
I usually do the same with `git pull -r`  (which imply rebase actually).

mtaillefumier:
Ok, I fixed the issue.

tools/docker/generate_dockerfiles.py
```diff
@@ -423,15 +420,23 @@ def install_cp2k_cmake() -> str:
...
 
 RUN echo "\nSummary: Compilation works fine.\nStatus: OK\n"
 
+# Run regression tests.
+ARG TESTOPTS=""
+
+# Output the report if the image is old and was therefore pulled from the build cache.
+# CMD cat $(find ./report.log -mmin +10) | sed '/^Summary:/ s/$/ (cached)/'
+ENTRYPOINT []
+
```
oschuett: This confuses the CI. I'd suggest we enable the regtests in a follow up PR as this one is already quite large. I could take care of it.

mtaillefumier:
Probably better to squash everything in one commit. 

- sanetizer is also there but I still need to test this

tools/toolchain/scripts/stage8/install_spfft.sh
```diff
@@ -108,16 +105,15 @@ case "${with_spfft}" in
...
-            install -d ${pkg_install_dir}/lib/cuda
-            [ -f src/libspfft.a ] && install -m 644 src/*.a ${pkg_install_dir}/lib/cuda >> install.log 2>&1
-            [ -f src/libspfft.so ] && install -m 644 src/*.so ${pkg_install_dir}/lib/cuda >> install.log 2>&1
+            make -j $(get_nprocs) install > install.log 2>&1 || tail -n ${LOG_LINES} install.log
+            cd ..
             ;;
           Mi50 | Mi100 | Mi200 | Mi250)
             [ -d build-hip ] && rm -rf "build-hip"
-            mkdir build-hip
+            mkdirdir build-hip
```
oschuett: typo

tools/toolchain/scripts/stage8/install_spla.sh
```diff
@@ -189,10 +186,11 @@ export SPLA_HIP_LDFLAGS="${SPLA_HIP_LDFLAGS}"
 export CP_DFLAGS="\${CP_DFLAGS} IF_HIP(-D__OFFLOAD_GEMM|) IF_CUDA(-D__OFFLOAD_GEMM|) ${OFFLOAD_DFLAGS} IF_MPI(-D__SPLA|)"
 export CP_CFLAGS="\${CP_CFLAGS} ${SPLA_CFLAGS}"
 export SPLA_LIBRARY="-lspla"
-export SPLA_ROOT="$pkg_install_dir"
-export SPLA_INCLUDE_DIR="$pkg_install_dir/include/spla"
+export SPLA_ROOT="IF_HIP("${pkg_install_dir}-hip"|IF_CUDA("${pkg_install_dir}-cuda"|"${pkg_install_dir}"))"
+export SPLA_INCLUDE_DIR="${pkg_install_dir}/include/spla"
 export SPLA_VERSION=${spla-ver}
 export CP_LIBS="IF_MPI(${SPLA_LIBS}|) \${CP_LIBS}"
+export CP_LDFLAGS="\${CP_LDFLAGS} IF_HIP(${SPLA_HIP_LDFLAGS}|IF_CUDA(${SPLA_CUDA_LDFLAGS}|${SPLA_LDFLAGS}))"
```
oschuett: This line seems to confuse `parse_if.py`. I don't understand why, which is a bit worrying.

Nevertheless, is this line really need? It seems redundant with the following if/elif/else block.

oschuett:
Wow, this PR is really a tough nut to crack! Is there anything I can to do help?
Should we maybe simplify by moving the COSMA upgrade to a separate PR?

tools/toolchain/scripts/stage8/install_spla.sh
mtaillefumier: Thanks for spotting this. I forgot to remove it. I had issue with parenthesis put in the wrong place and I could figure out where it was coming from. This the typical if then else came handy. I will remove the line and leave the if-them-else structure.

mtaillefumier:
> Wow, this PR is really a tough nut to crack! Is there anything I can to do help? Should we maybe simplify by moving the COSMA upgrade to a separate PR?

I could update cosma with a separate PR. It is probably a good idea anyway. Maybe do the same with SIRIUS so we can have a pure cmake PR without tinkering with the toolchain. I will try this tomorrow.

COSMA is not causing the trouble SIRIUS is. CMAKE / CUDA / CPU tests pass I just need to understand why it fails with rocm. I probably miss something when I specialize SIRIUS compilation for CUDA/ROCM. It works with CUDA but fails with ROCM. I will debug that tomorrow hopefully. 

Your double checking is already very useful :).

mtaillefumier:
@oschuett : do not trigger the others tests just yet. I opened a PR that updates cosma only. after being merged I will update this PR with the new master.
the SIRIUS thingy is also split to ease things. currently debugging the remaining issue.

mtaillefumier:
hmm git history is incorrect. How is that possible. It is one commit only and up to date with master

oschuett:
> hmm git history is incorrect. How is that possible. It is one commit only and up to date with master

Yeah, Github is sometimes slow to reset the `mergeable` attribute after a push. Unfortunately, their REST API does not seem to offer a way to recognize this situation. As a result the CI gets occasionally stuck.

As a workaround you can simply close and re-open the PR.

mtaillefumier:
CP2K_ENABLE_CONSISTENCY_CHECKS=ON will only check if these is inconsistencies between the CMakeLists.txt and the files found in src.... CP2K will not be compiled if on.

mtaillefumier:
@oschuett : thanks. it will be a bit hectic the next few days. I will try to finish this asap

oschuett:
I'd propose that we merge this PR now as is, because then I can add `CP2K_ENABLE_CONSISTENCY_CHECKS` to the Python/Misc test and we no longer have to worry about missing new files.

mtaillefumier:
please do if you have no further issue. Sorry for the very long delay

PR #2663, opened 2023-03-09
Add more initializers

fstein93:
This PR adds initializers to all types of the modules in the directories common, eri_mme, pw_env, subsys and introduces suitable simplifications the respective modules.


fstein93:
The failing tests of the Intel oneAPI tester are already known.

Issue #2117, opened 2022-05-18
Failing test with GCC 12 and OpenBLAS 0.3.20 (zgemv_n_4.c)


mkrack:
The test `SIRIUS/regtest-1/He-full-potential.inp` gives with GCC 12.1 and OpenBLAS 0.3.20 the following SIGSEGV error in the routine `zgemv_n_4.c:499`

```
[get_singular_components] smallest eigen-value of the singular components:   0.0013665705096840
[diag_full_potential_first_variation_davidson] step: 0, current subspace size: 61, maximum subspace size: 341
[diag_full_potential_first_variation_davidson] step: 1, current subspace size: 101, maximum subspace size: 341
[diag_full_potential_first_variation_davidson] step: 2, current subspace size: 141, maximum subspace size: 341
[diag_full_potential_first_variation_davidson] step: 3, current subspace size: 181, maximum subspace size: 341
[diag_full_potential_first_variation_davidson] step: 4, current subspace size: 221, maximum subspace size: 341

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x2b3ad79243ff in ???
#1  0x80bc617 in zgemv_n
	at ../kernel/x86_64/zgemv_n_4.c:499
#2  0x7eda21d in zgemv_
	at /data/user/krack/rt/gnu-latest-mpich/cp2k/tools/toolchain/build/OpenBLAS-0.3.20/interface/zgemv.c:263
#3  0x81bc379 in zlatrd_
	at /data/user/krack/rt/gnu-latest-mpich/cp2k/tools/toolchain/build/OpenBLAS-0.3.20/lapack-netlib/SRC/zlatrd.f:259
#4  0x81a67e2 in zhetrd_
	at /data/user/krack/rt/gnu-latest-mpich/cp2k/tools/toolchain/build/OpenBLAS-0.3.20/lapack-netlib/SRC/zhetrd.f:318
#5  0x81a3ddd in zheevx_
	at /data/user/krack/rt/gnu-latest-mpich/cp2k/tools/toolchain/build/OpenBLAS-0.3.20/lapack-netlib/SRC/zheevx.f:449
#6  0x75f6627 in _ZN18Eigensolver_lapack6solve_ISt7complexIdEEEiiiRN4sddk7dmatrixIT_EEPN4RealIS5_E4typeES7_
[...]

PR #2666, opened 2023-03-09
Retire older Intel testers

mkrack:


alazzaro:
Out of curiosity, not related to this PR (sorry for that), is there an explanation why the https://dashboard.cp2k.org/archive/psi-gnu-mpich-psmp/index.html is always failing?

mkrack:
No clue, there are always the same one or two tests running into the time limit of 400s when the regression test runs as a slurm batch job. The problem does not occur when I run the same test interactively on the login node.

alazzaro:
> No clue, there are always the same one or two tests running into the time limit of 400s when the regression test runs as a slurm batch job. The problem does not occur when I run the same test interactively on the login node.

Indeed, `QS/regtest-kp-2/cc2.inp` is always failing since this https://github.com/cp2k/cp2k/commit/452b3252e6f2376ff1818a3a99c9ccc117ee600e. But then I noticed the following in the output of the [dashboard](https://dashboard.cp2k.org/archive/psi-gnu-mpich-psmp/commit_452b3252e6f2376ff1818a3a99c9ccc117ee600e.txt):

```
# leak related to mpi or scalapack  triggers sometimes for regtest-kp-2/cc2.inp
leak:Cblacs_gridmap
leak:blacs_gridmap_
```
So maybe it is something related to leaks?

Update: Forget my previous comment, the output was always there...

mkrack:
The test `QS/regtest-kp-2/cc2.inp` fails with TIMEOUT already much longer, i.e. at least since 400 commits ([78096bc](https://github.com/cp2k/cp2k/commit/78096bc50ec3c41c1825630b7cc5ec9938bfc15f)) whereas [452b325](https://github.com/cp2k/cp2k/commit/452b3252e6f2376ff1818a3a99c9ccc117ee600e) is "only" 135 commits ago.

alazzaro:
Oh, I see, so it seems related to the switch to GCC 12.2 then...

mkrack:
Yes, that's most likely the reason. There has also been an [issue](https://github.com/cp2k/cp2k/issues/2117) with OpenBLAS and GCC 12 which did not occur with GCC 11.3. Therefore, I still recommend to use rather GCC 11.3 instead of GCC 12.2 to create CP2K production binaries.

mkrack:
@alazzaro The problem seems to appear because of the leak sanitizer. I have now disabled the leak detection with GCC 12.2 for testing.

PR #2673, opened 2023-03-10
ADDED_MOS for EC with MO response solver, some regtests for TDDFPT GAP…

juerghutter:
…W dipoles


oschuett:
The `regtest-debug-6/ch2o_xc_dip46.inp` ran too slow a [couple of times](https://dashboard.cp2k.org/archive/current-sdbg/index.html) last week. Maybe there is a way to speed it up?

juerghutter:
Will be fixed with the next update.

Issue #2668, opened 2023-03-09
Several RTP tests fail with wrong results using the Intel compiler


mkrack:
Several RTP tests show large deviations from the reference values using the Intel compiler (classic as well as oneAPI). This problem was introduced with commit [d463602](https://github.com/cp2k/cp2k/commit/d4636021b70e064aed1ca2668bd7bd77d889a5f3).

fstein93:
@edditler Can you check it?

edditler:
I can have a closer look tomorrow. Unfortunately, I'm not sure if I have access to an Intel compiler. 

The last time the Intel compiler behaved differently for me was in https://github.com/cp2k/cp2k/pull/1706#issuecomment-972937935, where I didn't nullify some pointers before using them. Would you have a pointer towards what else might be wrong?

Additionally, in https://github.com/cp2k/cp2k/pull/2577 a bug related to RTP, that I introduced, was fixed. (@mattiatj)

mkrack:
@edditler The Intel oneAPI HPC Toolkit can be downloaded [here](https://www.intel.com/content/www/us/en/developer/tools/oneapi/hpc-toolkit.html). Alternatively, you can run the [CP2K docker file](https://github.com/cp2k/cp2k/blob/master/tools/docker/Dockerfile.test_intel-psmp).

mkrack:
The CP2K binary created with the Intel compiler gives correct/consistent results with different MPI ranks for the test `QS/regtest-rtp-5/H2O-dens-pulse.3.inp`, but as soon as OMP_NUM_THREADS is larger than 1, the results show the variation detected in the regression tests.

edditler:
I'm not entirely sure how the errors in the RTP code are related to the changes I made. So far, I focussed on the VCD code by attaching to the docker build by

```
docker build --shm-size=1g -f Dockerfile.test_intel-psmp -t cp2k_intel ../../
docker run --shm-size=4g -it cp2k_intel:latest bash
```

Running the regtest `h2o_aat.inp` repeatedly by

```
root@b4e683f44dc3:/opt/cp2k/tests/QS/regtest-dcdr# source /opt/cp2k-toolchain/install/setup 
root@b4e683f44dc3:/opt/cp2k/tests/QS/regtest-dcdr# mpiexec -n 2 /opt/cp2k/exe/local/cp2k.psmp -i h2o_aat.inp
```

[...]

PR #2676, opened 2023-03-11
Conventions: Add check for ASSOCIATE inside OMP PARALLEL region

oschuett:
See #2668 for discussion.


oschuett:
The new check [found 11 places](https://storage.googleapis.com/cp2k-ci/run-cp2k-conventions-676dacda_report.txt) where an `ASSOCIATE` statements might interact badly with OpenMP.
@edditler, it would be great if you could remove those as a precaution.

fstein93:
Just a short question: how does it treat ASSOCIATE construct outside of OpenMP regions (wrapping OpenMP or independently but in the same translation unit)?

oschuett:
Thanks @edditler for the quick response!

> how does it treat ASSOCIATE construct outside of OpenMP regions (wrapping OpenMP or independently but in the same translation unit)?

The check looks at one procedure at a time and only complains when it finds an `ASSOCIATE` somewhere between `!$OMP PARALLEL` and  `!$OMP END PARALLEL`.

PR #2678, opened 2023-03-13
Remove associate constructs within OMP regions

edditler:
This addresses the new coding conventions (see https://github.com/cp2k/cp2k/pull/2676)


edditler:
The Intel oneAPI regtests are now passing on my machine (using the corresponding Dockerfile).

Issue #20, opened 2018-10-04
Fix manual triggering of check_runs.


oschuett:
Currently, this feature only works for PR that have their head_branch in the same repo.

oschuett:
Fixed.

PR #2679, opened 2023-03-13
Hotfix restart (#20)

glb96:
WARNING: MAY AFFECT THE RESTART PROCEDURE

Since a previous PR on the 3rd of February ( Linear density delta kick and restart https://github.com/cp2k/cp2k/pull/2543 ), the old restart files cannot be read properly.
I have removed previous modifications done on the qs_mo_io.F for backward compatibility.

The mo%uniform_occupation logical is now defined in the RTP part.

I apologize if you ran into problems with your restart files generated between the 3rd of February and now.
If it is the case, you can use the code from before this PR regarding the read_mos_restart_low function to load the restart file and write it again using the version of this PR of write_mo_set_low.

### DIFF
--- a/src/emd/rt_propagation_utils.F
+++ b/src/emd/rt_propagation_utils.F
@@ -307,6 +307,8 @@ CONTAINS
       CASE (use_restart_wfn)
          CALL read_mo_set_from_restart(mo_array, atomic_kind_set, qs_kind_set, particle_set, para_env, &
                                        id_nr=id_nr, multiplicity=dft_control%multiplicity, dft_section=dft_section)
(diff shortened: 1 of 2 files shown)


mkrack:
Does this mean that we would need a regression test which checks the reading of a legacy restart file in order to detect such issues in the future, e.g. by storing a small restart file for a water molecule in the repository? Would such a test have detected this issue?

juerghutter:
Yes, this problem could have been detected with such a test.

mkrack:
> Yes, this problem could have been detected with such a test.

Thanks, I will add such a test.

PR #2680, opened 2023-03-13
Test wavefunction restart file functionality

mkrack:
The CP2K regression tester should detect issues like #2679


mkrack:
The first test should fail, because the fix #2679 is not yet merged.

PR #2703, opened 2023-03-29
XAS_TDP| fix a bug with KERNEL%XC_FUNCTIONAL

abussy:
When defining the XC kernel in XAS_TDP, the program crashes if a shortcut (and only a shortcut) is used:

```
&KERNEL
   &XC_FUNCTIONAL PBE
   &END XC_FUNCTIONAL
&END KERNEL
```

The above fails, but the latter two options works smoothly:
```
&KERNEL
   &XC_FUNCTIONAL PBE
      &PBE
      &END
   &END XC_FUNCTIONAL
&END KERNEL
```
```
&KERNEL
   &XC_FUNCTIONAL
      &PBE
      &END
   &END XC_FUNCTIONAL
&END KERNEL
```

I was not able to determine exactly why this fails. Instead, there is now a check at the start of the calculation verifying that there is no shortcut used.


oschuett:
Yes, the XC shortcuts are bit of a hack. They are handled by [xc_functionals_expand](https://github.com/cp2k/cp2k/blob/dc34ec903ce29fa3e138f690b3e7fbdf40732b94/src/input_cp2k_check.F#L142).

So, to enable shortcuts for the `XAS_TDP/KERNEL/XC_FUNCTIONAL` section one needs to add it [here](https://github.com/cp2k/cp2k/blob/dc34ec903ce29fa3e138f690b3e7fbdf40732b94/src/input_cp2k_check.F#L117-L130).

An even better solution would be to introduce an intermediate representation of an XC section so that we no longer have to pass around and monkey patch the `section_vals_type`.

abussy:
Thanks for the hint! The bug is fixed now (rather than circumvented).

PR #2708, opened 2023-04-03
Write GW quasiparticle energies with 0.1 meV precision

JWilhelm:
instead of 1 meV precision


JWilhelm:
Hmm, not sure why there is a slow regtest / the sdbg regtester fails?

oschuett:
> Hmm, not sure why there is a slow regtest / the sdbg regtester fails?

That's a known problem, originating from #2673.

PR #2709, opened 2023-04-04
[toolchain] Update COSMA

mtaillefumier:
Update COSMA to the latest version.


mtaillefumier:
Thanks. I will update the cmake PR.

PR #2710, opened 2023-04-04
[toolchain] Use a dedicated directory when building SIRIUS for CPU/CUDA/HIP

mtaillefumier:
Use a separate directory when compiling SIRIUS, spla, and spfft with different flavor of GPU support.


oschuett:
Is this PR still relevant?

mtaillefumier:
it is but no time work on this. I use spack instead all the time. i will close it.

PR #22, opened 2018-10-05
fold cluster coordinates

rybkinjr:
Needed for interfacing to quantum chemistry software.

PR #2712, opened 2023-04-04
Fix restart rtp vg (#22)

glb96:
Create a new input to define an initial value for the vector potential for RTP calculations in the velocity gauge approach. 
This input key is also used to restart an RTP calculation.

### DIFF
--- a/src/cp_control_types.F
+++ b/src/cp_control_types.F
@@ -201,6 +201,8 @@ MODULE cp_control_types
       REAL(KIND=dp)                        :: strength
       REAL(KIND=dp)                        :: phase_offset
       REAL(KIND=dp)                        :: wavelength
+      REAL(KIND=dp), DIMENSION(:), &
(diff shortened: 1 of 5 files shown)


src/cp_control_types.F
```diff
@@ -201,6 +201,7 @@ MODULE cp_control_types
       REAL(KIND=dp)                        :: strength
       REAL(KIND=dp)                        :: phase_offset
       REAL(KIND=dp)                        :: wavelength
+      REAL(KIND=dp), DIMENSION(:), POINTER :: vec_pot_initial
```
fstein93: Can you turn that one into an ALLOCATABLE? Apparently, you do not really need the features of a POINTER. (Replace ASSOCIATED with ALLOCATED in the respective occasion). ALLOCATABLEs are safer in use than POINTERs.

oschuett:
Closing in favor of #2713.

PR #2713, opened 2023-04-05
Fix restart rtp vg

glb96:
Create a new input to define an initial value for the vector potential for RTP calculations in the velocity gauge approach.
This input key is also used to restart an RTP calculation.

### DIFF
--- a/src/cp_control_types.F
+++ b/src/cp_control_types.F
@@ -194,13 +194,14 @@ MODULE cp_control_types
 ! **************************************************************************************************
    TYPE efield_type
       REAL(KIND=dp)                        :: actual_time
-      REAL(KIND=dp), DIMENSION(:), POINTER :: polarisation
+      REAL(KIND=dp), DIMENSION(:), POINTER :: polarisation => NULL()
       INTEGER                              :: envelop_id
-      REAL(KIND=dp), DIMENSION(:), POINTER :: envelop_r_vars
+      REAL(KIND=dp), DIMENSION(:), POINTER :: envelop_r_vars => NULL()
       INTEGER, DIMENSION(:), POINTER       :: envelop_i_vars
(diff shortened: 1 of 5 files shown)


glb96:
Hey @fstein93 
Here is the modified version of the code to move the new object into an allocatable. However, when I compile I have  this error: 

/users/glebreton/Software/cp2k/src/cp_control_utils.F:2174:9:

          efield%vec_pot_initial = tmp_vals
         1

This is strange because I defined one line before the size of  efield%vec_pot_initial to be the one of tmp_vals. 
May you (or anyone) help me?

fstein93:
Modify it to `efield%vec_pot_initial(:) = tmp_vals`. Without providing any boundaries (including implicit one), the standard requires a reallocation of the array according to the size of the arrays on the r.h.s. whereas with the (implicit) boundaries, no reallocation will occur and the sizes of the copied (sub)arrays will be checked.

fstein93:
Why do you open a second pull request for the same feature? If you just want to update the remote branch, use `git push --force` to enforce this update. The CI will restart automatically without the need to open another PR or open/closing the old ones.

glb96:
In the meantime I found that: 

ALLOCATE (efield%vec_pot_initial, source=TRANSFER(tmp_vals, efield%vec_pot_initial))

also work. 
Is it better / equivalent / worst than: 

ALLOCATE (efield%vec_pot_initial(SIZE(tmp_vals))
efield%vec_pot_initial(:) = tmp_vals 

To be honest I am a bit lost about how to use the github properly. Though that opening a draft like this you be easier to discuss than working on the PR directly

glb96:
Sorry for the new PR request. I will work on the same for future PR. 

The new object is now defined to be an allocatable instead of a pointer as requested.

PR #2722, opened 2023-04-12
Inference of Allegro implemented into Fist

gabriele16:
Allegro is an equivariant NN interatomic potential for large-scale systems.


src/manybody_nequip.F
```diff
@@ -440,6 +440,8 @@ SUBROUTINE nequip_energy_store_force_virial(nonbonded, particle_set, cell, atomi
       DEALLOCATE (forces)
 
       CALL timestop(handle)
+      WRITE (*, *) "inside manybody_nequip"
+      WRITE (*, *) "pot_nequip = ", pot_nequip
```
fstein93: Left-over debug code?

src/manybody_allegro.F
```diff
@@ -0,0 +1,480 @@
...
+
+      CHARACTER(LEN=*), PARAMETER :: routineN = 'setup_allegro_arrays'
+
+      INTEGER                                            :: handle, i, iend, igrp, ikind, ilist, &
+                                                            ipair, istart, jkind, nkinds, nlocal, &
+                                                            npairs, npairs_tot
+      INTEGER, DIMENSION(:), POINTER                     :: temp_unique_list_a, work_list, work_list2
+      INTEGER, DIMENSION(:, :), POINTER                  :: list
+      REAL(KIND=dp), DIMENSION(3)                        :: cell_v, cvi
+      REAL(KIND=dp), DIMENSION(:, :), POINTER            :: rwork_list
```
fstein93: Is it possible to replace these POINTERs with ALLOCATABLEs?

src/pair_potential_types.F
```diff
@@ -1795,6 +1830,85 @@ SUBROUTINE pair_potential_nequip_release(nequip)
...
+   SUBROUTINE pair_potential_allegro_create(allegro)
+      TYPE(allegro_pot_type), POINTER                    :: allegro
+
+      CPASSERT(.NOT. ASSOCIATED(allegro))
+      ALLOCATE (allegro)
+      allegro%allegro_file_name = ""
+      allegro%unit_coords = ""
+      allegro%unit_forces = ""
+      allegro%unit_energy = ""
+      allegro%unit_cell = ""
```
fstein93: You can set the components just at the definition of the derived type itself. Do you than still need this routine?

src/pair_potential_types.F
```diff
@@ -1795,6 +1830,85 @@ SUBROUTINE pair_potential_nequip_release(nequip)
       END IF
    END SUBROUTINE pair_potential_nequip_release
 
+! **************************************************************************************************
+!> \brief Creates the ALLEGRO potential type
+!> \param allegro ...
+!> \author Teodoro Laino [teo] 11.2005
```
fstein93: Replace it with your name.

src/pair_potential_types.F
```diff
@@ -1795,6 +1830,85 @@ SUBROUTINE pair_potential_nequip_release(nequip)
...
+      allegro_dest%allegro_version = allegro_source%allegro_version
+      allegro_dest%rcutsq = allegro_source%rcutsq
+      allegro_dest%unit_coords = allegro_source%unit_coords
+      allegro_dest%unit_coords_val = allegro_source%unit_coords_val
+      allegro_dest%unit_forces = allegro_source%unit_forces
+      allegro_dest%unit_forces_val = allegro_source%unit_forces_val
+      allegro_dest%unit_energy = allegro_source%unit_energy
+      allegro_dest%unit_energy_val = allegro_source%unit_energy_val
+      allegro_dest%unit_cell = allegro_source%unit_cell
+      allegro_dest%unit_cell_val = allegro_source%unit_cell_val
```
fstein93: It could be simpler to just write `allegro_dest = allegro_source`

src/pair_potential_types.F
```diff
@@ -1795,6 +1830,85 @@ SUBROUTINE pair_potential_nequip_release(nequip)
...
+      allegro%allegro_version = 'NULL'
+      allegro%rcutsq = 0.0_dp
+      allegro%unit_energy_val = 0.0_dp
+      allegro%unit_coords_val = 0.0_dp
+      allegro%unit_forces_val = 0.0_dp
+      allegro%unit_cell_val = 0.0_dp
+      allegro%unit_coords = 'NULL'
+      allegro%unit_forces = 'NULL'
+      allegro%unit_energy = 'NULL'
+      allegro%unit_cell = 'NULL'
```
fstein93: You can use `allegro = allegro_pot_type()` instead after you have defined proper defaults at the definition of this type.

src/pair_potential_types.F
```diff
@@ -1795,6 +1830,85 @@ SUBROUTINE pair_potential_nequip_release(nequip)
...
+!> \param allegro ...
+!> \author Teodoro Laino [teo] 11.2005
+! **************************************************************************************************
+   SUBROUTINE pair_potential_allegro_release(allegro)
+      TYPE(allegro_pot_type), POINTER                    :: allegro
+
+      IF (ASSOCIATED(allegro)) THEN
+         DEALLOCATE (allegro)
+      END IF
+   END SUBROUTINE pair_potential_allegro_release
```
fstein93, line 1863: Do you really need this function (especially if you have removed the create routine)?

src/manybody_nequip.F
gabriele16: Thanks, it's fixed.

src/pair_potential_types.F
gabriele16: I am setting the defaults in the definition of the derived type for both Allegro and NequIP, but would like to keep this as well as other routines in ```src/pair_potential_types.F``` for consistency and overall clarity, since they are called for the different pair potentials and manybody potentials implemented before.

src/pair_potential_types.F
gabriele16: Done, thanks.

src/pair_potential_types.F
gabriele16: Done, thanks.

src/pair_potential_types.F
gabriele16: As for the create routine, I would like to keep it for consistency with previous code.

src/pair_potential_types.F
gabriele16: Done both for NequIP and Allegro, thanks. In general, should one claim authorship of a routine even if it was only slightly modified from a previous one, as it was the case, for example, for all the create, copy, release, clean routines in ```src/pair_potential_types.F``` for each manybody or pair potential?

fstein93: Because we have the git logs, there is not necessarily the need to add the authorship but if you do, just add a hint that it was copied from code by Teodoro.

src/manybody_allegro.F
```diff
@@ -0,0 +1,480 @@
...
+      INTEGER                                            :: iat, iat_use
+      REAL(KIND=dp), DIMENSION(3, 3)                     :: virial
+      TYPE(allegro_data_type), POINTER                   :: allegro_data
+
+      CALL fist_nonbond_env_get(fist_nonbond_env, allegro_data=allegro_data)
+
+      IF (use_virial) THEN
+         virial(1, 1) = 0.0_dp; virial(1, 2) = 0.0_dp; virial(1, 3) = 0.0_dp
+         virial(2, 1) = 0.0_dp; virial(2, 2) = 0.0_dp; virial(2, 3) = 0.0_dp
+         virial(3, 1) = 0.0_dp; virial(3, 2) = 0.0_dp; virial(3, 3) = 0.0_dp
```
fstein93: I am aware that this branch is not yet properly implemented. But just write `virial = 0.0_dp` and below `pv_nonbond = pv_nonbond + virial`

src/pair_potential_types.F
oschuett: > In general, should one claim authorship of a routine even if it was only slightly modified from a previous one,...

It depends ;-)

For simple boilerplate code like this, you can safely put your own name to let others know that you're the Allegro-expert. For more complex code you can share the credit via the `\history` tag.

Generally, we try to give [more visibility to recent work](https://github.com/cp2k/cp2k/blob/b234498075e47956c66564e93abbe409d5e88a7d/src/common/reference_manager.F#L230-L231).

src/manybody_allegro.F
gabriele16: Done, thanks.

src/manybody_allegro.F
gabriele16: I have done so where possible now. In some cases such as for ```glob_loc_list```, ```glob_cell_v```, ```glob_loc_list_a```, ```unique_list_a```, I have not changed to ALLOCATABLEs as they are defined above in ```src/manybody_potential.F``` and used throughout in different routines and modules.

PR #2747, opened 2023-04-21
RTP| fixed ADMM restart

abussy:
Restarting an ADMM RTP calculation would lead to a crash. This is fixed in this PR.


oschuett:
> Restarting an ADMM RTP calculation would lead to a crash.

Maybe this is worth adding a regtest for?

abussy:
I guess it does not hurt. Added.

Issue #2751, opened 2023-04-24
Limit disk space of regtests


oschuett:
Some regtests write an unreasonable amount of data to disk. For example, the tests in [Pimd](https://github.com/cp2k/cp2k/tree/master/tests/Pimd) produce 2.3 GB.

We should add a check to [do_regtest.py](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest.py) that limits the output to <1MB per test.

PR #2754, opened 2023-04-27
Print only 1 instead of 3 backup copies by default

mkrack:
This should save some disk space.


oschuett:
Thanks a lot for making progress towards #2751.

PR #2764, opened 2023-05-03
Exclude grid_unittest from keepalive mode

mkrack:
Intel oneAPI returns an error for the grid_unittest with --keepalive

### DIFF
--- a/tools/regtesting/do_regtest.py
+++ b/tools/regtesting/do_regtest.py
@@ -37,6 +37,7 @@ KEEPALIVE_SKIP_DIRS = [
     "QS/regtest-gpw-1",
     "QS/regtest-mp2-grad-1",
     "QS/regtest-mp2-grad-2",
+    "UNIT/grid_unittest",
 ]
 
 


oschuett:
I don't think this will have any effect. Unittests are standalone binaries that are [run separately](https://github.com/cp2k/cp2k/blob/cdd1f471827113d909df2ed3821d2ca6a431fa7c/tools/regtesting/do_regtest.py#L498). They don't participate in keepalive.

oschuett:
The sporadic failures of the [Intel test](https://dashboard.cp2k.org/archive/intel-psmp/index.html) might be because the `grid_unittest` binary is not MPI aware and does not call `MPI_Finalize`. I'll have a look.

mkrack:
That's weird. I can reproduce the runtime fail for grid_unittest locally with Intel oneAPI and the RUNTIME FAIL disappears with this PR.

mkrack:
The test itself is correct and shows no issue. It must be indeed something which creates an exit error during the run finalisation.

oschuett:
Closing in favor of #2765.

PR #2771, opened 2023-05-04
Reduce the output of the k point tests

mkrack:
Print a wavefunction restart file only if needed


oschuett:
Thanks a lot for reducing the test outputs!

Are they now below 2MiB? Because then we could remove them from [HUGE_TESTS_SUPPRESSIONS](https://github.com/cp2k/cp2k/blob/master/tests/HUGE_TESTS_SUPPRESSIONS).

mkrack:
The goal of 2 MiB per test is quite ambitious even for print level low. We will have to suppress the printing of any wavefunction restart file for most cases, if want to achieve that goal. Some tests (PIMD) went from about 200 MiB to a few MiB now. I will update the suppression file. It would be nice to have an option with do_regtest.py which allows to set the output limit per test and/or to trigger at least the printing of all tests surpassing that limit.

mkrack:
A further positive side effect is that some tests run much faster now.

oschuett:
> The goal of 2 MiB per test is quite ambitious even for print level low.

Yeah, I wasn't too sure about it either. Should we put it at a more generous 10MiB or something in-between?

> It would be nice to have an option with do_regtest.py which allows to set the output limit per test 

This would require a change of the file format of the `TEST_FILES` e.g. to json. I have it on my extended TODO list ;-)

> and/or to trigger at least the printing of all tests surpassing that limit.

For this you could just empty the `HUGE_TESTS_SUPPRESSIONS` and then run the tests with `--maxerrors=9999`.

PR #2783, opened 2023-05-11
Add a CP2K_BUILD_OPTIONS with some default dependencies.

mtaillefumier:
- Add -DCP2K_BUILD_OPTIONS=DEFAULT,MINIMAL,SERIAL,FULL to avoid specifying which dependency should be included in the build. The default behavior (when the argument is not speficied) lets the user decide what to include.
- Fix the installation of the libcp2k.h header file
- Add option to compile libcp2k shared version

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,11 +148,16 @@ if(NOT ${CP2K_SCALAPACK_VENDOR} IN_LIST CP2K_SCALAPACK_VENDOR_LIST)
   message(FATAL_ERROR "Invalid scalapack vendor backend")
 endif()
 
-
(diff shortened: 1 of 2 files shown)


mtaillefumier:
@oschuett @mkrack : Let me know if you wish more build options (just select the dependencies). 
NB : 

- the default is user full control. 
- cmake supports similar functionalities (Release, etc) but it is probably better to keep this separated as we allow combination minimal-debug etc...

mkrack:
@mtaillefumier Thanks, we can always elaborate further on this later.

mtaillefumier:
too fast. I wanted to include a minor change :/. Never mind I will open an other pr

mkrack:
@mtaillefumier No problem, I just want to avoid having PRs hang around too long for no good reason.

PR #2786, opened 2023-05-15
allegro/nequip: run with float32 or float64, check if TF32 is used, freeze model

gabriele16:
This pull request fixes potential issues arising from PyTorch when running inference with Allegro or NequIP. Some of the features have been added to ```src/torch_api.F``` and ```src/torch_c_api.F``` as calling torch functions may be required.  These features are consistent with [mir-group/pair_allegro/pair_allegro.cpp](https://github.com/mir-group/pair_allegro/blob/main/pair_allegro.cpp). Please let me know if there are some issues @oschuett :)

- Option to run inference with PyTorch either in single or double precision, as specified in the deployed model. Added tests for both single and double precision and adjusted thresholds accordingly.
- Freeze pytorch model, see [torch.jit.freeze.html](https://pytorch.org/docs/stable/generated/torch.jit.freeze.html)
- Check whether TF32 is being used as the default has changed in pytorch >1.11, see [https://pytorch.org/docs/stable/notes/cuda.html](https://pytorch.org/docs/stable/notes/cuda.html). Note that for Allegro/NequIP, using PyTorch v1.11 is recommended for now and pytorch v1.12 to v 2.0.0 are still being tested, see [mir-group/nequip/discussions/311](https://github.com/mir-group/nequip/discussions/311)


src/force_fields_input.F
```diff
@@ -2377,9 +2378,9 @@ SUBROUTINE read_nequip_data(nequip)
 
       CHARACTER(len=*), PARAMETER                        :: routineN = 'read_nequip_data'
 
-      CHARACTER(LEN=default_path_length)                 :: cutoff_str
-      INTEGER                                            :: handle
-      LOGICAL                                            :: found
+      CHARACTER(LEN=default_path_length)                 :: allow_tf32_str, config_str, cutoff_str
+      INTEGER                                            :: handle, i, idx, len_config
+      LOGICAL                                            :: allow_tf32 = .FALSE., found
```
oschuett: Initializing a variable at declaration time creates an implicit save in Fortran. See also https://cp2k.org/conv#c003.

src/force_fields_input.F
```diff
@@ -2395,15 +2396,45 @@ SUBROUTINE read_nequip_data(nequip)
...
+            nequip%do_nequip_sp = .FALSE.
+         ELSE
+            CALL cp_abort(__LOCATION__, &
+                          "NequIP default_dtype should be either <float32> or <float64>. "// &
+                          "Check the .yaml and .pth files.")
+         END IF
+      END IF
+
+      allow_tf32_str = torch_model_read_metadata(nequip%nequip_file_name, "allow_tf32")
+      IF (TRIM(allow_tf32_str) == "1") allow_tf32 = .TRUE.
```
oschuett: Just write `allow_tf32 = (TRIM(allow_tf32_str) == "1")`

src/force_fields_input.F
```diff
@@ -2395,15 +2396,45 @@ SUBROUTINE read_nequip_data(nequip)
...
       nequip%unit_forces_val = cp_unit_to_cp2k(nequip%unit_forces_val, nequip%unit_forces)
-      nequip%unit_energy_val = 1.0_dp
       nequip%unit_energy_val = cp_unit_to_cp2k(nequip%unit_energy_val, nequip%unit_energy)
-      nequip%unit_cell_val = 1.0_dp
       nequip%unit_cell_val = cp_unit_to_cp2k(nequip%unit_cell_val, nequip%unit_cell)
 
+      ! look in config which contains all the .yaml file options to see if we use float32 or float64
+      config_str = torch_model_read_metadata(nequip%nequip_file_name, "config")
+      len_config = LEN_TRIM(config_str)
+      idx = INDEX(config_str, "default_dtype:")
```
oschuett: Parsing YAML like this is really brittle. Maybe you could ask the model makers to add all the information that you need as individual metadata fields?

src/force_fields_input.F
```diff
@@ -2429,20 +2460,55 @@ SUBROUTINE read_allegro_data(allegro)
...
+      END IF
+
+      allow_tf32_str = torch_model_read_metadata(allegro%allegro_file_name, "allow_tf32")
+      IF (TRIM(allow_tf32_str) == "1") allow_tf32 = .TRUE.
+      IF (TRIM(allow_tf32_str) /= "1" .AND. TRIM(allow_tf32_str) /= "0") THEN
+         CALL cp_abort(__LOCATION__, &
+                       "The value for allow_tf32 <"//TRIM(allow_tf32_str)// &
+                       "> is not supported. Check the .yaml and .pth files.")
+      END IF
+      CALL torch_allow_tf32(allow_tf32)
```
oschuett: Instead of duplicating the code better move it into a subroutine.

src/torch_api.F
```diff
@@ -367,4 +370,52 @@ END FUNCTION torch_c_cuda_is_available
...
+
+#if defined(__LIBTORCH)
+      INTERFACE
+         SUBROUTINE torch_c_allow_tf32(allow_tf32) BIND(C, name="torch_c_allow_tf32")
+            IMPORT :: C_BOOL
+            LOGICAL(C_BOOL), VALUE                  :: allow_tf32
+         END SUBROUTINE torch_c_allow_tf32
+      END INTERFACE
+
+      CALL torch_c_allow_tf32(allow_tf32=LOGICAL(allow_tf32, 1))
```
oschuett: Use `LOGICAL(allow_tf32, C_BOOL)` instead.

src/torch_c_api.cpp
```diff
@@ -101,6 +115,17 @@ void torch_c_dict_get_int64(const torch_c_dict_t *dict, const char *key,
   torch_c_dict_get<int64_t>(dict, key, ndims, sizes, dest);
 }
 
+/*******************************************************************************
+ * \brief Retrieves array of floats from Torch dictionary.
```
oschuett: nit: floats -> doubles

src/input_cp2k_mm.F
```diff
@@ -1481,7 +1481,8 @@ SUBROUTINE create_NEQUIP_section(section)
 
       CALL keyword_create(keyword, __LOCATION__, name="ATOMS", &
                           description="Defines the atomic kinds involved in the NEQUIP potential. "// &
-                          "Provide a list of each element.", &
+                          "Provide a list of each element ordered by atomic number. "// &
```
oschuett: What happens when they are not ordered? Does it abort with a nice error message or simply return wrong results?

oschuett (commented):
Overall LGTM. I left a few minor comments.

src/force_fields_input.F
gabriele16: Thank you, I have put it into a subroutine instead of duplicating as suggested. There is ongoing development to make allegro work with lammps in double as well as single precision, so this might be done soon from their side. I will soon contact them about this also to let them know about the implementation. If parsing the YAML file is less than ideal, I could specify the `default_dtype` as a tag in the cp2k input file, as also done for the `ATOMS` tags where a list of chemical species is provided...

src/input_cp2k_mm.F
gabriele16: Good point, thanks a lot. I have modified the keyword description to be more general, since the list of elements and of the atomic coordinates themselves have to be provided in a way that is consistent with the YAML input file of nequip and allegro. If this is not done unphysical results will be obtained.

Please note that the same thing is done in LAMMPS for the case of both nequip and allegro, see for instance [mir-group/pair_allegro] (https://github.com/mir-group/pair_allegro/tree/main) (see FAQ there). I will add a few lines in the tutorial on [www.cp2k.org/howto:allegro](https://www.cp2k.org/howto:allegro) as well to warn the user about this.

Spotting such issues is quite straightforward as the energy is, at least for the tests I ran, significantly wrong. For example, by inverting the order of one of the elements in the test `regtest-nequip/NequIP_water.inp` the error with respect to the reference value is of the order of 1 eV. Additionally, running MD leads rapidly to an highly unstable simulation.

On the other hand, checking all the possible ways in which the atoms can be specified in the `ATOMS` tag and the way they can be ordered in the `&COORD...&END` or in the `&TOPOLOGY...&END` section in such a way that is consistent with the nequip/allegro input file does not provide much benefit in my view. Additionally, to check consistency with allegro/nequip one would have to parse the `config.yaml` file, at least as of right now. :)

PR #2792, opened 2023-05-22
gcc 12.2.0 -> 12.3.0

mkrack:


oschuett:
Almost all CI tests are currently using the system-provided GCC. Only the [OpenMPI test](https://github.com/cp2k/cp2k/blob/d4cb79545b4da7ef5960b7b35925d26c4d2d8d3d/tools/docker/generate_dockerfiles.py#L32) uses the toolchain to install GCC.

PR #2800, opened 2023-05-31
Bump SpLA to version 1.5.5

fstein93:
The new tarball is [here](https://github.com/eth-cscs/spla/archive/refs/tags/v1.5.5.tar.gz).


mkrack:
Done.

PR #2810, opened 2023-06-06
Docs: Add Sphinx-based manual

oschuett:


hfp:
Does this work allow docs to be built and hosted on RTD in addition to CP2K,org?

PR #2813, opened 2023-06-07
Docs: Add Sphinx-based manual

oschuett:


oschuett:
From #2810:
> Does this work allow docs to be built and hosted on RTD in addition to CP2K,org?

@hfp, not without extra efforts because the input reference is generated from an XML file which CP2K produces. Hence, building the manual actually requires building the toolchain and CP2K first.

Why do you think it would be good to have a second copy on RTD? I'd expect it'll only cause confusions.

So, my plan for now is to simply continue hosting it at manual.cp2k.org, where it has been for the last ~20 years.

hfp:
I think it's fine on CP2K.org only. I was just curious if the machinery for building the documentation can run everywhere with no extensive requirements.

( I am using RTD, and I had an issue in the past hooking FORD based documentation into MkDocs. To clarify this is now possible thanks to some more recent MkDocs capability, i.e., hooks to run Python based code for instance. On the other hand is was also an issue because of RTD's fixed flow for building documentation. )

PR #2831, opened 2023-06-20
Splitting of electronic states due to spin-orbit coupling

JWilhelm:
The SOC parameters from Hartwigsen, Goedecker, Hutter (https://doi.org/10.1103/PhysRevB.58.3641), Eq. 18, 19 + Table I, are used to compute the spin-orbit shift of electronic states (see Eq. 11 in https://iopscience.iop.org/article/10.1088/0953-8984/25/10/105503). 

If this PR is too close to the release, I am also happy to postpone.


tests/TEST_TYPES
```diff
@@ -101,6 +101,7 @@ Total Spread (Berry) : !6
 NVP |   1  2  !8
 Alpha GW direct gap !9
 Beta GW direct gap !9
+GW+SOC bandgap (eV) !4
```
oschuett, line 104: You also need to increment the counter on the first line when adding a new test type.

JWilhelm: Thanks, Ole!

oschuett:
Unfortunately, the CI jobs keep getting preempted right now...

JWilhelm:
@oschuett Thanks for having a look at this PR. The conventions seem to fail because of a missing initialization in exstates_types.F which has not been touched by this PR. Do you think there is still something to do to concerning this PR?

PR #2833, opened 2023-06-23
Add several more initializers

fstein93:
This PR adds initializers to (almost?) all derived types defined in the subdirectories of the src directory.

EDIT:
I got some insights: Derived types with ALLOCATABLE components stay uninitialized because these are initialized as soon as each component is initialized after the allocation of the array.


oschuett:
> I got some insights: Derived types with ALLOCATABLE components stay uninitialized because these are initialized as soon as each component is initialized after the allocation of the array.

I've updated [conv#c016](https://www.cp2k.org/conv#c016) to include this insight.

fstein93:
Apparently, the 'missing' initialization of ALLOCATABLEs does not allow a default initialization of the respective objects with some compilers.

PR #2836, opened 2023-06-26
Do not use the default initializer if it has an allocatable component

fstein93:
Follow up to #2833. This fixes testers with old compilers.

### DIFF
--- a/src/cp_eri_mme_interface.F
+++ b/src/cp_eri_mme_interface.F
@@ -78,7 +78,8 @@ MODULE cp_eri_mme_interface
 
    TYPE cp_eri_mme_param
       TYPE(cp_logger_type), POINTER                      :: logger => NULL()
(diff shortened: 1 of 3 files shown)


oschuett:
Thanks a lot for following up @fstein93 !

For the record, this also affected newer versions of GCC:

- [FEDORA TOOLCHAIN (PSMP)](https://dashboard.cp2k.org/archive/fedora-psmp/index.html) GNU Fortran 10.3.1 
- [DEBIAN, I386 (SSMP)](https://dashboard.cp2k.org/archive/i386/index.html) GNU Fortran 10.2.1 
- [CRAY-XC50-GNU.PSMP](https://dashboard.cp2k.org/archive/cray-xc50-gnu-psmp/index.html) GNU Fortran 9.3.0

So, it must depend also on other factors and might even still be present in the latest GCC version.

PR #2839, opened 2023-06-26
Update GTH pseudopotential database

mkrack:
- Add GTH PP database file including the spin-orbit coupling (SOC) parameters.
- Sync the CP2K GTH PP database file with the [cp2k-data repository](https://github.com/cp2k/cp2k-data) (adds mostly medium- and small-core GTH PPs for the 4f and 5f elements based on LDA and PBE)


oschuett:
Should we also [include](https://github.com/cp2k/cp2k/blob/c8a38c8314c6f43066d6e1b608696bdbcb5b4f35/Makefile#L441) the new `GTH_POTENTIALS_SOC` in the `POTENTIALS` file?

oschuett:
One would then also have to add it [here](https://github.com/cp2k/cp2k/blob/c8a38c8314c6f43066d6e1b608696bdbcb5b4f35/tools/precommit/precommit.py#L262).

Edit: Fixed link.

mkrack:
> Should we also [include](https://github.com/cp2k/cp2k/blob/c8a38c8314c6f43066d6e1b608696bdbcb5b4f35/Makefile#L441) the new `GTH_POTENTIALS_SOC` in the `POTENTIALS` file?

No, that would generate duplicate entries. We can later substitute GTH_POTENTIAL by GTH_POTENTIALS_SOC and always readin the SOC parameters, even if they are not used. I went for a separate file to keep the compatibility with the current and older CP2K versions.

PR #2850, opened 2023-07-07
Mpi fixes

mtaillefumier:
fix compilations happening when we use spack.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -306,6 +306,10 @@ endif()
 
 set(__cp2k_cmake_name "cmake_build_cpu")
 
+string(TOUPPER ${CP2K_USE_ACCEL} cp2k_tmp)
+set(CP2K_USE_ACCEL ${cp2k_tmp})
+unset(cp2k_tmp)
+
 if(CP2K_USE_ACCEL MATCHES "CUDA")
   set(__cp2k_cmake_name "cmake_build_cuda")
 endif()
--- a/src/almo_scf.F
+++ b/src/almo_scf.F
@@ -2031,7 +2031,7 @@ CONTAINS
             CALL dbcsr_get_info(matrix_t_tmp(ispin), group=para_group_handle, &
                                 nfullrows_total=Nrows, &
                                 nfullcols_total=Ncols)
-            CALL para_group%set_handle(para_group_handle)
+            CALL para_group%set_handle(INT(para_group_handle))
             CALL para_group%sum(retained(ispin))
 
             !devide by the total no. elements
--- a/src/almo_scf_methods.F
+++ b/src/almo_scf_methods.F
@@ -3244,7 +3244,7 @@ CONTAINS
 
       ndomains = dbcsr_nblkcols_total(dpattern)
       CALL dbcsr_get_info(dpattern, group=GroupID)
-      CALL group%set_handle(groupid)
+      CALL group%set_handle(INT(groupid))
 
       ALLOCATE (subm_no(ndomains), subm_nn(ndomains))
       CALL init_submatrices(subm_no)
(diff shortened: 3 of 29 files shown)


oschuett:
It seems odd that you have to run an integer variable through `INT()` before passing it as an [integer argument](https://github.com/cp2k/cp2k/blob/a7c7cb152c214f2aa9e9d66ac02b6f172b700f7e/src/mpiwrap/message_passing.F#L956).

Could you elaborate what kind of compilation problems you encountered?

mtaillefumier:
I get these kinds of errors with spack

```
4789    /home/mathieut/Projects/codes/cp2k/spack-build-hjm444z/src/qs_active_space_types.F:338:47:
     4790
     4791      338 |          CALL mp_group%set_handle(eri%mp_group)
     4792          |                                               1
  >> 4793    Error: Type mismatch in argument 'handle' at (1); passed TYPE(mp_comm_type) to INTEGER(4)
  >> 4794    make[2]: *** [src/CMakeFiles/cp2k.dir/build.make:10417: src/CMakeFiles/cp2k.dir/qs_active_space_types.F.o] Error 1
```

despite the presence of `-fallow-argument-mimatch` been present. Moreover libcp2k.f for instance does the conversion explicitely so I think it is better to do the conversion explicitely instead of relying on the compiler to do it for us.

mtaillefumier:
a better fix would be to actually to declare `mp_group` with the right type to start with.

oschuett:
> `Error: Type mismatch in argument 'handle' at (1); passed TYPE(mp_comm_type) to INTEGER(4)`

This looks to me like you're using incompatible version of DBCSR and CP2K.

mkrack:
@mtaillefumier Do you get that type mismatch because of the `-D__MPI_F08` flag or in general? Wouldn't it make sense to run at least one test using MPI for this PR?

alazzaro:
Indeed, I was saying the same: which version of DBCSR is used here? 
@hfp already reported some of these incompatibilities at https://github.com/cp2k/cp2k/discussions/2798#discussioncomment-6353624

I'm going to integrate a new DBCSR soon...

mkrack:
I recommend to wait for the new DBCSR release and then to resolve the issues related to MPI_F08 in cp2k. I also got the type mismatches observed by @hfp and @mtaillefumier when I used the DBCSR develop branch with CP2K and specifying the flag `-D__MPI_F08` and `-D__USE_MPI_F08` in the CP2K arch file.

mtaillefumier:
I see. Then this is not the right way to fix the issue. @mkrack I get the error in general as cmake does not enforce the MPI08 thing by default.

alazzaro:
I've just realized that CP2K uses `__MPI_F08` macro, while Frederick did `__USE_MPI_F08` in DBCSR... That means we need to entries in the CP2K arch file. I will make the change in DBCSR to accept `__MPI_F08` too...

oschuett:
We decided to make MPI_F08 opt-in for now because of a bug in GCC/Ubuntu. However, we're already testing it via [openmpi-psmp](https://dashboard.cp2k.org/archive/openmpi-psmp/index.html) and [intel-psmp](https://dashboard.cp2k.org/archive/intel-psmp/index.html).

See also #2486.

mtaillefumier:
> 

can we use `__USE_MPI_F08` or `__MPI_F08` but not both. Or better namespace it because we never know if `__USE_MPI_F08`  is used in dependencies or not.

mtaillefumier:
> 

that's good. Thanks. I should probably close this PR as it is not really the right way to solve the problem. Anyone has another opinion ?

alazzaro:
> We decided to make MPI_F08 opt-in for now because of a bug in GCC/Ubuntu. However, we're already testing it via [openmpi-psmp](https://dashboard.cp2k.org/archive/openmpi-psmp/index.html) and [intel-psmp](https://dashboard.cp2k.org/archive/intel-psmp/index.html).
> 
> See also #2486.

It is the same in DBCSR, of course. The problem is that I don't want to describe two macros to use for the standard makefile compilation via arch files... Therefore I will alias __MPI_F08 macro in DBCSR.

PR #2859, opened 2023-07-13
SOC absorption for TDDFPT

Tarandin:
We, @annahehn and I, have implemented a probational SOC correction for the tddfpt2 module. Basically, we changed the sums used in the xas_tdp module to use all Orbitals. We checked our implementation against a benchmark used by Christel M. Marian et al. (https://doi.org/10.1021/acs.jctc.6b01122) with reasonable deviations for 11 out of 14 small molecules. For the remaining three, dithiin, furan and quinoxaline, the corresponding excitations could not be identified.


src/input_cp2k_properties_dft.F
```diff
@@ -1603,6 +1603,31 @@ SUBROUTINE create_tddfpt2_section(section)
...
       CALL section_release(subsection)
 
+      ! SOC functional
+
+      CALL section_create(subsection, __LOCATION__, name="SOC", &
+                          description="Is jet to be implemented", &
+                          n_keywords=2, n_subsections=0, repeats=.FALSE.)
+
+      CALL keyword_create(keyword, __LOCATION__, name="SINGLETT", &
+         description="Number of singlett states", &
```
mkrack: I think `singlet` and `triplet` is the correct spelling.

src/cp_control_types.F
```diff
@@ -512,6 +512,12 @@ MODULE cp_control_types
       !> user-defined reference point
       REAL(kind=dp), DIMENSION(:), POINTER :: dipole_ref_point
       !
+      ! SOC subsection
+      ! Number of Singlett states to be used
+      INTEGER                              :: soc_singlett
+      ! Number of triplett states to be used
+      INTEGER                              :: soc_triplett
+      !
```
mkrack, line 517: soc_singlet soc_triplet

src/input_cp2k_properties_dft.F
Tarandin: Thank you very much, you are absolutely right.  This error should be resolved now.

src/cp_control_types.F
Tarandin: Thank you. This error should be resolved now.

oschuett:
Shall we (squash) merge this?

PR #2861, opened 2023-07-14
GAPW excited-state forces

annahehn:
Last 2 intermediates for GAPW excited-state forces


tests/SLOW_TESTS_SUPPRESSIONS
```diff
@@ -83,5 +83,13 @@ QS/regtest-sccs-3/H2O_sccs_td_cd5_geo_opt.inp
 QS/regtest-sccs-2/H2O_sccs_otdiis_cd5.inp
 QS/regtest-lsroks/ch2o_rs.inp
 QS/regtest-rtp-5/H2O-wfn-mix-dens-pulse-1.inp
+QS/regtest-tddfpt-force-gapw/h2o_f01_coulomb_only.inp
```
oschuett: I'd really like to avoid adding new tests to this suppression list. Maybe you could try to speed them up a little bit? I saw that your using `EPS_DEFAULT 1.E-30` and `CUTOFF 800`. Or perhaps you could reduce `NSTATES`?

annahehn: Dear Ole, I fully agree - I will stick to it in the future. Thanks Ole!

PR #2862, opened 2023-07-14
Fix tests failing because of use_ref_cell

mkrack:
4 tests failed with GCC 13.1.0 and OpenMPI 4.1.x, because use_ref_cell was not set correctly. For these tests, the consideration of the reference cell has a significant impact on the applied cutoff causing large deviations in the total energy.

### DIFF
--- a/src/cell_methods.F
+++ b/src/cell_methods.F
@@ -271,7 +271,7 @@ CONTAINS
                                   check_for_ref, para_env)
 
       TYPE(cell_type), POINTER                           :: cell, cell_ref
-      LOGICAL, INTENT(OUT), OPTIONAL                     :: use_ref_cell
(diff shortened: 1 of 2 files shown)


oschuett:
Great find!

So, this was again an uninitialized `INTENT(OUT)` variable.

I wonder if we could have found this by running with `-finit-logical=true` ?

https://gcc.gnu.org/onlinedocs/gfortran/Code-Gen-Options.html#index-finit-local-zero

hfp:
I am not sure how this carried forward to `ifx`, but `ifort` used to have some fair set of warnings. In particular for INTENT (out/inout) including writing to induction variables (loop) using out/inout (which can unintentionally screw up loop analysis or performance). I think Intel Fortran's warnings (let alone the "recommendations") are not leveraged systematically. Perhaps it's worth doing so like `-Werror` for Matthias' Dashboard entries/systems?

PR #2866, opened 2023-07-17
Include H100 in toolchain

### LINKED ISSUE #2871: Missing H100 support in install toolchain

jobs-git:
Fix #2871


alazzaro:
Could you elaborate on this PR?
I assume GH100 is Hopper, for which we don't have kernels in DBCSR, so no sure it is just to introduce the flag in the toolchain enough

oschuett:
It's now corrected to `H100`, which has indeed [compute capability 9.0](https://developer.nvidia.com/cuda-gpus).
So, while this PR should not cause any damage, I'm not sure we if should merge it before the release.

PS: Please run `make pretty` to fix the formatting issues.

hfp:
> Could you elaborate on this PR? I assume GH100 is Hopper, for which we don't have kernels in DBCSR, so no sure it is just to introduce the flag in the toolchain enough

( Wrt DBCSR and the OpenCL backend, I contributed H100 tuned parameters in April this year. )

alazzaro:
For the sake of completeness, the instructions on how to add GPU kernels in DBCSR is at https://cp2k.github.io/dbcsr/develop/page/3-developer-guide/3-programming/2-accelerator-backend/2-libsmm_acc/index.html
Without kernels in DBCSR, all the execution will go to the CPU.

hfp:
> Without kernels in DBCSR, all the execution will go to the CPU.

Is this (now) true for the HIP/CUDA backend?

I was under the impression the code fails if a suitable kernel is missing. We perhaps want to cover a few such negative cases or expected failures. It should be sufficient requesting a kernel below threshold and with no coverage by the means of [tuned] parameters.

alazzaro:
> > Without kernels in DBCSR, all the execution will go to the CPU.
> 
> Is this (now) true for the HIP/CUDA backend?
> 
> I was under the impression the code fails if a suitable kernel is missing. We perhaps want to cover a few such negative cases or expected failures. It should be sufficient requesting a kernel below threshold and with no coverage by the means of [tuned] parameters.

For the time being there is no default kernel, therefore we still rely on the CPU (hope to introduce a default kernel soon).
What I don't know is if people ask for H100 at the build, for which no files are provided (so I can assume it will not compile at all).

hfp:
> For the time being there is no default kernel, therefore we still rely on the CPU (hope to introduce a default kernel soon).

I am not even sure about falling back to CPU. Of course, one of the return codes is supposed to trigger the CPU-fallback. However, I am not sure if we bailed out even before/earlier.

> What I don't know is if people ask for H100 at the build, for which no files are provided (so I can assume it will not compile at all).

At some point I had "A100 support" for the CUDA/HIP backend compiling V100-kernels using the new compute capability. It had two components: (1) DBCSR/CMake, and (2) integration with CP2K (`exts/build_dbcsr/Makefile`).

alazzaro:
> > For the time being there is no default kernel, therefore we still rely on the CPU (hope to introduce a default kernel soon).
> 
> I am not even sure about falling back to CPU. Of course, one of the return codes is supposed to trigger the CPU-fallback. However, I am not sure if we bailed out even before/earlier.

This is for sure and tested, otherwise it will require to generate all kernels, which is impossible.

oschuett:
Please run `make pretty`.

oschuett:
Gentle ping :-)

PR #2870, opened 2023-07-19
Update offload_runtime.h

mtaillefumier:
[hip] fix for issue 2869

### DIFF
--- a/src/offload/offload_runtime.h
+++ b/src/offload/offload_runtime.h
@@ -338,7 +338,7 @@ static inline void offloadEnsureMallocHeapSize(const size_t required_size) {
   if (current_size < required_size) {
     OFFLOAD_CHECK(cudaDeviceSetLimit(cudaLimitMallocHeapSize, required_size));
   }
-#elif defined(__OFFLOAD_HIP)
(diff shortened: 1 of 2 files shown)


src/offload/offload_runtime.h
```diff
@@ -338,7 +338,8 @@ static inline void offloadEnsureMallocHeapSize(const size_t required_size) {
   if (current_size < required_size) {
     OFFLOAD_CHECK(cudaDeviceSetLimit(cudaLimitMallocHeapSize, required_size));
   }
-#elif defined(__OFFLOAD_HIP)
+#elif defined(__OFFLOAD_HIP) && (HIP_VERSION_MAJOR >= 5) &&                    \
+    (HIP_VERSION_MINOR >= 3)
```
oschuett: I think you'll need to combine both numbers:
```
#elif defined(__OFFLOAD_HIP) && (100 *  HIP_VERSION_MAJOR + HIP_VERSION_MINOR >= 503)
```

mtaillefumier: Yes it is possible to do this. One way or the other is correct. I will fix it later today or tomorrow

PR #2878, opened 2023-07-25
Add apptainer images for znver2 and znver3

nilsholle:
As Zen 2 and Zen 3 nodes are more and more widely used, I would propose also adding Apptainer images targeted at those platforms.


mkrack:
Thanks. Did you actually check the containers built with these definition files on the respective systems?

PR #2882, opened 2023-07-31
Move all ScaLapack routines to the fm folder

juerghutter:


juerghutter:
This commit also fixes a bug in cp_fm_to_fm

PR #2896, opened 2023-08-04
Add more initializers

fstein93:


fstein93:
Currently, we cannot provide default initializers if the libraries do not provide suitable default initializers themselves (including DBCSR).

PR #2904, opened 2023-08-07
Fix dashboard

fstein93:


fstein93:
Apparently, GCC 8 cannot deal with default initializers of a derived type with allocatable components of a type without a default initializer. In that case, I do not add a default initializer to the respective type.

alazzaro:
I think the most relevant problem is Daint-GPU which has GCC 9.3. @mkrack can we upgrade GCC on Daint?

mkrack:
> I think the most relevant problem is Daint-GPU which has GCC 9.3. @mkrack can we upgrade GCC on Daint?

While GCC 11.2.0 works fine for CP2K on daint-mc, the software stack (cudatoolkit) provided for daint-gpu did not work with that GCC version when I tried it the last time. I will give a try, because my last test was quite some time ago.

fstein93:
Another problematic architecture is Fedora.

alazzaro:
There is a way to force CUDA to compile with new compiler via `-allow-unsupported-compiler`, but I really don't know how far you can go... It seems the GCC 8 test is OK now.

oschuett:
> Another problematic architecture is Fedora.

I'll upgrade the Fedora test from version 33 to 38.

alazzaro:
The current Fedora is already using GCC 10.
@fstein93 is this something only for DBCSR? Something we can fix via [this trick](https://stackoverflow.com/questions/52712245/gfortran-complains-about-rank-of-element-in-structure-constructor-for-allocatabl)?

fstein93:
I can try that one. A better solution is the addition of default initializers to all derived types of DBCSR.

alazzaro:
> I can try that one. A better solution is the addition of default initializers to all derived types of DBCSR.

So, is it only DBCSR which gives errors? If so, please open a issue in DBCSR and we will proceed for a new release...

oschuett:
> The current Fedora is already using GCC 10.


Fedora 33 is running GCC 10.3.1 (which is failing), while our [Ubuntu test](https://dashboard.cp2k.org/archive/gcc10/index.html) is running GCC 10.4.0 (which is passing).

After the [upgrade to Fedora 38](https://github.com/cp2k/cp2k/pull/2907) it will be using GCC 13.2.1, which is working.

alazzaro:
> > The current Fedora is already using GCC 10.
> 
> Fedora 33 is running GCC 10.3.1 (which is failing), while our [Ubuntu test](https://dashboard.cp2k.org/archive/gcc10/index.html) is running GCC 10.4.0 (which is passing).
> 
> After the upgrade to Fedora 38 it will be using GCC 13.2.1, which is working.

Still, this is a weak solution, we are chasing for working compiler versions...

oschuett:
> Still, this is a weak solution, we are chasing for working compiler versions...

I agree. It's really concerning that this only works with certain minor versions. We need a more robust solution.

I'll upgrade Fedora nevertheless because I was planning to it anyways.

oschuett:
> I'll upgrade Fedora nevertheless because I was planning to it anyways.

On second thought, I'll hold off on merging #2907 until we find a good solution so we can continue testing GCC 10.3.1.

fstein93:
> > I can try that one. A better solution is the addition of default initializers to all derived types of DBCSR.
> 
> So, is it only DBCSR which gives errors? If so, please open a issue in DBCSR and we will proceed for a new release...

AFAICT, it is an issue related to default initializers of derived types with allocatable components. I do not have these issues if a component is a standard-type in Fortran.

I think DBCSR should provide default initializers for its types. In that way, you ensure that all types are somewhat initialized preventing certain types of bugs. I will open an issue.

alazzaro:
> > > I can try that one. A better solution is the addition of default initializers to all derived types of DBCSR.
> > 
> > 
> > So, is it only DBCSR which gives errors? If so, please open a issue in DBCSR and we will proceed for a new release...
> 
> AFAICT, it is an issue related to default initializers of derived types with allocatable components. I do not have these issues if a component is a standard-type in Fortran.
> 
> I think DBCSR should provide default initializers for its types. In that way, you ensure that all types are somewhat initialized preventing certain types of bugs. I will open an issue.

OK, thanks. Fedora is still failing, so we can use that test to check with a new DBCSR (and maybe the Daint-GPU too).
Then I propose to merge this PR. ETA for a new RC DBCSR will be mid-September (plan to introduce MPI G2G too).

PR #2905, opened 2023-08-07
Bump ELPA to version 2023.05.001

fstein93:


fstein93:
Could any of the admins at the new [tarball](https://elpa.mpcdf.mpg.de/software/tarball-archive/Releases/2023.05.001/elpa-2023.05.001.tar.gz)?

PR #2906, opened 2023-08-08
Energy decomposition analysis

juerghutter:


fstein93:
Your tests make use of HF (PBE0). Please add libint as a dependence of the respective directory.

oschuett:
The new `QS/regtest-ed/si8_pmix.inp` is often flagged as slow by the [pdbg tester](https://dashboard.cp2k.org/archive/current-pdbg/index_full.html). Maybe you could try to speed it up?

PR #2915, opened 2023-08-10
Cross-links between keyword list and descriptions added

mkrack:
- Modified CSS theme for admonition class note added
- Display of keyword list entries with underscores fixed
- Typo in variable name and for aliases fixed
- Use year range in copyright

### DIFF
--- a/docs/generate_input_reference.py
+++ b/docs/generate_input_reference.py
@@ -147,7 +147,8 @@ def process_section(
         # Render TOC for keywords
         output += ["## Keywords", ""]
         for keyword in keywords:
-            output += [f"* {get_name(keyword)}"]  # TODO cross-links with description
+            keyword_name = get_name(keyword)
+            output += [f"* <a href='#{keyword_name}'>{keyword_name}</a>"]
         output += [""]
         # Render keywords
         output += ["## Keyword descriptions", ""]


oschuett:
I'm not sure about this change. Replacing the `py:data` directive with a generic `admonition` moves us very far away from how Sphinx is supposed to be used. The fact that you had to fork the CSS style sheet is a testament to this.

What made you decide against `py:data` ?

mkrack:
I simply did not find a way to accomplish the cross-linking with `py:data` but with `admonition`. What's wrong in using the `admonition` class [note](https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/kitchen-sink/admonitions.html#note)? I modified the CSS style sheet slightly (come colors) for better readability. If one is happy with the default style like for `py:data`  (not sure myself), things are certainly easier.

oschuett:
> I simply did not find a way to accomplish the cross-linking with `py:data` but with `admonition`. 

Ok, let me take a stab at it.

> What's wrong in using the admonition class [note](https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/kitchen-sink/admonitions.html#note)?

Sphinx is based on the idea of logical text formatting. For this purpose it comes with a large libraries of [Domains](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html). The advantage of this approach is that one gets automatic cross-linking (between documents) and index pages.

While a `py:data` is not exactly a CP2K keyword, it's pretty close. If we really can't make it work, then I think the better way forward would be to implement our own Domain.


>  I modified the CSS style sheet slightly (come colors) for better readability.

The problem with forking the style sheet is that we'll have to maintain it as the [upstream theme](https://github.com/readthedocs/sphinx_rtd_theme) evolves.

Which colors did you change? Maybe we can include only those rules that you want to overwrite?

I think we should not stray too far though because otherwise we'll lose the familiar _Read the Docs_  look.

mkrack:
> Which colors did you change? Maybe we can include only those rules that you want to overwrite?

I used uppercase letters for the modified color codes, see line 4, columns 50083, 54333 and 55722.
I changed also the icon character in column 10538 to `0xF05A` in the same line.

oschuett:
What do you think about #2918?

mkrack:
That's just the first step of #2915 linking but not cross-linking keyword list and description. I do not insist on cross-linking as in the old documentation, but #2918 ignores also some further fixes (escape_markdown of link text, typos).

PR #2921, opened 2023-08-11
Docs: Fix cross-referencing

oschuett:
This change allows us to link to keywords from anywhere within our Sphinx docs and have them checked at build time.

Cross-references can be inserted like this: `[](MOTION.TMC.PRESSURE)` or like this: `[PRESSURE](MOTION.TMC.PRESSURE)`.

Thanks @mkrack for realizing that we can just use simple HTML anchors to preserve old links!

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -20,4 +20,4 @@ To build  a local version of the manual perform the following steps:
 
    `make html`
 
-1. Browse the HTML output in the `_build` directory.
+1. Browse the HTML output in the `_build/html` directory.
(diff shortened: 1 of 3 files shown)


mkrack:
@oschuett The effect of the `<a id="...">...</a>` tags is not obvious to me.

oschuett:
> The effect of the `<a id="...">...</a>` tags is not obvious to me.

They ensures that old links continue to work, like e.g. https://manual.cp2k.org/trunk/CP2K_INPUT/GLOBAL.html#desc_RUN_TYPE from [this wiki page](https://www.cp2k.org/howto:static_calculation).

PR #2924, opened 2023-08-14
Long-range quantum computing (WF) in short-range DFT embedding

stefabat:
This PR revamps the active space module and implements an embedding approach based on range-separated DFT.

An active space Hamiltonian describing a few electrons in a few orbitals embedded in the field of all remaining electrons is solved iteratively with external software.
The current implementation is interfaced to qiskit-nature, which enables the solution of the active space problem on classical and quantum hardware alike.


oschuett:
It would be great to have an integration test for qiskit-nature. Maybe you could draft something akin to [test_i-pi.sh ](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_i-pi.sh) or share an example input?

fstein93:
This PR introduced a compilation issue. The moduel hdf5_wrapper has to check for `__HDF5` (check cp2k_info for reference) and wrap the respective use statements and function bodies to prevent compilation errors if CP2K is not compiled with HDF5.

stefabat:
> It would be great to have an integration test for qiskit-nature. Maybe you could draft something akin to [test_i-pi.sh ](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_i-pi.sh) or share an example input?

Yes, I can do that. I have another repo with many test examples and I can easily pull one or a few of them out.

stefabat:
> This PR introduced a compilation issue. The moduel hdf5_wrapper has to check for `__HDF5` (check cp2k_info for reference) and wrap the respective use statements and function bodies to prevent compilation errors if CP2K is not compiled with HDF5.

You are right, I somehow completely forgot about hdf5_wrapper

stefabat:
> > It would be great to have an integration test for qiskit-nature. Maybe you could draft something akin to [test_i-pi.sh ](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_i-pi.sh) or share an example input?
> 
> Yes, I can do that. I have another repo with many test examples and I can easily pull one or a few of them out.

Actually, I first have to discuss with @mrossinek on how to do this, so it might take a few days before we have something similar in place.

PR #2930, opened 2023-08-17
Add docker files for CP2K production containers

mkrack:


oschuett:
We already have a couple of [production Dockerfiles](https://github.com/search?q=repo%3Acp2k%2Fcp2k+path%3A**%2FDockerfile.prod*&type=code). Should we remove them then to avoid confusion? 

Also, please add the new Python script to [test_misc.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_misc.sh).

mkrack:
@oschuett My proposal is to separate the docker files in the folder `tools/docker` into two sub-folders `testing` and `production`. The `testing` folder would collect the files primarily used for the dashboard and other tests. The current set of new two-stage production docker files should eventually replace the `prod_*` files. The current file set is, however, only a subset of the existing `prod_*` files, because the files for GPU systems are missing. I will continue there.

Do we need `prod_[sp]dbg` docker files?

Thanks for the hint, I will add the script to `test_misc.sh`. Is there any other interference? I hope that the CI workflows and the dashboard won't be affected by the new folder in docker.

Currently, I still push the docker images to `mkrack/cp2k` to not mess up `cp2k/cp2k` on Docker Hub. I am open to discuss the naming convention and how things will be organized on Docker Hub (maybe during tomorrow's meeting).

oschuett:
@mkrack This is great! There is certainly still a lot that could be improve on the old prod Dockerfiles.

>  The current set of new two-stage production docker files should eventually replace the prod_* files.

Separating the Dockerfiles for prod and testing means that the prod files receive less testing, but in return we get much smaller images due to the two-stage approach. I think that's a fair trade off. Furthermore, keeping those two use-cases apart will simplify maintenance.

> Do we need `prod_[sp]dbg` docker files?

Probably not.

>  I hope that the CI workflows and the dashboard won't be affected by the new folder in docker.

No worries. The CI only looks at the files linked in the [cp2k-ci.conf](https://github.com/cp2k/cp2k/blob/master/tools/docker/cp2k-ci.conf).

> Currently, I still push the docker images to mkrack/cp2k to not mess up cp2k/cp2k on Docker Hub. 

I'm happy to hand over the cp2k/cp2k repo on Docker Hub.  Let's discuss tomorrow.

PR #2931, opened 2023-08-17
[cmake] Add cmake module for finding the cusolver library

mtaillefumier:
Updated the cmake build system such that it can detect the cusolvermp library when the package is downloaded separately.


oschuett:
Please run `make pretty`.

PR #2936, opened 2023-08-22
Switch to latest oneAPI version, revise compiler command

mkrack:


mkrack:
The newest oneAPI HPC Toolkit version shows the same failures. Let's do this update anyway.

PR #2939, opened 2023-08-23
Use routines in pw_methods for some standard grid operations

fstein93:
This PR prepares later improvement steps for the OpenMP parallelization.

EDIT: This PR enforces the developer to provide proper spaces and data types when a PW grid is created. In that regard, there are also some inconsistencies within the PW methods.

### DIFF
--- a/src/pw/pw_methods.F
+++ b/src/pw/pw_methods.F
@@ -2039,15 +2039,17 @@ CONTAINS
 !>      JGH (14-Mar-2001) : Parallel sum and some tests, HALFSPACE case
 !> \author apsi
 ! **************************************************************************************************
-   FUNCTION pw_integral_ab(pw1, pw2, sumtype) RESULT(integral_value)
+   FUNCTION pw_integral_ab(pw1, pw2, sumtype, just_sum) RESULT(integral_value)
 
       TYPE(pw_type), INTENT(IN)                          :: pw1, pw2
       INTEGER, INTENT(IN), OPTIONAL                      :: sumtype
       REAL(KIND=dp)                                      :: integral_value
+      LOGICAL, OPTIONAL, INTENT(IN)                      :: just_sum
 
       CHARACTER(len=*), PARAMETER                        :: routineN = 'pw_integral_ab'
 
       INTEGER                                            :: handle, loc_sumtype
+      LOGICAL                                            :: my_just_sum
 
       CALL timeset(routineN, handle)
 
@@ -2058,6 +2060,10 @@ CONTAINS
          CPABORT("Grids incompatible")
       END IF
 
+      my_just_sum = .FALSE.
+      IF (PRESENT(just_sum)) my_just_sum = just_sum
+
+      ! do standard sum
       IF (loc_sumtype == do_standard_sum) THEN
 
          ! Do standard sum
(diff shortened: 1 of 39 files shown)


src/pw/pw_methods.F
```diff
@@ -2661,4 +2685,54 @@ SUBROUTINE pw_set_value(pw, value)
 
    END SUBROUTINE pw_set_value
 
+! **************************************************************************************************
+!> \brief ...
+!> \param pw1 ...
+!> \param pw2 ...
+! **************************************************************************************************
+   SUBROUTINE pw_multiply_with(pw1, pw2)
+      TYPE(pw_type), INTENT(IN)                          :: pw1, pw2
```
mkrack: I am wondering, if the INTENT(IN) attribute for `pw1` is appropriate here, because it appears on the lhs in this routine.

fstein93: That's one of the confusing aspects of the intent in Fortran. The intent of a pointer variable refers only to its association state, not its data. Speaking in C, a dummy variable of `INTEGER(C_INT), POINTER, INTENT(IN)` is equivalent to a C-variable of type `int const *`, not `const int *` (or `const int const *`). Thus, the intent used here is correct because only the data is affected, not its association state. But for clarity, I will change the intent of pw1 to inout.

mkrack: Thanks for considering my comment. Now, I am confused, because the dummy argument `pw1` has no `POINTER` attribute. I haven't studied the Fortran standard but only a common Fortran textbook which states:
Screenshot 2023-08-23 at 16 01 19
Well, no clue how Fortran compiler handle this eventually, but I am not sure, if we can rely on any analogy to C here.

mkrack:
@fstein93 In this context, I would like to rise another point concerning the combination of the `POINTER` attribute with an `INTENT` attribute. Following the same Fortran textbook an `INTENT` attribute should not appear together with a `POINTER` attribute for a dummy argument:

Screenshot 2023-08-23 at 16 08 23

I see, however, that a combination of these attributes is used in many places in the CP2K code. What's your opinion?

fstein93:
That used to be an issue with Fortran 95 and IIRC was fixed with Fortran 2003. In general, the intent of a pointer may lead to issues in case of `INTENT(OUT)` for two reasons:
- The association state of the pointer is undefined on entrance which is a problem for the caller if said state stays unchanged within the routine. Then, the behavior is compiler dependent if the pointer is accessed.
- The association state of the pointer is lost leading to accidental memory leaks if the target is not freed elsewhere (target is not a pointer or there is a different pointer to the same target).
`INTENT(IN/INOUT)` are safe in that respect. I never use `INTENT(OUT)` with pointers directly. As a component of a derived type, I am fine with its application to the respective derived type dummy variable if the purpose of the routine is clearly specified within its routine name (creation routines). Then, it is advantageous to define proper default initializers nullifying the pointer components automatically.

mkrack:
@fstein93 Thanks for the clarification.

fstein93:
I removed the constants NOSPACE and NODATA from the public interface of the pw_types module. This enforces the choice of proper data types and a proper space of a newly created PW grid.

PR #2943, opened 2023-08-23
Fix for Intel compiler (suppress OpenMP for loop)

mkrack:


fstein93:
Thank you for the fix. It is interesting, how conservative and how aggressive the intel compiler is regarding automatic parallelization.

PR #2964, opened 2023-09-01
Fix the error that the allocatable attribute globenv%gaussian_rng_stream is not allocated

wangtianxiang:
When compiling cp2k with compile options **_-fno-realloc-lhs_**  from the [cp2k-debugging](https://www.cp2k.org/dev:debugging)

```shell
FCFLAGS  = -O1 -fstrict-aliasing -g -fno-omit-frame-pointer -fno-realloc-lhs \
           -fcheck=bounds,do,recursion,pointer -ffree-form $(DFLAGS)
```

An error occurred when executing cp2k:
1693553744293_3763A21B-FFDE-493a-921D-AAE70A6C4084

The globenv%gaussian_rng_stream has an ALLOCATABLE  attribue
```Fortran
   TYPE global_environment_type
      INTEGER :: ref_count = 0
      TYPE(rng_stream_type), ALLOCATABLE      :: gaussian_rng_stream
      CHARACTER(LEN=default_string_length)    :: diag_library = "ScaLAPACK"
```

The globenv%gaussian_rng_stream should be allocated in subroutine globenv_create  
```Fortran
   SUBROUTINE globenv_create(globenv)
      TYPE(global_environment_type), POINTER             :: globenv

      CPASSERT(.NOT. ASSOCIATED(globenv))
      ALLOCATE (globenv)
      ALLOCATE (globenv%gaussian_rng_stream)
      globenv%ref_count = 1
      globenv%cp2k_start_time = m_walltime()

   END SUBROUTINE globenv_create
```

### DIFF
--- a/src/qs_active_space_utils.F
+++ b/src/qs_active_space_utils.F
@@ -124,8 +124,15 @@ CONTAINS
                DO i34l = 1, eri%nzerow_local(i12l)
                   i34 = eri%colind_local(irptr + i34l)
                   CALL csr_idx_from_combined(i34, nmo_max, i3, i4)
+! The FINDLOC intrinsic function of the Fortran 2008 standard is only available since GCC 9
+! That is why we use a custom-made implementation of this function for this compiler
+#if __GNUC__ < 9
+                  k = cp_findloc(active_orbitals(:, spin2), i3)
+                  l = cp_findloc(active_orbitals(:, spin2), i4)
+#else
                   k = FINDLOC(active_orbitals(:, spin2), i3, dim=1)
                   l = FINDLOC(active_orbitals(:, spin2), i4, dim=1)
+#endif
                   erival = eri%nzval_local%r_dp(irptr + i34l)
 
                   ! 8-fold permutational symmetry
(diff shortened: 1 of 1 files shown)


oschuett:
It seems that wiki page is a bit out of date. I believe `-fno-realloc-lhs` is equivalent to [`--Wrealloc-lhs-all`](https://gcc.gnu.org/onlinedocs/gfortran/Error-and-Warning-Options.html#index-Wrealloc-lhs-all). In the meantime we've switched to the more permissive [`--Wrealloc-lhs`](https://gcc.gnu.org/onlinedocs/gfortran/Error-and-Warning-Options.html#index-Wrealloc-lhs) for our [debug builds](https://github.com/cp2k/cp2k/blob/b75a7836de90d25328554d461b26e69ac0402ae4/tools/toolchain/scripts/generate_arch_files.sh#L60).

wangtianxiang:
`-Werror=realloc-lhs` only checks array. I think `-Werror=realloc-lhs-all` should be used instead of `-Werror=realloc-lhs`.
In the case, Why not `globenv%gaussian_rng_stream`  explicitly?

Explicit is better than implicit.

Also, I don't understand why the automatic test fails after just adding this line?

mkrack:
> Also, I don't understand why the automatic test fails after just adding this line?

The additional tests failed because of other issues which have to fixed.

mkrack:
@wangtianxiang Try to rebase and push again your change.

PR #2974, opened 2023-09-07
Fix forces output with REFTRAJ

mkrack:


oschuett:
I'm not sure about this PR. Now there seems to be no way to only evaluate the energy without the forces.

I'm afraid this will remain confusing until we change the input structure. We can retain backwards compatibility by ~~marking the old keywords as removed~~ adding a `deprecation_notice`. I'll take a stab at this.

PR #2979, opened 2023-09-08
Fix conventions

mkrack:


Anyborr:
This update broke compilation (tested using arch file for CRAY-XC50) :

> /users/anybergb/DEV/cp2k/src/mp2_types.F:15:7:
>
>   15 |    USE cp_eri_mme_interface,            ONLY: cp_eri_mme_finalize,&
>      |       1
> Error: The rank of the element in the structure constructor at (1) does not match that of the component (0/1)
> /users/anybergb/DEV/cp2k/src/mp2_types.F:15:7:
>
>   15 |    USE cp_eri_mme_interface,            ONLY: cp_eri_mme_finalize,&
>      |       1
> Error: The rank of the element in the structure constructor at (1) does not match that of the component (0/1)
> make[3]: *** [/users/anybergb/DEV/cp2k/Makefile:514: mp2_types.o] Error 1
> make[3]: *** Waiting for unfinished jobs....
> make[2]: *** [/users/anybergb/DEV/cp2k/Makefile:146: all] Error 2
> make[1]: *** [/users/anybergb/DEV/cp2k/Makefile:128: psmp] Error 2
> make: *** [Makefile:123: all] Error 2

mkrack:
@Anyborr Correct. Piz Daint is almost EOL when the next CP2K version will be released.

alazzaro:
This error is the same initialization error that @fstein93 fixed in other places. Maybe Frederick can take a look?

fstein93:
I have a fix which is just partially rewinding this PR (see #2988), otherwise the code would blow up for these types (see below). It is related to allocatable components of the other derived types in mp2_types.F, not to cp_eri_mme_types and requires to set all allocatable components explicitly to NULL() while calling the initializer.

PR #2987, opened 2023-09-12
Separate PW types

fstein93:
This PR incorporates rank, type and space of the grid into the pw_type. The naming scheme of the new types is "pw_<r/c for real/complex array><1D/3D depending on rank>_<rs/gs for realspace/reciprocal space>_type. This is justified by rank, type and space being fixed during the lifetime of the PW grids.
Code duplication is prevented as much as possible by letting Fypp write the necessary code starting from templates.

Advantages:
- These changes allow the promotion of several runtime checks regarding correct space and data types to compile time checks.
- This PR fixes two minor bugs in the code which were apparently never hit by users.
- Test coverage of the respective module is improved because of the reuse of code fragments.

This PR requires the creation of several copies of the pw_poisson_solve function because function interfaces must be distinguishable if optional arguments are not set. In exchange, I made some simplifications to the different copies.

The pw_copy function now covers all transfers between grids in the same space although not all of them are implemented (and will probably never be used). The old pw_transfer function is dropped in favor of an interface binding the same functions it used to call (pw_copy, pw_gather, pw_scatter, fft_wrap_pw1pw2).

Left for further PRs:
- Further refactoring of common bits used on several occasions (prepares improvements regarding OpenMP)
- Introduction of real2complex and complex2real FFT capabilities of FFTW to save FLOPs
- Acceleration of certain routines (especially scatter and gather operations)
- Switch to pw_types in XC and CDFT modules to prepare OpenMP


fstein93:
I will try to fix the initialization issue first to be able to have a test on CUDA because I made some adjustments to that part too. Unfortunately, my operating system does not support GCC8 anymore.

fstein93:
@oschuett Could you provide the full output or at least the cause of the failing compilation of the CUDA regtest?

oschuett:
> @oschuett Could you provide the full output or at least the cause of the failing compilation of the CUDA regtest?

I'm sorry. I forgot about this yesterday and right now I don't have time.
One thing you could do is to temporarily change [this line](https://github.com/cp2k/cp2k/blob/7e52d8fe64174c63d4297b1f87b53e7f026adc7a/tools/docker/scripts/test_regtest.sh#L59) to `cat make.out` as part of this PR.

fstein93:
Unfortunately, I have no way to test it locally, so I am attempting to bisect here.

oschuett:
You can run all the CI tests locally [using Docker](https://github.com/cp2k/cp2k/blob/master/tools/docker/README.md#test-images): 

```
cd cp2k/tools/docker
docker build --shm-size=1g -f Dockerfile.test_sdbg -t cp2k_test_sdbg ../../
```

fstein93:
@oschuett Thanks for the pointer to docker. I need a bit of time to get into it because I never used it and I am currently working with Ubuntu@WSL2 which brings its own issues (RAM, hard drive memory etc.). Tomorrow or next week, I will switch to a Linux-based notebook for work. I could get the sdbg running within docker.

fstein93:
I will close the PR for now and reopen it later after I am done fixing the relevant builds.

PR #2988, opened 2023-09-13
Fix initialization issue with GCC8 and GCC9.3

fstein93:
This PR partially rewinds #2979 to fix issues with GCC8 and GCC 9.3 (before 9.5, I do no whether 9.4 raises the same issue). As soon as we drop GCC8 and GCC9, we can remove the workaround.

### DIFF
--- a/src/cp_eri_mme_interface.F
+++ b/src/cp_eri_mme_interface.F
@@ -79,7 +79,11 @@ MODULE cp_eri_mme_interface
    TYPE cp_eri_mme_param
       TYPE(cp_logger_type), POINTER                      :: logger => NULL()
       ! There is a bug with some older compilers preventing a default initialization of  derived types with allocatable components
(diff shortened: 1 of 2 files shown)


mkrack:
Thanks @fstein93. I think `ri_rpa_im_time` and `ri_opt_param` can be initialised also with GCC <9.5.

fstein93:
As a workaround, I am initializing allocatable components explicitly now. The shorter versions without explicit initialization are the default for more recent compilers.

oschuett:
I believe what  trips up these older GCC versions are member fields that are **allocatable arrays of derived types**. Fortunately, this is a pretty rare combination. So, if the problem reappears then we could also stick to pointers for the next ~2 years until we can retire GCC 9.

fstein93:
The workaround is to initialize the allocatable components explicitly with a null pointer

PR #2990, opened 2023-09-14
compilation

annahehn:
compilation otherwise aborts on Caucluster Kiel


mkrack:
Which compiler version do you use on the Caucluster?

oschuett:
Related: https://www.cp2k.org/conv#c008

PR #2991, opened 2023-09-14
OpenBLAS 0.3.23 -> 0.3.24

mkrack:


dev-zero:
I suggest to follow distro packaging and raise the number of max threads from 64 to 128.
Unfortunately, the dynamical allocation of additional structures in OpenBLAS used when the number of threads is larger than the number of threads it was built for is still not race-free.

mkrack:
@dev-zero Thanks for the hint. I give it a try.

Issue #2913, opened 2023-08-10
Feature Request: Hybrid Functional (HSE06, PBE0, etc.) with k-points.


flwang1998:
Dear developers,
Is there any news about Hybrid Functional (HSE06, PBE0, etc.) with k-points in QS module?

flwang1998:
Dear developers,  
Is there any news about Hybrid Functional (HSE06, PBE0, etc.) with k-points in QS module?

abussy:
Hi,
it's currently in development, it should come online within the next couple of months.
Cheers

flwang1998:
Exciting with cp2k to make atoms dance!
Cheers

PR #2998, opened 2023-09-18
RI-HFX for K-points (with gradients and ADMM)

abussy:
This PR introduces a new method for RI-HFX with K-point sampling. The exact exchange matrix is calculated in real-space for all interacting periodic images, before it is Fourier transformed to reciprocal space. Then, the complex Hermitian KS matrix is built, and the eigenvalue problem solved for each K-point. Details of the method will be published in a paper in the near future.

The method relies on the DBT library for sparse-tensor algebra, and on the libint library for analytical 3- and 2-center ERI evaluation. Forces and stress tensor calculations are also implemented. The method can be sped-up with the ADMM approximation (ADMM2, ADMMP, ADMMQ and ADMMS).

Careful testing was conducted. Equivalence to Gamma-point supercell calculations was verified, as well as convergence with the number of k-points. DEBUG runs were systematically ran for forces and stress tensors. Four new regtest directories were added, covering KP-RI-HFX energy and forces, with and without ADMM.

@dev-zero was instrumental in helping develop the ADMM approximation with k-point sampling.

### DIFF
--- a/src/qs_scf_initialization.F
+++ b/src/qs_scf_initialization.F
@@ -630,8 +630,6 @@ CONTAINS
             ! possible for OT gamma point calculations
             qs_env%requires_mo_derivs = .FALSE.
          END IF
-         IF (dft_control%do_admm) &
-            CPABORT("No ADMM implemented with kpoints")
          IF (dft_control%do_xas_calculation) &
             CPABORT("No XAS implemented with kpoints")
          DO ik = 1, SIZE(kpoints%kp_env)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -220,6 +220,7 @@ list(
   hfx_pair_list_methods.F
   hfx_pw_methods.F
   hfx_ri.F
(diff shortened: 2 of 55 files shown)


src/qs_scf_initialization.F
```diff
@@ -630,8 +630,8 @@ SUBROUTINE qs_scf_ensure_mos(qs_env)
             ! possible for OT gamma point calculations
             qs_env%requires_mo_derivs = .FALSE.
          END IF
-         IF (dft_control%do_admm) &
-            CPABORT("No ADMM implemented with kpoints")
+         !IF (dft_control%do_admm) &
+         !   CPABORT("No ADMM implemented with kpoints")
```
dev-zero: maybe drop completely now :-)

fstein93:
Closes #2913

dev-zero:
@fstein93 fyi, the canonical HFXk reference implementation will come with a second PR from my side

PR #2984, opened 2023-09-11
Honor initial CMAKE_Fortran_FLAGS/FFLAGS

opoplawski:

PR #3000, opened 2023-09-20
[cmake] More modern ways to use cmake (partially)

mtaillefumier:
- Use message("" "") when multiple lines are printed instead of calling message() for each line.

- Use CMakePresents.json for defining default configurations. The options are identical to the old settings. To use them simply input 'cmake --preset MINIMAL,FULL,DEFAULT,SERIAL'. The user gets the default behavior (full control) if '--preset' is not set.

- Minor cleanups, alias definitions etc....

- Add dockerfiles (CPU, cuda, rocm) to create docker images of cp2k with spack. (openmpi is supported, adding mpich is trivial)

- move the compiler options to a json file so we can configure any compiler supported by cmake independently. Compiler versioning remain to be implemented.


mtaillefumier:
@oschuett and @RMeli i want your opinion on this. I moved the compilers options in  a json to ease configuration. It is fairly basic but works so far.
We could maybe improve it a little. Any suggestion ?

dev-zero:
> @oschuett and @RMeli i want your opinion on this. I moved the compilers options in a json to ease configuration. It is fairly basic but works so far. We could maybe improve it a little. Any suggestion ?

What is the rational behind replacing one markup language with another one?
In particular since the JSON-variant requires glue code in CMake (all the variables) and still is not as expressive as the former CMake-original (e.g. the Cray-specific branches to add `-h system_alloc`) got dropped?

mtaillefumier:
I really do not see what issue json has. All options are clearly presented in the json file and the three lines code of cmake code to read the json file are not complicated to read at all.

it is easier to edit a json file than having to go to cascaded `if then else` statements. It is expressive but difficult to follow in general and not maintainable when you have more than few compilers. I am all in for expressiveness but I find this 

```
"GNU": {
            "CMAKE_Fortran_FLAGS": "-mtune=native -ffree-line-length-512 -ffree-form -std=f2008 -fimplicit-none -Werror=aliasing -Werror=ampersand -Werror=c-binding-type -Werror=intrinsic-shadow -Werror=intrinsics-std -Werror=tabs -Werror=target-lifetime -Werror=underflow -Werror=unused-but-set-variable -Werror=unused-variable  -Werror=conversion -Werror=zerotrip -Wuninitialized -Wno-maybe-uninitialized -Wunused-parameter",
            "CMAKE_Fortran_FLAGS_RELEASE": "-O3 -g -funroll-loops",
            "CMAKE_Fortran_FLAGS_COVERAGE": "-O0 -g --coverage -fno-omit-frame-pointer -fcheck=all -ffpe-trap=invalid,zero,overflow -fbacktrace -finit-real=snan -finit-integer=-42 -finit-derived -Werror=realloc-lhs -finline-matmul-limit=0",
            "CMAKE_Fortran_FLAGS_DEBUG": "-fbacktrace -ffree-form -fimplicit-none -std=f2008 -fsanitize=leak -fcheck=all,no-array-temps -ffpe-trap=invalid,zero,overflow -finit-derived -finit-real=snan -finit-integer=-42 -Werror=realloc-lhs -finline-matmul-limit=0",
            "CMAKE_C_FLAGS_RELEASE": "-O3 -g -funroll-loops -Wall",
            "CMAKE_C_FLAGS_COVERAGE": "-O0 -g --coverage -Wall",
            "CMAKE_C_FLAGS_DEBUG": "-O2 -ggdb -Wall -fsanitize=undefined -fsanitize=address -fsanitize-recover=all -Wall -Wextra -Werror -Wno-vla-parameter -Wno-deprecated-declarations "
        }
```
clearer than 
```

if(CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
  set(CMAKE_Fortran_FLAGS
      "${CMAKE_Fortran_FLAGS} -mtune=native -ffree-line-length-512 -ffree-form -std=f2008 -fimplicit-none -Werror=aliasing -Werror=ampersand -Werror=c-binding-type -Werror=intrinsic-shadow -Werror=intrinsics-std -Werror=tabs -Werror=target-lifetime -Werror=underflow -Werror=unused-but-set-variable -Werror=unused-variable  -Werror=conversion -Werror=zerotrip -Wuninitialized -Wno-maybe-uninitialized -Wunused-parameter"
  )

  if(CMAKE_Fortran_COMPILER_VERSION VERSION_GREATER 10)
    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fallow-argument-mismatch"
    )# gcc 10+ has this automatically
  endif()

  set(CMAKE_Fortran_FLAGS_RELEASE "-O3 -g -funroll-loops")
  set(CMAKE_Fortran_FLAGS_COVERAGE
      "-O0 -g --coverage -fno-omit-frame-pointer -fcheck=all -ffpe-trap=invalid,zero,overflow -fbacktrace -finit-real=snan -finit-integer=-42 -finit-derived -Werror=realloc-lhs -finline-matmul-limit=0"
  )
  set(CMAKE_Fortran_FLAGS_DEBUG
      "-fbacktrace -ffree-form -fimplicit-none -std=f2008 -fsanitize=leak -fcheck=all,no-array-temps -ffpe-trap=invalid,zero,overflow -finit-derived -finit-real=snan -finit-integer=-42 -Werror=realloc-lhs -finline-matmul-limit=0"
  )
  if((NOT (USE_MPI)) OR (NOT ("${MPI_Fortran_LIBRARY_VERSION_STRING}" MATCHES
                              "Open MPI")))
    set(CMAKE_Fortran_FLAGS_COVERAGE
        "${CMAKE_Fortran_FLAGS_COVERAGE} -fsanitize=leak")
    set(CMAKE_Fortran_FLAGS_DEBUG
        "${CMAKE_Fortran_FLAGS_DEBUG} -fsanitize=leak")
  endif()
elseif(CMAKE_Fortran_COMPILER_ID STREQUAL "Intel")
```
although it is not the worst code ever. I would have use a case statement to make it cleaner if it existed. Apparently it does not.

json or yaml are dictionaries and I use them as such. Unfortunately cmake does not support compiler presets without been tedious. 

Including a new compiler is equivalent to adding a section in the json file is cleaner. I never really liked the initial setup but I kept it because I did not want to tackle too many problems at the same time.  

for nvhpc for instance (the options are bogus) : it is just an example

```
"nvhpc" : {
            "CMAKE_Fortran_FLAGS": "-Mfreeform -Mextend -Mallocatable=03",
            "CMAKE_Fortran_FLAGS_RELEASE": "-fast",
            "CMAKE_Fortran_FLAGS_DEBUG": "-g",
            "CMAKE_C_FLAGS_RELEASE": "-fast",
            "CMAKE_C_FLAGS_DEBUG": "-g"
}
```

We can even include a mechanism with conditional just as version number etc.... I did not do it but it is feasible.

Cray specific compilers can be put back easily although I do not even know if the current cray environment supports it. These options should probably be set outside though.

dev-zero:
> I really do not see what issue json has. All options are clearly presented in the json file and the three lines code of cmake code to read the json file are not complicated to read at all.

* It does not implement the same behaviour: now suddenly all variables are additive, not only `CMAKE_Fortran_FLAGS`. That was intended behaviour in the original in DBCSR to make it possible to easily override RELEASE/DEBUG/PROFILING flags completely.
* JSON does not permit comments which are sometimes used to describe why something was added (it is thanks to those that you know that you can likely drop some of the configurations as they may not be relevant anymore)
* The idea in DBCSR was to have working configurations for the most common compilers the project is being used on (extending on CMake's capabilities like C++ standard requirement flags)
* Since CMake naturally has branching constructs, we could also model more complex scenarios and make them available to others, disentangle *what is being built* from *how it is to be built*.
* lines 8-33 in the changeset, not 3. And this does not cover all the branching and cases in the current code nor version-dependent configurations.

oschuett:
I think replacing CMake code with JSON is almost always a win:

1. It's declarative.
2. It raises the level of abstraction.
3. Since it's easy to parse, we can build a bunch of sanity and consistency checks for it.
5. We can generate Sphinx documentation from it.

mtaillefumier:
> > I really do not see what issue json has. All options are clearly presented in the json file and the three lines code of cmake code to read the json file are not complicated to read at all.
> 
>     * It does not implement the same behaviour: now suddenly all variables are additive, not only `CMAKE_Fortran_FLAGS`. That was intended behaviour in the original in DBCSR to make it possible to easily override RELEASE/DEBUG/PROFILING flags completely.
>

The original code does not do it either.  `CMAKE_Fortran_FLAGS_***`  would be overwritten by the code. this PR https://github.com/cp2k/cp2k/pull/2984 does the right thing. If we want that user provided option have high priority then the logic is completely different and  we should not muggle with options at all.  

Most of these options are common default (I probably miss few additional options but is is easy to check). I can always make them additive or not depending on everyone will.... 

>     * JSON does not permit comments which are sometimes used to describe why something was added (it is thanks to those that you know that you can likely drop some of the configurations as they may not be relevant anymore)

While true there is a workaround for this. You can just add a `"!comment" : "my pretty comment"` where you want. 

The cray hack is outdated as well and most probably not relevant anymore....

>     * The idea in DBCSR was to have working configurations for the most common compilers the project is being used on (extending on CMake's capabilities like C++ standard requirement flags)

What you are describing is what cmake.Presets provide. This PR uses this mechanism actually.  However the compilers need to be specified for each configuration which is tedious to do by hand. Scripting this to generate the cmake.presets could be a way to go. 
To be honest, if we need to use cmake.presets to compile cp2k then the cmake build system is poorly designed. the preset functionality is very useful for reproducibility but a simple `cmake` should work as well (although for cp2k we need to add `-DCP2K_SCALAPACK_VENDOR=GENERIC`)

> 
>     * Since CMake naturally has branching constructs, we could also model more complex scenarios and make them available to others, disentangle _what is being built_ from _how it is to be built_.

I am not sure to understand the argument here. It is possible to do a lot of things with cmake but not all of them are good. 
 
>     * lines 8-33 in the changeset, not 3. And this does not cover all the branching and cases in the current code nor version-dependent configurations.

it is not a valid argument. My point is it is easy to read by anyone except the `string()` function the rest is clear.

```
foreach(__var long_list)
string(JSON cp2k_reading_value ERROR_VARIABLE cp2k_JSON_ERROR GET ${CP2K_COMPILER_CONFIGURATION} "compiler_id" "${CMAKE_Fortran_COMPILER_ID}" "${__var}")
  if(NOT cp2k_JSON_ERROR) # if we do not want additive properties then we can change this
    set(${__var} "${${__var}} ${cp2k_reading_value}")
  endif()
endforeach()
`````

it is also possible to add quirks when needed and add comment why we do it. I need to add the sanitizer check (openmpi) for instance. 

- the cmake.Presets could be generated by a script so we can offer multiple presets options with compilers versions cp2k options, etc.... Writing these by hand is really tedious considering the possibilities cp2k offers.

**Comment by oschuett** [comment]
> the cmake.Presets could be generated by a script

Why don't we cut out the middlemen (JSON) and simply generate the CMake code directly?
We could leverage the expressiveness of Python while producing very easy-to-read CMake code.

**Comment by mtaillefumier** [comment]
> > the cmake.Presets could be generated by a script
> 
> Why don't we cut out the middlemen (JSON) and simply generate the CMake code directly? We could leverage the expressiveness of Python while producing very easy-to-read CMake code.

Actually both are equivalent. The advantage of cmake.Presets is that we can use `cmake --preset gcc.debug.full.arch.psmp `for instance and cmake will have all information provided by the json file. If we go that path (meaning the idea of having preset configuration) we should definitely generate them with python. Now we can use the cmake infrastructure or we can generate cmake code (in that case not in CMakeLists.txt). Both are explicit. i see one advantage of the cmake code, which is that we can set compilers, architectures, parallelism completely independently.

Let me explain It is possible to parse the string "gcc.debug.full.arch.psmp" and pick all relevant information without generating all possible combinations. 

for instance we could have a file (json, yaml, direct cmake code) with different sections (i take json as reference)

```
"compiler" : {
"gcc11" : {
"BLA": "fff",
"RED": "eee"
}
}

"build_options" :{
"full" : "",
"minimal" : "",
"default" : "",
"debug" : ""
}
```
so when we parse the string, we know we should take gcc, build with OPTION A,B,C etc.... without having to search for the right settings.

I would have done it this way if I could combine different presets together but cmake does not seem to support this.

I restored the possibility to set up the `C` and `Fortran` compilers independently at the prize of a bit more boiler plate code. The sanitizer quirk of openmpi is also present but I kicked out the cray quirk as it dates back from 4 years ago and even daint (which is the oldest cluster with CRAY env I know) has more recent versions of the CRAY environment. 

I also think it is a good time to check the compiler options (and if we want compiler versioning as well, doable but more code is needed). Compilation options do not match between the arch files and cmake for the intel compiler.

**Comment by oschuett** [comment]
Let's divide and conquer. I think this PR could be split into four independent parts:

#### 1. Simple clean ups e.g. around `message("" "")`.

This is uncontroversial. Just open a separate PR or re-purpose this one.

#### 2. New Dockerfiles for Spack:

This is also fine. I only noticed that there is a lot of reducancy between `Dockerfile.gcc_spack_cuda` and `Dockerfile.gcc_spack`, but I can take care of that afterwards.

#### 3. Refactoring options for various compiler like gcc, intel, cray, etc.

I think here we're over engineering a bit. For non-gcc compilers we only need a single configuration. This means we don't have to support debug/converage/etc for all compilers. To support older compiler version we could rely on [check_fortran_compiler_flag](https://cmake.org/cmake/help/latest/module/CheckFortranCompilerFlag.html) similar to what we use [today](https://github.com/cp2k/cp2k/blob/dfc565023ea3187a044d14b931fefe7b58c72354/tools/toolchain/scripts/tool_kit.sh#L433C1-L433C1). The resulting cmake code should be simple enough that JSON is no longer needed.

#### 4. Refactoring preset like MINIMAL,FULL,DEFAULT,SERIAL

Here we should indeed start to apply code generation. Essentially the whole chain from presets (e.g. `FULL`) to cmake flags (e.g. `CP2K_USE_VORI`) to build flags (e.g.`-D__LIBVORI`) should be generated to ensure consistency. Do you want to take this on @mtaillefumier or should I give it a try?

**Comment by mtaillefumier** [comment]
P1 : I can do this no problem.
P2 : Let me give it a try. I am no python expert but I may be able to find a way to do it more cleanly.

P3: my intent with the json file was to separate the configuration from the processing so that devs could simply edit the json file without learning about the cmake syntax. The check_fortran_compiler flag is a very good suggestion and also easier to implement as it is possible to chain them one after the other. It would be a long list of compilation options to check but it is fine with me. I can give it a try.

P4 : I agree because it is really tedious to write this by hand....  there is no emergency for this though.

**Comment by mtaillefumier** [comment]
i reverted back some of the changes for the time being. Another small patch is coming. I will commit it after https://github.com/cp2k/cp2k/pull/3008 is merged. I left the preset configurations though as it is still cleaner than my hand written cmake code.

About the last point. the correspondence CP2K_USE_VORI -> -D__LIBVORI is done in the CMakeLists.txt. I am not sure that generating these correspondences with python helps actually because we do not modify them that often. They also only appear once in the file so the probability to get this wrong is slim. None the less I agree that generating default combination of options should be done with a script. It will probably be more productive if we move this discussion outside this PR. 

We can surely improve this incrementally. 

@oschuett  any suggestion ?

File: CMakePresets.json [hunk 1]
```diff
@@ -0,0 +1,181 @@
+{
```
oschuett: This and the other JSON file should probably also be reverted for now.

mtaillefumier: no major objection on my side.

oschuett:
>  the correspondence CP2K_USE_VORI -> -D__LIBVORI ... I am not sure that generating these ... helps

Yeah, you might be right. 

> We can surely improve this incrementally.

Yes, let's take this one step at a time. After this is merged I'll get started on a draft PR for generating the presets. Then we can discuss the further details there.

mtaillefumier:
Oh oh cmake test fails. I can not reproduce it locally.

mtaillefumier:
Oh they changed policy about env variables. I need to look at this more carefully. Note it is not really an issue but better fix it now.

Next thing I should probably look at is the FPGA stuff. This will be complicated...

oschuett:
Yes, it's complaining about [CMP0144](https://cmake.org/cmake/help/latest/policy/CMP0144.html). The root cause is #3004. 

We should probably fix this (either here or in a separate PR) before merging.

> Next thing I should probably look at is the FPGA stuff.

We haven't received a contribution for FPGA code in a long time. I'm not actually sure it's still in use.

mtaillefumier:
i will leave the FPGA stuff away from my list. I was actually thinking enforcing cmake 3.25 because of one useful thing the block() command... Let's not get carried away too much. I will fix the issue ASAP

PR #3006, opened 2023-09-25
HFX warning for KS sparsity

BelizSertcan:
modify KS sparsity warning for HFX calculations to recommend the more efficient MIN_PAIR_LIST_RADIUS keyword.

### DIFF
--- a/src/hfx_energy_potential.F
+++ b/src/hfx_energy_potential.F
@@ -586,8 +586,8 @@ CONTAINS
          IF (.NOT. ks_fully_occ) THEN
             CALL cp_warn(__LOCATION__, &
                          "The Kohn Sham matrix is not 100% occupied. This "// &
-                         "may result in incorrect Hartree-Fock results. Try to decrease EPS_PGF_ORB "// &
-                         "and EPS_FILTER_MATRIX in the QS section. For more information "// &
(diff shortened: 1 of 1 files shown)


oschuett:
I'm not sure about this change. 

The [FAQ page](https://www.cp2k.org/faq:hfx_eps_warning) goes to great lengths to explain why `EPS_PGF_ORB` and `EPS_FILTER_MATRIX` are needed. Whereas it's not clear to me from the [description](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/QS.html#MIN_PAIR_LIST_RADIUS) what exactly setting `MIN_PAIR_LIST_RADIUS` to -1 does.

BelizSertcan:
@oschuett I modified the message. I also added a short section to the faq page but cannot publish myself.

abussy:
I think the new message is clearer. I also think this change is quite important, because people tend to use exceedingly small values of EPS_PGF_ORB, leading to a lot of wasted CPU time. Provided that the link in the warning is also updated, I think this is ok to go.

BelizSertcan:
I published the new message to FAQ page.

oschuett:
Thank for updating the FAQ page. I didn't realize `MIN_PAIR_LIST_RADIUS` was a novel keyword. Now everything makes sense :-)

Issue #2996, opened 2023-09-17
Regtesting Improvements


oschuett:
I'd like to make a number of improvements to our regtesting script:

1. Give [test types](https://github.com/cp2k/cp2k/blob/master/tests/TEST_TYPES) meaningful identifiers instead of just  numbers.

2. Allow to check also other outputs besides the stdout.

3. Allow for multiple checks after running an input file.

As a first step, I'd like to move the [do_regtest.py](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/do_regtest.py) script from `cp2k/tools/regtesting` to `cp2k/tests`.

Are there any comments or concerns?

fstein93:
I am in favor of your ideas as it allows us to reduce the number of calls and thus runtime of the test suite to CP2K whereas you second idea allows us to test common mistakes in input files of CP2K assuming that we can then check them improving the coverage.

juerghutter:
What about to add a new macro ?

CPTEST(TAG, VALUE)


________________________________________

oschuett:
> What about to add a new macro ? `CPTEST(TAG, VALUE)`

What would be the advantage over our current pattern matching?

These patterns are surprisingly stable. The last time one had to be update was [more than two years ago](https://github.com/cp2k/cp2k/commit/0c4637b047664bdb0e05708583f3c20750a15988#diff-946efca8ef1f8b3ef9e820ad626c632a0812bb34ba8babca9825b71e21f2dc7e).

juerghutter:
This would allow for a single pattern rule (with a changing tag) instead
of 100 different rules.
It is transparent for the developer and easy to handle.
It could run with the current scheme in parallel (adds 1 new rule).

________________________________________

oschuett:
Ok, now I see where you're going. You want to eliminate the [TEST_TYPES](https://github.com/cp2k/cp2k/blob/master/tests/TEST_TYPES) entirely. That is indeed tempting.

However, I see two downsides:

1. The tagged output will clutter the regular output. We could collected it and print everything at the very end, but then we'd have nothing in case of a midway crash.
2. There is no way to migrated the existing tests automatically. Hence, the TEST_TYPES file will stay around for the foreseeable future and we'll have to deal with [two systems](https://xkcd.com/927/).

I agree that adding a new test pattern is  a bit  cumbersome. I'm hoping to alleviate this by moving to a more familiar syntax, i.e. changing `TEST_TYPES` to `test_types.py`.

[...]

PR #3008, opened 2023-09-26
Move do_regtest.py into tests dir

oschuett:
First step toward #2996.


RMeli:
While you are at it, would you mind updating the following message on CMake?
https://github.com/cp2k/cp2k/blob/c94756ddc75c22fb829516d4d1ee9958f089b1c9/CMakeLists.txt#L1029

Otherwise I will do it later. Thanks.

mtaillefumier:
i will do it in my open pr

mtaillefumier:
@oschuett could you merge this one first then I can update mine with this change so there is no inconsistency between the two PRs.

mtaillefumier:
thanks

PR #3015, opened 2023-10-03
Projection for EMD

glb96:
Extend the MO projection for EMD. 

Either the MO coeff is time-independent and the time-dependent (nuclei position dependent) overlap matrix is used for the projection. In this case, the reference MO is not the same in time because it depends on the nuclei position. This case is propagate_ref FALSE

Either, the MO coeff is propagated using the B matrix so that the reference wave function is time-independent. This case is propagate_ref TRUE. 

Regtest added


oschuett:
The failing Misc test is unrelated. I'll take care of it this evening.

src/input_cp2k_projection_rtp.F
```diff
@@ -55,6 +55,21 @@ SUBROUTINE create_projection_rtp_section(section)
                           repeats=.TRUE.)
 
       NULLIFY (keyword, subsection)
+      CALL keyword_create(keyword, __LOCATION__, name="PROPAGATE_REF", &
+                          description="In case of Ehrenfest dynamics, the atomic basis set is evolving with time. "// &
+                          "The reference MO can either be understood as a spatial-dependent wave-function which is "// &
+                          "time-independent or to be 'attached' with respect to the nuclei position, and "// &
+                          "thus evolve in space as the nuclei moves. For the first case, set this variable to TRUE. "// &
+                          "Note that in this case you shall have enough atomic orbital accross the whole space to "// &
+                          "describe this MO as the nuclei will move and may leave the space where the MO is defined."// &
```
oschuett: The [precommit check](https://storage.googleapis.com/cp2k-ci/run-cp2k-precommit-5c1b6102_report.txt) is failing because there is a space missing after the dot.

glb96: Are you pointing to "" is defined."// &"" ? In this case, I have just fixed this.

src/input_cp2k_projection_rtp.F
```diff
@@ -78,7 +93,7 @@ SUBROUTINE create_projection_rtp_section(section)
 
       CALL keyword_create(keyword, __LOCATION__, name="REF_MO_INDEX", &
                           description="Indexes of the reference MO read from the .wfn reference file (see REF_MO_FILE_NAME). "// &
-                          "Use this keywork if REFERENCE_TYPE=SCF."// &
+                          "Use this keyword if REFERENCE_TYPE=SCF."// &
```
oschuett: Could you please also add a space after `REFERENCE_TYPE=SCF.` ?
Unfortunately, [my recent change](https://github.com/cp2k/cp2k/commit/7c90856a675bdfbc0d30dfa5c6a14958c10a4a2d) led to a merge conflicts.

glb96: Done

PR #3016, opened 2023-10-03
HOTFIX: EMD + Time Dependent Electric Field

glb96:
In the current version of the code, the electric field is applied to the electrons only at the first iteration of each EMD step. 

For RTP: 
The core Hamiltonian is built at each RTP step at the first iteration, see line 133 and following of emd/rt_propagation_methods%propagation_step. In particular, once qs_energies_init is called, the core Hamiltonian is updated with the dipole operator for time-dependent fields (or velocity one for VG).  qs_energies_init is then never called before the end of the RTP step: we loop over the ''good'' core Hamiltonian through one RTP step. 

For EMD: 
It follows the same procedure, but the forces are also computed in qs_force. There, qs_energies_init is called to compute the forces. Which also reset the core Hamiltonian: it should thus be updated with the dipole operator. If not, the electronic Hamiltonian applied for the next iteration of the EMD step is field-free. Yet, the electric field is still applied to the nuclei. This produces a dynamics where the field interacts explicitly with the nuclei but not with the electrons. 

I have performed some tests to confirm that. For instance: run EMD with field and show the total energy for an isolated neutral molecule. This total energy should be independent with respect to the position of the nuclei since the nuclei-field and electron-field energy add up. In today's implementation it is not the case: moving the nuclei position changes the total energy value. With this fix the total energy is indeed position independent. 

This is a problem I have induced in the commit 78ff48d from 24 Feb: RTP: Add velocity gauge. There I have moved the electric field - electron interaction from a grid-based implementation to a matrix one in the core Hamiltonian (in order to have a similar implementation as the velocity gauge case). Therefore, this issue has been a problem since 24 Feb for EMD + field. 

Hence, I have also modified the regtests relative to the EMD + field since they were no longer correct. The obtained values with this fix are almost the same as the ones before the 24 Feb: the error may be explained by the new way of computing the electron-field interaction using matrices instead of the grid.  
 

As this is a potentially important fix, I allow myself to call this PR HOTFIX.  Should we communicate about this issue concerning the previous version of the code? 

Ps: I will soon push EMD for the velocity gauge

### DIFF
--- a/src/qs_force.F
+++ b/src/qs_force.F
@@ -28,7 +28,8 @@ MODULE qs_force
                                               dbcsr_set
    USE dft_plus_u,                      ONLY: plus_u
    USE ec_env_types,                    ONLY: energy_correction_type
-   USE efield_utils,                    ONLY: calculate_ecore_efield
+   USE efield_utils,                    ONLY: calculate_ecore_efield,&
+                                              efield_potential_lengh_gauge
    USE energy_corrections,              ONLY: energy_correction
    USE excited_states,                  ONLY: excited_state_energy
    USE hfx_exx,                         ONLY: calculate_exx
@@ -290,6 +291,11 @@ CONTAINS
       ELSE
          ! Dispersion energy and forces are calculated in qs_energy?
          CALL build_core_hamiltonian_matrix(qs_env=qs_env, calculate_forces=.TRUE.)
+         ! The above line reset the core H, which should be re-updated in case a TD field is applied:
+         IF (qs_env%run_rtp) THEN
+            IF (dft_control%apply_efield_field) &
+               CALL efield_potential_lengh_gauge(qs_env)
+         END IF
          CALL calculate_ecore_self(qs_env)
          CALL calculate_ecore_overlap(qs_env, para_env, calculate_forces=.TRUE.)
          CALL calculate_ecore_efield(qs_env, calculate_forces=.TRUE.)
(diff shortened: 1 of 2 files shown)


glb96:
Thank for merging. Would it also be possible to do a cherry-pick of this commit to the 2023.2 release version of CP2K? In this version, this issue is still there (this is not the case in the 2023.1 version if I am correct).

oschuett:
Releases are generally meant to be immutable. So, we'd have to make new release, which is quite a bit of work.
Since this issue affects only a few users, I'd prefer to just wait for then next regular release (2024.1) towards the end of the year.

mkrack:
@glb96 You can prepare a patch for `2023.2` and offer it for download.
@oschuett Do we have a convenient way for offering such release patches via GitHub?

oschuett:
We don't have a process for minor releases yet.

One would probably start by pushing a commit into the [support/v2023.2](https://github.com/cp2k/cp2k/tree/support/v2023.2) branch and then run a subset of the [release checklist](https://www.cp2k.org/dev:release_checklist). But again, given our short regular cadence, I'm a bit reluctant to spend time on this.

glb96:
My apologies but I am not sure to understand what a patch would be. For instance, just the two new versions of the file in this commit in the same directory architecture then the cp2k main directory so that someone can just simply do: 

cp correction_patch_v2023.2/* cp2k/*

and then compile the code again?  

I can also open a thread in the cp2k google community page to briefly point out this problem and to provide with such a patch.

mkrack:
> My apologies but I am not sure to understand what a patch would be. For instance, just the two new versions of the file in this commit in the same directory architecture then the cp2k main directory so that someone can just simply do:
> 
> cp correction_patch_v2023.2/* cp2k/*
> 
> and then compile the code again?

I meant the `patch` command using the result from a `diff` command saved in a patch file as described [here](https://opensource.com/article/21/11/linux-diff-patch) for instance. That patch file (containing only the differences) can be applied in `cp2k/` after downloading the release version 2023.2.

glb96:
Thanks for the explanations, I did not know about this process. 
Here are 2 patch files to be used with : 

patch cp2k/src/qs_force.F qs_force.patch
patch cp2k/tests/QS/regtest-rtp-2/TEST_FILES TEST_FILES.patch

It applies this commit to version 2023.2 and makes a version that passes the regtests. 

I propose to open a discussion in the Google group where I distribute these files and explain the problems. Hence, if a user struggles with EMD + field for version 2023.2 and looks at the forum, he/she will find the solution. 


[V2023_2_patch.tar.gz](https://github.com/cp2k/cp2k/files/12878699/V2023_2_patch.tar.gz)

PR #3038, opened 2023-10-13
set(CMAKE_C_STANDARD 11)

haampie:
Since 77977db053aa4dd91906f197b16b60c454c6ff3f (or maybe earlier?) CP2K requires C11

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,7 @@ endif()
 
 # set language and standard
 set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_C_STANDARD 11)
 
 find_package(PkgConfig)
 


mtaillefumier:
we also need to remove the https://github.com/cp2k/cp2k/blob/cd939261d1ab248def8312b6758d454238b76a3d/src/CMakeLists.txt#L1498C1-L1498C46 

`target_compile_feature ` in that case.

haampie:
Do you wanna take over this PR @mtaillefumier? I know too little about CP2K to fix it properly :p seems like there's quite a mix of C++11 vs C++14, and C99 vs C11. Probably one standard is sufficient ;p

mtaillefumier:
i will do the cleanup. Keep this PR open for the time being. @oschuett do not merge this PR for the moment.

mtaillefumier:
@haampie  can you close this PR. I opened a new one https://github.com/cp2k/cp2k/pull/3039

PR #3043, opened 2023-10-16
Helium-Solute NNP for interaction [cschran/lduran/hforbert]

hforbert:


src/motion/helium_methods.F
```diff
@@ -118,9 +126,9 @@ SUBROUTINE helium_create(helium_env, input, solute)
       REAL(KIND=dp)                                      :: cgeof, dx, he_mass, mHe, rtmp, T, tau, &
                                                             tcheck, x1, x_spline
       REAL(KIND=dp), ALLOCATABLE, DIMENSION(:, :)        :: pot_transfer
-      TYPE(cp_logger_type), POINTER                      :: logger
-      TYPE(mp_comm_type)                                 :: new_comm
-      TYPE(section_vals_type), POINTER                   :: helium_section, input_worm
+      TYPE(cp_logger_type), POINTER                      :: logger, tmplogger
+      TYPE(mp_para_env_type), POINTER                    :: new_comm
```
fstein93, line 130: Why do you make this communicator a pointer? In your case you know that all pointers point to the same object. The others are just mere copies. I also do not see when you deallocate the pointer.

hforbert: I make that a pointer, because cp_logger_create needs a pointer and I cannot make it a target since it's part of a structure which I didn't want to make a target. I didn't deallocate because I wrongly assumed %free is enough :)
But I agree the pointers are a mess here and also the parallization (helium assumes only one rank per environment
and doesn't use mpi, which clashes with the regular nnp (hence I had to create a logger with a single rank, so nnp is happy).

fstein93: Ah, thanks. No, %free just frees the data of the object but it cannot free the pointer (or the object itself). And I mixed up mp_para_env_type and mp_comm_type. Then, it is ok by design.

PR #3065, opened 2023-10-23
Actually fix MPI_F08 in CMake

RMeli:
https://github.com/cp2k/cp2k/pull/3063 did not contain the full fix, my bad. I opened it directly from GitHub because the actual changes I had locally were on a messy branch. This PR has been tested locally.

I don't think it would have helped here, but is it possible to automatically trigger the CMake CI action, when CMake files are changed?


oschuett:
> is it possible to automatically trigger the CMake CI action, when CMake files are changed?

That a good idea. I've just implemented it via https://github.com/cp2k/cp2k-ci/commit/84b5904882a7bf15e56b7ec6ec926cace19246e0 and #3067.

RMeli:
@oschuett fantastic, thanks a lot!

RMeli:
Changes included in https://github.com/cp2k/cp2k/pull/3070

PR #3070, opened 2023-10-23
Use MKL implementation of fftw3 by default when mkl is found

mtaillefumier:
Added the possibility to compile cp2k with the FFTW3 implementation found in MKL. It can be turned off with  -DCP2K_USE_FFTW3_WITH_MKL=ON


mtaillefumier:
there is a bug either in cp2k or mpich when MPI_F08 is activated. Turning it off temporarily for testing

oschuett:
> there is a bug either in cp2k or mpich when MPI_F08 is activated. 

Sorry, it took me a while to recall. There is indeed an issue with MPICH on Ubuntu: https://github.com/cp2k/cp2k/pull/2486#issuecomment-1375718354

mtaillefumier:
No worries. I also have the same issue on fedora actually. mpich version is 4.0.3

mtaillefumier:
thanks for confirming this is an issue with mpich. I could find anything wrong with the cp2k interface. I also wonder if some issues cray-mpich are also related to mpi_f08....

mtaillefumier:
I can revert my change if needed.

PR #3075, opened 2023-10-27
Docker files for spack stack

mtaillefumier:
simplification of the spack containers. I tested the cpu version and it builds. I will keep testing the others for good measure. @oschuett : would it be posslble to check the python code because I am really not a python expert.

Side note : docker build should be ran in the cp2k root source code. I designed it for CI/CD actually. If we want to build cp2k from scratch then we need to add `git clone ....`


oschuett:
@mtaillefumier, thanks a lot for your quick response!

We probably should have discussed a bit more what exactly we want to test here. I'd like to ensure that our users can use Spack instead of the toolchain to build the development version of CP2K.

Therefore I'd propose that we:
- use a release version of Spack, because our users should too.
- speed up the build by using an existing Docker image, e.g. [spack/ubuntu-jammy:v0.20.2](https://hub.docker.com/layers/spack/ubuntu-jammy/v0.20.2/images/sha256-9f82e2e154bfdbdce340c046059f62d6c1e61e26c3b514f6c8aa122e722c2c2a?context=explore).
- rely as little as possible on Ubuntu because many users will have a different Linux distribution.
- do not rely on the Spack CP2K package to avoid a circular dependency.

What do you think?

mtaillefumier:
My only attempt with this PR was to fix the docker image itself, most certainly **not** replace the docker file used by the CI/CD. I took inspiration from the file we use in our internal CI/CD but screwed up somehow. You also rightfully suggested to simplify the python script generating these docker files which this PR does. I also fixed the issues you reported at the same time and added the command for running the regtests (but we can remove it if needed).

I targeted ubuntu specifically because it is a common distro but nothing in the docker file is ubuntu specific except the `apt-get` command. We can easily replace it with any distribution of our choice. I can include this in the python script if you think it a good idea. 

This docker file is generic and can be used to run cp2k normally in container environment. The spack command can also be used to generate a developer environment if you wish. The advantage is that you get the environment tailored for a specific set of options. I think in particular about the CUDA and ROCM dependencies.

nearly all steps are identical what ever we want to 

- test the build with spack
- create a developer environment (then the last command is different)
- run performance tests and include them in the dashboard.

I added the regtests but they can be removed. In that case the docker container will contain a version of cp2k that can be loaded with the single `spack load cp2k`. Nothing needs to be done as spack sets the environment properly. Doing so also provide the necessary dependencies to the developers. 
 
> @mtaillefumier, thanks a lot for your quick response!
> 
> We probably should have discussed a bit more what exactly we want to test here. I'd like to ensure that our users can use Spack instead of the toolchain to build the development version of CP2K.
> 
> Therefore I'd propose that we:
> 
>     * use a release version of Spack, because our users should too.

We will have to wait for the next release at the very least. Using spack develop is fairly safe as well it might break from time to time. Moreover it is the only way to ge the latest versions of all recipes  (cp2k and dependencies) without having to maintain our own local copy of these. Note it is doable but just a bit more work in maintaining two versions of the same recipe.
 
>     * speed up the build by using an existing Docker image, e.g. [spack/ubuntu-jammy:v0.20.2](https://hub.docker.com/layers/spack/ubuntu-jammy/v0.20.2/images/sha256-9f82e2e154bfdbdce340c046059f62d6c1e61e26c3b514f6c8aa122e722c2c2a?context=explore).

We gain absolutely nothing using a docker image that include spack already. What take time is to build cp2k dependencies as spack is a python package and does not require any compilation by itself. Moreover doing so lock in the versions of the recipe used to build cp2k which means all the solutions to many problems we encountered are not necessarily included in these images.
 
>     * rely as little as possible on Ubuntu because many users will have a different Linux distribution.

spack does not rely on ubuntu at all. I chose ubuntu for not specific reason but the same docker file will work on fedora for instance (provided we have the correct header). I agree with you about this though (see my first few comments).
 
>     * do not rely on the Spack CP2K package to avoid a circular dependency.

It can not happen. CP2K has no circular dependencies. 

Side note, I would not install all dependencies independently for many reasons. First cp2k might require specific options (libint is a typical case) which are properly set in the cp2k recipe. Second spack never rebuild a dependency if it does not need to. The only advantage of splitting the dependencies (that are CPU, or NVIDIA or ROCM only) in separate commands is for docker containers.

RMeli:
> We will have to wait for the next release at the very least. Using spack develop is fairly safe as well it might break from time to time. Moreover it is the only way to ge the latest versions of all recipes (cp2k and dependencies) without having to maintain our own local copy of these.

I agree with this. Since Spack packages and Spack itself are bound together for the time being, restricting ourselves to a released version is too restrictive, unfortunately. However, instead of `develop` I'd suggest to use a specific commit on the `develop` branch, which can be updated when needed (i.e. when a new PR targeting the CP2K package is merged).

RMeli:
> do not rely on the Spack CP2K package to avoid a circular dependency

@oschuett I'm not totally sure what you mean here, but I agree with @mtaillefumier that the best way forward is indeed to use the CP2K Spack package to install all the dependencies. This will ensure all the right options are turned on (modulo bugs in the Spack package). For example, for `libint` mentioned above,
```
spack install libint
```
would not work. The correct command would be
```
spack install libint +fortran tune=cp2k-lmax-6
```
This is automatically taken into account by the Spack package, so that
```
spack install cp2k +libint
```
installs the correct version. The advantage of using Spack is not having to worry about this, since all is encoded in the Spack package.

If you want to build CP2K without the Spack package, you can always rely on the Spack package to build only the dependencies with
```
spack install --only=dependencies cp2k +libint
```

and then build CP2K manually within the [build environment](https://spack.readthedocs.io/en/latest/command_index.html#spack-build-env) with
```
spack build-env cp2k+libint -- cmake ..
spack build-env cp2k+libint -- make -j
```

The latter is what I do for development.

oschuett:
@mtaillefumier, the Python script is indeed much nicer now. I'll happily merge this PR.

Just to clarify my earlier remarks, I see three use-cases: 

### 1) Folks who want to build the latest master version of CP2K
They can simply use `spack install cp2k@master`.
I assume this is already covered by Spack's own nightly testing?

### 2) Folks who want to make code changes to CP2K
They can use `spack install --only=dependencies cp2k; spack build-env...`.
This should also be covered by the same tests as 1).

### 3) Folks who want to upgrade one of CP2K's dependencies or add a new one

This is the tricky case because it often requires accompanying code changes. As a result CP2K's code and the way its dependencies are install have to be changed simultaneously. This is what I meant earlier by *circular dependency*.

I believe the only way to cover this use case is to accept a bit of duplication by maintaining a separate spec string for CP2K's dependencies within its own repository.

mtaillefumier:
> @mtaillefumier, the Python script is indeed much nicer now. I'll happily merge this PR.

That's one important point for me. I can add a bit more flexibility for the distro if needed.
 
> Just to clarify my earlier remarks, I see three use-cases:
> ### 1) Folks who want to build the latest master version of CP2K
> 
> They can simply use `spack install cp2k@master`. I assume this is already covered by Spack's own nightly testing?

yes. it is covered by spack nightly build and us as well. Note that spack install cp2k@master will give the minimal set of features. 

> ### 2) Folks who want to make code changes to CP2K
> 
> They can use `spack install --only=dependencies cp2k; spack build-env...`. This should also be covered by the same tests as 1).

that's also reasonable. Note that `spack install --only-dependencies` and `spack install cp2k@master` only differ by the later building cp2k and its dependencies while the former will only build the dependencies.
 
> ### 3) Folks who want to upgrade one of CP2K's dependencies or add a new one
> 
> This is the tricky case because it often requires accompanying code changes. As a result CP2K's code and the way its dependencies are install have to be changed simultaneously. This is what I meant earlier by _circular dependency_.

It is clearer now. You can do that directly with the cp2k recipe if you want to change how dependencies are built. the recipe for the dependency may not need to be modified at all actually (except bumping version numbers).

> I believe the only way to cover this use case is to accept a bit of duplication by maintaining a separate spec string for CP2K's dependencies within its own repository.

Upgrading dependencies might simply work with `spack install cp2k+master +cosma ^cosma@new_version`. The maintenance burden for maintaining cp2k recipe within cp2k is not big either (one python file, maybe few others if we want the primary dependencies) but we should not forget to eventually push these changes to spack. 

spack will also ignore the local build rules of the project, unless told otherwise. 

I can already include the cp2k recipe rules within cp2k with this pr if you think it is the right for doing so.

mtaillefumier:
@oschuett :I can remove the regtests from the image as well if you want to use the docker file for the dashboard.

oschuett:
>> I assume this is already covered by Spack's own nightly testing?

> yes. it is covered by spack nightly build and us as well. 

Perfect! Are the test results public by chance?

> I can remove the regtests from the image as well if you want to use the docker file for the dashboard.

No, please leave the regtests in. They make for a better dashboard test.

> I can already include the cp2k recipe rules within cp2k with this pr if you think it is the right for doing so.

That would be awesome. 

> but we should not forget to eventually push these changes to spack.

Yes, once a change has landed (which sometimes takes a couple of PRs) the new recipe rules should definitely be up-streamed to the Spack repository.

mtaillefumier:
> > > I assume this is already covered by Spack's own nightly testing?
> 
> > yes. it is covered by spack nightly build and us as well.
> 
> Perfect! Are the test results public by chance?

It is better to ask @haampie for confirmation. The spack build is used pretty much every day either by me or @RMeli . If something break we will correct it very quickly.
> 
> > I can remove the regtests from the image as well if you want to use the docker file for the dashboard.
> 
> No, please leave the regtests in. They make for a better dashboard test.

Perfect. Could you have a look at the sh script as you know better how how the tests are run.

> > I can already include the cp2k recipe rules within cp2k with this pr if you think it is the right for doing so.
> 
> That would be awesome.

Ok, I will include it right away.
 
> > but we should not forget to eventually push these changes to spack.
> 
> Yes, once a change has landed (which sometimes takes a couple of PRs) the new recipe rules should definitely be up-streamed to the Spack repository.

oschuett:
I saw you just added a full copy of the Spack CP2K recipe. 

I was actually hoping for a more lightweight "copy". Wouldn't just a bunch of `RUN spack install foo@1.2.3`  work as well?

... sorry I should have been clearer to begin with.

mtaillefumier:
> I saw you just added a full copy of the Spack CP2K recipe.
> 
> I was actually hoping for a more lightweight "copy". Wouldn't just a bunch of `RUN spack install foo@1.2.3` work as well?
> 
> ... sorry I should have been clearer to begin with.

no worries. I only copied the cp2k recipe not the dependencies. I do not think it is possible to start from the spack recipe and then add only new rules locally. We should have the entire file.

mtaillefumier:
the other option is that we keep only the master version of cp2k. It is possible that way. We should just be careful not to publish this version of the recipe in spack as they usually keep old versions of software as well.

oschuett:
> the other option is that we keep only the master version of cp2k. 

Right, we don't need any of the backwards compatibility stuff like the Makefile-support or the patch files. 

> We should just be careful not to publish this version of the recipe in spack as they usually keep old versions of software as well.

True, but we won't be able to keep the recipes in perfect sync anyways. So, the simpler our copy, the easier it will be to upstream changes.

Also, could you please move the new `spack` directory below the `/tools` directory.

mtaillefumier:
- recipe cleaned. Far simpler to maintain
- reverted one change made in the precommit files
- moved the spack directory to tools. 
- libvori needs fixing on its side (fPIC is not activated) so i exclude it for the time been. We need to contact the author as there is no github or gitlab for this library

mtaillefumier:
Oh we need one more thing is we want to use the local version of the cp2k recipe in the ci/cd. Shall I add it to the docker files ?

oschuett:
This is starting to look really nice :-)

> Oh we need one more thing is we want to use the local version of the cp2k recipe in the ci/cd. Shall I add it to the docker files ?

Yes, please. Once this PR is merged I'll setup a nightly Dashboard run for `Dockerfile.gcc_spack`.

mtaillefumier:
done

Issue #3078, opened 2023-10-29
RFE: Need to be able to build and install multiple MPI implementations


opoplawski:
The Fedora cp2k package builds for multiple MPI implementations - currently openmpi and mpich.  However with cp2k's current cmake build systems the two builds write their executables to the same location.  Also, the fedora package currently suffixes the executables with the MPI names, e.g. `cp2k.psmp_openmpi`.  I'm not sure this is strictly necessary as they are installed (finally) into different directories (`/usr/lib64/{mpich,openmpi/}bin`), but might be one solution.

Another might be to able to define a suffix to the `__cp2k_cmake_name` variable.

oschuett:
Fixed via #3079.

PR #3079, opened 2023-10-29
Allow setting CP2K_CMAKE_SUFFIX to append to __cp2k_cmake_name

opoplawski:
Possible fix for #3078


mtaillefumier:
No objection. Both solution `/usr/lib64/{openmpi,mpich}` and this one are fine with me

PR #3081, opened 2023-10-30
VG for EMD  

glb96:
Time-dependent Field for EMD. 

- Gauge transformed pseudo potential force
- Re-write some EMD routines 
- Merge the Ehrenfest force for the MO or density-based approach  (calc_c_mat_force_fm function has been deleted)

Note that many tests have been changed because of the last change but with a very few impact overall. This implementation should be more readable and maintenable as we use the already computed density matrix.


src/rt_propagation_types.F
```diff
@@ -255,6 +257,26 @@ SUBROUTINE rt_prop_create(rtp, mos, mpools, dft_control, template, linear_scalin
 
    END SUBROUTINE rt_prop_create
 
+! **************************************************************************************************
+!> \brief Initialize SinvH_imag for rtp
+!> \param rtp ...
+!> \param nspins ...
+! **************************************************************************************************
+   SUBROUTINE rtp_create_SinvH_imag(rtp, nspins)
+      TYPE(rt_prop_type), POINTER                        :: rtp
```
fstein93: I think it is better to remove the `POINTER` attribute which is not required here. Instead, you may add INTENT(INOUT) for better documentation.

glb96: Thank for the help :)

PR #3088, opened 2023-11-02
Allow setting CP2K_DATA_DIR

opoplawski:
The Fedora package keeps a single copy of the data for all builds (serial, openmpi, mpich) in /usr/share/cp2k/data.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,6 +169,10 @@ set(CP2K_BUILD_OPTIONS
 set_property(CACHE CP2K_BUILD_OPTIONS PROPERTY STRINGS
                                                ${CP2K_BUILD_OPTIONS_LIST})
 
+set(CP2K_DATA_DIR
+    "default"
+    CACHE STRING "Set the location for cp2k data")
+
 string(TOUPPER ${CP2K_BUILD_OPTIONS} cp2k_build_options_up)
 
 if(NOT ${cp2k_build_options_up} IN_LIST CP2K_BUILD_OPTIONS_LIST)
@@ -759,10 +763,10 @@ if(CP2K_BUILD_DBCSR)
   add_library(DBCSR::dbcsr ALIAS dbcsr)
 endif()
 
+include(GNUInstallDirs)
+
 # subdirectories
 add_subdirectory(src)
-
-include(GNUInstallDirs)
 
 get_target_property(CP2K_LIBS cp2k_link_libs INTERFACE_LINK_LIBRARIES)
 configure_file(cmake/libcp2k.pc.in libcp2k.pc @ONLY)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1573,6 +1573,10 @@ target_link_libraries(
 
 string(TIMESTAMP CP2K_TIMESTAMP "%Y-%m-%d %H:%M:%S")
 
+if(CP2K_DATA_DIR STREQUAL "default")
+  set(CP2K_DATA_DIR "${CMAKE_INSTALL_FULL_DATAROOTDIR}/${PROJECT_NAME}/data")
+endif()
+
 target_link_libraries(cp2k PUBLIC cp2k_link_libs)
 target_compile_definitions(
   cp2k
@@ -1583,7 +1587,7 @@ target_compile_definitions(
     __COMPILE_DATE=\"${CP2K_TIMESTAMP}\"
     __COMPILE_HOST=\"${CP2K_HOST_NAME}\"
     __COMPILE_REVISION=\"${CP2K_GIT_HASH}\"
-    __DATA_DIR=\"${CMAKE_INSTALL_FULL_DATAROOTDIR}/cp2k/data\"
+    __DATA_DIR=\"${CP2K_DATA_DIR}\"
     __COMPILE_ARCH=\"${CMAKE_SYSTEM_PROCESSOR}\"
     $<$<CONFIG:COVERAGE>:__NO_ABORT>
     $<$<CONFIG:DEBUG>:__HAS_IEEE_EXCEPTIONS>
@@ -1777,8 +1781,7 @@ install(
   INCLUDES
   DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}")
 
-install(DIRECTORY "${CMAKE_SOURCE_DIR}/data"
-        DESTINATION "${CMAKE_INSTALL_FULL_DATAROOTDIR}/${PROJECT_NAME}")
+install(DIRECTORY "${CMAKE_SOURCE_DIR}/data/" DESTINATION "${CP2K_DATA_DIR}")
 
 # ##############################################################################
 # cp2kConfig.cmake, etc...


src/CMakeLists.txt
```diff
@@ -1769,8 +1769,7 @@ install(
   INCLUDES
   DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}")
 
-install(DIRECTORY "${CMAKE_SOURCE_DIR}/data"
-        DESTINATION "${CMAKE_INSTALL_FULL_DATAROOTDIR}/${PROJECT_NAME}")
+install(DIRECTORY "${CMAKE_SOURCE_DIR}/data/" DESTINATION "${CP2K_DATA_DIR}")
```
oschuett, line 1784: I think, this will copy the files to `${CMAKE_INSTALL_FULL_DATAROOTDIR}/cp2k/data/data`.

oschuett (commented):
Unfortunately, your change doesn't pass the regressions tests.

Note that for those regtests we don't actually install the binary, but just create a symlink to where the binary expects the data to be: https://github.com/cp2k/cp2k/blob/27465532c224d97ca9b5ed5de7173fb2f9dcfc01/tools/docker/scripts/test_regtest_cmake.sh#L58-L61.


CC: @mtaillefumier

src/CMakeLists.txt
opoplawski: The final `/` on the `DIRECTORY` argument tells cmake to replace the destination with that directory.  It works fine for me.

opoplawski:
Is `${PROJECT_NAME}` not getting set?  I changed `cp2k` to `${PROJECT_NAME}` in `__DATA_DIR` because that seemed like an appropriate, but perhaps it's needed for the builds?

opoplawski:
Is the log from the actual cmake build available somewhere?

opoplawski:
So I think the problem was ordering - and I hadn't tested the case then `CP2K_DATA_DIR` was not set.  I think because it was set after including the `src` directory the variable wasn't set when `__DATA_DIR` was defined and it was empty.

I'm not sure that this is the best location for the `set`, but it at least keeps the variable set and usage close.

src/CMakeLists.txt
```diff
@@ -1565,6 +1565,10 @@ target_link_libraries(
 
 string(TIMESTAMP CP2K_TIMESTAMP "%Y-%m-%d %H:%M:%S")
 
+if(NOT CP2K_DATA_DIR)
```
mtaillefumier: That's not a proper way to do it. If you want to change the default location for the data directory (and rightfully so) you have to add an compilation option to the CMakeLists.txt instead of relying on a variable that you define manually.

what i mean is this

`set(CP2K_DATA_DIR  "default" CACHE STRING "Set the location for cp2k data")` 

It has to moved to the main CMakeLists.txt (not the src/CMakeLists.txt)

then the if statement should be 

```
if (NOT CP2K_DATA_DIR MATCHES "default")
set(CP2K_DATA_DIR ${CMAKE_INSTALL_FULL_DATAROOTDIR}/cp2k/data)
endif
```
the rest is alright

mtaillefumier:
> Is `${PROJECT_NAME}` not getting set? I changed `cp2k` to `${PROJECT_NAME}` in `__DATA_DIR` because that seemed like an appropriate, but perhaps it's needed for the builds?

I agree with you. It is one of these small things we keep discovering over time. When i did not know cmake well enough (and good practices) we started to work on the cmake port so these beginners mistakes (using PROJECT_NAME, etc) get fixed very slowly.

src/CMakeLists.txt
opoplawski: Need to drop the `NOT` here, right?

opoplawski: And why not `option()` instead of `set()`?

opoplawski: Ah - option() is only for booleans.

mtaillefumier: there is a simpler option than what I suggested. It is possible to set CMAKE_INSTALL_FULLDATAROOTDIR when you configure cp2k itself

`cmake -DCMAKE_INSTALL_FULLDATAROOTDIR=/usr/share ..` should give the expected behavior. I realized afterwards that this solution should be working. it means we do not need to set an option and we can keep `${CMAKE_INSTALL_FULL_DATAROOTDIR}/${PROJECT_NAME}/data` everywhere or simply do what you did without the if statement

set(CP2K_DATA_DIR ${CMAKE_INSTALL_FULL_DATAROOTDIR}/cp2k/data)

information here https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

mtaillefumier: ${PROJECT_NAME} should  be defined by cmake automatically.

CMakeLists.txt
```diff
@@ -158,6 +158,10 @@ set(CP2K_BUILD_OPTIONS
 set_property(CACHE CP2K_BUILD_OPTIONS PROPERTY STRINGS
                                                ${CP2K_BUILD_OPTIONS_LIST})
 
+set(CP2K_DATA_DIR
+    "default"                                                                                                                                                 |
+    CACHE STRING "Set the location for cp2k data")                                                                                                            |
```
oschuett: These `|` characters seem out of place.

oschuett:
@opoplawski, gentle ping :-)

src/CMakeLists.txt
opoplawski: > there is a simpler option than what I suggested. It is possible to set CMAKE_INSTALL_FULLDATAROOTDIR when you configure cp2k itself
> 
> `cmake -DCMAKE_INSTALL_FULLDATAROOTDIR=/usr/share ..` should give the expected behavior. I realized afterwards that this solution should be working. it means we do not need to set an option and we can keep `${CMAKE_INSTALL_FULL_DATAROOTDIR}/${PROJECT_NAME}/data` everywhere 

This isn't working for me:
```
+ /usr/bin/cmake -S . -B redhat-linux-build_mpich -DCMAKE_C_FLAGS_RELEASE:STRING=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE:STRING=-DNDEBUG -DCMAKE_Fortran_FLAGS_RELEASE:STRING=-DNDEBUG -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DCMAKE_INSTALL_DO_STRIP:BOOL=OFF -DCMAKE_INSTALL_PREFIX:PATH=/usr -DINCLUDE_INSTALL_DIR:PATH=/usr/include -DLIB_INSTALL_DIR:PATH=/usr/lib64 -DSYSCONF_INSTALL_DIR:PATH=/etc -DSHARE_INSTALL_PREFIX:PATH=/usr/share -DLIB_SUFFIX=64 -DBUILD_SHARED_LIBS:BOOL=ON -DCP2K_BLAS_VENDOR=FlexiBLAS -DCP2K_ENABLE_REGTESTS=ON -DCMAKE_PREFIX_PATH:PATH=/usr/lib64/mpich -DCMAKE_INSTALL_PREFIX:PATH=/usr/lib64/mpich -DCMAKE_INSTALL_FULL_DATAROOTDIR:PATH=/usr/share -DCMAKE_INSTALL_Fortran_MODULES:PATH=/usr/lib64/gfortran/modules/mpich/cp2k -DCMAKE_INSTALL_LIBDIR:PATH=lib -DCP2K_CMAKE_SUFFIX=_mpich -DCP2K_USE_MPI_F08:BOOL=ON
```
Still results in data being installed in `/usr/lib64/mpich/share/cp2k/data`.  This is because while GNUInstallDirs sets and uses this variable, it makes no provision for overriding it.

CMakeLists.txt
opoplawski: That was weird, not sure where those came from

PR #3092, opened 2023-11-04
Fix parsing atom sites from CIF files

### LINKED ISSUE #3085: Wrong atom kinds with cif file generated by ase
Usually the atom_info of a cif file (e.g. Material Studio) will follow the following order
```
_atom_site_label
_atom_site_type_symbol
_atom_site_fract_x
_atom_site_fract_y
_atom_site_fract_z
Ti0    Ti    0.00000   0.00000   0.00000
```
But cif file generated by ase may follow order like
```
_atom_site_type_symbol
_atom_site_label
_atom_site_fract_x
_atom_site_fract_y
_atom_site_fract_z
Ti  Ti1       0.00000  1.00000  0.02294 
```
which will cause subroutine `read_coordinate_cif` in topology_cif.F take `_atom_site_label` as the atom type. For example, the type of oxygen atom is O1, but the other oxygen atom is O2.
The result is that atom kinds is equal to the number of atoms, which may lead a HF funcitonal calculation cost several times the time. My solution now is turn `USE_ELEMENT_AS_KIND` on. But I still wish this bug can be fixed. The following are my modifications to the code, please let me know if there are any problems. With my respect.
```patch
diff --git a/src/topology_cif.F b/src/topology_cif.F
index 7eab185f1..164a59f47 100644
--- a/src/topology_cif.F
+++ b/src/topology_cif.F
@@ -161,15 +161,15 @@ CONTAINS
       IF (.NOT. found) &
          CPABORT("The field _atom_site_label was not found in CIF file! ")
       ifield = 0
-      ii = 1
-      ifield(ii) = 0
+      ii = 0
+      ! ifield(ii) = 0
       CALL parser_search_string(parser, "_atom_site_", ignore_case=.FALSE., found=found, &
                                 begin_line=.FALSE., search_from_begin_of_file=.TRUE.)
       DO WHILE (INDEX(parser%input_line, "_atom_site_") /= 0)
-         IF (INDEX(parser%input_line, "_atom_site_label") /= 0) THEN
-            CALL parser_get_next_line(parser, 1)
-            CYCLE
-         END IF
+         ! IF (INDEX(parser%input_line, "_atom_site_label") /= 0) THEN
+            ! CALL parser_get_next_line(parser, 1)
+            ! CYCLE
+         ! END IF
          ii = ii + 1
          CPASSERT(ii <= 20)
```

oschuett:
Follow up to #2843, fixes #3085.


oschuett:
This fix ended up requiring more refactoring than I anticipated...

It turns out that the [_atom_site_type_symbol](https://www.iucr.org/__data/iucr/cifdic_html/1/cif_core.dic/Iatom_site_type_symbol.html) is optional, but if it's present then it takes precedence over the [_atom_site_label](https://www.iucr.org/__data/iucr/cifdic_html/1/cif_core.dic/Iatom_site_label.html).

PR #3093, opened 2023-11-05
Fix SIRIUS scripts for different sed on Darwin

tsaihsiaoyi:
This pull request fixes an issue with SIRIUS scripts on Darwin where sed behaves differently.

### DIFF
--- a/tools/toolchain/scripts/stage8/install_sirius.sh
+++ b/tools/toolchain/scripts/stage8/install_sirius.sh
@@ -116,7 +116,7 @@ case "$with_sirius" in
 
       # GCC 13 stopped including some common headers.
       # https://github.com/electronic-structure/SIRIUS/issues/854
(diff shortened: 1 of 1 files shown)


mkrack:
You can also use `gsed` on `Darwin`  (see [this arch file](https://github.com/cp2k/cp2k/blob/master/arch/Darwin-gnu-arm64.psmp#L36)) and create a link `/usr/local/bin/sed` to it instead of using the legacy `sed` version of `Darwin` which is only kept for license issues. Otherwise, any further use of `sed` within the CP2K toolchain has to be tested always with that `sed` legacy version on `Darwin` to ensure that `sed` and `gsed` produce the same output.

tsaihsiaoyi:
You r right, this fix only ensure `sed` on both arch would behave in the same way. Because the arch file cant link the `gsed` to `sed` automatically. That's why I send this pr to u. Only make it  easier to compile on Mac :)

PR #3104, opened 2023-11-10
Build plumed with all modules by default

mkrack:


oschuett:
Out of curiosity, which important modules are missing by default?

mkrack:
> Out of curiosity, which important modules are missing by default?

Obviously several PLUMED modules are not installed by default as listed [here](https://www.plumed.org/doc-v2.8/user-doc/html/mymodules.html). If PLUMED compiles fine with all modules within the CP2K toolchain and if that does not require too much extra compile time, then I see no reason why we should not provide PLUMED with all available modules. In case there is a problem with one of these non-default modules, this will be rather an issue for the PLUMED community than for us.

oschuett:
Generally, I have no problem with enabling all PLUMED modules. 

I'm mostly asking because of our production containers (incl. for NGC). Do you know how much additional space these non-default modules take up?

mkrack:
> I'm mostly asking because of our production containers (incl. for NGC). Do you know how much additional space these non-default modules take up?

Good point! The image size increases by 3 MB from [786 MB](https://hub.docker.com/r/cp2k/cp2k/tags) to [789 MB](https://hub.docker.com/r/mkrack/cp2k/tags) for the `master_mpich_generic_psmp` image.

PR #3106, opened 2023-11-11
PW: Bind functions to objects, remove reference counting from pw_poisson_type

fstein93:
Within this PR,
- the PW creation and release routines are bound to the pw_type and the pw_pool_type
- removes the reference counting from the pw_poisson_type


dev-zero:
@fstein93 awesome! Maybe drop the `_pw` suffix on the functions as well?

fstein93:
I decided against it because `<object>%create()` sounds like a constructor of `<object>` whereas the routine is supposed to allocate a pw_type from the pool instead.

PR #3108, opened 2023-11-13
Fix Wannier localization when using LOW SPIN ROKS

IlyaFed:
- Added the ability to calculate Wannier Сenters for the ROKS method (only LOCALIZE NONE and JACOBI).
- Added the error message when ROKS and LOCALIZE were used incorrectly (instead of Segmentation Fault).
- Replaced the output of two spins in ROKS by the output of spin 1 only, since the densities of spin_1/2 are the same.

I have previously raised this issue in the Google Group discussion https://groups.google.com/g/cp2k/c/yJPJS2vUkBI

This Wannier centres analysis is added in analogy to the implementation in CPMD code. See https://github.com/CPMD-code/CPMD/blob/main/src/ddipo_utils.mod.F90, The fragment “ELSEIF (lspin2%tlse) THEN…” means that for singly occupied orbitals the Jacobi rotation is disabled. In this case, the number of orbitals to which the Jacobi rotation can be applied has been reduced by 2 since SOMO-1 and SOMO-2 orbitals should not rotate.


src/qs_localization_methods.F
```diff
@@ -365,6 +370,11 @@ SUBROUTINE jacobi_rotations_serial(weights, zij, vectors, max_iter, eps_localiza
       CALL cp_fm_get_info(rmat, nrow_global=nstate)
       tolerance = 1.0e10_dp
 
+      IF (restricted > 0) THEN
+         WRITE (6, *) "JACOBI: The last ", restricted, " orbitals do not rotate"
```
mkrack: Note CP2K's coding convention [#c012](https://www.cp2k.org/conv#c012) concerning hardwired unit numbers.

src/qs_localization_methods.F
```diff
@@ -2208,6 +2220,13 @@ SUBROUTINE jacobi_rot_para(weights, zij, vectors, para_env, max_iter, eps_locali
 
       CALL cp_fm_get_info(rmat, nrow_global=nstate)
 
+      IF (restricted > 0) THEN
+         IF (output_unit > 0) THEN
+            WRITE (6, *) "JACOBI: The last ", restricted, " orbitals do not rotate"
```
mkrack: Same, see [here](https://www.cp2k.org/conv#c012).

mkrack:
Thanks for your contribution. Could you also add a regression test input for that new feature?

src/qs_localization_methods.F
IlyaFed: Sorry, I missed it. Replaced it in the code with cp_logger_get_default_unit_nr().

src/qs_localization_methods.F
IlyaFed: Sorry, I missed it. Replaced it in the code with cp_logger_get_default_io_unit().

IlyaFed:
I added two tests to QS/regtest-lsroks

The first test checks the output of localized Wannier Centers for ROKS. It would be more correct to check the centers themselves, but as far as I understand, the regression tests do not support this and I added a check for “Total Spread (Berry)”.

The second test checks the MO_CUBES output for ROKS. It would be more correct to check that it outputs cube files only for spin-1, but as I understand it, this is also not supported, so I added a check that the program worked through "ENERGY| Total FORCE_EVAL" and a comment to the input.

IlyaFed:
I'm sorry, could you explain what's wrong with Regtest sdbg, it says that all tests were successful...

mkrack:
> I'm sorry, could you explain what's wrong with Regtest sdbg, it says that all tests were successful...

Just one slow test was detected. That can happen sometimes without a serious reason. I requested to re-run the sdbg test.

IlyaFed:
> > I'm sorry, could you explain what's wrong with Regtest sdbg, it says that all tests were successful...
> 
> Just one slow test was detected. That can happen sometimes without a serious reason. I requested to re-run the sdbg test.

Thank you.

PR #3115, opened 2023-11-15
Fix FFTW3 from MKL usage

RMeli:
Follow up to #3070.

Fixes
```
 *** WARNING in environment.F:1046 :: FFT library FFTW3 is not available. ***
 *** Trying FFT library FFTSG.                                            ***
```
when using MKL only.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1567,45 +1567,47 @@ string(TIMESTAMP CP2K_TIMESTAMP "%Y-%m-%d %H:%M:%S")
 target_link_libraries(cp2k PUBLIC cp2k_link_libs)
 target_compile_definitions(
   cp2k
-  PUBLIC $<$<BOOL:${CP2K_USE_MPI}>:__parallel>
-         $<$<BOOL:${CP2K_USE_MPI}>:__SCALAPACK>
-         $<$<BOOL:${CP2K_USE_MPI_F08}>:__MPI_F08>
-         __COMPILE_DATE=\"${CP2K_TIMESTAMP}\"
-         __COMPILE_HOST=\"${CP2K_HOST_NAME}\"
-         __COMPILE_REVISION=\"${CP2K_GIT_HASH}\"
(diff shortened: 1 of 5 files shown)


src/CMakeLists.txt
```diff
@@ -1567,45 +1567,47 @@ string(TIMESTAMP CP2K_TIMESTAMP "%Y-%m-%d %H:%M:%S")
...
+    $<$<BOOL:${CP2K_USE_VORI}>:__LIBVORI>
+    $<$<BOOL:${CP2K_USE_SPGLIB}>:__SPGLIB>
+    $<$<BOOL:${CP2K_USE_SIRIUS}>:__SIRIUS>
+    $<$<BOOL:${CP2K_USE_SPLA}>:__SPLA>
+    $<$<BOOL:${CP2K_USE_SpFFT}>:__SPFFT>
+    $<$<BOOL:${CP2K_USE_SPLA_GEMM_OFFLOADING}>:__OFFLOAD_GEMM>
+    $<$<BOOL:${CP2K_USE_ELPA}>:__ELPA>
+    $<$<BOOL:${CP2K_USE_LIBXC}>:__LIBXC>
+    $<$<OR:$<BOOL:${CP2K_USE_FFTW3_}>,$<BOOL:${CP2K_USE_FFTW3_MKL_}>>:__FFTW3>
+    $<$<BOOL:${CP2K_USE_FFTW3_MKL_}>:__FFTW3_MKL>
```
RMeli, line 1596: This is the actual change in this file. Other changes are from `make pretty`.

oschuett: Yeah, that's indeed weird. We're using [cmake_format](https://github.com/cheshirekow/cmake_format), which is yet another formatter project that fell dormant.

RMeli: Sad. =( But I think here the formatting might have changed because of the length of line 1595. (Just a guess, I haven't looked into it.)

oschuett: It's probably because your new line would exceed 80 chars with the generous indentation.

RMeli: To clarify, my comment above is just meant to facilitate the review.

PR #3121, opened 2023-11-21
Explicitly initialize HIP/CUDA before MPI

RMeli:
On our system, running CP2K on multiple nodes with HIP (version `5.2.3`) results in the following error:
```
ERROR: hipErrorNoDevice /users/rmeli/git/cp2k/src/offload/offload_library.c 51
```

Initializing HIP before MPI circumvents the issue. This PR adds `offload_init` to call `cuInit`/`hipInit` explicitly, before `mpi_world_init`.

I've been told this was an issue on LUMI as well (and apparently it is no longer), but I wasn't able to find additional information.

---

Credit to @msimberg for suggesting the workaround.


src/f77_interface.F
```diff
@@ -226,6 +227,8 @@ SUBROUTINE init_cp2k(init_mpi, ierr)
          ! get runtime information
          CALL get_runtime_info()
 
+         CALL offload_init()
```
oschuett, line 232: Maybe add a comment that the GPU needs to be initialized before MPI.

oschuett:
The CUDA build fails because `cuInit` is a low-level routine and we're currently only importing `cuda_runtime.h`:

https://github.com/cp2k/cp2k/blob/39fc2feded38fb50f7b1c74dba66523010dcedba/src/offload/offload_runtime.h#L18

Maybe it would be sufficient to just call `offload_get_device_count()` before `mp_world_init()` ?

RMeli:
> Maybe it would be sufficient to just call offload_get_device_count() before mp_world_init() ?

It is. That was actually my first test and that's why I defined the `offload_device_count` variable. But I think being explicit is better in this case?

It should be sufficient to include `cuda.h` (and `hip_runtime_api.h`), I think. However, I don't mind moving
```
offload_device_count = offload_get_device_count()
```
where `CALL offload_init` currently is, if that's preferred.

oschuett:
>  But I think being explicit is better in this case?

You're right, explicit is better.

src/offload/offload_library.c
```diff
@@ -39,6 +45,23 @@ const uint32_t colormap[] = {0xFFFFFF00,  // Yellow
                              0xFF0000FF,  // Blue
                              0xFF000080}; // Navy
 
+/*******************************************************************************
+ * \brief Initialize runtime.
+ * \author Rocco Meli
+ ******************************************************************************/
+void offload_init(void) {
+#ifdef __OFFLOAD_CUDA
```
msimberg: Nit, for symmetry:
```suggestion
#if defined(__OFFLOAD_CUDA)
```
?

msimberg (approved):
Can't comment on the CP2K internals and organization, but can confirm that this "fixes" the problem. Specifically, it seems to be an issue with Cray's MPICH and/or Cray's libfabric but I also don't have more specific information than that.

RMeli:
The dashboard is a bit unhappy after this PR, but looks like most (all?) of the issues are related to #3122.

PR #3125, opened 2023-11-22
open-shell periodic GW

JWilhelm:


src/gw_methods.F
```diff
@@ -2747,13 +2908,14 @@ SUBROUTINE G0W0_hamiltonian(bs_env)
...
                          beta=0.0_dp, matrix_c=bs_env%fm_h_G0W0_Gamma)
 
-      IF (bs_env%do_soc) THEN
-         ! Spinor G0W0-matrix
-         ! H^G0W0_µν,σσ' = h^G0W0_µν*δ_σσ' + sum_α V^SOC_µν^(α)*Pauli-matrix^(α)_σσ',
-         ! see Hartwigsen, Goedecker, Hutter, Eq.(18) (doi.org/10.1103/PhysRevB.58.3641)
-         CALL H_KS_spinor(bs_env%cfm_h_G0W0_spinor_ao_Gamma, bs_env%fm_h_G0W0_Gamma, &
-                          bs_env%mat_V_SOC_xyz(:, 1))
-      END IF
+      ! JW think again how to best incorporate SOC in GW...
```
fstein93, line 2911: Is it necessary to keep the code then?

src/post_scf_bandstructure_methods.F
```diff
@@ -144,12 +144,14 @@ SUBROUTINE bandstructure_primitive_cell_all_methods(qs_env, bs_env)
...
                                                   "bandstructure_SCF_SOC.bs", &
                                                   bs_env%cfm_ks_spinor_ao_Gamma)
-         IF (bs_env%do_gw) THEN
-            CALL bandstructure_primitive_cell_spinor(qs_env, bs_env, &
-                                                     bs_env%eigenval_prim_cell_G0W0_soc, &
-                                                     "bandstructure_G0W0_SOC.bs", &
-                                                     bs_env%cfm_h_G0W0_spinor_ao_Gamma)
-         END IF
+
+! JW todo later: GW+SOC band structure of primitive cell
```
fstein93, line 148: Same here

JWilhelm:
Thanks for pointing this out. I will implement the GW+SOC, should be finished in a few days. It would be cool if you could leave this PR open...

fstein93:
As that seems to be a new/improved feature, I am in favor of having a separate PR then. That would also facilitate the testing depending on how much both features are intertwined with each other.

JWilhelm:
Ok, then I am also fine to merge this PR

PR #3130, opened 2023-11-24
periodic G0W0+SOC

JWilhelm:


tests/TEST_TYPES
```diff
@@ -106,6 +106,7 @@ Ground state stabilisation: !4
 TDDFT+SOC !5
 G0W0 direct band gap !6
 SCF+SOC direct band gap !6
+G0W0+SOC direct band gap !6
```
oschuett, line 109: You have to increment the counter in the first line when adding a new test type.

JWilhelm: Thanks, Ole!

PR #3132, opened 2023-11-27
[SIRIUS] Add on-site and intersite features.

mtaillefumier:
This PR contains changes allowing calculations of DFT+U and DFT+U+V in the plane wave domain and an updated version of SIRIUS accompanying it. It uses CP2K input files only and information are retrieved from the already present DFT+U input parameters. 

The U+V approximation has its own input section within the DFT section and only works in combination to SIRIUS as CP2K does not support this approximation at the time of writing.

To use the U+V section, include the following to the DFT section of the input file

```
&DFT_PLUS_U ON
      SUBSPACE_METHOD "full_orthogonalization"
&END DFT_PLUS_U
&DFT_U_PLUS_V ON
     & LINK
         ATOM_PAIR 1 2
         ANGULAR_MOMENTUM 1 2
         ATOMIC_LEVEL 2 3
         V 1.24532876
     &END LINK
&END DFT_U_PLUS_V
```
Each atom pair contributing the intersite hubbard coupling should be described with one `&LINK &END LINK` and all keywords within the `link`  section are mandatory.

the `&DFT_PLUS_U ON &END DFT_PLUS_U` controls how the hubbard subspace is orthogonalized. 

- full_orthogonalization : take all atomic orbitals found in the pseudo potentials to orthogonalize the hubbard subspace using the Lowin method
- orthogonalize : Orthogonalize the hubbard wavefunctions within themselves only
- normalize : normalize the hubbard wave functions
- none : take the bare wave functions.

Atoms are referenced by their indices as we have no simple mean to actually describe them without indicating their position. It means that the indexation should be the same all the time.

As the name suggests "intersite" means two distinct atoms. CP2K will exit if the two atoms referenced by the atom_pair parameter are the same and in the unit cell.

for instance 

```
ATOM_PAIR 1 1
TRANSLATION 1 0 0

```  
is allowed but 

```
ATOM_PAIR 1 1
TRANSLATION 0 0 0

```  
will cause an error. intersite should not be used to describe multi-orbitals onsite calculations....


src/input_cp2k_dft.F
```diff
@@ -9721,4 +9729,143 @@ SUBROUTINE create_cdft_opt_section(section)
...
+      ! SIRIUS_HUBBARD_SUBSPACE_NORMALIZE 2
+      ! SIRIUS_HUBBARD_SUBSPACE_ORTHOGONALIZE 3
+
+      CALL keyword_create(keyword, __LOCATION__, &
+                          name="SUBSPACE_METHOD", &
+                          description="Select which subspace method to use for generating the hubbard subspace "// &
+                          "NONE : Use bare atomic wave functions "// &
+                          "NORMALIZE : Normalize the wave functions contributing to the hubbard subspace "// &
+                          "ORTHOGONALIZE : Orthogonalize the wave functions contributing to the hubbard subspace "// &
+                          "FULL_ORTHOGONALIZATION : Use all atomic wave functions to orthogonalize the hubbard subspace ", &
```
oschuett: You can use `enum_desc` for the description of enums, see e.g. https://github.com/cp2k/cp2k/blob/bbd14a5fe6b002419d127c8ec63e3062d934e058/src/input_cp2k_dft.F#L276

src/input_cp2k_dft.F
```diff
@@ -9721,4 +9729,143 @@ SUBROUTINE create_cdft_opt_section(section)
...
+                          name="DFT_PLUS_U", &
+                          description="Define the parameters for the DFT+U correction implemented in the pwdft method", &
+                          n_keywords=4, &
+                          repeats=.FALSE.)
+      NULLIFY (keyword)
+
+      CALL keyword_create(keyword, __LOCATION__, &
+                          name="_SECTION_PARAMETERS_", &
+                          description="Controls the activation of the DFT+U section", &
+                          usage="&DFT_PLUS_U ON", &
```
oschuett: How does this new `DFT_PLUS_U` section relate to the existing [`PLUS_U_METHOD`](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT.html#CP2K_INPUT.FORCE_EVAL.DFT.PLUS_U_METHOD)? Could you maybe just extend the existing keyword?

src/input_cp2k_dft.F
```diff
@@ -9721,4 +9729,143 @@ SUBROUTINE create_cdft_opt_section(section)
...
+      CALL keyword_release(keyword)
+
+      CALL keyword_create(keyword, __LOCATION__, &
+                          name="V", &
+                          description="Effective parameter V (in eV)", &
+                          repeats=.FALSE., &
+                          n_var=1, &
+                          type_of_var=real_t, &
+                          default_r_val=0.0_dp, &
+                          unit_str="au_e", &
```
oschuett, line 9805: Here the unit is `au_e`, which is good because CP2K uses almost exclusively atomic units for its inputs. However, in the description above it says `eV`.

oschuett (changes requested):

mkrack:
> &DFT_PLUS_U ON
>       SUBSPACE_METHOD "full_orthogonalization"
> &END DFT_PLUS_U
> &DFT_U_PLUS_V ON
>      & LINK
>          ATOM_PAIR 1 2
>          ANGULAR_MOMENTUM 1 2
>          ATOMIC_LEVEL 2 3
>          V 1.24532876
>      &END LINK
> &END DFT_U_PLUS_V

Do we really need two new sections`&DFT_PLUS_U` and `&DFT_U_PLUS_V`?

mtaillefumier:
> > &DFT_PLUS_U ON
> > SUBSPACE_METHOD "full_orthogonalization"
> > &END DFT_PLUS_U
> > &DFT_U_PLUS_V ON
> > & LINK
> > ATOM_PAIR 1 2
> > ANGULAR_MOMENTUM 1 2
> > ATOMIC_LEVEL 2 3
> > V 1.24532876
> > &END LINK
> > &END DFT_U_PLUS_V
> 
> Do we really need two new sections`&DFT_PLUS_U` and `&DFT_U_PLUS_V`?

I either include the U_PLUS_V in the U section or I leave them split. It does not really matter to me but we need at least one section for it to happen. Many hubbard parameters for onsite U are part of the kind section (makes sense) but SIRIUS offer more fine control of hubbard that's why there is at least one new section.

src/input_cp2k_dft.F
mtaillefumier: thanks will fix it.

src/input_cp2k_dft.F
mtaillefumier: There is no relation between the two. There is no dedicated section for PLUS_U and it is a problem.

src/input_cp2k_dft.F
mtaillefumier: Not sure what those methods are but we are not speaking about the same thing. Anything related to orthogonalization, etc.... controls how the hubbard subspace is generated but there is no analysis of the subspace. SIRIUS has complete implementation of the various hubbard corrections (including the full theory) but it can not be summarized to one word only as many parameters control the different implementations.

mtaillefumier:
> > &DFT_PLUS_U ON
> > SUBSPACE_METHOD "full_orthogonalization"
> > &END DFT_PLUS_U
> > &DFT_U_PLUS_V ON
> > & LINK
> > ATOM_PAIR 1 2
> > ANGULAR_MOMENTUM 1 2
> > ATOMIC_LEVEL 2 3
> > V 1.24532876
> > &END LINK
> > &END DFT_U_PLUS_V
> 
> Do we really need two new sections`&DFT_PLUS_U` and `&DFT_U_PLUS_V`?

I changed the hierarchy of the two sections. DFT_U_PLUS_V is now a subsection of DFT_PLUS_U.

src/input_cp2k_dft.F
mtaillefumier: should be fixed now

mkrack:
> I changed the hierarchy of the two sections. DFT_U_PLUS_V is now a subsection of DFT_PLUS_U.

Great, that makes sense. What's about just `&PLUS_V` instead of `&DFT_U_PLUS_V` 
```
&DFT_PLUS_U
  . . .
  &PLUS_V
  . . . 
  &END PLUS_V
&END DFT_PLUS_U
```
to show that `PLUS_V` is an add on to `PLUS_U` ?

Does it make sense to have also `KIND_PAIRS` instead of `ATOM_PAIRS` to reduce the number of pair definitions?

mtaillefumier:
> > I changed the hierarchy of the two sections. DFT_U_PLUS_V is now a subsection of DFT_PLUS_U.
> 
> Great, that makes sense. What's about just `&PLUS_V` instead of `&DFT_U_PLUS_V`
> 
> ```
> &DFT_PLUS_U
>   . . .
>   &PLUS_V
>   . . . 
>   &END PLUS_V
> &END DFT_PLUS_U
> ```
> 
> to show that `PLUS_V` is an add on to `PLUS_U` ?

will do.

> Does it make sense to have also `KIND_PAIRS` instead of `ATOM_PAIRS` to reduce the number of pair definitions?

unfortunately the intersite involve atoms explicitly as we can define intersite V between two atoms of the same kind.

mtaillefumier:
it is also possible to call it overlap as we are really looking at overlap as in the tight binding model.

mkrack:
> > Does it make sense to have also `KIND_PAIRS` instead of `ATOM_PAIRS` to reduce the number of pair definitions?
> 
> unfortunately the intersite involve atoms explicitly as we can define intersite V between two atoms of the same kind.

That's not necessarily a showstopper if you define `KIND_PAIR 1 1` and exclude the self-interaction within cell (0,0,0), but, maybe I overlook some implication of the +V details.

mkrack:
> it is also possible to call it overlap as we are really looking at overlap as in the tight binding model.

So there is an cutoff radius for the interaction or a threshold value for the overlap?

mtaillefumier:
> > it is also possible to call it overlap as we are really looking at overlap as in the tight binding model.
> 
> So there is an cutoff radius for the interaction or a threshold value for the overlap?

not really. It is user responsibility to define which atom interact with their neighbors. It is not computed by sirius.

mtaillefumier:
I suggested overlap because intersite hubbard V terms are actually similar to the interactions between neighbors in the tight binding limit. the V term just represent the e-e interactions.

src/input_cp2k_dft.F
oschuett: Here, I was actually only refering to the documentation of enum keywords:
```
CALL keyword_create(keyword, __LOCATION__, &
                    name="SUBSPACE_METHOD", &
                    description="Select which subspace method to use for generating the hubbard subspace", &
                    enum_desc=s2a("Use bare atomic wave functions", 
                                  "Normalize the wave functions contributing to the hubbard subspace", &
                                  "Orthogonalize the wave functions contributing to the hubbard subspace", &
                                  "Use all atomic wave functions to orthogonalize the hubbard subspace"), &
                    enum_c_vals=s2a("NONE", "NORMALIZE", "ORTHOGONALIZE", "FULL_ORTHOGONALIZATION"), &
                    usage="FULL_ORTHOGANIZATION NONE", &
                    enum_i_vals=(/0, 1, 2, 3/), &
                    default_i_val=0)         
```

src/sirius_interface.F
```diff
@@ -420,54 +425,80 @@ SUBROUTINE cp_sirius_create_env(pwdft_env)
...
+            END IF
+         END IF
+         !
       END DO
 
+      IF (ASSOCIATED(dft_plus_u_section)) THEN
+         CALL section_vals_val_get(dft_plus_u_section, keyword_name="_SECTION_PARAMETERS_", l_val=use_u_plus_v)
+         CALL sirius_set_parameters(sctx, hubbard_correction=use_u_plus_v)
+         CALL section_vals_val_get(dft_plus_u_section, keyword_name="SUBSPACE_METHOD", i_val=subspace_method)
+         IF (subspace_method .EQ. 0) THEN
```
oschuett: Instead of hard-coding integers please define constants (`PARAMETER`) and also use them for the `enum_i_vals` when declaring the keywords.

mkrack:
> not really. It is user responsibility to define which atom interact with their neighbors. It is not computed by sirius.

That sounds a bit tedious for the user unless the system is small.
My point is if the definition of `KIND_PAIR`s (which could be internally expanded in `ATOM_PAIR` lists if SIRIUS does not support atomic kinds) could simplify/shorten the input? The explicit definition of `ATOM_PAIR`s could be allowed in addition and these definitions could be processed after the `KIND_PAIR`s overwriting that definitions in case explicit exceptions or additions are needed.

mtaillefumier:
I agree but the theory itself being what it is (holding on one leg) we still have to give the full list of pairs and the value of the interaction. the kind pair (if kind means atomic species) will only work if the V term is the same for all possible atoms of the two specified kind that are separated by the same distance. It means in practice 
- calculating the first neighbors of each atoms of the unit cell
- assume the V term is the same for all atom - pairs
- only assume first neighbors. Material scientists can come up with input files that would require two different V terms for the two same atomic species (first and second neighbors).

This list is also generated automatically. I doubt anyone was generating this by hand actually.

src/sirius_interface.F
mtaillefumier: Well it means having the sirius module included in the most basic cp2k_input file. I can do this but then there will be ifdef statements.

mtaillefumier: Probably better if I do it anyway.

src/input_cp2k_dft.F
mtaillefumier: my apologies, I misunderstood the intent. Sure it is cleaner that way. Thanks.

src/sirius_interface.F
mtaillefumier: should be solved with the most recent commit

oschuett:
> This list is also generated automatically. I doubt anyone was generating this by hand actually.

So then why don't you implement the most common algorithms for generating that list? It would greatly improve the usability of this feature.

mtaillefumier:
the list  results from calculations. They are not necessarily following the old neighbor search and the coefficients are not empirical. What you are asking is implementing a big chunk of QE in CP2K. Not going to do it. Now we can simply ignore this PR or I remove the plus_v correction from it.
Update : even if I generate the neighbor list automatically, assignment of the coupling parameters is link dependent so the list is needed.

mkrack:
So, QE is generating that list of linked atoms and not SIRIUS itself and thus SIRIUS needs all these details as input?

mtaillefumier:
Indeed. it is not just a matter of searching for the neighbors etc.... It is link chemical dependent in general. That's why I can not simply do a neighbor search then assign V for each atomic species pair... it does not work that way although QE might still do some of this internally and compute the relevant V parameters.

mtaillefumier:
once these coefficients are known we can still do other calculations such lattice relaxation etc.... Although the V should depend on this as well. A full self-consistent method is out of the scope of this PR though

mkrack:
That means, for the time being, we will have to live with that explicit DFT+U+V input for SIRIUS, if we want to have CP2K work as a fully-featured input parser for SIRIUS and thus the user has to calculate these parameters for a system in advance using QE. Correct?

mtaillefumier:
Yes. They need to know what they are doing. it is not a black box. I can simply drop it entirely and only leave the plus_u section

mkrack:
Yes, that requires indeed some expert knowledge, but it would be a pity, if your effort for `+V` is lost.

mtaillefumier:
the main effort was to include it in SIRIUS and it will not be the last time that my work is put in the drawer ;). 

Do you think this thing could be of interest to CP2K users. The Plus_U is certainly a good thing to have. the U_PLUS_V however is a bit more arguable.

oschuett:
Thanks for the explanation! Now it makes sense :-)

So, how would one extract the list from QE? Maybe we could streamline the process by directly parsing QE's output?

mtaillefumier:
> Thanks for the explanation! Now it makes sense :-)
> 
> So, how would one extract the list from QE? Maybe we could streamline the process by directly parsing QE's output?

yes that's possible. Unfortunately QE input are note the clearest possible. instead of using 

atom_pair [1 2]
translation [1 0 0]

they use 
atom_pair [1 1455]

where the atom index encode the distance between the unit cells. I can probably find where they do that in QE so I reverse back to the first description. will take me some time though.

mkrack:
I suggest to focus first on getting the current changes compiled.
If CP2K does not support (cannot parse) any `+V` information, then, I am afraid, nobody is going to use that feature with SIRIUS, because QE users will just use QE.

mkrack:
> yes that's possible. Unfortunately QE input are note the clearest possible. instead of using
> 
> atom_pair [1 2] translation [1 0 0]
> 
> they use atom_pair [1 1455]
> 
> where the atom index encode the distance between the unit cells. I can probably find where they do that in QE so I reverse back to the first description. will take me some time though.

Indeed, that's not very user-friendly. It would help, if QE could optionally dump the `+V` part in CP2K input format. Maybe, that's the easiest way for the time being without reinventing the wheel.

mtaillefumier:
just to be clear. QE has all this information internally and use it is practice for the calculations. But the input files are not....

oschuett:
> QE has all this information internally and use it is practice for the calculations. 

So, how did you create the `NiO-plus_v.inp` file for this PR? Did you run a modified version of QE?

mtaillefumier:
> > QE has all this information internally and use it is practice for the calculations.
> 
> So, how did you create the `NiO-plus_v.inp` file for this PR? Did you run a modified version of QE?

SIRIUS has a debugging functionality that generates a configuration file from the data provided through the interface. So since SIRIUS is integrated in QE (not officially) we pass all relevant information to SIRIUS using the API and enable the debugging option at runtime. We then get the configuration file (in json format) with all relevant information.

oschuett:
> SIRIUS has a debugging functionality that generates a configuration file ...

Ok, got it. I guess we can all agree that this is unfortunately not a viable user journey.

Do you have any plans to improve the usability, e.g. by copying or re-implementing the relevant code from QE?
Otherwise I'd be hesitant to merge a feature that effectively can't be used.

mtaillefumier:
> > SIRIUS has a debugging functionality that generates a configuration file ...
> 
> Ok, got it. I guess we can all agree that this is unfortunately not a viable user journey.
> 
> Do you have any plans to improve the usability, e.g. by copying or re-implementing the relevant code from QE? Otherwise I'd be hesitant to merge a feature that effectively can't be used.

not in the immediate future. I would offer to remove the plus_v stuff off the PR and only merge the on-site U as CP2K has most of this already. The V-stuff will go in a drawer for the time being.

mkrack:
I just learned from Iurii that QE defines a virtual 3x3x3 supercell to build the neighbour list for the `+V` interactions. A range cutoff could be used to build that neighbour list, but this caused trouble when atoms left the defined sphere during a vc-relax run and for this reason, the neighbour list was hard-coded. The numbering of the neighbour atoms, e.g. for the O atoms in NiO, follows a specific sequence which is defined in [this loop](https://gitlab.com/QEF/q-e/-/blob/develop/PW/src/intersite_V.f90?ref_type=heads#L84-113) of the PW routine `intersite_V.f90`. QE dumps a file `HUBBARD.dat` during a DFT+U+V calculation which contains all needed information. The only problem is the assignment of the atomic indices like `1 19` for the `+V` interactions
```
V Ni-3d O-2p 1 19 0.75
```
The loop in `intersite_V.f90` can be used, however, to identify which cell indices `(nx,ny,nz)` of the image cell belong to the atom 19. In principle, CP2K has only to parse that `HUBBARD.dat` file to load all needed information for the `&DFT_PLUS_U` and `&PLUS_V` sections.

mtaillefumier:
we would need to implement a QE parser within cp2k. That's really tedious...

mkrack:
> we would need to implement a QE parser within cp2k. That's really tedious...

I do not think that we need the QE parser to parse such lines. The CP2K parser can do that as well. We need only to read the two indices and the V value and then retrieve the cell indices using the loop from `intersite_V.f90`. The orbital momentum numbers `n` and `l` can be derived from the strings like `3d` after the `-`. `U` and `V` would be keywords in the `&DFT_PLUS_U` section. Alternatively, we could parse a line per term using instead of
```
&DFT_PLUS_U
  SUBSPACE_METHOD full_ortho
  U Ni 3d  7.70
  &PLUS_V
    V Ni-3d O-2p 1 19  0.75
    V ...
  &END PLUS_V
&END DFT_PLUS_U
```
the easier to parse representation
```
&DFT_PLUS_U
  SUBSPACE_METHOD full_ortho
  U Ni 3 2  7.70
  &PLUS_V
    V Ni 3 2  O 2 1  1 19  0.75
    V ...
  &END PLUS_V
&END DFT_PLUS_U
```
That translation would be acceptable as a user effort, but I think the CP2K parser can also handle the original QE syntax.
Please, correct me in case I miss an important point.

mtaillefumier:
Let close this PR and open a new one for some small fixes for U and SIRIUS updates. Then I will open a PR for U_PLUS_V after finding the way to solve this input problem

PR #3138, opened 2023-11-30
Manual: Add keyword mentions and stars

oschuett:


oschuett:
This PR links each keyword to pages in the methods section that mention it. Furthermore, any keyword that has at least one mention is emphasized with a star in the TOC to guide users to presumably important keywords. For a good example see the [CDFT](https://ci.cp2k.org/artifacts/run-cp2k-manual-58ae2b65/manual/CP2K_INPUT/FORCE_EVAL/DFT/QS/CDFT.html) section.

oschuett:
@mkrack, following your remark I've removed the stars from the keyword list and instead set mentioned keywords bold. Here is again the [CDFT](https://ci.cp2k.org/artifacts/run-cp2k-manual-71a8f713/manual/CP2K_INPUT/FORCE_EVAL/DFT/QS/CDFT.html) section as an example.

PR #1620, opened 2021-08-09
Add Deep Potential plugin for CP2K

Cloudac7:
Deep Potential (DP) is a machine learning potential based on deep neural network, first proposed by Zhang L. et al. (Doi: [10.1016/j.cpc.2018.03.016](https://doi.org/10.1016/j.cpc.2018.03.016)) By learning from coordinates as well as energy and forces of different structures from DFT calculations, a potential describing the whole system could be frozen to a file. By reading the file, program could yield energy, forces etc. from structures, and then MD simulation with DFT accuracy could be fastened greatly. This PR add Deep Potential support for CP2K, with potential file (e.g. `graph.pb`) supplied, CP2K could run MD simulation. 

**Only DeePMD-kit >= 2.0.0 is supportted.**

To enable the DeePMD-kit feature, `--with-deepmd` as well as `--with-tfcc` flag in toolchain should not be set to `no`. The following ways installing has been added.

 - Please set `--deepmd-mode` to `gpu` if GPU support is needed, and load CUDA environment before installing.
 - From exist DeePMD-kit C++ interface, with`$deepmd_root` and `$tensorflow_root` (Described [here](https://docs.deepmodeling.org/projects/deepmd/en/latest/install.html#install-the-deepmd-kit-s-c-interface) in detail), by setting `--with-deepmd=$deepmd_root` and `--with-tfcc=$tensorflow_root`.
  - From source code and library from Github and Anaconda Cloud, by setting the both flags to `install`. For now, DeePMD-kit v2.0.0.b1 and Tensorflow C++ interface 2.3.0 will be installed, and only CUDA Toolkit 10.1 is OK from this way. 
  - From system, by adding the header files and library to where `ld` command could read directly.

The result from CP2K has been tested to be the same with which from Lammps, supportted by Deepmodeling.

PR #3145, opened 2023-12-06
Deep potential plugin for CP2K implemented from C interface

Cloudac7:
Enhanced based on #1620 
For the implementation is fully changed, I think it would be better to just open a new PR instead of former old one.

> Deep Potential (DP) is a machine learning potential based on deep neural network, first proposed by Zhang L. et al. (Doi: [10.1016/j.cpc.2018.03.016](https://doi.org/10.1016/j.cpc.2018.03.016)) By learning from coordinates as well as energy and forces of different structures from DFT calculations, a potential describing the whole system could be frozen to a file. By reading the file, program could yield energy, forces etc. from structures, and then MD simulation with DFT accuracy could be fastened greatly. This PR add Deep Potential support for CP2K, with potential file (e.g. graph.pb) supplied, CP2K could run MD simulation.

As mentioned in #1620 , it was not a good practice to introduce C++ flags in CP2K, while at that time DeePMD-kit does not supply C interface. But now everything has changed. Since v2.2.x, [DeePMD-kit C Interface](https://docs.deepmodeling.com/projects/deepmd/en/master/install/install-from-c-library.html) is directly released together with each DeePMD-kit release, making it no need to install `libtensorflow_cc` any more. In this way, C wrapper has been rewritten and now only necessary parameters are added to arch file. 

We also support download from toolchain. To enable the DeePMD-kit feature, `--with-deepmd` flag in toolchain should not be set to no. The following ways installing has been added.
- Please set `--deepmd-mode` to `gpu` if GPU support is needed, and load CUDA environment before installing.
- From exist DeePMD-kit C interface, by setting `--with-deepmd=/path/to/libdeepmd_c`.
- From source code and library from Github, by setting it to `install`. For now, DeePMD-kit v2.2.7 will be installed.
- From system, by adding the header files and library to where ld command could read directly.


src/deepmd_wrapper.F
```diff
@@ -0,0 +1,125 @@
...
+      CHARACTER(len=1, kind=C_CHAR)                      :: c_model(LEN_TRIM(model) + 1)
+      INTEGER                                            :: i, N
+
+      N = LEN_TRIM(model)
+      DO i = 1, N
+         c_model(i) = model(i:i)
+      END DO
+      c_model(N + 1) = C_NULL_CHAR
+#if(__DEEPMD)
+      pot = dp_deep_pot_c(c_model)
```
fstein93: You will run into issues if this function is (accidentally) called if CP2K was built without DeepMD.

src/deepmd_wrapper.F
```diff
@@ -0,0 +1,125 @@
...
+      INTEGER(C_INT), DIMENSION(natom), INTENT(IN) :: atype
+      REAL(C_DOUBLE), DIMENSION(9), INTENT(IN), OPTIONAL :: cell
+      REAL(C_DOUBLE), INTENT(OUT), OPTIONAL :: energy
+      REAL(C_DOUBLE), DIMENSION(natom, 3), INTENT(OUT), OPTIONAL :: force
+      REAL(C_DOUBLE), DIMENSION(9), INTENT(OUT), OPTIONAL :: virial
+      REAL(C_DOUBLE), DIMENSION(natom), INTENT(OUT), OPTIONAL :: atomic_energy
+      REAL(C_DOUBLE), DIMENSION(natom, 9), INTENT(OUT), OPTIONAL :: atomic_virial
+
+#if(__DEEPMD)
+      CALL dp_deep_pot_compute_c(dp, natom, coord, atype, cell, energy, force, virial, atomic_energy, atomic_virial)
```
fstein93, line 116: same here

src/fist_nonbond_env_types.F
```diff
@@ -542,6 +558,15 @@ SUBROUTINE fist_nonbond_env_release(fist_nonbond_env)
          CALL torch_model_release(fist_nonbond_env%allegro_data%model)
          DEALLOCATE (fist_nonbond_env%allegro_data)
       END IF
+      IF (ASSOCIATED(fist_nonbond_env%deepmd_data)) THEN
+         IF (ASSOCIATED(fist_nonbond_env%deepmd_data%force)) THEN
+            DEALLOCATE (fist_nonbond_env%deepmd_data%force)
+         END IF
+         IF (ASSOCIATED(fist_nonbond_env%deepmd_data%use_indices)) THEN
+            DEALLOCATE (fist_nonbond_env%deepmd_data%use_indices)
```
fstein93, line 567: I would refactor that into a routine in the DeepMD module

src/manybody_deepmd.F
```diff
@@ -0,0 +1,228 @@
...
+      MARK_USED(pot_deepmd)
+      MARK_USED(para_env)
+      CALL cp_abort(__LOCATION__, "In order to use DeePMD-kit you need to download "// &
+                    "and install the DeePMD-kit (check CP2K manual for details)")
+#else
+      n_atoms = SIZE(particle_set)
+      ALLOCATE (use_atom(n_atoms))
+      ALLOCATE (use_atom_type(n_atoms))
+      use_atom = .FALSE.
+      use_atom_type = 100 ! just initial the array
```
fstein93: initialize

src/manybody_deepmd.F
```diff
@@ -0,0 +1,228 @@
...
+#ifdef __DEEPMD
+      INTEGER                                  :: iat, iat_use
+      TYPE(deepmd_data_type), POINTER            :: deepmd_data
+#endif
+
+#ifndef __DEEPMD
+      MARK_USED(fist_nonbond_env)
+      MARK_USED(force)
+      MARK_USED(pv_nonbond)
+      RETURN
```
fstein93: Maybe an abort is more suitable (like in deepmd_energy_store_force_virial)

src/manybody_deepmd.F
```diff
@@ -0,0 +1,228 @@
...
+!> \param fist_nonbond_env ...
+!> \param force ...
+!> \param pv_nonbond ...
+! **************************************************************************************************
+   SUBROUTINE deepmd_add_force_virial(fist_nonbond_env, force, pv_nonbond)
+      TYPE(fist_nonbond_env_type), POINTER     :: fist_nonbond_env
+      REAL(KIND=dp)                            :: force(:, :), pv_nonbond(3, 3)
+
+      !CHARACTER(LEN=*), PARAMETER :: routineN = 'deepmd_add_force_virial'
+      !   routineP = moduleN//':'//routineN
```
fstein93: Either time this routine or remove this parameter

src/pair_potential_types.F
```diff
@@ -172,6 +175,12 @@ MODULE pair_potential_types
       REAL(KIND=dp), POINTER, DIMENSION(:)    :: rhop => NULL(), phip => NULL(), frhop => NULL()
    END TYPE eam_pot_type
 
+! **************************************************************************************************
+   TYPE deepmd_pot_type
+      CHARACTER(LEN=default_path_length)     :: deepmd_file_name
+      INTEGER                                :: atom_deepmd_type
```
fstein93: Please add suitable defaults

src/pair_potential_types.F
```diff
@@ -1697,6 +1730,63 @@ SUBROUTINE pair_potential_eam_release(eam)
...
+! **************************************************************************************************
+!> \brief Copy two DEEPMD potential type
+!> \param deepmd_source ...
+!> \param deepmd_dest ...
+!> \author Yongbin Zhuang 07.2019
+! **************************************************************************************************
+   SUBROUTINE pair_potential_deepmd_copy(deepmd_source, deepmd_dest)
+      TYPE(deepmd_pot_type), POINTER                     :: deepmd_source, deepmd_dest
+
+      IF (.NOT. ASSOCIATED(deepmd_source)) RETURN
```
fstein93, line 1754: I would rather nullify deepmd_dest

INSTALL.md
```diff
@@ -433,6 +433,21 @@ Calls to `offload_dgemm` also accept pointers on GPU or a combination of them.
 - See <https://brehm-research.de> for more information.
 -->
 
+### 2v. DeePMD-kit (optional, wider range of interaction potentials)
+
+DeePMD-kit - Deep Potential Molecular Dyanmics. Support for DeePMD-kit can be enabled via the flag
+`-D__DEEPMD`.
+
+- Please set `--deepmd-mode` to `gpu` if GPU support is needed, and load CUDA environment before
+  installing.
```
fstein93: This seems to be a reference to the toolchain. So I would rather drop it. Same with the other two points.

oschuett:
It's so nice to see DeePMD land after all :-)

This PR looks very good to me. My only concern is with the [H2O-DPLR-model.pb](https://github.com/cp2k/cp2k/pull/3145/files#diff-209519bf89724b40a3c3a1a68591eb9249b40e01e45db0a87fc2a8411d959b44) file because its 10.3 MB. Is there maybe a smaller model we could use for testing?

Cloudac7:
> It's so nice to see DeePMD land after all :-)
> 
> This PR looks very good to me. My only concern is with the [H2O-DPLR-model.pb](https://github.com/cp2k/cp2k/pull/3145/files#diff-209519bf89724b40a3c3a1a68591eb9249b40e01e45db0a87fc2a8411d959b44) file because its 10.3 MB. Is there maybe a smaller model we could use for testing?

I just replaced the model as one of 1.6 MB in [AISSquare](https://www.aissquare.com/models/detail?pageType=models&name=W-DP-HYB-model&id=84). It could be a rather small one to reproduce results just similar to calculation result for a W cell as shown in materials project.

tools/toolchain/scripts/stage6/install_deepmd.sh
```diff
@@ -0,0 +1,115 @@
...
+      mv libdeepmd_c ${pkg_install_dir}
+      write_checksums "${install_lock_file}" "${SCRIPT_DIR}/stage6/$(basename ${SCRIPT_NAME})"
+    fi
+    DEEPMD_DFLAGS="-D__DEEPMD"
+    DEEPMD_CFLAGS="-I'${pkg_install_dir}/include'"
+    if [ "$DEEPMD_MODE" == "cpu" ]; then
+      DEEPMD_LDFLAGS="-L'${pkg_install_dir}/lib' -Wl,--no-as-needed -ldeepmd_op -ldeepmd_c -Wl,-rpath='${pkg_install_dir}/lib'"
+    elif [ "$DEEPMD_MODE" == "cuda" ]; then
+      DEEPMD_LDFLAGS="-L'${pkg_install_dir}/lib' -Wl,--no-as-needed -ldeepmd_op -ldeepmd_c -ldeepmd_dyn_cudart -ldeepmd_op_cuda -Wl,-rpath='${pkg_install_dir}/lib'"
+    fi
```
njzjz: This may not work as expected. `libdeepmd_op` is dynamically loaded during runtime. Linking or not linking it, or its dependency, will not change anything.

If the CUDA library is not found or the GPU is unavailable, the program will fall back to CPU mode. There is no need to have this option.

tools/toolchain/scripts/stage6/install_deepmd.sh
```diff
@@ -0,0 +1,115 @@
...
+source "${SCRIPT_DIR}"/tool_kit.sh
+source "${SCRIPT_DIR}"/signal_trap.sh
+source "${INSTALLDIR}"/toolchain.conf
+source "${INSTALLDIR}"/toolchain.env
+
+[ -f "${BUILDDIR}/setup_deepmd" ] && rm "${BUILDDIR}/setup_deepmd"
+
+DEEPMD_LDFLAGS=''
+DEEPMD_LIBS=''
+DEEPMD_CXXFLAGS='-std=gnu++11 '
```
njzjz: This isn't required because your interface is C, but not C++.

src/common/bibliography.F
```diff
@@ -4886,6 +4886,86 @@ SUBROUTINE add_all_references()
                          "ER"), &
                          DOI="10.48550/arXiv.2306.16066")
 
+      CALL add_reference(key=Han2018, ISI_record=s2a( &
```
njzjz: Why doesn't it call `Wang2018`

fstein93: Apparently, you do not call the respective sections. Did you set up your input file correctly and asked for a Deepmd calculation?

njzjz: Sorry I don't understand your meaning. I mean the key should be last name + year instead of first name + year.

fstein93: Ah, sorry for the misunderstanding.

tools/toolchain/scripts/stage6/install_deepmd.sh
Cloudac7: Yes. I just forgot to delete it.😂

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -200,6 +202,13 @@ The --with-PKG options follow the rules:
                           Default = no
   --with-quip             Enable interface to QUIP library
                           Default = no
+  --with-deepmd           Enable interface to DeePMD-kit library.
+                          Mode of library should be selected from --deepmd-mode
+                          for compatibility with CPU or CUDA version. 
+                          Default = no
+  --with-tfcc             Enable interface to TensorFlow.
+                          Should be used together with --with-deepmd.
+                          Default = no
```
njzjz: This flag seems to be not used anymore.

oschuett:
Seems like there is a [memory leak](https://storage.googleapis.com/cp2k-ci/run-cp2k-sdbg-9920eca4_report.txt). Can you fix this in DeepMD? Otherwise you could add a [suppression](https://github.com/search?q=repo%3Acp2k%2Fcp2k%20lsan.supp&type=code).

Cloudac7:
> Seems like there is a [memory leak](https://storage.googleapis.com/cp2k-ci/run-cp2k-sdbg-9920eca4_report.txt). Can you fix this in DeepMD? Otherwise you could add a [suppression](https://github.com/search?q=repo%3Acp2k%2Fcp2k%20lsan.supp&type=code).

As discussed in deepmodeling/deepmd-kit#3045, there is memory leak in Tensorflow and Protobuf, which could not be so easily fixed, while @njzjz has fixed a possible leak in DeePMD-kit (deepmodeling/deepmd-kit#3062, thanks a lot.)

At this moment, I think just adding suppression might be OK while the leak is not so severe that might cause the program crashing.

robinzyb:
It is nice to see this interface land!

oschuett:
I'm also very happy to finally merge this. Great work!

Btw, we now have a section on [machine learning](https://manual.cp2k.org/trunk/methods/machine_learning/index.html) in our manual ;-)

njzjz:
Thanks, @Cloudac7.

DeePMD-kit also has a [documentation page](https://docs.deepmodeling.com/projects/deepmd/en/master/third-party/out-of-deepmd-kit.html) for all external interfaces.

PR #3148, opened 2023-12-08
Abort run by default on SCF convergence failure

mkrack:
The new keyword IGNORE_CONVERGENCE_FAILURE can be set to obtain the old permissive behaviour. It replaces the keyword NOTCONV_STOPALL which enforced until now the strict behaviour which is now the default.


oschuett:
Oh wow, how did you format all our test input files? Should we maybe make that a separate commit so we can later add it to [.git-blame-ignore-revs](https://github.com/cp2k/cp2k/blob/master/.git-blame-ignore-revs)?

mkrack:
> Oh wow, how did you format all our test input files?

`vi` with some macro support.

> Should we maybe make that a separate commit so we can later add it to [.git-blame-ignore-revs](https://github.com/cp2k/cp2k/blob/master/.git-blame-ignore-revs)?

No objections.

oschuett:
To separate out the actual changes you could try  `git format-patch --ignore-space-change ...`.

mkrack:
> To separate out the actual changes you could try `git format-patch --ignore-space-change ...`.

Got two patch files ... ?

oschuett:
I was thinking you could do something like this: 
1. Use `git format-patch --ignore-space-change` to save your manual modification as patch files.
2. Reset your branch.
3. Run your vim macros again.
4. Make a first commit, which contains only the formatting.
5. Apply the patch files.
6. Make a second commit, which contains the `NOTCONV_STOPALL` changes.

mkrack:
Step 3 was not an automated process, because I still edited all files by hand.

oschuett:
> Step 3 was not an automated process, because I still edited all files by hand.

Ok, I think I found a way. Please do not merge.

mkrack:
> Ok, I think I found a way. Please do not merge.

@oschuett Thanks for the patch file. The test is clean now after some minor fixes. OK to merge?

oschuett:
Yes, all good. This nicely solves a long standing usage pitfall.

src/qs_scf.F
```diff
@@ -598,11 +598,12 @@ SUBROUTINE scf_env_do_scf(scf_env, scf_control, qs_env, converged, should_stop,
          dft_control%qs_control%cdft_control%total_steps = &
          dft_control%qs_control%cdft_control%total_steps + total_steps
 
-      IF (.NOT. converged) CPWARN("SCF run NOT converged")
```
oschuett, line 601: Maybe we should keep the old warning message? I'd expect it's hard coded in many places.

PR #3149, opened 2023-12-10
Docker: Install dependencies for CMake test via Spack

oschuett:
I decided to use [Spack environments](https://spack.readthedocs.io/en/latest/environments.html) instead of a dummy package with `install --only=dependencies` because it's simpler and more explicit. Later we can add more environment files for other common configurations like CUDA.

One remaining major issue is the build time of libint. In the old toolchain we address this by [pre-generating](https://github.com/cp2k/libint-cp2k) it.

### DIFF
--- a/tools/spack/cp2k-dependencies.yaml
+++ b/tools/spack/cp2k-dependencies.yaml
@@ -1,0 +1,40 @@
+# This is a Spack Environment file for developing CP2K.
+
+spack:
+  specs:
+   - "cmake @3.22:"  # external
+   - "mpi @4"  # external
+   - "openblas @0.3.24 +fortran threads=openmp"
+   - "dbcsr@2.6.0 +openmp +mpi"
+   - "netlib-scalapack @2.2.0"
+   - "libxsmm @1.17"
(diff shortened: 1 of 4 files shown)


tools/spack/cp2k-dependencies.yaml
```diff
@@ -0,0 +1,40 @@
...
+   - "dbcsr@2.6.0 +openmp +mpi"
+   - "netlib-scalapack @2.2.0"
+   - "libxsmm @1.17"
+   - "fftw @3.3.10 +openmp"
+   - "libxc @6.2.2"
+   - "spglib @1.16.2"
+
+   # Unfortunately, ScaLAPACK 2.2.1 has not yet been packaged by Spack.
+   # https://github.com/Reference-ScaLAPACK/scalapack/tree/v2.2.1
+   # which contains https://github.com/Reference-ScaLAPACK/scalapack/pull/26
```
RMeli, line 17: I think the issue here is that there is only a `v2.2.1` tag, but [no release](https://github.com/Reference-ScaLAPACK/scalapack/releases).

PR #3152, opened 2023-12-11
Add mi300 related work

mtaillefumier:
- Recent versions of hip emit a warning when header files are referenced with the location "hipfft.h" only. It will become an compilation error in the future.
- Introduce the Mi300a architecture to the CMakeLists.txt
- introduce unified memory support
- make pretty


src/grid/hip/grid_hip_context.h
```diff
@@ -33,50 +33,80 @@ class smem_parameters;
...
+      allocated_size_ = (size__ / 16 + 1) * 16;
     }
-    current_size_ = size_;
+    current_size_ = size__;
 
-    offloadMalloc((void **)&ptr, sizeof(T) * allocated_size_);
+#ifndef CP2K_USE_UNIFIED_MEMORY
+    offloadMalloc((void **)&device_ptr_, sizeof(T) * allocated_size_);
+#else
+    hipMallocManaged((void **)&device_ptr_, sizeof(T) * allocated_size_);
```
oschuett, line 55: What's the advantage of calling `hipMallocManaged` instead of `hipMalloc` ?

tools/precommit/check_file_properties.py
```diff
@@ -60,6 +60,7 @@
     r"__T_C_G0",
     r"__YUKAWA",
     r"__cplusplus",
+    r"CP2K_USE_UNIFIED_MEMORY",
```
oschuett: Please don't try to circumvent the checks. It's a new flag that merits proper documentation in [INSTALL.md](https://github.com/cp2k/cp2k/blob/master/INSTALL.md) and [cp2k_info.F](https://github.com/cp2k/cp2k/blob/master/src/cp2k_info.F).

src/CMakeLists.txt
```diff
@@ -1608,6 +1608,7 @@ target_compile_definitions(
     $<$<COMPILE_LANGUAGE:CUDA>:__OFFLOAD_CUDA>
     $<$<BOOL:${CP2K_USE_HIP}>:__HIP_PLATFORM_AMD__
     __OFFLOAD_HIP>
+    $<$<BOOL:${CP2K_USE_UNIFIED_MEMORY}>:CP2K_USE_UNIFIED_MEMORY>
```
oschuett: Please pick a name that follows our convention for flag, e.g.  `__OFFLOAD_UNIFIED_MEMORY`.

oschuett (changes requested):
It pains me a bit to introduce these new code paths for unified memory. Are your sure it's worth it? Don't these Mi300 still have several NUMA domains? How much speed up did you obverse?

tools/precommit/check_file_properties.py
oschuett: I just realized that `__OFFLOAD_CUDA` and `__OFFLOAD_HIP` are also on the suppression list. I guess this is why you added `CP2K_USE_UNIFIED_MEMORY` as well?

I'm sorry, I should have removed those two when I converted them from internal to external flags.

src/grid/hip/grid_hip_context.h
mtaillefumier: hipMalloc allocates on GPU only which means memory is not visible on the CPU side.

tools/precommit/check_file_properties.py
mtaillefumier: does not matter. I do not want normal users to try this as they have no access to these GPUs. I only have access to them and it is already complicated enough. The PR is to help AMD work.

mtaillefumier: I will remove this flag from the exclusion list.

src/CMakeLists.txt
mtaillefumier: will do.

mtaillefumier:
I can not give numbers right now (for various reasons). It vary from case to cases

The code is not used in a normal setting for now. We can test thing on Mi250 but keeping a separate branch and constantly keeping it up to date is unpractical as some vendor may want to create docker images with the official version of cp2k. it is experimental feature if you will much like DLAF.

mtaillefumier:
other argument for real unified memory is the memory consumption is lower. It is very useful for large matrices.

alazzaro:
I can add that AMD did a great job on Mi300 to share the HBM memory CPU-GPU, so definitely unified memory is the future to avoid the mess of explicit data movement.

src/CMakeLists.txt
```diff
@@ -1601,13 +1601,15 @@ target_compile_definitions(
     $<$<BOOL:${CP2K_USE_LIBTORCH}>:__LIBTORCH>
     $<$<BOOL:${CP2K_USE_COSMA}>:__COSMA>
     $<$<BOOL:${CP2K_USE_LIBXSMM}>:__LIBXSMM>
+    $<$<BOOL:${CP2K_USE_UNIFIED_MEMORY}>:__OFFLOAD_UNIFIED_MEMORY>
```
mtaillefumier: need to remove this line

mtaillefumier:
Everything looks alright now

oschuett (approved):

mtaillefumier:
thanks.

PR #3153, opened 2023-12-11
precommit: Add input file formatter

oschuett:


oschuett:
This PR adds a [simple formatter](https://github.com/cp2k/cp2k/blob/c14d0d3d551d2ce8419eed746fa5d911dfa2fd55/tools/precommit/format_input_file.py) for input files. It only changes white space and casing. 

While it would be nice to also sort the keywords and sections this is not trivial because of CP2K's built-in preprocessor.

oschuett:
I wrote an [improved formatter](https://github.com/cp2k/cp2k/blob/640f41d5e6cf51976dededf8e2bb909b300eb149/tools/precommit/format_input_file.py) that also sorts keywords and sections alphabetically whenever possible.
I made two exception for the top level sections `&GLOBAL` and `&FORCE_EVAL` by putting them first and last respectively.

PR #3154, opened 2023-12-11
Improve Spack environment

RMeli:
Follow-up to https://github.com/cp2k/cp2k/pull/3149. While I agree that using Spack environments and have everything defined in a YAML file is great, I think installing the dependencies one at a time is not the best way forward (but happy to be proven wrong).

## Rationale

While I agree that it is more explicit to list dependencies, I think it goes against Spack's philosophy. A single root spec in the environment (CP2K in this case) determines what we want to install, and Spack should take care of the dependencies. We shouldn't be worrying about the dependencies, since they are encoded in the Spack package. If some dependencies are not what we want, it likely means that we need to update the Spack package.

For example https://github.com/cp2k/cp2k/pull/3149 sets `+openmp` for the `fftw` dependency. But this is [already taken care by the CP2K Spack package](
https://github.com/spack/spack/blob/ce09642922f27e60dc47247a12f6455bd5aeb9b8/var/spack/repos/builtin/packages/cp2k/package.py#L164). So we are basically re-encoding a constraint that is already encoded in the Spack package, and if it wasn't we would probably want to improve the CP2K Spack package.

Additionally, using the CP2K Spack package instead of installing dependencies one-by-one also allows to realize that some constraint we want are not encoded. For example, https://github.com/cp2k/cp2k/pull/3149 rightfully set `cmake@3.22:`. I realized that this constraint was missing from the CP2K package, so I added it thus improving the package for everyone: https://github.com/spack/spack/pull/41542/files.

Additionally, setting multiple specs in an environment with variants might end up in unexpected concretizations. The concretizer can ignore/turn-on/turn-off some variants automatically, if it is the only way to satisfy some constraints. Using `requires:` to add constraints on the dependencies if we really want to is much stronger.

## PR Description

This PR does the following:
* Update the CP2K package to the one in Spack upstream
  * It is useful to have a package locally to avoid circular dependencies
  * If the package matches the upstream one closely, it will be easy to port changes upstream
* Use the local CP2K Spack repo (with the local CP2K Spack recipe)
* Define CP2K as a single root spec in the environment
* Use `spack install --only=dependencies` to automatically install the dependencies for CP2K
* Separates the definition of virtual providers from their constraints

## Next steps

* Unify `spack_env_toolchain` and `spack_toolchain`
* Use [Spack build caches](https://spack.readthedocs.io/en/latest/binary_caches.html#spack-buildcache) or other means to avoid re-building dependencies each time (libint is indeed particularly nasty)
* Speed-up Spack builds by [building multiple packages concurrently](https://spack.readthedocs.io/en/latest/environments.html#installing-an-environment) and other means


oschuett:
I don't know. The environment file seems like a very clean and simple solution. What's the advantage of additionally having the `cp2k/package.py`?

>  I think it goes against Spack's philosophy.

Yes, we are actually using Spack in an unusual way here.  We don't want to install CP2K - we only want to install its dependencies. And we don't want to choose from infinite combinations of versions and variants - we only want ~3 locked environments for reproducible testing.

> Using requires: to add constraints on the dependencies if we really want to is much stronger.

Oh, I didn't know this. Of course, the CI test should fail if any package from the environment file can't be installed.

RMeli:
> The environment file seems like a very clean and simple solution. What's the advantage of additionally having the cp2k/package.py?

They are not mutually exclusive. In this PR I'm still using a Spack environment to define
some constraints (i.e. the configuration we want, which should be essentially the same as yours). Having the `cp2k/package.py` in house allows to develop the CP2K package alongside CP2K, with changes to be contributed back in Spack's develop.

> We don't want to install CP2K - we only want to install its dependencies.

Yes, but what I'm trying to say is that Spack should be in charge of dealing with the dependencies (with minimal input from us, for example the versions).

oschuett:
> Having the `cp2k/package.py` in house allows to develop the CP2K package alongside CP2K, with changes to be contributed back in Spack's develop.

Right, but the problem is that it comes at a cost because the `cp2k/package.py` adds over 1k lines of code.

Therefore, I'd like to make the design decision that the CP2K CI should only test CP2K and not its distribution channels. 

While I agree that this might be a missed opportunity, I feel that testing CP2K alone is already difficult enough.

RMeli:
> Right, but the problem is that it comes at a cost because the cp2k/package.py adds over 1k lines of code.

That is a fair point. I was working on the assumption that one day Spack will replace the toolchain (see https://github.com/cp2k/cp2k/pull/3143#issuecomment-1839089800), so it would be good to use the Spack package and make it more difficult to update CP2K without updating the Spack package.

I still think using the CP2K Spack package as root spec is the best solution, but I'll close the PR and update #3143 according to https://github.com/cp2k/cp2k/pull/3149.

PR #3167, opened 2023-12-13
fix some warning issues when NDEBUG is used

mtaillefumier:
some cosmetics removing warning when NDEBUG is used. 

Note that the use of assert should be avoided if the test is a hard test as NDEBUG will simply define the assert macro as empty.


oschuett:
On the contrary, I think we should discourage the use of `NDEBUG`. In CP2K we strive to not have asserts in tight loops. 
So while there is no performance penalty, these asserts have saved the day countless times.


Perhaps we should make this more explicit:
```
#if defined(NDEBUG)
#error "Please do not build CP2K with NDEBUG - there is no performance advantage."
#endif
```

PR #3169, opened 2023-12-13
Fix libgrpp coding convention dashboard

abussy:
Add explicit `USE` statement for libgrpp routines

### DIFF
--- a/src/libgrpp_integrals.F
+++ b/src/libgrpp_integrals.F
@@ -10,10 +10,13 @@
 ! **************************************************************************************************
 
 MODULE libgrpp_integrals
-   USE kinds,                           ONLY: dp
-   USE mathconstants,                   ONLY: pi
-   USE orbital_pointers,                ONLY: nco,&
-                                              ncoset
+   USE kinds, ONLY: dp
+   USE mathconstants, ONLY: pi
+   USE orbital_pointers, ONLY: nco, &
+                               ncoset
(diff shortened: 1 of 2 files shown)


abussy:
@mkrack I slightly modified the `install_libgrpp.sh` toolchain script in this PR, which seems to lead to trouble on some of the Dashborad. For example, on the PSI machines:

```
==================== Installing libgrpp ====================
pkg_install_dir:
/data/user/krack/rt/gnu-latest-mpich/cp2k/tools/toolchain/install/libgrpp-main-20231206
wget  --quiet https://www.cp2k.org/static/downloads/libgrpp-main-20231206.zip
libgrpp-main-20231206.zip: OK
Checksum of libgrpp-main-20231206.zip Ok
Installing from scratch into /data/user/krack/rt/gnu-latest-mpich/cp2k/tools/toolchain/install/libgrpp-main-20231206
replace libgrpp-main/.gitignore? [y]es, [n]o, [A]ll, [N]one, [r]ename:  NULL
(EOF or read error, treating as "[N]one" ...)
ERROR: (./scripts/stage3/install_libgrpp.sh, line 42) Non-zero exit code detected.
```

This seems to only happen when an installation of libgrpp is already present.

mkrack:
A [PR](https://github.com/cp2k/cp2k/pull/3171) is already running to fix that. That problem was already present from the beginning, if libgrpp is not installed from scratch.

PR #3175, opened 2023-12-15
Put libgrpp calls in OMP CRITICAL

abussy:
There is a suspicion that libgrpp is not thread safe. If the regtests stabilize after this PR, this should confirm it (this is not meant to be a fix).

### DIFF
--- a/src/core_ppl.F
+++ b/src/core_ppl.F
@@ -527,6 +527,7 @@ CONTAINS
 
                         ELSE
                            !If the local part of the potential is more complex, we need libgrpp
+!$OMP CRITICAL(type1)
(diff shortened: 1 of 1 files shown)


oschuett:
Yes, libgrpp maintains various tables in global variables e.g. for [spherical harmonics](https://github.com/aoleynichenko/libgrpp/blob/811f3e0809146b1414418f9bc4738b06325b4ec3/libgrpp/spherical_harmonics.c#L32) and grows them [on demand](https://github.com/aoleynichenko/libgrpp/blob/811f3e0809146b1414418f9bc4738b06325b4ec3/libgrpp/spherical_harmonics.c#L56).

That is certainly not thread-safe.

PR #3176, opened 2023-12-17
Print only a warning on SCF convergence failure when the run is termi…

mkrack:
…nated by an external request

### DIFF
--- a/src/qs_scf.F
+++ b/src/qs_scf.F
@@ -599,7 +599,7 @@ CONTAINS
          dft_control%qs_control%cdft_control%total_steps + total_steps
 
       IF (.NOT. converged) THEN
-         IF (scf_control%ignore_convergence_failure) THEN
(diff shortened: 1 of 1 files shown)


oschuett:
I still think #3148 is a clear improvement. Nevertheless, this is a prominent breaking change that will catch many people by surprise. Hence, I'm wondering how we could soften the blow. 

Maybe we should change the [warning message](https://github.com/cp2k/cp2k/blob/4955cfc0d24997d34b1dc70cd9db47ebb77d3489/src/qs_scf.F#L603) back to the old text `"SCF run NOT converged"`?

And perhaps for the [abort message](https://github.com/cp2k/cp2k/blob/4955cfc0d24997d34b1dc70cd9db47ebb77d3489/src/qs_scf.F#L605) we should add a hint like `"To continue the calculation regardless, please set the keyword IGNORE_CONVERGENCE_FAILURE."`?

PR #3178, opened 2023-12-18
Speedup Spack build

RMeli:
Speedup Spack build by [generating depfiles](https://spack.readthedocs.io/en/latest/environments.html#generating-depfiles-from-environments).

### DIFF
--- a/tools/docker/Dockerfile.test_cmake
+++ b/tools/docker/Dockerfile.test_cmake
@@ -56,8 +56,8 @@ RUN spack compiler find
 WORKDIR /
 COPY ./tools/spack/cp2k-dependencies.yaml .
 RUN spack env create myenv ./cp2k-dependencies.yaml
-RUN spack -e myenv env depfile -o spack-makefile
-RUN make -j32 --file=spack-makefile
+RUN spack -e myenv concretize -f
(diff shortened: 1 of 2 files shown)


RMeli:
CMake action time down from 59m in https://github.com/cp2k/cp2k/pull/3143 to 49m. (Not sure how variable timings are though...)

oschuett:
This is a nice speedup!

I wonder if `-j32` is maybe a bit too much since the individual package installations also use up to 32 jobs, don't they?

And as a nit, I notice that the output now contains weird characters:
```
[0m[91m==> 
```

RMeli:
I had to [dig into the source code](https://spack.readthedocs.io/en/latest/_modules/spack/util/cpus.html#determine_number_of_jobs) to confirm what I've been told, but Spack uses `-j16` as default (or the number of CPUs, if lower). So `-j32` is where the speedup comes from. The documentation also says the following:
> Spack respects the Make jobserver and forwards it to the build environment of packages, meaning that a single -j flag is enough to control the load, even when packages are built in parallel.

> And as a nit, I notice that the output now contains weird characters

Good catch. I think this is related to colours, I'll look into it.

RMeli:
Thanks. Looks like `SPACK_COLOR=never` didn't do the trick, I'll fix the weird characters on another PR, provably after the holidays.

PR #3179, opened 2023-12-19
Remove old Spack Dockerfiles

RMeli:
PR #3177 removed some Spack files used by some of the Docker containers. This PR removes such containers and the generating code.

@mtaillefumier, can you please have a look if this is OK?

The next steps would be to add environments replacing the CUDA and ROCm Dockerfiles.


mtaillefumier:
I am a bit puzzled here. What's the rational behind this PR. We can get a next to a full featured build with these scripts using nothing but the official recipe (which we need to test).

They can be used as template for anyone wanting to build cp2k in a container with spack. Using environment for the CI/CD can make sense (provided that the dependencies are build with cp2k requirement) but no user will use such advanced feature if they just want cp2k in a container.

RMeli:
As mentioned, PR #3177 removes files used by these Docker containers, so they should be either updated or removed. My understanding from yesterday's discussion is that user-facing containers should go to [cp2k-containers](https://github.com/cp2k/cp2k-containers).

oschuett:
Thanks, I forgot about those Dockerfiles. The next step is indeed to write environment files with GPU support and then migrate our existing tests like [Dockerfile.test_cuda_P100](https://github.com/cp2k/cp2k/blob/master/tools/docker/Dockerfile.test_cuda_P100).

> I am a bit puzzled here.

@mtaillefumier The Dockerfiles this PR removes were super valuable as a starting point. Thanks again for providing them. Ultimately, I decided to use environments for our CI instead to avoid duplicating >1k lines of code. See also the discussion in #3154. 

> the official recipe (which we need to test).

I agree that the CP2K Spack recipe needs testing. However, this should primarily be the responsibility of the Spack project. The nightly releases of our production images from the [cp2k-containers project](https://github.com/cp2k/cp2k-containers) could serve as a second line of defense.

PR #3188, opened 2023-12-22
print LDOS in units of 1/(eV*Å^2)

JWilhelm:
(LDOS is integrated in z direction)


JWilhelm:
Small comment: Regtest sdbg fails because of the slow regtest 

    QS/regtest-debug-6/ch2o_dip46.inp                                                (  34.67 sec)

This PR should not affect this regtest.

Issue #3208, opened 2024-01-09
cp2k does not return the band gap when PWDFT is used


mtaillefumier:
The band gap is not communicated to cp2k when PWDFT is used.

mtaillefumier:
fixed with PR 3211.

PR #3211, opened 2024-01-11
[PWDFT] Print the band gap and total energy

mtaillefumier:
fix for issue #3208


oschuett:
I've uploaded the new tarball. It seems there is a build error now.

mtaillefumier:
my bad, I forgot that we namespaced the options in sirius. It should be fixed now.

mtaillefumier:
thanks

PR #3221, opened 2024-01-16
Bug fix screening in hfx_derivatives

juerghutter:
Disable density matrix screening in response forces calculations in hfx_derivatives.
Change EPS_SCHWARZ_FORCES default value to 100*EPS_SCHWARZ


abussy:
It looks like this PR breaks some regtests  over different machines. I'll go ahead and update the reference values.

PR #3222, opened 2024-01-16
Indent @-directives and set them upercase

mkrack:


tools/precommit/format_input_file.py
```diff
@@ -91,7 +91,9 @@ def __init__(self, preamble: List[str], line: str) -> None:
         self.sortkey = ""  # Pre-processor lines are not sortable.
 
     def render(self, verbatim: bool) -> List[str]:
-        return self.preamble + [self.line]
+        tokens = self.line.split(" ", 1)
+        tokens[0] = tokens[0].upper()
+        return self.preamble + [" ".join(tokens)]
```
oschuett, line 96: How about this:
```suggestion
        return self.preamble + [format_line(self.line)]
```

mkrack: I intentionally didn't use `format_line`, because I want to keep the rest of the directive lines as-is, including all spaces and comments do not work on directive lines anyway.

PR #3225, opened 2024-01-17
fix overlap derivatives

tsaihsiaoyi:
Fix a bug caused the derivatives of the overlap matrix to be unable to be print.

### DIFF
--- a/src/qs_core_hamiltonian.F
+++ b/src/qs_core_hamiltonian.F
@@ -569,13 +569,13 @@ CONTAINS
       INTEGER                                            :: after, handle, i, ic, iw, output_unit
       LOGICAL                                            :: omit_headers
       TYPE(cp_logger_type), POINTER                      :: logger
-      TYPE(dbcsr_p_type), DIMENSION(:), POINTER          :: matrix_s, matrix_v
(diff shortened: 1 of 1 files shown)


fstein93:
Thank you for the contribution. Is it possible to add a tiny test which tests this feature?

tsaihsiaoyi:
With this feature or not only cause difference in output file. It will not cause CP2k crashed or mistakes in energy. It seems any inputfile with `OVERLAP` and `DERIVATIVES` existed in `DFT` `PRINT` section simultaneously will make the differences. The testsuites [H2O-none.inp](https://github.com/cp2k/cp2k/blob/master/tests/QS/regtest-gpw-1/H2O-none.inp) or [H2O-noheader.inp](https://github.com/cp2k/cp2k/blob/master/tests/QS/regtest-gpw-1/H2O-noheader.inp) are enough to reproduce this bug. The current version can only output overlap matrix. After the repair, the overlap matrix of both xyz derivatives can be output normally.

tsaihsiaoyi:
PS: Thanks to Mr. Mingyang Chen for discovering this problem and submitting it on my behalf.

PR #3234, opened 2024-01-19
RI-HFX (incl. k-points) docs/tuto

abussy:


oschuett:
Thanks for contributing documentation! Please add that new page also to the [table of content](https://github.com/cp2k/cp2k/blob/master/docs/methods/dft/index.md?plain=1#L10).

Furthermore, it seems Sphinx doesn't like those empty links: `[accepted, not yet published.](<>)`.

oschuett:
Btw, you can preview the page [here](https://ci.cp2k.org/artifacts/run-cp2k-manual-51466c98/manual/methods/dft/ri_hfx.html).

abussy:
Thanks for the pointers!

oschuett:
I'm contemplating to create a dedicated subsection for HFX, similar to the one for [Optical Spectroscopy](https://manual.cp2k.org/trunk/methods/properties/optical.html).

- Hartree-Fock eXchange
  - HFX with ADMM
  - HFX with Resolution of Identity
  - HFX-RI with K-Points
  - ...

What do you think?

abussy:
Why not, I have no strong opinion about this, I just happend to bundle Gamma-point and k-point RI-HFX together because they rely on similar theory and machinery.

Issue #3228, opened 2024-01-18
Build error for SIRIUS 7.5.2 using GCC 9.3.0


mkrack:
The build of libsirius fails on Piz Daint using GCC 9.3.0. See the corresponding [regression test](https://dashboard.cp2k.org/archive/cray-xc50-gnu-psmp/index.html) for details.

mkrack:
@mtaillefumier could you have a look at this problem?

mtaillefumier:
problem is here : /store/psi/psi01/mkrack/rt/gpu-gnu-psmp/cp2k/tools/toolchain/build/SIRIUS-7.4.3/src/api/sirius_api.cpp:3062

mtaillefumier:
wrong version of SIRIUS.

mtaillefumier:
We won't be back porting these changes though so the linked version of sirius should be updated.

mkrack:
> problem is here : /store/psi/psi01/mkrack/rt/gpu-gnu-psmp/cp2k/tools/toolchain/build/SIRIUS-7.4.3/src/api/sirius_api.cpp:3062

That file does not exist. The current SIRIUS build for daint-gpu uses version 7.5.2. Therefore, it is not clear to me what you mean.

mtaillefumier:
CP2K is linked with the wrong version of SIRIUS. the error you see is triggered by SIRIUS itself. So the file might not exist but the binary still knows about it. The CP2K binary is linked with the wrong version of SIRIUS. 7.5.2 is build but the older version is still in use.

mkrack:
The problem is not related to CP2K. The compilation of SIRIUS 7.5.2 with GCC 9.3.0 itself is already failing:
```
==================== Installing SIRIUS ====================
wget  --quiet https://www.cp2k.org/static/downloads/SIRIUS-7.5.2.tar.gz
SIRIUS-7.5.2.tar.gz: OK
Checksum of SIRIUS-7.5.2.tar.gz Ok
Installing from scratch into /store/psi/psi01/mkrack/rt/gpu-gnu-psmp/cp2k/tools/toolchain/install/sirius-7.5.2
      |                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #3237, opened 2024-01-21
update CP2K arch file with deepmd for Piz Daint, CSCS, GPU partition

robinzyb:
I am currently working at EPFL and am able to access pizdaint with P100 GPU

here is the updated arch file with deepmd. 

 To enable deepmd installation 

append `--with-deepmd`  to `./install_cp2k_toolchain.sh` command at line 52
and uncomment `#USE_DEEPMD     := 2.2.7` at line 83


mkrack:
Thanks for contributing. Unfortunately, the test and build for Piz Daint (Cray XC50, daint-gpu) is currently broken because of a compilation error for SIRIUS 7.5.2 (see #3228).

PR #3243, opened 2024-01-23
MPI: Remove CONTIGUOUS from mp_free_mem to work around Flang bug

oschuett:
@fstein93 do you think this is reasonable? I'd expect that `free` fails in any case if the wrong pointer is passed.

### DIFF
--- a/src/mpiwrap/message_passing.fypp
+++ b/src/mpiwrap/message_passing.fypp
@@ -4711,8 +4711,8 @@
 !> \param[out] stat      (optional) allocation status result
 ! **************************************************************************************************
    SUBROUTINE mp_free_mem_${nametype1}$ (DATA, stat)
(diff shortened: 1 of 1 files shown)


fstein93:
I don't see any issues. The standard does not require the `CONTIGUOUS` keyword. As soon as the respected versions are not used anymore, we can reintroduce it as C always (?) returns pointers to contiguous chunks of memory and as if you have said, MPI will potentially complain if it receives incorrect pointers.

fstein93 (approved):

alazzaro:
out of curiosity, which Flang is this? Could you give a bit of details?

oschuett:
This was reported by AMD for their Fortran compiler, but the underlying issue is https://github.com/flang-compiler/flang/issues/673#issuecomment-763918602.

PR #3246, opened 2024-01-25
Bump Spack and DLA-Future in CI

RMeli:
Bump Spack to more recent commit and DLA-Future to `dla-future@0.4.0`.

### DIFF
--- a/tools/docker/Dockerfile.test_cmake
+++ b/tools/docker/Dockerfile.test_cmake
@@ -44,7 +44,7 @@ RUN apt-get update -qq && apt-get install -qq --no-install-recommends \
 WORKDIR /opt/spack
 RUN git init --quiet && \
     git remote add origin https://github.com/spack/spack.git && \
-    git fetch --quiet --depth 1 origin 8bcb1f8766ffbf097ec685bdafad53bc8b6e9e30 && \
+    git fetch --quiet --depth 1 origin 2721b4c10d27afde281ea0cdb0480a7343ce0863 && \
     git checkout --quiet FETCH_HEAD
 ENV PATH="/opt/spack/bin:${PATH}"
 
--- a/tools/docker/Dockerfile.test_cmake
+++ b/tools/docker/Dockerfile.test_cmake
@@ -51,7 +51,6 @@ ENV PATH="/opt/spack/bin:${PATH}"
 # Find all external packages and compilers.
 RUN spack compiler find
 RUN spack external find --all --not-buildable
-RUN cat ~/.spack/packages.yaml && sed -i 's/gcc:/gcc-runtime:/g' ~/.spack/packages.yaml && cat ~/.spack/packages.yaml
 
 # Install CP2K's dependencies via Spack.
 WORKDIR /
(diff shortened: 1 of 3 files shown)


oschuett:
What's the state of this PR? Anything I could do to help?

RMeli:
Ah. I did not realize this was failing. I'll look into it.

oschuett:
How should we proceed with this PR?

RMeli:
Sorry, I am working on lifting DLA-Future's Fortran interface out of CP2K into a standalone package so that we can have a slimmer integration and also add the standard and generalised eigensolver for complex types.

I was hoping to tackle this PR with the new integration, however it's taking a bit longer than planned. I'll try to bump the Spack version again and then look more closely into it if the issue persists.

tools/docker/Dockerfile.test_cmake
```diff
@@ -44,7 +44,7 @@ RUN apt-get update -qq && apt-get install -qq --no-install-recommends \
 WORKDIR /opt/spack
 RUN git init --quiet && \
     git remote add origin https://github.com/spack/spack.git && \
-    git fetch --quiet --depth 1 origin 8bcb1f8766ffbf097ec685bdafad53bc8b6e9e30 && \
+    git fetch --quiet --depth 1 origin 2721b4c10d27afde281ea0cdb0480a7343ce0863 && \
```
RMeli:
```suggestion
    git fetch --quiet --depth 1 origin 384e1983041094c8487d394f59dfe8a6848d60f2 && \
```

tools/docker/generate_dockerfiles.py
```diff
@@ -779,7 +779,7 @@ def spack_env_toolchain() -> str:
 WORKDIR /opt/spack
 RUN git init --quiet && \
     git remote add origin https://github.com/spack/spack.git && \
-    git fetch --quiet --depth 1 origin 8bcb1f8766ffbf097ec685bdafad53bc8b6e9e30 && \
+    git fetch --quiet --depth 1 origin 2721b4c10d27afde281ea0cdb0480a7343ce0863 && \
```
RMeli:
```suggestion
    git fetch --quiet --depth 1 origin 384e1983041094c8487d394f59dfe8a6848d60f2 && \
```

RMeli:
This is more annoying than I thought... I'll work locally with the Docker container and push a solution when it's ready.

RMeli:
I confirmed my suspicion that the problem appears after the introduction of the `gcc-runtime` package. On install, this packages copies the binaries internally, so the warning is spurious.

I think I never saw this locally because I'm using `spack build-env` instead of activating an environment with the default view. It is still not clear to me how to best fix this within the constraints discussed in https://github.com/cp2k/cp2k/pull/3149 and https://github.com/cp2k/cp2k/pull/3154. I'll have another look at this next week.

In the meantime, an ugly workaround is to remove the `libgomp.so` copy made by `gcc-runtime`.

RMeli:
Our local Spack aficionado @albestro helped me to investigate this issue, and he suggested to exclude `gcc-runtime` from the environment view. This prevents `libgomp.so` to be added to the view and therefore CMake only finds the original library in `/usr/lib/gcc/x86_64-linux-gnu/11`.

oschuett:
It seems odd that Spack still builds its own gcc-runtime after it picked up the external gcc. In any case, I'm glad you found a good workaround.

albestro:
> It seems odd that Spack still builds its own gcc-runtime after it picked up the external gcc. In any case, I'm glad you found a good workaround.

Just for clarity, spack is not "building" a `gcc-runtime`, indeed it is copying the compiler libraries needed for runtime from the compiler you are using. This is useful in case the compiler is removed, but whatever has been built will still work correctly.

The warning you get in the CMake is a legit warning, because it is not checking that those libraries are actually the same, but we know they are.

PR #3257, opened 2024-01-31
Refactor pw_type

fstein93:
This PR is the second attempt of #2987 . The changes are similar to this PR.
General changes:
- Splits the pw_type into separate types depending on the underlying grid (**r**eal/**c**omplex, 1d/3d) and the space (**r**eal**s**pace/reciprocal space(=**g**-**s**pace)) in which a function lives. The nomenclature is `pw_<r|c><1|3>d_<rs|gs>`. This change is motivated by the fact that the underlying grid or the space never change during the lifetime of a given pw_type.
- This separation revealed some bugs in routines which were never found by users or regtests. These are fixed with this PR.
- This PR requires that the several routines now require separate interfaces depending on the grids. This is most pronounced in pw_methods implementing general working routines for PW grids and pw_poisson_solve (see below). I often implemented the different copies using Fypp templates due the same general ideas in code structure.
- These changes allowed me to promote several runtime errors to compiletime errors.
- The routine pw_transfer was converted to a simple interface of the respective pw_copy/pw_scatter/pw_gather/fft_wrap_pw1pw2 routines.
- These changes reduce the memory overhead of the grids a bit. Potentially, some compilation errors may become a bit cryptic due to the many interfaces involved.
More information on pw_poisson_solve:
- Because almost all different combinations of grids are used in the code, I had to implement all combinations of optional parameters explicitly to prevent ambiguous interfaces which would have been present if I had optional parameters with all different combinations of the new plane wave types. Suitable simplifications were applied where possible.
- I restricted the combinations to combinations of real-valued realspace 3D grids (pw_r3d_rs_type, formerly pw_type with set REALDATA3D and REALSPACE) and complex-valued reciprocalspace 1D grids (pw_c1d_gs_type, formerly pw_type with set COMPLEXDATA1D and RECIPROCALSPACE). I enforced the same types for densities and auxiliary densities and the pw_c1d_gs_type for the arguments rho_core and greenfn because these are the only one in use in the code.
- Certain combinations were only partially implemented leading to incorrect values if an argument was absent although never needed (for instance in energy_corrections.F). I fixed that behavior and the routines should behave as expected independently whether the aux_density argument is present or not.
- The energy calculation could have failed for certain combinations of densities and potentials.

Changes were already squashed to several chunks. Probably the formatting changes several times (due to addition/removal of temporary routines required for the development).
These changes were tested locally and on the GPU partition of Noctua2.


fstein93:
Short update:
I could confirm the issue locally. It is related to OpenMP because the test results are correct if I employ only a single thread. Tracking down the test to the individual change, I found the deviations started when I removed the branches checking for COMPLEXDATA3D for which I am sure that they were not called anymore at that point. So, it could be a bug in the compiler. I am currently identifying the routine providing incorrect results and I am investigating the differences in the parallelization of the Intel compiler.

PR #3270, opened 2024-02-14
Use computed eigenvalues instead of required eigenvalues to determine ScaLAPACK fallback

RMeli:
Currently, input variables `DLAF_NEIGVEC_MIN` and `ELPA_NEIGVEC_MIN` use the number of requested eigenvalues instead of the number of computed eigenvalues to determine when to trigger the ScaLAPACK fallback. This PR changes the behavior to use the number of computed eigenvalues.

This is important for DLA-Future, where `DLAF_NEIGVEC_MIN` is currently set reasonably high. I changed it for all libraries for consistency.

### DIFF
--- a/src/fm/cp_fm_diag.F
+++ b/src/fm/cp_fm_diag.F
@@ -221,7 +221,7 @@ CONTAINS
          CALL cp_fm_syevd(matrix, eigenvectors, eigenvalues, info)
 
       ELSE IF (diag_type == FM_DIAG_TYPE_ELPA) THEN
-         IF (SIZE(eigenvalues) < elpa_neigvec_min) THEN
+         IF (matrix%matrix_struct%nrow_global < elpa_neigvec_min) THEN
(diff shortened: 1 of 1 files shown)


RMeli:
Looks like this broke CI: [CRAY-XC50-GNU](https://dashboard.cp2k.org/archive/cray-xc50-gnu-psmp/index.html). I'll look into it.

RMeli:
In the failing test it appears that both `SIZE(eigenvalues)` and  `matrix%matrix_struct%nrow_global` are set to `32`, so it should not have been affected by the changes here.

mkrack:
The [CRAY-XC50-GNU](https://dashboard.cp2k.org/archive/cray-xc50-gnu-psmp/index.html) test is fine now. Piz Daint shows sometimes a "noisy" behaviour or instabilities like occasional bus errors. So, this PR did not cause any problem so far.

RMeli:
@mkrack, thanks for checking.

PR #3274, opened 2024-02-16
Fix an error in the output of MO for ROKS

IlyaFed:
Added make_mo_eig(), which returns orbitals from OT space. Limited it to nspin=1. Now everything works consistently with CPMD code.

More details in the Google Group discussion:
https://groups.google.com/g/cp2k/c/yJPJS2vUkBI


IlyaFed:
Here I also need to add make_mo_eig() in cases where Wannier Сenters are used without outputting orbitals. I'll send it a little later.

oschuett:
> I'll send it a little later.

Ok, then I'll hold off merging this PR for now.

oschuett:
Any update on this?

IlyaFed:
After tests, it turned out that the make_mo_eig does not work correctly. As I understand it, this is due to the fact that (for some reason?) it searches the eigen vectors of the Kohn-Sham matrix and rotate MO to it, which does not work correctly in the case of ROKS. At the moment, I can’t fully understand why make_mo_eig is needed and how to fix it? But without it, the print orbitals and Wannier centers for ROKS does not work correctly...

oschuett:
@juerghutter what do you think?

juerghutter:
Hi
to give an fully informed answer I would have to have a deeper look into the
code. There are just too many ROKS options and I'm not sure they all need
the same treatment.
In general, when there is no single determinant reference function it gets
complicated (e.g. low spin ROKS). In this case there is no Hamiltonian and
the code works through energy gradients (that's why you need OT).
This means one should not use the Kohn-Sham matrix but go another
route in order to get meaningful eigenvalues.
regards
Juerg

oschuett:
@IlyaFed how do you want to proceed? Should I merge this PR as is or wait a little longer?

Another option would to block off the broken cases with `CPWARN` or `CPABORT`.

IlyaFed:
For now I think we should put CPABORT into PRINT_MO_CUBES and WANNIER in LOW SPIN ROKS, but I would like to improve this part further. Unfortunately, I somehow merged the new version incorrectly, can you tell me how to fix this? In general, it is enough to replace cp_warn with cp_abort on line 465 in qs_scf_post_gpw.F.

oschuett:
> Unfortunately, I somehow merged the new version incorrectly, can you tell me how to fix this? 

Sorry, I just saw your question. The simplest way to recover is usually to [start over](https://xkcd.com/1597/):

1. Make a backup.
2. `git fetch origin`
3. `git reset --hard origin/master`
4. Copy your changes back.

PR #3279, opened 2024-02-23
Manual: Add pages about X-Ray Spectroscopy

oschuett:


oschuett:
Preview: https://ci.cp2k.org/artifacts/run-cp2k-manual-2ccd1f77/manual/methods/properties/x-ray/index.html

@marci73, do you think this page structure makes sense?

marci73:
Hi Ole,

Well there is also another approach implemented in CP2K
https://www.cp2k.org/_media/exercises:2019_conexs_newcastle:conexs_fernanda.pdf

On the other hand I don’t think that with the cp2k manual we have to provide a computational chemistry book
It would be sufficient to explain the keywords and give examples on ow to use them.

Best
Marcella

From: Ole Schütt ***@***.***>
Date: Friday, 23 February 2024 at 11:41
To: cp2k/cp2k ***@***.***>
Cc: Marcella Mauri-Iannuzzi ***@***.***>, Mention ***@***.***>
Subject: Re: [cp2k/cp2k] Manual: Add pages about X-Ray Spectroscopy (PR #3279)

Preview: https://ci.cp2k.org/artifacts/run-cp2k-manual-2ccd1f77/manual/methods/properties/x-ray/index.html

@marci73<https://github.com/marci73>, do you think this page structure makes sense?

oschuett:
Thanks for the pointer! I've now added a placeholder page for the ~~Delta-DFT~~ Delta-SCF method. While we don't need to write an entire text book, I think it's important that we list all available methods  and explain their pros and cons.

PR #3292, opened 2024-02-29
Various smaller fixes or improvements

hfp:
* Avoid calling offloadEnsureMallocHeapSize (if offloading the GRID code is disabled).
* Suppress error message when using "which" for an unknown command (exts/build_dbcsr).
* Improved finding OpenCL header file (exts/build_dbcsr).

### DIFF
--- a/exts/build_dbcsr/Makefile
+++ b/exts/build_dbcsr/Makefile
@@ -299,22 +299,26 @@ OPENCL_DEFAULT := $(wildcard $(LIBSMM_ACC_ABS_DIR)/../opencl/smm/tune_multiply.c
 OPENCL_WITHGPU := $(wildcard $(LIBSMM_ACC_ABS_DIR)/../opencl/smm/params/tune_multiply_$(GPUVER).csv)
 OPENCL_PARAMS := $(if $(OPENCL_WITHGPU),$(OPENCL_WITHGPU),$(OPENCL_DEFAULT))
 OPENCL_PARDEP := $(if $(ACC_MAKEDEP),$(shell $(ACC_MAKEDEP) $(LIBSMM_ACC_ABS_DIR)/../opencl/smm/.with_gpu $(GPUVER)))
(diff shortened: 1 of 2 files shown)


src/grid/ref/grid_ref_collint.h
```diff
@@ -66,6 +66,7 @@ ortho_cx_to_grid(const int lp, const int k1, const int k2, const int j1,
 #if (GRID_DO_COLLOCATE)
     // collocate
     double reg[4] = {0.0, 0.0, 0.0, 0.0};
+#pragma omp simd reduction(+ : reg)
```
oschuett: The REF backend is now indeed only a reference implementation. The default CPU backend is now in:

https://github.com/cp2k/cp2k/tree/master/src/grid/cpu

docs/conf.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
```
oschuett: I don't think this file should have a shebang because it's not an executable.
Same for `tools/prettify/prettify_cp2k/__init__.py`.

src/grid/ref/grid_ref_collint.h
hfp: You think it should not carry any fancy directive? Indeed, I am collecting changes on my end (older and more recent).

docs/conf.py
hfp: I think this came up with warnings during precommit, but I can be mistaken.

hfp:
I will remove the two changes above from this PR, i.e.:

* Missing Python Shebang (docs/conf.py, prettify_cp2k).
* Some OpenMP SIMD directives (GRID reference code).

src/grid/ref/grid_ref_collint.h
oschuett: I'd like to keep the reference backend as simple as possible.

However, your fanzy directives are very much appreciated in the optimized CPU backend ;-)

https://github.com/cp2k/cp2k/blob/39d9274f51dfd4eb0249496921a785bb611b0c44/src/grid/cpu/grid_cpu_collint.h#L49

https://github.com/cp2k/cp2k/blob/39d9274f51dfd4eb0249496921a785bb611b0c44/src/grid/cpu/grid_cpu_collint.h#L64

hfp: Ahh, interesting point. Will do this asap.

PR #3293, opened 2024-02-29
Make use of SIMD directives

hfp:
- Compilers allow to toggle OpenMP SIMD separate from targeting multicore.
- Note: -fopenmp may not enable SIMD directives (-fopenmp-simd).


hfp:
See also https://github.com/cp2k/cp2k/pull/3292#discussion_r1507306915

PR #3321, opened 2024-03-15
DBM: introduced OpenCL support

hfp:
* Assembling batches now tracks the MxNxK ranges of all shapes in a batch.
* OpenCL kernel uses maximum M-extent (and other extents for verbose output).
* Marking up CUDA-kernel (to be code-shared with OpenCL) is certainly possible but leads to convoluted GPU-kernel source.
* Updated precommit to consider *.cl sources.


hfp:
( Adding support for GNU Make and CMake is subject of a follow-up PR. )

PR #3331, opened 2024-03-22
Enable Spack build cache

RMeli:
Try Spack build cache with `elpa` and `libint`.


RMeli:
Since we concretize multiple root specs together, high-level specs are not pulled from the Spack build cache. The build cache would only benefit low-level specs in this case, but most of them are pulled in as external packages anyways.

PR #3337, opened 2024-03-26
improve element comparison in reftraj

BelizSertcan:
This PR addresses an issue where the REFTRAJ was raising errors due to potential aliasing in trajectory file that is created with [PRINT_ATOM_KIND](https://manual.cp2k.org/trunk/CP2K_INPUT/MOTION/PRINT/TRAJECTORY.html#CP2K_INPUT.MOTION.PRINT.TRAJECTORY.PRINT_ATOM_KIND) set to True. Previously, only the element symbol of the initial configuration was compared to the reftraj trajectory. Now the element kind of the reference is also compared.

The tests run smoothly, as well as my own calculations. However, there can be some unforeseen cases that are not covered by tests where this may fail.


BelizSertcan:
How can I solve this slow test failure? It is not a file I created or modified.

oschuett:
Don't worry, this is a false positive. I just restarted the test.

BelizSertcan:
why does it happen?

mkrack:
> why does it happen?

Mostly bad luck, I guess. It also happens to me often, e.g. today. Overall it is a kind of annoying.

oschuett:
The runtimes of our tests are inherently noisy. The two main reasons are the parallelism in our [do_regtest.py](https://github.com/cp2k/cp2k/blob/master/tests/do_regtest.py) script and other cloud tenants. However, we're using a very generous criterion of `slow_test_threshold = 2 * percentile(timings, 0.95)`. So in principle all tests should be able to stay far away from the threshold.

> Overall it is a kind of annoying.

True. Maybe we should re-run directories with slow tests one more time before flagging them?

PR #3353, opened 2024-04-15
AS: Perform second contraction step using FM instead of DBCSR

fstein93:
This change significantly improves the performance of the half-transformed GPW method (Si 1x1x1: factor of 3 in total runtime).
I added supported for the truncated Coulomb potential.


fstein93:
This PR adds the truncated Coulomb potential to the active space model and replaces the current DBCSR-based second contraction step of the half-transformed GPW integration with a FM-based implementation. This reduces the computational costs significantly.

PR #3360, opened 2024-04-20
Slightly revised build_3c_integrals

hfp:
* Demoted max_contraction_i to scalar.

### DIFF
--- a/src/qs_tensors.F
+++ b/src/qs_tensors.F
@@ -2879,7 +2879,7 @@ CONTAINS
             END IF
 
             sp = SHAPE(block_t)
-            sp = sp([2, 3, 1])
+            sp([2, 3, 1]) = sp ! sp = sp([2, 3, 1]) performs worse
 
             CALL dbt_put_block(t3c(jcell, kcell), blk_idx, sp, &
                                RESHAPE(block_t, SHAPE=sp, ORDER=[2, 3, 1]), summation=.TRUE.)
--- a/src/qs_tensors.F
+++ b/src/qs_tensors.F
@@ -2806,8 +2806,8 @@ CONTAINS
                   rj = rij ! ri + rij
                   rk = rik ! ri + rik
 
-                  ALLOCATE (sijk(ncoj, ncok, ncoi))
                   IF (op_pos_prv == 1) THEN
+                     ALLOCATE (sijk(ncoj, ncok, ncoi))
                      sijk(:, :, :) = 0.0_dp
                      CALL eri_3center(sijk, &
                                       lmin_j(jset), lmax_j(jset), npgfj(jset), zetj(:, jset), rpgf_j(:, jset), rj, &
--- a/src/qs_tensors.F
+++ b/src/qs_tensors.F
@@ -2409,10 +2409,10 @@ CONTAINS
                                                             found, skip
       REAL(KIND=dp)                                      :: dij, dik, djk, dr_ij, dr_ik, dr_jk, &
                                                             kind_radius_i, kind_radius_j, &
-                                                            kind_radius_k, prefac, sijk_ext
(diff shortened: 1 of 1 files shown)


hfp:
CI seems totally crazy like:

> Compiling cp2k... done.
./exe/local/cp2k.sdbg: Relink `/opt/cp2k-toolchain/install/libdeepmd_c-2.2.7/lib/libtensorflow_framework.so.2' with `/lib/x86_64-linux-gnu/libz.so.1' for IFUNC symbol `crc32_z'
2024-04-22 12:41:31.298309: I tensorflow/tsl/cuda/cudart_stub.cc:28] Could not find cuda drivers on your machine, GPU will not be used.
2024-04-22 12:41:31.448291: E tensorflow/compiler/xla/stream_executor/cuda/cuda_dnn.cc:9342] Unable to register cuDNN factory: Attempting to register factory for plugin cuDNN when one has already been registered
2024-04-22 12:41:31.448336: E tensorflow/compiler/xla/stream_executor/cuda/cuda_fft.cc:609] Unable to register cuFFT factory: Attempting to register factory for plugin cuFFT when one has already been registered
2024-04-22 12:41:31.448360: E tensorflow/compiler/xla/stream_executor/cuda/cuda_blas.cc:1518] Unable to register cuBLAS factory: Attempting to register factory for plugin cuBLAS when one has already been registered
DeePMD-kit: Cannot find libcudart.so.11.0
implib-gen: libcudart.so.11.0: failed to load library 'libcudart.so.11.0' via callback 'DP_cudart_dlopen'
implib-gen: libcudart.so.11.0: failed to resolve symbol '__cudaRegisterFatBinary', library failed to load
implib-gen: libcudart.so.11.0: failed to load library 'libcudart.so.11.0' via callback 'DP_cudart_dlopen'
implib-gen: libcudart.so.11.0: failed to resolve symbol '__cudaRegisterFunction', library failed to load
implib-gen: libcudart.so.11.0: failed to load library 'libcudart.so.11.0' via callback 'DP_cudart_dlopen'
implib-gen: libcudart.so.11.0: failed to resolve symbol '__cudaRegisterFatBinaryEnd', library failed to load
implib-gen: libcudart.so.11.0: failed to load library 'libcudart.so.11.0' via callback 'DP_cudart_dlopen'
implib-gen: libcudart.so.11.0: failed to resolve symbol '__cudaUnregisterFatBinary', library failed to load

The `sdbg` test should have nothing to do with CUDA.

hfp:
I am worried about loading Python packages on demand or without any control of (indirect) dependencies; well cloud... This seems now to load proprietary bits "as if normal" aka CUDA everywhere.

hfp:
Note: similar changes can be applied to `build_?c_integrals`, etc.

oschuett:
> I am worried about loading Python packages on demand or without any control of (indirect) dependencies; well cloud... This seems now to load proprietary bits "as if normal" aka CUDA everywhere.

I don't think this is loading Python code. I belive it's loading DeepMD as a shared library, which is then looking for CUDA on its own. Nevertheless, I agree that this is not a great user experience.

What's a bit worrying though is that we don't verify the checksum of the tarball:
https://github.com/cp2k/cp2k/blob/adfb00c71eb664bd89deb2b7dc31bb9a093bc8a2/tools/toolchain/scripts/stage6/install_deepmd.sh#L41

hfp:
Thank you for the reply, insightful as always! For "the time being" it's probably fine. I mean, for inference it has yet to be shown what's "inevitable" and which dependencies are actually necessary to let people benefit.

oschuett:
Apparently [DeePMD v3](https://github.com/deepmodeling/deepmd-kit/releases/tag/v3.0.0a0) adds support for PyTorch. Hopefully we can then drop the dependency on Tensorflow.

PR #3363, opened 2024-04-23
Two new subroutines which give correct determinant signs even in para…

andsinya:
…llel environment with any number of ranks and threads

1) cp_fm_det computes the determinant of a real square matrix 2) cp_cfm_det computes the determinant of a complex square matrix Both routines make a copy of the input matrix, thus it is NOT modified in any way and can be safely used after. Also replaced the calls in two subroutines where determinant sign matters: calc_et_coupling and qs_moment_berry_phase.

### DIFF
--- a/src/fm/cp_cfm_basic_linalg.F
+++ b/src/fm/cp_cfm_basic_linalg.F
@@ -16,7 +16,11 @@
 MODULE cp_cfm_basic_linalg
    USE cp_blacs_env,                    ONLY: cp_blacs_env_type
    USE cp_cfm_types,                    ONLY: cp_cfm_get_info,&
-                                              cp_cfm_type
+                                              cp_cfm_type,&
+                                              cp_cfm_to_cfm,&
+                                              cp_cfm_create,&
+                                              cp_cfm_release
+   USE cp_log_handling,                 ONLY: cp_to_string
    USE cp_fm_struct,                    ONLY: cp_fm_struct_equivalent
    USE cp_fm_types,                     ONLY: cp_fm_type
    USE kahan_sum,                       ONLY: accurate_dot_product
@@ -49,7 +53,8 @@ MODULE cp_cfm_basic_linalg
              cp_cfm_triangular_multiply, &
              cp_cfm_rot_rows, &
              cp_cfm_rot_cols, &
-             cp_cfm_cholesky_invert
+             cp_cfm_cholesky_invert, &
+             cp_cfm_det ! determinant of a complex matrix with correct sign
 
    REAL(kind=dp), EXTERNAL :: zlange, pzlange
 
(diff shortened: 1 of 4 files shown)


mkrack:
@andsinya Thanks for contributing that fix.

src/fm/cp_cfm_basic_linalg.F
```diff
@@ -61,6 +66,87 @@ MODULE cp_cfm_basic_linalg
...
+      ncol_local = matrix_lu%matrix_struct%ncol_locals(mypcol)
+
+      DO irow_local = 1, nrow_local
+         i = indxl2g(irow_local, nrow_block, myprow, matrix_lu%matrix_struct%first_p_pos(1), nprow)
+         DO icol_local = 1, ncol_local
+            j = indxl2g(icol_local, ncol_block, mypcol, matrix_lu%matrix_struct%first_p_pos(2), npcol)
+            IF (i == j) diag(i) = matrix_lu%local_data(irow_local, icol_local)
+         END DO
+      END DO
+      CALL matrix_lu%matrix_struct%para_env%sum(diag)
```
fstein93, line 125: I think that there is no need to exchange the diagonal elements. Just multiply them in some container (let's say `determinant`) and exchange it using something like `para_env%prod(determinant)`.

src/fm/cp_cfm_basic_linalg.F
```diff
@@ -61,6 +66,87 @@ MODULE cp_cfm_basic_linalg
...
+         IF (ipivot(irow_local) /= i) P = P + 1
+      END DO
+      CALL matrix_lu%matrix_struct%para_env%sum(P)
+      ! very important fix
+      P = P/npcol
+#else
+      CALL zgetrf(n, n, a(1, 1), n, ipivot, info)
+      DO i = 1, n
+         diag(i) = matrix_lu%local_data(i, i)
+      END DO
```
fstein93, line 138: same here

src/fm/cp_fm_basic_linalg.F
```diff
@@ -88,6 +89,71 @@ MODULE cp_fm_basic_linalg
...
+      ALLOCATE (ipivot(n))
+      ipivot(:) = 0
+      P = 0
+      ALLOCATE (diag(n))
+      diag(:) = 0.0_dp
+#if defined(__SCALAPACK)
+      ! Use LU decomposition
+      desca(:) = matrix_lu%matrix_struct%descriptor(:)
+      CALL pdgetrf(n, n, a, 1, 1, desca, ipivot, info)
+      CALL cp_fm_get_diag(matrix_lu, diag)
```
fstein93, line 129: and here

src/fm/cp_fm_basic_linalg.F
```diff
@@ -88,6 +89,71 @@ MODULE cp_fm_basic_linalg
...
+         i = indxl2g(irow_local, nrow_block, myprow, matrix_lu%matrix_struct%first_p_pos(1), nprow)
+         IF (ipivot(irow_local) /= i) P = P + 1
+      END DO
+      CALL matrix_lu%matrix_struct%para_env%sum(P)
+      ! very important fix
+      P = P/npcol
+#else
+      CALL dgetrf(n, n, a, n, ipivot, info)
+      CALL cp_fm_get_diag(matrix_lu, diag)
+      determinant = PRODUCT(diag)
```
fstein93, line 146: and here

fstein93 (commented):
I have a potentially simpler idea to collect the "raw determinant". It does not require an intermediate array and reduces the communication although your solution is probably good enough.

andsinya:
> I have a potentially simpler idea to collect the "raw determinant". It does not require an intermediate array and reduces the communication although your solution is probably good enough.

Thanks. When I couldn't understand, why I was also getting different determinant signs with different number of MPI ranks like with old routines, I decided to rewrite my routine as clean as possible, not looking to save microseconds anymore, and it helped, that's how I arrived at this line "P = P/npcol". Determinants are usually used for postprocessing anyway, which is only done once, so I wouldn't worry about the speed.
Any optimizations require a lot of testing on different matrices with different number of ranks and threads to make sure it still works as before. And this time aside from checking the results, one must also measure the performance to confirm that it is indeed faster now. I have no time for that, sorry.

mkrack:
@andsinya I think that the code simplification proposed by @fstein93 is simple and safe for the parallel case, because it just replaces the global `MPI_SUM` for the diagonal matrix elements and the subsequent `PRODUCT` by a single `MPI_PROD` operation. Assuming similar timings for `MPI_SUM` and `MPI_PROD`, this will only reduce the timings by `PRODUCT` operation which is negligible. Likewise, I see no major advantage for the serial case. The simplification for the parallel case can be applied by a subsequent PR and the Intel oneAPI test fails because of another reason (see #3365). Therefore, I will rebase/merge this PR.

PR #3372, opened 2024-04-25
Introduce `SETPOSFILE` as a command in cp2k_shell

WardLT:
There seems to be a recurring issues [for the cp2k_shell stalling when the large structures](https://groups.google.com/g/cp2k/c/2C7_YEHFKPg/m/DfUyEZo9AwAJ). From my gleaning of the linked email thread and those ones it references, the workaround has been to modify the shell to read positions from files. 

This PR adds a `SETPOSFILE` command that reads the new positions for atoms based on a file. I've got [a matching PR to ASE](https://gitlab.com/ase/ase/-/merge_requests/3341) which will use this feature.

- [x] Increment shell number
- [x] Remember to close file after done with it

### DIFF
--- a/src/start/cp2k_shell.F
+++ b/src/start/cp2k_shell.F
@@ -68,7 +68,7 @@ MODULE cp2k_shell
    PRIVATE
 
    ! Queried by ASE. Increase version after bug-fixing or behavior changes.
-   CHARACTER(LEN=*), PARAMETER                           :: CP2K_SHELL_VERSION = "6.0"
+   CHARACTER(LEN=*), PARAMETER                           :: CP2K_SHELL_VERSION = "7.0"
 
    TYPE cp2k_shell_type
       REAL(dp)                                           :: pos_fact = 1.0_dp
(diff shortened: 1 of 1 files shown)


oschuett:
Thank you very much for your contribution! I wasn't aware of this workaround, but it makes a lot of sense. Since this adds a new feature to the shell, please also increment the [CP2K_SHELL_VERSION](https://github.com/cp2k/cp2k/blob/34ef4722c83cb2e23bc45676edada79b16bdb05e/src/start/cp2k_shell.F#L71). You can then later check it on the ASE side.

Perhaps the ASE calculator should even print a warning for large system to hint people towards this new feature?

WardLT:
Thanks! Should I roll from 6.0 to 6.1 on the shell version? 

Good idea about the warning and adding in the check of shell version on the ASE side. I'll put those on my to-do list.

oschuett:
> Should I roll from 6.0 to 6.1 on the shell version?

It doesn't really matter. In the past I usually went with major increments, so `7.0`.

WardLT:
Alright, the ASE tests for this work on my desktop and on one of my problematic clusters. I'm going to polish up ASE, but I think this side of things is ready for review.

PR #3380, opened 2024-04-30
Revised tensor contractions

hfp:
* Tensor contractions in CP2K currently leave some perf on the table; see https://github.com/cp2k/cp2k/pull/3380#issuecomment-2085573897
  - Other contraction routines can follow a similar flow and can probably be stream-lined (even names)
  - All contraction routines may be better called contract_xyz_something ("contract" in front).
  - Renamed libxsmm_abc_contract to abc_contract_xsmm (first shot/rename).
  - Interfaces to be revised like external/optional buffer allocation,
    allowing minimal/exact buffer size according to best code-path.
  - Arguments to allow more flexible result format like leading dimension.

* libxsmm_abc_contract (abc_contract_xsmm)
  - Implemented faster-path contraction if current interface permits
    like temporary/intermediate buffer being large enough.
  - Adjusted timeset to show which path was taken (uv vs vw).
  - Let abc_contract_xsmm support alpha and beta (pre/post-factor).
  - Revised code format and specified INTENT.
  - Documented subroutine arguments.

### DIFF
--- a/src/aobasis/ai_contraction_sphi.F
+++ b/src/aobasis/ai_contraction_sphi.F
@@ -9,13 +9,13 @@
 !> \brief Contraction of integrals over primitive Cartesian Gaussians based on the contraction
 !>        matrix sphi which is part of the gto_basis_set_type
 !> \par History
-!>      -added libxsmm_abc_contract routine, A. Bussy (04.2020)
+!>      -added abc_contract_xsmm routine, A. Bussy (04.2020)
 !> \author Dorothea Golze (05.2016)
 ! **************************************************************************************************
 MODULE ai_contraction_sphi
 
    USE kinds, ONLY: dp
-#if(__LIBXSMM)
+#if defined(__LIBXSMM)
    USE libxsmm, ONLY: LIBXSMM_PREFETCH_NONE, &
                       libxsmm_blasint_kind, &
                       libxsmm_dgemm, &
@@ -33,7 +33,7 @@ MODULE ai_contraction_sphi
 
    CHARACTER(len=*), PARAMETER, PRIVATE :: moduleN = 'ai_contraction_sphi'
 
-   PUBLIC :: ab_contract, abc_contract, abcd_contract, libxsmm_abc_contract
+   PUBLIC :: ab_contract, abc_contract, abcd_contract, abc_contract_xsmm
 
 CONTAINS
 
@@ -229,7 +229,7 @@ CONTAINS
 !>        the LIBXSMM library to be initialized somewhere before this routine is called.
 !> \param abcint contracted integrals
 !> \param sabc uncontracted integrals
-!> \param tsphi_a assumed to have dimensions nsgfa x ncoa
+!> \param sphi_a assumed to have dimensions nsgfa x ncoa
 !> \param sphi_b assumed to have dimensions ncob x nsgfb
 !> \param sphi_c assumed to have dimensions ncoc x nsgfc
 !> \param ncoa ...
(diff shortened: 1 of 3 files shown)


hfp:
This PR proposes to exploit more optimal tensor contractions. For three operands like "abc_contract", this is about mapping to (AB)C vs A(BC) which benefits temporary/intermediate/work arrays as well as FLOPS.

hfp:
Running some GW workload, this already shows some benefit and gives an insight of the impact by an improved profile/output like `abc_contract_xsmm_ab` vs `abc_contract_xsmm_bc`.

oschuett:
It's seems this PR caused some problems on ARM - under both [Linux](https://dashboard.cp2k.org/archive/arm64-psmp/commit_41b088e6f0a81c464025a22d9d9e66fb1cb93ad8.txt) and [macOS](https://dashboard.cp2k.org/archive/darwin-gnu-arm64-psmp/commit_7137784189958a95e8fafb04ba9c50ad8e4101f6.txt).

hfp:
> It's seems this PR caused some problems on ARM - under both [Linux](https://dashboard.cp2k.org/archive/arm64-psmp/commit_41b088e6f0a81c464025a22d9d9e66fb1cb93ad8.txt) and [macOS](https://dashboard.cp2k.org/archive/darwin-gnu-arm64-psmp/commit_7137784189958a95e8fafb04ba9c50ad8e4101f6.txt).

Thank you for taking a look and finding out!

There is a simple fix like not implementing the optimal contraction-flow for the BLAS-based code-path. This should completely restore the gold values. Though, we do not (yet) run LIBXSMM on ARM.

However, a different flow is only taken if the intermediate result-buffer is a smaller/equally sized matrix and if the number of FLOPS is less, i.e., current results are likely representing higher accuracy or less rounding errors. I can calibrate differences if you like.

I am somewhat worried about these numerical deviations. Perhaps a bad sign for OpenBLAS on ARM?

oschuett:
> I am somewhat worried about these numerical deviations.

I don't think these are just numerical deviations. Many of the tests seem to have not converged (huge output). And one even returns `'*****************'`. After >50 failures the run was aborted early.

mkrack:
This PR introduces an [if clause](https://github.com/cp2k/cp2k/pull/3380/files#diff-851f6fabfba92f159a786a365f7096ca87d089ef0cc86205cc1c2d8af7cf2e42R308-R322) for `ab` which causes the wrong results when the `else` block is considered, i.e. a contraction over `b` and `c` instead of `a` and `b` are performed. Setting always `ab = .TRUE.` seems to fix the problem. I think CP2K requires here only contractions over `a` and `b`.

hfp:
I will take a look!

hfp:
> This PR introduces an [if clause](https://github.com/cp2k/cp2k/pull/3380/files#diff-851f6fabfba92f159a786a365f7096ca87d089ef0cc86205cc1c2d8af7cf2e42R308-R322) for `ab` which causes the wrong results when the `else` block is considered, i.e. a contraction over `b` and `c` instead of `a` and `b` are performed. Setting always `ab = .TRUE.` seems to fix the problem. I think CP2K requires here only contractions over `a` and `b`.

Both branches are mathematically equivalent in multiplying three matrices A, B, and C. The first branch multiples A, B, and C in order but the 2nd branch first multiplies B and C and then A times the first result. I still wonder why errors only appear on ARM.

hfp:
Found it, I have some wrong leading dimensions in the else-case of the BLAS based code-path.

PR #3390, opened 2024-05-02
Reduced intermediate/temporary memory consumption of abc_contract

hfp:
The original idea was to preserve `cpp` and `cpc` buffers at least as POINTER/name. However, it's a bit intriguing to preserve the dimensionality if one just wants an offset on one of the three dimensions like `cpp => tmp(:, :, 2)`.


hfp:
All tests passed/correct but one slow test (`tests/QS/regtest-tddfpt-force-gapw/h2o_t07.inp`).

hfp:
( This PR has no different code-paths depending on the platform or such. )

PR #3402, opened 2024-05-06
PW: Refactor communicators and normalization factors

fstein93:
This PR
- removes one the non-cartesian communicator and all components which are mere copies of components of the communicator (positions, process grid).
- pushes normalization factors downstream

This PR was tested on Noctua2 (regtests, 128 H2O with PBE on GPUs vs CPU trunk for correctness)


fstein93:
The CUDA CI build is currently not available. But it runs correctly on the Noctua2 GPU partition.

PR #3413, opened 2024-05-10
dbm: leverage MPI_Alloc_mem/MPI_Free_mem

hfp:
* Introduced dbm_malloc/dbm_free using dbm_mpi_alloc_mem/dbm_mpi_free_mem.
* Replaced realloc since there is no MPI pendant (kept some comment).
* Removed unused qualification/types/casts (minor/general).
* Reuse some common code/definitions (OpenCL kernel).

### DIFF
--- a/src/dbm/dbm_matrix.c
+++ b/src/dbm/dbm_matrix.c
@@ -14,6 +14,7 @@
 #include <string.h>
 
 #include "dbm_hyperparams.h"
+#include "dbm_library.h"
 #include "dbm_matrix.h"
 #include "dbm_mpi.h"
 
@@ -34,21 +35,21 @@ void dbm_create(dbm_matrix_t **matrix_out, dbm_distribution_t *dist,
   matrix->dist = dist;
 
   size_t size = (strlen(name) + 1) * sizeof(char);
-  matrix->name = malloc(size);
+  matrix->name = dbm_malloc(size);
   memcpy(matrix->name, name, size);
 
   matrix->nrows = nrows;
   matrix->ncols = ncols;
 
   size = nrows * sizeof(int);
-  matrix->row_sizes = malloc(size);
+  matrix->row_sizes = dbm_malloc(size);
   memcpy(matrix->row_sizes, row_sizes, size);
 
   size = ncols * sizeof(int);
-  matrix->col_sizes = malloc(size);
+  matrix->col_sizes = dbm_malloc(size);
   memcpy(matrix->col_sizes, col_sizes, size);
 
-  matrix->shards = malloc(dbm_get_num_shards(matrix) * sizeof(dbm_shard_t));
+  matrix->shards = dbm_malloc(dbm_get_num_shards(matrix) * sizeof(dbm_shard_t));
 #pragma omp parallel for
   for (int ishard = 0; ishard < dbm_get_num_shards(matrix); ishard++) {
     dbm_shard_init(&matrix->shards[ishard]);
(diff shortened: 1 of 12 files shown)


hfp:
It has measurable benefit using MPI's memory allocation wrt send/receive buffers (tested with some bigger GW workload). In any case, this can be decided now at library level without touching the allocations again (`dbm_malloc` and `dbm_free`).

oschuett:
Is it really beneficial to replace *every* malloc with `MPI_Alloc_mem`? Usually, these things come with some kind of tradeoff. Note that for MPI all the data is copied into dedicated buffers (` dbm_pack_t`). So maybe `MPI_Alloc_mem` should only be used for those?

oschuett:
I'd propose to replace only the following mallocs:

https://github.com/cp2k/cp2k/blob/be1f5882581616708fdc25ee40722957c884746f/src/dbm/dbm_multiply_comm.c#L359-L360

https://github.com/cp2k/cp2k/blob/be1f5882581616708fdc25ee40722957c884746f/src/dbm/dbm_multiply_comm.c#L381-L382

https://github.com/cp2k/cp2k/blob/be1f5882581616708fdc25ee40722957c884746f/src/dbm/dbm_multiply_comm.c#L430-L431

https://github.com/cp2k/cp2k/blob/be1f5882581616708fdc25ee40722957c884746f/src/dbm/dbm_mempool.c#L37

https://github.com/cp2k/cp2k/blob/be1f5882581616708fdc25ee40722957c884746f/src/dbm/dbm_matrix.c#L152-L153

Since these are all MPI-related, you can just call `dbm_mpi_alloc_mem` directly.

hfp:
Makes sense!  I originally followed something like your proposal but found manipulating the shard's memory is a bit scattered and then wanted "to not judge" whether memory is involved into MPI comms or not. Anyhow, your proposal looks good! I will revise the PR accordingly.

src/dbm/dbm_matrix.c
```diff
@@ -34,21 +35,22 @@ void dbm_create(dbm_matrix_t **matrix_out, dbm_distribution_t *dist,
...
   memcpy(matrix->row_sizes, row_sizes, size);
 
   size = ncols * sizeof(int);
-  matrix->col_sizes = malloc(size);
+  matrix->col_sizes = (int *)malloc(size);
   memcpy(matrix->col_sizes, col_sizes, size);
 
-  matrix->shards = malloc(dbm_get_num_shards(matrix) * sizeof(dbm_shard_t));
+  matrix->shards =
+      (dbm_shard_t *)malloc(dbm_get_num_shards(matrix) * sizeof(dbm_shard_t));
```
oschuett: Is there an advantage to adding these casts? Since the type is already in the `sizeof`, I usually opted for the more compact and readable code.

hfp: I am not sure actually why it does not warn if casts are not there. Must be some C99 or later thing. Return-type of `malloc` is certainly `void*` which may not promote to any more qualified type without hint (cast).

hfp: Objective was to not rely on any "permissive" thing.

oschuett: I believe this is one of those subtle differences between C++ and C. According to [internet wisdom](https://stackoverflow.com/questions/605845/should-i-cast-the-result-of-malloc-in-c), it's considered good style to omit the cast in C.

oschuett:
Thanks a lot for looking into DBM's MPI performance!

I actually designed the communication scheme to allow for asynchronous GPU-to-GPU communication. Unfortunately, I never fully realized it and today there are still [mpi_sendrecv](https://github.com/cp2k/cp2k/blob/ee725ffc68fafd11fbf186e4af07b2db94632368/src/dbm/dbm_multiply_comm.c#L481) and [backend_upload_packs](https://github.com/cp2k/cp2k/blob/ee725ffc68fafd11fbf186e4af07b2db94632368/src/dbm/dbm_multiply.c#L361).

And in case you haven't seen it already, I wrote a bit of prose in section II of https://doi.org/10.1063/5.0144493.

PR #3428, opened 2024-05-23
Fix indexing and enable stress prediction in NequIP & Allegro

mariabilichenk0:
This PR introduces a number of fixes and enables stress prediction with NequIP and Allegro:

- fix indexing of elements in cp2k to correspond to the indexing of torch model

- update NequIP/Allegro models & TEST_FILES from NequIP==0.5.6 to NequIP==0.6.0

- change parsing of dtype from custom subroutine to torch_model_read_metadata, see #2786

- add kinds comparison cp2k torch

- add prediction of stress and add stress to regtest for nequip water-bulk-dp.inp


fstein93:
This PR introduced a bug in parallel builds (wrong results). Can you check what is going on there?

PR #3432, opened 2024-05-24
fixed cutoff and rel_cutoff tutorial

Anyborr:
Updated the cutoff/rel_cutoff to the new docs, and fixed some issues caused by recent updates to cp2k.

### DIFF
--- a/docs/methods/dft/cutoff.md
+++ b/docs/methods/dft/cutoff.md
@@ -1,0 +1,677 @@
+# How to Converge the CUTOFF and REL_CUTOFF
+
+## Introduction
+
+`QUICKSTEP`, as with nearly all ab initio Density Functional Theory simulation packages, requires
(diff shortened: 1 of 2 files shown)


Anyborr:
Question: This tutorial contains a 1.8 MB downloadable .tgz file. Do we consider it bad practice to include it in the docs/ since it will add to the total size of the cp2k package? I assume it's not a problem in this case, but might be for larger tarballs.

oschuett:
> Question: This tutorial contains a 1.8 MB downloadable .tgz file.

The tar-balls should not go into the git repo. We don't have a good solution yet. In the meantime, please upload them to the wiki and link to them. Example: https://github.com/cp2k/cp2k/blob/efbebd74523f78d037c730de201f15e042063791/docs/methods/qm_mm/image_charges.md?plain=1#L134

PR #3443, opened 2024-05-31
docker: intel oneapi and gnu based build

hfp:
* Docker image based on "rockylinux" (RPM distro) to supply something "HPC" (not necessarily HPE/Cray/Suse).
* Recipe aka HFP's "best" (minus obscure bits); effectively: GNU Fortran + Intel MPI + Intel MKL + LIBXSMM.
* Non-reproducible (as every image) because of pulling bits from the Internet (dependencies).
* Of interest: LIBINT/latest/2.9.0 bootstrap for CP2K.

### DIFF
--- a/tools/docker/Dockerfile.intel
+++ b/tools/docker/Dockerfile.intel
@@ -1,0 +1,73 @@
+FROM rockylinux:9
+
+RUN ( \
+  echo "[oneAPI]" && \
+  echo "name=Intel oneAPI repository" && \
+  echo "baseurl=https://yum.repos.intel.com/oneapi" && \
+  echo "enabled=1" && \
+  echo "gpgcheck=1" && \
+  echo "repo_gpgcheck=1" && \
+  echo "gpgkey=https://yum.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB" \
+) >/etc/yum.repos.d/oneAPI.repo
+
+RUN dnf upgrade -y
+RUN dnf install -y \
+  gcc-gfortran gcc-c++ git procps which bzip2 \
+  intel-oneapi-mpi intel-oneapi-mpi-devel \
+  intel-oneapi-mkl intel-oneapi-mkl-devel \
+  cmake automake libtool wget diffutils \
+  gmp-devel boost-devel
+
+ARG MYENV=/root/my.env
+RUN ( \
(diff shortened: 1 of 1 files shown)


hfp:
An OCI compliant image which is not necessarily tested with native Docker (I used sudo/podman). It should work with Docker out-of-the-box because of requiring less features (only OCI compliance).

hfp:
This is LIBINT/latest/2.9.0 bootstrapped for CP2K.
**Current LIBINT in CP2K**: 2.6.0

mkrack (commented):
All Dockerfiles are created by the generation script [generate_dockerfiles.py](https://github.com/mkrack/cp2k/blob/master/tools/docker/generate_dockerfiles.py). Therefore, this Dockerfile is special in this respect.

hfp:
> All Dockerfiles are created by the generation script [generate_dockerfiles.py](https://github.com/mkrack/cp2k/blob/master/tools/docker/generate_dockerfiles.py). Therefore, this Dockerfile is special in this respect.

Yes, that's true. Also, it deviates the way it pulls dependencies etc. Certainly it builds CP2K in a way not supported by the toolchain like GNU Compiler, Intel MPI, and Intel MKL. I am not insisting in taking the PR; I am actually not sure if should go in.

I can put the Dockerfile elsewhere (if I wanted to document the way I am building CP2K). Perhaps it's also useful if we want a recipe for generating CP2K's flavor of LIBINT.

oschuett:
Since this is a Dockerfile for running CP2K in production, it should go into the [cp2k-containers](https://github.com/cp2k/cp2k-containers)  repo.

For testing Intel's HPC Kit we use [Dockerfile.test_intel-psmp](https://github.com/cp2k/cp2k/blob/master/tools/docker/Dockerfile.test_intel-psmp)  - although it's broken (again) because of a missing gpg key.


> Perhaps it's also useful if we want a recipe for generating CP2K's flavor of LIBINT.

That is definitely super useful! With our current setup the best place for the info is probably the [libint-cp2k](https://github.com/cp2k/libint-cp2k) repo.

hfp:
PR will be retargeted to [cp2k-containers](https://github.com/cp2k/cp2k-containers) repository.

> although it's broken (again) because of a missing gpg key.

I had something like this too on my own system (Debian) and I needed to re-add the oneAPI repository according to the current documentation (perhaps just a new key).

PR #3445, opened 2024-05-31
Allegro: Fix parallelization with virials

mariabilichenk0:


oschuett:
The tests failed because allegro/nequip is taking too long. Ideally each test input should run in less than 10 seconds. This is something we also noticed on [the dashboard](https://dashboard.cp2k.org/archive/current-pdbg/index.html). 

Could you speed them up somehow? Is see that you are already doing just a single MD step. So maybe we need to make the system smaller, e.g. use only 16 water molecules.

fstein93:
Is the fix I added in #3434 erroneous?

mariabilichenk0:
> Is the fix I added in #3434 erroneous?

I checked stress with 1 and 4 cores (both Allegro and NequIP) and I get the same result, so I think all is good there 😊

mariabilichenk0:
> Could you speed them up somehow? Is see that you are already doing just a single MD step. So maybe we need to make the system smaller, e.g. use only 16 water molecules.

I have reduced all the models to 1 layer along with reducing few other training parameters, which makes the models not suitable for production runs but speeds up the tests. From `regtest-allegro/`  I have removed `water-gra-film-sp.inp` and left only `water-bulk-sp.inp` since they are using the same model, but the later one is faster. From `regtest-nequip/` I removed `water-bulk-dp.inp` energy test case and left only stress test case since they share the input.  This has improved the timings overall for both `sdbg` and `pdbg`. As for the system size, it took 14sec for a single water molecule and 19sec for 32 waters, so halving the system size,  unfortunately, does not make it much faster.  Thank you for the suggestions 😊

PR #3448, opened 2024-06-01
GW for small unit cells with full k-points (using DFT with k-points)

JWilhelm:


src/rpa_gw.F
```diff
@@ -4777,13 +4778,14 @@ SUBROUTINE get_sigma_c_bisection_pade(gw_energ, Eigenval_scf, Sigma_x_minus_vxc_
...
-            CPWARN(error_msg)
-            EXIT
-         END IF
+!         IF (icount > 500) THEN
+!            WRITE (n_level_gw_ref_char, '(I10)') n_level_gw_ref
+!            WRITE (error_msg, '(A,A,A)') " Self-consistent quasi-particle solution of "// &
+!               "MO ", TRIM(n_level_gw_ref_char), " has not been found."
+!            CPWARN(error_msg)
+!            EXIT
+!         END IF
```
fstein93: Wrap it in IF (.FALSE.) THEN ... END IF or remove it (same below)

fstein93 (commented):
It is just a minor thing. If you want to keep the debug code, wrap it such that the compiler can check it (but it will be removed later), otherwise, it will be kept around and might get out-of-sync some time later.

src/rpa_gw.F
JWilhelm: sure!

PR #3464, opened 2024-06-12
dbm: use squared epsilon instead of square-root of norm

hfp:
* OpenCL: account for GEN-kernel (not part of PR).

### DIFF
--- a/src/dbm/dbm_matrix.c
+++ b/src/dbm/dbm_matrix.c
@@ -278,6 +278,7 @@ void dbm_filter(dbm_matrix_t *matrix, const double eps) {
   if (eps == 0.0) {
     return;
   }
+  const double eps2 = eps * eps;
 
 #pragma omp parallel for schedule(dynamic)
   for (int ishard = 0; ishard < dbm_get_num_shards(matrix); ishard++) {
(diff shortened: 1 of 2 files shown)


src/dbm/dbm_matrix.c
```diff
@@ -298,7 +299,7 @@ void dbm_filter(dbm_matrix_t *matrix, const double eps) {
         norm += old_blk_data[i] * old_blk_data[i];
       }
       // For historic reasons zero-sized blocks are never filtered.
-      if (sqrt((float)norm) < eps && block_size > 0) {
+      if (block_size > 0 && norm < eps2) {
```
oschuett, line 302: Could you please also remove the comment above:
`// Compute norm as double, but compare as float.`

src/dbm/dbm_matrix.c
```diff
@@ -273,6 +273,7 @@ void dbm_clear(dbm_matrix_t *matrix) {
  * \author Ole Schuett
  ******************************************************************************/
 void dbm_filter(dbm_matrix_t *matrix, const double eps) {
+  const double eps2 = eps * eps;
```
oschuett: Maybe move this after the fast return.

```
if (eps == 0.0) {
    return;
}
```

hfp: I tend to write C89 with no declaration after executable statement.

( Similarly, on the F-side it means fixed format ;-) )

PR #3468, opened 2024-06-12
Fix AXK implementation and extend it to SOSEX

fstein93:
This PR
- fixes the legacy implementation of RPA-AXK
- adds the SOSEX kernel
- implements a second implementation employing the original integrals guaranteeing cancellation of terms in condensed phase with Coulomb correlation potential and truncated Coulomb HFX potential. This implementation has a worse scaling than the HFX-based implementation
- misc. refactorings to cp_fm_structs (bind index conversion functions to the type)

Changes to the input:
- the AXK keyword is replaced by an EXCHANGE_CORRECTION section
- The keyword USE_HFX_IMPLEMENTATION switches to the legacy-like HFX-based implementation
- BLOCK_SIZE increases the internal block size of the new implementation employing DGEMM/SpLA. The larger the blocksize the more memory is required but the better the performance.

Annotation:
- I observe numerical issues introducing potential errors at the order of 0.1 mEh for both implementations (the new implementation has lower errors).
- The new implementation could be replaced with a RI-MP2-like implementation which requires more refactoring on that side (future work) which could have better numerical properties.


hfp:
With IFORT the following error is issued:

> src/rpa_exchange.F(279): error #7999: The FROM or TO arguments of a MOVE_ALLOC reference must not be INTENT(IN).   [MOVE_ALLOC]
      CALL MOVE_ALLOC(qs_env%mp2_env%ri_rpa%mo_coeff_o, exchange_env%mo_coeff_o)
----------------------^
src/rpa_exchange.F(280): error #7999: The FROM or TO arguments of a MOVE_ALLOC reference must not be INTENT(IN).   [MOVE_ALLOC]
      CALL MOVE_ALLOC(qs_env%mp2_env%ri_rpa%mo_coeff_v, exchange_env%mo_coeff_v)

This seems to be a reasonable complaint. I would send a PR, but I am not sure if it is intended to change `qs_env` when calling `hfx_create_subgroup` (making `qs_env` an INOUT).

fstein93:
Standard-wise, the Intel compiler is wrong because the intent of a pointer variable refers only to its association state, not to the data it is pointing to. It is a common misconception which we have already found on several occasions with several compilers. I will submit a fix for this and other issues.

hfp:
I will convey your comment. i.e., opening an issue for it. As a side-note, I am getting next-gen Intel Fortran compiler to work with CP2K (logging issues and reproducers for our compiler team).

hfp:
Should `qs_env` be simply declared as `TYPE(qs_environment_type) :: qs_env`? I mean removing the POINTER and removing the INTENT like making it INTENT(INOUT)?

fstein93:
If the error is still present in that case, it can still be considered a bug. But I haven't tested it. I have now decided to drop the intent. But I am still not sure about the issue with GCC 8.

hfp:
Meanwhile, I found also a mistake on my side: I missed to apply `-std08` to IFORT's flags. Apparently, `-std08` allows your code to pass as-is. In this instance, IFORT seems to be better than I thought ;-).

PR #3473, opened 2024-06-13
Docker: Upgrade Intel test to hpckit:2024.1.1-devel-ubuntu22.04

oschuett:


oschuett:
The good news is that Intel fixed their hpckit image, the bad news is that it [doesn't work](https://storage.googleapis.com/cp2k-ci/run-cp2k-intel-psmp-e8e84338_report.txt) anymore after the ELPA upgrade (#3469):

```
checking whether Fortran compiler infers interfaces ... no
checking for fc option to support OpenMP... unsupported
configure: error: Could not compile a Fortran program with OpenMP, adjust FCFLAGS
```

I have a hard time debugging this because my dev machine has an AMD CPU.

@mkrack or @hfp maybe you could take a look?

mkrack:
I looked already at that. The ELPA configure seems to fail with `OpenMP` enabled, because it adds always the compiler flag `-openmp` which is not accepted by the Intel compiler. The build worked for me with `OpenMP` disabled. Maybe, this could be a work around, since there have been other issues with ELPA and `OpenMP` in the past anyway.

Edit: The problem exists also with the previous ELPA version and thus is not related to #3469

oschuett:
Oh, this reminds me of #1860. We do indeed already [disable OpenMP on Cray](https://github.com/cp2k/cp2k/blob/9d45d454af6559fc8492eae6d4833d6dca911374/tools/toolchain/scripts/stage5/install_elpa.sh#L51).

oschuett:
I'm merging this as is. The work on Intel+ELPA will continue at #3477.

hfp:
FYI, I am finally working on getting next-gen Intel Fortran (IFX) consume CP2K's code. I will take the ELPA issue as well or share with my colleague in this case. IFX-wise, I have one open issue causing ICEs with pw_methods.F and xc.F. Beside, there is at least one issue at runtime regarding ALLOCATABLE components in a derived type.

hfp:
ELPA-wise and as a workaround, I think relying on MKL's ScaLapack is just fine. However, ELPA is also targeting Intel GPUs but I haven't tested this for CP2K (we have good perf. with ELPA in other apps).

PR #3479, opened 2024-06-15
Switch toolchain to new Intel Fortran compiler ifx (ifort -> ifx)

mkrack:
This PR serves to reproduce the current problems in compiling CP2K with ifx 2024.1, namely the termination with an internal compiler error in pw_methods.F.

### DIFF
--- a/arch/Linux-intel-x86_64.psmp
+++ b/arch/Linux-intel-x86_64.psmp
@@ -2,7 +2,7 @@
 #
 # CP2K (Intel/MKL x86_64) arch file for Linux clusters
 #
-# Tested with: Intel 22.2 , Intel MPI, Intel MKL,
+# Tested with: Intel 2024.2.0, Intel MPI, Intel MKL,
 #              LIBINT 2.6.0, LIBXC 6.2.2, ELPA 2024.03.001,
 #              PLUMED 2.9.0, SPGLIB 2.3.1, LIBVORI 220621,
 #              GSL 2.7, COSMA 2.6.6, HDF5 1.14.2, SIRIUS 7.5.2,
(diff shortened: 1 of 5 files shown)


hfp:
Thank you, Matthias!

I am working on it, the `pw_methods.F` as well as `xc.F` translation units are failing (see [here](https://github.com/hfp/xconfigure/blob/main/config/cp2k/Linux-x86-64-intelx.arch#L1109)) with both likely sharing the same cause like using array-loops inside of a workshare. This is being resolved. Beside of adjusting the compilation rules, I tried some workaround (see [here](https://github.com/hfp/cp2k/commits/master/src/pw/pw_methods.F)). However, the latter did not solve it or catch all cases.

Further, there are runtime errors as well which will be logged asap and submitted for resolution.

mkrack:
@hfp thanks, great that you are working on this. Earlier versions of ifx succeeded to compile CP2K, but gave runtime errors. I guess recent source code changes challenged the ifx compiler again.

hfp:
> @hfp thanks, great that you are working on this. Earlier versions of ifx succeeded to compile CP2K, but gave runtime errors. I guess recent source code changes challenged the ifx compiler again.

I remembered so too like compiled successfully, but things move quickly. In addition to CP2K's dashboard, I will have an internal test to make sure things do not break. I guess, runtime issues are somewhat feature gaps not exercised elsewhere (copying a derived type with allocatable components along perhaps intermixed with POINTER vars in the same struct or whatever). We will get it ;-)

Even the ICE with array-loops inside of a workshare is sort of a feature gap issue since when it compiled successfully, it likely did not do any worksharing...

hfp:
( Side note [since this question came up elsewhere], IFX is not based on Flang but rather based on IFORT now "classic". I am all positive for the language itself to see investment into Fortran and to keep some viable variety with respect to compilers. )

PR #3480, opened 2024-06-16
Conventions: Fix uninitialised variables and missing default initializers

fstein93:
This PR:
- fixes uninitialised variables
- adds default initializers for most types
- adds an implicit interface to one module
- removes ~300 suppressions of the convention checker

Default initializers are still missing due if the respective type member variables of the libraries
- DBCSR
- SIRIUS (PR in progress: electronic-structure/SIRIUS#1005)
- PEXSI

For the record:
- Types derived from the abstract type eri_type_eri_element_func may be false positives as all their member variables have a default (the base type does not have any member variables)
- semi_empirical_int_debug could be reimplemented using submodules (requiring adjustments on the build system) due to dependencies between this file and semi_empirical_int_utils


hfp:
Some hero effort (IMHO), and it's all green! Hence, it should be perhaps merged quickly (touches 126 files).

Are you done, Frederick?

PR #3482, opened 2024-06-17
Infrastructure for Kpoint symmetries

juerghutter:
- Use K290 code to calculate reduced k-point set
- Use K290 to calculate symmetry operations relating k-points
- New code for rotations of basis functions


fstein93:
This PR introduced several convention violations (missing initializer, GOTOs, implicit save variables).

PR #3485, opened 2024-06-18
Fix conventions

fstein93:
Follow up to #3420


fstein93:
The left-over convention violations were introduced by #3482  and should be fixed by @juerghutter as he knows better which of them are intended and which are not.

PR #3502, opened 2024-06-25
Updated tutorials (metadynamics and delta scf)

fbranunes:


oschuett:
I made some tweaks and now Sphinx is happy. You can preview the pages [here](https://ci.cp2k.org/artifacts/run-cp2k-manual-2a047ae0/manual/index.html).
I also took the liberty to revert the title to _"X-Ray Absorption from ΔSCF"_ because I think it is more helpful when navigating the X-ray section. Hopefully you agree?

PR #3506, opened 2024-06-27
Update DBCSR v2.7.0-rc2

alazzaro:


oschuett:
~~Don't hold your breath for the CUDA performance test. It hasn't run for two months because V100 Spot GPUs have been stocked out at GCP. I don't expect the situation to improve anytime soon. So, I guess we should just retire the test.~~

Actually, never mind. Kubernetes is hard.

oschuett:
I got the CUDA performance test working again. The dashboard run just finished. It seems not much has changed :-)

alazzaro:
> I got the CUDA performance test working again. The dashboard run just finished. It seems not much has changed :-)

? 
Thanks from bringing up the test. However, the changes I'm proposing here are valid for GH (Santis at CSCS) or any new architecture, where we don't want to tune all kernels, and we are trying a more pragmatic approach: tune some kernels, use generic for others. For the V100 test in the CI I don't expect any change, V100 has a lot of kernels. Now I'm testing the extreme: use the generic for all kernels (only test). If it passes the test, I will make the RC and wait @RMeli and @abussy to test on Santis (when the system is back) before releasing an official DBCSR version for CP2K.

The real problem is that we are asking people to tune kernels and add them to the database (I've now added a warning for that), but this is something which requires to rebuild CP2K (or at least DBCSR), which is something people cannot do on a supercomputer with a centralized version of CP2K. 
It can be much better to move DBCSR standalone (at least we don't need a new CP2K version whenever we add a new kernel in DBCSR), but still people have to recompile and link something. The solution proposed here tries to mitigate the problem by introducing a generic ACC kernel.

oschuett:
Yes, a good generic kernel is super valuable. For tensors it's actually a necessity due to their diverse block-sizes. Furthermore, my understanding is that these days proper GPU-to-GPU communication is more important that optimized kernels. This is why DBM has so far only a [generic kernel](https://github.com/cp2k/cp2k/blob/master/src/dbm/dbm_multiply_gpu_kernel.cu).

alazzaro:
The wrong result is not due to this PR (already present). I'm going to make a new DBCSR RC.

@oschuett please let us know the CP2K release timeline, if needed I can turn the RC in an official DBCSR release, but I have only today for doing that, otherwise we go to the 3rd week of July (see https://github.com/cp2k/cp2k/discussions/3453#discussioncomment-9894447 )

oschuett:
> @oschuett please let us know the CP2K release timeline.

Sorry for the late reply. Let's move the release to July then. Thanks a lot for your GH efforts!

PR #3511, opened 2024-06-28
Accelerate toolchain build for ninja and dftd4

mkrack:
- Add parallel flag
- Enable log file dump to screen on failure


mkrack:
@oschuett Do you know why there is a timeout for the mergeability check?

oschuett:
> @oschuett Do you know why there is a timeout for the mergeability check?

I guess, Github had a partial outage / service degradation.

We wait up to [10 * 5 seconds](https://github.com/cp2k/cp2k-ci/blob/9edd04d962dfe85a81bccbedbdb4e3f2ff4594dd/backend/backend.py#L283) for the check, which is usually plenty. I don't want to wait much longer because the CI does not process any other events during that time.

PR #3525, opened 2024-07-02
CMake documentation

RMeli:
Add section on CMake to documentation. It is rather minimal at the moment, it can be expanded later. This PR also removes the previous `README_cmake.md` file since it's a bit outdated. Most of the information in there has been mirrored in the new documentation. However, in the new documentation I only list a handful of CMake variables, to minimise the chances to go out of sync with the `CMakeLists.txt` file itself.


RMeli:
@mtaillefumier

mtaillefumier (approved):
Looks alright to me. Few comments

- MPI is on by default
- ${CP2K_REPOSITORY} may need to be defined on commented on. Users might just copy paste and complain afterwards that it does not work.

PR #3540, opened 2024-07-10
simplified GW self-consistency input

JWilhelm:


tests/QS/regtest-gw/TEST_FILES
```diff
@@ -11,4 +11,6 @@ evGW_OH_PBE_svd.inp                                   11      1e-08            -
 evGW_H2O_PBE_RI_HFX_svd.inp                           11      1e-08            -13.112474185078236
 scGW0_H2O_PBE.inp                                     11      1e-08            -17.118441200805574
 G0W0_H2O_PBE0_Hedin_shift.inp                         78      1e-04                         16.806
+evGW0_H2O_PBE_set_sc_flag.inp                        111      1e-04                        16.5546
```
oschuett, line 14: Should this be test type 110?

tests/TEST_TYPES
```diff
@@ -1,4 +1,4 @@
-110
+111
```
oschuett: This should be 112.

PR #3541, opened 2024-07-11
Docs: Improve BSE methods section

MGraml:


oschuett:
Looking at the [Sphinx output](https://storage.googleapis.com/cp2k-ci/run-cp2k-manual-7927ca5a_report.txt), most issues were about missing bibliography entries. While these are stored in [bibliography.F](https://github.com/cp2k/cp2k/blob/master/src/common/bibliography.F), the citation keys are actually generated by [this code](
https://github.com/cp2k/cp2k/blob/da085fddd526fd2724bf4a5c0667acec3708425d/src/common/reference_manager.F#L159-L192).

You can figure out a new citation's key by running `cp2k --xml` and then looking at the produced `references.html`.

PR #3560, opened 2024-07-19
Tests: Turn --mpiexec argument into a template

oschuett:
Implements feature request from https://github.com/cp2k/cp2k/issues/2996#issuecomment-2231141062.

@mtaillefumier is this what you had in mind?


mtaillefumier:
Indeed. it is what I had in mind. Thanks a lot

PR #3561, opened 2024-07-20
RI basis sets for MOLOPT basis sets for whole periodic table (might be discussed in developer meeting)

JWilhelm:
For RI-MP2, RI-RPA, GW


JWilhelm:
it seems that there were no comments on the developer's meeting concerning the RI basis sets... for users this might be helpful, because for many RI-RPA and GW methods, the automatic RI basis set generation won't work because the automatic RI basis set generator generates contracted RI basis sets which at present cannot be handled by RI-RPA and GW (if I'm not wrong)

juerghutter:
A possible solution (or a fall back) could be provided by
create_primitive_basis_set in aobasis/basis_set_types.F

JWilhelm:
> A possible solution (or a fall back) could be provided by create_primitive_basis_set in aobasis/basis_set_types.F

Thanks, Jürg. We will check at a later stage. It is great to potentially get uncontracted RI basis by that routine.

oschuett:
So, shall I merge this PR?

JWilhelm:
> So, shall I merge this PR?

From my side that will be great!

PR #3576, opened 2024-07-25
Replace ISI format with plain Fortran in bibliography.F

oschuett:


src/common/reference_manager.F
```diff
@@ -47,21 +49,22 @@ MODULE reference_manager
...
-   INTEGER, PARAMETER :: ISI_length = 128
-
-   ! the way we store a reference, should remain fully private
    TYPE reference_type
       PRIVATE
-      ! the reference in a format as returned by the web of science
-      CHARACTER(LEN=ISI_length), DIMENSION(:), POINTER :: ISI_record => NULL()
-      ! the doi only, i.e. without "https://doi.org/"
-      CHARACTER(LEN=doi_length)                        :: DOI = ""
+      CHARACTER(LEN=default_string_length), DIMENSION(:), POINTER :: authors => NULL()
```
fstein93: I think you can turn it into an `ALLOCATABLE`

PR #3577, opened 2024-07-26
Updated LIBXSMM

hfp:
There are numerous smaller fixes targeting flawless build support. There is full functionality for ARM64 wrt CP2K's needs. The Fortran interface vs v1.17 is stable except minor glitches fixed. The functionality needed for CP2K is fully stable regardless of this unreleased version. Please note, there might be no v2.0 anytime soon (but wrt CP2K that's it already).


hfp:
This PR needs (beside of review) some help with uploading the bits to CP2K.org. I think the referred Tarball can be taken without change (hence the SHA256sum is alright).

hfp:
There might be subsequent work to leverage LIBXSMM for ARM64 wrt Toolchain and/or CMake.

oschuett:
I've uploaded the [tarball](https://www.cp2k.org/static/downloads/libxsmm-d009b33e8742a93c9e1549323587fb6197451294.tar.gz).

hfp:
Unfortunately, I need to refer to a new archive, i.e., upload to cp2k.org and deleting the old upload. I will ping here when ready.

hfp:
> Unfortunately, I need to refer to a new archive, i.e., upload to cp2k.org and deleting the old upload. I will ping here when ready.

Ready: https://github.com/libxsmm/libxsmm/archive/50982e992eefedd44ef9395f946dc73595b8579d.tar.gz

oschuett:
Voilà: https://www.cp2k.org/static/downloads/libxsmm-50982e992eefedd44ef9395f946dc73595b8579d.tar.gz

hfp:
> Voilà: https://www.cp2k.org/static/downloads/libxsmm-50982e992eefedd44ef9395f946dc73595b8579d.tar.gz

Thanks!  We need to redo once more. This should not gate the release.

hfp:
Please, one more time...  
https://github.com/libxsmm/libxsmm/archive/92b08d68724e9a2653056acb5a3c6667ecaeaea1.tar.gz

oschuett:
Done: https://www.cp2k.org/static/downloads/libxsmm-92b08d68724e9a2653056acb5a3c6667ecaeaea1.tar.gz

hfp:
This may not gate the release of CP2K. Tests look good now (two very minor change of margins/deviations). However, there is one SIGFPE with `tests/QS/regtest-ecp-2/SbH3_def2_gapw.inp` which looks somewhat unrelated (geoopt). I will take a look.

hfp:
My local tests with `tests/QS/regtest-ecp-2/SbH3_def2_gapw.inp` went all fine. I used current master/main revision of libgrpp which now requires `libgrpp_init()`.

oschuett:
>  I used current master/main revision of libgrpp which now requires libgrpp_init().

I've uploaded https://www.cp2k.org/static/downloads/libgrpp-main-20231225.zip.

oschuett:
Unfortunately, the floating point exception in `regtest-ecp-2/SbH3_def2_gapw.inp` still [occurs](https://storage.googleapis.com/cp2k-ci/run-cp2k-sdbg-6734680b_report.txt) with the latest libgrpp.

Since all other tests passed, my guess would be that this test is somehow numerically unstable. 

@abussy what do you think?

abussy:
Mmh, it might be. When I added `libgrpp` for ECP potentials, I did not use the gradients provided by the library because of numerical instabilities. Instead, I calculated integrals with quantum number l +- 1 and  used existing routines to get gradients from integrals (https://github.com/cp2k/cp2k/blob/fdc09aa00adcd72131a1e4c0a7097ddc39c57901/src/libgrpp_integrals.F#L664-L665).

However, I do not remember having issues with ECP integrals. Also I do not quite understand the connection between `libxsmm` and the present error message.

oschuett:
My theory is that the new libxsmm produces slightly different results, i.e. numerical noise. Due to bad luck we're now taking [this branch](https://github.com/aoleynichenko/libgrpp/blob/6e63e88f75385b811837efbbb143b2dcd83b00e8/libgrpp/specfunc_scaled_mod_sph_bessel.c#L83), which returns [zero](
https://github.com/aoleynichenko/libgrpp/blob/6e63e88f75385b811837efbbb143b2dcd83b00e8/libgrpp/specfunc_scaled_mod_sph_bessel.c#L92), which then leads to the [division by zero](https://github.com/aoleynichenko/libgrpp/blob/6e63e88f75385b811837efbbb143b2dcd83b00e8/libgrpp/screening.c#L356).

It might be that this is not an actual bug in libgrpp because it could still discard the `NaN` later on. While I think that floating point exceptions are an invaluable debugging tool, not all library authors feels the same way. We therefore had to disable them for some libraries e.g. libvori: 


https://github.com/cp2k/cp2k/blob/fdc09aa00adcd72131a1e4c0a7097ddc39c57901/src/voronoi_interface.F#L426-L430

abussy:
That's quite an unfortunate design. I am not sure whether libgrpp discards `NaNs` by itself though, as I seem to recall that the gradient issues I encountered where `NaN` related as well.

It might be worth trying the same approach as libvori, and check whether the regtests are affected.

oschuett:
> It might be worth trying the same approach as libvori, and check whether the regtests are affected.

Unfortunately, that will only work if the NaN is not passed back to CP2K. It seems this is an actual bug that needs to be fixed in libgrpp. I've opened [an issue](https://github.com/aoleynichenko/libgrpp/issues/5).

To unblock this PR in the meantime, we could try to tweak the [SbH3_def2_gapw.inp](https://github.com/cp2k/cp2k/blob/master/tests/QS/regtest-ecp/SbH3_def2_gapw.inp) input, e.g. by changing the grid settings, the box size, or moving an atom slightly.

hfp:
I have disabled `regtest-ecp/SbH3_def2_gapw.inp` and `regtest-ecp-2/SbH3_def2_gapw.inp` for the time being. I can log an issue as a reminder.

This PR brings LIBXSMM for ARM64, i.e., `m_cpuid_static`, `m_cpuid`, and `m_cpuid_name`.

oschuett:
> I have disabled regtest-ecp/SbH3_def2_gapw.inp and regtest-ecp-2/SbH3_def2_gapw.inp for the time being. I can log an issue as a reminder.

Yes, I think  that's reasonable. From what I can tell this is a bug in libgrpp, which would have surfaced inevitably.

oschuett:
> This PR brings LIBXSMM for ARM64, i.e., m_cpuid_static, m_cpuid, and m_cpuid_name.

This is awesome! You can enable it for our tests here:
https://github.com/cp2k/cp2k/blob/9b99305b54e559eedadc767cb35c3c2eab9e6b69/tools/docker/generate_dockerfiles.py#L81

hfp:
Thank you for the reply! I will go ahead and merge when ready. Next commit will fix the precommit/fileprops, and I will remove `--with-libxsmm=no` for all AARCH64 dockers/tests (doing so as well for Darwin will be a separate PR).

hfp:
Also, you can upload LIBXSMM to CP2K.org once more when merged (at your own pace ;-). I have incorporated `download_pkg_from_urlpath` in `tool_kit.sh` (which implements `download_pkg_from_cp2k_org`). However, `install_libxsmm.sh` now downloads from CP2K.org (as before) but can fallback to GitHub (using `download_pkg_from_urlpath`).

hfp:
Ok, I will take a look at the ARM64 status like failures. Might be SMMs with transpose are not ready. If no easy fix, I will revert testing on ARM64, merge, and fix in separate PR.

hfp:
This PR is now ready, just waiting for the ARM64 test to finish.

alazzaro:
It seems that the dashboard is still using 1.17 libxsmm in some tests (for example, the Ubuntu tests)...

hfp:
ACK, I came to the same conclusion. I wonder what needs to change. It's somehow CMake based (which was green for the PR).

oschuett:
Yes, I've moved many of our test off the toolchain to simplify and speed them up. So, it would be good if we could also support the libxsmm version that is distributed via Ubuntu.

hfp:
> Yes, I've moved many of our test off the toolchain to simplify and speed them up. So, it would be good if we could also support the libxsmm version that is distributed via Ubuntu.

Ok, normally I would add a script like `tools/docker/scripts/install_dbcsr.sh` and update the Docker images accordingly. Just realized current version is from upstream. For the latter, I have to touch the code again and conditionally go for the version mark.

hfp:
See #3676

PR #3592, opened 2024-07-30
Abbreviate journal names, add missing pages, introduce upper/lower case

JWilhelm:


hfp:
It's not this PR in particular, but the overall effort (a nice one!), I wonder why line-breaks would not break at certain characters like "space", "dash", or perhaps a whole class of naturally breakable (dash-able) characters. The current implementation appears to break (words) at any character, in particular inside of words. I am not asking for language awareness like when or not to insert dashes but just like breaking in a more reasonable fashion.

oschuett:
>  The current implementation appears to break (words) at any character, in particular inside of words.

Yes, that's a fair point. The [current rendering](https://github.com/cp2k/cp2k/blob/0698928559b1d04870232912a82e7c64b4f5fbdf/src/common/reference_manager.F#L326) of the references for the output file is a bit crude. I'll look into this.

PR #3597, opened 2024-07-31
add optical properties from BSE and improve cutoff handling

MGraml:


oschuett:
I belive the deallocations of  `fm_A` and `fm_B` are missing.

Furthermore, I found these double-frees of `fm_struct_A` and `fm_struct_W`:

https://github.com/cp2k/cp2k/blob/dbf22e368d5f244d866fe51431878aaf89130f25/src/bse_full_diag.F#L190-L191

MGraml:
Thank you Ole for having a look at it!

fm_A and fm_B should be fine, they are deallocated outside of the routines, e.g. in 
https://github.com/cp2k/cp2k/blob/dbf22e368d5f244d866fe51431878aaf89130f25/src/bse_main.F#L210

But there are dipole matrices from 
https://github.com/cp2k/cp2k/blob/dbf22e368d5f244d866fe51431878aaf89130f25/src/bse_util.F#L1448-L1450
, which are not deallocated. I will submit the patch for that and the double free of the structs shortly.

PR #3605, opened 2024-08-05
Add -D__MPI_F08 flag

mkrack:
OpenMPI 5.0.5 does not work with macOS (Darwin-gnu-arm64)

### DIFF
--- a/arch/Darwin-gnu-arm64.psmp
+++ b/arch/Darwin-gnu-arm64.psmp
@@ -100,6 +100,7 @@ CFLAGS         := -O2 -fopenmp -fopenmp-simd -ftree-vectorize -funroll-loops -g 
 
 DFLAGS         := -D__parallel
 DFLAGS         += -D__ACCELERATE
+DFLAGS         += -D__MPI_F08
 DFLAGS         += -D__HAS_IEEE_EXCEPTIONS
 DFLAGS         += -D__MAX_CONTR=$(strip $(MAX_CONTR))
 DFLAGS         += -D__NO_STATM_ACCESS


oschuett:
I've upgraded OpenMPI because of [problems on x86](https://dashboard.cp2k.org/archive/openmpi-psmp/commit_08a39e73b57ed5cb37e1414085188a8897d601ed.txt).

mkrack:
> I've upgraded OpenMPI because of [problems on x86](https://dashboard.cp2k.org/archive/openmpi-psmp/commit_08a39e73b57ed5cb37e1414085188a8897d601ed.txt).

The macOS regression test fails badly with OpenMPI 5.0.5 including one of the unit tests. We could try 5.0.4 hoping that both problems do not appear.

oschuett:
According to the [changelog](https://docs.open-mpi.org/en/v5.0.x/release-notes/changelog/v5.0.x.html#open-mpi-version-v5-0-5), OpenMPI 5.0.4 won't event compile on macOS.

Also, I'm kinda surprised that it fails so badly and in a variety of places. To me this smells more like a linking problem.

Another thing you could try is to enable `-D__MPI_F08`.

mkrack:
I will try `-D__MPI_F08`.
In general, I suggest to stay sync with Homebrew which updates rather quickly to newer versions. That's what I have done recently. Currently, [Homebrew](https://formulae.brew.sh/formula/open-mpi) still uses 5.0.3, most likely for good reason.

mkrack:
Adding `-D__MPI_F08` flag seems to solve the problem.

PR #3617, opened 2024-08-08
look for different scalapack names under linux

jpoto:
In this case it is a simple name check with ifs, alternatively we could identify the MPI and use that to look for it
(https://gitlab.com/dirac/dirac/-/blob/master/cmake/custom/IdentifyMPI.py?ref_type=heads).

### DIFF
--- a/cmake/modules/FindSCALAPACK.cmake
+++ b/cmake/modules/FindSCALAPACK.cmake
@@ -52,7 +52,13 @@ if(NOT CP2K_CONFIG_PACKAGE)
     # a vendor specific blas/lapack/scalapack implementation
 
     if(NOT CP2K_SCALAPACK_FOUND)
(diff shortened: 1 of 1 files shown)


oschuett:
Thanks for your PR!

I think it would be best to explicitly identify the MPI library because any hard-coded search order will inevitably fail for someone. Ideally we don't need to add a separate Python script, but can instead implement the logic within CMake. We already do something along those lines [here](https://github.com/cp2k/cp2k/blob/7adf92c51dad654ff821dedbc0c0279d14cd0091/CMakeLists.txt#L458C9-L458C43).


@mtaillefumier, what do you think?

mtaillefumier:
I am fine with this PR as well. But i agree with @oschuett  that it is better to identify which mpi is used instead of the specific scalapack naming convention. I would also avoid having a custom script for checking which mpi is used. Let cmake do the job for us.

PR #3619, opened 2024-08-09
Fix GCC14

fstein93:
Testing with GCC14 revealed some bugs fixed by this PR.


fstein93:
It does not seem that the failing tests are related to this PR.

PR #3625, opened 2024-08-12
Initialize DBCSR_related variables

fstein93:
After the addition of default initializers in DBCSR, we can default initialize the variables in CP2K.


oschuett:
Note that I've recently introduced yet another [wrapper around DBCSR](https://github.com/cp2k/cp2k/blob/f3901f63c685ea015a51ff90b61c1cf88bfbb70a/src/dbx/cp_dbcsr_api.F#L262).

oschuett:
This PR broke a bunch of builds - interestingly not all of them though.

fstein93:
I am not sure what exactly is breaking them. The error message tells me that the compiler does not find the default initializers introduced in the latest DBCSR version. To me, it seems that they pick up the wrong DBCSR version. Can you check that?

alazzaro:
> I am not sure what exactly is breaking them. The error message tells me that the compiler does not find the default initializers introduced in the latest DBCSR version. To me, it seems that they pick up the wrong DBCSR version. Can you check that?

So, my fault, I haven't updated the[ toolchain](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/stage9/install_dbcsr.sh). We are on v2.7 now.

alazzaro:
And since we are there, I would recommend to remove the DBCSR submodule to avoid such a situation...

oschuett:
> So, my fault, I haven't updated the [toolchain](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/stage9/install_dbcsr.sh). We are on v2.7 now.

Actually, we're not yet using the toolchain to install DBCSR, we are using [this script](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/install_dbcsr.sh).

So, don't blame yourself. It's truly a mess (mostly of my creation).

> I would recommend to remove the DBCSR submodule to avoid such a situation...

Yes, I have this high up on my list.

mkrack:
It seems that all failing regression tests have a cmake build of cp2k in common. The tests using make still work.

oschuett:
Indeed. Hopefully #3642 fixed the issue. I've just triggered another round of the daily dashboard runs.

PR #3630, opened 2024-08-15
Workarounds for the ICEs of the Intel ifx compiler

mkrack:
The internal compiler errors (ICEs) of the Intel ifx compiler (<=2024.2.1) disappear if two OpenMP clauses are excluded. Properly matching OpenMP END clauses seem to help, too.
@hfp Maybe, that helps the ifx compiler developers.
Unfortunately, the CP2K icx/ifx binary still shows runtime errors in dbcsr_mm.F:588.

### DIFF
--- a/src/deepmd_wrapper.F
+++ b/src/deepmd_wrapper.F
@@ -134,11 +134,11 @@ CONTAINS
       MARK_USED(coord)
       MARK_USED(atype)
       MARK_USED(cell)
-      MARK_USED(energy)
-      MARK_USED(force)
-      MARK_USED(virial)
-      MARK_USED(atomic_energy)
-      MARK_USED(atomic_virial)
+      energy = 0.0_dp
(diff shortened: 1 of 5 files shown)


hfp:
ACK, runtime is on top of ICEs. It is very helpful to see the workarounds you found!

* Parallel DO without END clause is an issue I will submit right away; should be supported.
* The other two issues are hopefully just the same cause (WORKSHARE).

hfp:
I prepared #3630 to not only get rid of misusing `__MKL` but to better recognize the workarounds.

hfp:
>  Unfortunately, the CP2K icx/ifx binary still shows runtime errors in dbcsr_mm.F:588.

What's the trick with the oneAPI test passing given the runtime error?

hfp:
Ahh, it's IFORT ;-)

PR #3632, opened 2024-08-16
IFX: use __INTEL_LLVM_COMPILER instead of __MKL

hfp:
* Note: it is common to use MKL even for non-Intel compiler


hfp:
I will update the `__INTEL_LLVM_COMPILER` condition as soon as the fix is released (version will be "2025.1").

PR #3637, opened 2024-08-20
Fix behavior of RTP/EMD restart file

Anyborr:
(Resubmitting this PR to restart checks)

* turn off any instantaneous perturbations which should only be applied once (deltakick)
* disable APPLY_WFN_MIX_INIT_RESTART explicitly (doesnt change behavior since the restart file also updates INITIAL_WFN to RT_RESTART, but makes the file more clear).
* update WFN_RESTART_FILE_NAME to be the .rtpwfn file generated by the previous RTP run.

### DIFF
--- a/src/motion/rt_propagation.F
+++ b/src/motion/rt_propagation.F
@@ -26,6 +26,9 @@ MODULE rt_propagation
                                               cp_to_string
    USE cp_output_handling,              ONLY: cp_add_iter_level,&
                                               cp_iterate,&
+                                              cp_p_file,&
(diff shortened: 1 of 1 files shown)


fstein93:
If you would like to test a branch after changing or adding a commit while a PR is opened, push your changes with `git push --force`.

Anyborr:
Thanks for the tip Frederick. I think in this instance there was an issue with the testing server which caused it to fail. I didn't change any of the code for the new submission so I assumed force pushing the exact same commit again wouldn't restart the tests. Hence making a new submission.

PR #3648, opened 2024-08-23
showing data folder for tests

jpoto:
show the way to set data folder


oschuett:
I'm wondering if this even merits a flag. Maybe we should just always show that message?

PR #3654, opened 2024-09-04
Accommodate toolchain for Alps

hfp:
* OpenBLAS
  - OpenBLAS detects Alps CPU as NeoverseV2 (LIBCORE), which is suddenly not in the TargetList of OpenBLAS
  - Building and installing OpenBLAS is now more robust relying on the DYNAMIC_ARCH feature.
  - On ARM in particular, "generic" was hard-coded as TARGET=NEHALEM.
  - The USE_THREAD setting is not necessary for CP2K (plain PThread flavor as opposed to OpenMP).
* MPICH/2
  - Apply --without-slurm as it otherwise is likely to meet an incompatible Slurm include file.
  - Previously: compilation error on Alps-like self-hosted SuperMicro system (w/ Ubuntu LTS).

### DIFF
--- a/tools/toolchain/scripts/stage1/install_mpich.sh
+++ b/tools/toolchain/scripts/stage1/install_mpich.sh
@@ -56,7 +56,7 @@ case "${with_mpich}" in
         MPICC="" \
         FFLAGS="${FCFLAGS} ${compat_flag}" \
         FCFLAGS="${FCFLAGS} ${compat_flag}" \
-        --without-x \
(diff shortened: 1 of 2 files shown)


hfp:
The next (forced) commit will just fix the precommit failure (formatting a Shell script). By then, the required regtests as well as ARM64 test have passed with the initial commit. I will not retest the ARM64-test (for time and cost reason), and the PR should be ready for merge.

PR #3655, opened 2024-09-04
Update fprettify git submodule

mabraham:
The fprettify project has been moved, which breaks the git clone behavior

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,5 +4,5 @@
 	branch = master
 [submodule "tools/prettify/fprettify"]
 	path = tools/prettify/fprettify
-	url = https://github.com/pseewald/fprettify.git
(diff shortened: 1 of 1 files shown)


mabraham:
Context: GROMACS project would like to build CI containers containing a pre-built CP2K for it to link against, starting from v9.1 when the support was formally offered by CP2K. But a recursive git clone fails to check out fprettify because the project has moved. Updating the submodule definition in your repo might be useful for more people than GROMACS.

mabraham:
Thanks!

mkrack:
@mabraham It seems that the fix does not work because of a typo: `fortan`

mabraham:
Sorry! https://github.com/cp2k/cp2k/pull/3657

oschuett:
I've updated the url also in all later support branches.

PR #3656, opened 2024-09-04
Update fprettify git submodule

mkrack:
Fix URL name


oschuett:
I've removed the fprettify submodule in #3603.

mkrack:
Indeed, but this causes problems when downloading older CP2K releases (2024.1 and older) with git. Switching between branches can now cause a mess.

oschuett:
> Switching between branches can now cause a mess.

Yes, this is one of the problems with git submodules. Unfortunately, it was kind of inevitable because we were going to remove the dbcsr submodule later this year anyways.

On the bright side, afterwards we won't have to deal with submodules ever again.

PR #3663, opened 2024-09-06
Update arch files for new libxsmm version

mkrack:


hfp:
I wonder if the ARCH-files can automatically pickup the version installed by the toolchain. Perhaps I will do more frequent updates of LIBXSMM in the future or at least when enabling it for macOS as well.

hfp:
Anyway, let's not make it complicated. I will update the ARCH-files next time I am changing LIBXSMM's version.

hfp:
LGTM

mkrack:
> I wonder if the ARCH-files can automatically pickup the version installed by the toolchain. Perhaps I will do more frequent updates of LIBXSMM in the future or at least when enabling it for macOS as well.

Yes, that would make sense. I will look into it. On the other hand, cmake will probably soon make the toolchain obsolete.

mkrack:
> Anyway, let's not make it complicated. I will update the ARCH-files next time I am changing LIBXSMM's version.

After PR #3666, the arch files should stay in sync with the toolchain automatically.

PR #3664, opened 2024-09-06
Add missing libxsmmext in arch files

mkrack:


hfp:
Thank you. Yes, libxsmmf suddenly depends on libxsmmext (which was not the case with v1.17 and earlier). It is an issue I might be able to prevent in the future but it should have no harm to link against libxsmmext.

PR #3676, opened 2024-09-12
LIBXSMM: keep compatibility with v1.17

hfp:
* Updated ignored/internal definitions.
* Cleanup of base/kinds.F


hfp:
Dashboard is red for [Current Toolchain (sdbg)](https://storage.googleapis.com/cp2k-ci/dashboard_sdbg_report.txt):
> ------------------------------- Summary --------------------------------
Number of FAILED  tests 0
Number of WRONG   tests 0
Number of CORRECT tests 4140
Total number of   tests 4140
Summary: correct: 4140 / 4140; slow: 1; 18min
Status: FAILED

... which is due to a slow test like `QS/regtest-pod/furane-pbe-4A.inp` (76.76 sec).

oschuett:
Thanks a lot for adding the libxsmm backwards compatibility!

> ... which is due to a slow test like QS/regtest-pod/furane-pbe-4A.inp (76.76 sec).

Yes, this is odd. While the regtest script already re-runs slow tests, there is not perfect protection against false positives.

I'll re-submit the sdbg tests. In all likelihood we just had a very noisy neighbor on our cloud machine.

PR #3677, opened 2024-09-13
LIBXSMM: enabled for Darwin

hfp:
* Updated LIBXSMM.
* Updated URL.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -175,10 +175,11 @@ required.
 
 ### 2h. LIBXSMM (optional, improved performance for matrix multiplication)
 
-- A library for matrix operations and deep learning primitives: <https://github.com/hfp/libxsmm/>.
+- A library for matrix operations and deep learning primitives:
+  <https://github.com/libxsmm/libxsmm/>.
 - Add `-D__LIBXSMM` to enable it, with suitable include and library paths, e.g.,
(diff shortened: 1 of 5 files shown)


hfp:
I think a real or specific test is via [Dashboard](https://dashboard.cp2k.org/archive/darwin-gnu-arm64-psmp/index.html) and hence the changes are only tested after merging the PR.

mkrack:
@hfp Don't get misled by the green dashboard entry for macOS. The `LIBXSMM` build [failed](https://www.cp2k.org/static/regtest/trunk/Darwin-gnu-arm64/latest-openmpi-psmp/regtest-0) and thus the variable `LIBXSMM_VER` was not set by the CP2K toolchain and thus the variable `USE_LIBXSMM` in the Makefile was also not set which disables `LIBXSMM` when the CP2K binary for macOS is built.

hfp:
> @hfp Don't get misled by the green dashboard entry for macOS. The `LIBXSMM` build [failed](https://www.cp2k.org/static/regtest/trunk/Darwin-gnu-arm64/latest-openmpi-psmp/regtest-0) and thus the variable `LIBXSMM_VER` was not set by the CP2K toolchain and thus the variable `USE_LIBXSMM` in the Makefile was also not set which disables `LIBXSMM` when the CP2K binary for macOS is built.

ACK, and thank you!  I will make a 2nd attempt asap (PR).

PR #3678, opened 2024-09-17
Update Darwin arch files for Sequoia 15.0

mkrack:
- Use ninja from system
- Disable libxsmm again

### DIFF
--- a/arch/Darwin-gnu-arm64.psmp
+++ b/arch/Darwin-gnu-arm64.psmp
@@ -3,13 +3,13 @@
 # CP2K Darwin arch file for a parallel arm64 binary
 # (https://www.cp2k.org/howto:compile_on_macos)
 #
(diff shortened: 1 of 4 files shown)


hfp:
I think #3680 made it ready for Darwin.

hfp:
@mkrack can you test if LIBXSMM compiles now for Darwin on Apple silicon using the intended/your environment?

I had to resolve several issues like bad distinction between pure GNU based toolchain (from brew) vs Apple's fake-gcc (Clang based). Also, finding the macOS SDK is now more generic rather than depending on hard-coded path. Plus, I found libxsmmext was not compiled and linked with OpenMP support under macOS. Though, the purpose of libxsmmext is to separate things potentially benefiting from OpenMP from foundational code that can be kept free from OpenMP (or any other multicore runtime).

mkrack:
@hfp The `psmp` regression test is already running on my Apple M1. It compiles now with `libxsmm` and the tests are all `OK` so far. I will update the arch files as soon as the test is completed.

PR #3680, opened 2024-09-18
Updated LIBXSMM

hfp:
Follow up on #3677.


hfp:
( I need to revise the LIBXSMM version again. I think it's almost validated now as for a major release. Let's see. )

PR #3682, opened 2024-09-20
Avoid using sha256sum in Darwin

tsaihsiaoyi:
Apple introduced the sha256sum command in a security update (I guess) for Mac OS X around August to September, but it can't be used directly with 'sha256sum --check'. Instead, one has to avoid using it by checking the version.

### DIFF
--- a/tools/toolchain/scripts/tool_kit.sh
+++ b/tools/toolchain/scripts/tool_kit.sh
@@ -609,8 +609,11 @@ checksum() {
   local __shasum_command='sha256sum'
   # check if we have sha256sum command, Mac OS X does not have
   # sha256sum, but has an equivalent with shasum -a 256
-  command -v "$__shasum_command" > /dev/null 2>&1 ||
+  if command -v "$__shasum_command" > /dev/null 2>&1 && ! ${__shasum_command} --version 2>&1 | grep -q 'Darwin'; then
+    __shasum_command='sha256sum'
+  else
     __shasum_command="shasum -a 256"
+  fi
   if echo "$__sha256  $__filename" | ${__shasum_command} --check; then
     echo "Checksum of $__filename Ok"
   else
@@ -651,8 +654,11 @@ verify_checksums() {
 
   # check if we have sha256sum command, Mac OS X does not have
   # sha256sum, but has an equivalent with shasum -a 256
-  command -v "$__shasum_command" > /dev/null 2>&1 ||
+  if command -v "$__shasum_command" > /dev/null 2>&1 && ! ${__shasum_command} --version 2>&1 | grep -q 'Darwin'; then
+    __shasum_command='sha256sum'
+  else
     __shasum_command="shasum -a 256"
+  fi
 
   ${__shasum_command} --check "${__checksum_file}" > /dev/null 2>&1
 }
@@ -665,8 +671,11 @@ write_checksums() {
 
   # check if we have sha256sum command, Mac OS X does not have
   # sha256sum, but has an equivalent with shasum -a 256
-  command -v "$__shasum_command" > /dev/null 2>&1 ||
+  if command -v "$__shasum_command" > /dev/null 2>&1 && ! ${__shasum_command} --version 2>&1 | grep -q 'Darwin'; then
+    __shasum_command='sha256sum'
+  else
     __shasum_command="shasum -a 256"
+  fi
 
   ${__shasum_command} "${VERSION_FILE}" "$@" > "${__checksum_file}"
 }


mkrack:
The CP2K toolchain should always use with gnu under Darwin the `sha256sum` from `homebrew` in `/opt/hombrew/bin` and not the generic `shasum` in `/usr/bin` from Darwin. A fallback to `shasum` is therefore not desirable and thus I do not see the point of this PR.

hfp:
It can be also helpful to rely on GNU specific names like `gsed` and test-up `command -v gsed` first and then (if empty) fall-back to `sed`. Perhaps that's true for brew's sha256sum too vs Apple's sha256sum. Usually, brew symlinks the "compatible" name into PATH while have it backed by some "g*" variant.

hfp:
Generally, the SHA based checksum protects a user of CP2K against false bits. Primarily if cp2k.org is hooked up but also when the Toolchain downloads source packages from elsewhere. If the URL-based content as well as the SHA correspond, they can be still wrong (if, say the scripts are not original)! The whole chain of trust must not be compromised. An incorrect SHA may not prevent our self-hosted infra from running "contributed" bits (we should check this). In any case, harmful bits may run on the cloud when testing a contribution. It's unfortunate if basic tools are not compatible to the last bit like sha256sum. I can see the point of checking/confirming `sha256sum` to be based on GNU bits (like provided by brew in case of Darwin).

mkrack:
> It can be also helpful to rely on GNU specific names like `gsed` and test-up `command -v gsed` first and then (if empty) fall-back to `sed`. Perhaps that's true for brew's sha256sum too vs Apple's sha256sum. Usually, brew symlinks the "compatible" name into PATH while have it backed by some "g*" variant.

Indeed, `homebrew` provides also a gnu version `gsha256sum` of `sha256sum`, but by contrast to `gsed` which must be used instead of the Darwin's very old default version of `sed`, that was never needed for `sha256sum`.

hfp:
> > It can be also helpful to rely on GNU specific names like `gsed` and test-up `command -v gsed` first and then (if empty) fall-back to `sed`. Perhaps that's true for brew's sha256sum too vs Apple's sha256sum. Usually, brew symlinks the "compatible" name into PATH while have it backed by some "g*" variant.
> 
> Indeed, `homebrew` provides also a gnu version `gsha256sum` of `sha256sum`, but by contrast to `gsed` which must be used instead of the Darwin's very old default version of `sed`, that was never needed for `sha256sum`.

Sure, it was just an example. Anyway, the g* version and above sketched flow may be a way to go forward or to at least provide an error message like "missing GNU sha256sum - please use Brew" (or similar).

tsaihsiaoyi:
That's not what I mean to do. As you can see the comments here. https://github.com/cp2k/cp2k/blob/c1b10bd6526dca92a6deb944b6159f1d8b1af7ba/tools/toolchain/scripts/tool_kit.sh#L610-L611
Mac OS X doesn't have command `sha256sum` in the past, whether in `/usr/bin` or else. If one doesn't install coreutils by homebrew, he must use `shasum -a 256` instead. But now, in a few month ago, Apple introduce the command `sha256sum` in `/sbin`, but not in GNU. So the patch I made mean to check if system use the `sha256sum` in GNU or not. If `sha256sum` in Apple, which will have Darwin in version, then abandon it, use `shasum -a 256` instead.

mkrack:
We do not support Mac OS X any longer and I am not sure if we ever did so properly using the toolchain. Mac support started with macOS 11 (Big Sur) and macOS 11 and 12 are even [not supported by Apple](https://en.wikipedia.org/wiki/MacOS) any longer. I do not think that we should add toolchain fixes for unsupported OS versions.

tsaihsiaoyi:
In that case, fix `Darwin-gnu-arm64.*` will be better than fix toolchain, right?

mkrack:
> In that case, fix `Darwin-gnu-arm64.*` will be better than fix toolchain, right?

The Darwin arch files enforce already the installation of the [coreutils](https://github.com/cp2k/cp2k/blob/master/arch/Darwin-gnu-arm64.psmp#L27) and `/opt/homebrew/bin` should precede `/sbin` in the `PATH`.

tsaihsiaoyi:
But `coreutils` only contains gsha256sum but not sha256sum. Maybe adding `alias sha256sum=gsha256sum` would be better?

mkrack:
> But `coreutils` only contains gsha256sum but not sha256sum. Maybe adding `alias sha256sum=gsha256sum` would be better?

There shouldn't be an issue, because `/opt/homebrew/bin` provides `sha256sum` and `/opt/homebrew/bin` should appear first in `$PATH`.

tsaihsiaoyi:
According to [coreutils](https://formulae.brew.sh/formula/coreutils) on homebrew, it requires manual operation.

> Commands also provided by macOS and the commands dir, dircolors, vdir have been installed with the prefix "g".If you need to use these commands with their normal names, you can add a "gnubin" directory to your PATH with:    PATH="$HOMEBREW_PREFIX/opt/coreutils/libexec/gnubin:$PATH"

`/opt/homebrew/bin` won't provide `sha256sum` by itself.

mkrack:
@tsaihsiaoyi Thanks for the patch.

PR #3684, opened 2024-09-20
Enable libxsmm with Darwin (see #3678)

mkrack:


hfp:
Nice, this worked out well!

PR #3689, opened 2024-09-23
Add Z-matrix formalism for linear response

rangsimanketkaew:
Hi there,

We would like to submit a pull request for the implementation of Z-matrix formalism as presented in [this paper](https://pubs.aip.org/aip/jcp/article/81/11/5031/87334/On-the-evaluation-of-analytic-energy-derivatives), which is a method that can speed up the DFPT (linear response) calculation. Z-matrix can reduce the computational complexity of the linear response calculations from 3N to 3, where N is the number of atoms in a system.

The users can turn on the Z-matrix method using the keyword `Z_MATRIX_METHOD` under the `DCDR`'s linear response deck. The regtest files for this implementation can also be found in `cp2k/tests/QS/regtest-dcdr/` folder.

Regarding the test, we have performed the compilations and full regtest using `sdbg`, `pdbg`, and `psmp` with no errors or memory leaks detected. Prettifying has also been done.

We believe that this contribution would benefit the CP2K's users.

Best regards,
Ravi (@ravikumar381709) and Rangsiman


juerghutter:
Thank you

PR #3694, opened 2024-09-25
Add output when SIRIUS is not converging

### LINKED ISSUE #3690: IGNORE_CONVERGENCE_FAILURE for SIRIUS SCF
This is rather a suggestion than an issue.
A parameter similar to IGNORE_CONVERGENCE_FAILURE in the DFT/SCF section is necessary for SIRIUS calculations. 
Currently, if SIRIUS SCF is converge after N iterations, the following message is printed:
[find] converged after N SCF iterations!
Otherwise, nothing is printed and the calculations (e.g. cell optimization) continue. 
I think a warning message should be printed if SCF is not converged and a parameter similar to  IGNORE_CONVERGENCE_FAILURE is also needed. 
Thanks!

mtaillefumier:
this PR should fix #3690


oschuett:
The `SIRIUS/regtest-1/Fe-upf.inp` test is taking a really long time. Is there any chance you could shorten it somehow?

PR #3696, opened 2024-09-25
Fixed compilation error about unknown fpp directive

hfp:


hfp:
( This came up with Intel Fortran Compiler )

PR #3691, opened 2024-09-23
Real time Bethe-Salpeter Propagation

StepanMarek:
We introduce the possibility of running the [real time Bethe-Salpeter](http://dx.doi.org/10.1103/PhysRevB.84.245110) propagation for the determination of the polarizability/absorption spectra of molecules.

This code enables an independent method for deriving formally identical results to the Casida equation based BSE approach, with possibilities for extension to stronger fields/time-dependent phenomena.

We have created elementary documentation, ran the regression testing and added new regtests, which were also checked for leaks using the valgrind tool. Precommit checks were run as well.

We plan to further develop and evaluate the performance of this method, especially compared to the Casida equation BSE approach, as we hope that our method will have a better memory scaling for larger systems. Nevertheless, we believe the state of the code right now is sufficient for an introduction to the CP2K community.

PR #3695, opened 2024-09-25
RPA based SIGMA Functionals

Raviraj26:
Added a support for Sigma functionals which can be accessed as a new keyword `SIGMA_FUNCTIONAL` in the `RI_RPA`-section.

PR #3698, opened 2024-09-26
RTBSE : Fixing dashboard issues from #3691

StepanMarek:
Fixing coding conventions

### DIFF
--- a/src/emd/rt_bse_types.F
+++ b/src/emd/rt_bse_types.F
@@ -247,6 +247,7 @@ MODULE rt_bse_types
       ! Workspace required for hartree_pw
       TYPE(dbcsr_type)                                          :: v_dbcsr = dbcsr_type(), &
                                                                    w_dbcsr = dbcsr_type()
+#if defined(__GNUC__) && defined(__GNUC_MINOR__) && CPVERSION(9, 5) > CPVERSION(__GNUC__, __GNUC_MINOR__)
       TYPE(dbt_type)                                            :: screened_dbt, &
                                                                    sigma_dbt, &
                                                                    greens_dbt, &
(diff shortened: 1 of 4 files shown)


fstein93:
There is also an issue with the GCC8 tester. This is related to a bug/missing feature of the earlier GCC compilers. This can be solved by not default-initializing the dbt_type-components if a GCC-compiler of version 9.5 or earlier is used
```
#if CPVERSION(9, 5) > CPVERSION(__GNUC__, __GNUC_MINOR__)
TYPE(dbt_type) :: my_dbt_object
#else
TYPE(dbt_type) :: my_dbt_object = dbt_type()
#endif
```
This will not affect the convention checker as it runs a later GCC compiler.

hfp:
If I am not mistaken, `CPVERSION(9, 5) > CPVERSION(__GNUC__, __GNUC_MINOR__)` may catch any non-GNU compiler as well, which is pure luck for Intel Compiler (apparently also affected by the issue). This is because FPP may expand `__GNUC__` and `__GNUC_MINOR__` to `0` (zero, if not defined). Hence, such check should normally be complemented by `defined(__GNUC__) && defined(__GNUC_MINOR__)` in addition to the actual version-check.

src/emd/rt_bse_types.F
```diff
@@ -247,13 +247,23 @@ MODULE rt_bse_types
...
+                                                                   t_3c_work2_RI_AO__AO
+#else
       TYPE(dbt_type)                                            :: screened_dbt = dbt_type(), &
                                                                    sigma_dbt = dbt_type(), &
                                                                    greens_dbt = dbt_type(), &
                                                                    t_3c_w = dbt_type(), &
                                                                    t_3c_work_RI__AO_AO = dbt_type(), &
                                                                    t_3c_work_RI_AO__AO = dbt_type(), &
                                                                    t_3c_work2_RI_AO__AO = dbt_type()
+#endif
```
StepanMarek, line 266: Many thanks to everyone for the discussion - I included the following preprocessor if statement in emd/rt_bse_types.F and a similar one in post_scf_bandstructure_types.F.

Let me know whether this is ok, I will also have to wait for the checks to rerun.

hfp: LGTM. If Intel Compiler is still affected, I will fix it by adding more conditions ;-). Also, I may introduce something like `CPVERSION_CHECK` which does the comparison plus check of the definition. However, this should not worry you as it's a distinct topic.

src/rpa_sigma_functional.F
```diff
@@ -731,15 +731,15 @@ INTEGER FUNCTION intervalnum(x_coord, n, sigma) RESULT(inum)
...
+      IF (sigma <= 0.0_dp) CPABORT('intervalnum: sigma should be positive')
 
       inum = -1
       IF (sigma > x_coord(n)) inum = n
       DO i = 1, n - 1
          IF ((sigma > x_coord(i)) .AND. (sigma <= x_coord(i + 1))) inum = i
       END DO
 
-      IF (inum == -1) STOP 'interval: something was wrong'
+      IF (inum == -1) CPABORT('interval: something was wrong')
```
StepanMarek, line 742: One more comment, I changed the STOP in this file to CPABORT, in order to satisfy the conventions checker. I assume this is fine with @Raviraj26 (this was introduced in #3695 )?

Raviraj26: The inputs are highly appreciated. Thank you

PR #3699, opened 2024-09-26
Introduced CPVERSION_CHECK macro

hfp:
See https://github.com/cp2k/cp2k/pull/3698#issuecomment-2376703705:
- Adjusted uses of plain CPVERSION accordingly.
- Introduced FTN_NO_DEFAULT_INIT workaround.


hfp:
> Duration threshold (2x 95th %ile): 30.89 sec
Found 1 slow tests (28 suppressed):
    QS/regtest-pod/furane-pbe-4A.inp                                                 (  70.38 sec)

Will restart SDBG tests.

hfp:
> Duration threshold (2x 95th %ile): 32.50 sec
Found 1 slow tests (24 suppressed):
    QS/regtest-ps-implicit-1-3/Ar_mixed_aa_planar-osc-field.inp                      (  61.49 sec)

Apparently, the deviations are shaky at least with what we get from the cloud.

I will restart again (`sdbg`).

oschuett:
It seems for some reason the [SLOW_TESTS_SUPPRESSIONS](https://github.com/cp2k/cp2k/blob/master/tests/SLOW_TESTS_SUPPRESSIONS) are being ignored. Very strange!

oschuett:
Ok, found it. I made a bug in the retry / suppression logic. I'll fix it.

hfp:
I will rebase when the fix is available. Thank you, Ole!

PR #3711, opened 2024-10-09
Fix segmentation fault when performing CDFT-MD with many constraints

chrisahart:
Fix for segmentation fault when performing CDFT-MD with many constraints, as reported by Leili Rassouli on the CP2K Google Groups
https://groups.google.com/g/cp2k/c/jaQdteFzSDc/m/trqpsgv9CQAJ

### DIFF
--- a/src/qs_cdft_methods.F
+++ b/src/qs_cdft_methods.F
@@ -755,8 +755,8 @@ CONTAINS
          CALL pw_set(cdft_control%group(igroup)%hw_rho_total_constraint, 1.0_dp)
 
          IF (igroup == 1) THEN
-            CALL auxbas_pw_pool%create_pw(cdft_control%group(1)%hw_rho_total)
-            CALL pw_set(cdft_control%group(1)%hw_rho_total, 1.0_dp)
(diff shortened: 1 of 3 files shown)


schulkov:
Thank you for fixing this issue.

As `hw_rho_total` component of `cdft_group_type` derived type is used exclusively within `hirshfeld_constraint_low()` subroutine, would it be better to simplify the code a little further by removing `hw_rho_total` from `cdft_group_type` and declaring it as a local variable?

It should get rid of the magic index in the expression `cdft_control%group(1)%hw_rho_total` at least.

chrisahart:
Thanks for the helpful suggestion Sergey. 

I've now moved `hw_rho_total` from `cdft_group_type` to `cdft_control_type`, as the promolecular density is the same for all CDFT groups so it should not be stored per CDFT group. 

`hw_rho_total` is also used in file qs_cdft_utils.F so keeping it in the CDFT control structure seems appropriate. If you would prefer it defined as a local variable and then passed to qs_cdft_utils.F I'm happy to do so

PR #3722, opened 2024-10-12
CMake should not check for libxcf90 since only libxcf03 is used.

susilehtola:
Although CP2K is correctly only using the interface provided by libxcf03, for some reason the CMake code tries to find libxcf90 and wants to link it in as well. This makes no sense, since libxcf90 and libxcf03 are different beasts, moreover, libxcf90 has been dropped in libxc 7.0.0.

This patch is necessary to build cp2k against up-to-date releases of Libxc.


oschuett:
Thanks a lot for contributing. LGTM. Please run `make pretty` and you're good to go.

susilehtola:
> Thanks a lot for contributing. LGTM. Please run `make pretty` and you're good to go.

done

PR #3729, opened 2024-10-17
FFT_TOOLS: thread-safety insurance (CPASSERT) and OMP ATOMIC

hfp:
The motivation for this PR stems from occasional error messages like:

> Fortran runtime error: Pointer actual argument 'fft_scratch_new' is not associated

In any case, the fft_tools manipulated global/SAVE variables without protection for data races.


hfp:
@JWilhelm I discovered this when running farming jobs.

PR #3736, opened 2024-10-22
Fix pw_derive with the Intel compilers

fstein93:
This PR fixes a few issues with the Intel compiler and an uninitialised variable.

### DIFF
--- a/src/pw/pw_methods.F
+++ b/src/pw/pw_methods.F
@@ -1014,7 +1014,7 @@ CONTAINS
                      LOGICAL, INTENT(IN), OPTIONAL                      :: just_sum, local_only
                      REAL(KIND=dp)                                      :: integral_value
 
-                     CHARACTER(len=*), PARAMETER                        :: routineN = 'pw_integral_ab'
+                     CHARACTER(len=*), PARAMETER                        :: routineN = 'pw_integral_ab_${kind}$_${kind2}$_${space}$'
 
                      INTEGER                                            :: handle, loc_sumtype
                      LOGICAL                                            :: my_just_sum, my_local_only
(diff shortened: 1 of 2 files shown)


src/pw/pw_methods.F
```diff
@@ -1034,6 +1034,8 @@ FUNCTION pw_integral_ab_${kind}$_${kind2}$_${space}$ (pw1, pw2, sumtype, just_su
                      my_just_sum = .FALSE.
                      IF (PRESENT(just_sum)) my_just_sum = just_sum
 
+WRITE(*,*) "DEBUG pw_integral_ab Start", loc_sumtype
```
oschuett: Left over. See also below.

hfp:
Just FYI, [this](https://github.com/cp2k/dbcsr/pull/859#issuecomment-2416460473) is surprisingly the main show stopper for Intel Fortran Compiler (not IFORT/classic but next-gen). If it's merged and eventually used by CP2K, IFX is ready to go.

fstein93:
Thank you for the pointer. My motivation was an issue a user observed. Thereby, I introduced the new notation to access real and imaginary parts of complex numbers via `a%re` and `a%im`. This is more handy than the former `REAL(a, KIND=dp)` and `AIMAG(a)`, respectively. Finally, I have found an uninitialised variable.

mkrack:
Based on my tests, this commit causes again an ICE in `pw_methods.F90` using the new Intel IFX compiler (classic ifort is working).

hfp:
> Based on my tests, this commit causes again an ICE in `pw_methods.F90` using the new Intel IFX compiler (classic ifort is working).

I fixed this with #3752 (one more ICE). FYI, all ICEs are gone with 2025 release (in roll-out already depending on distro).

PR #3737, opened 2024-10-23
Simplified calling dbcsr_init_lib

hfp:
- Rely on POINTER to optionally initialize accdrv_active_device_id
- Reduces Lines of Code (LOC)
- Complements #3730


hfp:
Rely on a NULL-pointer to keep an argument optional despite of passing the argument (standard compliant).

PR #3742, opened 2024-10-24
Fix GCC8

fstein93:
Follow up to #3736 . I simplified the workaround as it is apparently related to a bug in a compiler. On my machine, with my Intel compiler (OneAPI 2024.2), the code still works as expected, with the code before #3736 it does not.

### DIFF
--- a/src/pw/pw_methods.F
+++ b/src/pw/pw_methods.F
@@ -1062,7 +1062,7 @@ CONTAINS
                         #:elif kind2[0]=="r"
                            integral_value = accurate_sum(REAL(pw1%array, KIND=dp)*pw2%array) !? complex bit
                         #:else
-                           integral_value = accurate_sum(pw1%array%re*pw2%array%re + pw1%array%im*pw2%array%im) !? complex bit
+                           integral_value = accurate_sum(REAL(CONJG(pw1%array)*pw2%array, KIND=dp))
                         #:endif
 
                      END IF
@@ -1687,7 +1687,7 @@ CONTAINS
 !$OMP PARALLEL DO DEFAULT(NONE) PRIVATE(i,tmp) SHARED(pw,omega_2)
                      DO i = 1, SIZE(pw%array)
                         tmp = EXP(-pw%pw_grid%gsq(i)*omega_2)
-                        pw%array(i) = CMPLX(pw%array(i)%re*tmp, pw%array(i)%im*tmp, KIND=dp)
+                        pw%array(i) = pw%array(i)*tmp
                      END DO
 !$OMP END PARALLEL DO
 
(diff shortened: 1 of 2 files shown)


src/pw/pw_methods.F
```diff
@@ -2186,7 +2259,11 @@ SUBROUTINE pw_smoothing(pw, ecut, sigma)
                      DO ig = 1, cnt
                         arg = (ecut - pw%pw_grid%gsq(ig))/sigma
                         f = EXP(arg)/(1 + EXP(arg))
+#if defined(__GNUC__) && defined(__GNUC__MINOR__) && CPVERSION_CHECK(9, 5, >, __GNUC__, __GNUC__MINOR__)
+                        pw%array(ig) = CMPLX(f*REAL(pw%array(ig), KIND=dp), f*AIMAG(pw%array(ig)), KIND=dp)
```
oschuett: Couldn't we just write `pw%array(ig) = f * pw%array(ig)`?

fstein93: I am not sure. There was a recent report in the Google group with a similar issue. This is why I have decided to rewrite similar parts of the pw_methods module in a similar more explicitly manner. It may work standard-wise but the Intel compiler complains in certain circumstances. I will double-check the code as I can reproduce the original issue locally.

hfp: Is it worth using conditional compilation just to use `%re` and `%im` as opposed to `REAL` and `AIMAG`? The test `defined(__GNUC__) && defined(__GNUC__MINOR__)` is actually part of  `CPVERSION_CHECK` and can be dropped - so, things get surprisingly short ;-).

fstein93: After double-checking, we have probably hit some bug in the compiler. I am not sure what exactly caused it but replacing the `WORKSHARE`-construct with a `DO`-loop and rewriting the three steps as one, should have done the job, too.

hfp:
@oschuett I wonder if somebody knows of tools like [this](https://github.com/cp2k/dbcsr/issues/857#issuecomment-2416094575). Dumped AST-info does not seem to cover a static call-tree other than building it (which I think is quite some effort). Currently, conventions cannot ask, e.g., if an OMP-parallel code is already inside of a parallel region.

oschuett:
I think, finding nested parallelism through a static code analysis would be very challenging.
For DBM I resorted to adding lots of [asserts like these](https://github.com/cp2k/cp2k/blob/ffaae80b0820c5efae69dddfbe24bcf3606c8998/src/dbm/dbm_matrix.c#L26).

We could also consider to disable it globally via `omp_set_max_active_levels(1)`, but that will presumably tank the performance of some libraries like OpenBLAS.

hfp:
Assertions are likely the way to go like `.NOT.  omp_in_parallel() .OR. 0 == omp_get_thread_num()` for relaxed cases. Though, "parallel in master" seems illegal (see https://github.com/cp2k/dbcsr/pull/859). With more Fortran compilers based on LLVM (ifx, flang), we get potentially the ability to plug tools into compilation and to have custom analysis far beyond what's possible with AST.

hfp:
Limiting active OMP levels is probably too active like no code is better than (micro-)manage rare cases.

PR #3750, opened 2024-10-28
Revised CPVERSION_CHECK and related checks

hfp:
- Ban using "defined" as part of a macro's definition
- Updated __INTEL_LLVM_COMPILER workarounds


hfp:
_If the defined operator appears as a result of a macro expansion, the C standard says the behavior is undefined._

( See https://gcc.gnu.org/onlinedocs/cpp/Defined.html )

PR #3752, opened 2024-10-29
Enable libxsmm with AMD again after #3750

mkrack:


hfp:
See https://github.com/cp2k/cp2k/pull/3750#issuecomment-2443793622

PR #3762, opened 2024-11-06
Get AiiDA tests working again (except two tests with restart)

mkrack:


oschuett:
Oh nice! It seems this pretty much fixes https://github.com/aiidateam/aiida-cp2k/issues/220.
The remaining failures are probably due to https://github.com/aiidateam/aiida-cp2k/issues/214.

PR #3766, opened 2024-11-08
OMPT: trace OpenMP constructs and check for issues

hfp:
- Intel OpenMP implemented an OpenMP 5.0 facility called OMPT for tracing OpenMP constructs.
- Intel OpenMP (libiomp5) is open source (libomp), and LLVM compilers provide necessary bits.
- Effectively, Intel Compiler (icx/icc) works out-of-the box (note: ifx/ifort is no prerequisite).
- GNU compiler shall be possible when linking libomp instead of libgomp. Further, currently
  defined(_OPENMP) && defined(__clang__) can be relaxed or smarter/runtime check).
- Currently only one issue is tracked: if a parallel region is opened inside of a master section.
- The OMPT facility has no runtime impact if not enabled, otherwise the impact shall be low.
- The code can be a starting point to enforce more sophisticated conventions.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1357,7 +1357,7 @@ list(
   xc/xc_xpbe_hole_t_c_lr.F
   xc/xc_xwpbe.F)
 
-list(APPEND CP2K_SRCS_C sockets.c base/machine_cpuid.c)
+list(APPEND CP2K_SRCS_C sockets.c base/machine_cpuid.c base/openmp_trace.c)
 
 set(CP2K_DBM_SRCS_C
     dbm/dbm_distribution.c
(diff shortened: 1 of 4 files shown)


hfp:
@mkrack @oschuett comments are welcome. It does not have to be merged (can keep it in my branch). It's ~145 LOC incl. some code comments.

oschuett:
This looks like very useful addition. Ideally, we'd run it regularly - maybe as part of the [Intel oneAPI test](https://github.com/cp2k/cp2k/blob/master/tools/docker/Dockerfile.test_intel-psmp).

hfp:
Perhaps a break-out discussion, but if desired I can look into:

* Using GNU toolchain but ensure an option with `-fopenmp` *not* in the link-line but supplying `-lomp` (instead of GOMP). Note, Intel OpenMP aka LLVM's OpenMP has a minimal layer of compatibility with GNU's OpenMP implementation.
* Implement more conventions (or analysis information we can grep).

Design-wise, I tend to *not* add more functions to the F-interface but rather introduce a "print issues" function taking an output unit. This way, changes in analysis functionality are covered on the C-code side. Some examples:

* Info about deepest nested parallel region along with captured callstack/symbol info.
* [More ideas]

hfp:
Verbosity levels: see https://github.com/cp2k/cp2k/pull/3773#issuecomment-2470417301

PR #3774, opened 2024-11-12
Tighten thresholds for 2 LS regtests, reorder Test dirs

juerghutter:


juerghutter:
Consistently get a slow test from
QS/regtest-smeagol-1/au-2x2x3-bulk.inp

PR #3781, opened 2024-11-19
BSE: new features and update manual

MGraml:
- add full polarizability tensor
- add Natural Transition Orbitals (NTOs) for TDA and ABBA
- add exciton descriptors for TDA and ABBA
- update manual on new features


src/bse_print.F
```diff
@@ -0,0 +1,820 @@
...
+         WRITE (unit_nr, '(T2,A4)') 'BSE|'
+         IF (exc_descr(1)%flag_TDA) THEN
+            WRITE (unit_nr, '(T2,A4,T7,A)') 'BSE|', &
+               'where c_n = < 𝚿_n | 𝚿_n > = 1 within TDA.'
+         ELSE
+            WRITE (unit_nr, '(T2,A4,T7,A)') 'BSE|', &
+               'where c_n = < 𝚿_n | 𝚿_n > deviates from 1 without TDA.'
+         END IF
+
+         ! WRITE (unit_nr, '(T2,A4)') 'BSE|'
```
fstein93: Are these WRITE statements still needed? Remove if not or wrap in `IF (.FALSE.) THEN ... END IF` for type checking.

src/fm/cp_fm_diag.F
```diff
@@ -882,6 +899,140 @@ SUBROUTINE cp_fm_syevx(matrix, eigenvectors, eigenvalues, neig, work_syevx)
...
+!> \param matrix_eigvl ...
+!> \param matrix_eigvr_t ...
+!> \param eigval ...
+!> \param info ...
+!> \author Maximilian Graml
+! **************************************************************************************************
+   SUBROUTINE cp_fm_svd(matrix_a, matrix_eigvl, matrix_eigvr_t, eigval, info)
+
+      TYPE(cp_fm_type), INTENT(IN)              :: matrix_a
+      TYPE(cp_fm_type), INTENT(OUT)             :: matrix_eigvl, matrix_eigvr_t
```
fstein93: I am not sure whether other diagonalization/linear algebra routines in the fm package create matrices inside (without the obvious exceptions). I would rather push the calls to cp_fm_create outside of this routine to make the allocation step explicit. This makes other developers aware that the calling routing is responsible to release the matrix later.

MGraml:
Thank you @fstein93 for the review. I implemented all your suggestions and also fixed the memory leaks in the last commit.

PR #3786, opened 2024-11-25
ELPA: avoid multiple initializations

hfp:
- Note: potentially ensuring the original "active device" is down to the library (ELPA),
  and CP2K is only responsible for toggling the active device for its own code.
  This note refers to a potential side-effect of (re-)initializing ELPA.
- Avoid potentially high cost of initialization (context creation, etc.)
- Initialization SAVE state does not require atomics (expected to be called from master).
- Harmonized CFM and FM interfaces.
- Code/format cleanup.


hfp:
( this is some pre-work before introducing ELPA for the implementation of the CFM interface )

PR #3770, opened 2024-11-11
Add CMake support for cuSOLVERMp

jpoto:
@oschuett Can you check it?

PR #3787, opened 2024-11-27
Use cusolverMpSygvd - a generalized eigenvalue solver

sliwowitz:
In `cp_fm_geeig`, instead of reducing the generalized eigenvalue problem into standard from on the CPU and then solving the standard form on the GPU using `cusolverMpSyevd`, solve the whole generalized problem on the GPU using `cusolverMpSygvd`. This gives a nice speedup for all problems which are large enough to benefit from using cuSOLVERMp in the original code. 

The original API for solving only the standard form on the GPU has been kept in the code, should there be a need to use it from some other routine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -371,3 +371,7 @@ tags
 # End of https://www.gitignore.io/api/python,fortran,c,c++,cuda,emacs,vim,cmake
 
 !tools/fedora/cp2k.spec
+
+# CLion directories
+/.idea/
+/cmake-build-*
--- a/src/fm/cp_fm_cusolver.c
+++ b/src/fm/cp_fm_cusolver.c
@@ -316,6 +316,163 @@ void cp_fm_diag_cusolver(const int fortran_comm, const int matrix_desc[9],
   MPI_Barrier(comm);
 }
 
+/*******************************************************************************
+ * \brief Driver routine to solve A*x = lambda*B*x with cuSOLVERMp sygvd.
+ * \author Jiri Vyskocil
+ ******************************************************************************/
+void cp_fm_diag_cusolver_sygvd(const int fortran_comm,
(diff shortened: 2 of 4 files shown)


sliwowitz:
The automatic checks passed, but I suppose none of test are actually testing a build + run with cuSOLVERMp enabled, right?

For the record, using the fixes in PR #3770, I'm building CP2K with the Nvidia HPC SDK like `cmake -DCMAKE_PREFIX_PATH="/home/vyskoc65/cp2k/lib/dbcsr;/opt/nvidia/hpc_sdk/Linux_x86_64/24.7/compilers/bin;/opt/nvidia/hpc_sdk/Linux_x86_64/24.7/math_libs;/opt/nvidia/hpc_sdk/Linux_x86_64/24.7/comm_libs/12.6/hpcx/latest/ucc;/opt/nvidia/hpc_sdk/Linux_x86_64/24.7/comm_libs/12.6/hpcx/latest/ucx" -DCP2K_USE_ACCEL="CUDA" -DCP2K_USE_CUSOLVER_MP=ON ..`

oschuett:
>... none of test are actually testing a build + run with cuSOLVERMp enabled, right?

That's correct! There were a couple of reasons:

- cuSOLVERMp is not included in the CUDA docker image that we're [using](https://github.com/cp2k/cp2k/blob/master/tools/docker/Dockerfile.test_cuda_P100#L6) and the [manual install](https://github.com/oschuett/cp2k/blob/cusolvermp/tools/docker/scripts/install_cusolvermp.sh) seemed brittle.
- Some regtests failed with cuSOLVERMp.
- We can only afford one CUDA dashboard tests and I didn't want to lose coverage of ELPA's CUDA backend.

In all likelihood things have improve over the last year. So, maybe you want to give it another try?

PR #3792, opened 2024-12-03
TREXIO support

stefabat:
This PR enables support for writing [TREXIO](https://github.com/TREX-CoE/trexio) files from a SCF run.
TREXIO is a library that support interoperability between quantum chemistry codes, especially for programs implementing post-SCF methods such as QMC.

The PR adds an external library dependency, which itself depends on HDF5.

Currently, we only support writing trexio files, and not reading them (e.g. to restart a calculation previously run with a different code support trexio).


stefabat:
Tests simply run and make sure that the code written is executed without mistakes, but do not test the genuinity of the trexio file content.
Tests with dbg executables fail because of a bug in HDF5 1.14.3 (see https://github.com/HDFGroup/hdf5/issues/3831). The problem is the flag `-ffpe-trap=invalid`.

INSTALL.md
```diff
@@ -491,6 +491,14 @@ DeePMD-kit - Deep Potential Molecular Dynamics. Support for DeePMD-kit can be en
 - Add `-D__SMEAGOL` to DFLAGS, `-I$(LIBSMEAGOL_DIR)/obj` to FCFLAGS and
   `-L$(LIBSMEAGOL_DIR)/lib -lsmeagol` to LIBS
 
+### 2z. TREXIO (optional, unified computational chemistry format)
```
stefabat, line 494: The section is numbered 2z. even though it exists already, but we have run out of letters...

oschuett:
> Tests with dbg executables fail because of a bug in HDF5 1.14.3...

The workaround for this is to temporarily disable the trapping:

https://github.com/cp2k/cp2k/blob/2de90c3493f93ab5519e4c813f6f1ead4f3a97fc/src/fm/cp_fm_diag.F#L621-L633

tools/toolchain/scripts/stage7/install_hdf5.sh
```diff
@@ -6,8 +6,8 @@
 [ "${BASH_SOURCE[0]}" ] && SCRIPT_NAME="${BASH_SOURCE[0]}" || SCRIPT_NAME=$0
 SCRIPT_DIR="$(cd "$(dirname "$SCRIPT_NAME")/.." && pwd -P)"
 
-hdf5_ver="1.14.2"
-hdf5_sha256="ea3c5e257ef322af5e77fc1e52ead3ad6bf3bb4ac06480dd17ee3900d7a24cfb"
+hdf5_ver="1.14.3"
+hdf5_sha256="9425f224ed75d1280bb46d6f26923dd938f9040e7eaebf57e66ec7357c08f917"
```
oschuett: Where did you get this new archive? On [their repo](https://github.com/HDFGroup/hdf5/releases/tag/hdf5-1_14_3) I could only find .zip and .tar.gz files.
Also, is there a reason you did not upgrade to the latest version [1.14.5](https://github.com/HDFGroup/hdf5/releases/tag/hdf5_1.14.5) ?

tools/toolchain/scripts/stage8/install_trexio.sh
```diff
@@ -0,0 +1,107 @@
...
+    else
+      echo "Installing from scratch into ${pkg_install_dir}"
+      [ -d trexio-${trexio_ver} ] && rm -rf trexio-${trexio_ver}
+
+      if [ -f trexio-${trexio_ver}.tar.gz ]; then
+        echo "trexio_${trexio_ver}.tar.gz is found"
+      else
+        #download_pkg_from_cp2k_org "${trexio_sha256}" "trexio-${trexio_ver}.tar.gz"
+        # the release of trexio should be uploaded to the cp2k repository
+        wget -q https://github.com/TREX-CoE/trexio/releases/download/v${trexio_ver}/trexio-${trexio_ver}.tar.gz
```
oschuett: I've uploaded the tarball to: https://www.cp2k.org/static/downloads/trexio-2.5.0.tar.gz

src/trexio_module.F
```diff
@@ -0,0 +1,17 @@
+!--------------------------------------------------------------------------------------------------!
```
oschuett: I don't see any USE statements for this module file. Is it really needed?

oschuett (commented):
Just a few comments, generally LGTM.

tools/toolchain/scripts/stage7/install_hdf5.sh
stefabat: As a matter of fact, I realized yesterday night that I changed that. On the master branch we are still at 1.14.2, but on the cp2k downloads repo I found that [1.14.3 is uploaded](https://www.cp2k.org/static/downloads/hdf5-1.14.3.tar.bz2). Yesterday I tried again to revert to 1.14.2 and I remembered why I changed: with my machine (arch linux, gcc 14.2.1, and the rest from the toolchain) I cannot manage to compile 1.14.2.
On the other hand, I can compile 1.14.3 and everything also works with my system hdf5 library, which is 1.14.5.
As far as I am concerned, we can move on to 1.14.5.

Notice also that I changed the hdf5 script to search for the system library, as in arch linux (maybe in other distros as well), using simply https://github.com/cp2k/cp2k/blob/2de90c3493f93ab5519e4c813f6f1ead4f3a97fc/tools/toolchain/scripts/stage7/install_hdf5.sh#L58-L61
does not work. The path that `h5cc` returns is `/usr/lib`, ending up in the end with `/usr/lib/lib` and `/usr/lib/include` which are wrong. The `lib` and `include` paths on arch linux are `/usr/lib` and `/usr/include`, hence my changes.

oschuett: Thanks for extending the install_hdf5.sh to search for system libraries. That's very useful!

I've uploaded the tarball for 1.14.5. Note that the compression type has changed as they no longer offer bz2:

https://www.cp2k.org/static/downloads/hdf5-1.14.5.tar.gz

stefabat: Ok, thanks. So if I update the hdf5 version used in this PR, then it will be also changed in master once merged and I don't know whether that is gonna break other libraries which depend on hdf5, like SIRIUS. Is this anyhow fine?

src/trexio_module.F
stefabat: I talked with the developer of TREXIO and we figured that this is the (best?) way to use their fortran API. The library is written in C and compiled by the toolchain, and the fortran API is simply the module source code available in `fortran_f.f90`, which you have to compile yourself together with your program. `trexio_f.f90` is located in the `tools/toolchain/install/trexio` directory.
I don't know whether there is a better way to do this..?

stefabat:
> > Tests with dbg executables fail because of a bug in HDF5 1.14.3...
> 
> The workaround for this is to temporarily disable the trapping:
> 
> https://github.com/cp2k/cp2k/blob/2de90c3493f93ab5519e4c813f6f1ead4f3a97fc/src/fm/cp_fm_diag.F#L621-L633

Actually, with 1.14.5 this is not a problem anymore, so I don't need to disable the trapping.

src/trexio_module.F
oschuett: Ok, then I'd suggest you rename the file from `trexio_module.F` to `trexio.F`. This way it becomes clear that the USE-statement in `trexio_utils.F` refers to this file.

tools/toolchain/scripts/stage7/install_hdf5.sh
oschuett: Since it's only a patch level upgrade, I would not expect any issues with other libraries. 
However, if you'd like to play it safe then you could open a separate PR for just the HDF5 upgrade and afterwards return to this PR.

stefabat: I didn't think about that. It's true that for a patch level upgrade it should not be problematic.

PR #3795, opened 2024-12-04
Adjusted handling block size for FM matrices

hfp:
- Input keyword defaults now rely on internal setting (NROW/COL_BLOCKS).
- Use m_cpuid_vlen to make block size multiple of vector-length.
- Updated internal block size (32->64).
- Adjusted regtest-rtbse gold values.
- Improved code format (minor).

### DIFF
--- a/src/fm/cp_cfm_diag.F
+++ b/src/fm/cp_cfm_diag.F
@@ -58,7 +58,7 @@ CONTAINS
 
       COMPLEX(KIND=dp), DIMENSION(:), POINTER  :: work
       COMPLEX(KIND=dp), DIMENSION(:, :), &
-         POINTER                                :: m
+         POINTER                               :: m
       INTEGER                                  :: handle, info, liwork, &
                                                   lrwork, lwork, n
       INTEGER, DIMENSION(:), POINTER           :: iwork
@@ -66,7 +66,7 @@ CONTAINS
 #if defined(__parallel)
       INTEGER, DIMENSION(9)                    :: descm, descv
       COMPLEX(KIND=dp), DIMENSION(:, :), &
-         POINTER                                :: v
+         POINTER                               :: v
 #if defined (__HAS_IEEE_EXCEPTIONS)
       LOGICAL, DIMENSION(5)                    :: halt
 #endif
--- a/src/fm/cp_fm_diag.F
+++ b/src/fm/cp_fm_diag.F
@@ -678,8 +678,8 @@ CONTAINS
 
       ! Diagonalise the symmetric n by n matrix using the LAPACK library.
 
-      TYPE(cp_fm_type), INTENT(IN)              :: matrix
-      TYPE(cp_fm_type), OPTIONAL, INTENT(IN)    :: eigenvectors
+      TYPE(cp_fm_type), INTENT(IN)                 :: matrix
+      TYPE(cp_fm_type), OPTIONAL, INTENT(IN)       :: eigenvectors
       REAL(KIND=dp), OPTIONAL, INTENT(IN)          :: work_syevx
       INTEGER, INTENT(IN), OPTIONAL                :: neig
       REAL(KIND=dp), DIMENSION(:), INTENT(OUT)     :: eigenvalues
@@ -925,7 +925,7 @@ CONTAINS
       CHARACTER(LEN=2*default_string_length)   :: message
       INTEGER                                  :: handle, n, m, myinfo, lwork
       REAL(KIND=dp), DIMENSION(:, :), POINTER  :: a
-      TYPE(cp_fm_type)                :: matrix_lu
+      TYPE(cp_fm_type)                         :: matrix_lu
       REAL(KIND=dp), ALLOCATABLE, DIMENSION(:) :: work
 
 #if defined(__parallel)
@@ -1043,7 +1043,7 @@ CONTAINS
       ! - Creation (29.03.1999, Matthias Krack)
       ! - Parallelised using BLACS and ScaLAPACK (06.06.2001,MK)
 
-      TYPE(cp_fm_type), INTENT(IN)            :: matrix, work
+      TYPE(cp_fm_type), INTENT(IN)               :: matrix, work
       REAL(KIND=dp), INTENT(IN)                  :: exponent, threshold
       INTEGER, INTENT(OUT)                       :: n_dependent
       LOGICAL, INTENT(IN), OPTIONAL              :: verbose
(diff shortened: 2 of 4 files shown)


hfp:
This change has a measurable positive impact on performance. It not only documents the motivation of some LOC or gives more precise comments, e.g., (1) failure for non-square block-size in case of square matrices, or (2) forcing rectangular blocks must be respected.  Regarding (1), square matrices likely imply later-on-ops require a square matrix like Cholesky decomposition which in turn does not take a seconds/different dimension for the respective ScaLAPACK call and hence does not accept rectangular blocks as per ScaLAPACK design. Regarding (2), for instance Arnoldi related code fails with arrays out of bounds if a potentially square matrix automatically implies square blocks like how to interpret a square_blocks-argument set to false.

This PR also removes naive code bumping up the block size to the full local extent in case of lower rank-counts. ScaLAPACK's numroc function does not use the block-size in the first place to determine what's local or not like a block-size does not necessarily produces an imbalance except for some negligible remainder.

In any case, the new code added makes block-sizes a multiple of the vector-length of the CPU and thereby avoids not only odd splits when distributing work but also improves local performance.

hfp:
The code could be merged based on SDBG or non-parallel tests in general. However, it seems some functionality or in the order of five tests (regtest-gw-kpoints) needs updated gold values for regtests. The latter can also be a hidden instability not yet discovered.

oschuett:
Thanks for looking into this! I guess the code in `cp_fm_struct_create` predates vector instructions.

> five tests (regtest-gw-kpoints) needs updated gold values for regtests.

Yeah, I also had to adopt one test when I recently doubled the block size: #3184.

hfp:
> Yeah, I also had to adopt one test when I recently doubled the block size: #3184.

Thanks for the comment. Interesting. I will adjust say the gw-kpoints tests when ready. However, `tests/QS/regtest-gw-kpoints/G0W0_IH_SOC.inp` currently [fails](https://storage.googleapis.com/cp2k-ci/run-cp2k-psmp-156f3808_report.txt) with Emin becoming negative and subsequently a higher-root is taken (fractional exponent), introducing NaN. Cause is/are negative Eigenvalues in a MO-set.

oschuett:
> However, tests/QS/regtest-gw-kpoints/G0W0_IH_SOC.inp currently fails with Emin becoming negative ....

Yes, I actually had to give up on an earlier attempt to increase the block size because I ran into similar issues: Several parts of the code turned out to be surprisingly sensitive :-/

hfp:
> > However, tests/QS/regtest-gw-kpoints/G0W0_IH_SOC.inp currently fails with Emin becoming negative ....
> 
> Yes, I actually had to give up on an earlier attempt to increase the block size because I ran into similar issues: Several parts of the code turned out to be surprisingly sensitive :-/

Indeed, the "force flags" (when creating the handle/struct aka some optional arguments) are sort of "must be respected" (which was not documented or part of a comment); that's even the case when resulting blocks exceed the local size. Apparently, the call side sometimes "bumps" sizes in anticipation of "something" (redistribution?) and if not respected, code then ends up with OOB access.

I am essentially back to "do nothing" if something is "forced" even if it does make sense like block size end up larger than local size (that might be the key, aka re/new distribution). Luckily, it still means block sizes end up to be multiples of the vector-length in almost all cases even when block sizes are constrained by an "extreme" number of ranks (only the remainder is now affected with oddity).

hfp:
@StepanMarek Gold values of **regtest-rtbse** have been adjusted in this somewhat unrelated PR:

> h2_rt_cont.inp: old=0.86507260E-006  new=0.10294155E-005
h2_delta_cont.inp: old=0.88156804E-002  new=0.61079317E-002
h2_ft.inp: old=0.83649881E+001  new=0.97558548E+001

Above are the worst changes (suggesting an issue with **rtbse**).

Please also note, [QS/regtest-rtbse/TEST_FILES](https://github.com/cp2k/cp2k/blob/master/tests/QS/regtest-rtbse/TEST_FILES) duplicates `h2_rt.inp` with two/different 2nd column values like grep`ping it for different results?

hfp:
Hmm, PSMP-test suggests https://github.com/cp2k/cp2k/pull/3795#issuecomment-2527800050, however SSMP/SDBG reproduces the original Gold values. This blocks this PR and additional debug is needed.

StepanMarek:
Hello @hfp , I will have a look whether I can replicate the instability in the test files in my build of the CP2K and potentially change the parameters.

Edit : with regards to h2_rt.inp mentioned twice in QS/regtest-rtbse/TEST_FILES - first line checks for real part of
the result while second check for the imaginary part - since real and imaginary parts are typically printed as separate files, aggregating them on the standard output leads to two lines

hfp:
Indeed, `f930d121d40ab8a6039f41fed9bb49647d27ffcd` passed with psmp (and adjusted Gold values of rtbse regtest), but failed for sdbg. Now, the next change for this PR is about: (1) avoiding multiples of vlen, and (2) reverting back to a blocksize of 32.

hfp:
> Indeed, `f930d121d40ab8a6039f41fed9bb49647d27ffcd` passed with psmp (and adjusted Gold values of rtbse regtest), but failed for sdbg. Now, the next change for this PR is about: (1) avoiding multiples of vlen, and (2) reverting back to a blocksize of 32.

Interesting enough, disabling the code making the blocksize a multiple of vlen will reproduce the originally given Gold values for rtbse-regtest.

StepanMarek:
Dear @hfp, I did not manage to reproduce the differing results on sdbg build and psmp build based on my CP2K builds (ran on personal computer + Supermuc-NG) - I can provide the arch files if necessary. But it also seems that in the later commits, the tests in regtest-rtbse are now producing the correct results.

Let me know if I can help with the tests in another way.

hfp:
To wrap this up, the `cp_fm_struct_create` is quite sensitive to change and many call-sites have developed to fully use the subroutine's signature and options to achieve certain things (plus the meaning of some of the options is/was not clear). I will merge this PR now as is, i.e., generally there is little use of the new vlen-option. In fact, to make thing pass, I cannot generally revise block size to be a multiple if vlen, and there seem to be many cases where this would be beneficial. In essence, the perf. advantage claimed above is gone (so no better than what we had). On the other hand, a 2nd PR can still make it (hope is dying last).

RemiPasquier:
Hello,

I am commenting on this because the small cell GW code seems not to play well with the changes made in this commit.

Before the commit, the equivalent quantity (dumblock) was defined as the result of the integer division of the global row/col size by the row/col num_pe. After the commit, If I am not mistaken, nmax_block is essentially defined as the result of the integer division of the global row/col size by the row/col num_pe, but then it is remultiplied by this row/col num_pe. So it is essentially the biggest multiple of num_pe that is smaller than the global row/col sizes, which is way bigger than dumblock and usually not that far from the global size itself.

The issue is that, if I'm not mistaken, this basically creates massive blocks for any array that is smaller than the optimal size of 64 so that the number of blocks is much smaller than the number of MPI processes, and the small cell code then crashes because some small fm arrays used in the code are not correctly populated. The code does run if I greatly reduce the number of processes (which is not obviously not convenient for more intensive calculations), but I can fix this by removing the remultiplication by num_pe, so by redefining:

`nmax_block = (fmstruct%nrow_global + fmstruct%context%num_pe(1) - 1)/ &
                      (fmstruct%context%num_pe(1))` at line 207

`nmax_block = (fmstruct%ncol_global + fmstruct%context%num_pe(2) - 1)/ &
                      (fmstruct%context%num_pe(2))` at line 215

And then the small cell code works perfectly, but that's basically a drastic change in these quantities (it basically reverts the changes in the commit), so I was wondering if there was some misunderstanding on my side on these changes or another way to solve this issue ?

PR #3798, opened 2024-12-06
Fix links in CMake documentation

RMeli:


RMeli:
Looks like `make pretty` does not like capital letters for links, but it's not a big deal.

oschuett:
> Looks like make pretty does not like capital letters for links, but it's not a big deal.

We use [mdformat](https://github.com/hukkin/mdformat), which describes itself as being opinionated... so yeah ;-)

Personal, I just inline the URLs: `[CMake](https://cmake.org)`.

PR #3799, opened 2024-12-06
Add DLA-Future generalized eigensolver

RMeli:
Related to #3787.


RMeli:
Can someone please trigger the CMake test? Thank you.

RMeli:
I updated Spack to `v0.23.0` and `dla-future-fortran` to `0.2.0`. `libint` is not enabled in the Docker container with Spack (takes too long to build), so the relevant test for this PR (`QS/regtest-dlaf-2`) does not actually run.

PR #3804, opened 2024-12-11
WIP: Update DBCSR

alazzaro:


oschuett:
Note that we also refer to DBCSR versions in these two other places:
- [tools/docker/scripts/install_dbcsr.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/install_dbcsr.sh)
- [tools/spack/cp2k-dependencies.yaml](https://github.com/cp2k/cp2k/blob/master/tools/spack/cp2k-dependencies.yaml#L7)

RMeli:
@alazzaro the Spack version needs to be available on Spack. I'm about to open a PR. Then we will need to update the git commit here:
https://github.com/cp2k/cp2k/blob/3250c911c6bed5297392302b9d60973e1f485b8b/tools/docker/Dockerfile.test_cmake#L47

RMeli:
My comment was a few seconds too late. =) I'll take care of this on a separate PR, once the change in Spack is merged upstream.

alazzaro:
thanks @RMeli !
@oschuett could we document this procedure somewhere? Wiki?

PR #3811, opened 2024-12-14
Fix SIRIUS build flags

mkrack:


oschuett:
It seems this PR broken the [Intel oneAPI](https://dashboard.cp2k.org/archive/intel-psmp/index.html) test.

Issue #2573, opened 2023-02-10
Workers in psmp regtests do not run concurently


abussy:
Until this morning, I was using the old regtesting bash script `do_regtest` due to my working branch being behind. I just changed to the new python script `do_regtest.py` and encounter issues when running with the `psmp` version.

Namely, if I run `./do_regtest.py --maxtasks 100 --workbasedir ../../.. local psmp`, it seems that only one test is running at any given time, whereas 25 are expected to do so. It can tell because of the very slow timing and an obviously empty machine when running `htop`.

On the other hand, when I run the `ssmp` version, I observe that 50 tests are run at the same time. Any idea @oschuett ?

oschuett:
Can you share some of the outout? Especially, the `---- Settings ----` section and the `Launched X test directories and Y worker...` line?

Btw, a more convenient way run the tests is via `make VERSION=psmp test`.

fstein93:
I have exactly the same issue and I never run regtests by launching the regtest script but with `make <options> test` instead

abussy:
```
./do_regtest.py --maxtasks 100 --workbasedir ../../.. local psmp
*************************** Testing started ****************************

----------------------------- Settings ---------------------------------
MPI ranks:      2
OpenMP threads: 2
GPU devices:    0
Workers:        25
Timeout [s]:    400
Work base dir:  ../../../TEST-local-psmp-2023-02-10_12-27-48
MPI exec:       ['mpiexec']
Smoke test:     False
Valgrind:       False
Keepalive:      False
Flag slow:      False
Debug:          False
ARCH:           local
VERSION:        psmp
Flags:          omp,libint,fftw3,libxc,parallel,scalapack,cosma,xsmm,spglib,libvori,libbqb,libvdwxc
------------------------------------------------------------------------
Copying test files ... done
Skipping UNIT/nequip_unittest because its requirements are not satisfied.
Skipping TMC/regtest_ana_on_the_fly because its requirements are not satisfied.
Skipping SIRIUS/regtest-1 because its requirements are not satisfied.
Skipping QS/regtest-elpa-2 because its requirements are not satisfied.
Skipping QS/regtest-pexsi because its requirements are not satisfied.
Skipping QS/regtest-elpa-qr because its requirements are not satisfied.
Skipping QS/regtest-elpa because its requirements are not satisfied.
Skipping Fist/regtest-plumed2 because its requirements are not satisfied.
Skipping Fist/regtest-quip because its requirements are not satisfied.
[...]

PR #3813, opened 2024-12-16
DLA-Future complex eigensolvers

RMeli:
Introduce DLA-Future complex standard and generalized eigensolvers.

Other changes:
* Moved complex Cholesky decompositions to new module
* Moved `cp_cfm_upper_to_full` from `rpa_gw_kpoints_util` to `cp_cfm_basic_linalg`

As for #3799, tests in `regtest-dlaf-2` require ~`libxc`~ `libint` and therefore will not run in the current CMake CI action.


RMeli:
I think the CI timeout is an issue of thread binding. Locally, if I run with `--bind-to core:2` I get
```
>>> /tmp/rmeli/spack-stage/spack-stage-cp2k-master-lei4ue472mgchmksqn2uypijyqkh5sz5/spack-build-lei4ue4/bin/TEST-2024-12-16_11-40-33/QS/regtest-dlaf
    H2O-6.inp                                                                             -17.14603642           OK (   2.57 sec)
    c_2.inp                                                                               -45.68042106           OK (   5.89 sec)
```
While if I run with the docker container, which uses `--bind-to none` I do not hit the timeout but the test is indeed very slow:
```
#23 371.7 >>> /opt/cp2k/build/bin/TEST-2024-12-16_10-23-56/QS/regtest-dlaf
#23 371.7     H2O-6.inp                                                                             -17.14603642           OK (   2.48 sec)
#23 371.7     c_2.inp                                                                               -45.68042106           OK ( 187.95 sec)
```

oschuett:
> While if I run with the docker container, which uses --bind-to none I do not hit the timeout but the test is indeed very slow:

So far we had no issues with bind-to-none - on the [contrary](https://github.com/cp2k/cp2k/issues/2573#issuecomment-1425848613). Could you please verify that it's really the culprit by running the docker container without it.

oschuett:
> tests in regtest-dlaf-2 require ~~libxc~~ libint and therefore will not run in the current CMake CI action.

This is really not good. I believe we have the following options:
1. Add DLAF to the toolchain.
2. Get Spack build cache to work in the CI.
3. Switch to libcint.
4. Stomach the cost incl. the prolonged wait time for developers.

RMeli:
@oschuett I agree. A multi-stage build would prevent re-building the dependencies every time, which is what we used for DLA-Future.

We have recently discussed internally at CSCS about possible avenues for (part of) the CP2K CI (especially GPU tests), and we have some propositions. I'll write you and @mkrack and e-mail in the coming days. (Let me know if I should include anyone else in the discussion.)

For the timeout, I think this might be the issue now that we have multiple tests running concurrently:
https://github.com/cp2k/cp2k/blob/d59d53cc667a34cf18d0bb5fe1046647e96c1620/tools/docker/scripts/test_regtest_cmake.sh#L36-L37

I'll try to do more tests locally, and try to figure out exactly what is happening.

oschuett:
> CSCS about possible avenues for (part of) the CP2K CI (especially GPU tests), 

Sounds great! I'm looking forward to the email :-)

> For the timeout, I think this might be the issue now that we have multiple tests running concurrently:

Oh, I had totally forgotten about the `PIKA_PROCESS_MASK`. Ideally, pika could figure this out by itself. Otherwise, we could add some logic to the `do_regtest.py` script. We already do [something similar](https://github.com/cp2k/cp2k/blob/d59d53cc667a34cf18d0bb5fe1046647e96c1620/tests/do_regtest.py#L301) for GPUs.

RMeli:
I disabled binding for pika, and set the number of pika threads to the same number of OpenMP threads. Tests now run in a reasonable time:
```
>>> /opt/cp2k/build/bin/TEST-2024-12-16_17-20-05/QS/regtest-dlaf
    H2O-6.inp                                                                             -17.14603642           OK (   0.57 sec)
    c_2.inp                                                                               -45.68042106           OK (   5.29 sec)
```

To get this PR going, may I suggest to enable `libint` temporarily and do a single run just before merging? Then we can work on improving the CI for Spack + CMake in the new year.

oschuett:
> I disabled binding for pika, and set the number of pika threads to the same number of OpenMP threads.

Nice! That looks like a much better solution.

> To get this PR going, may I suggest to enable libint temporarily and do a single run just before merging? 

I think, we should just enable libint and leave it on. Since this will prolong the test by ~30 minutes, I'd move it to the weekly cadence.

RMeli:
> I think, we should just enable libint and leave it on. Since this will prolong the test by ~30 minutes, I'd move it to the weekly cadence.

Ok, I'll add it here then. While at it, I'll see if I can actually force Spack by temporarily reviving #3331.

tools/docker/generate_dockerfiles.py
```diff
@@ -851,6 +851,13 @@ def spack_env_toolchain() -> str:
 RUN spack compiler find
 RUN spack external find --all --not-buildable
 
+# Enable Spack build cache
+ARG SPACK_BUILD_CACHE=develop-2024-12-15
```
oschuett, line 855: Oh, I didn't realize that there is now a build cache for develop tags. That's awesome!
Should we maybe link it to the Spack version that we install above:

```
ARG SPACK_VERSION=develop-2024-12-15
...
git fetch --quiet --depth 1 origin ${{SPACK_VERSION}} --no-tags
...
spack mirror add ${{SPACK_VERSION}} https://binaries.spack.io/${{SPACK_VERSION}}
```

RMeli:
@oschuett as suggested above I enabled `libint` in the CMake CI action and now all DLA-Future tests run in CI:
```
>>> /opt/cp2k/build/bin/TEST-2024-12-17_12-35-15/QS/regtest-dlaf-2
    Ne-pbc-shortrange.inp                                                                   890.608044           OK (   1.89 sec)
    Ar-HF-2p-SOC-os.inp                                                                     261.347679           OK (   2.26 sec)
<<< /opt/cp2k/build/bin/TEST-2024-12-17_12-35-15/QS/regtest-dlaf-2 (8 of 352) done in 4.15 sec
>>> /opt/cp2k/build/bin/TEST-2024-12-17_12-35-15/QS/regtest-dlaf
    H2O-6.inp                                                                             -17.14603642           OK (   0.58 sec)
    c_2.inp                                                                               -45.68042106           OK (   6.10 sec)
<<< /opt/cp2k/build/bin/TEST-2024-12-17_12-35-15/QS/regtest-dlaf (9 of 352) done in 6.68 sec
```

I re-enabled Spack build cache, but I was unable to use `libint` from there because of some constraints in the dependency tree. We could look into enabling our own build cache, and later on moving the Spack + CMake CI action to a multi-stage build (possibly on top of CSCS CI).

tools/docker/generate_dockerfiles.py
RMeli: Spack does a weekly build. I think they should be separate, otherwise we will have to wait up to a week to be able to update the Spack version.

Unfortunately, as I just mentioned below, it is not possible to use `libint` from the build cache. I tried multiple ways to force Spack to use it, but it can't satisfy all the constraints with the version provided by the build cache.

This still saves a bit of time, since several dependencies are not built but taken directly from the build cache.

oschuett: > Unfortunately, as I just mentioned below, it is not possible to use libint from the build cache. I tried multiple ways to force Spack to use it, but it can't satisfy all the constraints with the version provided by the build cache.

Presumably, we're relying too much on Ubuntu packages. With the build cache in place I'd be ready to lean in and install most packages from Spack.

> Spack does a weekly build. I think they should be separate, otherwise we will have to wait up to a week to be able to update the Spack version.

Ok, that makes sense. My main worry is that there will be some change that leads to a lot of cache misses. Maybe as a safe-guard we could time box it:

```
timeout 10m spack -e myenv concretize -f
```

In any case, all of this could also happen in a separate PR.

RMeli: > Presumably, we're relying too much on Ubuntu packages.

That is a good point. I can try to experiment locally, and see if I can come up with a combination where `libint` is indeed pulled from the cache. Another option to look into would be to setup our own [OCI registry for the build cache](https://spack.readthedocs.io/en/latest/binary_caches.html#oci-docker-v2-registries-as-build-cache).

Happy to have a look at this in a separate PR of course.

oschuett:
> Happy to have a look at this in a separate PR of course.

Ok, then I'll merge this PR now.

Furthermore, I'll move the  "cmake" test to weekly cadence and I'll also rename it to "spack" because that has become its main role. For testing the CMake build system we can now use the "ssmp" test.

tools/docker/generate_dockerfiles.py
RMeli: FYI, I tried to build without tapping into Ubuntu packages but the problem persists. Spack is not very explicit in why it fails, but there must be some incompatibility between the cached `libint` dependencies and other packages. One workaround could be to use `unify: when_possible` instead of `unify: true`, but I feat this will create more problems than it solves (there will be multiple version of some libraries etc.).

PR #3814, opened 2024-12-16
GCC/14 to overcome codegen issue with coverage

hfp:


oschuett:
What exactly is the issue?
I'd prefer to stick with Ubuntu's default compiler for most of our tests, because that's most likely what our users are using.

hfp:
> What exactly is the issue? I'd prefer to stick with Ubuntu's default compiler for most of our tests, because that's most likely what our users are using.

It seems [tests/QS/regtest-gpw-8/si8_broy_wc_crazy.inp](https://storage.googleapis.com/cp2k-ci/dashboard_coverage-pdbg_report.txt) does not reproduce the reference value if `-O1 -coverage` is present. This PR is just to try it, but I misspelled the package like `g++-14` I guess. I can also try a different FM block size since the issue seems to be present since my changes related to blocksize.

oschuett:
>It seems `tests/QS/regtest-gpw-8/si8_broy_wc_crazy.inp` does not reproduce the reference...

Yes, that test has been unstable for a long time as it pops up every [once in a while](https://dashboard.cp2k.org/archive/current-ssmp/commit_724036aab431ed6c2c3183a7f1712b746c06dde0.txt). So, I don't think it's a compiler problem. Maybe somebody could try to stabilize the test. Otherwise we'll just have to lower the threshold.

PR #3829, opened 2024-12-27
Smeared occupation TDA

annahehn:


oschuett:
Would it be possible to shorten the ` CH3_tddfpt_smeared_occ.inp` test a bit more? It still took 33.14 sec on the first run and only stayed below slowness threshold in the re-run at the end.

annahehn:
Dear Ole, thanks and please excuse, Anna

oschuett:
Hi Anna, no worries. Writing good tests that are also fast is an art :-)

PR #3831, opened 2024-12-29
Libxc 6.2.2 -> 7.0.0

mkrack:


mkrack:
A new SIRIUS release is needed including this [PR](https://github.com/electronic-structure/SIRIUS/commit/1e73cf60863177141a18601d9fcd44f725c88432) to get CP2K work with SIRIUS and Libxc >=7.

oschuett:
Oof, classical [diamond dependency](https://en.wikipedia.org/wiki/Dependency_hell). Thanks for looking into this!
I guess, we better hold the release until we figure out this upgrade?

mkrack:
The Apple M1 macOS CP2K regression tester using the latest Homebrew version of `Libxc` updated automatically to version `7.0.0` without any problem. Therefore, I do not expect any issue with CP2K and `Libxc 7.0.0` once a new SIRIUS release including the patch for `Libxc 7.0.0` is available. I suggest to wait for the next SIRIUS release to keep all tester in sync. @toxa81, what's the schedule for the next SIRIUS release?

PR #3844, opened 2025-01-07
Integrate libsmeagol into regression testing

mkrack:


mkrack:
The OpenMPI test shows random failures though that test should not be affected by this commit.

PR #3851, opened 2025-01-09
dbm: improved dbm_shard_copy

hfp:
- Keep allocation if sufficient.
- Only allocate used size.

### DIFF
--- a/src/dbm/dbm_shard.c
+++ b/src/dbm/dbm_shard.c
@@ -81,28 +81,36 @@ void dbm_shard_init(dbm_shard_t *shard) {
  * \author Ole Schuett
  ******************************************************************************/
 void dbm_shard_copy(dbm_shard_t *shard_a, const dbm_shard_t *shard_b) {
-  free(shard_a->blocks);
+  assert(shard_a != NULL && shard_b != NULL);
+
+  if (shard_a->nblocks_allocated < shard_b->nblocks) {
+    free(shard_a->blocks);
+    shard_a->blocks = malloc(shard_b->nblocks * sizeof(dbm_block_t));
+    shard_a->nblocks_allocated = shard_b->nblocks;
+  }
   shard_a->nblocks = shard_b->nblocks;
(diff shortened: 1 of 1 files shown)


hfp:
For example, `benchmarks/QS_low_scaling_GW/GW.inp` with `dbm_copy` in the timings, the time average self-time went down from 11.6s to 9.2s.

src/dbm/dbm_shard.c
```diff
@@ -81,28 +81,36 @@ void dbm_shard_init(dbm_shard_t *shard) {
...
+
+  if (shard_a->nblocks_allocated < shard_b->nblocks) {
+    free(shard_a->blocks);
+    shard_a->blocks = malloc(shard_b->nblocks * sizeof(dbm_block_t));
+    shard_a->nblocks_allocated = shard_b->nblocks;
+  }
   shard_a->nblocks = shard_b->nblocks;
-  shard_a->nblocks_allocated = shard_b->nblocks_allocated;
-  shard_a->blocks = malloc(shard_b->nblocks_allocated * sizeof(dbm_block_t));
   assert(shard_a->blocks != NULL);
```
oschuett, line 92: Could you please move these asserts to be right after the malloc. This will make it easier for users to understand that they ran out of memory when they trigger.

oschuett (commented):
This is really an obvious optimization. I guess, I underestimated the importance of copy. Thanks for picking up the slack!

src/dbm/dbm_shard.c
hfp: Sure, will do another PR. Since the malloc is conditional, it won't check the `blocks`, `hashtable` and `data` in any case; just saying.

oschuett: I mostly added these asserts to prevent spurious segfaults when we run out of memory. I think, it's a good pattern and I was planing to add a convention check for it. The proper way to implement that would be as a clang-tidy extension, but I simply haven't gotten around to it yet.

PR #3859, opened 2025-01-09
more explicit upper triangular form

hfp:
- Use invmat_symm if complete matrix desired.
- Renamed upper_to_full to uplo_to_full.
- Call invmat_symm instead of POTRx.
- Removed own_cfm_upper_to_full.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -113,7 +113,7 @@ Accelerate and reference BLAS/LAPACK.
 
 When building on/for Windows using the Minimalist GNU for Windows (MinGW) environment, you must set
 `-D__MINGW`, `-D__NO_STATM_ACCESS` and `-D__NO_SOCKETS` to avoid undefined references during
-linking, respectively errors while printing the statistics.
+linking, and to respectively avoid errors while printing the statistics.
 
 ### 2e. MPI and ScaLAPACK (optional, required for MPI parallel builds)
 
--- a/src/admm_methods.F
+++ b/src/admm_methods.F
@@ -22,11 +22,10 @@ MODULE admm_methods
    USE cp_cfm_basic_linalg,             ONLY: cp_cfm_scale,&
                                               cp_cfm_scale_and_add,&
                                               cp_cfm_scale_and_add_fm,&
-                                              cp_cfm_transpose
+                                              cp_cfm_uplo_to_full
    USE cp_cfm_cholesky,                 ONLY: cp_cfm_cholesky_decompose,&
                                               cp_cfm_cholesky_invert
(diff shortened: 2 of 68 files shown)


hfp:
This PR is motivated by the variety of hard-coded triangular forms used, but mostly by the fact that some device-accelerated libraries may only (efficiently) support the lower triangular form, e.g., DLAF requests the lower triangular form. Further, there seem to be no packed flavors of (Sca-)LAPACK functions used in CP2K, i.e., it is about full matrices represented regardless of lower/upper triangular part used. Therefore, the arbitrary choice is streamlined by the PR.

fstein93:
Just for confirmation: Did you check whether the tests are still correct if you change the default triangular form?

hfp:
> Just for confirmation: Did you check whether the tests are still correct if you change the default triangular form?

Thank you for asking! This is a draft PR and not there yet, i.e., there are many occurrences of hard-coded `uplo`. Indeed, the plan is to at least check toggling from `U` to `L` before changing back and merging the PR. Please also note, if anything falls short along this plan - I may just close the PR (without merging it).

hfp:
The FM/UPLO is stored in common `mathlib` module because FM actually falls apart in FM and CFM in the current design. Further, storing FM/UPLO in, e.g., the `cp_fm_types` module, leads to cyclic dependency wrt common code (`src/common`) which may also rely on UPLO.

hfp:
If, e.g., certain GPU enabled library wants a particular storage format, initialization can call `mathlib`'s `setup` routine. Alternatively, this PR introduces a preprocessor definition to set global default triangular format (uplo). This way, an ARCH-file can setup the global UPLO as well. Otherwise, any input can now request UPLO by the means of the global FM section.

hfp:
This PR also contains some minor modernization mostly for `?getpotr?` like calling `invmat_symm` in cases where the former was followed by completing the lower/upper symmetric part. Also note, `invmat_symm` is somewhat "evil" as it does not offer UPLO but if alike-UPLO is supplied (optional), it omits the initial step of the Cholesky decomposition aka UPLO given for what has been done before.

hfp:
This is almost hopeless since using the upper triangular form is implicit at global scale. There are many places that have been fixed/tried in this PR among some code consolidation and cleanup. I may repurpose this PR to keep the useful bits.

hfp:
> This is almost hopeless since using the upper triangular form is implicit at global scale. There are many places that have been fixed/tried in this PR among some code consolidation and cleanup. I may repurpose this PR to keep the useful bits.

I can perhaps introduce it for specific domains on a centralized basis. However, in general the upper format is implicitly present in QS (like everywhere). Even DBCSR<->FM looks affected.

Issue #3416, opened 2024-05-15
CMake can't find several packages from the toolchain


oschuett:
The following packages are currently not found by CMake when provided by the [toolchain](https://github.com/cp2k/cp2k/tree/master/tools/toolchain):

- [ ] ELPA
- [ ] COSMA
- [ ] SIRIUS
- [ ] Torch
- [ ] libgrpp
- [ ] libxsmm
- [ ] dftd4
- [ ] trexio
- [ ] hdf5
- [ ] DeepMD

oschuett:
I've updated the list above by comparing the `Flags:` from [sdbg](https://storage.googleapis.com/cp2k-ci/dashboard_sdbg_report.txt) with those from [ssmp](https://storage.googleapis.com/cp2k-ci/dashboard_ssmp_report.txt), which is our Toolchain+CMake guinea pig.

RMeli:
Thanks @oschuett for the updated list. I'm working on DFTD4 (it's broken on the CMake side), and @mtaillefumier has PR #3864 up for TrexIO.

RMeli:
https://github.com/cp2k/cp2k/pull/3874#issuecomment-2613991745

RMeli:
| lib | ssmp (tc) | psmp (tc) | psmp (spack) |
| --- | :-------: | :-------: | :----------: |
| ELPA | | ✅ | ✅ |
| COSMA |  | ✅ | ✅ |
| SIRIUS | | ✅ | ✅ |
| Torch | ✅ | ✅ | ✅ |
| libgrpp | 🛠 | 🛠 | 🛠 |
| libxsmm |  ✅ | ✅ | ✅ |
| dftd4 | ✅ | ✅ | ✅ |
| trexio | ✅ | ✅ | ✅ |
| hdf5 | ✅ | ✅ | ✅ |
| deepmd | ✅ | ✅ | 🛠 |

- SIRIUS in the Spack build is without the `pugixml` dependency.
- Torch need to be enabled in the Spack build (couldn't find a version that concretizes correctly)
- DeePMD Spack package is missing
- #3864 should take care of `libgrpp`, `trexio`, and `hdf5`

mtaillefumier:
I can open a PR for the deepmd spack package. But it is a monster package.

mtaillefumier:
SIRIUS is detected by cmake with the toolchain. it can be turned on with #3864.

RMeli:
[...]

Issue #3414, opened 2024-05-15
CMake can't find spglib on Ubuntu/Debian


oschuett:
When building with `-DCP2K_USE_SPGLIB=ON` on Ubuntu 2024.04, I get the following error message:
```
CMake Error at CMakeLists.txt:686 (find_package):
  Could not find a package configuration file provided by "Spglib" with any
  of the following names:

    SpglibConfig.cmake
    spglib-config.cmake

  Add the installation prefix of "Spglib" to CMAKE_PREFIX_PATH or set
  "Spglib_DIR" to a directory containing one of the above files.  If "Spglib"
  provides a separate development package or SDK, be sure it has been
  installed.
```

It turns out that the [libspglib-f08-dev](https://packages.ubuntu.com/hu/noble/amd64/libspglib-f08-dev/filelist) package ships cmake files with somewhat different names:

```
/usr/lib/x86_64-linux-gnu/cmake/Spglib/SpglibTargets_fortran-none.cmake
/usr/lib/x86_64-linux-gnu/cmake/Spglib/SpglibTargets_fortran.cmake
```

@mtaillefumier what do you think?

mtaillefumier:
We should not do anything at all except be sure that Spglib::fortran is added to the cp2k libraries dependencies. 

https://github.com/cp2k/cp2k/blob/c5411e048c1d97964d822cc470060c619c47960d/src/CMakeLists.txt#L1560C5-L1560C49

we need to add `Spglib::fortran` to this line. I can do it if you want.

mtaillefumier:
Basically, cmake will take care of this all by itself unless the devs decided otherwise. it might be a module to be searched with `find_package(Spglib required fortran)`

mtaillefumier:
The error is really ubuntu specific actually. `/usr/lib/x86_64-linux-gnu/cmake` should be added to `CMAKE_PREFIX_PATH`

oschuett:
> we need to add Spglib::fortran to this line. I can do it if you want.

Yes, it would be great if you could help out with these `find_package` issues. Then I'll focus on migrating the remaining CI scripts.

oschuett:
> we need to add Spglib::fortran to this line. I can do it if you want.

[...]

PR #3868, opened 2025-01-16
Enable spglib in ssmp CMake build

RMeli:
It is already enabled in the Spack build. Related to https://github.com/cp2k/cp2k/issues/3416.

https://github.com/cp2k/cp2k/issues/3414 might still be outstanding, but if this is the case the best course of action would be to report the issue and have a fix upstream.

### DIFF
--- a/tools/docker/scripts/build_cp2k_cmake.sh
+++ b/tools/docker/scripts/build_cp2k_cmake.sh
@@ -191,7 +191,7 @@ elif [[ "${PROFILE}" == "minimal" ]] && [[ "${VERSION}" == "ssmp" ]]; then
     -DCP2K_USE_VORI=OFF \
     -DCP2K_USE_COSMA=OFF \
     -DCP2K_USE_DLAF=OFF \
-    -DCP2K_USE_SPGLIB=OFF \
+    -DCP2K_USE_SPGLIB=ON \
     -DCP2K_USE_LIBTORCH=OFF \
     .. |& tee ./cmake.log
   CMAKE_EXIT_CODE=$?


tools/docker/scripts/build_cp2k_cmake.sh
```diff
@@ -191,7 +191,7 @@ elif [[ "${PROFILE}" == "minimal" ]] && [[ "${VERSION}" == "ssmp" ]]; then
     -DCP2K_USE_VORI=OFF \
     -DCP2K_USE_COSMA=OFF \
     -DCP2K_USE_DLAF=OFF \
-    -DCP2K_USE_SPGLIB=OFF \
+    -DCP2K_USE_SPGLIB=ON \
```
oschuett, line 194: This is the minimal build which is supposed to be - well minimal ;-)
Please enable it instead above for the following case:
```elif [[ "${PROFILE}" == "toolchain" ]] && [[ "${VERSION}" == "ssmp" ]]; then```

RMeli: Sorry, I looked at the wrong branch in this case. Looks like the `ssmp` version with the `toolchain` is already enabled. Sorry about the noise.

oschuett: NP, I also missed this. Yes, spglib already works everywhere except on Ubuntu:

https://github.com/cp2k/cp2k/blob/f8b3abcd3ac413146004a9ebf8d71317c9449627/tools/docker/scripts/build_cp2k_cmake.sh#L138-L139

PR #3873, opened 2025-01-17
Fix CMake build with libtorch and enable it in CI (toolchain)

RMeli:
I tried to also add `py-torch@1.12.1` to the Spack build but it fails to build, so more work is needed on that front.


oschuett:
I'm not sure what the Spack package includes, but CP2K depends on the C++11 API of PyTorch.

RMeli:
A lot of optional dependencies are on by default in the `py-torch` Spack package. I tried the most minimal build I could get to concretize successfully, but it fails to build in the CP2K container. (I did manage to build a similar spec locally, but with version `2.x`.) I'll look into this next week.

RMeli:
> I'm not sure what the Spack package includes, but CP2K depends on the C++11 API of PyTorch.

I think the issue is that `py-torch@1.12` is rather old, and it is difficult to find the right constraints on other packages to it builds. I managed to build (and run the tests) with `py-tprch@2.5.1`, but it requires to bump C++ from C++14 to C++17. Would that be an acceptable change?

RMeli:
They added the C++ standard needed by PyTorch as a property of the target, so it shouldn't be an issue to bump the C++ support only when required by PyTorch (which I think would be acceptable).

oschuett:
Yes, go ahead. In CP2K we use C++ very sparingly, so the version shouldn't really matter to us.

PR #3874, opened 2025-01-18
Docker: Migrate psmp test to CMake

oschuett:
Lets make psmp our guinea pigs for parallel builds. See also #3416.


oschuett:
As expected, the number of tests dropped from 4594 down to 4222. So, there is still some way to go.

RMeli:
For reference, old flags:
```
omp,libint,fftw3,libxc,libgrpp,pexsi,elpa,parallel,scalapack,cosma,quip,deepmd,xsmm,plumed2,spglib,libdftd4,sirius,libvori,libbqb,libtorch,libvdwxc,hdf5,trexio
```
New flags:
```
omp,libint,fftw3,libxc,parallel,scalapack,mpi_f08,xsmm,plumed2,spglib,libvori,libbqb
```

PR #3876, opened 2025-01-20
use blas-like extensions

hfp:
- Out-of-place transpose (mkl_?omatcopy).
- Cleanup.


hfp:
It was considered to use zgemm3m however there is no corresponding ScaLAPACK function (pzgemm3m).

PR #3890, opened 2025-01-24
DBM: MAX_BATCH_SIZE 10000 -> 30000

hfp:
- INITIAL_NBLOCKS_ALLOCATED/INITIAL_DATA_ALLOCATED can be zero.
- MAX_BATCH_SIZE tuned on CPUs assuming GPUs benefit even more.
  - Both mp_sync and mp_waitall_2 decreased.
  - Of course, dbm_multiply decreased.
- Driver/Makefile/HIP: missing object file/compilation.


src/dbm/dbm_shard.c
```diff
@@ -64,15 +64,20 @@ static void hashtable_init(dbm_shard_t *shard) {
 void dbm_shard_init(dbm_shard_t *shard) {
   shard->nblocks = 0;
   shard->nblocks_allocated = INITIAL_NBLOCKS_ALLOCATED;
-  shard->blocks = malloc(shard->nblocks_allocated * sizeof(dbm_block_t));
-  assert(shard->blocks != NULL);
+  shard->blocks = NULL;
+  if (0 < shard->nblocks_allocated) {
```
oschuett: I'm confused. Isn't this branch always taken because `0 < INITIAL_NBLOCKS_ALLOCATED`? Same for below.

hfp: Yes, it's always taken (and the compiler can even see this). The branch is there since I played with the parameters and I wanted to allow INITIAL_NBLOCKS_ALLOCATED=0 (conceptional) as a valid choice. In fact, it works with those params set to zero since the shards are evaluated for growth before use.

oschuett: Ok, got it! Did you find that zero is a good value? Otherwise, I'd prefer to keep the code as simple as it is.

hfp: Zero was not worse but also no advantage. I can get the original code back given that the initial size is non-zero.

oschuett: Thanks!
If there really is not advantage then I'd also be fine with removing those initial mallocs and the two hyper-parameters. I simply don't want to clutter the code with unnecessary if statements.

hfp: Sounds even better, I will actually run some tests to be really sure. Removing the initial mallocs is perhaps best for simplicity.

hfp:
I will send a separate PR for removing the unnecessary tuning parameters.

PR #3893, opened 2025-01-24
Upgrade to DeePMD 3.0.1 and switch to PyTorch backend

oschuett:


oschuett:
I believe the only thing missing is a PyTorch version of the [model](https://github.com/cp2k/cp2k/blob/master/data/DeePMD/W.pb) for the [regression test](https://github.com/cp2k/cp2k/blob/master/tests/Fist/regtest-deepmd/DeePMD_W.inp).

I tried `dp convert-backend ./W.pb ./W.pth` but it failed with the following error:

```
KeyError: "The name 'train_attr/training_script:0' refers to a Tensor which does not exist. The operation, 'train_attr/training_script', does not exist in the graph."
```

@njzjz, could you help?

njzjz:
This model seems to be generated by the old version and doesn't store the training parameters. This is necessary to convert between backends, though.

oschuett:
Ok, I'm now using [examples/water/dpa2/frozen_model.pth](https://github.com/deepmodeling/deepmd-kit/blob/r3.0/examples/water/dpa2/frozen_model.pth).

Unfortunately, it still doesn't work with leak sanitizer - presumably because PyTorch does some weird dynamic linking.

PR #3895, opened 2025-01-25
Simple Toolchain/CMake build script (no container)

mkrack:


oschuett:
Your new script has a lot of overlap with [build_cp2k_cmake.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/build_cp2k_cmake.sh) and [install_dbcsr.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/install_dbcsr.sh).
Maybe it could call these exiting scripts to avoid duplication?

PR #3897, opened 2025-01-25
Enable more dependencies in CMake psmp CI build (toolchain)

RMeli:


RMeli:
SIRIUS need a bit more work, I'll tackle it next.

PR #3901, opened 2025-01-27
DBM: removed unnecessary hyper-parameters

hfp:
- Removed INITIAL_NBLOCKS_ALLOCATED and INITIAL_DATA_ALLOCATED.
- Both allocations are handled by buffer-growth routine.


hfp:
ACK, Dashboard status...

mtaillefumier:
Can you revert this change. It breaks the CI/CD and introduce a bug along the way.

hfp:
I noticed the issue with [Dashboard](https://github.com/cp2k/cp2k/pull/3901#issuecomment-2615347195)), and I can revert it meanwhile (and reintroduce it later according to intent). The breakage is actually interesting as it (1) was not caught by the default test, and (2) it points to some deeper issue which needs to be resolved.

PR #3904, opened 2025-01-27
DBM: removed unnecessary hyper-parameters

hfp:
DBM: removed unnecessary hyper-parameters

- This reverts commit ff01eb320a35700a236790801c64722908352714.
- Reapplied and fixed #3901.

Other:
- Only allocate "data_size" on device rather than "data_allocated" on CPU (dbm_multiply_gpu_start).


hfp:
Regtest pdbg actually [passed](https://storage.googleapis.com/cp2k-ci/run-cp2k-pdbg-d3dd179d_report.txt); just shown as "failed" due to one slow test.

mtaillefumier:
You should not have merged this. ssmp tests will fail.

JWilhelm:
@hfp @oschuett @fstein93 @abussy @RemiPasquier

Hi Hans, hi all, 

Rémi has also tested your commit and he found a memory reduction by a factor 2 for the small cell, full k-points GW code (computing time is identical as before)!! Great 🎉🎉 We are planning to run large-scale GW calculations soon with the Gamma-only code. We will keep you updated.

hfp:
@JWilhelm That's great news. Thank you for the feedback!

I hope this was not only #3904 but also #3913 (or at least the improvement was not eroded by the latter). I think Frederick's change in-between increased memory consumption (but bettered things on a different front).

Btw, #3917 added statistics separate for both host and device memory allocation (pooled memory in DBM).

PR #3907, opened 2025-01-27
[docs] Add note about compilers and Spack

RMeli:


docs/getting-started/build-with-spack.md
```diff
@@ -6,6 +6,11 @@ HPC.
 
 To install CP2K with [Spack], you need to [install Spack].
 
+.. note::
```
oschuett: Since we're using MyST, the syntax is a bit different:

https://github.com/cp2k/cp2k/blob/master/docs/README.md#notes-and-warnings

RMeli:
Not sure why the `precommit` action wants to get rid of the link...

oschuett:
> Not sure why the precommit action wants to get rid of the link...

Probably because it's enclosed by `` ``` ``. The two options I see are to either inline the link or to add [colon-fence](https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#syntax-colon-fence) to the [myst_enable_extensions](https://github.com/cp2k/cp2k/blob/a1b1b75c35b7d5f14894dbfc2ab7ad369452db12/docs/conf.py#L26).

PR #3910, opened 2025-01-28
BSE/RI-RPA: improve abort messages and fix bug for ERI_METHOD OS

MGraml:
- BSE: add meaningful abort messages for users
- RI-RPA: [Google group issue](https://groups.google.com/g/cp2k/c/9e8Y55_XW6E) reported a problem, which is related to the [Obara-Saika method](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/XC/WF_CORRELATION/INTEGRALS.html) (invoked for 0D systems by default) and the automatically generated RI_AUX basis sets. Here, I just added a graceful abort, which states their incompatability as already written in the [manual](https://manual.cp2k.org/trunk/methods/post_hartree_fock/preliminaries.html#step-5-setup-of-integral-calculation-step).

### DIFF
--- a/src/bse_full_diag.F
+++ b/src/bse_full_diag.F
@@ -475,7 +475,12 @@ CONTAINS
       ALLOCATE (Exc_ens(homo*virtual))
 
       CALL choose_eigv_solver(fm_C, fm_eigvec_Z, Exc_ens, diag_info)
-      CPASSERT(diag_info == 0)
+
+      IF (diag_info /= 0) THEN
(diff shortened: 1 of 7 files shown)


MGraml:
The issue here was two-fold:
Commit 67ea2cf3 enables automatically generated RI basis sets to work with OS (original issue mentioned in google groups) without an unhandled error.
However, these basis sets seem to be often not converged for OS. Therefore, I added a warning when OS and the basis set generator for RI are invoked.

PR #3911, opened 2025-01-28
Fix occurrence of ghost states with SVDs

### LINKED ISSUE #3823: Occupied ghost states with smearing without Cholesky
For tracking:
If I turn on smearing and turn off the Cholesky-based eigenvalue solver, CP2K introduces eigen states at energy zero. These states are then partially occupied resulting in wrong energies. The reason is that the root of the inverse overlap matrix still works in the AO base/full MO base projecting out the numerically small components leading to these components still being present but with eigenvalue zero. The workaround is to work in a reduced MO base. I already have a fix for it but still have to check all other places on how to deal with this issue (or if it is relevant at all).

fstein93:
Fixes #3823 . If the Cholesky decomposition is turned off for numerical reasons, CP2K approximates the inverse square roots of the overlap matrix using SVDs. This keeps the null-space during the solution of the eigenvalue problem such that additional eigenvectors with eigenvalue zero are found which are occupied if the Fermi-level $\mu\ge -kT$.
My PR removes the null-space entirely by working in the linear-independent subspace spanned by the eigenvectors outside of the null-space of the overlap matrix. I applied this scheme also to the level-shifting scheme and eigenvalue calculation for MP2. The reference values of the regtests were updated accordingly and are in better agreement with comparable regtests. I reintroduced the EPS_SVD keyword in the MP2 case.
@JWilhelm I tried Tikhonov regularization but found that it does not solve the problem in our case.

### DIFF
--- a/src/input_cp2k_mp2.F
+++ b/src/input_cp2k_mp2.F
@@ -2421,6 +2421,13 @@ CONTAINS
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
+      CALL keyword_create(keyword, __LOCATION__, name="EPS_SVD", &
+                          description="Determines the upper bound of eigenvectors to be removed during the SVD (see DO_SVD).", &
+                          usage="EPS_SVD  1E-5", &
+                          default_r_val=0.0_dp)
+      CALL section_add_keyword(section, keyword)
+      CALL keyword_release(keyword)
+
       CALL keyword_create(keyword, __LOCATION__, name="ERI_BLKSIZE", &
                           description="block sizes for tensors (only used if ERI_METHOD=MME). First value "// &
                           "is the block size for ORB basis, second value is the block size for RI_AUX basis.", &
--- a/src/mp2.F
+++ b/src/mp2.F
@@ -29,9 +29,12 @@ MODULE mp2
    USE cp_dbcsr_api,                    ONLY: dbcsr_get_info,&
                                               dbcsr_p_type
    USE cp_dbcsr_operations,             ONLY: copy_dbcsr_to_fm
-   USE cp_fm_basic_linalg,              ONLY: cp_fm_triangular_invert
+   USE cp_fm_basic_linalg,              ONLY: cp_fm_column_scale,&
+                                              cp_fm_syrk,&
+                                              cp_fm_triangular_invert,&
+                                              cp_fm_upper_to_full
    USE cp_fm_cholesky,                  ONLY: cp_fm_cholesky_decompose
-   USE cp_fm_diag,                      ONLY: cp_fm_power
+   USE cp_fm_diag,                      ONLY: choose_eigv_solver
    USE cp_fm_struct,                    ONLY: cp_fm_struct_create,&
                                               cp_fm_struct_release,&
                                               cp_fm_struct_type
@@ -39,6 +42,7 @@ MODULE mp2
                                               cp_fm_get_submatrix,&
                                               cp_fm_release,&
                                               cp_fm_set_all,&
+                                              cp_fm_to_fm,&
                                               cp_fm_type
    USE cp_log_handling,                 ONLY: cp_get_default_logger,&
                                               cp_logger_type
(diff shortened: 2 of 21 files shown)


hfp:
I am not yet fully sure about the cause, but with this PR, I am getting:

```text
 RI-RPA section
 --------------

  Number of removed MOs:                                                      68
  Number of available MOs:                                                  1342

 *  \___/                         Diagonal has wrong size                      *
 * / \                                                 dbcsr_operations.F:2922 *


            5 dbcsr_get_diag
            4 compute_vec_Sigma_x_minus_vxc_gw
            3 mp2_main
            2 qs_energies
            1 CP2K
```

My test case is `benchmarks/QS_low_scaling_GW/GW.inp` with 16 ranks on a single node/system (PSMP).

fstein93:
I am currently checking it. It is possible. Apparently, DBCSR silently creates matrices of the wrong (?) size.

fstein93:
I think it is in [here](https://github.com/cp2k/cp2k/blob/a1b1b75c35b7d5f14894dbfc2ab7ad369452db12/src/rpa_gw_sigma_x.F#L536). The code implicitly assumes that the number of MOs is the same as the number of AOs. Before this change, this was actually true, whereas I changed it to the size of the linear independent subspace.
@JWilhelm Is it correct/reasonable to resize mo_coeff_b and matrix_tmp to nao x nmo and matrix_tmp_2 to nmo x nmo?

hfp:
Another question is, if there is any test we can add which is showing the [issue](https://github.com/cp2k/cp2k/pull/3911#issuecomment-2621830767).

hfp:
( Btw, good work! )

JWilhelm:
> I think it is in [here](https://github.com/cp2k/cp2k/blob/a1b1b75c35b7d5f14894dbfc2ab7ad369452db12/src/rpa_gw_sigma_x.F#L536). The code implicitly assumes that the number of MOs is the same as the number of AOs. Before this change, this was actually true, whereas I changed it to the size of the linear independent subspace. @JWilhelm Is it correct/reasonable to resize mo_coeff_b and matrix_tmp to nao x nmo and matrix_tmp_2 to nmo x nmo?

Yes, the code here assumes n_AO = n_MO. Agreed, it should work to resize mo_coeff_b and matrix_tmp as you suggested.

fstein93:
Apparently, it is easier to just get all diagonal elements and copy the first n_MO elements to the actual array. I checked the GW-corrected energies for a regtest-like setup and found the same values as before. But the sigma_C contribution changes. Is this behavior expected? How does the GW-code deal with ghost states? If you want to run some tests, checkout [my branch](https://github.com/fstein93/cp2k/tree/Reduce_svd).

hfp:
I merged your branch and I can confirm it works for the mentioned case!

( As a side-note, either the fix or the SVD change itself seems to have caused a slight regression in performance [`benchmarks/QS_low_scaling_GW/GW.inp`]. Though, this is about generality/correctness rather than performance. )

fstein93:
I have opened the PR. I suppose that the differences in the correlation contributions to the self-energy are related to the falsely considered ghost states in the calculation.

hfp:
To wrap up https://github.com/cp2k/cp2k/pull/3911#issuecomment-2623793648, I tested current master against master with reverted commits `44d3f296a06ed44b6f3b263e9364c69f9d58d067`, `a24614b15a54e7cbb2bc67b5bd6f9c47c73e96e1`, and `cac185c219e31f8d4df52f216abe8cf7406e3b13` (in this order). I ran `benchmarks/QS_low_scaling_GW/GW.inp` found current master to be somewhat slower (~4%) with DBM consuming 40% more (device-)memory according to the recently introduced DBM memory statistics (see below).

```text
 -                                DBM STATISTICS                               -
 [...]
 Memory consumption                            Number of allocations  Size [MiB]
 Device                                                          108        1627   <-- 98 and 1166 before
 Host                                                             34        1085
```

Slower run is according to `tools/diff_cp2k.py`:

* compute_self_energy_cubic_gw
* dbt_reshape
* dbt_tas_reserve_blocks_index
* dbm_reserve_blocks

Perhaps these function just point-out somewhat more work rather than having gotten slower.

fstein93:
That's interesting because I only shrinked the MO matrices a bit (maybe 5-10 %). In `compute_self_energy_cubic_gw`, `dbcsr_get_diag` is called with a larger array for the diagonal elements which will probably not affect the timings of DBM/DBT.

hfp:
Yeah, it's hard to conclude something actionable (besides of some voodoo). I may take a look as we go forward.

Unrelated to this issue:
I noticed the get-diag routines in our linalg bits (not just dbcsr) are all about collecting diagonal elements into an array like allocated just for the sake of getting the elements. In this case, if the routine had an option to only take N elements you could take them into the desired destination right away. Anyway, I hope also DBM takes over more going forward.

PR #3917, opened 2025-01-29
DBM: statistics for pooled memory

hfp:
- Print number of allocations, and memory consumption.
- MPI: avoid memory allocation for reductions (count=1).
- MPI: dbm_mpi_max_uint64, dbm_mpi_sum_uint64 (unused).
- Stats are based on the max-metric over all ranks.

### DIFF
--- a/src/dbm/dbm_library.c
+++ b/src/dbm/dbm_library.c
@@ -8,7 +8,6 @@
 #include <assert.h>
 #include <inttypes.h>
 #include <omp.h>
-#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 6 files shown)


hfp:
Current output/print aggregates memory consumption over all ranks; not quite intuitive for multiple nodes/systems.

hfp:
PR #3913 had an open TODO (which is this PR). It also claimed improvements wrt memory consumption ("expectations").

Output **without** searching for suitable chunk (`benchmarks/QS_low_scaling_GW/GW.inp` @ 16 ranks @ 8 GPUs @ 1 system):

```
 Memory consumption                            Number of allocations  Size [MiB]
 Device                                                          152         935
 Host                                                             29        1207
```

Output **with** searching for suitable chunk (#3913):

```
 Memory consumption                            Number of allocations  Size [MiB]
 Device                                                          100        1085
 Host                                                             34        1085
```

In summary, a PSMP-based build which is targeting 8 GPU devices with 16 ranks of the mentioned workload yields:

* About 34% less number of allocation events per device.
* About 10% lower memory consumption on the host.

( updated output according to max-metric per rank )

PR #3921, opened 2025-01-30
DBM: activated optimization for single-value MPI reductions

hfp:
- Removed dbm_mpi_sum_uint64 (unused).
- Implemented #3917, now activated.


hfp:
Only `dbm_mpi_sum_int` is ever called with `count != 1` and hence allocates a temporary buffer.

PR #3924, opened 2025-01-30
[cmake] Enable hdf5 support in cmake

mtaillefumier:
- Add hdf5 dependency to cmake
- Updated the `cp2kConfig.cmake.in` file with current supported dependencies
- minor cleanup in the `CMakeLists.txt`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -841,7 +841,9 @@ if(CP2K_USE_LIBTORCH)
 endif()
 
 if(CP2K_USE_HDF5)
-  message("  - HDF5\n")
+  message("  - HDF5\n" "   - HDF5_VERSION : ${HDF5_VERSION}\n"
+          "   - HDF5_INCLUDE_DIRS : ${HDF5_INCLUDE_DIRS}\n"
+          "   - HDF5_LIBRARIES : ${HDF5_LIBRARIES}\n\n")
 endif()
 
 if(CP2K_USE_FFTW3)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,6 +118,7 @@ option(
   CP2K_USE_DLAF
   "Use DLA-Future for Cholesky decompositions and the solution of eigenvalue problems"
   OFF)
+option(CP2K_USE_HDF5 "Enable HDF5 support" OFF)
 option(CP2K_USE_LIBXSMM "Use libxsmm for small gemms" OFF)
 option(CP2K_USE_LIBSMEAGOL "Enable libsmeagol support" OFF)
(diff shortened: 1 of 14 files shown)


CMakeLists.txt
```diff
@@ -591,6 +593,10 @@ if(CP2K_USE_LIBXC)
   find_package(LibXC 7 REQUIRED)
 endif()
 
+if(CP2K_USE_HDF5)
+  find_package(HDF5 REQUIRED COMPONENTS C Fortran)
```
RMeli, line 585:
```suggestion
  find_package(HDF5 REQUIRED CONFIG COMPONENTS C Fortran)
```
Worth making the `CONFIG` requirement explicit?

CMakeLists.txt
```diff
@@ -832,6 +838,10 @@ if(CP2K_USE_LIBTORCH)
   message("  - LIBTORCH\n" "    - libraries : ${CP2K_LIBTORCH_LIBRARIES}\n\n")
 endif()
 
+if(CP2K_USE_HDF5)
+  message("  - HDF5\n")
```
RMeli: Worth printing out the following variables? `HDF5_VERSION`, `HDF5_INCLUDE_DIRS`, and `HDF5_LIBRARIES`?

CMakeLists.txt
RMeli: Actually, this does't work. Better to use the CMake-provided one (i.e. without `CONFIG`).

CMakeLists.txt
mtaillefumier: Yep. Just adding them. At least libraries and header files. I will add the version

CMakeLists.txt
mtaillefumier: I use the cmake one although there are inconsistencies with the targets naming. But that ok

CMakeLists.txt
```diff
@@ -832,6 +838,13 @@ if(CP2K_USE_LIBTORCH)
   message("  - LIBTORCH\n" "    - libraries : ${CP2K_LIBTORCH_LIBRARIES}\n\n")
 endif()
 
+if(CP2K_USE_HDF5)
+<<<<<<< HEAD
```
RMeli:
```suggestion
```

mtaillefumier: yeah I saw it :(

RMeli:
The latest CI failure is being fixed in https://github.com/cp2k/cp2k/pull/3919.

oschuett:
For the Ubuntu test you also need to add `libhdf5-dev` to here: 
https://github.com/cp2k/cp2k/blob/7dd82e8fc79ba901be116dc3b1d9a453e079d168/tools/docker/generate_dockerfiles.py#L453

PR #3926, opened 2025-01-30
Updated tools/diff_cp2k.py

hfp:
- Expected input format seems to have changed, or was not exercised.
- Split ENERGY line along white spaces and take the last item.
- Some code comments (terminal colors).

### DIFF
--- a/tools/diff_cp2k.py
+++ b/tools/diff_cp2k.py
@@ -86,8 +86,9 @@ def read_file(filename, field, special_keys, stats_keys):
                 if "NAMEOUT=" in line:
                     nameout[0] = line.split("=", 2)[1].strip()
                     continue
-                if "ENERGY| Total FORCE_EVAL ( QS ) energy " in line:
-                    nameout[1] = line.split(":", 2)[1].strip()
+                if "ENERGY| Total FORCE_EVAL ( QS ) energy" in line:
+                    split = line.split()
+                    nameout[1] = split[-1]  # last item
                     continue
                 if "DBCSR STATISTICS" not in line and nstats == 0:
                     continue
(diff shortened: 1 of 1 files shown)


hfp:
It can be considered to change to semaphore semantics (red/higher/worse, green/lower/better). IMHO current coloring appears a bit counter-intuitive.

alazzaro:
Thanks @hfp , I will check tomorrow.

hfp:
This PR is motivated since the tool crashes if two outputs are compared, e.g., stemming from `benchmarks/QS_low_scaling_GW/GW.inp`.

alazzaro:
The format of the energy I considered is:

```
ENERGY| Total FORCE_EVAL ( QS ) energy [a.u.]:           -35222.022153955636895
```

therefore I split on [":" to get the value](https://github.com/cp2k/cp2k/blob/44d3f296a06ed44b6f3b263e9364c69f9d58d067/tools/diff_cp2k.py#L90). With your change, you will split on "]", am I right?

I'm not sure which pattern we have in `benchmarks/QS_low_scaling_GW/GW.inp`, I would assume there is no ":"... 
Then I suggest a nested condition, where we first check `"ENERGY| Total FORCE_EVAL ( QS ) energy "`, then internally we check on which character to split.

hfp:
> I'm not sure which pattern we have in `benchmarks/QS_low_scaling_GW/GW.inp`, I would assume there is no ":"... 

Indeed, there is no `:` and the pattern for the mentioned workload looks like:

```text
 ENERGY| Total FORCE_EVAL ( QS ) energy [hartree]          -1847.427724668849805
```

hfp:
I will rework it according to your recommendation.

alazzaro:
About the colors, I agree with you. I remember that "blue" was used instead of "red" because it was easier to read on some terminal color combinations :) But, yes, I think we can use red (maybe we can add an option to disable colors at all).

There is another consideration: currently we are not checking the number of calls of a given timer, I have a version which shows the number of calls (`-f 0`), I can push it at some point...

tools/diff_cp2k.py
```diff
@@ -86,8 +86,9 @@ def read_file(filename, field, special_keys, stats_keys):
                 if "NAMEOUT=" in line:
                     nameout[0] = line.split("=", 2)[1].strip()
                     continue
-                if "ENERGY| Total FORCE_EVAL ( QS ) energy " in line:
-                    nameout[1] = line.split(":", 2)[1].strip()
+                if "ENERGY| Total FORCE_EVAL ( QS ) energy" in line:
+                    split = line.split()
+                    nameout[1] = split[-1]  # last item
```
alazzaro, line 91: very clever solution ;)

hfp:
Ok, I now split ENERGY line along white spaces and take the last item, which is hopefully the desired number. So, no separator particular separator is assumed. Also, stripping whitespaces is not needed then.

tools/diff_cp2k.py
hfp: Ready to merge?

alazzaro:
LGTM

PR #3927, opened 2025-01-31
GRID: avoid integer overflow

hfp:
- On higher core-count systems, the adjusted expressions can overflow.
- For example (384 threads), the lower bound appeared negative.
- Avoid subsequent OOB access.

### DIFF
--- a/src/grid/cpu/grid_cpu_task_list.c
+++ b/src/grid/cpu/grid_cpu_task_list.c
@@ -379,10 +379,11 @@ static void collocate_one_grid_level(
       const int actual_group_size = imin(group_size, nthreads - dest_thread);
       // Parallelize summation by dividing grid points across group members.
       const int rank = modulo(ithread, group_size); // position within the group
-      const int lb = (npts_local_total * rank) / actual_group_size;
-      const int ub = (npts_local_total * (rank + 1)) / actual_group_size;
+      const int64_t lb = ((int64_t)npts_local_total * rank) / actual_group_size;
+      const int64_t ub =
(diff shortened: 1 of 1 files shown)


hfp:
The bad news seems, there are more locations overflowing, e.g., DBCSR's `dbcsr_mm_multrec.F` (line 309).

mtaillefumier:
Out of curiosity, in what case would we run cp2k with 384 threads ?

hfp:
> Out of curiosity, in what case would we run cp2k with 384 threads ?

I did not use 384 threads per-se, I tried 16 and 32 MPI-ranks and had threads per rank only accordingly. I am looking more closely into the reproducer. I think OMP_NUM_THREADS was set explicitly, but perhaps calculated wrong in the environment script. In any case, the adjusted expressions are a bit prone to overflowing in an intermediate product.

hfp:
My local system which is causing the overflow also spawned/motivated a bug-report for DBCSR (https://github.com/cp2k/dbcsr/issues/886).

oschuett:
Sorry for the late reply. LGTM.

I think it's good to stress test the code and learn about its limits. That being said,  I doubt that we'll reach competitive performance with hundreds of threads in the foreseeable future, due to the mercilessness of Amdahl's law.

So while small fixes like these make sense, for larger problems I'd just leave an assert for future generations.

For the record, the problem with the current threading scheme is that it replicates the grid, which needs a lot of memory and a slow reduction. The solution would to partition the grid across threads - just like we already do for MPI ranks.

PR #3934, opened 2025-02-03
Add DeePMD support in CMake and enable it in CMake + toolchain CI

RMeli:
Mathieu is working on the Spack package, so CMake + Spack CI will come in a separate PR.

The patch is needed to fix a warning, so that we can keep `-Werror=dev`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -927,7 +927,7 @@ if(NOT CP2K_USE_DFTD4)
   message("   - DFTD4")
 endif()
 
-if(NOT CP2K_USE_DFTD4)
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -918,6 +927,10 @@ if(NOT CP2K_USE_DFTD4)
   message("   - DFTD4")
 endif()
 
+if(NOT CP2K_USE_DFTD4)
```
RMeli:
```suggestion
if(NOT CP2K_USE_DEEPMD)
```

oschuett:
Wow, you're really killing it 🥳

PR #3937, opened 2025-02-05
DBM: improvements and optimizations

hfp:
- Avoid stream-synchronization for copy-completion (GPU); at the expense of higher memory consumption (lifetime).
- Implement SMM-threshold (CPU/LIBXSMM/BLAS).

OpenCL backend:
- Adjusted valid range of kernel-parameter (BN).
- Adjusted default value (BN).

Makefile/Miniapp
- Consolidated MPI-compilation-
- Ensure gcc rather than cc.
- Adjusted CFLAGS (-fPIC).
- Fixed clean-target.

Other
- Purified some getter-like functions.

### DIFF
--- a/src/dbm/dbm_multiply_comm.c
+++ b/src/dbm/dbm_multiply_comm.c
@@ -210,13 +210,14 @@ static void fill_send_buffers(
       const dbm_shard_t *shard = &matrix->shards[ishard];
       const double *blk_data = &shard->data[blk->offset];
       const int row_size = plan->row_size, col_size = plan->col_size;
+      const int plan_size = row_size * col_size;
       const int irank = plan->rank;
 
       // The blk_send_data is ordered by rank, thread, and block.
       //   data_send_displ[irank]: Start of data for irank within blk_send_data.
       //   ndata_mythread[irank]: Current threads offset within data for irank.
       nblks_mythread[irank] -= 1;
-      ndata_mythread[irank] -= row_size * col_size;
+      ndata_mythread[irank] -= plan_size;
       const int offset = data_send_displ[irank] + ndata_mythread[irank];
       const int jblock = blks_send_displ[irank] + nblks_mythread[irank];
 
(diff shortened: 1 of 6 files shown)


src/dbm/dbm_multiply_comm.c
```diff
@@ -226,17 +227,17 @@ static void fill_send_buffers(
...
-            norm += element * element;
             data_send[offset + i * col_size + j] = element;
+            norm += element * element;
           }
         }
         blks_send[jblock].free_index = plan->blk->col;
         blks_send[jblock].sum_index = plan->blk->row;
       } else {
-        for (int i = 0; i < row_size * col_size; i++) {
+        memcpy(data_send + offset, blk_data, sizeof(double) * plan_size);
```
oschuett: Is the `memcpy` really cheaper if we're looping over the same data again right afterwards? I would expect that the `element` could just remain in the register.

src/dbm/dbm_multiply_gpu.c
```diff
@@ -116,47 +115,55 @@ void dbm_multiply_gpu_process_batch(const int ntasks, const dbm_task_t *batch,
...
   // Reallocate shard_c_dev->data if necessary.
+  double *old_data_dev = NULL;
   if (shard_c_host->data_promised > shard_c_dev->data_allocated) {
-    double *old_data_dev = shard_c_dev->data;
     shard_c_dev->data_allocated =
         ALLOCATION_FACTOR * shard_c_host->data_promised;
+    old_data_dev = shard_c_dev->data;
     shard_c_dev->data =
         dbm_mempool_device_malloc(shard_c_dev->data_allocated * sizeof(double));
+    // Omit to wait for copy before freeing old buffer.
```
oschuett, line 135: Nice!

src/dbm/dbm_multiply_comm.c
hfp: No it's likely not cheaper. It actually looks nicer in the profiler now that all such zeroing tasks are summarized under the memset primitive. It's also not costlier since it is (hopefully) in-cache after memset and owned by the core in question. Also, memset is highly optimized in GLIBC; just in case...

hfp: ... but I can change it back if you prefer.

hfp: Reverted back to in-loop update.

PR #3938, opened 2025-02-05
DBM: improved filtering matrix

hfp:
- Early-exit if norm is above threshold.

### DIFF
--- a/src/dbm/dbm_matrix.c
+++ b/src/dbm/dbm_matrix.c
@@ -300,7 +300,11 @@ void dbm_filter(dbm_matrix_t *matrix, const double eps) {
       const int block_size = row_size * col_size;
       double norm = 0.0;
       for (int i = 0; i < block_size; i++) {
-        norm += old_blk_data[i] * old_blk_data[i];
+        const double value = old_blk_data[i];
+        norm += value * value;
+        if (eps2 <= norm) {
+          break;
+        }
       }
       // For historic reasons zero-sized blocks are never filtered.
       if (block_size > 0 && norm < eps2) {
@@ -311,7 +315,7 @@ void dbm_filter(dbm_matrix_t *matrix, const double eps) {
           shard, old_blk.row, old_blk.col, block_size);
       assert(new_blk->offset <= old_blk.offset);
       if (new_blk->offset != old_blk.offset) {
-        // Using memmove instead of memcpy because it handles overlap correctly.
+        // Using memmove because it can handle overlapping buffers.
         double *new_blk_data = &shard->data[new_blk->offset];
         memmove(new_blk_data, old_blk_data, block_size * sizeof(double));
       }


src/dbm/dbm_matrix.c
```diff
@@ -294,26 +294,30 @@ void dbm_filter(dbm_matrix_t *matrix, const double eps) {
...
-      }
-      // Re-create block.
-      dbm_block_t *new_blk = dbm_shard_promise_new_block(
-          shard, old_blk.row, old_blk.col, block_size);
-      assert(new_blk->offset <= old_blk.offset);
-      if (new_blk->offset != old_blk.offset) {
-        // Using memmove instead of memcpy because it handles overlap correctly.
-        double *new_blk_data = &shard->data[new_blk->offset];
-        memmove(new_blk_data, old_blk_data, block_size * sizeof(double));
+      if (block_size > 0) {
```
oschuett: Wouldn't it be enough to just slip in a single if-statement like this:

```
double norm = 0.0;
for (int i = 0; i < block_size; i++) {
   norm += old_blk_data[i] * old_blk_data[i];
   if(norm > eps2) break;  // we know enough - leave early
}
```

hfp: Yes, seems like - you got me! It is a remainder of my transformation (not removing it). Now that I look at, the condition can be removed entirely because of `for (int i = 0; i < block_size; i++)`, i.e., block_size is tested already.

src/dbm/dbm_matrix.c
```diff
@@ -299,21 +299,22 @@ void dbm_filter(dbm_matrix_t *matrix, const double eps) {
...
+          // Re-create block.
+          dbm_block_t *new_blk = dbm_shard_promise_new_block(
+              shard, old_blk.row, old_blk.col, block_size);
+          assert(new_blk->offset <= old_blk.offset);
+          if (new_blk->offset != old_blk.offset) {
+            // Using memmove because it can handle overlapping buffers.
+            double *new_blk_data = &shard->data[new_blk->offset];
+            memmove(new_blk_data, old_blk_data, block_size * sizeof(double));
+          }
+          break;
```
oschuett, line 306: What's the advantage of conflating the norm calculation with the later copy?
Personally, I prefer to keep the planing and the execution parts separate.
Wouldn't just slipping in that `if (norm > eps2) break;` be all the change we need?

hfp: Yeah, very good point!

I guess it is bad habit/style on my end. Motivation is to have single-exit style, but it comes at the cost of stacking/indenting conditions.

hfp: > Wouldn't just slipping in that if (norm > eps2) break; be all the change we need?

Hmm, then I need to retest the condition because I could have existed the loop for just reaching block_size.

oschuett: So, you're saying that addition floating point comparison matters? But then it'll also hurt the loop performance, right?

Maybe we could do this:
```
double eps_minus_norm = eps2;
for (int i = 0; i < block_size; i++) {
   const double value = old_blk_data[i];
   eps_minus_norm -= value * value;
  if(eps_minus_norm < 0.0) break;  // we only need to know the sign
}

if (block_size > 0 && eps_minus_norm  < 0.0) {
   continue; // filter the block
}
```

This way the condition should be much cheaper as only the sign bit has to be inspected.

hfp: How about now?

The clang-format makes it a bit bulky, but loop is not inflected anymore. Anyway, the main saving with this PR is skipping unnecessary loads from `old_blk_data`.

oschuett: Hmm, I don't know. Re-purposing the `block_size` is a bit confusing. (btw, there is a `break` statement missing).

If you just want to avoid the loads then `if (norm > eps2) break;` would be least invasive. If you also care about the floating point comparisons then the `eps_minus_norm` solution should be just as fast - or is the compiler not smart enough to look only at the sign bit?

hfp: Now it's the minimal change, just the early-out for collecting the norm.

PR #3940, opened 2025-02-05
ccECP

juerghutter:
CP2K integral routine for local part of ECPs. Replaces and extends local (Type 1) integrals from libgrrp.
Makes pseudopoentials used in Quantum Monte Carlo accessible to CP2K,
see pseudopotentials.org


annulen:
For those coming here from release notes of CP2K v2025.2:

Correct link (right now, maybe it has moved since February) is https://pseudopotentiallibrary.org, not pseudopotentials.org

PR #3943, opened 2025-02-05
Enable hardware support for atomic operations on mi250 

mtaillefumier:
The compilation flag controlling the hardware acceleration of atomic operations on Mi250X was deleted. This PR solves this issue. This flags is on by default on Mi300.

The PR also contains few cleanups

- The `CP2K_DEBUG_MODE` is removed
- Use `CP2K_ENABLE_{PW,GRID,DBM}_GPU` to turn on/off `GPU` support.
The default value is `ON` for `GRID` and `DBM` and `OFF` for `PW`. It should be turned back on when the HIP issue about multiple mpi ranks per GPU is solved.
- Set HIP compilation flags properly

PR #3948, opened 2025-02-06
Optimize the transformation spherial -> cartesian in collocate

mtaillefumier:
The collocate function (not the kernel, the entire function) is 3 times as slow on on Mi250x than on V100. This pr fixes the problem with added bonus it is up to 10 times faster than earlier on Mi250x.

### DIFF
--- a/src/grid/hip/grid_hip_collocate.cu
+++ b/src/grid/hip/grid_hip_collocate.cu
@@ -44,41 +44,45 @@ __device__ __inline__ void block_to_cab(const kernel_params &params,
   // two products are fused to conserve shared memory.
   const int tid =
       threadIdx.x + blockDim.x * (threadIdx.y + blockDim.y * threadIdx.z);
-  for (int i = 0; i < task.nsgf_setb; i++) {
-    for (int j = 0; j < task.nsgf_seta; j++) {
-      T block_val;
-      if (task.block_transposed) {
-        block_val = task.pab_block[j * task.nsgfb + i] * task.off_diag_twice;
-      } else {
-        block_val = task.pab_block[i * task.nsgfa + j] * task.off_diag_twice;
+
+  for (int jco = task.first_cosetb + tid / 16; jco < task.ncosetb; jco += 4) {
+    for (int ico = task.first_coseta + (tid % 16); ico < task.ncoseta;
+         ico += 16) {
+      T pab_val = 0.0;
+      for (int i = 0; i < task.nsgf_setb; i++) {
+        const T sphib = task.sphib[i * task.maxcob + jco];
+        T tmp_val = 0.0;
+        for (int j = 0; j < task.nsgf_seta; j++) {
+          T block_val;
+          if (task.block_transposed) {
+            block_val =
+                task.pab_block[j * task.nsgfb + i] * task.off_diag_twice;
+          } else {
+            block_val =
+                task.pab_block[i * task.nsgfa + j] * task.off_diag_twice;
+          }
+          //          const T sphia = task.sphia[j * task.maxcoa + ico];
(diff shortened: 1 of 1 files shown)


mtaillefumier:
We get the following timings for collocate/integrate using the same input file than for PR #3943 (large l). MPI_RANK=1, OMP=4

Without this PR

```
grid_collocate_task_list           124  9.3   60.267   60.267   60.267   60.267
 grid_integrate_task_list           124 12.1   19.710   19.710   19.710   19.710
```

with this PR

```
grid_integrate_task_list           124 12.1   19.760   19.760   19.760   19.760
grid_collocate_task_list           124  9.3    6.121    6.121    6.121    6.121
```

oschuett:
I'm pretty sure I also experimented with the loop order when developing the CUDA kernels. However, I used a V100 (which has native atomic adds) and only looked at small basis-sets. IIRC, in that regime the coalesced access still gave a slight edge. Nevertheless, your solution is obviously better as it works wonders for larger basis sets by avoiding a ton of atomics.

Great work!

mtaillefumier:
I have the numbers for V100. I will test the conventional benchmarks H2O-64 tomorrow as well so that we have complete picture of what is going on.

mtaillefumier:
same input file with hip on v100. 

without this pr
```
grid_collocate_task_list           124  9.3   22.155   22.155   22.155   22.155
grid_integrate_task_list           124 12.1   26.940   26.940   26.940   26.940
```
with this PR
```
grid_integrate_task_list           124 12.1   26.934   26.934   26.934   26.934
grid_collocate_task_list           124  9.3   10.290   10.290   10.290   10.290
```
I will test the small basis set case tomorrow hopefully

mtaillefumier:
hip and gpu on master give similar timings.

mtaillefumier:
I ran the `H2O-64-nonortho` benchmark to measure the difference and I get these numbers on V100. It is not really a small basis set but still.

hip improvements (63df77bb0e)
```
grid_integrate_task_list           107 12.3    6.579    6.579    6.579    6.579
grid_collocate_task_list           107  9.7    6.578    6.578    6.578    6.578
```

master

GPU:
```
grid_integrate_task_list           107 12.3    7.925    7.925    7.925    7.925
grid_collocate_task_list           107  9.7    7.812    7.812    7.812    7.812
```
hip:
```
grid_collocate_task_list           107  9.7    7.084    7.084    7.084    7.084
grid_integrate_task_list           107 12.3    6.587    6.587    6.587    6.587
```

I would be more than happy to measure a small basis set case.

mtaillefumier:
we will need to propagate these modifications in the GPU backend eventually as users from ALPS are using it by default.

hfp:
> This is reproducible by taking the usual H2O-XX.inp DFT benchmarks, and changing to larger basis sets and meta-GGA functionals.

@mtaillefumier Can you be so kind and attach such a demanding input file for reference like high quantum number?

oschuett:
> we will need to propagate these modifications in the GPU backend

Yes, if you have the time, please feel free to swap the loop order also on the GPU backend. Otherwise, I'll try to do it in the next weeks.

As for your other optimization regarding [cxyz_to_gridpoint](https://github.com/cp2k/cp2k/blob/a2ea924e12dcaa12bfa2d0afae7fe185538a91df/src/grid/gpu/grid_gpu_collocate.cu#L35), I'd prefer to wait for the SymPy solution.

mtaillefumier:
> > This is reproducible by taking the usual H2O-XX.inp DFT benchmarks, and changing to larger basis sets and meta-GGA functionals.
> 
> @mtaillefumier Can you be so kind and attach such a demanding input file for reference like high quantum number?

I will do it next week or so. I will not touch collocate kernel except for the loop inversion.

PR #3958, opened 2025-02-09
added INTENSITY_LIST, START_FRAME and END_FRAME keywords for periodic efield

LeoDecking:
With these changes, it is now possible to give a list of intensity values for the periodic efield which are applied one after each other starting and ending in the specified frames. If the number of frames exceeds the number of values, they are cyclically repeated.
These keywords are implemented only for qs_efield_berry and efield_tb_methods. Implementation for eeq_method is still open.

Fixed the problems of #3957, the start_frame defaults to 0 now.


src/efield_tb_methods.F
```diff
@@ -91,7 +91,12 @@ SUBROUTINE efield_tb_matrix(qs_env, ks_matrix, rho, mcharge, energy, calculate_f
       energy%efield = 0.0_dp
       CALL get_qs_env(qs_env, dft_control=dft_control)
       IF (dft_control%qs_control%dftb .OR. dft_control%qs_control%xtb) THEN
-         IF (dft_control%apply_period_efield) THEN
+         if_apply: IF (dft_control%apply_period_efield) THEN
+            ! check if the periodic efield should be applied in the current step
+            IF (dft_control%period_efield%start_frame > qs_env%sim_step) EXIT if_apply
```
oschuett: I didn't realize one can use `EXIT` on if-blocks. This smells a lot like GOTO. A better solution would be to use something like this: `IF... ELSE IF...ELSE IF... ELSE..END IF`.

src/input_cp2k_field.F
```diff
@@ -89,13 +90,43 @@ SUBROUTINE create_per_efield_section(section)
...
       CALL keyword_create(keyword, __LOCATION__, name="D_FILTER", &
-                          description="Filter for displacement field (x,y,z-dirction)", &
+                          description="Filter for displacement field (x,y,z-direction)", &
                           usage="D_FILTER  1.0 0.0 0.0", &
                           repeats=.FALSE., n_var=3, &
                           type_of_var=real_t, default_r_vals=(/1.0_dp, 1.0_dp, 1.0_dp/))
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
+      CALL keyword_create(keyword, __LOCATION__, name="INTENSITY_LIST", &
```
oschuett, line 100: Could you please add a regtest that covers this new features.

oschuett (commented):
Thanks a lot for contributing!

> Fixed the problems of #3957, ...

There is no need for a new PR, you can still push changes to a branch after opening a PR.

tests/QS/regtest-p-efield/TEST_FILES
```diff
@@ -8,6 +8,7 @@ H2O-field-gopt.inp                                     1      3e-11
 H2O-field-gopt-lsd.inp                                 1      4e-12             -16.84348812829990
 H2O-field.inp                                          1      4e-13             -15.90320150373305
 H2O-field-lsd.inp                                      1      4e-14             -15.92914916395207
+H20-field-list.inp                                     2      1e-14            -0.171095788220e+02
```
oschuett: This is the number zero instead of the capital letter O.

LeoDecking: oops, thank you

src/efield_tb_methods.F
LeoDecking: I fixed that, thanks for noticing

PR #3965, opened 2025-02-10
Append token for MPI implementation

mkrack:


oschuett:
Maybe we should move the toolchain part into a separate `build_cp2k_toolchain.sh` script.

Or could we perhaps change the behavior of the existing [install_cp2k_toolchain.sh](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/install_cp2k_toolchain.sh) so that a wrapper is not needed?

mkrack:
It would indeed better to disentangle the `toolchain/cmake` builds with and without a container environment. I will look into that.

oschuett:
Yeah, it's a bit of a mess right now (as with every migration).

Eventually, I'd like to reach the following state:
1. Remove [docker/scripts/install_dbcsr.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/install_dbcsr.sh) and use [toolchain/scripts/stage9/install_dbcsr.sh](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/stage9/install_dbcsr.sh) instead. Note, that the later script will have to be able to build multiple version of DBCSR (ssmp, psmp, gpu, ...).
2. Have the toolchain generate the command line for CMake,  similar to how it now [generates the arch files](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/generate_arch_files.sh).

Since the toolchain is already quite complex, it won't be easy to add those features immediately. Fortunately, most of the complexity will go away once we can drop the old Makefile. So, the [build_cp2k_cmake.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/build_cp2k_cmake.sh) script was mostly intended as an interim solution to resolve this chicken-and-egg situation.

mkrack:
> Yeah, it's a bit of a mess right now (as with every migration).
> 
> Eventually, I'd like to reach the following state:
> 
>     1. Remove [docker/scripts/install_dbcsr.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/install_dbcsr.sh) and use [toolchain/scripts/stage9/install_dbcsr.sh](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/stage9/install_dbcsr.sh) instead. Note, that the later script will have to be able to build multiple version of DBCSR (ssmp, psmp, gpu, ...).
> 
>     2. Have the toolchain generate the command line for CMake,  similar to how it now [generates the arch files](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/generate_arch_files.sh).
> 
> 
> Since the toolchain is already quite complex, it won't be easy to add those features immediately. Fortunately, most of the complexity will go away once we can drop the old Makefile. So, the [build_cp2k_cmake.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/build_cp2k_cmake.sh) script was mostly intended as an interim solution to resolve this chicken-and-egg situation.

Thanks @oschuett I will address that in future commits.

PR #3978, opened 2025-02-12
Include DBCSR library in toolchain build

mkrack:


alazzaro:
This is a great change! I would assume we can remove the submodule (the entire [exts](https://github.com/cp2k/cp2k/tree/master/exts) directory?). 
Are the changes propagated to the arch files? (we can drop some options, but then we have to change LIBS and FCFLAGS)

mkrack:
Indeed, the goal is to get rid of the last submodule and the related `exts` folder. This requires also a major change of the Makefile and of all arch files. I am wondering, however, if it is worth the effort to update Makefile and arch files, given the fact that we are in the process of moving to `cmake`?

The docker file generator for the regression tests is not affected by this PR, because `stage9` of the toolchain installation is not yet included. This will be part of a follow-up PR.

oschuett:
> I am wondering, however, if it is worth the effort to update Makefile and arch files, given the fact that we are in the process of moving to cmake?

We are on track to deprecate the Makefile with the next release. Once the release is out we can remove the Makefile and the submodule. So, I'd just leave it until then.

tagging #351

PR #3983, opened 2025-02-14
cmake: missing link-directory for libxsmm

hfp:
- Mention CP2K_LIBXSMM_LIBRARY_DIRS (mark advanced).
- Revised related CMake module slightly.


mtaillefumier:
Thanks. One thought. It is better to associate the `prefix/lib` to the libxsmm target itself not the final library although here it is a bit less relevant. 

```
add_library(cp2k::LibXSMM::${__lib} INTERFACE IMPORTED)
target_link_directories([cp2k::LibXSMM::${__lib} INTERFACE
                          ${CP2K_LIBXSMM_LIBRARY_DIRS})
```
instead of 
```
if(CP2K_USE_LIBXSMM AND CP2K_LIBXSMM_LIBRARY_DIRS)
  target_link_directories(cp2k_linalg_libs INTERFACE
                          ${CP2K_LIBXSMM_LIBRARY_DIRS})
endif()
```

hfp:
Still a good point, will adjust it.

PR #3991, opened 2025-02-15
Use 2-stage dockerfiles for CSCS CI/CD

mkrack:
to reduce/minimize image size for production container deployment


mkrack:
cscs-ci run

Issue #3986, opened 2025-02-14
Docs: Typos in usage strings


oschuett:
As reported by https://github.com/cp2k/cp2k/discussions/3979, some of the usage strings in the manual have misspellings.
I've added a warning to detect them in #3982. In total there are currently 385 bad usage strings.

In case somebody wants to start chipping away, the currently list can be obtained like this:

```
wget -q  -O-  https://storage.googleapis.com/cp2k-ci/dashboard_manual_report.txt | grep "has bad usage:"
```

Once all the typos are fixed, please arm the check in [keyword_create](https://github.com/cp2k/cp2k/blob/b0259176ecd3425dd931caf2a3ec7271d03effe6/src/input/input_keyword_types.F#L233-L236).

PR #3996, opened 2025-02-17
Fix typos in usage strings

AsukaHanaT:
As per #3986, fixed all the typos in usage string in manuals according to their keyword names.

Only one problem, the keyword CP2K_INPUT.MOTION.TMC.GROUP_ANLYSIS_NR itself is misspelled (should be CP2K_INPUT.MOTION.TMC.GROUP_ANALYSIS_NR) and might need additional fixing. I changed the usage string according to the misspelled version of the keyword.


oschuett:
Awesome! Thanks a lot for taking care of this :-)

As for `GROUP_ANLYSIS_NR`, [I'll just rename it](https://github.com/cp2k/cp2k/pull/4002) because I don't think it's much used.

PR #3997, opened 2025-02-17
Partially revert #3992 and apply alternative fix

mkrack:


mkrack:
Fix typo in AiiDA test in follow-up PR.

PR #4006, opened 2025-02-18
Install MPI compatible with Sarus; add minimal/psmp

mkrack:


ci/docker/build_cp2k_psmp.Dockerfile
```diff
@@ -20,16 +29,55 @@ RUN apt-get update -qq && apt-get install -qq --no-install-recommends \
     wget \
     zlib1g-dev
 
+# Retrieve the number of available CPU cores
+ARG NUM_PROCS=32
+
+# Install an MPI version ABI-compatible with the host MPI on Cray at CSCS
+ARG MPICH_VER="4.2.3"
```
RMeli: On Eiger, a full MPICH replacement is performed by Sarus. Since Cray-MPICH is based on `mpich@3`, this version should be used in the container. `mpich@3.1.4` is known to work. `mpich@4` might or might not work, but it needs to be tested and I'm unable to provide a version known to work.

On Daint, the [container engine](https://confluence.cscs.ch/spaces/KB/pages/776306695/Container+Engine) is used and only libfabric is replaced. Therefore, `mpich@4` does work, as long as the container uses a `libfabric` version lower than `1.20`.

This is not really documented yet on our side unfortunately, because there will soon be a system upgrade which will likely change things.

mkrack: Thanks @RMeli I tried different MPICH versions and I can confirm what you write. The suggested version `mpich@3.1.4` works, but the recent versions `mpich@4.3.0` and `mpich@4.2.3` (currently installed by the CP2K toolchain) do not work. That's a bit unfortunate, because the `mpich` installation has to be done in the CI docker file.

PR #4010, opened 2025-02-20
Update CSCS CI workflow for eiger

mkrack:
- Add a new test folder CI with a small, medium and large water benchmark test input
- Reduce test output size
- Run medium test only
- Apply more specific pipeline task names


mkrack:
cscs-ci run

PR #4013, opened 2025-02-20
Test Cray environment

mkrack:


mkrack:
cscs-ci run

mkrack:
cscs-ci run

RMeli:
I suggest not to spend time trying to use the CPE on Eiger. While it is still technically supported, in the coming months Eiger will be updated to better align with Daint. 

We recommend using [Spack-based User Environments](https://confluence.cscs.ch/spaces/KB/pages/701726829/uenv+user+environments) or the [containers](https://confluence.cscs.ch/spaces/KB/pages/811466794/Containers) on our systems. System-wide modules might disappear in the near future, and it is likely that the CPE will only be available within a container going forward.

mkrack:
Thanks @RMeli Although the CI build of CP2K followed by a benchmark test run on Eiger at CSCS works now, there are still some issues:
- I do not see how to run a CP2K regression test using the `do_regtest.py` script within a single SLURM batch job without a sub-scheduler. The internal farming of CP2K, which was used in the past as a work around, is not a good alternative and moving to ctest is not a quick solution.
- While single-node jobs run fine, multi-node jobs crash with an BLACS error in the PW setup. It looks like an MPI problem due to inter-node communication.

RMeli:
> I do not see how to run a CP2K regression test using the do_regtest.py script within a single SLURM batch job without a sub-scheduler.

We were discussing this just the other day. At the moment, when we deploy on Alps, we use `do_regtest.py` with a single worker, which consists in submitting one job at a time; it's suboptimal and very time consuming. I noticed in the [old (?) documentation](https://www.cp2k.org/dev:regtesting) that there was a `--farming` option, but this is no longer available, is it? @mtaillefumier has a custom script he uses on Frontiers, he might be able to help out here. 

Migrating to `ctest` as suggested in https://github.com/cp2k/cp2k/issues/3314 would indeed be ideal. However, there is currently an issue on Daint right now, where when running with the container engine one needs `--mpi=pmi2`, which has a caveat that you can run only exactly once an MPI program with `srun`. But this is not the case on Eiger, which is what we should focus on first (and we have some workarounds, that we can discuss once we tackle Daint).

> While single-node jobs run fine, multi-node jobs crash with an BLACS error in the PW setup. It looks like an MPI problem due to inter-node communication.

Do you have a link to the pipeline that failed by any chance, so that I can have a deeper look?

oschuett:
While I don't think it's feasible to adopt ctest, maybe we can cherry-pick from it. Do you know how it manages to run multiple MPI jobs within the confines of Alps' SLURM setup?

mkrack:
@RMeli PR #4015 should consume two Eiger nodes for the test run in the [CI pipeline](https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/3546923033037693/3295760759903019/-/pipelines/1681433250).
The [farming option](https://manual.cp2k.org/trunk/CP2K_INPUT/FARMING.html) is still available in CP2K. It is not as elegant and versatile as `do_regtest.py`, but a test run does not take ages which is the case for `do_regtest.py` with a single worker. Is there any SLURM sub-scheduler available ([Greasy](https://user.cscs.ch/tools/high_throughput/)?) and supported by CSCS which could be an alternative?

Another issue, which I forgot to mention, is the performance of the CP2k container binary. The first SCF step is very slow due to a very bad performance of `cp_fm_cholesky_invert` used in that step. If I build the docker container with the same docker file on my local PC, I do not observe that performance bottleneck.

oschuett:
Since  for the regtests a single node is more than enough, I wonder if we could just call `mpiexec` instead of `srun`

alazzaro:
Maybe unrelated.
Last time I tried the do_regtest.py on LUMI it was running fine (although it was without containers)... That was ~ a year ago.
You can do salloc to reserve the nodes and then `srun --overlap` to get multiple srun's execution. Need to make sure that there is no binding (--bind=none).

mkrack:
The CSCS CI pipeline is already launching an `srun` command and running `mpiexec` from the container or another `srun` does not work. Tried these options already, but maybe I am missing something. 

Given these restrictions, I will give a try to farming again.

mtaillefumier:
it is possible to use 'srun' for the regtests but it is costly because each call to srun imply communication with the scheduler even in an already allocated job. the overlap option without binding is also problematic because each job see the entire node. My solution is not perfect either as I still call srun in the background. I would really like to have the possibility to call mpiexec directly on frontier for instance but i am not even sure it is possible.

oschuett:
The farming solution is really not great because any failure brings down the entire job, which is a pain to debug.

If the CSCS-CI is too restrictive, we could also explore to hookup  [FirecREST](https://user.cscs.ch/tools/firecrest/) to the CP2K-CI.

Or we just settle for some daily dashboard tests.

alazzaro:
> it is possible to use 'srun' for the regtests but it is costly because each call to srun imply communication with the scheduler even in an already allocated job. the overlap option without binding is also problematic because each job see the entire node. My solution is not perfect either as I still call srun in the background. I would really like to have the possibility to call mpiexec directly on frontier for instance but i am not even sure it is possible.

mpixec should not be there... There can be mpirun, but it is a wrapper around srun...
On the other side, if you stay in a node, it should be possible to install your own MPI with mpirun...

mkrack:
Let's move this discussion to #4019

PR #4015, opened 2025-02-20
Launch CI test run on two eiger nodes

mkrack:


mkrack:
cscs-ci run

alazzaro:
@mkrack Interesting enough, we saw on LUMI the same error message reported at https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/3546923033037693/3295760759903019/-/jobs/9202439783 . It turns out that the offending part was due to SPLA...

mkrack:
> @mkrack Interesting enough, we saw on LUMI the same error message reported at https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/3546923033037693/3295760759903019/-/jobs/9202439783 . It turns out that the offending part was due to SPLA...

Interesting, do you have further inside or do you know a fix or a work-around?

alazzaro:
LUMI decision was to drop SPLA, see https://github.com/Lumi-supercomputer/LUMI-EasyBuild-contrib/commit/ae77a33a89b69198409122c15df507d61a5edaec . Note that the problem was only on the GPU stack...

mkrack:
> LUMI decision was to drop SPLA, see [Lumi-supercomputer/LUMI-EasyBuild-contrib@ae77a33](https://github.com/Lumi-supercomputer/LUMI-EasyBuild-contrib/commit/ae77a33a89b69198409122c15df507d61a5edaec) . Note that the problem was only on the GPU stack...

OK, then that's not related to this problem at CSCS on Eiger. The test run is also crashing with the `minimal/psmp` CP2K build which does not include `SpLA`.

PR #4032, opened 2025-02-27
DBM: hyper-parameters and other improvements

hfp:
- Streamlined dbm_mempool_host_malloc vs dbm_mpi_alloc_mem.
  - Leave contended/temporary allocation up to MPI runtime.
- Removed hashtable_mask from internal dbm_shard_t.
- Moved duplicate imax into internal header
  - Made imax a macro for some type-flexibility.
  - Renamed dbm_multiply_internal.h.
- Prefixed hyper-parameters and avoid name clashes.
- Made OpenMP dynamic schedule a hyper-parameter.

### DIFF
--- a/src/dbm/dbm_internal.h
+++ b/src/dbm/dbm_internal.h
@@ -5,8 +5,13 @@
 /*  SPDX-License-Identifier: BSD-3-Clause                                     */
 /*----------------------------------------------------------------------------*/
 
-#ifndef DBM_MULTIPLY_INTERNAL_H
-#define DBM_MULTIPLY_INTERNAL_H
+#ifndef DBM_INTERNAL_H
+#define DBM_INTERNAL_H
+
+/*******************************************************************************
+ * \brief Returns the larger of two given integer (missing from the C standard)
+ ******************************************************************************/
+#define DBM_MAX(X, Y) ((X) > (Y) ? (X) : (Y))
 
 /*******************************************************************************
  * \brief Internal struct for storing a dbm_block_t plus its norm.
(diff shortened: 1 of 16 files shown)


hfp:
This the initial change of a series aiming to significantly improving memory consumption as well as lowering the number of allocations necessary (pooled memory). This is the preparation for a follow-on PR of #3904.

src/dbm/dbm_internal.h
```diff
@@ -5,8 +5,13 @@
...
 
-#ifndef DBM_MULTIPLY_INTERNAL_H
-#define DBM_MULTIPLY_INTERNAL_H
+#ifndef DBM_INTERNAL_H
+#define DBM_INTERNAL_H
+
+/*******************************************************************************
+ * \brief Returns the larger of two given integer (missing from the C standard)
+ ******************************************************************************/
+#define DBM_MAX(X, Y) ((X) > (Y) ? (X) : (Y))
```
oschuett, line 14: Could we turn this into a function again? Performance-wise it should be the same, but it's easier to read and has additional type-safety.

hfp: Will do. I was about to implement the respective function yet again in another module, then I put it into the header, then I realized, I need another type-combination not just `int` (but rather unsigned long int or size_t). Currently, it's only used as int-flavor. I will turn it back into a function.

hfp: ( btw, I removed the author name since I did not want to "let you write that macro". I will bring it back also ;-)

PR #4034, opened 2025-02-27
DBM: adjusted heuristic of pooled allocation

hfp:
- On top of #3904.


hfp:
To qualify the claim, `benchmarks/QS_low_scaling_GW/GW.inp` with 16 ranks on 8 GPU-devices (single-node) shows 24% reduction in device-memory consumption and about 18% less allocation events. For CPU-only, the improvements are expected to be similar.

PR #4035, opened 2025-02-27
GW/BSE: Fix indices in case of removed null-spaces for SVDs 

MGraml:
In this PR, I adapt parts of the GW and BSE code to the changes introduced in commits cac185c219e31f8d4df52f216abe8cf7406e3b13 (#3911) and a24614b15a54e7cbb2bc67b5bd6f9c47c73e96e1 (#3922), which cause the GW/BSE module to abort if `CHOLESKY OFF` is invoked.
Mainly, I fix the matrix sizes to take into account the removed null-space, which also cause the GW part to abort if `CORR_MOS_VIRT ` is set negative or to large numbers. I add tests for BSE calculations in which MOs are removed and indices are derived within the code.
I checked that the new BSE results are identical with BSE results before cac185c219e31f8d4df52f216abe8cf7406e3b13.
I also adapt the lines [rpa_gw_sigma_x.F#L966](https://github.com/MGraml/cp2k/blob/5dc9f155f97fc66276de3f1822d73d91b69ed0b4/src/rpa_gw_sigma_x.F#L966) since they seem fairly similar and I assume the same abort will occur there (for large `CORR_MOS_VIRT `).
Besides that, I add two lines in src/qs_scf_methods.F, which take care of `CHOLESKY OFF` in combination with no removed MOs (as for example for very small `EPS_EIGVAL`).
I would be happy about a second opinion @fstein93. If there are any suggestions, I am happy to include them.


oschuett:
I'm afraid I've created a conflict for this PR because I didn't want to merge it before @fstein93 has reviewed it. Sorry!

I hope you will find that the new [TEST_FILES.toml](https://github.com/cp2k/cp2k/blob/master/tests/QS/regtest-bse/TEST_FILES.toml) is rather similar.

MGraml:
@oschuett Don't worry, should be fixed. Sorry, I have never done a rebase during a PR, but the history seems to be fine now.

oschuett:
>  Sorry, I have never done a rebase during a PR, but the history seems to be fine now.

Yeah, rebaseing is tricky. It seems you are missing the two follow up PRs I made to fix the import of `toml`.

MGraml:
Yes, apparently ^^ Now, the new do_regtest.py should include these PRs.

fstein93:
Sorry, I am on holidays. LGTM.

fstein93 (approved):

MGraml:
Thank you for the review @fstein93!

PR #4045, opened 2025-03-03
DBM: prefer function over macro

hfp:
- Complements #4032.


hfp:
As a side-note, the macro may not necessarily be more unsafe, e.g., if imax quietly take-converts one unsigned (no warning generated), the macro in contrast likely generates "comparison of different signs".

oschuett:
Thanks a lot!

PR #4048, opened 2025-03-04
DBM_VALIDATE_AGAINST_LIBXSMM to enable internal validation

hfp:
- Compile-time/opt-in with -DDBM_VALIDATE_AGAINST_LIBXSMM.
- TODO: more sophistication like injecting error margin.


hfp:
This PR may be complemented with adjusted code in the miniapp to have non-trivial test data (TODO). On top, this can be enabled as part of CP2K regression tests. Gold data is produced by LIBXSMM and validated against accelerator results (GPU). Validation adds cost due to additional calculation on the host plus intermittent read-back from the accelerator.

PR #4052, opened 2025-03-04
DBM/miniapp: leverage DBM_VALIDATE_AGAINST_LIBXSMM

hfp:
- Individual matrix elements (values).
- Randomized block sizes.
- Complements #4048.


hfp:
Checking `#if defined(__OFFLOAD)` in the miniapp does not work since only `offload_library.h` is included, i.e., `__OFFLOAD` is not defined if `__OFFLOAD_CUDA`, `__OFFLOAD_HIP`, `__OFFLOAD_OPENCL` is defined. A follow-up PR will not ask for `#if defined(__OFFLOAD)`. Alternatively, `offload_library.h` could define it accordingly.

PR #4056, opened 2025-03-06
Adjust regtests HF-loc-field-gopt.inp and RI_RPA_grad_MP2_H2O_HF_PBE_ADMM.inp

oschuett:
Follow up to #4030, which let to some wrong results:

- https://dashboard.cp2k.org/archive/gcc14/index.html
- https://dashboard.cp2k.org/archive/minimal-sdbg/index.html
- https://dashboard.cp2k.org/archive/i386/index.html

### DIFF
--- a/tests/QS/regtest-ri-rpa-grad/TEST_FILES.toml
+++ b/tests/QS/regtest-ri-rpa-grad/TEST_FILES.toml
@@ -1,5 +1,5 @@
 "RI_RPA_grad_MP2_H2O.inp"               = [{matcher="M011", tol=8e-08, ref=-16.994618476831036}]
 "RI_RPA_grad_MP2_CH3.inp"               = [{matcher="M011", tol=8e-08, ref=-7.515607384605213}]
 "RI_RPA_grad_MP2_H2O_HF_PBE.inp"        = [{matcher="M011", tol=8e-08, ref=-17.057843090414949}]
-"RI_RPA_grad_MP2_H2O_HF_PBE_ADMM.inp"   = [{matcher="M011", tol=8e-08, ref=-17.066232490836303}]
+"RI_RPA_grad_MP2_H2O_HF_PBE_ADMM.inp"   = [{matcher="M011", tol=1e-07, ref=-17.066232490836303}]
 #EOF
--- a/tests/xTB/regtest-2/HF-loc-field-gopt.inp
+++ b/tests/xTB/regtest-2/HF-loc-field-gopt.inp
@@ -6,7 +6,7 @@
 
 &MOTION
   &GEO_OPT
-    MAX_ITER 3
+    MAX_ITER 2
   &END GEO_OPT
 &END MOTION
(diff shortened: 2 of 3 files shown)


hfp:
Beside of removing `-O3` in the mentioned commit, it added `-march=native` (perhaps we should think about checking the latter and make test pass with it?). Do we know which of the change caused it? I mean `-march=native` may enable wider vector instruction set architecture (vector ISA), but `-O3` or not is pretty much "unpredictable" like umbrella flag. For instance mighty tree-vectorize is part of O3 and like-wise other stuff.

mkrack:
Adding the `-O3` flag for the CMake configuration `Release` is useless, because CMake adds that flag by default with `Release`. That is why I removed the flag in PR #4030 to get rid of the redundancy of that flag.
I added in the same PR the `-march=native` flag, because it is needed with GCC for CP2K to recognise the CPUID. Without that flag, CP2K prints always the HINT that it was not specifically compiled for the actual target CPU. The `-march` flag is used by the `Linux-gnu-*` arch files since a long time without problem, but there the `-march` flag is combined with `-O2` enhanced by the flags `-ftree-vectorize -funroll-loops`. Maybe, the combination of `-O3` with `-march` causes a different numerical noise.
I think that `-O2` enhanced with selected flags activated by `-O3` would be better as the default `Release` Fortran compiler flags. In the past, I observed that CP2K binaries compiled with `-O3` were not faster or sometimes even slower than the binaries compiled with "enhanced" `-O2` flags. Well, that might have changed.

hfp:
@mkrack Thank you, Matthias!

Also, only one test gave wrong result not looking too wrong (only checked the GCC14 result). Perhaps the test can be calibrated?

oschuett:
My goal for this PR was just to get the dashboard back to green.

I agree that we should put some more thought into how we want to organize our flags in the future. Furthermore, we should also make sure that we carry over all the hard-won lesson that led to the set of flags that [we're currently using](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/scripts/generate_arch_files.sh).

mkrack:
I get no error, if I run the tester [Minimal arch-file (sdbg)](https://dashboard.cp2k.org/archive/minimal-sdbg/index.html) and [Ubuntu, GCC 14 (ssmp)](https://dashboard.cp2k.org/archive/gcc14/index.html) locally.

oschuett:
> I get no error, if I run the tester Minimal arch-file (sdbg) and Ubuntu, GCC 14 (ssmp) locally.

I managed to reproduce the GCC 14 issue at home, but did not try the other. Perhaps it only occurs on AMD CPUs.

>  Perhaps the test can be calibrated?

Yes, that seems like the better solution. I'll just increase the tolerances on `RI_RPA_grad_MP2_H2O_HF_PBE_ADMM.inp` and `HF-loc-field-gopt.inp` and remove the i386 test entirely.

PR #4057, opened 2025-03-07
[dftd3] Improve error messaging when input parameters are missing

mtaillefumier:
The error message returned by cp2k is unclear when these two combination of parameters are present in a input file.

 ```
REFERENCE_FUNCTIONAL
 PARAMETER_FILE_NAME dftd3.dat
```

The error message is also misleading when  "PARAMETER_FILE_NAME" is empty causing confusion. The default value should be dftd3.dat or the error message should be more explicit.

### DIFF
--- a/src/qs_dispersion_utils.F
+++ b/src/qs_dispersion_utils.F
@@ -93,7 +93,10 @@ CONTAINS
             IF (.NOT. explicit) THEN
                CALL section_vals_val_get(pp_section, "REFERENCE_FUNCTIONAL", explicit=exfun)
                IF (.NOT. exfun) THEN
-                  CPABORT("D2 vdW REFERENCE_FUNCTIONAL expected (example REFERENCE_FUNCTIONAL PBE). Go to https://www.chemie.uni-bonn.de/grimme/de/software/dft-d3/ for a full list of supported functionals")
+                  CALL cp_abort(__LOCATION__, "D2 vdW REFERENCE_FUNCTIONAL expected a second parameter "// &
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mtaillefumier** [comment]
Any input from dev with knowledge in DFTD3 is welcome because the code is not particularly clear if the two parameters are needed or not. The documentation is not clear either.

File: src/qs_dispersion_utils.F [hunk 1]
```diff
@@ -92,7 +92,9 @@ SUBROUTINE qs_dispersion_env_set(dispersion_env, xc_section)
             CALL section_vals_val_get(pp_section, "SCALING", explicit=explicit)
             IF (.NOT. explicit) THEN
                CALL section_vals_val_get(pp_section, "REFERENCE_FUNCTIONAL", explicit=exfun)
-               CPASSERT(exfun)
+               IF (.NOT. exfun) THEN
+                  CPABORT("D2 vdW REFERENCE_FUNCTIONAL expected (example REFERENCE_FUNCTIONAL PBE). Go to https://www.chemie.uni-bonn.de/grimme/de/software/dft-d3/ for a full list of supported functionals")
```
oschuett: For aborts with long messages you can use the following:
```
CALL cp_abort(__LOCATION__, "A long error message "//&
              "with lots of valuable advice.")  
```

See also https://www.cp2k.org/dev:error_handling

mtaillefumier: OK Thnk you for the information. I would like to make the error message more explicit because it took me some time to figure out what was the error.

PR #4058, opened 2025-03-07
Add openPMD output

franzpoeschel:
This is an effort to add openPMD as a method for creating simulation output (and at a later point also reading input) from CP2K. In here, PMD is short for particle-mesh data.

For reference:

* https://openpmd-api.readthedocs.io/en/0.16.1/
* https://github.com/openPMD/
* https://www.openpmd.org/

Roughly, this will be an alternative to writing to (and at a later point reading from) .cube files.
Being a data schema rather than a file format, openPMD is implementable in various physical file formats. The openPMD-api, used here, implements the standard in terms of [HDF5](https://www.hdfgroup.org/solutions/hdf5/) (binary files), [ADIOS2](https://adios2.readthedocs.io) (binary files, staging methods), JSON and TOML (text files).

Tasks:

- [x] Add an `openPMD.F` module that binds to the (C++) openPMD-api via C. Support for this is only added via CMake and can be activated through specification of `-DCP2K_USE_OPENPMD=ON` at configuration. The bindings are compiled to a separate static library in order to simplify potential future extraction into C bindings.
- [x] Add necessary structure for file creation and handling inside `cp_output_handling.F`. A particular challenge has been that openPMD files contain internal structure and will, if used idiomatically, correspond with multiple outputs as written by CP2K in .cube format; and that additionally the openPMD-api supports streaming IO methods next to file writing. In consequence, openPMD output files need to be kept open internally during multiple cycles of dataset writing performed by CP2K, meaning that multiple previously created .cube files need to be resolved to a single openPMD output Series.
    The functions in the `cp_output_handling` module expose handles to callers mostly in terms of `unit_nr`s. Since this project aims to mostly keep the interfaces of existing functionality unchanged, this means that "virtual unit numbers" are handed out that are resolved internally to an openPMD file and information on the associated datasets therein. Additionally, openPMD output should be distinguished by callsite in order to have physically distinct outputs for logically distinct callsites.
    In summary, data members inside the cp_output_handling modules are maintained in terms of (1) a map `unit_nr → associated data` and (2) a map `callsite identifier → associated data`
- [x] Add actual writing of n-dimensional array data in `realspace_grid_cube.F`
- [x] Parallel output writing (this is ideally trivial since produced data is composed into a single output by the backends)
- [ ] Physical metadata definitions (self-describing datasets via attributes)
- [x] Conditional compilation (exclude openPMD functionality when openPMD is not linked)
- [x] Runtime configuration from input file (currently hardcoded to a single test case)
- [x] Testing, CI integration, documentation
- [x] Restrict least supported openPMD version to current dev (use of `json::merge()`)

Tasks for a follow-up PR (not required for basic functionality):

* Tooling: Conversion functionality .cube <-> openPMD
* Add openPMD output to modules other than `qs_scf_post_gpw.F` (currently used for testing)
* Input reading


oschuett:
Thanks for opening this PR. A few comments:

I'm wondering why you didn't add the bindings for C and Fortran to the openPMD library itself? There the bindings could also be used by other codes. Fortran is, after all, quite prevailing in the chemistry community.

I saw that you are piggybacking on the existing [MO_CUBES](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/PRINT/MO_CUBES.html) printkey. I think, it would be better to leave the existing code paths untouched and instead add a new printkey named e.g. `MO_OPENPMD` .

Finally, I'm really worried about the tight coupling between the prinkey code and openPMD. In my view openPMD is just another file format (no offense) and should therefore not receive any special treatment. I do realize that you need a way to know when output should be written without actually opening a file in Fortran. For this you could use the following:

```
USE cp_output_handling,                        ONLY: cp_print_key_should_output, cp_p_file
...
should_output = BTEST(cp_print_key_should_output(logger%iter_info, basis_section, print_key_path), cp_p_file)
```

franzpoeschel:
Thank you for your feedback!

> I'm wondering why you didn't add the bindings for C and Fortran to the openPMD library itself? There the bindings could also be used by other codes. Fortran is, after all, quite prevailing in the chemistry community.

This can be the far goal. I'll argue that starting this development inside an actual end-user code is the best way to get there. During the time that the bindings are CP2K-internal, those bindings have one user, ourselves. Integrating the bindings into the library code means that we have to wait for a release to use them, have API stability guarantees to hold from then on, and that we cannot quickly add new features once we notice that we miss something.
Additionally, for a code as long-living as CP2K is, defining its own internal interface to external libraries is even desirable for maintainability by minimizing the inclusion of external library code "outside your control" (and potentially also for use cases such as logging).

Fyi, there is a [PR for C bindings](https://github.com/openPMD/openPMD-api/pull/1537) in the openPMD-api as a third-party contribution which I plan on completing in the upcoming time; the `openPMD.h` and `openPMD.cpp` files in CP2K can then be easily swapped with that. Until then, those files are small and straight-forward enough.

> I saw that you are piggybacking on the existing [MO_CUBES](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/PRINT/MO_CUBES.html) printkey. I think, it would be better to leave the existing code paths untouched and instead add a new printkey named e.g. MO_OPENPMD .

> Finally, I'm really worried about the tight coupling between the prinkey code and openPMD. In my view openPMD is just another file format (no offense) and should therefore not receive any special treatment. I do realize that you need a way to know when output should be written without actually opening a file in Fortran. For this you could use the following:

This is the reason why I opened this PR early, I'm new to the internal structure of CP2K, so this is a good time to check if I'm going the right direction.
I'm not fully sure I understand the alternative that you suggest. I'll try to summarize how I understood what you wrote, maybe you can point out where I got it wrong:

1. Revert all changes made in `cp_print_key_unit_nr`, that function is for creating files from Fortran and should stay like that.
2. Move the logic added into `cp_print_key_unit_nr` to a new function, e.g. `cp_print_openpmd_key_unit_nr` as part of the same module.
3. At the callsite, call *either* `cp_print_key_unit_nr` *or* `cp_print_openpmd_key_unit_nr`.
4. I don't understand where to put the `should_output = BTEST(cp_print_key_should_output(logger%iter_info, basis_section, print_key_path), cp_p_file)` line. This already exists inside `cp_print_key_unit_nr`. But if `should_output` is true, then the call to `cp_print_key_unit_nr` is a no-op, so I don't see why I would even still call that function.
5. You suggest adding a new printkey MO_OPENPMD. I guess that this is about the input section? I did not change anything about the input section yet simply because I just did not get there yet; so yes, this should be done properly at some point.

oschuett:
> Until then, those files are small and straight-forward enough.

I'm fine with keeping the C and Fortran bindings within CP2K for now. However, then I'd ask you to condense them into just two files:  `src/openpmd_c_api.cpp` and `src/openpmd_api.F`.

> 1. Revert all changes made in cp_print_key_unit_nr, that function is for creating files from Fortran and should stay like that.

Yes, please don't modify any of the existing IO code.

> 2. Move the logic added into cp_print_key_unit_nr to a new function, e.g. cp_print_openpmd_key_unit_nr as part of the same module.

Yes, move your logic into a separate function, e.g. `openpmd_printkey_write`, but please put that function into a new file, e.g. `openpmd_printkey.F`.

> 3. At the callsite, call either cp_print_key_unit_nr or cp_print_openpmd_key_unit_nr.

Please leave the existing calls to `cp_print_key_unit_nr` unchanged. Instead you can add a seperate call to your new `openpmd_printkey_write` function.

> 4. I don't understand where to put the should_output = BTEST(cp_print_key_should_output(...), cp_p_file) line. ...

I was thinking that you could call `cp_print_key_should_output` from your new `openpmd_printkey_write` function.

> 5. You suggest adding a new printkey MO_OPENPMD. I guess that this is about the input section?

Yes, [printkeys](https://www.cp2k.org/dev:printkey) are essentially input sections with a standardized semantic. 

The `MO_CUBES` section is created here:

https://github.com/cp2k/cp2k/blob/6843abca0f77ce480c668f110010bf46211d2f12/src/input_cp2k_properties_dft.F#L1343

I'd suggest you create your own `MO_OPENPMD` section next to it.

oschuett:
It's great to see you're at this again. In CP2K we're keeping our git history linear. So, it would be good if you could remove any merge commits. That would also allow the CI to start running.

franzpoeschel:
Hello Ole, I am planning to completely clean up the commit history at some point anyway. By now, this PR mostly adds new files and modifies some few others, so this cleanup should be relatively straightforward. For the moment, merging was easier due to the many little commits and the ongoing changes in the CMake scripts that caused conflicts.

franzpoeschel:
The code is now mostly ready for review.

The "user interface" is relatively simple: For sections `DFT%PRINT%MO_CUBES`, `DFT%PRINT%ELF_CUBE`, `DFT%PRINT%E_DENSITY_CUBE`, there now exist corresponding alternative sections `DFT%PRINT%MO_OPENPMD`, `DFT%PRINT%ELF_OPENPMD` and `DFT%PRINT%E_DENSITY_OPENPMD`. The configuration structures are roughly the same, except:

- Cube-specific keys are missing: APPEND
- openPMD-specific keys are added: OPENPMD_EXTENSION, OPENPMD_CFG, OPENPMD_CFG_FILE. (I should maybe add OPENPMD_ACCESS as an alternative to the removed APPEND key.)

For each output method, Cube and openPMD may not be activated at the same time (adding this option would not be too complicated, but I don't see the need right now).
More output methods may receive an openPMD alternative, it's just these three right now because we needed a place to start. Any output that currently writes to Cube via the triple of `cp_print_key_unit_nr`, `cp_pw_to_cube` and `cp_print_key_finished_output` should be straightforward to transform to choosing between openPMD and Cube for output.

The code is now structured such that most openPMD-specific functionality is part of new files. Only file `src/pw/qs_scf_post_gpw.F` contains a larger amount of changes to existing code: It contains those routines that actually call the Cube output routines. I changed those routines such that they now interact with a newly-introduced type `cp_section_key` that forwards calls to either Cube or openPMD.

I just ran the regtests, they finish with the same results as the base commit of this branch, i.e. with two errors:

```
<<< /home/franzpoeschel/git-repos/cp2k/build/bin/TEST-2025-09-25_15-09-47/QS/regtest-harris (215 of 215) done in 9.88 sec
------------------------------- Errors ---------------------------------
/home/franzpoeschel/git-repos/cp2k/build/bin/TEST-2025-09-25_15-09-47/QS/regtest-gpw-4/H2O_wavelet_XZ.inp.out
Spec: {'matcher': 'E_total', 'tol': 3e-14, 'ref': -16.5870513472298}
Difference too large: 4.39e-04 > 3e-14, value: -16.59433276616352.

/home/franzpoeschel/git-repos/cp2k/build/bin/TEST-2025-09-25_15-09-47/xTB/regtest-2/HF-loc-field-gopt.inp.out
Spec: {'matcher': 'E_total', 'tol': 2e-08, 'ref': -5.65791503021294}
Difference too large: 1.04e-07 > 2e-08, value: -5.65791561585296.


------------------------------- Timings --------------------------------
Plot: name="timings", title="Timing Distribution", ylabel="time [s]"
PlotPoint: name="100th_percentile", plot="timings", label="100th %ile", y=39.53, yerr=0.0
PlotPoint: name="99th_percentile", plot="timings", label="99th %ile", y=11.19, yerr=0.0
PlotPoint: name="98th_percentile", plot="timings", label="98th %ile", y=8.72, yerr=0.0
PlotPoint: name="95th_percentile", plot="timings", label="95th %ile", y=6.59, yerr=0.0
PlotPoint: name="90th_percentile", plot="timings", label="90th %ile", y=4.76, yerr=0.0
PlotPoint: name="80th_percentile", plot="timings", label="80th %ile", y=3.27, yerr=0.0

------------------------------- Summary --------------------------------
Number of FAILED  tests 0
Number of WRONG   tests 2
Number of CORRECT tests 3010
Total number of   tests 3012

Summary: correct: 3010 / 3012; wrong: 2; 20min
```

I guess that these two errors stem from somewhere else?

Still missing are:

- Test cases (relatively easily generated by replacing Cube output in existing input files with openPMD output)
- Integration with CI (CI currently failing because it sees the openPMD Fortran files, but does not actually build them)
- Some physics metadata in the output concerning physical units (SI unit, scaling factor); can be added later, not really relevant for code review.
- Documentation

I think that the documentation is generated automatically from the definition of the input section structure? `docs/methods/dft` mentions "Cube" only in passing. If there is a specific place to document this, please point me there. I also added a new environment variable `CP2K_USE_THREADED_MPI` (needed for some IO methods) which should be documented.
For the CI integration, I might also need some pointers on where to add things.

If you want to actually run things, you can follow the [installation instructions](https://openpmd-api.readthedocs.io/en/0.16.1/install/install.html) for the openPMD-api (there are some [CP2K-specific performance optimizations](https://github.com/openPMD/openPMD-api/pull/1741) not yet part of that release) and then use `-DCP2K_USE_OPENPMD=ON` in CMake for building. openPMD is not part of the toolchain.

Pinging @oschuett

src/f77_interface.F
```diff
@@ -18,134 +18,137 @@
...
+                                cp_dlaf_free_all_grids
+   USE cp_error_handling, ONLY: cp_error_handling_setup
+   USE cp_files, ONLY: init_preconnection_list, &
+                       open_file
+   USE cp_log_handling, ONLY: &
+      cp_add_default_logger, cp_default_logger_stack_size, cp_failure_level, &
+      cp_get_default_logger, cp_logger_create, cp_logger_get_default_unit_nr, cp_logger_release, &
+      cp_logger_retain, cp_logger_type, cp_rm_default_logger, cp_to_string
+   USE cp_output_handling, ONLY: cp_iterate
+#ifdef __OPENPMD
```
oschuett: High-level files like `f77_interface.F` should not contain any `#ifdef`. Please move them into `cp_output_handling_openpmd.F` or below.

src/qs_scf_post_gpw.F
```diff
@@ -13,191 +13,197 @@
...
+   USE s_square_methods, ONLY: compute_s_square
+   USE scf_control_types, ONLY: scf_control_type
+   USE stm_images, ONLY: th_stm_image
+   USE transport, ONLY: qs_scf_post_transport
+   USE trexio_utils, ONLY: write_trexio
+   USE virial_types, ONLY: virial_type
+   USE voronoi_interface, ONLY: entry_voronoi_or_bqb
+   USE xray_diffraction, ONLY: calculate_rhotot_elec_gspace, &
+                               xray_diffraction_spectrum
+#ifdef __OPENPMD
```
oschuett: Please push this `#ifdef` also down into the low-level files.

src/input_cp2k_print_dft.F
```diff
@@ -163,6 +163,11 @@ MODULE input_cp2k_print_dft
    USE qs_mom_types, ONLY: create_mom_section
    USE string_utilities, ONLY: newline, &
                                s2a
+
+#ifdef __OPENPMD
```
oschuett: Please remove the `#ifdef` also from this file and just always create the input keywords.

This allows you to print a nice error message in case openPMD is not available - opposed to a confusing input parser error.

See e.g. [cp_fm_dlaf_api.F](https://github.com/cp2k/cp2k/blob/b396de2029893bdd02783a52e14ca541ec8cb85b/src/fm/cp_fm_dlaf_api.F#L73).

src/mpiwrap/message_passing.F
```diff
@@ -1146,18 +1148,41 @@ SUBROUTINE mp_world_init(mp_comm)
...
+!$          required_threading_level = MPI_THREAD_MULTIPLE
 #else
-!$          CALL mpi_init_thread(MPI_THREAD_SERIALIZED, provided_tsl, ierr)
-#endif
+!$          required_threading_level = MPI_THREAD_SERIALIZED
+#endif
+            ! Allow users to raise the MPI threading level if needed
+            ! Introduced because some (but not all) IO routines in openPMD require threaded MPI
+            ! (AsyncWrite in ADIOS2/BP5, MPI data plane in ADIOS2/SST, subfiling in HDF5)
+!$          CALL get_environment_variable( &
```
oschuett: I really don't like CP2K's behavior depending on environment variables.

Couldn't we just make `MPI_THREAD_MULTIPLE` a hard requirement for openPMD:

```
#if defined(__DLAF) || defined(__OPENPMD)
!$          CALL mpi_init_thread(MPI_THREAD_MULTIPLE, provided_tsl, ierr)
#else
...
```

oschuett (commented):
Generally looks good. Thanks for making the requested changes.

> I think that the documentation is generated automatically from the definition of the input section structure?

Yes, once the code compiles we can have the CI generate a preview manual for this PR.

> Test cases (relatively easily generated by replacing Cube output in existing input files with openPMD output)

Yes, since this adds a lot of new code, it would definitely be good to have tests. The first step towards that would be to add openPMD to our Spack package.

src/f77_interface.F
franzpoeschel: I have moved the #ifdef guards into the openPMD-specific source files now (which are now all unconditionally included into the build)

src/mpiwrap/message_passing.F
franzpoeschel: Done. Theoretically, an additional CMake variable could be introduced to control this instead; but I do not know if the performance penalty by threaded MPI is still as large as it used to be. This now enables threaded MPI for use cases where it is not needed.

franzpoeschel:
After rebasing the branch today, the compiler is giving me various errors from unrelated source files that look to me like a wrong Fortran standard being selected by CMake (e.g. `Error: Line truncated at (1) [-Werror=line-truncation]`). This happens for me locally as well as [in the CI](https://storage.googleapis.com/cp2k-ci/run-cp2k-sdbg-07131c8e_report.txt).

oschuett:
> Error: Line truncated at (1) [-Werror=line-truncation]

Yes, I recently re-enable the check for overly long lines that violate the Fortran standard (#4468).

You can just break up long strings with the concatenation operator `//` like this:

https://github.com/cp2k/cp2k/blob/ad3ce71a450b1cd81fd944488742f4378e7c5979/src/input_cp2k_dft.F#L213-L236

franzpoeschel:
> > Error: Line truncated at (1) [-Werror=line-truncation]
> 
> Yes, I recently re-enable the check for overly long lines that violate the Fortran standard (#4468).
> 
> You can just break up long strings with the concatenation operator `//` like this:
> 
> https://github.com/cp2k/cp2k/blob/ad3ce71a450b1cd81fd944488742f4378e7c5979/src/input_cp2k_dft.F#L213-L236

Done now. I was a bit surprised because for my local builds to succeed, I also need to fix a handful of lines unrelated with this PR, check [this commit](https://github.com/cp2k/cp2k/pull/4058/commits/762ca17092704f27754d321368c7406e2815a024). The compiler is gfortran 13.3.0.

oschuett:
> I also need to fix a handful of lines unrelated with this PR, check [this commit](https://github.com/cp2k/cp2k/pull/4058/commits/762ca17092704f27754d321368c7406e2815a024).

Yeah, this is a [known issue](https://github.com/cp2k/cp2k/blob/ad3ce71a450b1cd81fd944488742f4378e7c5979/src/CMakeLists.txt#L1646) due to our ongoing migration from GNU Make to CMake. Your workaround is fine. Alternatively you could call  `cp_abort`. Sorry for all the hassle. 

See also https://www.cp2k.org/dev:error_handling.

franzpoeschel:
> Yes, since this adds a lot of new code, it would definitely be good to have tests. The first step towards that would be to add openPMD to our Spack package.

openPMD is now added as optional dependency in the [Spack package](https://github.com/spack/spack-packages/pull/1996). I have uploaded a simple input file that activates a handful of openPMD output routines. How would I go forward with tying this together in the CI?

oschuett:
Congrats for getting openPMD into Spack!

Now there are just a few more steps needed to get this working in our CI:

- Change the [SPACK_PACKAGES_VERSION](https://github.com/cp2k/cp2k/blob/26c18786f45870d44f6ac26397e30ffc4d931660/tools/docker/generate_dockerfiles.py#L728) to `2025.11.0`.
- Run the `generate_dockerfiles.py` script you just edited.
- Add openPMD to [cp2k_deps_psmp.yaml](https://github.com/cp2k/cp2k/blob/master/tools/spack/cp2k_deps_psmp.yaml).
- Add `-DCP2K_USE_OPENPMD=ON` to [this section](https://github.com/cp2k/cp2k/blob/26c18786f45870d44f6ac26397e30ffc4d931660/cmake/cmake_cp2k.sh#L39-L52).

Then we should be finally able to run your new tests.

If you want to run the container locally first, I'd highly recommend a Spack cache. See [here](https://manual.cp2k.org/trunk/getting-started/build-with-spack.html#spack-and-cp2k-docker-containers) for details.

cmake/cmake_cp2k.sh
```diff
@@ -47,6 +47,7 @@ if [[ "${PROFILE}" == "spack" ]] && [[ "${VERSION}" == "psmp" ]]; then
     -DCP2K_USE_DLAF=OFF \
     -DCP2K_USE_LIBXSMM=OFF \
     -DCP2K_USE_TBLITE=OFF \
+    -DCP2K_USE_OPENPMD_ON \
```
oschuett: Typo: `CP2K_USE_OPENPMD_ON` -> `CP2K_USE_OPENPMD=ON`

franzpoeschel: Ah, thanks

franzpoeschel:
Alright, openPMD is now part of the Spack CI run, and the example input file runs successfully in it. Two notes:

* I had to activate a dummy output engine, otherwise (even for very small outputs only) the CI would complain about "HUGE_OUTPUT". I guess there might be a setting somewhere for specifying expected output size?
* For getting the input file into the spack run, I added the input file to `tests/QS/regtest-ot/TEST_FILES.toml`. Of course, this breaks CI runs such as sdbg. There is probably an option to add the input file more specifically to just some selected CI runs?
* openPMD currently is part only of the psmp CI run, but there is fundamentally no problem with adding it to ssmp, too. Should I?

src/cp2k_info.F
```diff
@@ -295,6 +295,10 @@ FUNCTION cp2k_flags() RESULT(flags)
       flags = TRIM(flags)//" greenx"
 #endif
 
+#if defined(__OPENPMD) && defined(OPENPMDAPI_VERSION_GE) && openPMD_HAVE_MPI
```
oschuett: This condition should match the one used elsewhere in the code:
```suggestion
#if defined(__OPENPMD)
```

oschuett (commented):
> Alright, openPMD is now part of the Spack CI run,...

Woohoo!

> For getting the input file into the spack run, I added the input file to tests/QS/regtest-ot/TEST_FILES.toml.

Please create a new directory, e.g. `tests/QS/regtest-openpmd`, and add it to [TEST_DIRS](https://github.com/cp2k/cp2k/blob/master/tests/TEST_DIRS):

```
QS/regtest-openpmd                     openpmd
```

>    I had to activate a dummy output engine, otherwise (even for very small outputs only) the CI would complain about "HUGE_OUTPUT".


Yes, we have a default limit of 2 MiB. If your output is still reasonable, say <100MiB, then you can add your tests to [HUGE_TESTS_SUPPRESSIONS](https://github.com/cp2k/cp2k/blob/master/tests/HUGE_TESTS_SUPPRESSIONS).

> openPMD currently is part only of the psmp CI run, but there is fundamentally no problem with adding it to ssmp, too. Should I?

The second column in the `TEST_DIRS` is actually a selector for the flags that are printed by `cp2k_info.F`. So, as we're migrating more of our CI testers to Spack the `regtest-openpmd` directory will be automatically picked up.

src/cp2k_info.F
franzpoeschel: That is how I had it originally, but the CI wants all preprocessor flags that are used anywhere to be present also in `src/cp2k_info.F` and `INSTALL.md`, and this was the most direct way.

src/cp2k_info.F
```diff
@@ -274,6 +274,15 @@ FUNCTION cp2k_flags() RESULT(flags)
       flags = TRIM(flags)//" greenx"
 #endif
 
+! Preprocessor macros used in openPMD.cpp:
```
oschuett: Sorry, I forgot about this. We usually add exception [here](https://github.com/cp2k/cp2k/blob/f0e2e694aacf72b5fbb4a35afc4edb64dda8b7ec/tools/precommit/check_file_properties.py#L20), but your solution also gets the job done ;-)

franzpoeschel: Thank you for the hint. It turned out that my solution did not really get the job done, because those two preprocessor macros were of course not included into the Fortran code, so I have added these two exceptions now.

oschuett:
Sorry for producing all those merge conflicts.
Better hold off until #4619 is in and then add your bit to the new `docs/technologies/libraries.md` page.

franzpoeschel:
> Better hold off until #4619 is in and then add your bit to the new `docs/technologies/libraries.md` page.

Sure!
That will probably be next year then, as I am off work after today.

oschuett:
Ok #4619 is merged. 

> That will probably be next year then, as I am off work after today.

No problem, but you will probably miss the upcoming release.

Btw, where did you end up hiding those environment variables for the test? I can't find them anymore.

franzpoeschel:
> No problem, but you will probably miss the upcoming release.

It is probably better anyway to add a large change at the beginning, not the end of a release cycle?
> 
> Btw, where did you end up hiding those environment variables for the test? I can't find them anymore.

Which do you mean? The test does not require environment variables for execution.

oschuett:
> It is probably better anyway to add a large change at the beginning, not the end of a release cycle?

Up to you. Many of our contributors are actually eager to [catch the release train](https://en.wikipedia.org/wiki/Release_early,_release_often), but playing it safe certainly has its merits. 


> Which do you mean? The test does not require environment variables for execution.

Sorry, I briefly confused this PR with the one for MiMiC.

Happy holidays 🎄

franzpoeschel:
> Happy holidays 🎄

Happy new year, hope you had nice holidays, too ;)

I have rebased the PR now

PR #4063, opened 2025-03-10
Remove CMake build presets

RMeli:
I'm starting to look at ways to simplify, cleanup, and consolidate the CMake build system. This PR removes the `CP2K_BUILD_OPTIONS` presets because they are not particularly useful and rather arbitrary. It's better to be explicit about the different dependencies one wants to build CP2K with.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,7 +97,7 @@ option(CMAKE_POSITION_INDEPENDENT_CODE "Enable position independent code" ON)
 option(CP2K_ENABLE_CONSISTENCY_CHECKS
        "Check that the list of compiled files and files contained in src match"
        OFF)
-option(CP2K_USE_DFTD4 "Enable dispersion coorections DFTD4" ON)
+option(CP2K_USE_DFTD4 "Enable dispersion corrections with DFTD4" ON)
 option(CP2K_USE_DEEPMD "Enable DeePMD" ON)
 option(CP2K_USE_FFTW3 "Use fftw3 for the calculating fast fourier transforms"
        ON)
@@ -117,6 +117,7 @@ option(
   "MKL has its own compatible implementation of the fftw library. This option, when ON, will use the separate and original fftw library."
   OFF)
 
+# MPI-enabled options
 cmake_dependent_option(CP2K_USE_COSMA "Enable COSMA support" ON "CP2K_USE_MPI"
                        OFF)
 cmake_dependent_option(CP2K_USE_DLAF "Enable DLA-Future support" ON
@@ -125,8 +126,6 @@ cmake_dependent_option(CP2K_USE_ELPA "Enable ELPA support" ON "CP2K_USE_MPI"
                        OFF)
 cmake_dependent_option(CP2K_USE_LIBSMEAGOL "Enable libSMEAGOL support" ON
                        "CP2K_USE_MPI" OFF)
-cmake_dependent_option(CP2K_USE_LIBXSMM "Enable libxsmm support" ON
-                       "NOT CP2K_USE_ACCEL MATCHES \"OPENCL\"" ON)
 cmake_dependent_option(CP2K_USE_MPI_F08 "Enable MPI Fortran 2008 interface" ON
                        "CP2K_USE_MPI" OFF)
 cmake_dependent_option(CP2K_USE_PLUMED "Enable PLUMED2 support" ON
@@ -136,6 +135,8 @@ cmake_dependent_option(CP2K_USE_SIRIUS "Enable SIRIUS support" ON
 cmake_dependent_option(CP2K_USE_SPLA "Enable SPLA support" ON "CP2K_USE_MPI"
                        OFF)
 
+cmake_dependent_option(CP2K_USE_LIBXSMM "Enable libxsmm support" ON
+                       "NOT CP2K_USE_ACCEL MATCHES \"OPENCL\"" ON)
 cmake_dependent_option(CP2K_USE_LIBVDWXC "Enable libvdwxc support with SIRIUS"
                        ON "CP2K_USE_SIRIUS" OFF)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -191,76 +191,9 @@ if(NOT ${CP2K_SCALAPACK_VENDOR} IN_LIST CP2K_SCALAPACK_VENDOR_LIST)
   message(FATAL_ERROR "Invalid scalapack vendor backend")
 endif()
 
-set(CP2K_BUILD_OPTIONS_LIST "CUSTOM" "DEFAULT" "MINIMAL" "FULL" "SERIAL")
-
-set(CP2K_BUILD_OPTIONS
-    "CUSTOM"
-    CACHE
-      STRING
-      "Build cp2k with a predefined set of dependencies. The default setting is full user control"
-)
-set_property(CACHE CP2K_BUILD_OPTIONS PROPERTY STRINGS
-                                               ${CP2K_BUILD_OPTIONS_LIST})
-
 set(CP2K_DATA_DIR
     "default"
     CACHE STRING "Set the location for cp2k data")
-
-string(TOUPPER ${CP2K_BUILD_OPTIONS} cp2k_build_options_up)
-
-if(NOT ${cp2k_build_options_up} IN_LIST CP2K_BUILD_OPTIONS_LIST)
-  message(FATAL_ERROR "Invalid value for cp2k build options")
-endif()
-
-if(${cp2k_build_options_up} STREQUAL "MINIMAL")
-  set(CP2K_USE_FFTW3 ON)
-  set(CP2K_USE_MPI ON)
-endif()
-
-if(${cp2k_build_options_up} STREQUAL "FULL")
-  set(CP2K_USE_FFTW3 ON)
-  set(CP2K_USE_MPI ON)
-  set(CP2K_USE_DFTD4 ON)
-  set(CP2K_USE_SIRIUS ON)
-  set(CP2K_USE_LIBXSMM ON)
-  set(CP2K_USE_VORI ON)
-  set(CP2K_USE_COSMA ON)
-  set(CP2K_USE_SPLA ON)
-  set(CP2K_USE_LIBXC ON)
-  set(CP2K_USE_LIBINT2 ON)
-  set(CP2K_USE_ELPA ON)
-  set(CP2K_USE_SPGLIB ON)
-  set(CP2K_USE_LIBTORCH ON)
-  set(CP2K_USE_HDF5 ON)
-  set(CP2K_USE_LIBGRPP ON)
-  set(CP2K_USE_TREXIO ON)
-endif()
-
-if(${cp2k_build_options_up} STREQUAL "SERIAL")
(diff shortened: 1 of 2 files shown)


oschuett:
I agree that the current presets are not very useful. However, I'm also not happy about where [cmake_cp2k.sh](https://github.com/cp2k/cp2k/blob/master/cmake/cmake_cp2k.sh) is heading.

So, I was thinking that we could maybe make the presets overwrite-able. Then one could select e.g. `FULL` and supply only a few additional `-DCP2K_foo_bar=OFF` flags on top of that.

Furthermore, we could use `cmake_dependent_option` more widely. Then e.g. `-DCP2K_USE_MPI=OFF` would automatically disable all libraries that require MPI.

What do you think?

RMeli:
> So, I was thinking that we could maybe make the presets overwrite-able. Then one could select e.g. FULL and supply only a few additional -DCP2K_foo_bar=OFF flags on top of that.

I haven't touched the current defaults, but if we can stick with a single default configuration we could tweak the default to get what you want. Would that be enough? If yes, which default setting would you suggest?

> Furthermore, we could use cmake_dependent_option more widely. Then e.g. -DCP2K_USE_MPI=OFF would automatically disable all libraries that require MPI.

Yes, this is something on my TODO list. We can definitely use `cmake_dependent_option` more aggressively.

oschuett:
> if we can stick with a single default configuration we could tweak the default to get what you want.

Interesting idea! How about we enable everything except for GPUs by default?

RMeli:
cscs-ci run

mtaillefumier (approved):
All these changes look alright. Using `cmake_dependent_option` for `MPI` related packages is a good suggestion. I should have done that long time ago. Thanks for fixing this. 

Unless something breaks I think we can merge this PR

mtaillefumier:
Other suggestions:
- GPU targets:. We should probably use the cmake convention and use `sm_70`, etc... instead off `V100`, `A100` or have a mechanism that explicitly avoid the checks when -DCMAKE_CUDA_ARCHITECTURE is provided.
- language standards: The `if` statement is the best way in my view to handle standards. It allows us to define them during configuration stage or use a default value if not explicitly given. We should make sure that we do not give `-std=cxx` explicitly and if some mini-app requires a different standard use target properties instead. I am almost certain we already do this but a glitch can always happens.

```
if(NOT DEFINED CMAKE_CUDA_STANDARD)
  set(CMAKE_CUDA_STANDARD 11)
  set(CMAKE_CUDA_STANDARD_REQUIRED ON)
endif()

# set language and standard
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_C_STANDARD 11)
set(CMAKE_HIP_STANDARD 14)
set(CMAKE_CUDA_STANDARD 14)
```
All these points are suggestions only though.

RMeli:
> GPU targets:. We should probably use the cmake convention and use sm_70, etc... instead off V100, A100 or have a mechanism that explicitly avoid the checks when -DCMAKE_CUDA_ARCHITECTURE is provided.

Yes, I was thinking the same and wanted to discuss with you. I'll try to make small and self-contained PRs, so this one doesn't touch CMake GPU logic.

Very good point about the standard. I don't think we use `-std=...` explicitly anywhere in CMake, but I will double check.

CMakeLists.txt
```diff
@@ -97,41 +97,46 @@ option(CMAKE_POSITION_INDEPENDENT_CODE "Enable position independent code" ON)
 option(CP2K_ENABLE_CONSISTENCY_CHECKS
        "Check that the list of compiled files and files contained in src match"
        OFF)
-option(CP2K_USE_DFTD4 "Enable dispersion coorections DFTD4" OFF)
-option(CP2K_USE_DEEPMD "Enable DeePMD" OFF)
-option(CP2K_USE_SIRIUS "Enable plane wave dft calculations with sirius" OFF)
+option(CP2K_USE_DFTD4 "Enable dispersion coorections DFTD4" ON)
```
oschuett: typo: coorections -> corrections

CMakeLists.txt
```diff
@@ -97,41 +97,46 @@ option(CMAKE_POSITION_INDEPENDENT_CODE "Enable position independent code" ON)
...
-cmake_dependent_option(CP2K_USE_LIBXSMM "Use libxsmm for small gemms" OFF
+cmake_dependent_option(CP2K_USE_COSMA "Enable COSMA support" ON "CP2K_USE_MPI"
+                       OFF)
+cmake_dependent_option(CP2K_USE_DLAF "Enable DLA-Future support" ON
+                       "CP2K_USE_MPI" OFF)
+cmake_dependent_option(CP2K_USE_ELPA "Enable ELPA support" ON "CP2K_USE_MPI"
+                       OFF)
+cmake_dependent_option(CP2K_USE_LIBSMEAGOL "Enable libSMEAGOL support" ON
+                       "CP2K_USE_MPI" OFF)
+cmake_dependent_option(CP2K_USE_LIBXSMM "Enable libxsmm support" ON
```
oschuett: Maybe move out of this code block as it does not depend on `CP2K_USE_MPI`.

oschuett (commented):
I love it! The `cmake_cp2k.sh` is now quite reasonable.

hfp:
Regarding `cmake_cp2k.sh`, I introduced for instance `CP2K_SCALAPACK_VENDOR=auto` (similar to `CP2K_BLAS_VENDOR=auto`). I was planning to leverage this in `cmake_cp2k.sh` as it currently hard-codes some settings "knowing" what the mighty toolchain prepared upfront.

oschuett:
Those build errors are unrelated to this PR: https://dashboard.cp2k.org/archive/current-ssmp/index.html

I don't think it was any of our other changes either. I'm not yet sure what's going. Usually, our docker builds are quite reproducible.

mtaillefumier:
What this dependency updated recently?

RMeli:
Not sure either, but I think `dftd4` uses CMake's `fetch_content` to fetch the `multicharge` dependency, therefore that might be where the issue stems from.

RMeli:
@oschuett is the full output of the toolchain installation available somehow?

RMeli:
Actually, I think I might have been the one breaking things in https://github.com/grimme-lab/multicharge/pull/44, but it is odd that `dftd4` is pulling the `main` branch of `multicharge`.

oschuett:
> @oschuett is the full output of the toolchain installation available somehow?

You can click "Restart w/o Cache", which re-runs the test without populating the docker build cache. Then you get to watch the entire toolchain build.

RMeli:
> Then you get to watch the entire toolchain build.

Usually the toolchain build doesn't show what is happening during the build steps, though, does it? If this information is not stored anywhere no worries, I can re-build locally.

RMeli:
> Actually, I think I might have been the one breaking things in https://github.com/grimme-lab/multicharge/pull/44, but it is odd that dftd4 is pulling the main branch of multicharge.

I can confirm this is the issue. Version `dftd4@3.6.0` [doesn't pin `multicharge`](https://github.com/dftd4/dftd4/blob/3844dc141936e801ca059d1daea1773c3e6103ec/config/cmake/Findmulticharge.cmake#L20).

I need to fix `multicharge` itself (I tested my PR building `dftd4`, but not using `dftd4` as a dependency of something else), but a quick workaround for CI would be to update to `dftd4@3.7.0` which [does pin the `multicharge` dependency](https://github.com/dftd4/dftd4/blob/7b2ff85a71a3630808fd8a2d972933f75b743f3c/config/cmake/Findmulticharge.cmake#L20).

oschuett:
>  If this information is not stored anywhere no worries, I can re-build locally.

No, currently the build dir is completely removed after every stage to keep the docker image small:
https://github.com/cp2k/cp2k/blob/6843abca0f77ce480c668f110010bf46211d2f12/tools/docker/generate_dockerfiles.py#L793

This could certainly be improved. Let's keep it in mind for when we clean up the toolchain once the Makefile is gone.

RMeli:
cscs-ci run

RMeli:
> How about we enable everything except for GPUs by default?

I had some reservations on this suggestion, but I went with it because of the CI simplifications. To me it makes more sense to enable what I want, instead of disabling what I don't want (and maybe don't know about).

However, in hindsight, I'm not sure this was a good idea. It is great for CI, but it might be rather annoying for an end user, which doesn't have all the dependencies installed.

Does the toolchain installs _all_ the dependencies by default?  In any case, DLA-Future is not installed by the toolchain, so maybe it should be turned OFF, at least for the time being.

* [ ] Fix documentation with the new logic

oschuett:
Yeah, both defaults make sense.

My feeling is that we're already at a point where one can not build CP2K by hand anymore because there are now 25+ dependencies. Since the toolchain and Spack are essentially the only ways to build CP2K, it's ok to optimize for them.

Generally, I'm quite worried about this trend. At the current growth rate we'll soon have 50 dependencies. So, if we decide to continue down this path then we'll need a plan for scaling our dependency management.

hfp:
May thinking was making the CMake discover things as best as possible and introduce more "auto" settings. Ideally, the Toolchain or Spack environment would then dictate what's automatically enabled with CMake. With this, we can also harden and exercise detecting our desired packages/dependencies and this would even carry forward when not using Spack or the Toolchain.

mtaillefumier:
`spack` dictates what dependency is included in the build already. 

My preference is to have all dependencies set to 'off' by default, except for `mpi`, `openmp`, and `dbcsr`. Right now, all options are enabled by default, which I'm fine with, but it can be frustrating, especially when using Spack. For example, if a dependency is not included in the Spack recipe, it causes issues.

The auto-suggestion feature sounds good in theory, but it rarely works well in practice. In my view, it's always better to specify explicitly which dependencies we have. If we do go with the 'auto' option, then dependencies should be made optional, and there should be a mechanism that allows us to ignore the 'auto' selection when needed. 

spack can handle 50 dependencies, cmake as well, it is not a problem.

RMeli:
> For example, if a dependency is not included in the Spack recipe, it causes issues.

This is indeed what happened with this PR, which broke building CP2K from `master` with Spack (already fixed in https://github.com/spack/spack/pull/49489). However, I don't think this is necessarily bad, because it remind us that the package needs to be updated to handle the additional dependencies. However, I think we should at some point re-visit the discussion we had in https://github.com/cp2k/cp2k/pull/3154, and probably have our own copy of the Spack package that is co-developed with CP2K (and then ported upstream).

mtaillefumier:
Indeed, you reacted quickly to this which is good. The argument against making all options ON by default is that it will break all package managers not only `spack` as we have to consider the global picture (apt-get, rpm, dnf, etc..) not only `spack`. 
I would be very upset as a spack user to see a cp2k build not working because of an additional dependency that is not supported by the `package.py` file.

oschuett:
> However, I think we should at some point re-visit the discussion we had in https://github.com/cp2k/cp2k/pull/3154, and probably have our own copy of the Spack package that is co-developed with CP2K (and then ported upstream).

I'm starting to warm up to that idea. However, we would need guardrails like only allowing copies of Spack packages that already exist upstream and are direct dependencies of CP2K. Furthermore, we'll need to set up a build cache to keep the cloud costs in check. Unfortunately, I'm a bit busy right now. So, I suggest we revisit the topic in May.

PR #4065, opened 2025-03-11
DBM: removed supporting pooled host-memory

hfp:
Reduces LOC further after #4064.

### DIFF
--- a/src/dbm/dbm_library.c
+++ b/src/dbm/dbm_library.c
@@ -188,33 +188,19 @@ void dbm_library_print_stats(const int fortran_comm,
 
   dbm_memstats_t memstats;
   dbm_mempool_statistics(&memstats);
-  dbm_mpi_max_uint64(&memstats.device_mallocs, 1, comm);
-  dbm_mpi_max_uint64(&memstats.host_mallocs, 1, comm);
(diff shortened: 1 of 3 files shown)


oschuett:
Is this because we're now using the MPI allocator which also pins the memory?

hfp:
There was no call to the host-flavor anymore because in an earlier PR (have to check which; will comment), the only remaining pooled host-allocation was replaced with the MPI-flavor. This was for code cleanup and performance like the allocation was contended and the pool only added overhead and called MPI-alloc anyway.

hfp:
https://github.com/cp2k/cp2k/pull/4032/files#diff-9e11739de71772f0a09ecca408f2577eb558ea517a9435ba207e5b27dd02613fR405

oschuett:
Ok, thanks. We have to be careful here. Transfers between host and device are much slower with un-pinned memory (presumably because they can't be handed off to the DMA controller).

hfp:
Indeed, so this point broke perhaps when using MPI-alloc for pooled host-allocations. I will check the impact.

hfp:
I will revert this and implement using `offloadMallocHost`, which was never the case.

PR #4066, opened 2025-03-11
Remove old TDDFPT code

juerghutter:
Remove old TDDFPT code
Clean up some code (2nd derivatives XC)


tests/do_regtest.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python3.11
```
oschuett: left over?

juerghutter: Thank you! Yes Opensuse has Python 3.6 as system default. do_regtest needs 3.8 and prettyfier will not allow explicit versions. So it is a constant change of versions in testing cycles and this one was left.

PR #4067, opened 2025-03-11
update xastdp PDOS manual entry

Anyborr:
Change PDOS description to reflect that only the spectrum MOs are projected

### DIFF
--- a/src/input_cp2k_xas.F
+++ b/src/input_cp2k_xas.F
@@ -966,9 +966,103 @@ CONTAINS
       CALL section_add_subsection(subsection, print_key)
       CALL section_release(print_key)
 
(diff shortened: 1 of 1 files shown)


Anyborr:
I was toying with the idea to rename the PDOS section in xas_tdp to something like P_SPEC, but this would break all current input files using this keyword, and require a lot of refactoring in the code to accommodate the changes.

Also, as this PR removes the use of the create_pdos_section function, any changes in that code would have to be manually applied to this code as well. I assume that the PDOS section is basically static, but if i'm wrong then feel free to ignore this PR.

oschuett:
I'd be fine with merging this. @juerghutter ?

PR #4073, opened 2025-03-12
DBM: use host_malloc, introduced host_free and device_free

hfp:
- Internally, use separate pools for device/host memory.
- Let hyperparams enable/disable pooled memory.
- Let hyperparam control offload memory (host).


hfp:
This PR already passes all tests etc. However, performance regressed compared to the current solution (master). This points to the relatively unsophisticated lock in mempool's implementation (critical section). The current solution was profiled and a thread-optimized allocator has shown benefit. I already measured transfer timings with this PR, and performance of transfers is better. However, better transfer times seem to cannot make up for the contention (mempool).

hfp:
To improved the mempool, I may do the following:

* Distinct freeing host/device; currently we have a single `dbm_mempool_free`.
* Split the pool internally into host and device pool; reduces search times.
* Introduce config/hyperparam to disable pooled allocation for host/device.

oschuett:
Btw, this also showed up in the [Performance CUDA Volta](https://dashboard.cp2k.org/archive/perf-cuda-volta/index.html) dashboard. So we can use that test to verify this PR.

hfp:
> Btw, this also showed up in the [Performance CUDA Volta](https://dashboard.cp2k.org/archive/perf-cuda-volta/index.html) dashboard. So we can use that test to verify this PR.

I noticed this too, but was hoping it goes away (naive it is). I will try reproducing it and bisect, but it's likely related to the mempool's strategy of reusing memory. The latter should be easy to verify/rule-out without bisection. It can be also unrelated and due to other changes, e.g., my first turn on mempool's strategy overlapped with some GW changes which in turn impacted things as well.

hfp:
With this PR, we get controls like (`dbm_hyperparams.h`):

* DBM_OFFLOAD_ALLOC
* DBM_MEMPOOL_DEVICE
* DBM_MEMPOOL_HOST

oschuett:
> It can be also unrelated and due to other changes, ...

Yeah, optimizing multi-arch code is not an easy task. Thanks for taking this on!

hfp:
Further investigation/characterization will be post-merge. I can share performance figures using PM.

PR #4075, opened 2025-03-13
[cmake] Cleanup the GPU architecture logic

mtaillefumier:
`CMAKE_HIP|CUDA_ARCHITECTURES` **or** `CP2K_WITH_GPU` should be given explictly. No default values are set.

- simplify the GPU setup logic
- remove a debug message
- Properly handle the C, CXX, HIP, and CUDA standards


CMakeLists.txt
```diff
@@ -92,7 +109,7 @@ endforeach()
 
 # =================================================================================================
 # OPTIONS
-
+option(BUILD_SHARED_LIBS "Build shared library" OFF)
```
RMeli: This one is already set below, with the default value `ON`. Maybe worth separating out CMake options and CP2K-specific options for clarity?

CMakeLists.txt
RMeli, line 1: I think it would be good to define `CP2K_SUPPORTED_HIP_ARCHITECTURES` using `CP2K_SUPPORTED_CUDA_ARCHITECTURES`, instead of repeating the content manually.

mtaillefumier:
I found something not working (edge case). I will fix it later today

mtaillefumier:
OK, the mistake is fixed now. Mixed up the two syntaxes `if ${variable}` and `if variable`. Should be the later

oschuett:
Is this PR ready to be merged?

mtaillefumier:
Yes it is ready.

PR #4078, opened 2025-03-13
Interface for greenX library

dgolze:
Contains the interface to the minimax and analytic continuation component of the [greenX library](https://github.com/nomad-coe/greenX). 
- minimax component is currently used; AC planned to be used
- no integration in toolchain yet
- for cmake build with greenX, main or v2.1 (https://github.com/nomad-coe/greenX/tree/v2.1) of greenX is required.


mtaillefumier (approved):
This PR looks alright to me. I misinterpreted the cmake stuff, apologies for that

RMeli:
GreenX is not available on Spack, but it looks like a solid project that should be easy to add. Once this PR is merged, I'll add it to Spack and then in CI (Spack + CMake).

tests/TEST_DIRS
```diff
@@ -4,7 +4,7 @@
 # in case a new directory is added just add it at the top of the list..
 # the order will be regularly checked and modified...
 QS/regtest-double-hybrid-stress-laplace                     libint libxc
-QS/regtest-rpa-sigma                                        libint
+QS/regtest-rpa-sigma                                        libint greenx=.false.
```
oschuett: The correct syntax here would be `!greenx`, but we currently only use it for leak sanitizer and I don't think we should start using it for features.

I think, an input file should either produce the same result regardless of the presents of a library or fail if the required library is missing.

So, if GreenX produces significantly different results then it's essentially a different method and therefore merits an input keyword to enable/disable it.

dgolze: Hi Ole, I see your point and I was thinking about this, too. For a few cases, greenX and the internal minimax have slightly different grids. Because of numerical instabilities, this affects the results, but only for small grids. For those, I relaxed either the tolerance a little or I used a larger, numerically more stable grids. I can fix the "regtest-rpa-sigma" tests in a similar manner; I have to admit I was a little lazy here. The other case "QS/regtest-gw-kpoints" I would leave because the tests are already relatively long and a larger grids makes them even longer...

oschuett: Sounds like well balanced plan 👍

I'd suggest we use `!greenx` only for the transition period and once we have GreenX in the toolchain we change the reference values and switch to `greenx` (without the `!`).

oschuett:
The recent PRs have unfortunately created some merge conflict. Once those are resolved this seems ready to be merged.

PR #4081, opened 2025-03-14
[hotfix] fix a memory leak

mtaillefumier:
This patch fixes a memory leak that made cp2k crashes after 4 hours of MD simulations. This bug was introduced by commit 49d8227 but was not detected so far.


mtaillefumier:
thanks

PR #4082, opened 2025-03-15
Docker: Fix builds of GCC and minimal tests

oschuett:
Follow up to #4063.


RMeli:
I just saw the e-mails from the dashboard. Thanks for fixing this already!

PR #4083, opened 2025-03-17
PDOS input description

juerghutter:


juerghutter:
Replaces
"update xastdp PDOS manual entry #4067"
This PR generalizes the input description for PDOS from "MO" to "States"

PR #4084, opened 2025-03-18
DBM: print used in addition to committed memory

hfp:
- No restriction on thread-safety (dbm_mempool_statistics).
- Reduced contention of critical section.
- Introduced "used" memory statistic.


hfp:
To benefit from previously introduced offload memory allocation, an actual allocation shall not happen inside of the contended critical section of the memory pool. For thread-safety, it is also not necessary to perform the actual memory allocation inside of the critical section.

hfp:
Looks like "PSI, Merlin" is broken with OpenMPI's implementation of MPI-alloc whereas "MacStadium" tests are broken with limited ability to diagnose like no useful stacktrace. I will replicating the issue. What changed is the allocation of memory for a new chunk of the pool moved out of the critical section, which should fine unless MPI-alloc is not thread-safe...

PR #4089, opened 2025-03-20
DBM: adjusted MPI-based memory allocation

hfp:
- Apparently, MPI_Alloc_mem/free are not fully thread-safe.
- Debug shows issue similar to issues reported elsewhere.
- Improved CHECK macro (however, crash in OpenMPI).
- Note: issue seems specific to OpenMPI 5.x

### DIFF
--- a/src/dbm/dbm_mpi.c
+++ b/src/dbm/dbm_mpi.c
@@ -19,7 +19,7 @@
  ******************************************************************************/
 #define CHECK(status)                                                          \
   if (status != MPI_SUCCESS) {                                                 \
-    fprintf(stderr, "MPI error in %s:%i\n", __FILE__, __LINE__);               \
+    fprintf(stderr, "MPI error #%i in %s:%i\n", status, __FILE__, __LINE__);   \
     MPI_Abort(MPI_COMM_WORLD, EXIT_FAILURE);                                   \
   }
 #endif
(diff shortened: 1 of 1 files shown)


hfp:
I tested [GLOBAL.DBCSR.USE_MPI_ALLOCATOR](https://manual.cp2k.org/trunk/CP2K_INPUT/GLOBAL/DBCSR.html#CP2K_INPUT.GLOBAL.DBCSR.USE_MPI_ALLOCATOR), and it seems to be not affected. Perhaps only the main thread allocates memory.

alazzaro:
> I tested [GLOBAL.DBCSR.USE_MPI_ALLOCATOR](https://manual.cp2k.org/trunk/CP2K_INPUT/GLOBAL/DBCSR.html#CP2K_INPUT.GLOBAL.DBCSR.USE_MPI_ALLOCATOR), and it seems to be not affected. Perhaps only the main thread allocates memory.

The entire DBCSR-MPI is based on the Funneled mode, so this is correct...

hfp:
Thanks for reading!

I am close to the conclusion that MPI allocation hinges on the thread model as well. So far, Intel MPI works fine but for sure it uses a thread-scaling allocator if not TBB's scalable allocator. I will have another turn to fix this for DBM because stock-MPICH on ARM64 so far seems to fail also.

hfp:
See #4090.

PR #4093, opened 2025-03-22
Fix minimal regression test

mkrack:


mkrack:
cscs-ci run eiger-spack,eiger-toolchain

PR #4101, opened 2025-03-25
Add py-torch and update DLA version for Spack/CMake CI

mkrack:


mkrack:
cscs-ci run eiger-spack

PR #4103, opened 2025-03-25
Search for Torch cmake config files

mkrack:


mkrack:
cscs-ci run eiger-spack

PR #4104, opened 2025-03-25
Greenx toolchain

StepanMarek:
Addition of GreenX library to the toolchain script, following #4078 , which introduced the interface in the CMake script + the Makefile infrastructure. There are several things to be ironed out before I feel comfortable turning this into a full pull request, namely:

 - GreenX has optional dependency on GMP. I have included a GMP install script, but I have seen that at least on archlinux, gmp is required dependency of coreutils - is it necessary to be included in this way/should I check for presence of gmp in a different way?
 - GreenX requires linking to stdc++ (at least when using GMP) - should this be explicitly handled or is this already included somewhere else in the toolchain?
 - shellcheck complains about inaccessible files sourced at the starts of the install scripts - I setup ignore for these warnings, but I am unsure whether this is the correct approach

Furthermore, the test for the use of AC component of GreenX in RTBSE (one of the motivations for inclusion of GreenX as a library) is not yet present. I will work on this, but in the meantime, could maybe someone have a look at the current state of the code and let me know if it looks good?

### DIFF
--- a/tools/toolchain/scripts/stage2/install_gmp.sh
+++ b/tools/toolchain/scripts/stage2/install_gmp.sh
@@ -4,16 +4,16 @@
 SCRIPT_DIR="$(cd "$(dirname "$SCRIPT_NAME")/.." && pwd -P)"
 
 gmp_ver="6.3.0"
-gmp_sha256="a3c2b80201b89e68616f4ad30bc66aee4927c3ce50e33929ca819d5c43538898"
-# shellcheck source=/dev/null
+gmp_sha256="e56fd59d76810932a0555aa15a14b61c16bed66110d3c75cc2ac49ddaa9ab24c"
+# shellcheck disable=SC1091
 source "${SCRIPT_DIR}"/common_vars.sh
-# shellcheck source=/dev/null
+# shellcheck disable=SC1091
 source "${SCRIPT_DIR}"/tool_kit.sh
-# shellcheck source=/dev/null
+# shellcheck disable=SC1091
 source "${SCRIPT_DIR}"/signal_trap.sh
-# shellcheck source=/dev/null
+# shellcheck disable=SC1091
 source "${INSTALLDIR}"/toolchain.conf
-# shellcheck source=/dev/null
+# shellcheck disable=SC1091
 source "${INSTALLDIR}"/toolchain.env
 
 [ -f "${BUILDDIR}/setup_gmp" ] && rm "${BUILDDIR}/setup_gmp"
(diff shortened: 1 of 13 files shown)


tools/toolchain/scripts/stage2/install_gmp.sh
```diff
@@ -0,0 +1,112 @@
...
+    echo "==================== Installing GMP ===================="
+    pkg_install_dir="${INSTALLDIR}/gmp-${gmp_ver}"
+    install_lock_file="$pkg_install_dir/install_successful"
+    if verify_checksums "${install_lock_file}"; then
+      echo "gmp-${gmp_ver} is already installed, skipping it."
+    else
+      if [ -f gmp-${gmp_ver}.tar.xz ]; then
+        echo "gmp-${gmp_ver}.tar.xz is found"
+      else
+        # TODO : Exchange once the gmp is present on the CP2K website
```
oschuett: I've uploaded a copy (note I switched to .gz):
https://www.cp2k.org/static/downloads/gmp-6.3.0.tar.gz

oschuett (commented):
>  GreenX has optional dependency on GMP. I have included a GMP install script, ...  it necessary to be included in this way/should I check for presence of gmp in a different way?

This is already taken care of the by the _"Finding GMP from system paths"_ part. It's triggered by supplying `--with-gmp=system`

>  GreenX requires linking to stdc++ (at least when using GMP) - should this be explicitly handled or is this already included somewhere else in the toolchain?

We're already linking to the stdc++ because it's needed by many other packages too.

> shellcheck complains about inaccessible files sourced at the starts of the install scripts - I setup ignore for these warnings, but I am unsure whether this is the correct approach

You can also use `# shellcheck disable=SC1091` to be more specific. Unfortunately, we started using shellcheck after the toolchain was written. So, in most scripts we simply do `# shellcheck disable=all`.

StepanMarek:
Thanks a lot for the review @oschuett !

> 
> This is already taken care of the by the _"Finding GMP from system paths"_ part. It's triggered by supplying `--with-gmp=system`
> 

I understood the GMP dependence therefore correctly it seems - it should be an additional dependency.

> 
> We're already linking to the stdc++ because it's needed by many other packages too.
> 

Great!

> 
> You can also use `# shellcheck disable=SC1091` to be more specific. Unfortunately, we started using shellcheck after the toolchain was written. So, in most scripts we simply do `# shellcheck disable=all`.
>

I will try to incorporate this to be more explicit.

tools/toolchain/scripts/stage2/install_gmp.sh
StepanMarek: Great, I will change this in the next commit. Thank you!

StepanMarek:
I think the pull request is now ready to be opened. Summary of the additional changes:

 - I included a new unit test gx_ac_unittest, which fails when launched after compilation without -D__GREENX flag and does a very small Padé-Thiele interpolation otherwise.
 - I specified the exception for shellcheck to SC1091

I am not sure whether other CI checks should not be triggered, since I am changing the toolchain? Maybe `Regtest psmp` or one of the `Ubuntu, GCC` checks?

oschuett:
LGTM. I love that you added a unit test!

Should we now also change the reference values for the regtests in `QS/regtest-rpa-sigma` and `QS/regtest-gw-kpoints` and then switch their entries in [TEST_DIRS](https://github.com/cp2k/cp2k/blob/master/tests/TEST_DIRS) from `!greenx` to  `greenx`?

StepanMarek:
I have updated the reference values for the regtests that involve the minimax grids from GreenX and changed the conditions for the tests - `QS/regtest-rpa-sigma` and `QS/regtest-gw-kpoints` now only runs when CP2K is compiled with GreenX support.

Also tagging @dgolze , since her pull request (#4078) is closely related.

StepanMarek:
Dear @oschuett , I noticed that the `Spack` and `Regtest psmp` checks compile without the greenx flag, should I somehow change that?

oschuett:
> I noticed that the Spack and Regtest psmp checks compile without the greenx flag, should I somehow change that?

Yes, please enable GreenX by default [here](https://github.com/cp2k/cp2k/blob/0c3e00032cd252947bef8d190319d9d30b3798fa/CMakeLists.txt#L131) and then disable it for Spack and minimal builds [here](https://github.com/cp2k/cp2k/blob/master/cmake/cmake_cp2k.sh).

StepanMarek:
The GreenX compulation option was added to default builds with exception of spack builds and minimal builds.

PR #4117, opened 2025-04-02
Use Spack fork for the time being

mkrack:
The spack develop branch has been quite volatile recently and required changes for the CP2K CI build are still in the pipeline. The dust will hopefully settle once Spack v1.0 is released.

### DIFF
--- a/tools/docker/generate_dockerfiles.py
+++ b/tools/docker/generate_dockerfiles.py
@@ -863,9 +863,9 @@ RUN apt-get update -qq && apt-get install -qq --no-install-recommends \
 # Install a recent Spack version
 WORKDIR /root/spack
 ARG SPACK_VERSION
-ENV SPACK_VERSION=${{SPACK_VERSION:-test_cp2k}}
+ENV SPACK_VERSION=${{SPACK_VERSION:-develop}}
 RUN git init --quiet && \
-    git remote add origin https://github.com/mkrack/spack.git && \
+    git remote add origin https://github.com/spack/spack.git && \
     git fetch --quiet --depth 1 origin ${{SPACK_VERSION}} --no-tags && \
     git checkout --quiet FETCH_HEAD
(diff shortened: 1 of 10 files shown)


RMeli:
Thanks @mkrack, I think this is a good idea for the time being, until things become a bit more stable and ironed out upstream. (At work I've been keeping `spack@0.23.1` and ported the latest versions of the needed recipes to a custom [Spack repository](https://spack.readthedocs.io/en/latest/repositories.html).)

oschuett:
Would it make sense to host a spack fork under https://github.com/cp2k ?

mkrack:
> Would it make sense to host a spack fork under https://github.com/cp2k ?

Sure, that's possible, but I currently try to follow the suggestion of @RMeli (see above) to keep specific (experimental) Spack build recipes local in the CP2K tree under `tools/spack` as an additional `repo` which is more efficient than maintaining a full Spack fork.

tools/docker/generate_dockerfiles.py
```diff
@@ -863,7 +863,7 @@ def install_deps_spack() -> str:
 # Install a recent Spack version
 WORKDIR /root/spack
 ARG SPACK_VERSION
-ENV SPACK_VERSION=${{SPACK_VERSION:-a3abc1c492f2431f477a63bbccb48aa3a2d34199}}
+ENV SPACK_VERSION=${{SPACK_VERSION:-develop}}
```
oschuett, line 866: I believe this will make the build unreproducible. Ideally we'd use a tag like `develop-2025-03-30`, but a commit sha would also work.

mkrack: Yes, indeed, we can move back to a fixed tag for the `SPACK_VERSION` after we can keep now very recent (proprietary) spack recipes within the CP2K tree if needed.

PR #4120, opened 2025-04-03
Bugfix in Davidson solver

NachoXmex:
The content of evals is needed to calculate the error after a restart. See qs_tddfpt2_eigensolver.F:878

### DIFF
--- a/src/qs_tddfpt2_methods.F
+++ b/src/qs_tddfpt2_methods.F
@@ -1074,7 +1074,6 @@ CONTAINS
             IF ((conv <= tddfpt_control%conv) .OR. iter >= niters) EXIT
 
             ! ... otherwise restart Davidson iterations
-            evals = 0.0_dp
             IF (log_unit > 0) THEN
                WRITE (log_unit, '(1X,25("-"),1X,A,1X,25("-"))') "Restart Davidson iterations"
                CALL m_flush(log_unit)


juerghutter:
This line was recently added as a bug fix for internal restarts. Can you give an example of wrong behavior that is fixed with deleting the line again?
Thank you.

NachoXmex:
[debug.txt](https://github.com/user-attachments/files/19583072/debug.txt)
[output_master.txt](https://github.com/user-attachments/files/19583073/output_master.txt)
[output_fixed.txt](https://github.com/user-attachments/files/19583074/output_fixed.txt)

Check convergence output of Davidson iterations after the restart

BelizSertcan:
Hi, the removed line exists because as you can see in output_fixed before the restart only 6 states converged and after the restart all states are converged. Without this line, the evals from the previous step are compared with themselves after the restart, due to line
         `evals(1:nstates) = evals_last(1:nstates)` 
in [qs_tddfpt2_eigensolver.F](https://github.com/cp2k/cp2k/blob/1b487436cea33cc77d18136cb8fd9d0916f57139/src/qs_tddfpt2_eigensolver.F#L941C10-L941C50) at the end of Davidson solver.
This leads to false convergence in line
`conv = MAXVAL(ABS(evals_last(1:nstates) - evals(1:nstates)))`
in [qs_tddfpt2_eigensolver.F](https://github.com/cp2k/cp2k/blob/1b487436cea33cc77d18136cb8fd9d0916f57139/src/qs_tddfpt2_eigensolver.F#L878) after the restart.
Therefore before a restart the evals should be cleared for a correct convergence.

NachoXmex:
I see that even though evals_last is evaluated again after the restart, it is done with the previous approximation to the eigenvectors. Thus, getting the same evals from the last iteration, within numerical noice, hence the ~1.E-16 Convergence.
I now close this pull request.

PR #4123, opened 2025-04-04
CMake: Introduce CP2K_USE_EVERYTHING option

oschuett:
This is an attempt to address https://github.com/cp2k/cp2k/pull/4063#issuecomment-2733998709 by essentially bringing back the preset to include _everything_.

What do you think?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,21 +114,26 @@ option(CMAKE_POSITION_INDEPENDENT_CODE "Enable position independent code" ON)
 option(CP2K_ENABLE_CONSISTENCY_CHECKS
        "Check that the list of compiled files and files contained in src match"
        OFF)
-option(CP2K_USE_DFTD4 "Enable dispersion corrections with DFTD4" ON)
-option(CP2K_USE_DEEPMD "Enable DeePMD" ON)
-option(CP2K_USE_FFTW3 "Use fftw3 for the calculating fast fourier transforms"
-       ON)
-option(CP2K_USE_MPI "Enable MPI support" ON)
-option(CP2K_USE_LIBINT2 "Enable libint2 support" ON)
-option(CP2K_USE_VORI "Enable libvori support" ON)
-option(CP2K_USE_SPGLIB "Enable spglib support" ON)
-option(CP2K_USE_LIBXC "Enable libxc support" ON)
-option(CP2K_USE_LIBTORCH "Enable libtorch support" ON)
+
+option(CP2K_USE_EVERYTHING
+       "Enable all dependencies. They can be individually turned off again."
+       OFF)
+
+option(CP2K_USE_DFTD4 "Enable DFTD4 support" ${CP2K_USE_EVERYTHING})
+option(CP2K_USE_DEEPMD "Enable DeePMD" ${CP2K_USE_EVERYTHING})
+option(CP2K_USE_FFTW3 "Enable fftw3 support" ${CP2K_USE_EVERYTHING})
+option(CP2K_USE_MPI "Enable MPI support" ${CP2K_USE_EVERYTHING})
+option(CP2K_USE_LIBINT2 "Enable libint2 support" ${CP2K_USE_EVERYTHING})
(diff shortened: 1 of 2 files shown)


abussy:
I like the idea of having a minimalistic default, with the option of turning everything on at once. However, I think that a few packages are necessary for a default build, such that meaningful calculations can still be done. Namely, I think that `MPI`, `LIBXC` and `LIBINT2` should be there.

mtaillefumier:
- I think this PR offers a good compromise by setting all the dependencies off by default. However, I would suggest enabling MPI and a few other basic libraries by default. This way, we'd have a base version of CP2K that supports most common calculations, providing a more complete starting point for users. 
- Would it be possible to move the cp2k_cmake.sh file to a different location? Since it's a tool primarily used for CI/CD processes and not directly related to the core cmake modules, relocating it might help maintain clearer separation between tools and build configurations.

RMeli (approved):
I think it's a good idea, thank you for taking this on.

mkrack:
The Spack/CMake CI builds use a `CP2K_BUILD_TYPE` variable. Currently, I am using only `minimal` and `all`, but there could be some further variations following the `toolchain` arch files. What's about having a few CP2K_BUILD_TYPEs which represent each a certain `ON`-`OFF` pattern of `-DCMAKE_foo` flags, e.g. defined by `-DCP2K_BUILD_TYPE="all"`. We could have

- `all` (all dependencies included)
- `all_static` (for the serial static cp2k.ssmp binary, see toolchain static arch file)
- `benchmark` (includes basic features needed for the benchmarks like MPI, libint, libxc as proposed above by @abussy)
- `production` (default, basically `all` without some exotic (rarely used or still tested) features)
- `minimal` and `minimal_static` (quick CI testing)

RMeli:
I removed similar presets in https://github.com/cp2k/cp2k/pull/4063 because they are always a bit arbitrary (i.e. everyone will have a slightly different opinion on what should be `ON` and what should be `OFF`) and I don't see much value in them (I often have my Spack configuration in an environment file, or have the CMake invocation readily available in a script as it is the case in CI). But if there is consensus on a few configurations, I'm not opposed. I was just trying to start simplifying and streamlining the `CMakeLists.txt` files.

mtaillefumier:
@RMeli I agree that we can not have too many presets. Minimal and all are in my view fine. If the build is not specified then only include whatever dependency makes sense (MPI, libxc, libint for instance). if we all in one state by default then all `OFF` is much better.

mtaillefumier:
> The Spack/CMake CI builds use a `CP2K_BUILD_TYPE` variable. Currently, I am using only `minimal` and `all`, but there could be some further variations following the `toolchain` arch files. What's about having a few CP2K_BUILD_TYPEs which represent each a certain `ON`-`OFF` pattern of `-DCMAKE_foo` flags, e.g. defined by `-DCP2K_BUILD_TYPE="all"`. We could have
> 
>     * `all` (all dependencies included)
> 
>     * `all_static` (for the serial static cp2k.ssmp binary, see toolchain static arch file)
> 
>     * `benchmark` (includes basic features needed for the benchmarks like MPI, libint, libxc as proposed above by @abussy)
> 
>     * `production` (default, basically `all` without some exotic (rarely used or still tested) features)
> 
>     * `minimal` and `minimal_static` (quick CI testing)

it is possible to do this with `cmake` without modifying the `CMakeLists.txt`. `CMake` supports this natively (cmake presets I think).

mtaillefumier:
https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html

oschuett:
Thanks for all your feedback. It seems this wasn't a totally terrible idea :-)

I shared the hesitation towards adding more build-types/presets directly to cmake because of the potential clutter.

So, I'd just leave it to the package maintainers to decide what are sensible defaults for their users.

And for the few people brave enough to build CP2K from scratch, the minimal config is actually a good starting point.

> Would it be possible to move the cp2k_cmake.sh file to a different location?

Since `cp2k_cmake.sh` is now much simpler, I actually wanted to explore eliminating it altogether. Otherwise, we can of course move it elsewhere.

PR #4126, opened 2025-04-08
DBM/OpenCL: DBCSR/DBM kernel interoperability

hfp:
- Revised parameter format description (support DBCSR-style in kernel).
- DBM_MULTIPLY_SMM=-1: use DBM-kernels for DBCSR.
- DBM_MULTIPLY_SMM=1: use DBCSR-kernels for DBM.

Other:
- Bulk-Lock c_dbcsr_acc_opencl_info_devptr (c_dbcsr_acc_opencl_info_devptr_lock).
- Call dbm_multiply_gpu_launch_info only if necessary.
- Pass stream in non-const fashion.
- Miniapp: drop CFLAGS (linkage).

### DIFF
--- a/src/dbm/Makefile
+++ b/src/dbm/Makefile
@@ -45,8 +45,8 @@ CFLAGS += -fopenmp -Wno-vla-parameter
 CC := $(if $(filter-out 0,$(MPI)),mpicc,gcc)
 else
 MKL_FCRTL := intel
-LIBS += $(if $(OMPRT),-l$(OMPRT),-qopenmp)
-CFLAGS += -qopenmp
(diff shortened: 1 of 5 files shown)


hfp:
Preliminary insight: universal DBM-kernel is typically ~20% behind specialized/auto-tuned DBCSR-kernel in real workloads. The artificial dataset used for auto-tuning DBCSR (CUDA/HIP/OpenCL) has higher sharing of C-indexes which enables longer accumulation phases without write-back to global memory (FP-atomics). The artificial dataset typically shows 2x for the auto-tuned kernels vs the universal DBM-kernel. This confirms practical performance for DBM's approach vs highly parameterized/specialized kernels.

PR #4127, opened 2025-04-09
Check variable only if defined

mkrack:


mkrack:
@oschuett The `Misc` test is failing, but this seems unrelated to this commit. Any idea?

oschuett:
It's the release of e3nn. I'll fix it and then pin the version.

PR #4136, opened 2025-04-11
DBM: reformulated control-flow to workaround an issue in Intel Compiler

hfp:
- Issue is reported and will be resolved. However:
- Released compilers are enabled by the workaround (WA), and:  
  ( we like the WA so much, we want to keep it :-)


hfp:
It's the best compiler ever, it won't break, and it always wants to continue no matter how difficult the conditions are.

PR #4145, opened 2025-04-17
Add local Spack recipe for SIRIUS v7.7.0

mkrack:


mkrack:
The two failing tests are caused by the earlier PR #4139.

PR #4148, opened 2025-04-17
[toolchain] cleanup

mtaillefumier:
- removed the two sirius patches from the toolchain
- Added a local copy of the cp2k package.py file


tools/precommit/check_file_properties.py
```diff
@@ -64,6 +64,7 @@
     r"__YUKAWA",
     r"__cplusplus",
     r"HIP_VERSION",
+    r"__NO_IPI_DRIVER",
```
oschuett, line 67: Are you sure? The flags has been rename to `__NO_SOCKETS` in https://github.com/cp2k/cp2k/commit/cd366a4022a.

oschuett (commented):
I understand that we sometimes need to fork Spack packages of our dependencies, but why do you want to fork the package for CP2K itself?

tools/precommit/check_file_properties.py
mtaillefumier: It shows up when I run make pretty as this symbol is included in one patch

mtaillefumier:
spack was updated. I will close this pr

PR #4181, opened 2025-05-02
add inital ace support

hforbert:


hforbert:
regtest pdbg fails but it failed before the same way, can't merge though because failed test even if not required (note to self, don't request non-required regtests that current fail in master.... opening new pull request to reset test list.

oschuett:
Sorry for the hassle. We should definitely fix that pdbg test ASAP.

In the future you can also tag @cp2k/cp2k-admins as we have the ability to merge PRs despite failing tests.

PR #4185, opened 2025-05-06
Switch to recent COSMA commit from master branch

mkrack:


tools/spack/cp2k_deps_all_psmp.yaml
```diff
@@ -95,8 +95,8 @@ spack:
     - "mpich@4.3.0"
     - "openblas@0.3.29"
     - "netlib-scalapack@2.2.2"
-#   - "cosma@2.6.6" Use master until a new release will be available
-    - "cosma@master"
+#   - "cosma@2.6.6" Use recent master version until a new release will be available
+    - "cosma@master-20250402"
```
RMeli, line 99: FYI, since the `cosma` package has a `git` attribute, the following syntax can be used to install a given commit:
```suggestion
    - "cosma@git.3e06ce2b0fc04253e6e249403a8f6c2bb4689ffd"
```
This would avoid having to create a custom version (and possibly needing a custom `cosma` package).

RMeli (commented):
As an aside (I just noticed `cosma` is a custom package), I think that as much as possible should be contributed to upstream Spack instead of using custom Spack packages, although I totally understand that the current transition of Spack towards 1.0.0 makes things hard. But we should aim at reducing/eliminating the custom packages in the long term (once Spack is stable).

tools/spack/cp2k_deps_all_psmp.yaml
RMeli: [Spack - Basic Usage - Git Versions](https://spack.readthedocs.io/en/v0.23.1/basic_usage.html#git-versions)

mkrack: Agree. A new (more frequent) COSMA release would facilitate things as well. For the time being, I have stopped contributing to spack upstream, because it is somewhat discouraging. Maybe, it will become better after Spack version 1.0.0 is released.

RMeli: > A new (more frequent) COSMA release would facilitate things as well.

Totally! I will discuss this internally and we will try to release more frequently.

> I have stopped contributing to spack upstream, because it is somewhat discouraging.

I totally understand. Locally I've been also using Spack 0.23.1 and a custom repository of packages if I need the latest version. Hopefully this will change soon (I already did a few tests with 1.0.0 and it seems they ironed out a lot of things already).

Let's revisit when Spack is more stable. And if you need a quick review and merge upstream, feel free to tag me.

mkrack: Thanks. I will drop the local custom cosma recipe when it is possible.

Issue #3424, opened 2024-05-22
Drop support for QUIP and PEXSI


oschuett:
I think we should drop support for QUIP and PEXSI because for all I can tell they are not really being used. 

While in the past the maintenance effort was rather low, we would now have to invest quite a bit of work to migrate them to CMake and Spack.

Are there any objections?

fstein93:
I am in favor of discussing about it (maybe at the next developers' meeting). I have never used any of these packages but I am not an average user anyways.
A workaround for PEXSI could be to move on to PEXSI 2.0.0. According to its Manual, it does build with CMake (see [here](https://pexsi.readthedocs.io/en/latest/install.html?highlight=cmake#build-pexsi)).

RMeli:
I'm usually in favor of removing unused/barely used code. However, I just want to point out that PEXSI is available in Spack.

In https://github.com/spack/spack/pull/36049 I added version `1.2` and `2.0`, but never got around to test them with CP2K. As @fstein93 pointed out, these two versions use CMake and we could get rid of the custom `FindPEXSI.cmake` we have.

We clearly never tested `cp2k+pexsi` since I can see that `CP2K_USE_PEXSI` is missing from the CP2K Spack package, so a bit more work will be needed there too.

QUIP, however, doesn't seem to be available in Spack, so it will need to be added if it is needed as a dependency.

oschuett:
Thanks for chiming in. I have no doubt that we _could_ keep supporting these libraries, but there is an opportunity cost.

Patrick already concluded in [his thesis](https://dx.doi.org/10.3929%2Fethz-a-010738698) that PEXSI is essentially always outperformed by diagonalization - with the exception of extreme strong scaling:

> Currently, the main limitation for a broad application of the PEXSI approach
> to a wide range of systems is its high computational costs in the sense that
> PEXSI starts to be beneficial only at large system sizes and requires massive
> parallelisation.

He also noted that other codes fare better because of their more localized basis sets:

> The results obtained in this work can not be directly compared to the SIESTA-
> PEXSI approach because SIESTA uses numerical basis functions [45, 46] while
> CP2K is based on Gaussian-type functions.

As for QUIP, I believe its crown jewel GAP, while groundbreaking at the time, has since been superseded by more sophisticated ML potentials such as NNP, NequIP/Allegro, and DeePMD.

It's also noteworthy that both packages are only ever mentioned in the forum in the context of installation issues.

[...]

PR #4203, opened 2025-05-14
Recover PEXSI support (with Spack)

mkrack:


oschuett:
Out of curiosity, have you found an actual application for CP2K+PEXSI?

mkrack:
> Out of curiosity, have you found an actual application for CP2K+PEXSI?

No, just revived it, because it was easy with Spack and it still seems to work. Moreover, [CP2K is still referenced](https://pexsi.readthedocs.io/en/v2.0.0/introduction.html#pexsi-used-in-external-packages) as a package using PEXSI and there has been obviously some effort to clean up and improve the PEXSI build and documentation for version 2.0.0. Alternatively, we could drop PEXSI support completely, i.e. we would have to remove the PEXSI related CP2K source code.

oschuett:
The main motivation for dropping PEXSI was to unblock the CMake/Spack migration (#3424). I'm actually quite surprised that the upgrade to PEXSI 2 went so smoothly. I was under the impression that their API had changed significantly.

Since there is still no evidence that PEXSI+CP2K is actually being used, maybe we'll just leave it as deprecated (#3600) until something tips the scales one way or the other.

PR #4210, opened 2025-05-19
Update Spack version to develop-2025-05-18

mkrack:


oschuett:
Btw, I noticed that we're not actually using the public build cache (because it doesn't work for some reason):
https://github.com/cp2k/cp2k/blob/2e57c770f48e434ccf41ff660f62c0ff4c89c16b/tools/docker/Dockerfile.test_spack#L63

Since we now have our own local cache, the public cache really only matters when upgrading Spack - like this PR does.

mkrack:
> Btw, I noticed that we're not actually using the public build cache (because it doesn't work for some reason):
> 
> https://github.com/cp2k/cp2k/blob/2e57c770f48e434ccf41ff660f62c0ff4c89c16b/tools/docker/Dockerfile.test_spack#L63
> 
> Since we now have our own local cache, the public cache really only matters when upgrading Spack - like this PR does.

Indeed, I think we can safely delete [these lines](
https://github.com/cp2k/cp2k/blob/2e57c770f48e434ccf41ff660f62c0ff4c89c16b/tools/docker/Dockerfile.test_spack#L60C1-L63)?

PR #4211, opened 2025-05-19
Add documentation for local Spack cache

RMeli:
Add documentation from the local Spack cache set up by @oschuett in #4189.

Some documentation is already present in `tools/docker/README.md`, but this PR adds it to the manual in the Spack section.

### DIFF
--- a/docs/getting-started/build-with-spack.md
+++ b/docs/getting-started/build-with-spack.md
@@ -133,7 +133,7 @@ From here, you can iterate CP2K development using the native build system as Spa
 Some CP2K Docker containers install all the necessary dependencies with Spack.
 Building with Spack can be time-consuming, because all dependencies are built from source.
 In order to speed-up development and testing with the Spack-based Docker containers,
-you can enable a [Spack cache] locally:
+you should enable a [Spack cache] locally:
 
 ```bash
 ./tools/docker/spack_cache_start.sh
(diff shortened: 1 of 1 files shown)


docs/getting-started/build-with-spack.md
```diff
@@ -128,6 +128,29 @@ spack build-env CP2K_SPEC -- bash
...
+you can enable a [Spack cache] locally:
+
+```bash
+./tools/docker/spack_cache_start.sh
+```
+
+This command starts a local [MinIO] server that can be used as cache for the Spack packages.
+The `spack_cache_stop.sh` script stops the server.
+
+You can then build the the Docker container (here the `test_spack` container, as an example)
```
oschuett: typo: `the the`

docs/getting-started/build-with-spack.md
```diff
@@ -128,6 +128,29 @@ spack build-env CP2K_SPEC -- bash
 
 From here, you can iterate CP2K development using the native build system as Spack would use it.
 
+### Spack and CP2K Docker Containers
+
+Some CP2K Docker containers install all the necessary dependencies with Spack.
+Building with Spack can be time-consuming, because all dependencies are built from source.
+In order to speed-up development and testing with the Spack-based Docker containers,
+you can enable a [Spack cache] locally:
```
oschuett: Actually, the `Dockerfile.test_spack` expects the cache by default. So, maybe mention that building without a cache can be done via `--build-arg SPACK_CACHE=""`.

Thinking about this, it's really a tripping hazard. So, maybe we should add small script that prints a more useful error message when the cache is missing?

Edit: `--build-arg SPACK_CACHE=""` won't work right now, but it really should.

oschuett (approved):
LGTM, thanks for adding!

docs/getting-started/build-with-spack.md
oschuett: I've fixed `--build-arg SPACK_CACHE=""` and added an error message in #4216.

RMeli: Thanks Ole, sorry for the delay on this one.

PR #4213, opened 2025-05-19
Drop public Spack cache

mkrack:


oschuett:
Thanks for cleaning this up. It's a real pity that the public cache doesn't work though. 

While the local cache helps with the development, for new users running `spack install cp2k` still takes a looong time.

PR #4224, opened 2025-05-28
Add regression tester for new Intel ifx compiler

mkrack:
Test the latest Intel oneAPI HPC Toolkit version with ssmp and psmp using the new ifx Fortran compiler. The existing Intel tests use still the old ifort compiler.

### DIFF
--- a/tools/dashboard/dashboard.conf
+++ b/tools/dashboard/dashboard.conf
@@ -202,6 +202,18 @@ report_url:  https://storage.googleapis.com/cp2k-ci/dashboard_hip-rocm-build_rep
 
 # ==============================================================================
 
+[intel-oneapi-hpckit-ssmp]
+sortkey:     1028
+name:        Intel oneAPI HPC Toolkit (ssmp)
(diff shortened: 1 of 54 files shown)


tools/dashboard/dashboard.conf
```diff
@@ -202,6 +202,18 @@ report_url:  https://storage.googleapis.com/cp2k-ci/dashboard_hip-rocm-build_rep
 
 # ==============================================================================
 
+[intel-oneapi-hpckit-ssmp]
```
oschuett, line 205: The `dashboard.conf` is only for displaying test results.
To actually run the new tests, you'll also need to add them to [tools/docker/cp2k-ci.conf](https://github.com/cp2k/cp2k/blob/master/tools/docker/cp2k-ci.conf).

mkrack:
Merging in spite of test failure when using the Intel Fortran compiler ``ifx`` to keep track of the status in the [CP2K dashboard](https://dashboard.cp2k.org/index.html).

PR #4230, opened 2025-06-03
GW: Minor changes for input handling

MGraml:
- add a warning for new low-scaling GW code when GW restart files are used, but SCF is reiterated
- enable (canonically scaling) GW calculations with BASIS_SET NONE, e.g. for usage of point charges.


JWilhelm (approved):
Looks good to me

PR #4233, opened 2025-06-04
PROT: fallback from PROT to ROT

hfp:
- Avoids error condition for single-rank/non-MPI execution (positive info-value).
- Code cleanup.


hfp:
Waiting for the [Dashboard](https://dashboard.cp2k.org/index.html) to catch up (before eventually reverting this PR).

PR #4245, opened 2025-06-13
Fix Gromacs build for the QM/MM tester

mkrack:


oschuett:
It should be possible to build Gromacs without the Python modules. Maybe `-DGMX_PYTHON_PACKAGE=OFF` does the [trick](https://github.com/cp2k/cp2k/blob/2c50328c88b3ff0000aaa390397027da6e790083/tools/docker/scripts/test_gromacs.sh#L28)?

mkrack:
> It should be possible to build Gromacs without the Python modules. Maybe `-DGMX_PYTHON_PACKAGE=OFF` does the [trick](https://github.com/cp2k/cp2k/blob/2c50328c88b3ff0000aaa390397027da6e790083/tools/docker/scripts/test_gromacs.sh#L28)?

Thanks for the hint, @oschuett, I will try that.

mkrack:
> > It should be possible to build Gromacs without the Python modules. Maybe `-DGMX_PYTHON_PACKAGE=OFF` does the [trick](https://github.com/cp2k/cp2k/blob/2c50328c88b3ff0000aaa390397027da6e790083/tools/docker/scripts/test_gromacs.sh#L28)?
> 
> Thanks for the hint, @oschuett, I will try that.

It did not work for me.

PR #4246, opened 2025-06-13
Fixed data races and undefined behavior

hfp:
- Assert presence of certain arguments.
- Fixed potential data races (Helgrind).
- Avoid undefined behavior (UBSan).

Minor (qs_harmonics_atom.F):
- reduced memory consumption.

### DIFF
--- a/src/aobasis/ai_oneelectron.F
+++ b/src/aobasis/ai_oneelectron.F
@@ -547,6 +547,8 @@ CONTAINS
                   END DO
                END DO
 
+               CPASSERT(PRESENT(iatom) .AND. PRESENT(jatom) .AND. PRESENT(katom))
+               CPASSERT(PRESENT(deltaR))
                DO idir = 1, 3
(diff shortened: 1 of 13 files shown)


hfp:
Regarding the data races, none of the typical compilers seem to detect the `lock_num` missing from `SHARED` or `PRIVATE` clause, i.e., `DEFAULT(NONE)` did not safe us and some compilers may not proactively privatize `lock_num`.

oschuett:
> `DEFAULT(NONE)` did not safe us 

Oof!

Out of curiosity, how long did it take to run Helgrind and UBScan?
For comparison, the regtests with [address sanitizer](https://dashboard.cp2k.org/archive/asan-psmp/index.html) take current 53 minutes.

hfp:
Sorry, late reply (PASC). I only ran select workloads with UBSan. I will try it for the entire set of regtests and report on it. For the DEFAULT(NONE), I may report it as a defect at least for Intel Fortran. As an excuse, this is likely caused by the "weird" way of declaring the aforementioned variable.

PR #4259, opened 2025-06-23
Update Spack and add support for Spack packages repository

RMeli:
This PR updates Spack and adds support for the [Spack packages](https://github.com/spack/spack-packages) repository. The transition transition to Spack `v1.0` is getting more stable (a release is coming soon) and the packages have been separated out from the main repository.

Changes:
* Get specific versions of `spack` and `spack-packages`
* Rename custom repo `cp2k_repo` to avoid ambiguity
* Update custom packages to new syntax (eventually we should get rid of them)

Additionally, this PR updates the versions of `dla-future` and `dla-future-fortran`.


tools/docker/Dockerfile.test_spack
```diff
@@ -41,15 +42,27 @@ RUN python3 -m venv /opt/venv
...
+ENV SPACK_VERSION=${SPACK_VERSION:-2e7168b4cfe9222a94becf0c2c5b401f513181ec}
+ARG SPACK_PACKAGES_VERSION
+ENV SPACK_PACKAGES_VERSION=${SPACK_PACKAGES_VERSION:-ec6c0f3435d3a862337a7af71d07790f4d116274}
+ARG SPACK_REPO=https://github.com/spack/spack
+ENV SPACK_ROOT=/opt/spack-$SPACK_VERSION
+ARG SPACK_PACKAGES_REPO=https://github.com/spack/spack-packages
+ENV SPACK_PACKAGES_ROOT=/opt/spack-packages-$SPACK_PACKAGES_VERSION
+RUN mkdir -p $SPACK_ROOT \
+    && curl -OL $SPACK_REPO/archive/$SPACK_VERSION.tar.gz \
+    && tar -xzvf $SPACK_VERSION.tar.gz -C /opt && rm -f $SPACK_VERSION.tar.gz \
```
mkrack: I suggest to drop the `v` flag for the `tar` commands (see also below) to reduce the amount of output.

RMeli:
```suggestion
    && tar -xzf $SPACK_VERSION.tar.gz -C /opt && rm -f $SPACK_VERSION.tar.gz \
```

tools/docker/Dockerfile.test_spack
```diff
@@ -41,15 +42,27 @@ RUN python3 -m venv /opt/venv
...
+    && mkdir -p $SPACK_PACKAGES_ROOT \
+    && curl -OL $SPACK_PACKAGES_REPO/archive/$SPACK_PACKAGES_VERSION.tar.gz \
+    && tar -xzvf $SPACK_PACKAGES_VERSION.tar.gz -C /opt && rm -f $SPACK_PACKAGES_VERSION.tar.gz
```
RMeli:
```suggestion
    && tar -xzf $SPACK_PACKAGES_VERSION.tar.gz -C /opt && rm -f $SPACK_PACKAGES_VERSION.tar.gz
```

tools/docker/Dockerfile.test_spack
```diff
@@ -41,15 +42,27 @@ RUN python3 -m venv /opt/venv
...
 ARG SPACK_VERSION
-ENV SPACK_VERSION=${SPACK_VERSION:-develop-2025-05-18}
-RUN git init --quiet && \
-    git remote add origin https://github.com/spack/spack.git && \
-    git fetch --quiet --depth 1 origin ${SPACK_VERSION} --no-tags && \
-    git checkout --quiet FETCH_HEAD
-ENV PATH="/root/spack/bin:${PATH}"
+ENV SPACK_VERSION=${SPACK_VERSION:-2e7168b4cfe9222a94becf0c2c5b401f513181ec}
+ARG SPACK_PACKAGES_VERSION
+ENV SPACK_PACKAGES_VERSION=${SPACK_PACKAGES_VERSION:-ec6c0f3435d3a862337a7af71d07790f4d116274}
```
RMeli:
```suggestion
ENV SPACK_PACKAGES_VERSION=${SPACK_PACKAGES_VERSION:-c63151a5baffcd96978966cd5ed83a6b8328e026}
```

tools/spack/cp2k_deps_all_psmp.yaml
```diff
@@ -102,13 +100,13 @@ spack:
     - "dbcsr@2.8.0"
     - "deepmdkit@3.0.2"
     - "dftd4@3.7.0"
-    - "dla-future@0.9.0"
+    - "dla-future@0.10.0"
     - "dla-future-fortran@0.4.0"
```
RMeli:
```suggestion
    - "dla-future-fortran@0.5.0"
```

tools/docker/generate_dockerfiles.py
```diff
@@ -886,15 +887,27 @@ def install_deps_spack(version: str) -> str:
...
 ARG SPACK_VERSION
-ENV SPACK_VERSION=${{SPACK_VERSION:-develop-2025-05-18}}
-RUN git init --quiet && \
-    git remote add origin https://github.com/spack/spack.git && \
-    git fetch --quiet --depth 1 origin ${{SPACK_VERSION}} --no-tags && \
-    git checkout --quiet FETCH_HEAD
-ENV PATH="/root/spack/bin:${{PATH}}"
+ENV SPACK_VERSION=${{SPACK_VERSION:-2e7168b4cfe9222a94becf0c2c5b401f513181ec}}
+ARG SPACK_PACKAGES_VERSION
+ENV SPACK_PACKAGES_VERSION=${{SPACK_PACKAGES_VERSION:-ec6c0f3435d3a862337a7af71d07790f4d116274}}
```
RMeli:
```suggestion
ENV SPACK_PACKAGES_VERSION=${{SPACK_PACKAGES_VERSION:-c63151a5baffcd96978966cd5ed83a6b8328e026}}
```
Forgot to update the Python script...

tools/spack/cp2k_repo/repo.yaml
```diff
@@ -0,0 +1,3 @@
+repo:
+  namespace: cp2k_repo
```
oschuett: Does the repo really need its own sub-directory? It seems kinda wasteful to add a directory level for just two entries. Note that, you can still place the files into a subdir when copying them into the container.

If you decide to keep the current directory structure then you should also adopt the following:

https://github.com/cp2k/cp2k/blob/68ff919/tools/precommit/precommit.py#L219-L220

oschuett (approved):
Generally LGTM, just one nit.

tools/spack/cp2k_repo/repo.yaml
RMeli: I think this is a requirement for the new `v2.0` API. The directory name should match the namespace. This is to integrate with Python import system. I can try to revert to the `v1.0` API if it's preferred, but I think having a sub-directory for the Spack repo is cleaner and better going forward.

If this is OK, I'll adapt the pre-commit. Otherwise revert to `v1.0`.

mkrack: I suggest to use the new API `v2.0` and thus to adapt pre-commit. I came already across that new constraints when switching to API `v2.0.` The folder structure is now tight like the namespace which has to match the folder name and I think the name convention should now be `foo_repo`. The new API v2.0 is less permissive.

oschuett: > I think this is a requirement for the new v2.0 API. The directory name should match the namespace. 

Ok, that's a good reason!

One last thought: The name `cp2k_repo` feels a bit generic. Maybe `dev_repo` would be better to emphasize that we intent to upstream all these packages in a timely manner.

RMeli: I agree that `cp2k_repo` might be a bit generic, but `dev_repo` is even more so especially if the internal CP2K repo is added by users to their own Spack installation (with `spack repo add`): it would appear as `dev_repo` in their `spack repo list`, which might contain many other repositories.

What about `cp2k_dev_repo`?

oschuett: > What about cp2k_dev_repo?

Perfect :-)

PR #4261, opened 2025-06-25
Switch from minimal to full build on daint

mkrack:


RMeli:
The Spack error is not super clear here, but it is because `libxsmm@1.17` requires `x86_64`. I think more recent versions do support `aarch64`, but I don't know what the status currently is (a comment in the `libxsmm` Spack package mention `libxsmm@2.0.0` planned for Jan/Feb 2024...). On GH200 we don't use `libxsmm` for now.

RMeli:
Thanks for porting the Spack changes of #4259 to the CSCS CI as well, I totally forgot to do it, sorry.

mkrack:
> The Spack error is not super clear here, but it is because `libxsmm@1.17` requires `x86_64`. I think more recent versions do support `aarch64`, but I don't know what the status currently is (a comment in the `libxsmm` Spack package mention `libxsmm@2.0.0` planned for Jan/Feb 2024...). On GH200 we don't use `libxsmm` for now.

For the time being, `libxsmm` is disabled with Spack as part of this PR until `libxsmm@2.x` will be available.

PR #4279, opened 2025-07-08
CMake: revised compiler configuration

hfp:
* Provide compiler configurations for Intel LLVM-based compilers.
* Definitions for GNU and Intel Compilers:
  - Define __HAS_IEEE_EXCEPTIONS (Release builds).
  - Define __MAX_CONTR=4.
* Minor format cleanup.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -378,10 +378,10 @@ if(CP2K_USE_MPI)
   if(NOT MPI_Fortran_HAVE_F90_MODULE)
     message(
       FATAL_ERROR
-        "\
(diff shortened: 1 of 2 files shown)


hfp:
The compiler ID in CMake changed for Intel "classic" to "next-gen" compilers (`Intel` -> `IntelLLVM`).

hfp:
Now, `__MAX_CONTR=4` is leveraged for GNU and Intel Compilers. If it would not be used, we could potentially drop this sort of "manual code unroll" (entire `__MAX_CONTR` to be dropped eventually).

PR #4282, opened 2025-07-10
sanitizer: fixed more undefined behavior

hfp:
- Usually sanitizer can see subsequent exit conditions aka assertions; however, this may not be true to CPASSERT.
- In any case, included CPASSERTs instead of only/implicit assumptions.
- malloc: resolved implementation specific assumption about zero-sized allocation resulting into NULL-pointer.

### DIFF
--- a/src/dbm/dbm_distribution.c
+++ b/src/dbm/dbm_distribution.c
@@ -30,8 +30,10 @@ static void dbm_dist_1d_new(dbm_dist_1d_t *dist, const int length,
   dist->nranks = dbm_mpi_comm_size(comm);
   dist->length = length;
   dist->index2coord = malloc(length * sizeof(int));
-  assert(dist->index2coord != NULL);
-  memcpy(dist->index2coord, coords, length * sizeof(int));
+  assert(dist->index2coord != NULL || length == 0);
+  if (length != 0) {
+    memcpy(dist->index2coord, coords, length * sizeof(int));
+  }
 
   // Check that cart coordinates and ranks are equivalent.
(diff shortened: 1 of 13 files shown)


oschuett:
> Usually sanitizer can see subsequent exit conditions aka assertions; however, this may not be true to CPASSERT.

We have [this ~~EXIT~~ STOP statement](https://github.com/cp2k/cp2k/blob/2f51be0cd664684a312da5e722ab7f98bfce3a24/src/base/base_hooks.F#L165), which is never reached and merely intended as a hint for compilers and tools. However, it's in a different module and might therefore get overlooked?

hfp:
> > Usually sanitizer can see subsequent exit conditions aka assertions; however, this may not be true to CPASSERT.
> 
> We have [this EXIT statement](https://github.com/cp2k/cp2k/blob/2f51be0cd664684a312da5e722ab7f98bfce3a24/src/base/base_hooks.F#L165), which is never reached and merely intended as a hint for compilers and tools. However, it's in a different module and might therefore get overlooked?

I have to look a bit deeper, but C's `assert` is recognized and able to mute "issues" whereas CPASSERT is not able to achieve the same in the context of UBSan.

hfp:
Infra looks like "macro -> wrapper -> funptr -> exit" which might break things because of the function pointer indirection. Nowadays, compilers have attributes like "never returns" ([`NORETURN`](https://gcc.gnu.org/onlinedocs/gfortran/ATTRIBUTES-directive.html)) which is probably what's collected by tools. I will take a look at our infra/flow and check if we can make use of such attributes.

PR #4285, opened 2025-07-11
FFTW: removed distinction with MKL

hfp:
- Kept FFTW3_MKL for CMake to allow separate FFTW library (even if MKL is linked).
- Removed source code distinctions for FFTW3_MKL (-D__FFTW3_MKL).
- Removed code related to MKL 10.x (thread-safety).
- Enabled Guru interface for MKL as well.
- Updated INSTALL.md.


fstein93:
Sidenote: I am currently working a new FFT backend for CP2K superseding the old one. My current schedule is its incorporation with the release at the end of the year.

hfp:
> Sidenote: I am currently working a new FFT backend for CP2K superseding the old one. My current schedule is its incorporation with the release at the end of the year.

Thank you for the note, Frederick!  I hope merging this PR is not disturbing too much. In any case, I believe revisiting some of the hand-tuned cases like flattening 3d to 2d/1d FFTs may yield different performance today when keeping it 3d for instance. If you come up with special cases or batched cases, I would like to hear too; maybe I have something to contribute.

hfp:
Merging this, will not only get of `__FFTW3_MKL` but also get rid of the CP2K flag denoting this case. On an extended note, our CMake setup with `CP2K_USE_FFTW3_WITH_MKL` denotes the original/separate FFTW (instead of the MKL built-in), which is essentially the opposite of `__FFTW3_MKL` like simply flagging MKL to deliver it. That confusion is hopefully now gone too.

fstein93:
This PR does not disturb my progress as it is an entire rewrite in C (and another license) exploiting 1D, 2D and 3D-FFTs, R2C/C2R-FFTs and distributed FFTs in FFTW (see [my repository](https://github.com/fstein93/cp2k/tree/Migrate_RS_grids_profiling)). The R2C/C2R-FFTs help for sure, but it's not so clear for distributed FFTs. Standalone, it seems to be at least as performant as the old one but I still need to integrate the backend into CP2K.
Regarding the Guru-interface, we have the same issues with cuFFTW (different from cuFFT) and I still need to integrate the Guru interface in my code. I am currently facing exceedingly large planning times (even with planning mode ESTIMATE) in serial mode and larger grids which could be solved with the Guru interface (or a similar approach).

hfp:
@fstein93 Perhaps this [work](https://github.com/intel/double-batched-fft-library) can be of interest.

fstein93:
> @fstein93 Perhaps this [work](https://github.com/intel/double-batched-fft-library) can be of interest.

Thank you, I will have a look at it.

hfp:
> > @fstein93 Perhaps this [work](https://github.com/intel/double-batched-fft-library) can be of interest.
> 
> Thank you, I will have a look at it.

Don't get distracted with this ;-)  This library is used by BigDFT because it's interesting but mostly because of some cool feature related to "callbacks". It has also the potential to better cover some of the cases which are "guru" otherwise. I think, MKL may incorporate this work in the future while providing established interfaces like FFTW3 and MKL's own [DFTi interface](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2025-2/fourier-transform-functions.html). The latter for instance is leveraged by QE. On a side-note, I still think MKL does not implement FFTW's guru interface but this PR cleaned it up anyway like turned to a function "is guru support" (creating a test-plan) by failing with it and going ahead. Also, regarding planning time - I remember work going into reducing planning time at least with MKL but in general that can be an interesting point of optimization like "in bounds" vs "hardly amortizable" depending on the FFT(W) implementation.

PR #4289, opened 2025-07-14
Fixed undefined behavior

hfp:
- Avoid zero-sized allocation (c99/stack/alloca).
- Avoid NULL-pointer for notnull attributed arg.

### DIFF
--- a/src/dbm/dbm_shard.c
+++ b/src/dbm/dbm_shard.c
@@ -105,15 +105,18 @@ void dbm_shard_copy(dbm_shard_t *shard_a, const dbm_shard_t *shard_b) {
   }
   shard_a->data_size = shard_b->data_size;
 
-  if (shard_a->blocks != NULL) {
+  if (shard_b->nblocks != 0) {
+    assert(shard_a->blocks != NULL && shard_b->blocks != NULL);
     memcpy(shard_a->blocks, shard_b->blocks,
            shard_b->nblocks * sizeof(dbm_block_t));
   }
-  if (shard_a->hashtable != NULL) {
+  if (shard_b->hashtable_size != 0) {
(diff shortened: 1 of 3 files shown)


hfp:
This seems to be the last of actionable issues. Current remainder looks like:

> src/common/memory_utilities.F:113:38: runtime error: load of value 278800992, which is not a valid value for type 'logical(kind=4)'
src/common/memory_utilities.F:113:38: runtime error: load of value 3235785312, which is not a valid value for type 'logical(kind=4)'
src/common/memory_utilities.F:113:38: runtime error: load of value 538976288, which is not a valid value for type 'logical(kind=4)'
src/core_ppl.F:758:32: runtime error: load of value 4294967295, which is not a valid value for type 'logical(kind=4)'
src/hfx_compression_methods.F:236:111: runtime error: member access within null pointer of type 'struct hfx_container_node'
src/input/input_keyword_types.F:351:71: runtime error: negation of -9223372036854775808 cannot be represented in type 'integer(kind=8)'; cast to an unsigned type to negate this value to itself
src/motion/rt_propagation.F:542:32: runtime error: member access within null pointer of type 'struct admm_type'

The integer related issues are mostly about hashing and subsequent cast of integer value. The other issues about null pointer I cannot get rid of using CPASSERT; they may look like false positives or thing to ignore for the time being.

hfp:
@oschuett I had a look at [ATTRIBUTES](https://gcc.gnu.org/onlinedocs/gfortran/ATTRIBUTES-directive.html) and tested https://github.com/hfp/cp2k/commit/b6331b49e93bff1a5ec427b038b7d446c4d42bb7, which does not help wrt CPASSERT (also given you already had compiler hints aka `STOP`).

PR #4292, opened 2025-07-15
Match command comments with the gpu-ver supported list

Metaphorme:
Match command comments with the [gpu-ver supported list](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/install_cp2k_toolchain.sh#L498).


hfp:
LGTM.

PR #4293, opened 2025-07-15
MKL/FFTW3: append fftw to MKL's include directory

hfp:
- Code cleanup (fftw3_lib.F).


hfp:
This accounts for the missing include directory like include "fftw3.f03" when MKL provides FFTW3 interface.

PR #4294, opened 2025-07-15
Fix writing forces to .xyz files

### LINKED ISSUE #3956: Broken formatting of force xyz files
I believe #3778 broke Phonopy (and possibly other tools) because it also changed the format of the forces written to file:

```
&FORCE_EVAL
  &PRINT
    &FORCES
      FILENAME forces
    &END FORCES
  &END PRINT
  ...
&END FORCE_EVAL
```

Phonopy [expects](https://github.com/phonopy/phonopy/blob/master/example/Si-CP2K/Si-supercell-001-forces-1_0.xyz) a force xyz file like this: 

```
 ATOMIC FORCES in [a.u.]

 # Atom   Kind   Element          X              Y              Z
      1      1      Si         -0.00258246    -0.00000000     0.00000000
 ...
     64      1      Si          0.00000630     0.00000684    -0.00000588
 SUM OF ATOMIC FORCES           0.00000002     0.00000000     0.00000000     0.00000002
Si-supercell-001-forces-1_0.xyz (END)
```

but our force files currently look like this:

```
 FORCES| Atomic forces [hartree/bohr]
 FORCES|   Atom     x               y               z               |f|
 FORCES|      1 -2.58245602E-03 -1.44924495E-09  4.98348825E-13   2.58245602E-03
 ...
 FORCES|     64  6.30480306E-06  6.83541468E-06 -5.87596425E-06   1.10000178E-05
 FORCES| Sum     1.86532986E-08  4.16491159E-12  3.94881104E-12
 FORCES| Total atomic force                                       1.86532995E-08
Si-supercell-001-forces-1_0_mine.xyz (END)
```

oschuett:
Fixes #3956 and presumably other workflows.

### DIFF
--- a/src/force_env_utils.F
+++ b/src/force_env_utils.F
@@ -16,6 +16,7 @@ MODULE force_env_utils
    USE constraint,                      ONLY: rattle_control,&
                                               shake_control
    USE constraint_util,                 ONLY: getold
+   USE cp_log_handling,                 ONLY: cp_logger_get_default_io_unit
    USE cp_subsys_types,                 ONLY: cp_subsys_get,&
                                               cp_subsys_type
    USE cp_units,                        ONLY: cp_unit_from_cp2k
@@ -34,7 +35,8 @@ MODULE force_env_utils
    USE particle_list_types,             ONLY: particle_list_type
    USE particle_types,                  ONLY: update_particle_set
    USE physcon,                         ONLY: angstrom
-   USE string_utilities,                ONLY: lowercase
+   USE string_utilities,                ONLY: lowercase,&
+                                              uppercase
 #include "./base/base_uses.f90"
 
    IMPLICIT NONE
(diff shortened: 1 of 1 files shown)


oschuett:
@mkrack, as simplification I removed the options `zero_force_core_shell_atom` and `grand_total_force` from `write_forces()`. I hope that's ok for you?

oschuett:
I'm having second thoughts...

Since we [default](https://github.com/cp2k/cp2k/blob/ae240b2ef64b703281b19f8f1314bce4e7b7779d/src/force_env_methods.F#L368) to the .xyz extension, I figured it would be a good idea to write the forces in the offical [XYZ format](https://en.wikipedia.org/wiki/XYZ_file_format).

However, the more user-friendly move is probably to restore our old (custom) format that we had before #3778.


Meanwhile the Phonopy developers have already adopted to our new (custom) format (https://github.com/phonopy/phonopy/pull/538). Fortunately, they also kept the parser for the old format.

Any opinions?

mkrack:
> @mkrack, as simplification I removed the options `zero_force_core_shell_atom` and `grand_total_force` from `write_forces()`. I hope that's ok for you?

I don't see how this simplification for the forces printout can work for the special case of core-shell model runs which include three particle sets, i.e. cores, shells, and atoms, instead of just one (atoms).

mkrack:
> I'm having second thoughts...
> 
> Since we [default](https://github.com/cp2k/cp2k/blob/ae240b2ef64b703281b19f8f1314bce4e7b7779d/src/force_env_methods.F#L368) to the .xyz extension, I figured it would be a good idea to write the forces in the offical [XYZ format](https://en.wikipedia.org/wiki/XYZ_file_format).

CP2K can write already the particle forces like the particle coordinates or velocities in [XMOL format](https://github.com/cp2k/cp2k/blob/master/src/motion_utils.F#L392) (and [here](https://github.com/cp2k/cp2k/blob/master/src/particle_methods.F#L227url)). 

> 
> However, the more user-friendly move is probably to restore our old (custom) format that we had before #3778.

Why is the old format more user-friendly?

> Meanwhile the Phonopy developers have already adopted to our new (custom) format ([phonopy/phonopy#538](https://github.com/phonopy/phonopy/pull/538)). Fortunately, they also kept the parser for the old format.
> 
> Any opinions?

It's generally a bad idea to parse proprietary output files which are primarily written for human readers. The `.restart` files or the `frc` output files in XMOL or DCD format would be a better source as these are written in a well defined format.

If these files in standard formats are not written for each `RUN_TYPE` and `METHOD` combination, then we should use the existing routines for printing these files in a standard format.

oschuett:
> It's generally a bad idea to parse proprietary output files which are primarily written for human readers.

While I generally agree, the format has been stable for the last 15 years. Hence, it's not a surprise that people built [workflows](https://xkcd.com/1172) around it. And popular run types, like `ENERGY_FORCE`, do indeed not write trajectories at the moment.

> Why is the old format more user-friendly?

My hope is that by sticking to the old format when writing to a file we could avoid breaking all those workflows.

PR #4298, opened 2025-07-16
Save Mulliken atomic orbital populations

rangsimanketkaew:
Save Mulliken atomic orbital populations to results


fstein93:
Is this feature already covered within the regression test suite?

rangsimanketkaew:
Not really. These saves are done automatically when `PRINT%MULLIKEN` is requested. I think it can be useful and used in other modules.

PR #4305, opened 2025-07-17
Add CSCS tests to CP2K-CI

oschuett:


oschuett:
Since all CSCS tests are now Spack-based, I took the liberty to drop "spack" from their names.

PR #4313, opened 2025-07-18
Re-structure compiler configs and migrate sdbg and pdbg tests to CMake

oschuett:


hfp:
Hmm, not sure why - but I am getting:

> src/common/cp_files.F:563:132:
     2938
     2939      563 |       data_dir_path = __DATA_DIR
     2940          |                                                                                                                                    1
     2941    Warning: Line truncated at (1) [-Wline-truncation]
     2942    src/common/cp_files.F:563:24:
     2943
     2944      563 |       data_dir_path = __DATA_DIR
     2945          |                        1
  >> 2946    Error: Unterminated character constant beginning at (1)

oschuett:
Seems like you data dir path is too long. As a work around you can set `-DCMAKE_INSTALL_PREFIX=` to a reasonably sized path.

hfp:
It's not fully under my control, I am using Spack and it uses paths like `/tmp/hpabst/spack-stage/spack-stage-cp2k-master-4sjfznywkra72iwsvbv35sa5nxfpymbf/spack-build-4sjfzny/src/common`. So, the data path might be quite long given the path to sources.

hfp:
I wonder if dropping `-Werror=line-truncation` for at least `cp_files.F` helps. Though, this worked when I tried until and including last week.

oschuett:
We could move `-Werror=line-truncation` from the base line to the debug build.

hfp:
I am currently debugging something for the Spack based build which looks like:

> spack install cp2k@master+opencl ldflags="-Wl,--wrap=_gfortran_runtime_warning_at" build_type=Coverage %gcc %c,cxx,fortran=gcc ^libxsmm+debug
 ^dbcsr@develop build_type=Coverage ^libint tune=cp2k-lmax-6 ^intel-oneapi-mpi ^[virtuals=scalapack] intel-oneapi-mkl@2025 ^[virtuals=lapack] intel-oneapi-m
kl@2025 ^[virtuals=blas] intel-oneapi-mkl@2025 ^[virtuals=fftw-api] intel-oneapi-mkl@2025

I will document the useful learning from it like how to "allow" `Coverage` at all when using Spack (needs change for the Spack recipe). Btw, I think our compiler configs are skewed towards something unusual. For instance, "Coverage" is the only config combining "-g -O0" while omitting sanitizers. The sanitizers are somewhat disturbing normal debug. I think our "Debug" build should not use anything **beyond** "-O0" as people expect unoptimized code when debugging and best matches for debug symbols. We invented "Coverage" do achieve this since our commonly used "Debug" shows certain issues. This is true btw for DBCSR "as well".

For instance, GFortran issues tons of output for "temporary arrays" making debug a **nightmare**. For this, I have `-Wl,--wrap=_gfortran_runtime_warning_at` which resolves with LIBXSMM (awkward!). Going forward, CP2K might provide "_gfortran_runtime_warning_at" to suppress these messages...

hfp:
> We could move `-Werror=line-truncation` from the base line to the debug build.

Let's do this! It's safer when we want to give best experience for folks compiling the new release.

hfp:
One learning from "old days with libxsmm", building **releases** might drop all "Werror" flags. We can detect straight releases by SHA or whatever and drop Werror. This is because future compilers may be "more accurate with warnings" and hence we potentially get more and more errors as compilers evolve with our aging release. This is somewhat counter-intuitive since people think "it worked before" plus "release code did not change".

hfp:
> I am currently debugging something for the Spack based build which looks like:
> 
> > spack install cp2k@master+opencl ldflags="-Wl,--wrap=_gfortran_runtime_warning_at" build_type=Coverage %gcc %c,cxx,fortran=gcc ^libxsmm+debug
> > ^dbcsr@develop build_type=Coverage ^libint tune=cp2k-lmax-6 ^intel-oneapi-mpi ^[virtuals=scalapack] intel-oneapi-mkl@2025 ^[virtuals=lapack] intel-oneapi-m
> > kl@2025 ^[virtuals=blas] intel-oneapi-mkl@2025 ^[virtuals=fftw-api] intel-oneapi-mkl@2025
> 
> I will document the useful learning from it like how to "allow" `Coverage` at all when using Spack (needs change for the Spack recipe). Btw, I think our compiler configs are skewed towards something unusual. For instance, "Coverage" is the only config combining "-g -O0" while omitting sanitizers. The sanitizers are somewhat disturbing normal debug. I think our "Debug" build should not use anything **beyond** "-O0" as people expect unoptimized code when debugging and best matches for debug symbols. We invented "Coverage" do achieve this since our commonly used "Debug" shows certain issues. This is true btw for DBCSR "as well".
> 
> For instance, GFortran issues tons of output for "temporary arrays" making debug a **nightmare**. For this, I have `-Wl,--wrap=_gfortran_runtime_warning_at` which resolves with LIBXSMM (awkward!). Going forward, CP2K might provide "_gfortran_runtime_warning_at" to suppress these messages...

Perhaps, we invented "Debug" with optimization level to get rid of GFortran's messages about temporary arrays.... then, we added Sanitizers to the "Debug" config ("screwed in any case") and introduced "Coverage" to get back some sort of "Debug" experience...

oschuett:
> Perhaps, we invented "Debug" with optimization level to get rid of GFortran's messages about temporary arrays.... 

My intention was simply to migrate our [old arch files](https://github.com/cp2k/cp2k/blob/12a8a74dd0fb31d489f8772558d306af1b7ebfca/tools/toolchain/scripts/generate_arch_files.sh#L427-L445) to CMake.

> building releases might drop all "Werror" flags.

Yes, that's a good idea. I'll move all the `Werror` flags to the debug build type.

>  "Debug" config ("screwed in any case") and introduced "Coverage" to get back some sort of "Debug" experience...

Personally, I always found the `sdbg` and `pdbg` builds well suited for debugging, but we could of course invent more build types if there is a need.

hfp:
> Personally, I always found the `sdbg` and `pdbg` builds well suited for debugging, but we could of course invent more build types if there is a need.

Indeed, Toolchain- and Makefile based builds are understood and settled. I to soon flawlessly contribute to spack-packages. It can take a moment to contribute let alone Spack's CI (or not rebasing against dev; or latest tested head). There is quite some perfectionism in an imperfect world. Exposing "Coverage" (or any other custom name of a config) together with Spack is quite a machinery. ATM, I am debugging something that only Spack produces; an interesting experience.

oschuett:
I would expect CP2K developers to use Spack only for installing dependencies and then invoke CMake directly.

Hence, I'd only expose "Release" via Spack and consider all other build types for internal-only use. For example, the "Coverage" build type was only meant to produce our nightly [coverage report](https://www.cp2k.org/static/coverage/).

hfp:
I still have this issue. Though, CI tests create dedicated environments with Spack and perhaps line length is more limited. I will try `-ffree-line-length-none`, and it might be needed in general (baseline).

hfp:
As a side-note, apparently only baseline, i.e., GNU-only, enables OpenMP; will check.

oschuett:
> As a side-note, apparently only baseline, i.e., GNU-only, enables OpenMP; will check.

Admittedly, I only polished the flags for GNU. I believe CMake sets `-fopenmp` as well, but we should maybe not rely on it.

hfp:
I will send a PR asap. The line-length thing is indeed resolved by `-ffree-line-length-none`.

mtaillefumier:
We should avoid adding options that are directly controlled by cmake as it can really lead to weird errors especially with openmp (openmp offload tartget comes to mind).

PR #4315, opened 2025-07-18
Resonant Inelastic X-ray Spectroscopy (RIXS) Module

BelizSertcan:
This pull request introduces a new module for Resonant Inelastic X-ray Spectroscopy (RIXS), built upon the existing `xas_tdp` and `tddft` modules. The implementation follows the theoretical framework described in [Vaz da Cruz et al](https://pubs.rsc.org/en/content/articlelanding/2021/cp/d0cp04726k).

Upcoming work will include extension to roks/uks.


src/rixs_methods.F
```diff
@@ -0,0 +1,469 @@
...
+!> \param rixs_env ...
+!> \param rixs_section ...
+! **************************************************************************************************
+   SUBROUTINE print_rixs_to_file(donor_state, mu_i0, mu_if, w_i0, w_if, &
+                                 rixs_env, rixs_section)
+
+      TYPE(donor_state_type), POINTER                    :: donor_state
+      REAL(dp), ALLOCATABLE, DIMENSION(:, :)             :: mu_i0
+      REAL(dp), ALLOCATABLE, DIMENSION(:, :, :)          :: mu_if
+      REAL(dp), ALLOCATABLE, DIMENSION(:)                :: w_i0, w_if
```
fstein93: You can drop the `ALLOCATABLE` attribute here

src/rixs_methods.F
```diff
@@ -0,0 +1,469 @@
...
+!> \brief Rotate vectors. Returns rotated mo_occ and evects.
+!> \param evects ...
+!> \param mo_ref ...
+!> \param mo_occ ...
+!> \param overlap_matrix ...
+!> \param unit_nr ...
+! **************************************************************************************************
+
+   SUBROUTINE rotate_vectors(evects, mo_ref, mo_occ, overlap_matrix, unit_nr)
+      TYPE(cp_fm_type), DIMENSION(:, :), POINTER         :: evects
```
fstein93: You can drop the `POINTER` here

src/rixs_methods.F
```diff
@@ -0,0 +1,469 @@
...
+      diff = MAXVAL(ABS(rotated_mo_coeffs%local_data - mo_occ%local_data))
+      IF (unit_nr > 0) THEN
+         WRITE (unit_nr, FMT="(T9,A,F10.6,/)") "Max difference between orbitals = ", diff
+      END IF
+
+      CALL cp_fm_to_fm(rotated_mo_coeffs, mo_occ)
+
+      nstates = SIZE(evects, DIM=2)
+      DO istate = 1, nstates
+         current_evect => evects(1, istate)
```
fstein93: Use `ASSOCIATE` instead to be able to drop the `POINTER` attribute of `evects`

fstein93 (commented):
I made some specific comments.
I personally try to reduce the use of `POINTER`s in Fortran if it is easily possible (less memory leaks, less worries about incorrectly setup pointers, ...). Furthermore, you could also think of dropping the `POINTER` attributes of your types and allocate them outside their routines such that the creation/release routines are only concerned with the actual basic setup/release of the components of the types.

src/rixs_methods.F
BelizSertcan: is there a way to declare the size without `REAL(dp), DIMENSION(4, nvirt, nstates) :: mu_if`? I never saw declaring size based on other variables this way in cp2k.

fstein93: It is not required to specify the size which is used in most parts of CP2K. By default, the sizes of the array are the same as the one passed to the routine and their lower bounds are 1 if not specified otherwise.

BelizSertcan:
i don't understand what is causing the mem leak @oschuett @fstein93

fstein93:
I compiled an sdbg myself and found the same leaks. I attach the output of one test
[CO-PBE0.inp.log](https://github.com/user-attachments/files/21363911/CO-PBE0.inp.log)

src/rixs_methods.F
```diff
@@ -0,0 +1,470 @@
...
+         CPABORT("XAS_TDP calculation missing")
+      END IF
+      IF (.NOT. ASSOCIATED(tddfp2_section)) THEN
+         CPABORT("TDDFPT calculation missing")
+      END IF
+
+      CALL rixs_core(rixs_section, qs_env)
+
+      IF (output_unit > 0) THEN
+         WRITE (UNIT=output_unit, FMT="(/,(T2,A79)),/") &
```
oschuett: Please remove the spurious closing bracket from the format string:
```
/opt/cp2k/src/rixs_methods.F:109:51:

  109 |          WRITE (UNIT=output_unit, FMT="(/,(T2,A79)),/") &
      |                                                   1
Warning: Extraneous characters in format at (1)
```

oschuett:
@BelizSertcan, did the log that Frederick posted put you on the right track? I know that hunting memory leaks of reference-counted objects can be quite tedious.

BelizSertcan:
> @BelizSertcan, did the log that Frederick posted put you on the right track? I know that hunting memory leaks of reference-counted objects can be quite tedious.

It took me some time to figure the problem was coming from `mo_coeff`, I also had some help from Augustin. It should be fixed now.

PR #4332, opened 2025-07-22
[cmake] Fix fortran compilers options passed to C

mtaillefumier:
- Remove explicit reference to openmp as the target already include it
- Restored -Wextra -Wall for C and C++
- Fix allow-argument-mismatch passed to C and C++ while it is a Fortran option only
- Fix short name issue


oschuett:
As discussed [here](https://github.com/cp2k/cp2k/pull/4313#issuecomment-3096083475), we should not include `-Wall;-Werror` in release builds to be future-proof for new compiler versions.

Nevertheless, it would be nice to include it for debug builds.

oschuett:
> Nevertheless, it would be nice to include it for debug builds.

Although, we'll probably have to fix a bunch of warnings before we can actually turn `-Wall;-Werror` on.

As for `-fbacktrace` and `-fallow-argument-mismatch`, it seems CMake added those flags on its own.

mtaillefumier:
> Although, we'll probably have to fix a bunch of warnings before we can actually turn `-Wall;-Werror` on.

Yes. I found a few already. I can fix some of them for sure.

cmake/CompilerConfiguration.cmake
```diff
@@ -88,12 +91,6 @@ add_link_options(
   "$<$<AND:$<CONFIG:DEBUG>,$<COMPILE_LANG_AND_ID:CXX,GNU>>:-fsanitize=leak>"
   "$<$<AND:$<CONFIG:DEBUG>,$<COMPILE_LANG_AND_ID:C,GNU>>:-fsanitize=leak>")
 
-# TODO Prevent CMake from wrongly adding -fbacktrace and
```
oschuett, line 91: Please leave these comment lines in as the problem still exists.

src/CMakeLists.txt
```diff
@@ -1591,14 +1591,14 @@ add_library(cp2k_linalg_libs INTERFACE)
...
   cp2k_linalg_libs
   INTERFACE cp2k::LAPACK::lapack
-            $<$<BOOL:${CP2K_USE_LIBXSMM}>:cp2k::LibXSMM::libxsmmf>
-            $<$<BOOL:${CP2K_USE_LIBXSMM}>:cp2k::LibXSMM::libxsmmext>
+            $<$<BOOL:${CP2K_USE_LIBXSMM}>:cp2k::LibXSMM::libxsmmf
+            cp2k::LibXSMM::libxsmmext>
             cp2k::BLAS::blas
             $<$<BOOL:${CP2K_USE_CUDA}>:cp2k_cuda_libs>
             $<$<BOOL:${CP2K_USE_HIP}>:cp2k_hip_libs>
-            $<$<BOOL:${CP2K_USE_MPI}>:MPI::MPI_Fortran>
```
oschuett, line 1599: I actually like the repeated conditionals we had before - despite them repeating. Otherwise one always has to check the previous line for a closing `>`

PR #4334, opened 2025-07-22
CMake: adjusted baseline/other compilers

hfp:
- Unlimited line-length for GNU (__DATA_DIR).
- Enabled OpenMP for Intel Compilers.

### DIFF
--- a/cmake/CompilerConfiguration.cmake
+++ b/cmake/CompilerConfiguration.cmake
@@ -35,18 +35,18 @@ endif()
 
 # Baseline
 add_compile_options(
-  "$<$<COMPILE_LANG_AND_ID:Fortran,GNU>:-std=f2008;-fopenmp;-ffree-form;-fimplicit-none>"
+  "$<$<COMPILE_LANG_AND_ID:Fortran,GNU>:-std=f2008;-ffree-form;-ffree-line-length-none;-fimplicit-none>"
   "$<$<COMPILE_LANG_AND_ID:Fortran,GNU>:-g;-fno-omit-frame-pointer;-fbacktrace>"
   "$<$<COMPILE_LANG_AND_ID:Fortran,GNU>:$<$<VERSION_GREATER_EQUAL:${CMAKE_Fortran_COMPILER_VERSION},11>:-fallow-argument-mismatch>>"
   "$<$<COMPILE_LANG_AND_ID:Fortran,GNU>:-Wno-deprecated-declarations;-Wno-maybe-uninitialized;-Wuninitialized;-Wuse-without-only>"
 )
 add_compile_options(
-  "$<$<COMPILE_LANG_AND_ID:CXX,GNU>:--std=c++17;-fopenmp>"
+  "$<$<COMPILE_LANG_AND_ID:CXX,GNU>:--std=c++17>"
   "$<$<COMPILE_LANG_AND_ID:CXX,GNU>:-g;-fno-omit-frame-pointer>"
   "$<$<COMPILE_LANG_AND_ID:CXX,GNU>:-Wno-deprecated-declarations;-Wno-vla-parameter>"
 )
 add_compile_options(
-  "$<$<COMPILE_LANG_AND_ID:C,GNU>:-std=c11;-fopenmp>"
+  "$<$<COMPILE_LANG_AND_ID:C,GNU>:-std=c11>"
   "$<$<COMPILE_LANG_AND_ID:C,GNU>:-g;-fno-omit-frame-pointer>"
   "$<$<COMPILE_LANG_AND_ID:C,GNU>:-Wno-deprecated-declarations;-Wno-vla-parameter>"
 )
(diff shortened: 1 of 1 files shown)


mtaillefumier:
we do not need to specify `-qopenmp` explicitly as cmake adds this compilation flags automatically when `openmp::openmp` is passed as a library dependency.

hfp:
> we do not need to specify `-qopenmp` explicitly as cmake adds this compilation flags automatically when `openmp::openmp` is passed asa library dependency.

So, it can be removed entirely like also from baseline/gnu.

mtaillefumier:
> > we do not need to specify `-qopenmp` explicitly as cmake adds this compilation flags automatically when `openmp::openmp` is passed asa library dependency.
> 
> So, it can be removed entirely like also from baseline/gnu.

indeed we can remove it. I did this in my open pr that fixed the `fallow-argument-mismatch`

hfp:
> > > we do not need to specify `-qopenmp` explicitly as cmake adds this compilation flags automatically when `openmp::openmp` is passed asa library dependency.
> > 
> > 
> > So, it can be removed entirely like also from baseline/gnu.
> 
> indeed we can remove it. I did this in my open pr that fixed the `fallow-argument-mismatch`

Hot development today ;-)  Sorry, for potentially causing rebase issue...

mtaillefumier:
no worries

mtaillefumier:
OK this test 'fails' but considering the difference I would not block the merge

```
/opt/cp2k/build/bin/TEST-2025-07-22_11-28-37/QS/regtest-ls-rtp/H2O-ls-emd-ewindow.inp.out
Spec: {'matcher': 'M067', 'tol': 6e-09, 'ref': 2.23001754e-06}
Difference too large: 8.97e-09 > 6e-09, value: 2.23001756e-06.
```

hfp:
> OK this test 'fails' but considering the difference I would not block the merge
> 
> ```
> /opt/cp2k/build/bin/TEST-2025-07-22_11-28-37/QS/regtest-ls-rtp/H2O-ls-emd-ewindow.inp.out
> Spec: {'matcher': 'M067', 'tol': 6e-09, 'ref': 2.23001754e-06}
> Difference too large: 8.97e-09 > 6e-09, value: 2.23001756e-06.
> ```

Noticed. I will send a separate PR to adjust the tolerance, and I will force this one in...

mtaillefumier:
i thought we are using free form for fortran but gcc complains

```
  204 |       CPABORT("Error! You've to specify either ORAC, CP2K or CPMD!")
      |                                                                                                                                    1
Error: Line truncated at (1) [-Werror=line-truncation]
```

hfp:
Interesting, for me things started to break down approx. with #4313. I have no full explanation or root cause but this PR fixes the issue I had since then. Perhaps, previous compiler configs had some "initial defaults" or something like this. Pretty esoteric.

To be more specific, `common/cp_files.F` ran into an issue with a "very long" data path (despite removing warning/error about line truncation). The Spack compilation line failed with the preprocessed source (I checked the data dir was already expanded and relatively long but not under my control), and line-truncation warnings (plus warning turned into error) was not present on the flag-/command-line.

mtaillefumier:
doing this instead works.

```
IF (COUNT((/l_orac, l_cp2k, l_cpmd/)) /= 1) then
      CPABORT("Error! You've to specify either ORAC, CP2K or CPMD!")
ENDIF
```

mtaillefumier:
something is wrong with gfortran (gcc 14.2.1).

hfp:
> something is wrong with gfortran (gcc 14.2.1).

Possible. I used 14.2.0.

hfp:
( For the record, my primary interest is GNU GCC. Everything is "nice to have" )

Well, GCC + Intel Libs...

oschuett:
> something is wrong with gfortran (gcc 14.2.1).

No, it's a bug in our CMakeLists:

https://github.com/cp2k/cp2k/blob/856d28c997d984a34d5dc115a0dea737ce455ac7/src/CMakeLists.txt#L1628

When `__SHORT_FILE__` is missing then we fall back to `__FILE__` which can easily overflow the line:

https://github.com/cp2k/cp2k/blob/856d28c997d984a34d5dc115a0dea737ce455ac7/src/base/base_uses.f90#L24-L27

Maybe instead of the fall back we should just raise an `#error`?

mtaillefumier:
To be fair we should avoid modifying the source code before compilation. And we never had this issue before so clearly something has changed in between.

mtaillefumier:
Raising an error because the line is too long is also not a solution as we should not care at all about the length of the line to begin with. It is an issue with fortran only.

hfp:
> Raising an error because the line is too long is also not a solution as we should not care at all about the length of the line to begin with. It is an issue with fortran only.

Let's relax and make fun of it. I like source code to be razor sharp on the right hand side like "80 characters are enough". Joke aside, the unclear thing is why this popped up with "mentioned PR" (I am not even sure about it). We may admit: things became more complex since then (out of our control). The build-stack is more spooky with CMake plus Spack than GNU Make alone (not saying one is better than the other).

oschuett:
> To be fair we should avoid modifying the source code before compilation.

Well, this is what the pre-processor does...

CP2K expects that `__SHORT_FILE__` is always set to the base name of the source file. Right now this is not the case for top-level Fortran files (those that start with `PROGRAM`). 

> And we never had this issue before so clearly something has changed in between.

Presumably the path length of your build directory has changed.

mtaillefumier:
the razor sharp is a bit old indeed ;). My point being I do not even understand why we have this to begin with. It is not an issue with spack or any other build system in my view. I get this error when I compile locally with cmake + make. So the issue is not the complexity of spack but the code or the compiler that is not happy for some strange reason.

hfp:
Yeah, "strange reason" qualifies for "complexity". Now, "we are with you" - beside, we would never have this issue with 80 columns only ;-)

Joke aside:

> To be fair we should avoid modifying the source code before compilation.

We have FYPP, like "template meta-programming" ("you wanted it - you got it"). Also, look at https://github.com/cp2k/dbcsr/pull/915 - GNU's preprocessing is broken for corner cases at least. Only exotic 

I (we?) have other issues as well:

* DBCSR is not in lock-step anymore and "not fixed there" means "not fixed here" for the time being/deprecation.
* Spack packages are not in sync; I doubt our pinning covers end-user experience (if they used Spack at all).

mtaillefumier:
short_file issue is fixed now.

oschuett:
I've opened #4338 so we don't forget about the `-ffree-line-length-none` workaround.

PR #4346, opened 2025-07-28
build(docker): use cache mounts for apt and dnf

e-kwsm:
to optimize cache usage

<https://docs.docker.com/build/cache/optimize/#use-cache-mounts>


oschuett:
Thanks a lot for this suggestion. However, there are a number of reasons why I don't want to apply it:

- Downloading packages is actually pretty quick because our CI runs in the Google Cloud.
- We're planing to move to Podman and Spack in the near future.
- Mounting host directories for caching won't help much because the CI containers run in an ephemeral environment (Kubernetes Pods on preemptible VMs).

oschuett:
I'm going to close this PR now. Please do not take this as discouragement.

PR #4352, opened 2025-08-08
Improved check_diag

hfp:
- Only print one deviating value (if there is one).
- Reduced amount of code and stuff in arith. loop.


hfp:
Beside if WARNing or ABORTing the code (if the check fails), this also prints the deviating values. This PR changes the routine to only print any value if it exceeds the threshold. Also, the WRITE is taken out of the loop; printing happens after exiting the loop-nest. Avoids duplication of similar code and reduces LOC.

PR #4357, opened 2025-08-11
Update of active space module

stefabat:
Major update of the active space module.
Added a new longrange truncated potential for the ERI (with test).
Several changes in the input keywords, removed a lot of obsolete code.


src/common/errfun.F
```diff
@@ -0,0 +1,485 @@
...
+! Faddeyeva and Voight functions": efficiency improvements and
+! Fortran translation, ACM Trans. Math. Software 42, Article 26, 2016.
+
+! **************************************************************************************************
+!> \brief Module to compute the error function of a complex argument
+!> \par History
+!>      08.2025 Adapted to use CP2K intrinsics and constants
+!> \author Stefano Battaglia
+! **************************************************************************************************
+MODULE errfun
```
fstein93: Can you find a better name as the functions provided in this module is not the same as the ordinary error function

src/common/errfun.F
```diff
@@ -0,0 +1,485 @@
...
+   REAL(kind=dp), PARAMETER :: zero = 0.0_dp, one = 1.0_dp, &
+                               two = 2.0_dp, half = 0.5_dp, rmin = TINY(one), &
+                               eps0 = EPSILON(one), sqrt_log_rmin = SQRT(-LOG(rmin)), &
+                               pi2 = pi*pi, one_sqrt_pi = one/SQRT(pi)
+   COMPLEX(kind=dp), PARAMETER :: &
+      cmplxj = CMPLX(zero, one, kind=dp), &
+      cmplx0 = CMPLX(zero, zero, kind=dp)
+
+   PRIVATE
+   PUBLIC :: wpop, wzag, erf_pop, erf_zag
```
fstein93: Please choose better names than w or erf which is either too generic or a bit misleading for other developers. I would also rather vote for something like faddeyeva_fast and faddeyeva_accurate to build in the difference between both implementations.

src/common/errfun.F
stefabat: This was the original name from the authors, and I actually only use the **complex** error function from it.

src/common/errfun.F
stefabat: Sure, I'll update that now.

PR #4359, opened 2025-08-13
OpenCL: support Spack/externally built DBCSR

hfp:


hfp:
DFTD4 asset cannot be fetched and hence tests are failing.

oschuett:
> DFTD4 asset cannot be fetched and hence tests are failing.

Yes, I've opened #4362.

PR #4360, opened 2025-08-13
Cleaned up profile output

hfp:


hfp:
https://github.com/cp2k/cp2k/pull/4359#issuecomment-3182704914

Issue #4353, opened 2025-08-09
Feature request: molden output with cell/ECP/ghost/orbital(OT) information


he-zilong:
Greetings. The molden file is a common plaintext file for exchanging wavefunction information, and its output from CP2K calculation is specified by `&FORCE_EVAL/&DFT/&PRINT/&MO_MOLDEN` keywords. However, due to the current lack of some key information in molden output, this feature request is proposing an addition or revision of several sections for ease of post analysis.

- The cell or box of the periodic boundary condition: the settings in `&FORCE_EVAL/&SUBSYS/&CELL` that represents periodicity of geometry and wavefunction is so vital that it is worth including. A new `[Cell]` section with three cartesian components of the three cell vectors written as a 3x3 matrix, where A, B, C each taking a row, and unit in (angstroem|atomic unit) defined by (Angs|AU) keyword like the native `[Atoms]` section, may have a syntax like the following example. Fluctuating cell vectors in a CELL_OPT job, or an MD job with NPT ensemble, are to be captured with molden files of different frames as requeted by `&FORCE_EVAL/&DFT/&PRINT/&MO_MOLDEN/&EACH`.
```
[Cell] Angs
 2.45600000     0.00000000     0.00000000
-1.22800000     2.12695839     0.00000000
 0.00000000     0.00000000     6.69600000
```
- The ECP or valence electron count for each atom kind: since the native `[Atoms]` section contains only atomic number that is not equal to valence electron count in the presence of pseudopotentials, a separate section for that information is needed so that programs parsing the molden file know how many core electrons are covered by ECP and how many are not. According to README of [Molden2AIM](https://zorkzou.github.io/Molden2AIM/readme.html), there are existing `[Pseudo]` or `[NVal]` syntax for this purpose. 
- The presence of ghost atoms: when `&FORCE_EVAL/&SUBSYS/&KIND/GHOST` is set to T for an atom kind, the molden file can mark it as well in one of several formats in the native `[Atoms]` section, going by README of Molden2AIM linked above.
- The occupation number and eigenvalues for _both occupied and vacant_ orbitals from an OT calculation: for some reason, at the moment this is basically only available with diagonalization algorithm. Both the print from `&FORCE_EVAL/&DFT/&PRINT/&MO` (perhaps commit #3230 is related?) and the `[MO]` record in molden files give eigenvalues/energies as **0.00** after an OT calculation in version 2025.1, and no vacant/virtual/unoccupied orbital is in either output.

[...]

PR #4364, opened 2025-08-13
Fix bug part of issue #4353

mkrack:
When using OT, the occcupied orbital eigenvalues are now
- written after SCF convergence to the CP2K output file if requested
- dumped to the binary wavefunction restart file and are printed upon restart if requested
- printed to the MOLDEN output file
- not printed during the OT SCF cycle (only the final orbitals after convergence are printed for the given index range)


mamaskybaby1989-prog (commented):
Help

Issue #4362, opened 2025-08-13
Toolchain: tblite fails to install because of CMake FetchContent


oschuett:
The toolchain [fails to build tblite](https://storage.googleapis.com/cp2k-ci/dashboard_psmp_report.txt) because [CMake can't fetch](https://github.com/tblite/tblite/blob/8f6ed634a4e4656d603720f877307f83c5840516/config/cmake/tblite-utils.cmake#L94) dftd4 since their repository contains some [git submodules](https://github.com/dftd4/dftd4/tree/main/assets/aur) pointing at an old Arch Linux repository that just disappeared.

To fix this we have to stop relying on CMake FetchContent and instead install DFTD4 ourselves.

hfp:
It is at least unreliable as it fetches successfully sometimes...

jpoto:
It seems there is a fix:

https://github.com/dftd4/dftd4/pull/306

with regards to building dftd4 ourselves:
We have toolchain script that can build dftd4, but I would need to look into linking it correctly.

I am not certain if I then should also add build scripts for other dependencies (mctc-lib, toml-f).

hfp:
The fix was merged two hours ago and there is no new release just yet. Also, the release cadence of DFTD4 does not suggest a fix "tomorrow". Is it possible to defer the questions about the toolchain script, mctc-lib, toml-f, and whatever if CP2K would rely on a "SHA-release" for the time being?

oschuett:
> I am not certain if I then should also add build scripts for other dependencies (mctc-lib, toml-f).

Hmm, maybe we should just focus on moving to Spack instead then.

> ... rely on a "SHA-release" for the time being?

Yes, we could try this and hope that the DFTD4 API hasn't changed since their 3.7.0 release. We'd have to patch tblite [here](https://github.com/tblite/tblite/blob/8f6ed634a4e4656d603720f877307f83c5840516/config/cmake/Finddftd4.cmake#L20).

jpoto:
Hi Ole,

I have a few problems with interfaces, but the latest version of tblite was released a few weeks ago. 

Can you add tblite v05.0 to static/downloads?

best,
Johann

mkrack:
[https://www.cp2k.org/static/downloads/tblite-0.5.0.tar.gz](https://www.cp2k.org/static/downloads/tblite-0.5.0.tar.gz)

[...]

PR #4375, opened 2025-08-15
first try to avoid arch error

jpoto:


jpoto:
start to fix #4362

oschuett:
Since we now have a workaround (#4653), I'm going to close this PR.

PR #4379, opened 2025-08-18
Partially reverted #4378

hfp:


hfp:
There seems to be some issue with CMake's `add_custom_command` and `add_custom_target`. The latter runs as expected; in fact it runs like a phony Makefile target, i.e., always. This seems to be the documented behavior like the timestamps of the `BYPRODUCTS` "may or may not be considered" (rebuild). On the other hand, `add_custom_command` produces `OUTPUTS` but does not seem to run (unless some manual "stunts" are made to have some dependency on the output expressed).

PR #4390, opened 2025-08-21
Use CMAKE_BUILD_TYPE generic for RPM build

mkrack:


siteshwar:
There are [several findings](https://openscanhub.fedoraproject.org/task/67555/log/added.html) reported by static analyzers that are introduced by changes in this pull request. Feel free to fix them in upstream if they may be real issues. Thanks!

Issue #4389, opened 2025-08-21
Possible documentation improvement on RELAX_MULTIPLICITY


PolyachenkoYA:
I wanted to do `UKS`, and the role of `RELAX_MULTIPLICITY` was not immediately clear.

The [docs](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT.html#CP2K_INPUT.FORCE_EVAL.DFT.RELAX_MULTIPLICITY) only say `Larger tolerance values increase the probability for a spin flip`.

After searching for a while, I can see many threads about it where people are confused about the underlying mechanics of spin relaxation in CP2K.

I have not found any satisfying explanations in these threads, so I had to dig up the code. My current understanding is that the only way `RELAX_MULTIPLICITY` affects calculations is through being assigned to `flexible_electron_count` through multiple routes, which all then lead to the [qs_mo_occupation.F](https://github.com/cp2k/cp2k/blob/master/src/qs_mo_occupation.F), where on line ~310 (for future updates: search for `Aufbau`) the actual Aufbau implementation is done. The `flexible_electron_count`-s (which hold the `RELAX_MULTIPLICITY` values) give `threshold = MAX(mo_array(1)%flexible_electron_count, mo_array(2)%flexible_electron_count)` which is then used to determine how far the difference in energies of LUMO-s of orbitals of a- and b- spin-orientations can grow before their occupations start to balance each-other.

Now, the logic above gives uneven a- and b- occupations (with a- preferred with the ~`RELAX_MULTIPLICITY` energy gap). However, I don't exactly see how this leads to an `increase the probability for a spin flip` mentioned in the docs. I did not investigate the source code on this one, but the most likely interpretation I can think of is:
The uneven occupations of a- and b- (caused by `RELAX_MULTIPLICITY > 0`) act as something like an initial guess for SCF or other non-linear optimization. Such optimization probably has many local minima, so starting it in a spin-polarized state (a- preferred to b-) makes it "more likely" the converged optimization result ends up in a local minima with the same type of bias (a- preferred to b-). But if it's (close to) true, then I think it can be confusing to say `increase the probability for a spin flip` because:
1) There are no flips of single spins (as I would expect from some monte-carlo style equilibration based on "probabilities"), the optimization just ends up in a spin-polarized state
2) There are no random numbers involved (given the nuclei' positions), so the "probability" mentioned in the docs is not some controlled probability threshold but just a qualitative concept here.

[...]

PR #4391, opened 2025-08-21
Update RELAX_MULTIPLICITY description

PolyachenkoYA:
I did a bit of investigation on how `RELAX_MULTIPLICITY` works (which I found many threads about, and some people I work with were also a bit confused). I outlined my findings and got them [confirmed](https://github.com/cp2k/cp2k/issues/4389) by a CP2K author, so I am now proposing to add the details to the docs. I used a link to the discussion in the description, but I am not sure if I formatted it correctly. I did not find any in-text links in the docs, so I don't know if it's possible to do it. The format I used is from longer CP2K guide-style webpages.


mkrack:
@PolyachenkoYA Thanks for your proposal resolving #4389. LGTM.

PR #4403, opened 2025-08-26
CNEO-DFT energy and force

zc62:
Dear CP2K developers,

I would like to contribute the constrained nuclear-electronic orbital density functional theory (CNEO-DFT) implementation that I presented in the publication https://doi.org/10.1021/acs.jctc.5c00837

CNEO-DFT belongs to multicomponent DFT. In this case, it also quantizes the nuclei of user's choice, but in most cases, hydrogen nuclei. The idea is to take into account nuclear zero-point energy, as well as nuclear delocalization effects directly into DFT calculations. The constraint part gives well-defined molecular structures and classical dynamics (classical equation of motion but with nuclear quantum effects incorporated, see https://doi.org/10.1021/acs.jpclett.2c02905 and https://doi.org/10.1021/jacs.1c12932). The proposed advantages are:

- With geometry optimization, the stable geometry with ZPE correction can be directly identified. As shown in the JCTC paper, the hydrogen adsorption on Pt(111) surface is correctly recognized. In molecular cases and bio-related systems with QM/MM (non-PBC, droplets), our group has demonstrated the success of CNEO-DFT on the description of shared-protons in low-barrier hydrogen bonds.
- Though not thoroughly tested in periodic systems, the hope is CNEO molecular dynamics can generate statistical and dynamical quantities that are similar to path-integral based methods, while keeping the computational cost close to pristine DFT calculations.

The CP2K implementation of CNEO-DFT relies on the GAPW method and the all-electron potential code. We use the distinguishable-particle approximation for quantum nuclei, therefore for each quantum nucleus, only a few localized basis functions are used. It then creates quantum nuclear densities that have angular momenta higher than zero. The electron-nuclear mean-field Coulomb interaction is therefore implemented by bumping up the angular momentum from zero to a general one in the GAPW all-electron smeared ion code. For soft nuclear densities that extend beyond the PAW sphere, I also followed the GAPW code to put them on the global grid. Though for regular nuclei, that part is negligible contributions.

### Code review

I certainly need experts to carefully review the code, because I modified a bunch of contents that may not seem relevant to the method. For example, I just followed every place that added soft rho0, then I also add the soft nuclear density contributions there. Also for QM/MM, I don't know if the modification is really as simple as what I have done now. We have ordinary molecular CNEO+QM/MM implementation in PySCF, but for CP2K the GEEP QM/MM is very different (to our eyes). CP2K CNEO+QM/MM is basically untested.

I also moved some function execution orders in GAPW. I hope that won't create troubles.

### Questions

I found there is no good place to write input documentations. For example, how should I let the users know `POTENTIAL CNEO` can trigger CNEO-DFT calculations (which also needs nuclear basis set input)?

Should I provide test run files and/or examples? Here I attach a file that does an ENERGY_FORCE calculation for H on Pt(111): [CNEO.inp](https://github.com/user-attachments/files/21998280/CNEO.inp.txt)

Zehua Chen, from [Yang group](https://yang.chem.wisc.edu) at University of Wisconsin-Madison

Commit messages:

> Energy and force for constrained nuclear-electronic orbital density functional theory (J. Chem. Theory Comput. 2025, 21, 16, 7865–7877).
>
> `POTENTIAL CNEO` triggers CNEO calculations for certain atom kinds (hydrogen is the most common, but no limit in principle) and nuclear basis functions are supplied by `BASIS_SET NUC basis_name`. PB series basis sets for proton are provided in `data/NUCLEAR_BASIS_SETS`. Note that GAPW is mandatory for CNEO-DFT calculations, and pseudopotentials can be used for atoms types with classical nuclei.
>
> For the electronic part, k-points can be enabled. For quantum nuclei, because they are very localized, we use a distinguishable-particle approximation. This is very similar to the Gamma-point approximation for indistinguishable quantum nuclei, with the added benefit that each quantum nucleus only involves a very small Kohn-Sham equation with localized basis functions. Therefore, k-points settings for the electronic part will not affect the quantum nuclear part.

EDIT:

TODOs for tests and documentation:

- [x] Tests
- [ ] Documentation (will do in a separate PR)


fstein93:
Dear @zc62 ,
Thank you for your contribution. Please add regression tests to the `tests/` directory. We recommend to add additional documentation explaining setup, workflow and parameters to tune accuracy and/or performance in the `docs` directory.

fstein93:
Please run `make -j <N> pretty`

fstein93:
Add the new files to the `src/CMakeLists.txt`

juerghutter:
Dear Zehua Chen

thank you very much for your contribution. Very impressive work.

Getting this PR ready for merging will probably need some more work. You will have to make changes locally.
You can push those changes, they will be automatically added to the current PR, no need to delete/reopen the PR.

The first steps needed are:
1) Run "make pretty" on your code to pass the formatting rules.
2) Rebase your branch to the current master in the CP2K repository.

Tests, examples, and documentation can be added at the end.

src/qmmm_gpw_forces.F
```diff
@@ -216,10 +217,28 @@ SUBROUTINE qmmm_forces(qs_env, qmmm_env, mm_particles, calc_force, mm_cell)
                CALL auxbas_pool%create_pw(rho_tot_r2)
                CALL pw_transfer(rho0_s_gs, rho_tot_r2)
                CALL pw_axpy(rho_tot_r2, rho_tot_r)
+               IF (ASSOCIATED(rhoz_cneo_s_gs)) THEN
+                  NULLIFY (rho_tot_r3)
+                  ALLOCATE (rho_tot_r3)
```
fstein93: rho_tot_r3 need not be a POINTER anymore.

zc62:
I have run the prettifier and added the new files to `src/CMakeLists.txt`. I can probably just copy the hydrogen-containing tests in `tests/QS/regtest-gapw` to make up CNEO tests, but I am less confident about adding documentations. Should it be a separate method under `docs/methods`?

src/qmmm_gpw_forces.F
zc62: I was copying the code above, which used `rho_tot_r` and `rho_tot_r2` and they are pointers. `rho_tot_r3` is a similar temporary variable that I created to avoid reusing a not-yet-released pointer. I think it should still be a pointer? Otherwise the others should be changed too.

fstein93: The routines with which you call `rho_tot_r3`, do not require pointers and the variable is only allocated within the branches. The pools do not require points because they copy the PW grid objects themselves (which is only a few pointers). The code you copy-pasted indeed does not need pointers although it requires a simple modification of two routines. I have already done some cleanup but apparently missed a few spots.

fstein93:
It is up to you to setup regtests. They are just supposed to be fast enough.
I think it is better to have a separate chapter for CNEO except you deem it tightly related to another one.

src/qmmm_gpw_forces.F
fstein93: I removed the pointer attributes of `rho_tot_r` and `rho_tot_r2` in the master branch.

fstein93:
Nevertheless, I will not merge this PR without any regtests to prevent "surprises" afterwards. If possible, every line of code should be tested; the bare minimum should be regtests to cover the core functionalities with the most common combinations of features.

oschuett:
> I can probably just copy the hydrogen-containing tests in tests/QS/regtest-gapw to make up CNEO tests,

Yes, new tests are often derived from existing ones. Go for it :-)
Usually, just a few tests already go a long way towards protecting a feature from future regressions. 
The large test suite is what allows us to quickly accept even broad changes like yours.

zc62:
@fstein93 

> Nevertheless, I will not merge this PR without any regtests to prevent "surprises" afterwards.

Sure! I am not in a hurry to get this PR merged at all. I also expect the manual inspection of the code from core developers will take some time. In the meantime I will think about the tests and documentations. I added two TODOs as a reminder. Also I can convert this to a draft if you prefer.

> If possible, every line of code should be tested

This seems a bit impractical to me, especially for a fortran project without external unit test libraries. For instance, many utility functions or code snippets I added are simply copied and modified from other places in CP2K, while I don't know how to unit test them, and I don't seem to see they get tested in the test suite (CG coefficient generation, atom solver, GAPW local density matrix related operations, etc.). `tests` seems to only test the final outcome of the method, and if there are bugs in some of the functions, as the bugs remain, the results won't change and the tests will pass.

Some general questions about the tests (also for @oschuett):

1. Many of the tests seem to have a tolerance that is unbelievably strict. For example the energy must agree to 1e-14 within the reference value. This is way tighter than the SCF convergence requested and might even be disturbed by numerical errors from MPI or OMP parallel addition. In other projects I use much looser criteria. Is there any requirement on this or can I just determine my tolerance? In the recently force-pushed commit I even tightened the criteria for position constraint (from 1e-10 bohr to 1e-12 bohr) as I don't see I can reach 1e-14 accuracy or even consistency at all.

2. How do I reliably and easily test numerical forces and stress tensors? In addition to test against a known gradient value in the test files, letting the test automatically test numerical gradients adds an extra layer of verification.

fstein93:
1. We do have these issue in other parts of the code, too. I determine the thresholds by running a test in serial and parallel mode and choose the reference value to be the result in serial mode and the threshold such that the result in parallel mode is still correct. If a test of the CI or the dashboard fails, we will simply adjust the threshold.
2. We have a `RUN_TYPE DEBUG` (see [here](https://manual.cp2k.org/cp2k-2024_2-branch/CP2K_INPUT/DEBUG.html)) to compare analytical and numerical gradients.
3. You can add the documentation in a follow-up PR. The regression tests serve as a simple documentation by example but usually with lower accuracy demands to restrict the runtime.

fstein93:
To me, it is sufficient to have regtests for the most relevant parts/combinations of functionalities, usually basic setup in one closed shell and one open-shell setup. In a follow-up PR, you may add regtests for other combinations than the basic setup you deem relevant.

oschuett:
> Many of the tests seem to have a tolerance that is unbelievably strict. 

A tight tolerance only means that a test is very reproducible. The results of most regtests are physically meaningless because they are tuned for speed, e.g. by exiting the SCF loop early. So, the test results are more like a fingerprint of the executed algorithm - which is good enough for an integration test :-)

> I am not in a hurry to get this PR merged at all.

Since you are touching a lot of files, we should try to merge this PR rather soon to avoid merge conflicts.

I'll do another read-through, but my first impression is that your code is already pretty good.

So, just add a couple of tests and once they pass, I think we can merge this PR.

zc62:
Thanks for the info! I can see mine can fluctuate around 1e-14 to 1e-13 for sdbg and psmp (with up to 64 cores), so I chose a relatively safe 1e-12, which I have seen being used by some other tests. Then, four simple total energy tests are added.

I ran a `DEBUG` run and gradient/stress verification was successful.

I also wanted to verify dipole and polarizability, but somehow dipole had disagreements in numerical and analytical ones. First, CNEO dipole is still well defined and easy to compute. Even if we have delocalized quantum nuclei, their expectation positions coincide with the classical geometry, hence the original code for nuclear part should just work (both the calculation of dipole and the interaction energy with the field). Then I thought if it was the reason that I did not consider `EFIELD` in my quantum nuclear code. After some tests (really added the electric field term to nuclear Kohn-Sham Hamiltonian), I realize that it is not needed at all. Any uniform electric field just cancels with the position constraint that is in the form of `f*r`, so without any `EFIELD` implementation it already takes care of any uniform field. It is then a mystery to me why they don't agree when both sound correct. This part might need some further debugging. By the way, `EXTERNAL_POTENTIAL` should just work, because it is put on the global Hartree grid and therefore integrated with GAPW and hence CNEO.
EDIT: mystery resolved, see below.

For polarizability, it seems to require TDDFT, and currently I disable the CNEO+TDDFT combination. Electronic TDDFT with frozen CNEO nuclear density is actually quite simple and we do have implementations in molecular codes and a paper pending. The thing is, I am not quite sure about how TDDFT in CP2K works (for instance, why `zcore=0` in TDDFT) so I decided to postpone it. Also, any further derived quantities will not directly work as CNEO-CPKS or CNEO Z-vector equations will kick in (e.g., static polarizability if calculated by CPKS, and gradients of electronic TDDFT).

For the code review, I almost forgot to mention the following:
https://github.com/cp2k/cp2k/blob/b11b8edb9a25f63dcef58216c2d5cda0b15e040e/src/qs_rho0_methods.F#L450-L451
`lrho1_eq_lrho0` seems to be used nowhere and is always false? To ensure `max_iso_not0` is not messed up in my part if `lrho1_eq_lrho0` happens to be true, I added
```fortran
         IF (ASSOCIATED(cneo_potential)) THEN
            IF (cneo_potential%harmonics%max_iso_not0 > 0) THEN
               CPASSERT(.NOT. gapw_control%lrho1_eq_lrho0)
            END IF
         END IF
```
but I feel like it is not the most elegant way.

oschuett:
Thanks a lot for adding the tests!

> `lrho1_eq_lrho0` seems to be used nowhere and is always false?

Yeah, this looks like a left-over from a refactoring. Feel free to remove it entirely or just leave it and I'll take care.

> It is then a mystery to me why they don't agree when both sound correct. This part might need some further debugging.

Do you consider this a blocker? Otherwise, I'd go ahead and merge this PR now to avoid conflicts.

Then you can follow up to polish things. We usually merge small PRs very quickly.

zc62:
@oschuett The dipole mystery is gone now. I was using a too small box (4A each side). When I use a box with 10A each side, the same molecule, now analytical and numerical dipoles agree precisely. Also the absolute values of both numerical and analytical would change after the change of simulation box, apparently showing the finite box size effect while choosing to use a "local" electric field.

Can you update the master branch regarding `lrho1_eq_lrho0`? Then I will rebase my branch, and delete my code related to that too.

fstein93:
I am taking care of that

fstein93:
@zc62 You can now perform a rebase.

zc62:
@fstein93 Thanks! I have updated the code.

zc62:
@fstein93 I found out the commit message of the merged commit is not the commit message I had in my commit (with appropriate 72-character line breaks):
```
CNEO: CNEO-DFT energy and force

Energy and force for constrained nuclear-electronic orbital density
functional theory (J. Chem. Theory Comput. 2025, 21, 16, 7865–7877).

`POTENTIAL CNEO` triggers CNEO calculations for certain atom kinds
(hydrogen is the most common, but no limit in principle) and nuclear
basis functions are supplied by `BASIS_SET NUC basis_name`. PB series
basis sets for proton are provided in `data/NUCLEAR_BASIS_SETS`.

The `MASS` can be tuned to simulate isotopes such as deuterium, however,
one should be careful that basis functions need to be changed
accordingly. One simple modification to PB4-D with scaled exponents that
is suitable for deuterium is also added to `data/NUCLEAR_BASIS_SETS`

Note that GAPW is mandatory for atoms with quantum nuclei in CNEO-DFT
calculations, and pseudopotentials or even `GPW_TYPE T` can be used for
atom kinds with conventional classical nuclei.

For the electronic part, k-points can be enabled. However, this k-points
setting will not affect the quantum nuclear part. This is because even
quantum nuclei are delocalized as compared to conventional classical
point charges, they are still relatively localized. Therefore, a
distinguishable-particle approximation is used for quantum nuclei. This
is similar to the Gamma-point approximation for indistinguishable
quantum nuclei, with the added benefit that each quantum nucleus only
involves a small Kohn-Sham equation with localized basis functions, and
multiple quantum nuclei create multiple coupled Kohn-Sham equations,
instead of a giant one in the indistinguishable fermion case.
```
(note that I did some last-minute edit so it differs from the one in the post) but a copy of the things I wrote in the beginning of the PR (and without line breaks), is that intended?

Personally I took them differently, commit message to explain what the commit does, and the "letter" in the beginning of the PR to explain why I was opening the PR. Also the questions I had about code review etc. entered the commit message, which may seem a bit awkward.

fstein93:
My bad. I used squash-merge instead of rebase-merge. But still, the letter at the start of your PR documents the change as well and the commit message will be buried by hundreds of other messages later anyways. But it is an excellent starting point for the manual/docs where such an extensive description (of the keywords) is better suited for.

oschuett:
> I found out the commit message of the merged commit is not the commit message I had in my commit

Sorry about that! I fixed it.

PR #4404, opened 2025-08-27
Fixed compilation (ssmp) on Ubuntu GCC 8

hfp:
- Complements #4397.


jpoto:
Thanks for fixing it.

PR #4336, opened 2025-07-22
DBM: Simplify memory pool

oschuett:
Simplifies the DBM memory pool by hard-cording the following hyper-parameters that have proven generally useful: `DBM_MEMPOOL_DEVICE`, `DBM_MEMPOOL_HOST` and `DBM_ALLOC_OFFLOAD`.

Hopefully, this also fixes the memory leak #4270.

PR #4410, opened 2025-08-28
DBM: simplify memory pool and resize chunk within critical section

### LINKED ISSUE #4270: Possible memory leak in DBM (GPU)
I've been revisiting the RI-HFXk method to improve it's memory footprint and performance. The method heavily relies on DBT and DBM.

While debugging on a GPU system (GH200), I have noticed a steady increase of memory usage across SCF steps, sometimes killing the run with OOM events. This does not happen on CPU only machines, or on GH200 with CP2K compiled without GPU support (steady memory usage).  The rate of memory usage increase is proportional to the number of OMP threads.

I have been playing with DBM [hyperparameters](https://github.com/cp2k/cp2k/blob/master/src/dbm/dbm_hyperparams.h#L13C1-L14C27), and noticed that the issue only occurs when both `DBM_MEMPOOL_DEVICE` and `DBM_MEMPOOL_HOST` are set to 1.

If I set:

```
#define DBM_MEMPOOL_DEVICE 0
#define DBM_MEMPOOL_HOST 1
```

or

```
#define DBM_MEMPOOL_DEVICE 1
#define DBM_MEMPOOL_HOST 0
```

memory usage remains steady during the run. Performance wise, the second option is much better and matches the default settings (as far as I can tell). Should we disable `DBM_MEMPOOL_HOST` by default whenever compiling with GPU support?

Note: I have only tested this on GH200 (4 GPUs per node), I cannot exclude this being a platform specific bug yet.

hfp:
- Previous objective memory allocation contending the section for a long time.
- This may resolve #4270 (ARM has weaker ordering on regular LD/ST, etc).
- Resizing chunk since outside of critical was perhaps x86-only.
- Use preincrement in face of atomic instructions.
- Atomic postincrement is more costly.
- Subsumes #4336.

### DIFF
--- a/src/dbm/dbm_mempool.c
+++ b/src/dbm/dbm_mempool.c
@@ -75,10 +75,10 @@ static void *actual_malloc(const size_t size, const bool on_device) {
   // Update statistics.
   if (on_device) {
 #pragma omp atomic
-    device_malloc_counter++;
+    ++device_malloc_counter;
   } else {
 #pragma omp atomic
-    host_malloc_counter++;
+    ++host_malloc_counter;
   }
 
   assert(memory != NULL);
@@ -122,7 +122,7 @@ static void *internal_mempool_malloc(dbm_mempool_t *pool, const size_t size,
 
 #pragma omp critical(dbm_mempool_modify)
   {
-    // Find a possible chucks to reuse or reclaim in available list.
+    // Find a possible chunk to reuse or reclaim in available list.
     dbm_memchunk_t **reuse = NULL, **reclaim = NULL; // ** for easy list removal
     dbm_memchunk_t **indirect = &pool->available_head;
     while (*indirect != NULL) {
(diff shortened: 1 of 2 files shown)


src/dbm/dbm_mempool.c
```diff
@@ -122,7 +122,7 @@ static void *internal_mempool_malloc(dbm_mempool_t *pool, const size_t size,
 
 #pragma omp critical(dbm_mempool_modify)
   {
-    // Find a possible chucks to reuse or reclaim in available list.
+    // Find a possible chunks to reuse or reclaim in available list.
```
oschuett: singular: chunks -> chunk

oschuett (commented):
> - This may resolve https://github.com/cp2k/cp2k/issues/4270 (ARM has weaker > ordering on regular LD/ST, etc).

Oh, I didn't know this. I wonder where else this is going to bite us.
Thanks a lot for picking this up!

hfp:
One (unrelated) comment, debugging https://github.com/cp2k/cp2k/pull/4336#issuecomment-3182648149 can run out of stack space in case of no optimizations (aka `-O0`). This happens because, e.g., `neighbor_list_3c_iterate` (or related recursive functions) can go very deep for large iteration spaces. Such recursive functions can potentially be modified to be iterative at certain levels of the iteration space and only recurse otherwise.

src/dbm/dbm_mempool.c
```diff
@@ -41,290 +30,257 @@ typedef struct dbm_memchunk {
...
-    if (NULL != reuse) {
+    // Select an existing chunk or allocate a new one.
+    if (reuse != NULL) {
+      // Reusing an exising chunk that's already large enought.
       chunk = *reuse;
-      *reuse = chunk->next;
-    } else { // Allocate a new chunk.
+      *reuse = chunk->next; // remove chunk from available list.
+    } else if (reclaim != NULL) {
+      // Reclaiming an existing chunk (resize will happen outside crit. region).
```
oschuett: This comment also needs to be updated.

PR #4419, opened 2025-09-02
Fixed short-cycled condition

hfp:


hfp:
A series of similar issues was discovered multiple times. I may ask compiler folks to work on warning when relying on short-cycled IF-conditions. In Fortran, IF-conditions are not short-cycled by standard. Apparently, GNU Fortran short-cycles (like C languages) and thereby hides/avoids related issues.

PR #4420, opened 2025-09-02
CNEO documentation

zc62:
CNEO-DFT documentation as promised.

I am unsure about the order in `dft/index.md`. I am okay with any order.

A little concern is I had to use ```` ```math ```` to render `\{\}` on GitHub. If the latex to html engine works fine for `\{\}` but does not recognize ```` ```math ````, I can change it to `$$`.

(The discussion here should not go into the commit message. My own commit message has only the title line `CNEO documentation` but any elaboration is welcome.)

### DIFF
--- a/docs/methods/dft/cneo.md
+++ b/docs/methods/dft/cneo.md
@@ -1,0 +1,157 @@
+# Constrained Nuclear-Electronic Orbital DFT
+
+## Introduction
+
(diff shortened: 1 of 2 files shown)


zc62:
I see sphinx does not recognize `math` so I am changing to `$$` and see if the output is correct. Ignoring GitHub markdown rendering issue.

PR #4421, opened 2025-09-04
DBM/OFFLOAD: prepared for hybrid execution

hfp:
- Split backend_upload_packs into two phases on GPU.
- Initial phase is chosen to have not triggered upload.
- Offload: offloadEventQuery to check if GPU is busy.
- Removed dbm_multiply_gpu_download_results
- Transfers are launched as early as possible.
- Improved overlap between transfer/compute.
- Idea: proceed on CPU if GPU is busy.

Other changes:
- Documented way to perform validation (ctx=NULL).
- Permit DBM_ prefix for preprocessor symbols.
- Removed C-shard pointer from GPU-context.
- Per-shard event stored/provided by shard.
- Migrated CPU-options into context.
- Miniapp: allow to omit validation.
- Use preincrement (atomic).


hfp:
To try it, two lines can be changed:

* https://github.com/cp2k/cp2k/blob/7face5689eecea11e162328edf493523d06fda3a/src/dbm/dbm_multiply_gpu.c#L90
* https://github.com/cp2k/cp2k/blob/7face5689eecea11e162328edf493523d06fda3a/src/dbm/dbm_multiply.c#L372

It's approx. 95% done except of validation fails due to not combining results correctly. Also the idea using `offloadEventQuery` to check if transfers are contended is not necessarily the best. Usually, hybrid execution needs a heuristic to balance work.

hfp:
The PR was merged because it's almost neutral in LOC measures and there are related changes that can be beneficial.

PR #4427, opened 2025-09-09
VIB| Bugfix for force constant

BelizSertcan:
Fixes [the issue](https://groups.google.com/g/cp2k/c/cZAcjwo-TxA/m/yzsjWAEWBQAJ) reported on google groups

### DIFF
--- a/src/common/physcon.F
+++ b/src/common/physcon.F
@@ -77,7 +77,7 @@ MODULE physcon
    PUBLIC :: a_bohr, a_fine, a_mass, angstrom, atm, bar, bohr, boltzmann, c_light_au, &
              debye, e_charge, e_gfactor, e_mass, evolt, femtoseconds, h_bar, &
              hertz, joule, kcalmol, kelvin, kjmol, massunit, mu_perm, n_avogadro, newton, &
-             p_mass, pascal, picoseconds, seconds, vibfac, wavenumbers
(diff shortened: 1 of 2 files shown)


BelizSertcan:
`make pretty` did not catch the change in physcon.F

hfp:
Ready?

hfp:
Uoops, I rebased the PR instead of squashing both commits.

oschuett:
Thanks for fixing this. Should we maybe also add a test?

BelizSertcan:
> Thanks for fixing this. Should we maybe also add a test?

Should I make a matcher for force constant?

oschuett:
> Should I make a matcher for force constant?

Yes, that would be awesome. And feel free to pick a more descriptive name than `M128` ;-)

BelizSertcan:
> > Should I make a matcher for force constant?
> 
> Yes, that would be awesome. And feel free to pick a more descriptive name than `M128` ;-)

Are there any plans to give the matchers reasonable and descriptive names?

oschuett:
> Are there any plans to give the matchers reasonable and descriptive names?

Yes, I was hoping that over time the community would reassigned descriptive names to all the matchers. It usually requires a bit of digging to figure out what a machter actually does to then pick a good name.

The actual search-and-replace can done with `sed`. I left an example command in the [README](https://github.com/cp2k/cp2k/blob/master/tests/README.md).

PR #4428, opened 2025-09-10
DBM: reduced consuming host/offload aka page-locked memory

hfp:
- Replaced DBM_OVERCOMMIT_DEVICE/HOST with DBM_ALLOCATION_FACTOR-
- Thereby reduce overcommit factor for host (scarce resource).
- Comm related routines to rely on MPI's malloc.
  - Even if buffer is subsequently transferred (GPU-transfer).
  - Because dbm_mempool_host_malloc is delicate resource.


oschuett:
Looking at our [cuda benchmark](https://dashboard.cp2k.org/archive/perf-cuda-volta/index.html) I noticed that the timings for `dbm_multiply` went up by 50% after this PR.

hfp:
I noticed this too, but I was not sure how this applies to other/larger workloads. I will investigate on Alps, and first try to reproduce and see what the minimal change would be. I checked the performance impact of this PR for GW-workloads, and it was no impact or certain WLs were able to run on less resources like reduced consumption of page-locked memory.

hfp:
Can you reproduce this regression on your own system?

I may try the Docker images, but so far I have not figured out the cause of this. Whether it is contended MPI memory allocation in OpenMP regions like a specific MPI flavor/version or perhaps only present in older CUDA versions (with current Dashboard this seems to be ruled out like CUDA 12.x still showing the regression). That aside, I used A100/H100 systems and non-Docker setup but not seeing any difference "before&after".

PR #4429, opened 2025-09-10
Renamed oneAPI testers and fixed/suppressed issues

hfp:
- Distinction is IFORT vs IFX or "classic vs new/next-gen".
- Intel Fortran Classic is available up to and incl. 2024.2 packages.
- Intel Fortran (next-gen) is default (still not ready for CP2K).
- Fix bug about Autoconf spilling -loopopt option into link list.
- Drop names derived from packaging images like "hpckit".
- Renamed Dockerfiles for tests and internal names.
- Suppress "libint" flag until regtests pass.
- Introduced "ifx" flag and suppress tests.

### DIFF
--- a/src/cp2k_info.F
+++ b/src/cp2k_info.F
@@ -93,7 +93,12 @@ CONTAINS
       END IF
 
 !$    flags = TRIM(flags)//" omp"
-#if defined(__LIBINT)
+
+      ! TODO: remove __INTEL_LLVM_COMPILER conditions (regtests)
+#if defined(__INTEL_LLVM_COMPILER)
+      flags = TRIM(flags)//" ifx"
+#endif
+#if defined(__LIBINT) && !defined(__INTEL_LLVM_COMPILER)
       flags = TRIM(flags)//" libint"
 #endif
 #if defined(__FFTW3)
@@ -129,35 +134,27 @@ CONTAINS
 #if defined(__COSMA)
       flags = TRIM(flags)//" cosma"
 #endif
-
 #if defined(__QUIP)
       flags = TRIM(flags)//" quip"
 #endif
-
 #if defined(__HAS_PATCHED_CUFFT_70)
       flags = TRIM(flags)//" patched_cufft_70"
 #endif
-
 #if defined(__ACE)
       flags = TRIM(flags)//" ace"
 #endif
-
 #if defined(__DEEPMD)
       flags = TRIM(flags)//" deepmd"
 #endif
-
 #if defined(__PW_FPGA)
       flags = TRIM(flags)//" pw_fpga"
 #endif
-
 #if defined(__PW_FPGA_SP)
       flags = TRIM(flags)//" pw_fpga_sp"
 #endif
-
 #if defined(__LIBXSMM)
       flags = TRIM(flags)//" xsmm"
 #endif
-
 #if defined(__CRAY_PM_ACCEL_ENERGY)
       flags = TRIM(flags)//" cray_pm_accel_energy"
 #endif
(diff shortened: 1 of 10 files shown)


hfp:
Updated IFX image shows some more breakage during compilation. I will investigate it.

oschuett:
Thanks for taking a look at the ifx tests.

If there are still failing tests after this update, then I'd propose we skip those directories so that we at least notice when more tests start failing.

Furthermore, these tests would also need to be migrated to CMake....

hfp:
> Updated IFX image shows some more breakage during compilation. I will investigate it.

This seems an old issue even with recent Autoconf (reintroduced?). It's a bug about Autoconf spilling -loopopt option into the link-list of the Fortran Makefile.

I have fixed this in `tools/toolchain/scripts/stage3/install_libint.sh` and thereby enabled FMA. Though, compiling LIBINT 2.6 is obsolescent or future deprecation of the toolchain. In any case, LIBINT shall use FMAs given I changed the upstream Spack recipe.

hfp:
> If there are still failing tests after this update, then I'd propose we skip those directories so that we at least notice when more tests start failing.

Perhaps an option, and for sure there are still failing cases. I am working on resolving them. Also, there was progress made at compiler level. Some of the related issues are also relevant for GNU. I will wrap it up too like the Workshare thing (PR).

> Furthermore, these tests would also need to be migrated to CMake....

I will do this incl. any fix-up not yet done at the level of the CMakeFile(s).

tools/dashboard/dashboard.conf
```diff
@@ -225,28 +225,28 @@ report_url:  https://storage.googleapis.com/cp2k-ci/dashboard_hip-rocm-build_rep
 
 [intel-oneapi-hpckit-ssmp]
```
oschuett: For consistency you could also change the internal names e.g.:

- `intel-oneapi-hpckit-ssmp` -> `intel-ifx-ssmp`
- `intel-ssmp` -> `intel-ifort-ssmp`

Just make sure to keep dashboard and CI in sync. I'll then create a symlink for the dashboard archive.

hfp:
Four directories with `RUNTIME FAIL` were listed in `TEST_DIRS` using `!ifx` relying on the new flag introduced by this PR: `QS/regtest-debug-5`, `QS/regtest-debug-6`, `QS/regtest-tddfpt-force-gapw`, and `QS/regtest-hfx`.

hfp:
The "WRONG RESULT" tests are also flagged for the time being. The hope is, they disappear once runtime errors are gone or can be adjusted with confidence (error margin).

tools/dashboard/dashboard.conf
hfp: This is now completed as well (force-pushed next revision).

hfp:
I will finally also trigger the "Intel oneAPI (ssmp, ifx)" test and hopefully the failures and wrong results are already covered like suppressed by the means of the PSMP version.

hfp:
As a side note, the updated IFX-SSMP can now test https://github.com/cp2k/cp2k/pull/4365 (MKL 2025 is needed).

hfp:
It is observed, all workloads suppressed for `ifx` are related to `libint`.

hfp:
The next revision of the PR (if tests still fail/wrong) will drop the `ifx` flag and lie about the `libint` flag like not advertising it in case of IFX.

src/cp2k_info.F
```diff
@@ -93,7 +93,12 @@ FUNCTION cp2k_flags() RESULT(flags)
       END IF
 
 !$    flags = TRIM(flags)//" omp"
-#if defined(__LIBINT)
+
+      ! TODO: remove __INTEL_LLVM_COMPILER conditions (regtests)
+#if defined(__INTEL_LLVM_COMPILER)
+      flags = TRIM(flags)//" ifx"
+#endif
+#if defined(__LIBINT) && !defined(__INTEL_LLVM_COMPILER)
```
oschuett, line 101: Why would you hide the `libint` flag for ifx?

hfp: All crashes or wrong results were suddenly related to tests carrying the `libint` flag. Also, errors and crashes fluctuated among the `libint` labeled test directories. Now, the "rule" for IFX is everything with `libint` plus a few labeled with `ifx`. I really hope this is short-lived as things resolve.

hfp: I think already one of the `!ifx` labeled tests (only a few) vanished because of https://github.com/cp2k/cp2k/pull/4434.

hfp: Btw, the breakage of IFX with LIBINT does not mean it's miscompiled LIBINT; it might be just the code-path passes an issue. I already have some OOP code we use in CP2K, which IFX does not like.

oschuett: I'm fine with excluding all libint dirs from the ifx tests. However, I'd prefer if we did this explicitly in `TEST_DIRS` instead of having a binary report wrong flags.

hfp: Ok, then I will send another PR making this explicit.

hfp: > Btw, the breakage of IFX with LIBINT does not mean it's miscompiled LIBINT; it might be just the code-path passes an issue. I already have some OOP code we use in CP2K, which IFX does not like.

https://github.com/cp2k/cp2k/blob/eab93308cbe21c14bedb4be9d6697c2daf72b597/src/dbt/dbt_types.F#L751-L752

PR #4430, opened 2025-09-10
tests for force constant

BelizSertcan:


tests/1e-05,
oschuett, line 1: This file should probably not be here. Otherwise, this PR is ready to go!

PR #4431, opened 2025-09-11
OpenMP named critical sections and nowait

hfp:
- Named criticals in an attempt to slightly reduce contention (qs_tensors.F).
- Rely on NOWAIT to omit implicit barriers when unnecessary.


hfp:
Seems to be broken with GCC 8..12 according to the Dashboard. A fix will be prepared.

PR #4435, opened 2025-09-12
Avoid INTENT(IN) with ALLOCATABLE and rely on CONTIGUOUS instead

hfp:
Interesting read: https://fortran-lang.discourse.group/t/just-say-no-to-non-default-lower-bounds/6108/4

### DIFF
--- a/src/mp2_gpw_method.F
+++ b/src/mp2_gpw_method.F
@@ -636,7 +636,7 @@ CONTAINS
       TYPE(cp_fm_type), INTENT(IN)                       :: fm_BIb_jb
       REAL(KIND=dp), DIMENSION(:, :), INTENT(OUT)        :: BIb_jb
       INTEGER, INTENT(IN)                                :: max_row_col_local
-      INTEGER, ALLOCATABLE, DIMENSION(:, :), INTENT(IN)  :: local_col_row_info
+      INTEGER, CONTIGUOUS, DIMENSION(:, :), INTENT(IN)   :: local_col_row_info
       INTEGER, INTENT(IN)                                :: my_B_virtual_end, my_B_virtual_start
 
       INTEGER                                            :: i_global, iiB, j_global, jjB, ncol_rec, &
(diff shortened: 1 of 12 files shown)


hfp:
This PR made the code not working anymore since the ALLOCATABLE preserves the lower bound even of INTENT(IN), which CONTIGUOUS does not. This lets me think IFX may not honor ALLOCATABLE INTENT(IN) the same way as GNU. IFX based binaries are crashing like if GNU gets CONTIGUOUS. Bad enough, the related code here seems to depend on preserving array bounds (see https://fortran-lang.discourse.group/t/just-say-no-to-non-default-lower-bounds/6108/4).

oschuett:
While I agree that non-default bounds are error-prone, I don't think we should replace `ALLOCATABLE` with `CONTIGUOUS`.

The problem with `CONTIGUOUS` arguments is that the compiler will create a temporary copy if a non-continuous array is passed, which in turn can lead to use-after-free bugs. I encountered this problem e.g. with [torch_tensor_from_array](https://github.com/cp2k/cp2k/blob/0446f17dce5019af245b6797d450ba78f1aebfc0/src/torch_api.F#L89).

hfp:
> While I agree that non-default bounds are error-prone, I don't think we should replace `ALLOCATABLE` with `CONTIGUOUS`.

I thought ALLOCATABLE implies CONTIGUOUS, but I can be wrong. The only motivation was to have the "least downgrade". I wanted to mainly get rid of the combination ALLOCATABLE and INTENT(IN) which looked cumbersome to me. I learned it's *the* way to preserve the bounds of the incoming array...

> The problem with `CONTIGUOUS` arguments is that the compiler will create a temporary copy if a non-continuous array is passed, which in turn can lead to use-after-free bugs. I encountered this problem e.g. with [torch_tensor_from_array](https://github.com/cp2k/cp2k/blob/0446f17dce5019af245b6797d450ba78f1aebfc0/src/torch_api.F#L89).

I noticed this line ;-)  Still, somewhat "esoteric" behavior one or the other way.

oschuett:
> I thought ALLOCATABLE implies CONTIGUOUS...

Yes, ALLOCATABLES are always contiguous. The only place one cares about this though is when handing the pointer over to C. Usually we then make a copy of the data on the C side, but when the array is large then we just hold on to the pointer instead. This is obviously where a temporary copy snuck in by the compiler can really ruin the day.

> I wanted to mainly get rid of the combination ALLOCATABLE and INTENT(IN) which looked cumbersome to me.

Have you tried to just remove the `INTENT(IN)` ?

> I learned it's the way to preserve the bounds of the incoming array...

Yes, we should avoid non-default bounds. I'm not sure we can enforces it via the convention checker, but we can certainly remove it when we come across it.

PR #4436, opened 2025-09-12
Make suppressed tests explicit

hfp:
- See #4429.


oschuett:
Thanks for this followup. It's awesome that the Intel tests on the dashboard are now running through again!

PR #4437, opened 2025-09-12
Fix molecule indexing when using MULTIPLE_UNIT_CELL

mkrack:
Revise forcefield output

### DIFF
--- a/src/nnp_acsf.F
+++ b/src/nnp_acsf.F
@@ -71,7 +71,7 @@ CONTAINS
       REAL(KIND=dp), ALLOCATABLE, DIMENSION(:, :)        :: forcetmp
       REAL(KIND=dp), ALLOCATABLE, DIMENSION(:, :, :)     :: force3tmp
       REAL(KIND=dp), DIMENSION(3)                        :: rvect1, rvect2, rvect3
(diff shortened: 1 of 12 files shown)


src/nnp_acsf.F
```diff
@@ -71,7 +71,7 @@ SUBROUTINE nnp_calc_acsf(nnp, i, dsymdxyz, stress)
       REAL(KIND=dp), ALLOCATABLE, DIMENSION(:, :)        :: forcetmp
       REAL(KIND=dp), ALLOCATABLE, DIMENSION(:, :, :)     :: force3tmp
       REAL(KIND=dp), DIMENSION(3)                        :: rvect1, rvect2, rvect3
-      TYPE(nnp_neighbor_type)                            :: neighbor
+      TYPE(nnp_neighbor_type)                            :: neighbor = nnp_neighbor_type()
```
fstein93, line 74: I think the initialization should go to a separate line. Otherwise, `neighbor` becomes implicitly a static variable.

mkrack: @fstein93 Thanks for the feedback. I will move the initialisation to a separate in my next PR.

PR #4443, opened 2025-09-18
Follow-up on #4414

hfp:
- No assertions with side-effects (even if NDEBUG is never defined).

### DIFF
--- a/src/grid/grid_replay.c
+++ b/src/grid/grid_replay.c
@@ -44,7 +44,9 @@ static int parse_int(const char key[], FILE *fp) {
   char line[100], format[100];
   read_next_line(line, sizeof(line), fp);
   snprintf(format, sizeof(format), "%s %%i", key);
-  assert(sscanf(line, format, &value) == 1);
+  if (sscanf(line, format, &value) != 1) {
+    assert(!"parse_int failed");
+  }
   return value;
 }
 
@@ -56,7 +58,9 @@ static void parse_int3(const char key[], FILE *fp, int vec[3]) {
   char line[100], format[100];
   read_next_line(line, sizeof(line), fp);
   snprintf(format, sizeof(format), "%s %%i %%i %%i", key);
-  assert(sscanf(line, format, &vec[0], &vec[1], &vec[2]) == 3);
+  if (sscanf(line, format, &vec[0], &vec[1], &vec[2]) != 3) {
+    assert(!"parse_int3 failed");
+  }
 }
 
 /*******************************************************************************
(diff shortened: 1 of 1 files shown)


src/grid/grid_replay.c
```diff
@@ -79,8 +85,10 @@ static double parse_double(const char key[], FILE *fp) {
 static void parse_double3(const char key[], FILE *fp, double vec[3]) {
   char line[100], format[100];
   read_next_line(line, sizeof(line), fp);
-  snprintf(format, sizeof(format), "%s %%le %%le %%le", key);
-  assert(sscanf(line, format, &vec[0], &vec[1], &vec[2]) == 3);
+  snprintf(format, sizeof(format), "%s %%f %%f %%f", key);
```
oschuett: While admittedly `scanf` also accepts scientific notation for `%f`, we do need the `l` modifier because we're reading doubles.

hfp: Yes. I changed the title. It's not about incompatible format specifiers. I may try on a Mac system at hand and play around a bit to see if the issue can be avoided.

hfp: Unfortunately, we do not seem to have `macOS on Apple M1 (psmp)` in our list of tests (like Dashboard only).

I am on a x86-64 macOS, and the issue seems to be not present there (beside of the OS/SDK/CLI not bleeding edge aka regression). I have to find an M1/M2/M3 mac...

hfp:
Let's merge it and get rid of the side-effects. I doubt going from `e` to `f` helps in this case; so might be still red...

hfp:
Ok, it failed again ([Dashboard](https://www.cp2k.org/static/regtest/trunk/Darwin-gnu-arm64/latest-openmpi-psmp/regtest-0)), however revisiting the conditions, it seems brittle results are asserted. For instance asserting `sscanf(line, format, &value) == 1` looks like it intends to check for one value/integer (`%i`), but the actual return value is about the *number of characters*.

oschuett:
> but the actual return value is about the number of characters.

Are you sure? [Here](https://cplusplus.com/reference/cstdio/sscanf/) it says otherwise:

_On success, the function returns the number of items in the argument list successfully filled._

hfp:
No, I was wrong. I looked at the wrong reference. Anyhow, I think we need an M-mac system for debug or get one into the GitHub list of CI-targets.

oschuett:
> we need an M-mac system

We do have the machine from MacStadium that is also running the dashboard tests. It would be nice to also have it in the cp2k-ci, but it's complicated because containers don't really work on macOS. In any case, you can ask Matthias for access.

PR #4445, opened 2025-09-18
Sternheimer BSE (1. Version with W matrix given)

annahehn:


tests/QS/regtest-tddfpt-bse/TEST_FILES.toml
```diff
@@ -0,0 +1,7 @@
+# runs are executed in the same order as in this file
```
oschuett, line 1: Please add this new test directory to the [TEST_DIRS](https://github.com/cp2k/cp2k/blob/master/tests/TEST_DIRS) file.

annahehn: Yes, I forgot, thank you Ole!

hfp:
It's hard to see at the moment (all red Dashboard), but I think [this](https://storage.googleapis.com/cp2k-ci/dashboard_pdbg_report.txt) is related to the PR like Out-Of-Bounds accesses (OOBs):

> At line 299 of file /opt/cp2k/src/qs_tddfpt2_bse_utils.F
Fortran runtime error: Index '12' of dimension 1 of array 'csvirt%local_data' above upper bound of 11
At line 298 of file /opt/cp2k/src/qs_tddfpt2_bse_utils.F
Fortran runtime error: Index '9' of dimension 1 of array 'wxvirtao%local_data' above upper bound of 8

oschuett:
@annahehn, could you please take a look at the failing `regtest-tddfpt-bse/H2O.inp` test.

hfp:
The code somehow seems to mess up global/local indexes/bounds to some extent (this is only a very high-level first look). Perhaps it was not tested thoroughly with PSMP.

oschuett:
Yes, there a few places where `%local_data` is being accessed, which is always error prone. Ideally, high-level routines from the fm package would be called instead.

@annahehn is already working on this and I believe @fstein93 is helping. However, since it's now been ~over a month~ three weeks, we should maybe consider a rollback to recover the dashboard.

hfp:
Well, the branch/code is not lost if reverted. It's however an inconvenience to reapply with a new PR. In any case, it's a new PR and I am leaning towards a rollback if there are no conflicts or additional work for the person who reverts it. The current CI or Dashboard state seems degenerated like no clear visibility anymore if current/later contributions are valid or not. That's fine, I think, if we had no contributions or upcoming release. Depending on the timeline (far release or resolution), I would leave it or not.

fstein93:
I am working on fixing the bugs.

fstein93:
I do have an idea on how to fix it but it will take a bit of time to mimic the intended tensor contraction. I am currently in exchange with @annahehn on how to proceed.

PR #4451, opened 2025-09-23
CI: improve Time to Solution

hfp:
- TTS on Alps using CPU only (single node) to be improved.
- The workload is best with square number for rank-count.

### DIFF
--- a/ci/cscs_daint_spack_psmp.yaml
+++ b/ci/cscs_daint_spack_psmp.yaml
@@ -102,14 +102,13 @@ benchmark cp2k 1 daint node:
   script:
     - cp2k /opt/cp2k/benchmarks/CI/H2O-512_md.inp
   variables:
-    SLURM_CPU_BIND: cores
-    SLURM_CPUS_PER_TASK: 1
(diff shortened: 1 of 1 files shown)


mkrack:
I am not sure, if it is a good idea to change the `CPU_BIND` from `cores` to `sockets`. I would also keep the explicit `MPI_TYPE` specification `pmi2` instead of relying on the default one even if it might be the same currently.

hfp:
I tested CPU_BIND=sockets since then. Though, it was suggested by CSCS Documentaiont for ALPS and CP2K in particular (https://docs.cscs.ch/software/sciapps/cp2k/#running-on-daint). I can change it back once the PR passes for the first time.

In fact, I would like to avoid binding/pinning procs to CPU cores in a consecutive fashion (mainly because of OpenMP). I think the underlying MPI implementation has a default scheme for placing the procs (once forced to a socket ;-).

oschuett:
I'm going to close this PR as it seems to have fallen dormant.

hfp:
No problem!  I may come up again with it (when ready).

PR #4464, opened 2025-10-01
Updated real-valued diagonalization

hfp:
- Introduced cp_fm_error and avoided intermediate allocations.
- Query workspace for PDSYEVD via PDORMTR (workaround).
- Introduced __SCALAPACK_NO_WA (disable workaround/s).
- Cleanup workaround (cp_fm_syevd_base).
- Cleanup (cp_fm_cholesky.F).


hfp:
For PDBG, only `tests/QS/regtest-tddfpt-bse/H2O.inp` fails (#4445).

PR #4467, opened 2025-10-02
Resolve CUDA performance regression

hfp:
- Tried OpenMPI vs MPICH in several version flavors.
- Tried Docker "to reproduce exactly".
- Tried Spack and CUDA 11.8.
- Related to #4428.

### DIFF
--- a/tools/docker/Dockerfile.test_cuda_A100
+++ b/tools/docker/Dockerfile.test_cuda_A100
@@ -3,7 +3,7 @@
 # Usage: podman build --shm-size=1g -f ./Dockerfile.test_cuda_A100 ../../
 #
 
-FROM nvidia/cuda:11.8.0-devel-ubuntu22.04
+FROM nvidia/cuda:13.0.1-devel-ubuntu24.04
 
 # Setup CUDA environment.
 ENV CUDA_PATH /usr/local/cuda
--- a/tools/docker/Dockerfile.test_cuda_P100
+++ b/tools/docker/Dockerfile.test_cuda_P100
@@ -3,7 +3,7 @@
 # Usage: podman build --shm-size=1g -f ./Dockerfile.test_cuda_P100 ../../
 #
 
-FROM nvidia/cuda:11.8.0-devel-ubuntu22.04
+FROM nvidia/cuda:13.0.1-devel-ubuntu24.04
 
 # Setup CUDA environment.
 ENV CUDA_PATH /usr/local/cuda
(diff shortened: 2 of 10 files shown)


hfp:
Let me merge this PR anyway as it is a reasonable update jumping two major versions (11->13). Also, there are signs the mentioned issue is not easily reproducible otherwise.

oschuett:
This change [broke the CUDA build](https://dashboard.cp2k.org/archive/cuda-pascal/commit_b396de2029893bdd02783a52e14ca541ec8cb85b.txt) giving the surprising error message:

```
nvcc fatal   : Unsupported gpu architecture 'sm_60'
```

hfp:
Updating to CUDA 13 seems to cause issues (e.g., `sm_60`). I will try to replace it with `-arch=all`. See [here](https://dashboard.cp2k.org/archive/cuda-pascal/commit_b396de2029893bdd02783a52e14ca541ec8cb85b.txt) when compiling ELPA.

hfp:
CUDA 13 aka `nvcc` has `sm_75` as baseline (just tested it). Also, our infra wants `sm_` and it's no simple way to introduce/try `-arch=all`, i.e., "all" is not "sm_all". In any case, adjusting for it, is a larger change than I anticipated. Going forward, it seems to be inevitable and we may have three options: (1) try `-arch=all` and see if it works before SM75, (2) strip all older ARCHs, or (3) figure out something else aka holy grail.

oschuett:
> CUDA 13 aka nvcc has sm_75 as baseline

Oh [you're right](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#deprecated-architectures). I didn't see this one coming...

Well, I'm still developing on Volta and for our CI we can't really afford anything newer either.

So, I'd suggest we stick to `nvidia/cuda:12.9.1-devel-ubuntu24.04` for now.

hfp:
My weak attempt for a minimal change: #4469. Let me try a few iterations and see how it goes. If it does not go well, I will revert to CUDA 12 for the CI images targeting older architectures.

oschuett:
> (1) try -arch=all and see if it works before SM75,

Worth a try, but reading NVIDIA's announcement I'm not hopeful at all:

"...newer toolkits will be unable to target these architectures."

hfp:
PR #4469 changes the image to `nvidia/cuda:12.9.1-devel-ubuntu24.04`.

PR #4469, opened 2025-10-02
Account for baseline architecture

hfp:
- Rely on CUDA 12 instead of CUDA 13.
- Avoid missing case labels (#4462).

PR #4475, opened 2025-10-06
Updated to SIRIUS 7.7.1

hfp:
- See https://github.com/cp2k/cp2k/pull/4469#issuecomment-3361694889

### DIFF
--- a/tools/toolchain/scripts/stage8/install_sirius.sh
+++ b/tools/toolchain/scripts/stage8/install_sirius.sh
@@ -6,8 +6,8 @@
 [ "${BASH_SOURCE[0]}" ] && SCRIPT_NAME="${BASH_SOURCE[0]}" || SCRIPT_NAME=$0
 SCRIPT_DIR="$(cd "$(dirname "$SCRIPT_NAME")/.." && pwd -P)"
 
-sirius_ver="7.7.0"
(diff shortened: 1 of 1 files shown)


hfp:
DFTD4 still seems to suffer from fetching content and unavailable download locations. If of interest, I once implemented:
> if ! download_pkg_from_cp2k_org "${libxsmm_sha256}" "libxsmm-${libxsmm_ver}.tar.gz"; then
  download_pkg_from_urlpath "${libxsmm_sha256}" "${libxsmm_ver}.tar.gz" \
  https://github.com/libxsmm/libxsmm/archive \
  "libxsmm-${libxsmm_ver}.tar.gz"
fi

However, the issue seems to be using CMake's fetch content...

oschuett:
> DFTD4 still seems to suffer from fetching content and unavailable download locations. 

Yes, this is #4362. We are [waiting](https://github.com/cp2k/cp2k/issues/4362#issuecomment-3200032627) for a new tblite release, which in turn waits for a new dftd4 release, which happens roughly once per year.

Since this is starting to significantly hamper our development, maybe it's time to temporarily disable tblite in the toolchain?

hfp:
> Since this is starting to significantly hamper our development, maybe it's time to temporarily disable tblite in the toolchain?

It's perhaps as easy as changing the default of "build all dependencies", i.e., someone working with mentioned deps can still build a CP2K variant enabling such bits.

PR #4477, opened 2025-10-06
Adjusted print_func similar to Offload and DBM library

hfp:
- Print grid statistics (only if non-zero counter exists).
- Removed dependency from string_utilities (strlcpy_c2f).
- No copy of the message is necessary.
- Reduced PACKAGE dependencies.
- Print offload memory stats.

### DIFF
--- a/src/grid/common/PACKAGE
+++ b/src/grid/common/PACKAGE
@@ -1,5 +1,5 @@
 {
     "description": "Common parts shared by the grid backends",
-    "requires": ["../../offload"],
+    "requires": ["../../offload", "../../mpiwrap"],
     "archive": "libcp2kgridcommon",
 }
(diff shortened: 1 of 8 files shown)


oschuett:
I think this PR somehow [broke](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-pascal-996e0b5c_report.txt) the `grid_unittest.c`.

hfp:
> I think this PR somehow [broke](https://storage.googleapis.com/cp2k-ci/run-cp2k-cuda-pascal-996e0b5c_report.txt) the `grid_unittest.c`.

Stemmed from now wanting "no MPI" even if `__parallel` is defined (grid miniapp and unit test do not support MPI). See #4481.

PR #4478, opened 2025-10-06
Don't assume MPI_COMM_WORLD in init_cp2k in f77 API

trdurrant:
Hi cp2k team,

I have been running into a problem where the function `init_cp2k` in the f77 API is assuming that the initialisation is occurring on the `MPI_COMM_WORLD` communicator. This is a problem for my application, where only a subset of MPI ranks are involved in running cp2k.

I have made a minor change to the relevant function to add an optional argument, so that the user can supply an MPI communicator to this function, without the assumption that `MPI_COMM_WORLD` will be used.

Looking at the rest of the API functions, I think this should be following the style of other functions that allow the same thing.

I am not very experienced with cp2k's API, so just let me know if I am using the wrong function or this change doesn't make sense.

All the best,
Tom

### DIFF
--- a/src/f77_interface.F
+++ b/src/f77_interface.F
@@ -212,11 +212,14 @@ CONTAINS
 !>      other functions when using cp2k as library
 !> \param init_mpi if the mpi environment should be initialized
 !> \param ierr returns a number different from 0 if there was an error
+!> \param mpi_comm an existing mpi communicator (if not given mp_comm_world
+!>      will be used)
 !> \author fawzi
 ! **************************************************************************************************
-   SUBROUTINE init_cp2k(init_mpi, ierr)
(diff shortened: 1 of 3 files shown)


oschuett:
I think this PR is generally ok. However, the `f77_interface.F` is deprecated. Please use [libcp2k.F](https://github.com/cp2k/cp2k/blob/master/src/start/libcp2k.F) or [libcp2k.h](https://github.com/cp2k/cp2k/blob/master/src/start/libcp2k.h) instead.

Also note that, while we most of CP2K supports getting reinitialized, [some methods](https://github.com/cp2k/cp2k/blob/master/tests/do_regtest.py#L45) do not.

trdurrant:
Thank you for the information. I wasn't aware that `f77_interface.F` is deprecated, so I will update `libcp2k.F` and `libcp2k.h` to allow this too

PR #4481, opened 2025-10-07
Revised #4477

hfp:
- Enabled some of the MPI-C wrappers to handle MPI_COMM_NULL.
- MPI_COMM_NULL used as if single-rank application.


hfp:
Regtest pdbg passed except for:

> At line 299 of file /opt/cp2k/src/qs_tddfpt2_bse_utils.F
Fortran runtime error: Index '12' of dimension 1 of array 'csvirt%local_data' above upper bound of 11
At line 298 of file /opt/cp2k/src/qs_tddfpt2_bse_utils.F
Fortran runtime error: Index '9' of dimension 1 of array 'wxvirtao%local_data' above upper bound of 8

Issue #3925, opened 2025-01-30
GRID GPU backend extremely slow for forces with  high l quantum numbers


abussy:
We recently ran into an issue on the GH200 architecture with the GRID CUDA backend. The calculation of forces in a MD run was about 30 times more expansive with the default CUDA backend than with the CPU backend.

This behavior is only triggered with high angular momenta. Looking at the `GRID STATISTICS` in the output, the highest l quantum number was 11. After switching the functional from SCAN to PBE (meta-GGA to GGA), the highest reported angular momentum was 9, and timings became normal again.

Note that this issue is not bound to the CUDA backend, it was also reproduced on AMD GPUs with the HIP backend.

This is reproducible by taking the usual `H2O-XX.inp` DFT benchmarks, and changing to larger basis sets and meta-GGA functionals.

oschuett:
This this is a know limitation of the grid GPU code. For `lp >= 7` the performance falls off these two cliffs:

https://github.com/cp2k/cp2k/blob/26096efc000316683be093141185a0bd14d2043f/src/grid/gpu/grid_gpu_integrate.cu#L211

https://github.com/cp2k/cp2k/blob/26096efc000316683be093141185a0bd14d2043f/src/grid/gpu/grid_gpu_collocate.cu#L162

Unfortunately, one can not simply unroll more of the loops because that would increase the [register usage](https://github.com/cp2k/cp2k/tree/master/src/grid#cuda-register-usage) and tank the performance for smaller basis sets. A naive solution would be to introduce specialized kernels for large `lp`, but I think there should also be a smarter way to handle the generic case.

Btw, for very large basis sets you might also run into [this limit](https://github.com/cp2k/cp2k/blob/26096efc000316683be093141185a0bd14d2043f/src/grid/common/grid_library.c#L61), which is a workaround for shared memory constraints https://github.com/cp2k/cp2k/issues/1785.

abussy:
I see. This quite unfortunate, I assume that anyone running into this issue will eventually switch to the CPU backend. From there, it is only a small step from systematically adding
```
&GRID
    BACKEDN CPU
&END GRID
```
for piece of mind.

oschuett:
I agree that the situation is not great, but `lp < 7` does cover most use-cases.

Btw, in the meantime others have started to work on spherical harmonics as well: https://github.com/lab-cosmo/sphericart

juerghutter:
Hi
[...]

PR #4488, opened 2025-10-10
Merge device calls for better performance

fstein93:
This PR improves the grid integration kernel. I measured improvements of ~15 % to grid_integrate_task_list in case of QS/H2O-512 and 25 % in case of hBN.

### DIFF
--- a/src/grid/gpu/grid_gpu_integrate.cu
+++ b/src/grid/gpu/grid_gpu_integrate.cu
@@ -28,8 +28,9 @@
 #error "OpenMP should not be used in .cu files to accommodate HIP."
 #endif
 
-// Teen registers are sufficient to integrate lp <= 2 with a single grid sweep.
-#define GRID_N_CXYZ_REGISTERS 10
+// Twenty registers are sufficient to integrate lp <= 3 with a single grid
+// sweep.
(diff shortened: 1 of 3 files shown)


oschuett:
The number of registers is indeed a crucial parameter for performance. Unfortunately, the optimal value depends very much on the workload and the hardware. Presumably, it would be better to have multiple kernels that are specialized for fixed values of `lp`.

See also #3925 for more ideas on how to improve the GPU backend.

fstein93:
Ok, then it is probably better to rewind my change and wait for a better solution instead.

oschuett:
> Ok, then it is probably better to rewind my change and wait for a better solution instead.

Not necessarily. I tuned those parameters by running DZVP water on a Volta GPU. Both are ~~not~~ kinda outdated by now.

fstein93:
I ran on A100.

oschuett:
Sorry, I miss-typed. I meant that Volta is very much outdated by now (Nvidia just dropped support for it in CUDA 13) and DZVP water is not really a representative workload.

I in particular underestimated how for larger basis sets the timings are completely dominated by the d-functions.

All that is to say that I'd keep your change ;-)

fstein93:
Thank you for the clarification.

PR #4495, opened 2025-10-17
Restore BSD-3 license for C-based MPI-wrapper

hfp:


hfp:
@fstein93 Let me know if anything else is missing. The last we want is missing a contribution because of the license!

I did not implement the "exclusion list" like files falling back to GPL inside of a non-GPL directory, i.e., DBM or GRID API related F-files are still under BSD-3 license. If you want those to be GPL - let me know! So far, we have the most permissive solution implemented, i.e., BSD-3 even if it refines the F-API.

fstein93:
No, we should stick to BSD (or a similar license) as much as possible.

fstein93:
I will be in need of the new MPI-C interface for my code, so it is important that the license on this file (or at least its header file) is compatible with BSD-3.

PR #4496, opened 2025-10-17
Use git ls-files instead of os.walk with pattern

hfp:
- Avoids processing unrelated files.


hfp:
Note, `git ls-files` also lists the staged files like just added by the commit in question (for which the precommit script runs).

PR #4499, opened 2025-10-20
Get the number of distributed systems

hfp:
- Removed unused mp_get_node_global_rank routine.
- Determine and print node count.
- Improved output format.


hfp:
With PSMP, or hybrid execution (MPI and OpenMP), it can be hard to reproduce the execution of a workload. Prior to this PR, there was no information printed (log) about the number of distributed systems (nodes). This was even harder when CPU info was not present/unknown as part of the log.

Issue #4503, opened 2025-10-22
Dashboard: performance regressions


hfp:
Looking at the CUDA Volta benchmarks like https://dashboard.cp2k.org/archive/perf-cuda-volta/index.html, I wonder if we should generally track and denote the total time to solution and potentially improve the graphing of results. If timings redistribute between parts, graphs can be discontinuous let alone cases when timings drop from the profile. For example `diag_cu144_broy_6cpu_1gpu`, it seems "something happened approx. `-40` commits ago (by "today"), in fact, total time to solution regressed from `189.323` to `327.517` seconds (very severe!). This was one example, whereas we are still hunting a regression with `RI-HFX_H2O-32_6cpu_1gpu`, which only changed from `202.556` to `215.077` (`-92` to "today" which is 2025/10/22).

abussy:
I agree, total time should be tracked, and ideally an alarm raised in case of bad regression. This might save a lot of time debugging (e.g. #4373). 

I've also spotted some regressions in the RPA and MP2 benchmarks going from `v2025.1` to `v2025.2`. I have not yet ruled out influence from my system, but when I come up with reproducers, I expect a long session of git bisect. It would be nice to consistently catch such things early.

hfp:
As a note, our Dashboard/CI (when based on cloud), can be non-reproducible or disruptive if we allow "equivalent or better SKUs" to be scheduled. Perhaps, we can also run more performance regressions. For instance, PSMP (non-debug) jobs for Dashboard/CI may take ~30 minutes, and say two additional benchmarks with about five minutes runtime each would not matter much especially if that test does not run often.

oschuett:
>  total time should be tracked

The problem with total time is that it's inherently much larger than the other timings and therefore doesn't fit in the same graph. Hence, I decided to track the `rest` time instead. While it's a bit harder to parse - especially when the top-5 change - I still feel that it's a good compromise.

> ideally an alarm raised in case of bad regression

Yes, that would be nice, but it's not trivial. Currently, I just glance at the graphs ever other week or so, which catches most regressions but obviously not all.

>  For example diag_cu144_broy_6cpu_1gpu, it seems "something happened approx. -40 commits ago (by "today")

Yeah, there is a bit of history:

1. It all started with a change that reduced the use of host locked memory. While favorable on other machines, it slowed down `dbm_multiply` in the dashboard performance test:
  #4428
2. Since we couldn't reproduce it on other machines, we decided to upgrad the outdated base container:
[...]

PR #4504, opened 2025-10-22
Performance: show total time as part if the title

hfp:
- Related to #4503.


hfp:
Totally untested; needs review/confirmation/go...

oschuett:
Nice hack 😃 Some of the titles might overflow, but let's give it a try.

hfp:
I removed "Timings of" since the y-axis is about timing. The title now looks like "Title: 42 s".

hfp:
I will refine it and print the min-max range like "Title: 36..42 s".

PR #4506, opened 2025-10-23
Dashboard: plot range of total timings

hfp:
- Related to #4503
- Improves #4504


oschuett:
Unfortunately, this is not going to work because we also have plots that don't show timings, e.g. [here](https://dashboard.cp2k.org/archive/coverage/index.html).

I think, the best way to improve the dashboard plots is to replace the static images with something like e.g. [Plotly](https://plotly.com/javascript/line-charts/).

hfp:
Ok, I will revert #4504 since it had a typo also.

PR #4509, opened 2025-10-23
Enabled DBM_MULTIPLY_COMM_MEMPOOL

hfp:
- Reproduces #4428 (or not).
- Related to #4503 effort.

### DIFF
--- a/src/dbm/dbm_multiply_comm.c
+++ b/src/dbm/dbm_multiply_comm.c
@@ -12,7 +12,7 @@
 #include <stdlib.h>
 #include <string.h>
 
-#if 0
+#if 1
 #define DBM_MULTIPLY_COMM_MEMPOOL
 #endif
 


hfp:
Since we have "Performance CUDA Volta" in CI, we can collect a little table of performance for every try:

| Benchmark time [s] | Test 1 | Test 2 |
| :----------- | -------------: | -------------: |
| H2O-64_6cpu_1gpu | 102.320 | 102.923 |
| H2O-64_nonortho_6cpu_1gpu | 98.453 | 97.450 |
| GW_PBE_4benzene_6cpu_1gpu | 170.324 | 170.401 |
| RI-HFX_H2O-32_6cpu_1gpu | 196.095 | 201.356 |
| RI-MP2_ammonia_6cpu_1gpu | - | 108.447 |

The "RI-HFX_H2O-32_6cpu_1gpu" benchmark is of particular interest (`dbm_multiply`):

| DBM multiply time [s] | Test 1 | Test 2 |
| :----------- | -------------: | -------------: |
| RI-HFX_H2O-32_6cpu_1gpu | 49.300 | 48.957 |

Baseline for `dbm_multiply` without this PR is about **70 s** (regression) with a total time of about **215 s**. Still, I was not able to reproduce the regression on a number of own systems with A100, H100, and H200 GPUs.

oschuett:
So this PR is essentially a rollback of #4428, right? I think that's the right decision as it'll buy us time to analyze the situation more thoroughly.

While host-pinned memory was always expensive to allocate, on older GPUs it was well-worth it. However, on newer architectures with tighter CPU-GPU integration, like Grace-Hopper, this tradeoff might no longer hold.

So, we're left with the usual three choices:
1. Drop support for older hardware.
2. Leave performance on the table with newer hardware.
3. Support both, which increases complexity and slows down development.

To make matters worth, GPU prices have gone through the roof and many folks are holding on to their old Volta cards.

So again, let's rollback and then start a broader discussion on the topic.

PR #4510, opened 2025-10-24
Let builds succeed

hfp:
- It seems -Wno-error does not cover line truncation warning.
- It seems -ffree-line-length-none is necessary on top.
- Issues came up with Spack/CMake based builds.
- Cleanup/typos.

### DIFF
--- a/cmake/CompilerConfiguration.cmake
+++ b/cmake/CompilerConfiguration.cmake
@@ -35,7 +35,7 @@ endif()
 
 # Baseline
 add_compile_options(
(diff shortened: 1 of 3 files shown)


oschuett:
I'd really like to avoid using `-ffree-line-length-none` again as I just got ride of it. The underlying issue is this:
https://github.com/cp2k/cp2k/blob/2a4e9c4b3c7c69d1149df2321e309cc6af21f99f/src/CMakeLists.txt#L1646

hfp:
Ok, fine. It takes some more/different work then as I need to apply it to unit test files, e.g., `src/gx_ac_unittest.F`.

oschuett:
Let's do #4512 instead.

PR #4523, opened 2025-11-05
precommit: Add Fortitude linter

oschuett:
This add the new [Fortitude](https://github.com/PlasmaFAIR/fortitude/) linter to our precommit tool.

To get started, I added all the rules that are currently violated to the ignore list. Many of them can be fixed automatically with `--unsafe-fixes` but need to be reviewed. Furthermore, a number of files can not be parsed - typically because Fortran statements are inter-leafed with pre-processor macros. So, we can chip away at these issue in the coming weeks.

Huge shoutout to @LiamPattinson and @ZedThree for creating the amazing Fortitude tool!

### DIFF
--- a/src/pw/fft/fftw3_lib.F
+++ b/src/pw/fft/fftw3_lib.F
@@ -1004,7 +1004,7 @@ CONTAINS
       COMPLEX(KIND=dp), DIMENSION(*), INTENT(IN)         :: zout
       INTEGER, INTENT(IN)                                :: plan_style
 #if defined(__FFTW3)
-      INTEGER                                            :: ii, di, io, DO, num_threads, num_rows
+      INTEGER                                            :: ii, di, jj, dj, num_threads, num_rows
 
       INTEGER :: fftw_plan_type
       SELECT CASE (plan_style)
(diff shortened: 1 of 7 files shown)


src/pw/fft/fftw3_lib.F
```diff
@@ -1049,32 +1049,32 @@ SUBROUTINE fftw3_create_plan_1dm(plan, zin, zout, plan_style)
 !$    plan%num_rows = num_rows
       ii = 1
       di = plan%n
-      io = 1
-      DO = plan%n
+      jj = 1
```
fstein93, line 1052: The old variable names made more sense as ii/di refer to Input sizes and leading dimensions whereas io/do indicate the respective output quantities.

fstein93 (commented):
Great idea to incoporate Fortitude! Let's see how it improves (or can improve) CP2K.

src/pw/fft/fftw3_lib.F
oschuett: I changed them because the `DO` was tripping up the Fortitude parser.

fstein93: True. Still, `jj` is not the best new name. Maybe something like `[size|distance]_[in|out]` is better.

oschuett: Ok, I'll make the change.

PR #4524, opened 2025-11-06
Get Fortitude auto-fixing to work and enable some more rules

oschuett:
Follow up to #4523.


mkrack:
Just out of curiosity, is there any reason for keeping a semicolon at the end of a line in Fortran?

oschuett:
> Just out of curiosity, is there any reason for keeping a semicolon at the end of a line in Fortran?

No, there isn't and Fortitude also has a rule for [superfluous semicolons](https://fortitude.readthedocs.io/en/stable/rules/superfluous-semicolon/), but it's still in preview.

For now I've not enabled preview rules because we are still violating a lot of the [stable rules](https://github.com/cp2k/cp2k/blob/master/tools/precommit/fortitude.toml#L6).

As a next step we could enable [old-style-array-literal](https://fortitude.readthedocs.io/en/stable/rules/old-style-array-literal), [deprecated-relational-operator](https://fortitude.readthedocs.io/en/stable/rules/deprecated-relational-operator), and [unnamed-end-statement](https://fortitude.readthedocs.io/en/stable/rules/unnamed-end-statement), which will affect pretty much every Fortran file.

PR #4527, opened 2025-11-07
RT-TDDFT+RT-BSE : Fourier Transform Outputs

StepanMarek:
Following discussion at #4489 (@Anyborr ), this pull request refactors the calculation of Fourier transforms in real-time propagation calculations and allows for printing of polarizability components and Fourier transform of dipole moments in both RT-BSE and RT-TDDFT.

So far, I have only implemented the already present methods for RT-BSE - namely, the printing of the mentioned observables is only available for non-periodic systems.

On a more technical level, I also refactored the FTs, so now the FFTW3 library is used if available.

The input structure changes a little bit - the parameters setting for the Fourier transforms of the observables are now set in a subsection directly in `&REAL_TIME_PROPAGATION` section, instead of subsection of `&RTBSE`, since they apply for both RT-BSE and RT-TDDFT. Specifically, new the subsection `&FT` and its subsection `&PADE` set these parameters, with reasonable defaults.


src/emd/rt_propagation_ft.F
```diff
@@ -0,0 +1,320 @@
...
+         DEALLOCATE (result_series)
+         ALLOCATE (result_series(nseries, nsamples), source=CMPLX(0.0, 0.0, kind=dp))
+      END IF
+
+      ! Calculate the omega series values, ordered from negative to positive
+      IF (PRESENT(omega_series)) THEN
+         CALL fft_freqs(nsamples, t_total, omega_series, fft_ordering_opt=.FALSE.)
+      END IF
+
+#if defined(__FFTW3)
```
fstein93: I think it is easier to check the error code `stat` of the FFT and use the naive implementation as a backfall if necessary. Then, there is no need to pull dependencies on the available libraries into this module.

src/emd/rt_propagation_output.F
```diff
@@ -1037,4 +1074,504 @@ SUBROUTINE print_rtp_energy_components(qs_env, dft_section)
...
+      TYPE(rtp_control_type), POINTER                    :: rtc
+      INTEGER, OPTIONAL                                  :: info_opt
+      TYPE(cell_type), OPTIONAL, POINTER                 :: cell
+
+      CHARACTER(len=11), DIMENSION(2)                    :: file_extensions
+      CHARACTER(len=20), DIMENSION(:), POINTER           :: headers
+      CHARACTER(len=21)                                  :: prefix
+      CHARACTER(len=5)                                   :: prefix_format
+      COMPLEX(kind=dp), DIMENSION(:), POINTER            :: omegas_complex, omegas_pade
+      COMPLEX(kind=dp), DIMENSION(:, :), POINTER :: field_results, field_results_pade, &
```
fstein93: Hint: Most (all?) of these pointers can be downgraded to `ALLOCATABLE, TARGET`. This prevents accidental memory leaks although you deallocate all variables AFAICS.

src/emd/rt_propagation_ft.F
StepanMarek: Thank you for the suggestion, I have removed the preprocessor directives, the routine now uses only the stat code to determine whether the naive implementation should run.

src/emd/rt_propagation_output.F
StepanMarek: I rewrote the local arrays to be `ALLOCATABLE, TARGET`, but due to `multi_fft` requiring pointers, I included additional pointer that point to them.

fstein93: `multi_fft` does not actually need a `POINTER`. In most cases, you can drop it (and probably also in the interface of some other routines) except from its third argument (`result_series`) which you can downgrade to `ALLOCATABLE` because it may be reallocated within `multi_fft`.
Another issue is the planner routine `fft_create_plan_1dm` which should not be called with the actual input array because it may be overwritten with some FFTW planning modes (default in CP2K is FFTW_ESTIMATE) (see [here](https://www.fftw.org/doc/Complex-DFTs.html)). Apparently, this behavior is assumed at several occasions.

StepanMarek: Thank you for the specific suggestions - I rewrote `multi_fft` so that it now accepts the arrays directly and the pointers are therefore reduced. Inside `multi_fft`, I left a few pointers to arrays used in `fft_alloc` - I assume since this is coupling to C arrays, these really have to be pointers.

For the planner routine, I now call `fft_create_plan_1dm` with empty (allocated but not initialized) arrays and `fft_1dm` with additional array, which contains copy of the input data (for the results, I reuse the array used in `fft_create_plan_1dm`). I tested determination of `POLARIZABILITY` with different FFTW plans (`ESTIMATE`, `MEASURE` and `PATIENT`) on a simple system and the results are identical in all cases, so I hope this approach will be valid.

PR #4530, opened 2025-11-09
NEGF: bug correction of f4fc7e3.

dryndyk:


oschuett:
Thanks for the quick follow up!

dryndyk:
I'm sorry for not having enough testing.... ))

PR #4533, opened 2025-11-13
RIXS| Open-Shell

BelizSertcan:
This PR
- Enables RIXS calculations with UKS and ROKS
- Fixes dipole moment calculation (was incorrectly squared)
- Adds excited atom index to RIXS output printing
- Updates XAS excitation energy matcher name
- Fixes memory leak in TDDFT where ao_ao_fm_struct was created but only released when [do_eigen=.TRUE.](https://github.com/cp2k/cp2k/blob/c01e6a4e89d3b4b67a117c4170084206f8bd57c5/src/qs_tddfpt2_utils.F#L333) This leak was found for TDDFT+ROKS, where there was no test coverage.
- Adds test coverage for ROKS+TDDFT


BelizSertcan:
Once in a while, the test [QS/regtest-gapw-4/CO_xastpxfh_150Ry.inp](https://github.com/cp2k/cp2k/blob/c01e6a4e89d3b4b67a117c4170084206f8bd57c5/tests/QS/regtest-gapw-4/CO_xastpxfh_150Ry.inp) was failing on the local machines due to some numerical inconsistency:
 "Difference too large: 4.04e-12 > 4e-12". 

It was consistently calculating `-92.8473936558286` while the reference value is `-92.84739365620344`. Should we maybe reduce the threshold slightly?

hfp:
> It was consistently calculating -92.8473936558286 while the reference value is -92.84739365620344. Should we maybe reduce the threshold slightly?

It is possible to adjust the Gold value or (alternatively) the margin. Calibrating the Gold value is more ambitious because the "correct" value aka Gold is usually taken from some SDBG run and then the margin is widened based on PDBG and others. If you adjust the Gold value, you get risk the "error" returns.

BelizSertcan:
psmp test passes with --mpiranks 1
i haven't found the source of the error

PR #4534, opened 2025-11-13
Cleanup workarounds for T3E and old OMPI

hfp:
- USE statements for `mpi_f08` cannot be relied on (similar reason as for `mpif`).
- Cleanup workaround for unaligned access errors on T3E.
- Avoid unnecessary allocation in key-routine(s).
- Allocations negatively impact scalability.
- Cleanup broadcasting array of strings.
- Use SOURCE argument (ALLOCATE).
- Fixed subroutine comments.

* Supposed unaligned access errors were wrong
  implementations in the first place. Instead,
  it is necessary to clear the bcast-message
  if src-msglen is less (or zero) than dst-msglen.

* No need to broadcast size (string-length)
  - Instead of LEN_TRIM and actual length,
    broadcast entire variable (SIZEOF).
  - Size is now a quarter the previous size
    because of MPI_CHARACTER vs MPI_INTEGER.
  - This can be applied to array of strings as well,
    which are required to be blobs like
    CONTIGUOUS strings of the same length.

### DIFF
--- a/src/mpiwrap/message_passing.F
+++ b/src/mpiwrap/message_passing.F
@@ -21,14 +21,12 @@
 !> \author JGH
 ! **************************************************************************************************
 MODULE message_passing
-   USE ISO_C_BINDING, ONLY: C_F_POINTER, &
-                            C_PTR
+   USE ISO_C_BINDING, ONLY: C_F_POINTER, C_PTR
    USE kinds, ONLY: &
       dp, int_4, int_4_size, int_8, int_8_size, real_4, real_4_size, real_8, &
       real_8_size, default_string_length
    USE machine, ONLY: m_abort
-   USE mp_perf_env, ONLY: add_perf, &
-                          add_mp_perf_env, rm_mp_perf_env
+   USE mp_perf_env, ONLY: add_perf, add_mp_perf_env, rm_mp_perf_env
 
 #include "../base/base_uses.f90"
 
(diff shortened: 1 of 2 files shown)


src/mpiwrap/message_passing.F
```diff
@@ -59,55 +57,15 @@ MODULE message_passing
 #endif
 
 #if defined(__parallel)
+! subroutines: unfortunately, mpi implementations do not provide interfaces for all subroutines (problems with types and ranks explosion),
+!              we do not quite know what is in the module, so we can not include any....
+!              to nevertheless get checking for what is included, we use the mpi module without use clause, getting all there is
```
fstein93, line 62: Is there a combination of compiler+MPI implementation where this is not possible with the `mpi_f08` module?

hfp: Thanks for asking!

This PR is just a draft yet, but it seems messy for instance OpenMPI 5.x misses `mpi_get_address` in the F08 interface (though, it does not harm the `USE` statement requesting it). The `MPI_STATUSES_IGNORE` situation is even more crowded, i.e., using an own variable with some kind of content denoting "ignore" seems to be unportable in general beside of `MPI_STATUSES_IGNORE` missing even from the pre-F08 interface.

I will play a bit more with this PR and worst case closing it without merge. What I am testing locally (beside of CI here), is ARM64 w/ OpenMPI from Ubuntu distro, Intel MPI and OpenMPI on x86. I extend this to more specific tests with OMPI4 and stock-MPICH.

hfp: I am only testing GNU Fortran at the moment (applies to Intel MPI also).

fstein93: Let's see when or whether this part of the MPI standard is available entirely (apart from hitting compiler bugs as we had them with MPICH+GCC 11 (or was it GCC 12?)).

hfp:
Broadcasting a series of MPI_CHARACTERs with actual meaning to be a string can end wrongly. What was supposedly an "unaligned access errors" which needed a workaround actually avoided two flaws:

* If the incoming string had length of zero (`LEN_TRIM`), the broadcast did not transfer anything. However, if the destination string had some content, it hence remained. In turn this yields a wrong results.
* If the destination string had some (old) content longer than what was broadcasted, it remained and also gave a wrong result.

The whole workaround with temporary integer array just avoided both problems since the broadcasted message variable (`INOUT`) was cleared before filling it from the intermediate/temporary result (`msg = ""`).

There might have been an unaligned access error in the past, but the workaround actually only served to avoid above two problems (for likely the past 25 years).

hfp:
Failed for [PDBG](https://storage.googleapis.com/cp2k-ci/run-cp2k-pdbg-9b3ce7df_report.txt) because of #4537.

hfp:
This PR also drops ~180 LOC from the codebase.

PR #4535, opened 2025-11-13
Monovalent atoms

antalik:
Added an option to turn any atomic kind into monovalent, i.e., with nuclear charge 1.0 and contributing a single electron. It is primarily intended for saturating dangling bonds in QM/MM and should ideally be used in conjunction with pseudopotentials developed for such a purpose.

### DIFF
--- a/src/input_cp2k_subsys.F
+++ b/src/input_cp2k_subsys.F
@@ -1038,7 +1038,7 @@ CONTAINS
 
       CALL section_create(section, __LOCATION__, name="KIND", &
                           description="The description of the kind of the atoms (mostly for QM)", &
-                          n_keywords=19, n_subsections=1, repeats=.TRUE.)
+                          n_keywords=20, n_subsections=1, repeats=.TRUE.)
 
       NULLIFY (keyword)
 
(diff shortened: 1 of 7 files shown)


tests/TEST_DIRS
```diff
@@ -3,6 +3,7 @@
 # Directories have been reordered according the execution time needed for a gfortran pdbg run using 2 MPI tasks
 # in case a new directory is added just add it at the top of the list..
 # the order will be regularly checked and modified...
+QS/regtest-monovalent                                       libint !ifx
```
fstein93: If I see it right, this test does not depend on libint (for Hartree-Fock calculations). What happens if you run this test with ifx?

src/input_cp2k_subsys.F
```diff
@@ -1304,6 +1304,17 @@ SUBROUTINE create_kind_section(section)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
+      CALL keyword_create(keyword, __LOCATION__, &
+                          name="MONOVALENT", &
+                          description="This keyword makes all atoms of this kind monovalent, i.e., with "// &
```
fstein93: I think you can remove the comma after `i.e.`

tests/TEST_DIRS
antalik: Sorry about that! Since I'm new to the testing environment in CP2K, I just took a copy of the nearest test dir without giving it too much thought. So it really doesn't and I haven't compiled it with Intel, but I don't expect anything strange going on. I'll fix it in a minute!

antalik:
I'm actually not sure what to do about the CI, as the pipeline fails when downloading openblas...

PR #4537, opened 2025-11-18
Check if deriv is associated

hfp:
This fixes accessing non-associated pointers, and it can be reproduced by running for instance the PDBG regtests (runtime failure). The checks introduced (whether pointers are associated or not) are similar to other parts of the same module.

### DIFF
--- a/src/xas_tdp_atom.F
+++ b/src/xas_tdp_atom.F
@@ -121,6 +121,7 @@ MODULE xas_tdp_atom
                                               xc_dset_get_derivative,&
                                               xc_dset_release
    USE xc_derivative_types,             ONLY: xc_derivative_get,&
+                                              xc_derivative_p_type,&
                                               xc_derivative_type
    USE xc_derivatives,                  ONLY: xc_functionals_eval,&
                                               xc_functionals_get_needs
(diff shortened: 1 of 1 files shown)


src/xas_tdp_atom.F
```diff
@@ -2899,31 +2899,32 @@ SUBROUTINE integrate_sc_fxc(int_fxc, iatom, ikind, deriv_set, xas_atom_env, qs_e
...
       ! Allocate some work arrays
       ALLOCATE (fxc(na, nr))
       ALLOCATE (int_so(nsotot, nsotot))
 
       ! Integrate for all three derivatives, taking the grid weight into account
       ! If closed shell, do not need to integrate beta-beta as it is the same as alpha-alpha
-      ub = 2; IF (nspins == 2) ub = 3
-      DO i = 1, ub
-
+      DO i = 1, MERGE(2, 3, nspins /= 2)
```
oschuett: `DO i = 1, nspins+1`  ?

juerghutter:
The regtest failing for psmp (RIXS) is caused by a bug. The code is missing a 
xc_functionals_expand
for the new XC section.
We will fix this later today.

PR #4547, opened 2025-11-20
Adjusted CRITICIAL section names for Intel Fortran Compiler

hfp:
With Intel Fortran Compiler (ifx and ifort), the name of an OpenMP critical section cannot be named like the subroutine containing it. The issue will be reported.

### DIFF
--- a/src/common/gamma.F
+++ b/src/common/gamma.F
@@ -545,7 +545,7 @@ CONTAINS
                        "incomplete Gamma function is invalid")
       END IF
 
-!$OMP CRITICAL(init_md_ftable)
+!$OMP CRITICAL(omp_init_md_ftable)
       !   *** Check, if the current initialization is sufficient ***
       IF (nmax > current_nmax) THEN
          CALL deallocate_md_ftable()
@@ -553,7 +553,7 @@ CONTAINS
          !     *** for the Taylor series expansion      ***
          CALL create_md_ftable(nmax, 0.0_dp, 12.0_dp, 0.1_dp)
       END IF
-!$OMP END CRITICAL(init_md_ftable)
+!$OMP END CRITICAL(omp_init_md_ftable)
 
    END SUBROUTINE init_md_ftable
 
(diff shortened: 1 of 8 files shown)


hfp:
The script `tools/toolchain/scripts/generate_arch_files.sh` formed the wrong dependencies for replacement rules aka workarounds (`.F` instead of triggering `.F90` aka FYPP). Disabled workarounds for certain compilation units when using Intel Compiler. Updated workaround rules even if commented out; avoid similar issues if WAs are ever needed for whatever compiler.

oschuett:
I'd like to remove the `Makefile` in a week or two, which will render `generate_arch_files.sh` obsolete. So, maybe we should convert the Intel tests to CMake before spending more time on them?

hfp:
> I'd like to remove the `Makefile` in a week or two, which will render `generate_arch_files.sh` obsolete. So, maybe we should convert the Intel tests to CMake before spending more time on them?

Yes, that's in mind and I will help with this whether it's here, with CMake, or with the Spack repo/recipes. Anyway, I will squeeze out the "value" (issues) of the failures here (Makefile) before moving on. I already reported the name clash for critical (this will be an extra comment).

hfp:
Regarding name of CRITICAL sections, this is what our people say:

> The names of critical constructs are global entities of the program. If a name conﬂicts with any other entity, the behavior of the program is unspeciﬁed.

Citation and reference is [here](https://www.openmp.org/spec-html/5.2/openmpse90.html#x227-23900015.2).

As we all interpret the standard, there are different interpretations (let alone issues in the standard). Also, "there is hope for forgiveness and this shall become a warning rather than an error".  It might be Intel follows GNU Fortran to accept a clashing name. That's my hope btw (aka IMHO).

oschuett:
I don't mind as long as it reliably throws an error - ideally with a meaningful message. 
The part that worries me is: _"the behavior of the program is unspeciﬁed"_.

hfp:
It's not too bad to get an error and resolve via this PR (sure, I am affiliated ;-).

This now triggers other/unrelated issues, i.e., the [ICE](https://ci.cp2k.org/artifacts/run-cp2k-intel-ifx-psmp-9eb354a6/make.out) for `hfx_contraction_methods.F`, which will be reported too.

PR #4549, opened 2025-11-22
Docker: Migrate HIP test to CMake

oschuett:
I'm trying to migrate our HIP test to CMake. Unfortunately, CMake adds invalid compiler options when MPI is enabled due to [this bug]( https://gitlab.kitware.com/cmake/cmake/-/issues/27231). While gcc simply ignores unknown options, the clang++ compiler used by HIP stops with an error:

```
clang++: error: unknown argument: '-fallow-invalid-boz'
clang++: error: unknown argument: '-fallow-argument-mismatch'
```

@alazzaro @mtaillefumier any thoughts?

### DIFF
--- a/cmake/CompilerConfiguration.cmake
+++ b/cmake/CompilerConfiguration.cmake
@@ -154,7 +154,14 @@ add_compile_options(
   "$<$<AND:$<CONFIG:DEBUG>,$<COMPILE_LANG_AND_ID:C,AppleClang>>:-O0;-g>"
   "$<$<AND:$<CONFIG:DEBUG>,$<COMPILE_LANG_AND_ID:Fortran,NAG>>:-C=all>")
 
-# Tweaks
+# =================================== Tweaks ===================================
+# Workaround https://gitlab.kitware.com/cmake/cmake/-/issues/27231
+get_target_property(opts MPI::MPI_Fortran INTERFACE_COMPILE_OPTIONS)
+set_target_properties(
+  MPI::MPI_Fortran PROPERTIES INTERFACE_COMPILE_OPTIONS
+                              "$<$<COMPILE_LANGUAGE:Fortran>:${opts}>")
+unset(opts)
+
 if(CMAKE_C_COMPILER_ID STREQUAL "AppleClang")
   set(CMAKE_Fortran_MODOUT_FLAG "-ef") # override to get lower-case module file
                                        # names
--- a/cmake/cmake_cp2k.sh
+++ b/cmake/cmake_cp2k.sh
@@ -136,7 +136,7 @@ elif [[ "${PROFILE}" == "toolchain_cuda_"* ]] && [[ "${VERSION}" == "psmp" ]]; t
     -DCMAKE_INSTALL_PREFIX="${INSTALL_PREFIX}" \
     -DDBCSR_DIR="${DBCSR_CUDA_ROOT}/lib/cmake/dbcsr" \
     -DSPLA_ROOT="${SPLA_ROOT}-cuda" \
-    -DCP2K_WITH_GPU="${PROFILE:(-4)}" \
+    -DCP2K_WITH_GPU="${PROFILE:15}" \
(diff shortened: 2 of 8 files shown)


mtaillefumier:
a potential workaround could be to search for these options and suppress them manually. Dirty but it might work.

oschuett:
I managed to work around the CMake bug, but now I have apparently hit [a compiler bug](https://storage.googleapis.com/cp2k-ci/run-cp2k-hip-rocm-build-4848ec33_report.txt):
```
during IPA pass: static-var
lto1: internal compiler error: Segmentation fault
0x1013aba internal_error(char const*, ...)
	???:0
0x1288247 DFS::DFS_write_tree(output_block*, DFS::sccs*, tree_node*, bool, bool)
	???:0
0x12887c9 DFS::DFS_write_tree_body(output_block*, tree_node*, DFS::sccs*, bool)
	???:0
0x1286597 DFS::DFS(output_block*, tree_node*, bool, bool, bool)
	???:0
0x128622d lto_output_tree(output_block*, tree_node*, bool, bool)
	???:0
0x15e2f6c output_symtab()
	???:0
0x15e0611 lto_output()
	???:0
0x7a7605 ipa_write_optimization_summaries(lto_symtab_encoder_d*)
	???:0
0x157ee31 lto_main()
	???:0
Please submit a full bug report, with preprocessed source (by using -freport-bug).
Please include the complete backtrace with any bug report.
See <file:///usr/share/doc/gcc-13/README.Bugs> for instructions.
lto-wrapper: fatal error: /usr/bin/gcc returned 1 exit status
compilation terminated.
```

mtaillefumier:
is it gcc 13 or 14 because I also encounter compilers bug on ubuntu 24.

oschuett:
Yes, it's  gcc 13.
Ok, then I'll try gcc 12.

oschuett:
I tried gcc 12 and 11, but neither worked. Hence, I went back to `rocm/dev-ubuntu-22.04:5.3.2-complete`.

Now I'm stuck with a linker error:

```
src/CMakeFiles/dbm_miniapp.dir/dbm/dbm_multiply_gpu_kernel.cu.o: file not recognized: file format not recognized
```

I believe the culprit is that clang++ is called with `-flto=auto`. Unfortunately `add_compile_options("-fno-lto")` doesn't work because it's too early in the command line. I tried various other ways to disable LTO/IPO, but without success so far.

mtaillefumier:
OK, thanks for this information. I will check as well as I have a full build of cp2k with spack and ubuntu working.  ican check which option is used. I will also check if clang is used or not. I am really surprized it does not work because I build cp2k by hand quite often.

alazzaro:
Interesting, I did a comment but it was not uploaded... 
The problems of the lto comes from Ubuntu >=22 forcing LTO for all packages. Now, if you are using MPICH (for example), then the same option is applied when you use mpicc/mpifort. This is not a cmake issue.
The way I found to solve this in DBCSR is reported at:

https://github.com/cp2k/dbcsr/blob/a1bdab5ad0de62254078d7f0413ff8a0f758c486/tools/docker/Dockerfile.build-env-ubuntu#L67

Hope it helps

oschuett:
Thanks a lot Alfio! This indeed did the trick.

alazzaro:
So, the HIP test on ROCM 5.x is really too old... We should aim at least for ROCM 6.3. Is there something @gsitaram can check maybe?

mtaillefumier:
maybe this will help https://github.com/amd/InfinityHub-CI/pull/22/files

oschuett:
Yes, I tried upgrading to `rocm/dev-ubuntu-24.04:7.1-complete`, but got compiler crashes.

It would be great if someone else could take a stab at it. In an ideal world only a single line would need to be changed:

https://github.com/cp2k/cp2k/blob/b7551b79a1ba22cd0ec740d3302589133e04df5e/tools/docker/generate_dockerfiles.py#L536

mtaillefumier:
OK. I will give it a try.

PR #4446, opened 2025-09-20
SF-TDDFT implementation

NachoXmex:
The linear-response TDDFT module was extended to spin-flip excitations. This extension is capable to make collinear and noncollinear SF-TDDFT calculations. Some key details of the implementation are:

  - The noncollinear kernel used is the Wang & Ziegler kernel with a screening in the denominator to stabilize it
  - The noncollinear kernel derivative is calculated by finite differences

PR #4551, opened 2025-11-24
Revert: restore original tddft force printing

BelizSertcan:
Restores the changes done in #4446 for tddft force printing.


hfp:
LGTM, ready for merge?

Issue #4396, opened 2025-08-24
Some MO for k-points missing in output after multi-process parallel execution with MPI


he-zilong:
Dear CP2K developers,
The printing of MO information for k-points as requested by `&FORCE_EVAL/&DFT/&PRINT/&MO` is affected by parallel execution, as there are fewer items from a multi-process run than from a single-process single-thread run. The following calculations on an fcc-Ni structure with k-point set to 11x11x11 demonstrate the behavior:
[Ni_prim.tar.gz](https://github.com/user-attachments/files/21956196/Ni_prim.tar.gz)
Program version, execution method and result comparison are attached.
```
# cp2k.ssmp --version
 CP2K version 2025.1
 Source code revision git:9635df4
 cp2kflags: omp libint fftw3 libxc libgrpp xsmm spglib libdftd4 libvori libbqb
 compiler: GCC version 11.5.0 20240719 (Red Hat 11.5.0-5)
...[compiler options omitted]
# mpirun --version
mpirun (Open MPI) 5.0.6
# cp2k.ssmp Ni_prim.inp > Ni_prim_ssmp.out
# mpirun -np 1 -x OMP_NUM_THREADS=1 cp2k.psmp Ni_prim.inp > Ni_prim_psmp_p1t1.out
# mpirun -np 2 -x OMP_NUM_THREADS=1 cp2k.psmp Ni_prim.inp > Ni_prim_psmp_p2t1.out
# mpirun -np 4 -x OMP_NUM_THREADS=1 cp2k.psmp Ni_prim.inp > Ni_prim_psmp_p4t1.out
# mpirun -np 6 -x OMP_NUM_THREADS=1 cp2k.psmp Ni_prim.inp > Ni_prim_psmp_p6t1.out
# mpirun -np 1 -x OMP_NUM_THREADS=2 cp2k.psmp Ni_prim.inp > Ni_prim_psmp_p1t2.out
# for i in Ni_prim_*.out ; do echo $i ; grep "DBCSR| MPI" $i -n ; grep "DBCSR| OMP" $i -n ; grep "EIGENVALUES AND OCCUPATION NUMBERS FOR K POINT" $i -c ; done
Ni_prim_psmp_p1t1.out
14: DBCSR| MPI: My process id                                                     0
15: DBCSR| MPI: Number of processes                                               1
16: DBCSR| OMP: Current number of threads                                         1
17: DBCSR| OMP: Max number of threads                                             1
1332
[...]

PR #4557, opened 2025-11-25
kpoint : Transformations for trivial kpoint parallelism

StepanMarek:
Certain schemes we would like to implement in the future will use trivial kpoint parallelism - computational bottleneck of calculations such as real-time semiconductor-Bloch equations propagation will likely be the evaluation of propagation at large kpoint grid, while information from different kpoints will not required by other kpoints. This means that assigning a set of kpoints to each mpi rank leads to efficient parallelism, since the the mpiranks essentially do not require communication.

In order to truly eliminate the communication, we propose replication of real-space matrices on each rank, which is viable for unit cells with small number of atoms/small overlaps with neighbouring image cells. This pull request prepares the routines for such replication.

Since two developers from our group work with these methods, it made sense to us to separate the methods into a standalone pull request, on top of which we will build actual new features.

Co-written with @ShridharShanbhag.


StepanMarek:
Sorry, I realized that the convention for the mirror image cell index differs between the replication and transform routines, I will fix it shortly.

he-zilong:
Dear developers, apologies if interjection is inappropriate but would this be an opportunity to fix #4396 as well?

StepanMarek:
> Dear developers, apologies if interjection is inappropriate but would this be an opportunity to fix #4396 as well?

Dear @he-zilong , I think this is best left to a developer more familiar with the kpoints code - this pull request essentially prepares ground for future code which will use different parallelism scheme from the current kpoints code, so I probably won't use the MO printing code in the code I will write in the future.

Hopefully this issue will get resolved in a separate pull request.

PR #4566, opened 2025-11-28
Fixed a rare matrix distribution bug which sometimes caused crashes i…

andsinya:
Fixed a rare matrix distribution bug which sometimes caused crashes in linres_solver when num_pe(1) /= num_pe(2).
Here is what caused it for this case (309 AOs, 32 occupied, num_pe = [4, 8]):
- some (but not all) matrices involved in response calculation were created using _square_ 309 x 309 mo_coeff matrix as a template
- square matrix gets row block size (309 + 4 - 1) / 4 = 78 > default 64 = 64 and col block size (309 + 8 - 1) / 8 = 38, **they are forced to be equal**, so the smaller one is used for both dimensions, thus 38 x 38
- when you create a _rectangular_ 309 x 32 matrix struct using a _square_ one as a template, you get 38 row block size and (32 + 8 - 1) / 8 = 4 col block size
- when you create a _rectangular_ 309 x 32 matrix struct without a template, you get default 64 row block size and 4 col block size
- matrices with different block sizes are incompatible, hence the crash during cp_fm_to_fm_matrix
- this fix changes most work matrices to use the same template and get 38 row block size, my example works now
- I may have missed a few, so please point me if you find another rectangular matrix created without a template 

See Alps output attached:
[_tddfpt_fm_dist_fix.zip](https://github.com/user-attachments/files/23821747/_tddfpt_fm_dist_fix.zip)

### DIFF
--- a/src/qs_tddfpt2_methods.F
+++ b/src/qs_tddfpt2_methods.F
@@ -1623,7 +1623,6 @@ CONTAINS
          nmo = gs_mos(ispin)%nmo_active
          CALL cp_fm_struct_create(fm_struct, template_fmstruct=gs_mos(ispin)%mos_occ%matrix_struct, &
                                   ncol_global=nmo, context=blacs_env)
-         ! CALL cp_fm_struct_create(fm_struct, nrow_global=nao, ncol_global=nmo, context=blacs_env)
          NULLIFY (gs_mos(ispin)%mos_active)
          ALLOCATE (gs_mos(ispin)%mos_active)
          CALL cp_fm_create(gs_mos(ispin)%mos_active, fm_struct)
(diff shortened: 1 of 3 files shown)


src/qs_tddfpt2_methods.F
```diff
@@ -1621,7 +1621,9 @@ SUBROUTINE init_res_method(qs_env, gs_mos, tddfpt_control, tddfpt_section, iouni
          CALL get_qs_env(qs_env, blacs_env=blacs_env)
          CALL cp_fm_get_info(gs_mos(ispin)%mos_occ, nrow_global=nao)
          nmo = gs_mos(ispin)%nmo_active
-         CALL cp_fm_struct_create(fm_struct, nrow_global=nao, ncol_global=nmo, context=blacs_env)
+         CALL cp_fm_struct_create(fm_struct, template_fmstruct=gs_mos(ispin)%mos_occ%matrix_struct, &
+                                  ncol_global=nmo, context=blacs_env)
+         ! CALL cp_fm_struct_create(fm_struct, nrow_global=nao, ncol_global=nmo, context=blacs_env)
```
oschuett: Please remove the original code instead of just commenting it.

andsinya: done

PR #4585, opened 2025-12-03
Initialize SIRIUS library only if needed (fixes #4564)

### LINKED ISSUE #4564: CP2K fails running its self-dumped input files
Running the following minimal CP2K input with the `--echo` flag
```
&GLOBAL
  PROJECT H2O
  RUN_TYPE energy
&END GLOBAL

&FORCE_EVAL
  &DFT
    BASIS_SET_FILE_NAME BASIS_SET
    POTENTIAL_FILE_NAME POTENTIAL
    &XC
      &XC_FUNCTIONAL PADE
      &END XC_FUNCTIONAL
    &END XC
  &END DFT
  &SUBSYS
    &CELL
      ABC 6.0 6.0 6.0
    &END CELL
    &COORD
      O   0.000000    0.000000   -0.065587
      H   0.000000   -0.757136    0.520545
      H   0.000000    0.757136    0.520545
    &END COORD
    &KIND H
      BASIS_SET DZVP-GTH-PADE
      POTENTIAL GTH-PADE-q1
    &END KIND
    &KIND O
      BASIS_SET DZVP-GTH-PADE
      POTENTIAL GTH-PADE-q6
    &END KIND
  &END SUBSYS
&END FORCE_EVAL
```
dumps a CP2K input file which does not run with CP2K, because the first input line
```
SIRIUS 7.7.1, git hash: https://api.github.com/repos/electronic-structure/SIRIUS/git/ref/tags/v7.7.1
```
is not valid. That line is always printed in the CP2K output when CP2K is built with SIRIUS, even if SIRIUS is not used.
This issue could be fixed either on the SIRIUS side by prepending a comment character like `#` or (better) by suppressing any output during the [SIRIUS initialization](https://github.com/mkrack/cp2k/blob/master/src/f77_interface.F#L301) which is done (for whatever reason) even if the SIRIUS is not used.

After fixing that first line, the dumped input is still not working, because of the deprecated input keywords included in the dump like
```
 *  \___/  The specified keyword 'AUX_BASIS_SET' is not available anymore: use *
 *    |                         'BASIS_SET AUX ...' instead                    *
 * / \                                               input/input_parsing.F:262 *
```
This issue could be fixed either by substituting the deprecated keywords with the new ones, which appears to me not that easy to implement or by simply dropping the deprecated keywords, because the grace period for deprecated keywords should end at some point.

Any comments or suggestions are welcome.

mkrack:


mtaillefumier:
Thanks for this work. My PR on the SIRIUS side was merged today. We will create a new minor release as soon as possible.

mkrack:
@mtaillefumier Thanks for pushing things on the Sirius side, though that is no longer crucial if this PR works.

PR #4613, opened 2025-12-18
TREXIO 2.5.0 -> 2.6.0

mkrack:


stefabat:
#4615 should also fix the failure of the trexio tests

PR #4615, opened 2025-12-18
Trexio bugfix

stefabat:
This PR fixes wrong basis set and atomic orbital normalization factors dumped to the trexio file for the case of spherical GTOs.


mkrack:
The problem with writing the state_energy has been reported as issue https://github.com/TREX-CoE/trexio/issues/231

PR #4621, opened 2025-12-18
Implementation of moments for k-point calculations 

ShridharShanbhag:
We implement a method to calculate non-diagonal moment matrix elements for k-point calculations.

The subroutine "qs_moment_kpoints" in qs_moments.F calculates the moments for each k-point added in the input in the &MOMENTS section.
The k-points for the calculations can be added in the &MOMENTS section as they are added in the &KPOINTS section in DFT or, as in the &KPOINT_SET section in the BAND_STRUCTURE section. If no k-points are added in the &MOMENTS section, &KPOINTS from DFT are used.

The subroutine "build_local_moment_matrix_rs_img" calculates the local moment matrix for a periodic system across image cells. and stores them in a DBCSR matrix of type dbcsr_type_no_symmetry.

The subroutine "get_xkp_for_dipole_calc" obtains the list of k-point coordinates needed for these calculations.

Two new subroutines have been added in the mathlib.F to facilitate these calculations. The subroutine "geeig_right" solves the generalized eigenvalue equation to compute the right generalized eigenvectors of complex matrices.

New tests were created in "QS/regtest-moments-kpoints" to ensure the stability of the outputs and to ensure that the three kpoint input methods mentioned above remain valid.
"C2_pbe_scf_kp.inp" tests the default SCF input.
"C2_pbe_moment_kp.inp" tests the KPOINT input in the MOMENTS section.
"C2_pbe_moment_kpset.inp" tests the KPOINT_SET input in the MOMENTS section.

This code was developed under the supervision of @JWilhelm with help from @StepanMarek


tests/matchers.py
```diff
@@ -262,4 +262,6 @@ def __setitem__(self, key: str, value: Matcher) -> None:
 registry["M126"] = GenericMatcher(r" # Total charge ", col=5)
 
 registry["M127"] = GenericMatcher(r"Checksum (Acoustic Sum Rule):", col=5)
+registry["M128"] = GenericMatcher(r" 1  1  2", col=4)
```
oschuett: Please pick descriptive names for new matchers and maybe also leave a comment line.
See e.g.  `BSE_1st_excit_ener_TDA`.

ShridharShanbhag: I changed the names of the matchers and added comments describing them.
The names are now: Dipole_at_kp_1, BC_near_K_point

he-zilong:
Great job! Just asking from a user's perspective: now that dipole moments with k-points are available from a single-point calculation, would IR intensities (from static vibrational analysis or from autocorrelation during molecular dynamics) with k-points be available as well?

JWilhelm:
Thanks for the question. The dipole matrix elements that are now available with k-point sampling make electronic optical spectra of solids accessible, but they do not directly give access to infrared vibrational spectra of crystals. Details:

The quantities commited in this PR are Bloch-state dipole matrix elements

$$
d_{nm}(\mathbf{k}) = i \langle u_{n\mathbf{k}} \mid \nabla_{\mathbf{k}} \mid u_{m\mathbf{k}} \rangle
$$

These $d_{nm}(\mathbf{k})$ are dipole matrix elements between Bloch states. They are the basic ingredients for computing optical absorption, dielectric functions, and optical conductivity arising from electronic transitions in solids. In that sense, electronic optical spectra (including electronic absorption in the IR, visible, or UV range) become accessible once these matrix elements are available with proper k-point sampling.

Infrared vibrational spectra are a different problem. There, absorption is caused by phonons, and the relevant quantity is how the macroscopic polarization changes when atoms are displaced. This is described for example by Born effective charges together with the phonon eigenvectors. These quantities are obtained from linear-response calculations or from finite differences of the Berry-phase polarization, not from explicit Bloch-to-Bloch dipole matrix elements like the one above.

In phonon IR calculations, dipole-like quantities do appear, but they are ionic effective charges, not electronic transition dipoles. The electronic dipole matrix elements $d_{nm}(\mathbf{k}) $ do not appear explicitly in the final expressions for harmonic IR intensities.

Issue #4611, opened 2025-12-17
Some build flags inaccessible through CMake


oschuett:
Some build flags are not yet accessible through CMake.

#### Add to CMake:
- [x]  `__CRAY_PM_ACCEL_ENERGY`
- [x]  `__CRAY_PM_ACCEL_ENERGY`
- [x]  `__CRAY_PM_ENERGY`
- [x]  `__CRAY_PM_ENERGY`
- [x]  `__NO_STATM_ACCESS`
- [x]  `__OFFLOAD_PROFILING`
- [ ]  `__SCALAPACK_NO_WA`
- [ ]  `__NO_SOCKETS`

#### Replace with OFFLOAD flag:
- [x]  `__ELPA_AMD_GPU`
- [x]  `__ELPA_INTEL_GPU`
- [x]  `__ELPA_NVIDIA_GPU`

#### Remove:
- [x]  `__HAS_PATCHED_CUFFT_70`
- [x]  `__MINGW`
- [x]  `__NO_MPI_THREAD_SUPPORT_CHECK`
- [x]  `__QUIP`
- [x] `__FFTW3_UNALIGNED`

#### Unclear:
- [ ]  `__PW_FPGA`
- [ ]  `__PW_FPGA_SP`
- [ ]  `__STATM_RESIDENT`
- [ ]  `__STATM_TOTAL`

alazzaro:
The `_CRAY_*` are still OK. I think the `*STATM*` are there because of MacOS?
No idea about the others...

fstein93:
`__ELPA_*_GPU` used to be set by the arch file. Probably, we can replace it using the offloading flags.
`__HAS_PATCHED_CUFFT_70` was a workaround to a bug in the cuFFT interface (version 7.0) and we should be able to remove it (current version is 13.0).
`__MINGW` is used in [`machine.F`](https://github.com/cp2k/cp2k/blob/b99ca74f88d83c4c395fc4f7ec9c049eb12ca25a/src/base/machine.F#L425) and is probably a reminiscence to the old MINGW builds on Windows. I think the current recommendation is to use WSL instead (should be always available).
`__NO_MPI_THREAD_SUPPORT_CHECK` disables threading for MPI in case that the MPI library does not offer threading support. Is this still a thing?
[...]

PR #4624, opened 2025-12-19
[cmake] Added CRAY power management counters support

mtaillefumier:
- Added CRAY_PM_ENERGY and CRAY_PM_ACCEL_ENERGY. Patch related to issue #4611
- Fix an issue when defining cp2k::cosma target


docs/getting-started/build-from-source.md
```diff
@@ -43,3 +43,6 @@ cmake --build build -j 32
 - `-DCP2K_USE_EVERYTHING=ON` Enables all dependencies.
 - `-DCP2K_ENABLE_CONSISTENCY_CHECKS=ON` Only used for
   [testing](https://dashboard.cp2k.org/archive/misc/index.html).
+- `-DCP2K_USE_CRAY_PW_ENERGY` Used to monitor pwoer consumption on Cray systems. =
```
oschuett: typo `pwoer` -> `power`.

docs/technologies/libraries.md
```diff
@@ -236,3 +236,9 @@ greenX - Open-source file format and library. Support for greenX can be enabled
 - Pass `-DCP2K_USE_HDF5=ON` to CMake to enable HDF5 support.
 - HDF5 is a hard dependency for SIRIUS and TREXIO, but can also be used by itself to allow
   read/write functionalities of QCSchema files in the active space module.
+
+## Energy monitoring on Cray systems
```
oschuett, line 240: I think it's sufficient to document these options only on the `build-from-source.md` page.
Also, it's not actually a library.

mtaillefumier: That's what I did initially but the make_pretty script was still complaining.

docs/getting-started/build-from-source.md
mtaillefumier: it should be fixed. Thanks for spotting it

docs/technologies/libraries.md
oschuett: Oh I see, there is actually another typo `PW` -> `PM`.

PR #4638, opened 2025-12-28
Fix HDF5 installation path detection

Growl1234:
The command `cut -d: -f2` results in a leading space in the output path, such as "` /opt/hdf5`" instead of `/opt/hdf5`. This causes CMake to fail in locating the HDF5 path, reporting the error: "`Could NOT find HDF5 (missing: HDF5_LIBRARIES HDF5_INCLUDE_DIRS C Fortran) (found version "")`". Alternatively, replacing `cut -d: -f2` with `awk '{print $3}'` can avoid the problem.


Growl1234:
Here is a screenshot that can prove this: 
图片

PR #4652, opened 2026-01-02
Docker: Re-enable tblite test after solving #4362

Growl1234:
Docker: Re-enable tblite test after solving #4362


oschuett:
Thanks a lot for this PR! After your suggestion yesterday, I have actually also been working on re-enabling tblite.

Usually I would give your PR precedence (you discovered the workaround after all), but there are a number of additional changes needed. So, for the sake of simplicity, I'll close this PR in favor of my own #4653. 

Kudos again for finding this solution!

PR #4677, opened 2026-01-08
Restoring OSX arm64 support

tlaino:
This Pull Request restores comprehensive support for macOS on ARM64 (Apple Silicon), specifically tested and optimized for the M4 Max architecture.

Key Improvements:

Compiler Flags: Optimized for Apple Silicon by using -mcpu=native for GCC, resolving invalid architecture errors.

Runtime Fix: Added -D__NO_STATM_ACCESS and -D__MACOSX to bypass Linux-specific /proc filesystem calls that cause crashes on Darwin.

Toolchain Portability: Patched several stage scripts (ELPA, DeePMD, Libxsmm, OpenBLAS) to resolve GNU-specific linker flag incompatibilities on macOS.

Modernization: Updated CMake logic for FindBlas/FindLapack to support modern alias targets required by recent dependencies like tblite.

Still pending issues: libtorch (and consequently deepmd).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -855,9 +855,8 @@ if(CP2K_USE_GREENX)
   # GreenX with AC requires GMP library
   find_library(GMP_LIBRARY NAMES gmp REQUIRED)
   get_filename_component(GMP_LIBRARY_PATH ${GMP_LIBRARY} DIRECTORY)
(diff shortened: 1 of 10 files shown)


oschuett:
Welcome back!
The slow test is a false positive that I'll fix today.
LGTM from my side, but I'd like to give @mtaillefumier a chance to take a look as well.

PR #4683, opened 2026-01-09
ELPA: compile-time choice for solver-kind

hfp:
- For the OpenCL based GPU code-path, or by choice, ELPA may not be built with GPU support.
- Introduced __NO_OFFLOAD_ELPA to allow CPU-only ELPA (related to #4617).
- Allow compile-time selection of ELPA solver-kind (1-stage vs 2-stage).
- For ELPA/SYCL, use 1-stage ELPA solver by default for the time being.
- ELPA/SYCL (MKL in general?) has an index-fix currently only in a branch.

Index/ILP issue (undefined behavior):
- ELPA eventually mismatches index-kinds (LP64 vs ILP64 when linking BLAS/MKL).
- For static linkage it usually does not not matter (at least CPU-only).
- CP2K links the correct BLAS flavor (SO and GPU-build may be affected?).
- ELPA 2-stage may be even affected beyond MKL (or for GPU in general).
- ELPA 1-stage is apparently not affected by index issue.


hfp:
I will merge "as is" but perhaps follow-up making 1-stage vs 2-stage a runtime choice. A runtime choice (with keyword) perhaps enables cross-testing wrt correctness and performance. On top, 1-stage is likely always better for "smaller" matrices.

oschuett:
Thanks for following up on my #4617. Yeah, ELPA has a lot of moving parts...

I've also triggered the pdbg to actually test your changes.

I'm starting to think that we should maybe run pdbg by default on PRs.

hfp:
> I'm starting to think that we should maybe run pdbg by default on PRs.

For the sake of resource consumption, it could be a full switch from sdbg to pdbg. I wonder if pdbg runs longer and why. Anyway, to get "correct" reference values is no reason to prefer sdbg.

PR #4699, opened 2026-01-15
Rework spack configuration files

mkrack:


tools/spack/cp2k_deps_psmp.yaml
```diff
@@ -1,160 +1,196 @@
...
     all:
       buildable: true
-      prefer:
-        - ~cuda
-        - +mpi
-        - +openmp
-        - +pic
-        - ~rocm
-      require: target=""
+      variants:
```
RMeli: [`prefer` is stronger than `variants`](https://spack.readthedocs.io/en/latest/frequently_asked_questions.html#why-does-spack-pick-particular-versions-and-variants), so if this is intended, maybe it would be best to add these variants directly to the `require`s of the explicit dependencies where we really want these variants to be enabled? This would avoid possible surprises in concretization.

mkrack: Good point. It would be easier to stick to `prefer` as it has not caused problems so far. An explicit definition of the variants for each dependency blows up the configuration, but is less misleading. A `prefer` `+openmp` for instance does not ensure that this becomes really effective for a dependency eventually. I guess if a `preferred` variant is added to the build recipe for a dependency, it will be picked up by `spack` automatically and our software stack for building CP2K is affected implicitly without any change on the CP2K side?

RMeli: Sorry for the late reply.

> A prefer +openmp for instance does not ensure that this becomes really effective for a dependency eventually.

To ensure that we would need `all:require:+openmp`. Requirements on variants for all packages are only enforced for the packages defining such variant.

> I guess if a preferred variant is added to the build recipe for a dependency, it will be picked up by spack automatically and our software stack for building CP2K is affected implicitly without any change on the CP2K side?

Yes, that would be the case. But it would also be the case for other changes in the recipe of a dependency (addition of a constraint, deprecation of one version, ...), so you need to assume that dependencies can change if you change the `spack-packages` commit.

`prefer` is just stronger in terms of requirements than `variant`, which I feel is the intended goal here, and can be overwritten by a `require`.

PR #4710, opened 2026-01-20
ELPA: introduced ELPA_ONE_STAGE and revised ELPA_QR/UNSAFE

hfp:
- Introduced ELPA_ONE_STAGE (see https://github.com/cp2k/cp2k/pull/4683).
- QR factorization (unsafe is now internal).
- Removed keyword ELPA_QR_UNSAFE.
- Updated to ELPA 2025.06.002.
- Disabled IEEE exceptions.
- Test coverage.


hfp:
Can someone please upload https://elpa.mpcdf.mpg.de/software/tarball-archive/Releases/2025.06.002/elpa-2025.06.002.tar.gz ?

oschuett:
It's already there: https://www.cp2k.org/static/downloads/elpa-2025.06.002.tar.gz

mkrack:
> Can someone please upload https://elpa.mpcdf.mpg.de/software/tarball-archive/Releases/2025.06.002/elpa-2025.06.002.tar.gz ?

That tar file is already [available](https://www.cp2k.org/static/downloads/).

hfp:
Something like `ELPA_QR_UNSAFE` qualifies as a workaround and should not be honored with a keyword (unless it is an intrinsic property of a cleanly implemented method). This PR aims to remove this workaround at the expense of breaking compatibility with old/outdated versions.

hfp:
Truth is, regtests do not pass with QR decomp. enabled on ELPA's side (beside of the crafted regtests for ELPA/QR). This is independent of the QR_UNSAFE (which is now internal; no keyword anymore). Blocksize restrictions (questionable already in terms of correctness) were/are honored especially since the default for the minimum number of Eigenvalues (to trigger ELPA) was raised (to match minumum ScaLAPACK blocksize). I will ask to get QR in ELPA fixed.

hfp:
Another cleanup could be to demote `PRINT_ELPA` to a lone keyword rather than a section.

hfp:
I hereby confirm, QR factorization is not sufficiently tested in ELPA (according to authors). Therefore, it was right to **not** anticipate correct behavior (or using keywords for this purpose), but to report the issue(s) and seek resolution. ELPA's QR decomp. is turned **off by default**, qr_safe is subsumed by "qr", and CP2K's/our tests for QR are simply crafted (to be correct). The latter is shown by the fact of failing regtests if ELPA's QR is enabled (let alone "qr_safe").

hfp:
This PR should be watched wrt https://dashboard.cp2k.org/archive/perf-openmp/index.html (Timings of diag_cu144_broy with 32 OpenMP Threads). Note, performance relevant can be the new default of `ELPA_NEIGVEC_MIN` (raised from 16 to 64).

oschuett:
> This PR should be watched wrt https://dashboard.cp2k.org/archive/perf-openmp/index.html (Timings of diag_cu144_broy with 32 OpenMP Threads).


Unfortunately, the time spend in `cp_fm_diag_elpa_base` with OpenMP [went up](
https://dashboard.cp2k.org/archive/perf-openmp/diag_cu144_broy_timings_32omp.txt) from ~20 to ~37 seconds.

Furthermore, the [CUDA performance test](https://dashboard.cp2k.org/archive/perf-cuda-volta/index.html) and the [CUDA regtests](https://dashboard.cp2k.org/archive/cuda-pascal/index.html) are failing.

How should we proceed?

hfp:
Oh, I missed this so far. I will send a PR restoring the original threshold. Let's see then.

oschuett:
Any news on this? I'm a bit anxious about going without the CUDA performance test for so long.

Maybe we just rollback?

oschuett:
I think, it's time we rollback this PR to restore our CUDA tests.

PR #4721, opened 2026-01-22

Growl1234:
Added a new script that simply detects installed dependencies in the toolchain and generates CMake options according to the settings specified in "install/toolchain.conf".


Growl1234 (commented):
Tried modify a bit so that it can pass precommit check

PR #4724, opened 2026-01-22
Make Toolchain generate CMake options

Growl1234:
Added a new script that simply detects installed dependencies in the toolchain and generates CMake options according to the settings specified in "install/toolchain.conf".


oschuett:
I think this is looking really good.

Btw, you can continue to push changes to your git branch even after you have opened the pull request.

Growl1234:
Thank you. 

I think there should be a more convenient way instead of manually adding too many "if-then" commands🤔, so I am trying to have a look on this and do some tests to see if another idea can work, and will submit another PR when it's almost ready (should be soon). The basic idea is to directly make the directives in toolchain.conf perfectly correspond to the respective options in src/CMakeLists.txt by using some extraction commands, thereby making the script less redundant.

Issue #4667, opened 2026-01-06
Toolchain: miscellaneous details


he-zilong:
Dear CP2K developers,

In light of recent efforts towards upcoming version, some details in the toolchain are worth revisiting. Sorry if this is looking rather long - there are four checkboxes marking separate points. (Edit on Jan 29: fixed indentation in task list and added a bold keyword for each task) (Edit on Feb 8: ticked off items, see discussion)

- [x] **CMake flags** There are fewer CMake flags currently described in `./docs/getting-started/build-from-source.md` compared with those available in `CMakeLists.txt`, with all the `-DCP2K_USE_{PACKAGE_NAME}` flags having to be entered manually. A more user-friendly goes like this:
  * Print suggested flags towards end of running `./tools/toolchain/install_cp2k_toolchain.sh` automatically based on input options, so that it can be copy-pasted later; and
  * List all flags explicitly from `CMakeLists.txt` in `build-from-source.md` complete with what they are set to by default.

- [x] **Package manager** The script `./tools/toolchain/scripts/install_requirements.sh` uses package manager to prepare some dependencies, but there is no note for their individual purpose and necessity. For instance, it can be confusing when the process reports `Error: Unable to find a match: autogen`, which has indeed been removed from newer distributions of RHEL according to [a stackexchange discussion](https://unix.stackexchange.com/questions/705878/autogen-is-no-longer-available-in-centos-9), yet subsequent program usage seems fine (maybe as long as there is no need of making a local copy of manual?). This is also a chance to review `autoconf`, `automake` and `make` now that GNU make is retired in favor of CMake in building.

- [x] **PKG_CONFIG_PATH** The line `export PKG_CONFIG_PATH="$PKG_CONFIG_PATH:${COSMA_LIBDIR}/pkgconfig"` in the install script for COSMA that is to be written to setup file is unusual for two reasons: one, it appears _twice_ inside and outside of a condition statement; and two, it differs from other scripts where `prepend_path PKG_CONFIG_PATH` (with `prepend_path` defined in `./tools/toolchain/scripts/tool_kit.sh`) is used to modify the environment variable, here resulting in premature variable expansion in the setup file.
https://github.com/cp2k/cp2k/blob/5e54ba2094a67dcb8ef44d5dfee8ef91e07eebfb/tools/toolchain/scripts/stage4/install_cosma.sh#L251-L278
And they are not the only instances in the toolchain.
  ```shell
[...]

PR #4726, opened 2026-01-22

Growl1234:
Added a new script that can detect installed dependencies in the toolchain and generate CMake options according to `tools/toolchain/install/toolchain.conf` and `src/CMakeLists.txt` (in response to a question mentiond in issue #4667).

### DIFF
--- a/tools/toolchain/install_cp2k_toolchain.sh
+++ b/tools/toolchain/install_cp2k_toolchain.sh
@@ -1054,22 +1054,5 @@ else
   ./scripts/stage7/install_stage7.sh
   ./scripts/stage8/install_stage8.sh
   ./scripts/stage9/install_stage9.sh
-  cat << EOF
-========================== usage =========================
-Done!
-To use the installed tools and libraries and cp2k version
-compiled with it you will first need to execute at the prompt:
-  source ${SETUPFILE}
-
-If you invoked the toolchain with --install-all then you can proceed to build cp2k like this:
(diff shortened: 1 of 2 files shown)


Growl1234 (commented):
corrected a small mistake

Growl1234 (commented):
I found that libvori will be ignored by mistake if the "second grep" in `ADDED_CMAKE_OPTION` exists, so I have to list DFT-D4 as a special consideration so that both problems can be avoided. 
Now there should be no other problems (at least for packages listed as "--with-PKG" in `install_cp2k_toolchain.sh --help` except for Intel compilers and ones that's not mentioned in `CMakeLists.txt`).

Growl1234:
I don't know why the check "Regtest pdbg" ran repeatedly and failed at last. Maybe re-opening the PR helps?

Growl1234:
Oops I didn't notice that merge commits is not allowed and will be synced to the PR. I have to open a new PR now. Sorry for that.

oschuett:
> I have to open a new PR now. Sorry for that.

Actually, you can just `git push --force` to your existing branch and keep the PR open.

> corrected a small mistake

No worries. I get notified whenever you push code changes and the CI tests will also restart automatically.

Generally, there is no reason to close a PR unless you want to  abandon a change entirely.

Growl1234:
Thanks a lot for the reminder. I'll remind myself of these when pushing work in the future.

PR #4727, opened 2026-01-23

Growl1234:
Added a new script that can detect installed dependencies in the toolchain and generate CMake options according to `tools/toolchain/install/toolchain.conf` and `src/CMakeLists.txt` (in response to a question mentiond in issue #4667).

### DIFF
--- a/tools/toolchain/scripts/generate_cmake_options.sh
+++ b/tools/toolchain/scripts/generate_cmake_options.sh
@@ -21,7 +21,7 @@ export CP2K_ROOT=$(cd ${ROOTDIR}/../.. && pwd)
 CMAKE_OPTIONS="-DCMAKE_INSTALL_PREFIX=../install -DCP2K_DATA_DIR=${CP2K_ROOT}/data"
 if [ -n "$(grep -- "--install-all" ${INSTALLDIR}/setup)" ]; then
   CMAKE_OPTIONS="${CMAKE_OPTIONS} -DCP2K_USE_EVERYTHING=ON -DCP2K_USE_DLAF=OFF -DCP2K_USE_PEXSI=OFF"
-  # Since "--install-all" can be used together with "--with-PKG=no" or "--with-PKG=system", an extra safeguard is added here
+  # Since "--install-all" can be used together with "--with-PKG=no", an extra safeguard is added here
   for toolchain_option in $(grep -i "dontuse" ${INSTALLDIR}/toolchain.conf | cut -d'_' -f2 | cut -d'=' -f1); do
     if [ $(eval echo "$with_"'$toolchain_option') != "__DONTUSE__" ]; then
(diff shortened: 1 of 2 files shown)


tools/toolchain/scripts/generate_cmake_options.sh
```diff
@@ -0,0 +1,87 @@
...
+source "${SCRIPT_DIR}"/signal_trap.sh
+source "${INSTALLDIR}"/toolchain.conf
+source "${INSTALLDIR}"/toolchain.env
+
+export CP2K_ROOT=$(cd ${ROOTDIR}/../.. && pwd)
+
+# ------------------------------------------------------------------------
+# generate cmake options for compiling cp2k
+# ------------------------------------------------------------------------
+export CMAKE_OPTIONS="-DCMAKE_INSTALL_PREFIX=../install"
```
oschuett: Since this variable is only used within this file it doesn't need to be exported. To further clarify this you could mark it [local](https://tldp.org/LDP/abs/html/localvar.html).

tools/toolchain/scripts/generate_cmake_options.sh
```diff
@@ -0,0 +1,87 @@
...
+else
+  # If MPI is used, set "CP2K_USE_MPI" to "ON"
+  if [ "${MPI_MODE}" != "no" ]; then
+    export CMAKE_OPTIONS="${CMAKE_OPTIONS} -DCP2K_USE_MPI=ON"
+  fi
+  # If GPU acceleration is used, add the option about GPU acceleration
+  if [ "${ENABLE_CUDA}" = "__TRUE__" ]; then
+    export CMAKE_OPTIONS="${CMAKE_OPTIONS} -DCP2K_USE_ACCEL=CUDA -DCP2K_WITH_GPU=${GPUVER}"
+  else
+    if [ "${ENABLE_HIP}" = "__TRUE__" ]; then
```
oschuett: These nested if statements could be simplified with [elif](https://tldp.org/LDP/abs/html/testconstructs.html).

oschuett (commented):
Some minor comments, otherwise LGTM. Nice work!

tools/toolchain/scripts/generate_cmake_options.sh
Growl1234: It seems that using a dedicated `local` works better when using a function, so I decided to remove the word `export` here.

tools/toolchain/scripts/generate_cmake_options.sh
Growl1234: Thank you and I've simplified this.

Growl1234:
Thank you and I've resolved these minor comments. 
Additionaly, I found "--install-all" can be used together with "--with-PKG=no", so an extra safeguard is added.

PR #4731, opened 2026-01-24
Add two testers for make_cp2k.sh build script

mkrack:


tools/docker/generate_dockerfiles.py
```diff
@@ -659,6 +665,53 @@ def install_deps_spack(version: str, mpi_mode: str) -> str:
     return output
 
 
+# ======================================================================================
+def install_make_cp2k(version: str, mpi_mode: str) -> str:
```
oschuett, line 669: The `mpi_mode` argument appears to be unused.

tools/dashboard/dashboard.conf
```diff
@@ -318,4 +318,20 @@ timeout:     170
 host:        GCP
 report_url:  https://storage.googleapis.com/cp2k-ci/dashboard_asan-psmp_report.txt
 
+# ==============================================================================
+
+[make-cp2k-psmp]
```
oschuett, line 323: We typically keep the testers that run daily further up on the dashboard. For daily testers you can also remove the `timeout: 170` lines and just rely on the 24h default.

tools/docker/generate_dockerfiles.py
mkrack: The `mpi_mode` will become available in the `make_cp2k.sh` script soon.

tools/dashboard/dashboard.conf
mkrack: I plan to replace the `spack (psmp)` and `spack (ssmp)` testers by these new testers soon.

PR #4738, opened 2026-01-26
qs_moment_kpoints open shell implemented

ShridharShanbhag:
MOMENTS calculations for KPOINTS can be done for both open shell and closed shell calculations.
Adding keyword MAX_NMO which determines maximum number of molecular orbitals closest to the Fermi level to be used for printing of the dipole moment matrix elements. (MAX_NMO 0 prints all the elements, default 10).
This new keyword is ignored if it is not a KPOINT calculation.


oschuett:
Looks good. How about adding a regtest for this new feature?

ShridharShanbhag:
Agreed. I will add a regtest for the open shell code.

PR #4741, opened 2026-01-27
Toolchain: add --enable-sse2 option for fftw with target-cpu=native

he-zilong:
This simple PR is based on one item in #4667 I made as well as the [fftw document on installation](https://fftw.org/fftw3_doc/Installation-on-Unix.html). Only `sse2` is detected for now to keep support the single/double precision.

### DIFF
--- a/tools/toolchain/scripts/stage3/install_fftw.sh
+++ b/tools/toolchain/scripts/stage3/install_fftw.sh
@@ -52,6 +52,7 @@ case "$with_fftw" in
           grep '\bavx\b' /proc/cpuinfo 1> /dev/null && FFTW_FLAGS="${FFTW_FLAGS} --enable-avx"
           grep '\bavx2\b' /proc/cpuinfo 1> /dev/null && FFTW_FLAGS="${FFTW_FLAGS} --enable-avx2"
           grep '\bavx512f\b' /proc/cpuinfo 1> /dev/null && FFTW_FLAGS="${FFTW_FLAGS} --enable-avx512"
+          grep '\bsse2\b' /proc/cpuinfo 1> /dev/null && FFTW_FLAGS="${FFTW_FLAGS} --enable-sse2"
         fi
       fi
       ./configure CFLAGS="${CFLAGS} -std=c17" \


oschuett:
The sse2 instruction set is from the year 2000. AFAIU, enabling it in FFTW would just bloat the binary but give no performance advantage whatsoever - unless your CPU is from before 2011 when AVX was introduced.

Maybe I'm missing something?

he-zilong:
The idea of building fftw with `--enable-sse2` flag has been suggested by gromacs for some time and is still in the most recent 2026.0 version, see [their code repository](https://gitlab.com/gromacs/gromacs/-/blob/96b77326ef002a98825b8ce477ca72be8c8a067a/cmake/gmxManageFFTLibraries.cmake#L83) and [installation guide](https://manual.gromacs.org/documentation/2026.0/install-guide/index.html#using-fftw).

oschuett:
Thanks for those pointers. The Gromacs folks certainly know a thing about performance.

However, it really doesn't fit my understanding. I think, the SSE2 kernel will always be outperformed by the AVX kernel because of the wider vector. So, enabling SSE2 might even slow down FFTW because it takes more time for the planing.

@hfp, what do you think?

Growl1234:
> The idea of building fftw with `--enable-sse2` flag has been suggested by gromacs for some time and is still in the most recent 2026.0 version, see [their code repository](https://gitlab.com/gromacs/gromacs/-/blob/96b77326ef002a98825b8ce477ca72be8c8a067a/cmake/gmxManageFFTLibraries.cmake#L83) and [installation guide](https://manual.gromacs.org/documentation/2026.0/install-guide/index.html#using-fftw).

In my opinion, the topic about performance should always be discussed with the specified program the tool is used with. Maybe GROMACS developers have their own considerations?

he-zilong:
Well, commits relevant to the linked gromacs warning about sse2 in fftw, like [d33720e6](https://gitlab.com/gromacs/gromacs/-/commit/d33720e6c87063242fa851ef7908b0720f1e80ae) and [ff03a699](https://gitlab.com/gromacs/gromacs/-/commit/ff03a6997c020742ba2cd32e8f1ed22ac601f747), date back to years ago. A quick search on [their forums](https://gromacs.bioexcel.eu) did not find more recent discussion on this part.

This PR is motivated not by performance concern, but by wish to suppress a warning when configuring the same fftw library installed in toolchain for cp2k and gromacs. Yet if it has no benefit for the majority of use cases it can be closed.

hfp:
> @hfp, what do you think?

Usually configure options like `enable-something` do not mean something else is disabled (e.g., AVX). It sounds indeed awkward to enable SSE2 specifically. Though, SSE2 is mandatory for 64-bit support on x86, e.g., anything previously x87 is SSE2 scalar instruction flavor (not necessary vectorized). I could imagine this configure option enables some handwritten/old intrinsic based code or certain kernels. I wonder what this would help over any kind of auto-vectorization.

hfp:
It could be interesting to look at https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/fftw/package.py if FFTW in CP2K would be in demand for improvement. I would prefer looking into the Spack recipe rather than enabling potentially less tested or old hand-written kernels.

Growl1234:
> Well, commits relevant to the linked gromacs warning about sse2 in fftw, like [d33720e6](https://gitlab.com/gromacs/gromacs/-/commit/d33720e6c87063242fa851ef7908b0720f1e80ae) and [ff03a699](https://gitlab.com/gromacs/gromacs/-/commit/ff03a6997c020742ba2cd32e8f1ed22ac601f747), date back to years ago. A quick search on [their forums](https://gromacs.bioexcel.eu) did not find more recent discussion on this part.

According to these commits, it seems that the point is about the coding of GROMACS itself, specifically, how GROMACS do FFT work with FFTW. As I understand, it is indeed not necessary for CP2K to require SSE2 flag with FFTW. 

> Usually configure options like `enable-something` do not mean something else is disabled (e.g., AVX).

Unfortunately, all these flags about instruction set have to be manually opened if needed when configuring FFTW.

Growl1234:
> This PR is motivated not by performance concern, but by wish to suppress a warning when configuring the same fftw library installed in toolchain for cp2k and gromacs. Yet if it has no benefit for the majority of use cases it can be closed.

Please also note that GROMACS mainly use single-precision version of FFTW (especially when using GPU acceleration it only supports single-precision version), which needs an additional `--enable-float` flag. If you use both CP2K and GROMACS in one machine, I think the best thing is to compile both single-precision (with the `--enable-sse2` flag recommended by GROMACS developers) and double-precision versions of FFTW in two different directories and let those programs use the proper one for each. Two versions are not conflicted with each other since the libraries of the single-precision version are named like `libfftw3_f.so` rather than `libfftw3.so`.

he-zilong:
> It could be interesting to look at https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/fftw/package.py if FFTW in CP2K would be in demand for improvement.

Here SSE2 is in the pool of considered SIMD supports, but it should appear in `spec.target` to be enabled in configuring.
```python
simd_features = ["sse2", "avx", "avx2", "avx512", "avx-128-fma", "kcvi", "vsx", "asimd"]
```
> If you use both CP2K and GROMACS in one machine, I think the best thing is to compile both single-precision and double-precision versions of FFTW in two different directories and let those programs use the proper one for each.

Good point. That would not work, though, for those who want cp2k QM/MM interface with gromacs, as the [installation guide](https://manual.gromacs.org/2026.0/install-guide/index.html#building-with-cp2k-qm-mm-support) requires explicitly that "FFT, BLAS and LAPACK libraries should be the same between CP2K and GROMACS".

Growl1234:
> That would not work, though, for those who want cp2k QM/MM interface with gromacs, as the [installation guide](https://manual.gromacs.org/2026.0/install-guide/index.html#building-with-cp2k-qm-mm-support) requires explicitly that "FFT, BLAS and LAPACK libraries should be the same between CP2K and GROMACS".

Thanks for the reminder. Since the manual says "Double precision in general is better than single for QM/MM" and the double-precision version "gmx_d" doesn't conflict with single-precision version, that's not a big problem. In this case, it maybe better when `--enable-sse2` is set during building of double-precision FFTW indeed.

oschuett:
To summarize, while this change won't increase performance it will improve the user experience for Gromacs+CP2K. 

That's good enough for me. Thanks for the discussion!

PR #4747, opened 2026-01-28
GW: performance improvements

MGraml:
In this PR, we introduce several changes to enable low-scaling GW (lsGW) to geometries with 10k+ atoms:
- adapt integers in dbm/dbm_multiply_comm.c to unsigned long int (overflow in earlier versions); after consultation with @oschuett 
- introduce all bounds to dbt_contract's in lsGW for improved scaling and overhead, based on sparsity of 3-center Coulomb integral
- fix some lsGW operations to be thread-safe
- add 2D mp_min/max in mpiwrap to handle 2D bound arrays needed in lsGW

### DIFF
--- a/src/dbm/dbm_multiply_comm.c
+++ b/src/dbm/dbm_multiply_comm.c
@@ -94,8 +94,9 @@ static void create_pack_plans(const bool trans_matrix, const bool trans_dist,
       for (int iblock = 0; iblock < shard->nblocks; iblock++) {
         const dbm_block_t *blk = &shard->blocks[iblock];
         const int sum_index = (trans_matrix) ? blk->row : blk->col;
-        const int itick = (1021 * sum_index) % nticks; // 1021 = a random prime
-        const int ipack = itick / dist_ticks->nranks;
+        unsigned long long itick64 = (unsigned long long)sum_index 
+                  * 1021ULL % (unsigned long long)nticks; // 1021 = a random prime
+        const int ipack = itick64 / dist_ticks->nranks;
         nblks_mythread[ipack]++;
       }
     }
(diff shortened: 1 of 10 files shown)


src/dbm/dbm_multiply_comm.c
```diff
@@ -94,8 +94,10 @@ static void create_pack_plans(const bool trans_matrix, const bool trans_dist,
       for (int iblock = 0; iblock < shard->nblocks; iblock++) {
         const dbm_block_t *blk = &shard->blocks[iblock];
         const int sum_index = (trans_matrix) ? blk->row : blk->col;
-        const int itick = (1021 * sum_index) % nticks; // 1021 = a random prime
-        const int ipack = itick / dist_ticks->nranks;
+        unsigned long long itick64 =
```
oschuett: Since this is now three lines instead of one, should we move it to a static inline function?

MGraml: Sure, if this fits the convention better.
So new lines 97-99 into a separate static inline function? Same replacement for lines 129-131?
Should I propose the change?

oschuett: Yes, just include it in this PR. Thanks!

src/dbm/dbm_multiply_comm.c
```diff
@@ -154,6 +150,12 @@ static void create_pack_plans(const bool trans_matrix, const bool trans_dist,
   } // end of omp parallel region
 }
 
+static inline unsigned long long calculate_tick_index(long sum_index,
```
oschuett: Nit: At the call site the `sum_index` and `nticks` are of type `int`. So depending on how smart the compiler is you might be doing two conversions: `int -> long -> unsigned long long`.

Maybe just define it as: 

`static inline unsigned long long  calculate_tick_index(int sum_index, int nticks) {...}`.

PR #4748, opened 2026-01-28
Toolchain: Remove redundant "export PKG_CONFIG_PATH" commands

Growl1234:
Related to the 3rd checkbox of issue #4667.

### DIFF
--- a/tools/toolchain/scripts/stage4/install_cosma.sh
+++ b/tools/toolchain/scripts/stage4/install_cosma.sh
@@ -256,7 +256,6 @@ prepend_path LIBRARY_PATH "${COSMA_LIBDIR}"
 prepend_path CPATH "$pkg_install_dir/include"
 export COSMA_INCLUDE_DIR="$pkg_install_dir/include"
 export COSMA_ROOT="${pkg_install_dir}"
-export PKG_CONFIG_PATH="$PKG_CONFIG_PATH:${COSMA_LIBDIR}/pkgconfig"
 prepend_path PKG_CONFIG_PATH "$pkg_install_dir/lib/pkgconfig"
 prepend_path CMAKE_PREFIX_PATH "$pkg_install_dir"
 EOF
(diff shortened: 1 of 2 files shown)


he-zilong:
I thought it was necessary to preserve the one `PKG_CONFIG_PATH` setting outside of the `if [ "$with_pkg" != "__SYSTEM__" ]; then ... ; fi` condition. Why is it that in the case of libvdwxc the inside one is preserved, and in the case of cosma both are deleted?

Growl1234:
> I thought it was necessary to preserve the one `PKG_CONFIG_PATH` setting outside of the `if [ "$with_pkg" != "__SYSTEM__" ]; then ... ; fi` condition. Why is it that in the case of libvdwxc the inside one is preserved, and in the case of cosma both are deleted?

What I deleted is the command "export", not "prepend_path", and `install_cosma.sh` already has one "prepend_path" command in the inside one. Furthermore, it's unnecessary to export PKG_CONFIG_PATH when searching package from system path, especially the case that CMake has replaced the old Makefile and arch files.

he-zilong:
> Furthermore, it's unnecessary to export PKG_CONFIG_PATH when searching package from system path, especially the case that CMake has replaced the old Makefile and arch files.

Is there a source of this statement? In that case, every occurrence of `prepend_path PKG_CONFIG_PATH` in the toolchain, regardless of inside or outside of the if condition, needs to be reviewed.

Growl1234:
> Is there a source of this statement? In that case, every occurrence of prepend_path PKG_CONFIG_PATH in the toolchain, regardless of inside or outside of the if condition, needs to be reviewed.

Maybe you're right. It's indeed strange that except for what's modified in this PR, all scripts exported PKG_CONFIG_PATH via `prepend_path` only inside that `if` condition. 
For now I'm still not sure if these command should be reviewed, though, but if so, maybe we could for the first consider those which may rely more on pkg-config such as DFT-D4? As far as I know, some programs such as HDF5 v1.14.6 will not have `lib/pkgconfig` if built with not CMake but `./configure` and `make`.

PR #4755, opened 2026-01-29
Toolchain: Several tweaks and fixes

Growl1234:
1. Print expected cmake options when running toolchain with `--dry-run`
2. Improve the message printed at last
3. Toolchain scripts: Fix the issue that `PKG_ROOT`, `PKG_CFLAGS` and `PKG_INCLUDES` are blank when finding from paths of self-installed PKG refers to FFTW, OpenBLAS and ScaLAPACK. (Some people may prefer compiling these basic math libraries by themselves rather than installing directly from package manager for better performance, in which case the installation paths of them may not be in `/usr` but paths like `/opt/openblas` and `/opt/fftw`. This fix ensure these paths can be read in install scripts of toolchain.)
4. Make `generate_cmake_options.sh` use `${CP2K_ROOT}/CMakeLists.txt` instead of `${CP2K_ROOT}/src/CMakeLists.txt` to fix regtest issue brought by operation 1

### DIFF
--- a/tools/toolchain/install_cp2k_toolchain.sh
+++ b/tools/toolchain/install_cp2k_toolchain.sh
@@ -1037,6 +1037,7 @@ write_toolchain_env ${INSTALLDIR}
 
 # write toolchain config
 echo "tool_list=\"${tool_list}\"" > ${INSTALLDIR}/toolchain.conf
+echo "dry_run=\"${dry_run}\"" >> ${INSTALLDIR}/toolchain.conf
 for ii in ${package_list}; do
   install_mode="$(eval echo \${with_${ii}})"
   echo "with_${ii}=\"${install_mode}\"" >> ${INSTALLDIR}/toolchain.conf
@@ -1047,6 +1048,7 @@ done
 # ------------------------------------------------------------------------
 if [ "${dry_run}" = "__TRUE__" ]; then
   echo "Wrote only configuration files (--dry-run)."
+  ./scripts/generate_cmake_options.sh
 else
   echo "# Leak suppressions" > ${INSTALLDIR}/lsan.supp
   ./scripts/stage0/install_stage0.sh
--- a/tools/toolchain/install_cp2k_toolchain.sh
+++ b/tools/toolchain/install_cp2k_toolchain.sh
@@ -1049,7 +1049,7 @@ done
 if [ "${dry_run}" = "__TRUE__" ]; then
   echo "Wrote only configuration files (--dry-run)."
   # Avoid error in Regtest_pdbg
-  if [ -n "$(pwd |grep toolchain)" ]; then
+  if [ -n "$(pwd |grep 'tools/toolchain')" ]; then
     ./scripts/generate_cmake_options.sh
   fi
 else
--- a/tools/toolchain/install_cp2k_toolchain.sh
+++ b/tools/toolchain/install_cp2k_toolchain.sh
@@ -1048,7 +1048,10 @@ done
 # ------------------------------------------------------------------------
 if [ "${dry_run}" = "__TRUE__" ]; then
   echo "Wrote only configuration files (--dry-run)."
-  ./scripts/generate_cmake_options.sh
+  # Avoid error in Regtest_pdbg
+  if [ -n "$(pwd |grep toolchain)" ]; then
+    ./scripts/generate_cmake_options.sh
+  fi
 else
   echo "# Leak suppressions" > ${INSTALLDIR}/lsan.supp
   ./scripts/stage0/install_stage0.sh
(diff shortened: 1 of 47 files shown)


tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -1047,6 +1048,10 @@ done
 # ------------------------------------------------------------------------
 if [ "${dry_run}" = "__TRUE__" ]; then
   echo "Wrote only configuration files (--dry-run)."
+  # Avoid error in Regtest_pdbg
+  if [ -n "$(pwd | grep 'tools/toolchain')" ]; then
+    ./scripts/generate_cmake_options.sh
```
oschuett: To fix this you'd have to add `generate_cmake_options.sh`  to [generate_dockerfiles.py:511](https://github.com/cp2k/cp2k/blob/5a17cb81585225b1879f458477d369e6221b0789/tools/docker/generate_dockerfiles.py#L511-L517) and then run script to re-generate all the Dockerfiles.

However, the `src/CMakeLists.txt` is not available that early in the docker build to improve caching. 

So, personally I'd just leave it as is and simply not call `generate_cmake_options.sh` during `--dry-run`.

he-zilong: Why not simply call the script from a path relative to the top-level `cp2k` directory, like `./tools/toolchain/scripts/generate_cmake_options.sh`, so that no if condition on the working directory is needed?

It is definitely nice to see `--dry-run` to output more information than current from a user's perspective!

Growl1234: > However, the src/CMakeLists.txt is not available that early in the docker build to improve caching.

To deal with this I can simply use `CMakeLists.txt` instead of `src/CMakeLists.txt` (if available). A small problem about this is there are several options that can be found in `src/CMakeLists.txt` but is not included in `CMakeLists.txt` (such as spFFT), and I'm not sure if simply editing `CMakeLists.txt` with adding the lacked options to the `option(CP2K_USE_xxx ... )` part have other affects.

Growl1234: The problems above should be solved with DockerFiles regenerated and `CMakeLists.txt` (in the root directory) modified. The description and build option of SpFFT library has been added to `docs/technologies/libraries.md` to prevent precommit failure.

oschuett:
It seems this PR has caused the _"Ubuntu, GCC * (ssmp)"_ testers on the [dashboard](https://dashboard.cp2k.org) to fail - for example [here](https://dashboard.cp2k.org/archive/gcc14/commit_57d0aac4b7e15b949c481a4be89b3d2d8ad5527e.txt).

Growl1234:
I'll have a look. I guess it's because the system-installed libopenblas in Ubuntu isn't named like "libopenblas.a" or "libopenblas.so". 
Maybe the similar changes about FFTW and ScaLAPACK should also be reviewed.

Growl1234:
Here is the reason: (based on my observations on Ubuntu 24.04.3 LTS)

Both `libopenblas.a` and `libopenblas.so` exist, but they are located in `/usr/lib/x86_64-linux-gnu`, which is not in the list of `SYS_LIB_PATH` defined in `scripts/toolkit.sh`. Therefore, this issue can be resolved with simply editing `scripts/toolkit.sh` and add `/usr/lib/x86_64-linux-gnu` to `SYS_LIB_PATH`. It's the same condition with libfftw3. 

As for ScaLAPACK, besides what's explained above, the system library of it is named `libscalapack-openmpi.so` or `libscalapack-mpich.so`, not `libscalapack.so`. So `scripts/stage4/install_scalapack.sh` also needs to be edited, changing "libscalapack.\*" in `find_in_paths` to "libscalapack\*" with the dot removed. 

That's all. If no further questions, I'll push these changes in a new PR shortly.

oschuett:
Thanks for the great analysis. Looking forward to your PR.

Growl1234:
Unfortunately, it turned out to be more complex than I had expected. 

It seems there is a compatibility issue in shell behavior between Ubuntu and Red Hat–based distributions (I test all my changes on Rocky Linux 10.1 I generally use), although the problems explained above do exist. 

What's strange when testing with Ubuntu is that these nested bash command works well in terminal but fails with a non-zero exit code when included in the script even if it starts with "`#!/bin/bash`". 

As a result, I have to fall back to `if` section so this problem can be avoided.

Growl1234:
> What's strange when testing with Ubuntu is that these nested bash command works well in terminal but fails with a non-zero exit code when included in the script even if it starts with "`#!/bin/bash`".

Well it's simply because the changes to `LIB_PATHS` in `tool_kit.sh` not being propagated to `common_vars.sh`.😅 (Fixing it in #4793)

PR #4766, opened 2026-02-02
Add NCCL support for CuSolverMP

sliwowitz:
CuSolverMp versions >= 0.7 have removed support for the Nvidia CAL communication library, and instead rely on NCCL. 

The new cuSolverMp versions are e.g. bundled with Nvidia HPC SDK 25.11+ with CUDA 13.0, but since you could install a more-or-less arbitrary version of cuSolverMp along with any CUDA version, we need to explicitly check against the actual installed cuSolverMp headers to get its version, and based on that decide if we're looking for CAL or NCCL dependencies.

More info about the migration here https://docs.nvidia.com/cuda/cusolvermp/usage/initialization/cal_to_nccl.html


oschuett:
Thanks for keeping the cusolvermp code up-to-date!

Just a few comments. To start, please update the year in the copyright banner and then run `make_pretty.sh`.

Furthermore, please add the preprocessor flag (`__CUSOLVERMP_NCCL`) to [cp2k_info.F](https://github.com/cp2k/cp2k/blob/e644ccc3b3697aa65a66613e7df3c8d2b4af7599/src/cp2k_info.F#L257) and document the CMake flag (`CP2K_CUSOLVERMP_USE_NCCL`) in [cusolvermp.md](https://github.com/cp2k/cp2k/blob/master/docs/technologies/eigensolvers/cusolvermp.md).


Unfortuantely, we can not yet test cusolvermp in the CI, but there is now a [Spack package](https://packages.spack.io/package.html?name=cusolvermp), which puts it within reach.

PR #4770, opened 2026-02-03
Toolchain: Decrease unnecessary environment variables for either building or running CP2K in `install/setup`

Growl1234:
Related to the comments in #4667

### DIFF
--- a/tools/toolchain/install_cp2k_toolchain.sh
+++ b/tools/toolchain/install_cp2k_toolchain.sh
@@ -231,10 +231,10 @@ The --with-PKG options follow the rules:
   --with-libsmeagol       Enable interface to SMEAGOL NEGF library (requires MPI).
                           Default = no
   --with-dftd4            Enable the DFTD4 package by Grimme.
-                          This package requires CMake and Ninja.
+                          This package requires CMake.
                           Default = install
   --with-tblite           Enable the tblite package by Grimme.
-                          This package requires CMake and Ninja.
+                          This package requires CMake.
                           Default = no
   --with-trexio           Enable the trexio library (read/write TREXIO files).
                           Default = no
@@ -827,13 +827,7 @@ if [ "${ENABLE_OPENCL}" = "__TRUE__" ]; then
   fi
 fi
 
-#dftd4 / tblite installation requires ninja
-if [ "${with_dftd4}" = "__INSTALL__" ] ||
-  [ "${with_tblite}" = "__INSTALL__" ]; then
-  [ "${with_ninja}" = "__DONTUSE__" ] && with_ninja="__INSTALL__"
-fi
-
-#tblite includes dftd4 - deactivate it
+# tblite includes dftd4 - deactivate it
 if [ "${with_tblite}" != "__DONTUSE__" ]; then
   with_dftd4="__DONTUSE__"
 fi
@@ -1048,7 +1042,6 @@ done
 # ------------------------------------------------------------------------
 if [ "${dry_run}" = "__TRUE__" ]; then
   echo "Wrote only configuration files (--dry-run)."
-  ./scripts/generate_cmake_options.sh
 else
   echo "# Leak suppressions" > ${INSTALLDIR}/lsan.supp
   ./scripts/stage0/install_stage0.sh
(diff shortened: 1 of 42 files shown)


Growl1234:
along with a small fix about writing mpi_f08 settings

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -1061,5 +1060,11 @@ else
   ./scripts/stage7/install_stage7.sh
   ./scripts/stage8/install_stage8.sh
   ./scripts/stage9/install_stage9.sh
-  ./scripts/generate_cmake_options.sh
+  # Decrease unnecessary environment variables for cmake in install/setup
+  grep -v "# \|FLAGS\|CPATH\|CP_LIBS" ${INSTALLDIR}/setup > ${INSTALLDIR}/setup.tmp
```
oschuett: I don't know. This seems hacky. While I realize that it's more work, I'd prefer to remove the unnecessary  variables at their origin.

tools/toolchain/scripts/stage1/install_openmpi.sh
```diff
@@ -150,7 +150,7 @@ export MPI_LIBS="${OPENMPI_LIBS}"
 export CP_DFLAGS="\${CP_DFLAGS} IF_MPI(-D__parallel|)"
 # For proper mpi_f08 support, we need at least GCC version 9 (asynchronous keyword)
 # Other compilers should work
-  if ! [ "$(gfortran -dumpversion | cut -d. -f1)" -lt 9 ]; then
+  if ! [ "\$(gfortran -dumpversion | cut -d. -f1)" -lt 9 ]; then
```
oschuett, line 150: Doesn't this backslash turn the whole expression into a literal string?

Have you tried removing the quotes altogether?

```suggestion
  if ! [ $(gfortran -dumpversion | cut -d. -f1) -lt 9 ]; then
```

Growl1234: Yes I've tried, and what's in build/setup_openmpi is: 
```
if ! [ 14 -lt 9 ]; then
```
That "\\" only works for the `cat` command, and what's written in the file doesn't contain it.

tools/toolchain/install_cp2k_toolchain.sh
Growl1234: I'm afraid that may affect some other files like `install/toolchain.env`, and I haven’t evaluated the impact of this yet. 

P.S. Maybe some files previously set to be genarated is no longer needed? If there are clearly better optimization approaches worth pursuing, I can consider converting this PR to a draft first and then exploring more substantial changes. Thank you for your attention!

Growl1234: I found that function `require_env` defined in the `scripts/tool_kit.sh` uses the file `install/toolchain.env`, and since `scripts/stage8/install_sirius.sh` needs a lot of flags via `require_env`, not all flags can be removed at their origin (even thogh they are not used by CP2K at all). Therefore, unless editing and testing `install_sirius.sh`, I'm afraid the similar command as above may still be needed in order to meet the expectation, but I generally do not use SIRIUS... In any case, I will give it a try.

Growl1234: > I'm afraid that may affect some other files like `install/toolchain.env`, and I haven’t evaluated the impact of this yet.

What I've found is indeed those variables will be written to `install/toolchain.env`, which will be loaded every time a new installation script is executed and may be used. Directly editing the `install_xxx.sh` and removing those  variables may cause problem during toolchain period. Therefore, for the moment, I still think just making `$SETUPFILE` differ from others may be a proper choice, as it strikes a balance between avoiding potential issues in the toolchain process and reducing the length of `$SETUPFILE` (making environment management easier).

oschuett:
Before we move on with this PR, I'd like to get the [dashboard](https://dashboard.cp2k.org) back to more green - including https://github.com/cp2k/cp2k/pull/4755#issuecomment-3852603384.

Growl1234:
It seems this regtest fails with a network issue. 😵‍💫

PR #4779, opened 2026-02-04
GAPW increase accuracy for XC integration

juerghutter:
This PR includes a method to increase the XC integration accuracy.
First tests show that the methods leads to improved overall accuracy when compared to other QC codes.

Example HF (TZVP+ all electron) PBE
Reference   -100.39347548
GAPW                  0.00002867
GAPW (new)      0.00000463

However, results are not fully consistent and probably need more testing and investigations.

Second, the method leads to "improved" forces. Meaning, it reduces the dependence of the forces on used PW grid.
This can be tested by looking at the total force (Sum of all forces on all atoms).

Example CH2O (DZVP + PP BLYP 600 Ry 
GPW                             2.31334368E-03
GAPW                           1.73036212E-05
GAPW ACCINT          2.28855308E-06

Example 32 water molecules
32H2O    300   60       4.45286768E-02   GPW
32H2O    300   60       1.15218495E-03   GPW spline3
32H2O    300  100      2.30603670E-03   GAPW
32H2O    300  100      1.63134523E-04   GAPW spline3
32H2O    300   60        1.13581628E-04   GAPW ACCINT
32H2O    600   60        5.84875218E-05   GAPW ACCINT
32H2O    800   80        6.13324476E-06   GAPW ACCINT


juerghutter:
This PR touches almost every QS method. I tested it for many energy/force methods but there is a large probability
that I have missed some combinations.

juerghutter:
How it works:

XC integral  in GPW $E_{xc} = \int F_{xc} dr \approx \sum_R F_{xc}(R)$ is approximated on the PW grid points R.

We define an atomic weight function $w_A(r)$ that is 1 at the nucleus and goes to 0 within the atomic radius. 
We will use a Gaussian for $w_A(r)$. Then we can write the XC integral as

$$
          E_{xc} = \int (1- \sum_A w_A(r)) F_{xc} dr + \sum_A \int w_A(r) F_{xc} dr 
$$

We now replace the XC integral over the smooth density in GAPW (or GAPW_XC) by the above 

$$
          E_{xc} = \int (1- \sum_A w_A(r)) F_{xc}[\tilde \rho] dr + \sum_A ( \int F_{xc}[\rho_A] dr - \int [1- w_A(r)] F_{xc}[\tilde \rho_A] dr )
$$

Potentials and nuclear forces have to be adjusted correspondingly.

PR #4785, opened 2026-02-05
HDF5: initialize output_unit in wrapper

hfp:
- Rely on cp_logger_get_default_io_unit()


hfp:
Testing with CP2K-CI / Spack (psmp, GCC 15).

PR #4786, opened 2026-02-05
TDDFT| delete outdated description

BelizSertcan:


BelizSertcan:
could the tests be triggered again or merged as is? the failure is due to slow test QS/regtest-stda-force-4/h2o_v01.inp

PR #4787, opened 2026-02-05
Fix Ubuntu GCC test issue

Growl1234:
Follow up to #4755 and #4779.

### DIFF
--- a/tests/TEST_DIRS
+++ b/tests/TEST_DIRS
@@ -193,10 +193,10 @@ QS/regtest-ps-implicit-1-1                                  fftw3
 QMMM/QS/regtest-cdft                                        parallel mpiranks%2==0
 DFTB/regtest-nonscc
 QS/regtest-acc-1                                            libint !ifx
-QS/regtest-acc-2                                            libint !ifx
+QS/regtest-acc-2                                            libint libxc !ifx
 QS/regtest-acc-3                                            libint !ifx
-QS/regtest-acc-4                                            libint !ifx
-QS/regtest-acc-5                                            libint !ifx
+QS/regtest-acc-4                                            libint libxc !ifx
+QS/regtest-acc-5                                            libint libxc !ifx
 QS/regtest-tddfpt                                           libint !ifx
 Fist/regtest-5-vib                                          mpiranks==1||mpiranks%2==0
 QS/regtest-tddfpt-lri                                       libint !ifx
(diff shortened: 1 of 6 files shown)


Growl1234:
Btw I have a small curiosity about some tests marked as needing "libint", because the Ubuntu regtests have set `--with-libint=no` but these test jobs still run and even succeed. 🤔

fstein93:
I think it was just copy-pasted but tests without the non-RI-HFX-code do not require libint.

Growl1234:
Oh I see that libint is in `/usr` path.
```
  - Libint2
    - include directories: /usr/include;/usr/include/libint2;/usr/include/eigen3
    - libraries: /usr/lib/x86_64-linux-gnu/libint2.so
```

tests/TEST_DIRS
```diff
@@ -192,12 +192,12 @@ QS/regtest-dm-ls-scf-1
...
-QS/regtest-acc-3                                            libint !ifx
-QS/regtest-acc-4                                            libint !ifx
-QS/regtest-acc-5                                            libint !ifx
-QS/regtest-tddfpt                                           libint !ifx
+QS/regtest-acc-1                                            libint libxc !ifx
+QS/regtest-acc-2                                            libint libxc !ifx
+QS/regtest-acc-3                                            libint libxc !ifx
+QS/regtest-acc-4                                            libint libxc !ifx
+QS/regtest-acc-5                                            libint libxc !ifx
+QS/regtest-tddfpt                                           libint libxc !ifx
```
oschuett: The `regtest-tddfpt` used to work without libxc.


As for [the others](https://github.com/search?q=repo%3Acp2k%2Fcp2k+path%3A%2F%5Etests%5C%2FQS%5C%2Fregtest-acc-.%5C%2F%2F+%2F%28LDA%7CGGA%7CMGGA%7CHYB%29_%2F&type=code), it seems libxc is really only used by `regtest-acc-2` and `regtest-acc-5`.

oschuett:
> Btw I have a small curiosity about some tests marked as needing "libint", because the Ubuntu regtests have set --with-libint=no but these test jobs still run and even succeed.

The idea behind these _"Ubuntu/GCC"_ tests is to pull all dependencies from the Ubuntu distribution instead of using the Toolchain. However, when we switch to CMake we started using the Toolchain after all, but only to build DBCSR.

tests/TEST_DIRS
Growl1234: > it seems libxc is really only used by `regtest-acc-2` and `regtest-acc-5`.

`/opt/cp2k/regtesting/TEST-2026-02-06_09-21-52/QS/regtest-acc-4/ch2o_dip48.inp.out` also fails due to not having libXC. After modifying, there are `regtest-acc-2`, `regtest-acc-4`, and `regtest-acc-5`, which need libXC.

PR #4788, opened 2026-02-05
low-scaling GW: fix blocksize-dependent deviations

MGraml:
This PR fixes a deviation in G0W0 gaps computed from the low-scaling GW module for different user-specified computational resource parameters.
The bug is related to the sizes of atom blocks, which are influenced, among other things, by the input keyword `MEMORY_PER_PROC` and the number of MPI ranks.

### DIFF
--- a/src/gw_large_cell_gamma.F
+++ b/src/gw_large_cell_gamma.F
@@ -7,6 +7,8 @@
 
 ! **************************************************************************************************
 !> \brief Routines from paper [Graml2024]
+!> \par History
+!>      01.2026 Maximilian Graml: add more bounds to exploit sparsity in 3c integrals, fixes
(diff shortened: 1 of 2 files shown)


oschuett:
Is there a regtest we could add to catch this kind of bugs in the future? We recently started [testing with 4 MPI ranks](https://dashboard.cp2k.org/archive/psmp-4ranks/index.html).

MGraml:
I also thought about that, but I don't see an easy way to create such a test.
We have seen this bug only for geometries of 1000+ atoms, where sparsity comes into play, and on top of that variations of computational resources typical for HPC systems. One of these calculations takes around 1500 coreh. For smaller geometries, I was not able to reproduce the bug.

PR #4791, opened 2026-02-06
Toolchain: mark segments of setup file by source

he-zilong:
Related to fourth item in the task list of #4667: a delimiter marking start of
configurations from each individual setup files is now written to the overall
setup file so as to make it more human readable and dependency handling easier.
A variable WHAT is now used to refer to the package name in each install script.

### DIFF
--- a/tools/toolchain/install_cp2k_toolchain.sh
+++ b/tools/toolchain/install_cp2k_toolchain.sh
@@ -900,15 +900,6 @@ export CP_CFLAGS=""
 export CP_LDFLAGS="-Wl,--enable-new-dtags"
 
 # ------------------------------------------------------------------------
-# Start writing setup file
-# ------------------------------------------------------------------------
-cat << EOF > "$SETUPFILE"
-#!/bin/bash
-source "${SCRIPTDIR}/tool_kit.sh"
-export CP2K_TOOLCHAIN_OPTIONS="${TOOLCHAIN_OPTIONS}"
-EOF
-
-# ------------------------------------------------------------------------
 # Special settings for CRAY Linux Environment (CLE)
 # TODO: CLE should be handle like gcc or Intel using a with_cray flag and
 #       this section should be moved to a separate file install_cray.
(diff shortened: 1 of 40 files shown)


he-zilong:
I also noted two details:
* In `scripts/stage6/install_libtorch.sh`, one line says `if [ "$ENABLE_CUDA" = "__TRUE__" ]` but the difference between conditions is not clear to me. For now the line that writes to `setup_libtorch` is taken out of the if statements.
* In `scripts/stage8/install_sirius.sh`, one line says `echo 'with_sirius="__FALSE__"' >> ${BUILDDIR}/setup_sirius` if MPI is disabled and SIRIUS is not built. Why is this the only instance to set a `with_something` variable as `__FALSE__` in the setup file, is it to be read or tested later?

Growl1234:
I do not fully understand the purpose of explicitly adding `${WHAT}`. This does not appear to introduce practical difference.

Another question: Is there a follow-up plan for you to generate a more minimal environment setup file?

he-zilong:
Yes, this PR does not contain substantial change, exactly as intended; apart from some additional comments as delimiter, nothing in the overall setup file is different from before. The motive is to help whoever may need to read or process the setup file for dependency management and figure out what happens when it gets sourced in the shell. (In some scripts there are further edits to the individual `setup_something` files after its contents gets copied to the overall setup file.)

The `${WHAT}` variable is simply to ensure that the name of the package is used consistently in the headers for each script. Also all instances of the same header can now be found by something like `grep "Linking \${WHAT} to user paths" ./tools/toolchain/ -nr`.

As for figuring out what flags are necessary for what stage, that could be the subject of other PRs.

Growl1234:
> As for figuring out what flags are necessary for what stage, that could be the subject of other PRs.

All flags are not necessary for building CP2K (but may used during toolchain buildings), and that's what I want to do in #4770.

Growl1234:
> In `scripts/stage6/install_libtorch.sh`, one line says `if [ "$ENABLE_CUDA" = "__TRUE__" ]` but the difference between conditions is not clear to me. For now the line that writes to `setup_libtorch` is taken out of the if statements.

This does look redundant, as `ENABLE_CUDA` is already taken into consideration earlier in this script:
```
if [ "$ENABLE_CUDA" = "__TRUE__" ]; then
  LIBTORCH_LDFLAGS="-Wl,--no-as-needed,-L'${LIBTORCH_LIBDIR}' -Wl,--no-as-needed,-rpath='${LIBTORCH_LIBDIR}'"
else
  LIBTORCH_LDFLAGS="-L'${LIBTORCH_LIBDIR}' -Wl,-rpath='${LIBTORCH_LIBDIR}'"
fi
```
 I’ve just pushed the corresponding change to my PR.

Growl1234:
> The `${WHAT}` variable is simply to ensure that the name of the package is used consistently in the headers for each script. 

I don't know. It doesn't seem to make things easier in my view. 

> Also all instances of the same header can now be found by something like `grep "Linking \${WHAT} to user paths" ./tools/toolchain/ -nr`.

Why not just `grep "Linking" ./tools/toolchain/ -nr`? 🤔

he-zilong:
I am closing this PR due to having been superseded by already merged #4770 in which the `filter_setup()` function in `tool_kit.sh` is modified to achieve the same effect. However, my concern above about the purpose of `echo 'with_sirius="__FALSE__"' >> ${BUILDDIR}/setup_sirius` remains unaddressed.

PR #4794, opened 2026-02-07
Fix OpenMP in qs_moment_kpoints

oschuett:
Follow up to #4738.

I noticed that `ispin` was declared as shared, which is most likely a bug. Hence, I've now moved it to private.

CC: @ShridharShanbhag


ShridharShanbhag:
Thank you for noticing this bug and fixing it.

Issue #4795, opened 2026-02-07
CP2K compilation with Intel oneAPI


Growl1234:
After PR #4799, CP2K can now be compiled successfully with Intel oneAPI 2025.2.1 (base Toolkit with MKL, and HPC Toolkit with icx, icpx, ifx and MPI library). 

However, during a test of cell optimization task, I found the Intel-compiled CP2K runs not as fast as what's compiled using GCC with OpenMPI or MPICH. Even more strangely, there occurs a memory issue as shown below. (increasing slowly and abnormally) When I try falling Intel oneAPI version back to 2025.2.1, the memory leak still occurs.
<img width="1209" height="324" alt="Screenshot From 2026-02-07 22-05-21" src="https://github.com/user-attachments/assets/65ebcd3b-892b-48ab-8f59-4a215dc73168" />
While here's the condition with GCC/OpenMPI, which is normal and uses only half of the memory compared to the Intel-compiled one: 
<img width="1209" height="324" alt="Screenshot From 2026-02-07 22-06-35" src="https://github.com/user-attachments/assets/5964663f-87bf-48a6-aaf3-a97212d2c199" />

It's also worth noticing that regtests with ifx are currently disabled with jobs requiring libint for runtime failure and wrong results, and optimize level of several files has to be reduced.

Therefore, in addition to compilation-related issues, I would like to further discuss whether there might be some compatibility issues between the current CP2K code and oneAPI.

Growl1234:
I also wonder whether what these maeesages say during building affects (actually the similar warning appears not only once):
```
/opt/software/cp2k-2026.1-intel/build/src/mpiwrap/message_passing.F(23176): warning #5474: Statement in an OpenMP PARALLEL WORKSHARE construct cannot be parallelized; the statement will only be executed by one thread of one team.
            base(:) = win_data(disp_aint + 1:disp_aint + len)
------------^
/opt/software/cp2k-2026.1-intel/build/src/mpiwrap/message_passing.F(27887): warning #5474: Statement in an OpenMP PARALLEL WORKSHARE construct cannot be parallelized; the statement will only be executed by one thread of one team.
            base(:) = win_data(disp_aint + 1:disp_aint + len)
------------^
/opt/software/cp2k-2026.1-intel/build/src/mpiwrap/message_passing.F(32598): warning #5474: Statement in an OpenMP PARALLEL WORKSHARE construct cannot be parallelized; the statement will only be executed by one thread of one team.
[...]

PR #4799, opened 2026-02-08
Attempts to resolve compilation issue with oneAPI(ifx)

Growl1234:
Changes settings about linking in `src/CMakeLists.txt` to fix linker error; 
Makes building with oneAPI use "-xHOST" so that the highest instruction set supported by CPU can be detected and used; 
Along with a small attempt regarding a performance issue described in #4795.


Growl1234:
Based on my tests, it seems not easy to resolve performance issue with running oneAPI-compiled CP2K (with ifx). It may require changes to the Fortran source code in some components? While I am not confident about this. 
Currently I'm not sure if the editing of `install_mkl.sh` really helps, although linking to `-lmkl_gf_lp64` when using Intel compiler does not seem appropriate.

oschuett:
Thanks a lot for taking on the Intel compilers!

In the past we also had to [lower the optimization level](https://github.com/cp2k/cp2k/blob/2c79fd46c42d57064f04908c3383fba7e3810ab4/tools/toolchain/scripts/generate_arch_files.sh#L397-L408) on some files to pass the regtests.

CC: @hfp

Growl1234:
Unfortunately, both regtests seem to stay failing in `dbt_unittest` regardless of how I adjust the optimization flags, along with some other tests in regtest task. I am now more inclined to think that this may be due to a bug of ifx/ifort itself.

Growl1234:
And here's the issue when I try to debug that why `dbt_unittest` fails to run via gdb: 
```
dbt_unittest.psmp: /opt/software/cp2k-2026.1-intel/src/dbm/dbm_multiply.c:329: void dbm_multiply(const _Bool, const _Bool, const double, const dbm_matrix_t *, const dbm_matrix_t *, const double, dbm_matrix_t *, const _Bool, const double, int64_t *): Assertion `num_sum_index_a == num_sum_index_b' failed.

Program received signal SIGABRT, Aborted.
0x00001555406bef8c in __pthread_kill_implementation () from /lib64/libc.so.6
(gdb) bt
#0  0x00001555406bef8c in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x0000155540668f56 in raise () from /lib64/libc.so.6
#2  0x00001555406508fa in abort () from /lib64/libc.so.6
#3  0x000015554065081e in __assert_fail_base.cold () from /lib64/libc.so.6
#4  0x00001555406614b6 in __assert_fail () from /lib64/libc.so.6
#5  0x000015554e8a29b1 in dbm_multiply (transa=false, transb=true, alpha=1, matrix_a=0x643f80, matrix_b=0x686350, beta=0, 
    matrix_c=0x66c410, retain_sparsity=false, filter_eps=9.9999999999999998e-13, flop=0x7ffffffe0360)
    at /opt/software/cp2k-2026.1-intel/src/dbm/dbm_multiply.c:329
#6  0x000015554d911699 in dbm_api::dbm_multiply (transa=.FALSE., transb=.TRUE., alpha=1, matrix_a=..., matrix_b=..., beta=0, 
    matrix_c=..., retain_sparsity=.FALSE., filter_eps=6.9533558014172892e-310, flop=0)
    at /opt/software/cp2k-2026.1-intel/build/src/dbm/dbm_api.F:788
#7  0x000015554d9bef57 in dbt_tas_mm::dbt_tas_multiply (transa=.FALSE., transb=<optimized out>, transc=<optimized out>, alpha=1, 
    matrix_a=..., matrix_b=..., beta=1.1588901064496588e-310, matrix_c=..., optimize_dist=.FALSE., split_opt=..., 
    filter_eps=1.1588901064472872e-310, flop=0, move_data_a=.TRUE., move_data_b=.TRUE., retain_sparsity=.FALSE., simple_split=.TRUE., 
    unit_nr=0, log_verbose=.FALSE.) at /opt/software/cp2k-2026.1-intel/build/src/dbt/tas/dbt_tas_mm.F:677
#8  0x000015554d9c694d in dbt_tas_mm::dbt_tas_estimate_result_nze (transa=.FALSE., transb=.TRUE., transc=.TRUE., matrix_a=..., 
    matrix_b=..., matrix_c=..., estimated_nze=140737488234664, filter_eps=1.1588901064472872e-310, unit_nr=0, retain_sparsity=.FALSE.)
    at /opt/software/cp2k-2026.1-intel/build/src/dbt/tas/dbt_tas_mm.F:1435
#9  0x000015554d9bd0c2 in dbt_tas_mm::dbt_tas_multiply (transa=.FALSE., transb=.TRUE., transc=.TRUE., alpha=0.90000000000000002, 
    matrix_a=..., matrix_b=..., beta=2.5790384813919739e-317, matrix_c=..., optimize_dist=.FALSE., split_opt=..., 
    filter_eps=1.1588901064472872e-310, flop=0, move_data_a=.TRUE., move_data_b=.TRUE., retain_sparsity=.FALSE., 
    simple_split=.FALSE., unit_nr=-111836, log_verbose=.TRUE.) at /opt/software/cp2k-2026.1-intel/build/src/dbt/tas/dbt_tas_mm.F:262
#10 0x000015554d940d31 in dbt_methods::dbt_contract_expert (alpha=0.90000000000000002, tensor_1=..., tensor_2=..., 
    beta=0.10000000000000001, tensor_3=..., contract_1=..., notcontract_1=..., contract_2=..., notcontract_2=..., map_1=..., 
    map_2=..., bounds_1=..., bounds_2=..., bounds_3=..., optimize_dist=.FALSE., pgrid_opt_1=0x0, pgrid_opt_2=0x0, pgrid_opt_3=0x0, 
    filter_eps=1.1588901064472872e-310, flop=0, move_data=.FALSE., retain_sparsity=.FALSE., nblks_local=0, unit_nr=-92716, 
    log_verbose=.TRUE.) at /opt/software/cp2k-2026.1-intel/build/src/dbt/dbt_methods.F:904
#11 0x000015554d93bb76 in dbt_methods::dbt_contract (alpha=0.90000000000000002, tensor_1=..., tensor_2=..., beta=0.10000000000000001, 
    tensor_3=..., contract_1=..., notcontract_1=..., contract_2=..., notcontract_2=..., map_1=..., map_2=..., bounds_1=..., 
--Type <RET> for more, q to quit, c to continue without paging--c
    bounds_2=..., bounds_3=..., optimize_dist=.FALSE., pgrid_opt_1=0x0, pgrid_opt_2=0x0, pgrid_opt_3=0x0, 
    filter_eps=1.1588901064472872e-310, flop=0, move_data=.FALSE., retain_sparsity=.FALSE., unit_nr=-92716, log_verbose=.TRUE.)
    at /opt/software/cp2k-2026.1-intel/build/src/dbt/dbt_methods.F:533
#12 0x000015554d982da6 in dbt_test::dbt_contract_test (alpha=0.90000000000000002, tensor_1=..., tensor_2=..., 
    beta=0.10000000000000001, tensor_3=..., contract_1=..., 
    notcontract_1=<error reading variable: value requires 93822685571712 bytes, which is more than max-value-size>, contract_2=..., 
    notcontract_2=..., map_1=<error reading variable: value requires 27189737926880 bytes, which is more than max-value-size>, 
    map_2=<error reading variable: value requires 22730301077660 bytes, which is more than max-value-size>, unit_nr=-82948, 
    bounds_1=..., bounds_2=..., bounds_3=..., log_verbose=.TRUE., write_int=.TRUE.)
    at /opt/software/cp2k-2026.1-intel/build/src/dbt/dbt_test.F:938
#13 0x000000000042869e in dbt_unittest () at /opt/software/cp2k-2026.1-intel/build/src/dbt/dbt_unittest.F:428
```

Growl1234:
After fully disabling optimizations (setting "-O0") for certain DBT modules, the test passed on my machine. However, the regtests show several “WRONG RESULT”s, with some deviations being quite large: 
```
>>> /opt/software/cp2k-2026.1-intel/tests/regtesting/TEST-2026-02-09_12-31-26/TMC/regtest_ana_on_the_fly
    TMC_ana_density.inp                                                                   0.1233121927           OK (   0.87 sec)
    TMC_ana_G_R.inp                                                                     0.002560696691           OK (   0.86 sec)
    TMC_ana_dip_cl.inp                                                                    0.1259847557           OK (   0.85 sec)
    TMC_ana_deviation.inp                                                                 0.3151914052           OK (   0.89 sec)
    TMC_ana_create_traj_without_ana.inp                                                              -           OK (   0.85 sec)
    TMC_ana_start_with_exist_traj.inp                                                      3.135784688 WRONG RESULT (   0.87 sec)
    TMC_ana_restart.inp                                                                    2.574210419 WRONG RESULT (   0.84 sec)
```
And here's output `TMC_ana_start_with_exist_traj.inp.out`, when seems to be due to Intel compilers:
```
 -                                TMC setting                                  -
 -                                distribution of cores                        -
 TMC| number of all working groups                          1                   
 TMC| number of groups (ener|cc)             1                    |      0      
 TMC| cores per group (ener|cc)              1                    |      0      
 TMC| Analysis groups                                       1                   
 TMC| Temperature(s) [K]                                  300.00  380.00  460.00
 TMC| # of Monte Carlo Chain elements:                    100                   
 TMC| exact potential input file:                                        H2O.inp
 TMC| Pressure [bar]:                                   1.0000E+00              
 TMC| Numbers of atoms/molecules moved 
 TMC|   within one conf. change                             0                   

  type  temperature  |     trajec   PT(swap) atom trans  mol trans    mol rot   
   volume
    I       I        |   weight->      0.429      0.143      0.143      0.143   
    0.143
    V       V        |   size  ->        XXX      0.050      0.100      4.000   
    0.100

 TMC| restarting at Markov Chain element(s):           78          30
          50          34
 TMC|first_global_tree_rnd_nr_X=   5.905896245601220E-003

 *** WARNING in tmc/tmc_worker.F:401 :: analysis old trajectory up to elem ***
 ***     30                   . Read trajectory file.                      ***

 TMC_ANA| read xyz file                             ./tmc_trajectory_T300.00.xyz
 TMC_ANA| read cell file                           ./tmc_trajectory_T300.00.cell

 *** WARNING in tmc/tmc_file_io.F:646 :: end of position file reached at ***
 *** line  4.5700E+02              , last element     30                 ***


 *** WARNING in tmc/tmc_worker.F:401 :: analysis old trajectory up to elem ***
 ***     50                   . Read trajectory file.                      ***

 TMC_ANA| read xyz file                             ./tmc_trajectory_T380.00.xyz
 TMC_ANA| read cell file                           ./tmc_trajectory_T380.00.cell

 *** WARNING in tmc/tmc_file_io.F:646 :: end of position file reached at ***
 *** line  1.2490E+03              , last element     50                 ***


 *** WARNING in tmc/tmc_worker.F:401 :: analysis old trajectory up to elem ***
 ***     34                   . Read trajectory file.                      ***

 TMC_ANA| read xyz file                             ./tmc_trajectory_T460.00.xyz
 TMC_ANA| read cell file                           ./tmc_trajectory_T460.00.cell

 *** WARNING in tmc/tmc_file_io.F:646 :: end of position file reached at ***
 *** line  9.6100E+02              , last element     34                 ***


 ===============================================================================
 Total energy:   0.525942531884051     

 ===============================================================================
 =                                finalizing TMC                               =
 acceptance rates:
  type  temperature  |     trajec   PT(swap) atom trans  mol trans    mol rot   
   volume
 probs  T=    300.00 |      0.552      0.724      0.167      1.000      0.667   
    0.143
 counts T=    300.00 |         58         29         12          4          6   
        7
 nr_acc T=    300.00 |         32         21          2          4          4   
        1
 probs  T=    380.00 |      0.657      0.775      0.333      0.500      0.857   
    0.429
 counts T=    380.00 |         70         40         12          4          7   
        7
 nr_acc T=    380.00 |         46         31          4          2          6   
        3
 probs  T=    460.00 |      0.483        ---      0.111      1.000      0.750   
    0.500
 counts T=    460.00 |         29          0          9          4          4   
       12
 nr_acc T=    460.00 |         14          0          1          4          3   
        6

 Efficiencies:
 (MC elements/calculated configuration) global: 1.91 sub tree(s):  2.15 2.59 1.04
 (MC elements/energy calculated configuration): 1.78 0.66 1.12 0.79
 computing time of one Markov chain element       0.00
 TMC run time[s]:       0.04

 ===============================================================================
 stopping workers
```

I also noticed that pushing changes sometimes results in certain manually enabled tests in CP2K CI being automatically turned off, and I don't know why this happens.

Growl1234:
Another curiosity: Above testings are in my modified cp2k-2026.1 package (with all changes of toolchain scripts and cmake files), but when I switch to development branch they all fail with `SIGNAL: 7 (Bus error)` or `SIGNAL: 11 (Segmentation fault)`. 
Even my own previous testing task shown in #4795 fails with above error when using development branch. It's quite strange.

Growl1234:
Unfortunately, further attempts to resolve the issue by tweaking compilation options turned out to be much more difficult than expected and do not seem to address the root cause. I have therefore reverted those experimental attempts and kept only the initial changes that ensure the build process itself succeeds (even though the resulting binaries are entirely not usable yet).

At this point, the remaining investigation likely requires deeper insight into the compiler or source-level behavior, which is beyond what I can confidently handle. I really appreciate the developers’ continued efforts in this area.

If everyone agrees that the remaining initial changes are still meaningful, I will mark this PR as ready for review so that we can bring it to a timely conclusion.

Growl1234:
It seems the segmentation fault has to do with the changes in #4789. During string concatenation (doc = doc // ...), the length of doc seems to be misinterpreted as a memory address.
```
(gdb) run
Starting program: /opt/dev/cp2k-intel/build/bin/libcp2k_unittest.psmp 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Unit test starts ...
Testing cp_c_get_version(): CP2K version 2026.1 (Development Version).

Program received signal SIGSEGV, Segmentation fault.
0x000015554c5095bf in fparser::docf (
    doc=<error reading variable: value requires 140737488164360 bytes, which is more than max-value-size>)
    at /opt/dev/cp2k-intel/build/src/common/fparser.F:1160
1160	      doc = doc//"A functional form is specified. Mathematical Operators recognized are "
(gdb) bt
#0  0x000015554c5095bf in fparser::docf (
    doc=<error reading variable: value requires 140737488164360 bytes, which is more than max-value-size>)
    at /opt/dev/cp2k-intel/build/src/common/fparser.F:1160
#1  0x000015554aae9b00 in input_cp2k_mm::create_genpot_section (section=0x0) at /opt/dev/cp2k-intel/build/src/input_cp2k_mm.F:1304
#2  0x000015554ab693cb in input_cp2k_tb::create_xtb_nonbonded_section (section=0x15552c312200)
    at /opt/dev/cp2k-intel/build/src/input_cp2k_tb.F:527
#3  0x000015554ab65ef8 in input_cp2k_tb::create_xtb_control_section (section=0x15552c325800)
    at /opt/dev/cp2k-intel/build/src/input_cp2k_tb.F:142
#4  0x000015554aa9bacd in input_cp2k_qs::create_qs_section (section=0x15552c33df80)
    at /opt/dev/cp2k-intel/build/src/input_cp2k_qs.F:520
#5  0x000015554aa51d8d in input_cp2k_dft::create_dft_section (section=0x155530505d00)
    at /opt/dev/cp2k-intel/build/src/input_cp2k_dft.F:364
#6  0x000015554aab9cd1 in input_cp2k_force_eval::create_force_eval_section (section=0x15553054e200)
    at /opt/dev/cp2k-intel/build/src/input_cp2k_force_eval.F:143
#7  0x000015554d0b9ac3 in input_cp2k::create_cp2k_root_section (root_section=0x1555303ae980)
    at /opt/dev/cp2k-intel/build/src/start/input_cp2k.F:110
#8  0x000015554d0d96aa in libcp2k::cp2k_create_force_env (new_env_id=0, input_file_path=..., output_file_path=...)
    at /opt/dev/cp2k-intel/build/src/start/libcp2k.F:158
#9  0x0000000000407c77 in main () at /opt/dev/cp2k-intel/src/start/libcp2k_unittest.c:78
```

And this does not seem to be something that can be resolved simply by adjusting optimization flags.

Growl1234:
This segmentation fault should now be fixed. To address the issue, I had to replace allocatable deferred-length string with fixed-length buffer and manual append routine However, I would appreciate it if @SebastianWellig could take a look at [src/common/fparser.F](https://github.com/cp2k/cp2k/pull/4799/changes/95214649b143b0b31fed3579d0ebd6142b8d4568#diff-2b8a23e61fb664c60bb42278c45b6d99f71615e4c790edcedd11f4c1ae9bbc06) and confirm whether the code after modification aligns with your original intention.

In addition, there are still a few regtests on my machine that produce unreasonable results (as mentioned in the previous comment). I will investigate this further.

@oschuett, could you please re-enable the Intel regtests? Thank you very much!

Growl1234:
I was really excited to see that the two Intel regtests have passed! This suggests that, at least for most typical use cases, CP2K compiled with Intel oneAPI appears to work reasonably well. However, the performance and memory usage issues, as well as the “wrong results” observed in the TMC test mentioned above, is remaining unresolved. Further investigation and experimentation would still be worthwhile.

I will make one final attempt about something like switching the optimization level back to `-O3` (I found what really matters seems to be `-fp-model;consistent`) and see whether the tests still pass (they pass on my machine). If everything looks acceptable, perhaps this PR could be concluded for now, as it has already been open for quite some time.

Growl1234:
Strangely, the Intel regtests seem to have disappeared again.

Here is the log from a regtest I just ran locally (oneAPI 2025.2.1 with ifx). It may serve as a reference or supporting evidence to some extent.

[regtest.log](https://github.com/user-attachments/files/25290482/regtest.log)

oschuett:
> Strangely, the Intel regtests seem to have disappeared again.

This happens when you force-push. It's due to how the Github API for check-runs works and our CI not having its own data store.

src/common/fparser.F
```diff
@@ -1148,27 +1148,44 @@ END FUNCTION evalfd
...
-      !----- -------- --------- --------- --------- --------- --------- --------- -------
-      CHARACTER(LEN=:), ALLOCATABLE                      :: doc
-
-      INTEGER                                            :: i
+      ! Manually allocate length to avoid segmentation fault with Intel compilers
+      CHARACTER(LEN=3000)    :: doc
+      INTEGER                :: i, pos
 
-!----- -------- --------- --------- --------- --------- --------- --------- -------
+      doc = ''
```
SebastianWellig: Did you check whether just adding this line wouldn't fix it already, as it will initialize the variable, which I forgot?

Growl1234: Maybe you're right, this line should make the following `doc // ...` legal. 

I’ll test it on my machine tomorrow. While I personally prefer the modified version as it seems cleaner and more robust to me, I fully respect your approach and would like to minimize changes to your contribution, provided that testing confirms that adding just that one line is sufficient.

SebastianWellig: Or better just remove `doc//` on line 1160 to initialize it there.

Growl1234: > Or better just remove `doc//` on line 1160 to initialize it there.

That works on my machine. Thank you!

Growl1234:
When doing regtests of psmp with ifort (oneAPI 2024.2.1) on my machine, `mpiexec` fails to start because of unrecognized argument `--bind-to`, while 2025.2.1 with ifx does not occur:

https://github.com/cp2k/cp2k/blob/9ea93394cc22b7813d728ab445506cebcf5c4567/tests/do_regtest.py#L70

However, I also found this `--bind-to none` is needed according to #3191 and the issues it fixed. It maybe better to make `--bind-to` an extra command that is only applied when not using older versions of oneAPI, or find a similar oprion instead.

src/common/fparser.F
SebastianWellig: Glad to hear. Thanks for fixing my mistake!

oschuett:
> When doing regtests of psmp with ifort (oneAPI 2024.2.1) on my machine, mpiexec fails to start because of unrecognized argument --bind-to, while 2025.2.1 with ifx does not occur:

Yeah, that one is really difficult to get right.

We could either give [do_regtest.py](https://github.com/cp2k/cp2k/blob/master/tests/do_regtest.py) smarter defaults or handle this in the calling CI script [test_regtest.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_regtest.sh).

Growl1234:
> We could either give [do_regtest.py](https://github.com/cp2k/cp2k/blob/master/tests/do_regtest.py) smarter defaults

Yes and this is what I try to do in commit [5b94b23](https://github.com/cp2k/cp2k/pull/4799/commits/5b94b238b0afafdab9d0f8dabae6f636b0e6bdec), though it's just for Intel MPI.

oschuett:
Great work!

oschuett:
It seems this PR somehow broke the [Spack psmp build](https://dashboard.cp2k.org/archive/spack-psmp/index.html).

Growl1234:
Strange. It seems that failure is because libtorch is not found when starting regtests, but this PR has nothing to do with any behavior about searching these libraries. 🤔

oschuett:
Admittedly, I'm a bit puzzled too. This might very well be a false positive. Unfortunately, our CI is not 100% hermetic.

@mkrack, what do you think?

mkrack:
> Admittedly, I'm a bit puzzled too. This might very well be a false positive. Unfortunately, our CI is not 100% hermetic.
> 
> @mkrack, what do you think?

I don't know what is going. I will test locally the master version and also send a PR to get further inside. 
We should consider that Spack and toolchain still share the CMake configuration.

mkrack:
> > Admittedly, I'm a bit puzzled too. This might very well be a false positive. Unfortunately, our CI is not 100% hermetic.
> > @mkrack, what do you think?
> 
> I don't know what is going. I will test locally the master version and also send a PR to get further inside. We should consider that Spack and toolchain still share the CMake configuration.

The current [spack_psmp test](https://storage.googleapis.com/cp2k-ci/run-cp2k-spack-psmp-44f53134_report.txt) seems to work.

PR #4804, opened 2026-02-09
Fix for tblite to not use OT optimizer

juerghutter:
Let's tblite (gfn1 tested only) to use OT optimizer.

@jpoto The code at lines 527 in qs_scf.F looks to me odd. No other method needs updates at that point in the SCF loop.

### DIFF
--- a/src/qs_environment.F
+++ b/src/qs_environment.F
@@ -133,6 +133,7 @@ MODULE qs_environment
                                               write_structure_data
    USE particle_types,                  ONLY: particle_type
    USE pw_env_types,                    ONLY: pw_env_type
+   USE pyscon,                          ONLY: kelvin
    USE qmmm_types_low,                  ONLY: qmmm_env_qm_type
    USE qs_basis_rotation_methods,       ONLY: qs_basis_rotation
    USE qs_dftb_parameters,              ONLY: qs_dftb_param_init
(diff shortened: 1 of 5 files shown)


jpoto:
I introduced the code in lines 527 in qs_scf.F because there was an update of the charges necessary during one iteration to get the same convergence as in the reference calculation. In the case of a nearly converged computation the update doesn't produce a relevant change of the charges. 

We could try to remove it, but then initial convergence needs to be tested. 

As for finding a different location, I would have to run some tests.

jpoto:
Should we add some tests?

juerghutter:
Yes, some tests are needed. Maybe also check what options of tblite work with OT?

I did some quick tests for gfn1 and gfn2, both seem to work.

@jpoto If you prepare a test set, I would merge now .

jpoto:
I have the tests in https://github.com/jpoto/cp2k/tree/jvp_gfn_ot_tests
but it will probably be easier if you merge first.

juerghutter:
Issue #4818 has a valid point. tblite Hamiltonians are spin-free. The tblite library treats
all cases using restricted orbitals and non-uniform occupation numbers.
This collides with CP2K handling occupation numbers in the optimizer part.

I suggest to solve the problem in a different way then done in this PR
- Disable OT for tblite
- Disable LSD for tblite
- Enforce ADDED_MOS -1 for tblite (make all orbitals available)

jpoto:
Should we leave the OT option for closed shell systems?

juerghutter:
I don't know the internals of tblite, but if we use OT the number of MO's is given by  (number of electrons)/2. As Fermi smearing
is default, this will cause differences to the standard usage, especially for systems with small HOMO-LUMO gap.
It's not so bad (it's the xTB default in CP2K) but I would keep it consistent with the external reference, at least for the moment.

PR #4810, opened 2026-02-10
CMake: Restore creation of popt/sopt symlink

Growl1234:
Previously, the popt/sopt symlink was created by the old GNU Makefile, but this step was missing after the transition to CMake. This change restores that behavior.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1259,3 +1259,20 @@ install(
   DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/cp2k"
   FILES_MATCHING
   PATTERN "*.cmake")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1259,3 +1259,20 @@ install(
   DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/cp2k"
   FILES_MATCHING
   PATTERN "*.cmake")
+
+# Create symlink: psmp -> popt, ssmp -> sopt
```
oschuett:
```suggestion
# Create symlink: popt -> psmp, sopt -> ssmp
```

Issue #4792, opened 2026-02-06
Toolchain: Available Updates of Dependencies


Growl1234:
Here I would like to list several dependency updates that are available to be built via toolchain, along with my own perspective. Some of them can be directly implemented into toolchain scripts, while some have to wait. 

- [x] **MPICH** The new major version of MPICH, 5.0.0, has released just these days. According to my own tests, a few common jobs work fine but most of regtests fail, which is found to be a bug of MPICH and I've submit a PR https://github.com/pmodels/mpich/pull/7722 to MPICH repo to fix it. May next version of MPICH can fix it.
- [x] **OpenBLAS** 0.3.31, needs nothing but simply to update the version and SHA256 in toolchain script.
- [x] **HDF5** has a new major version 2.0.0 which doesn't seem to have breaking changes that inffect CP2K use. However, current 2.0.0 version has some bugs about searching and showing libraries supported, which have been resolved in current develop branch (See https://github.com/HDFGroup/hdf5/pull/6152 and https://github.com/HDFGroup/hdf5/pull/6191). Therefore, I think it's better to wait for a new [2.1.0 release](https://github.com/HDFGroup/hdf5/milestone/17). 
- [x] **DFT-D4** 4.0.0 has some breaking changes which makes CP2K incompatible with it. To solve this, `qs_dispersion_d4.F` has to be modified, which in turn makes CP2K incompatible with older versions of dftd4. Based on a simple try I did earlier, at least for current D4 model, it would be quite easy (see [this fork branch](https://github.com/Growl1234/cp2k/tree/dftd4-v4) for detail); as for new D4S model, I have no idea if this is very meaningful in implementating for CP2K, but if so, I guess it woundn't be hard. However, the latest version of **tblite**, 0.5.0, still uses DFT-D4 v3.7.0, so a new version of tblite has to be waited for before putting this update on the agenda.
***P.S. I have a small question: As shown in https://github.com/dftd4/dftd4/issues/329, the performance of DFT-D4 in MPI-parallel CP2K calculations is rather limited, and there's no plan from the side of dftd4 developers for MPI implementation, so I wonder if there's a way to directly push DFT-D4 code to the CP2K source with MPI support, like DFT-D3.***
- [x] **SIRIUS** Since I generally don't use this tool, I have no idea what effect does this upgrade have. Given that https://www.cp2k.org/static/downloads/ has already had this new version package, I guess there may still be some issues preventing the new version from being applied to CP2K?
[...]

PR #4812, opened 2026-02-10
Toolchain: Updates of OpenBLAS, PLUMED and Spglib

Growl1234:
OpenBLAS: 0.3.30 -> 0.3.31
PLUMED: 2.9.3 -> 2.10.0
Spglib: 2.5.0 -> 2.7.0

As mentioned in #4792.


Growl1234:
Actually, I am looking forward to these new package versions being uploaded to the CP2K website, as modifying the download URLs is just a temporary workaround (if they are not on CP2K website) to check whether this change would trigger regtest failures.

oschuett:
Voilà:
https://www.cp2k.org/static/downloads/OpenBLAS-0.3.31.tar.gz
https://www.cp2k.org/static/downloads/plumed-src-2.10.0.tgz
https://www.cp2k.org/static/downloads/spglib-2.7.0.tar.gz

PR #4841, opened 2026-02-16
Attempt to enable DFT+U for k-point calculations

Growl1234:
The current version of CP2K does not yet support DFT+U calculations with k-point sampling. As far as I understand, though, DFT+U is essentially an on-site correction. Therefore, its correction behavior should, in principle, be the same whether k-points are considered or only the Gamma point is used. Extending DFT+U to support k-points should thus not introduce conceptual difficulties.

This PR adds a `do_kpoints` branching in the DFT+U code path and enables DFT+U to use the 2D accessor to implement the correction scheme under k-point sampling. The Gamma-point implementation keeps the original approach, otherwise it leads to segmentation faults in practice.

I am new to Fortran, and this is my first attempt at making substantial modifications to the main CP2K source code. If anything is incorrect or inconsistent with the design intentions, I would greatly appreciate any corrections or suggestions.

### DIFF
--- a/src/dft_plus_u.F
+++ b/src/dft_plus_u.F
@@ -220,7 +220,7 @@ CONTAINS
       INTEGER, DIMENSION(1)                              :: iloc
       INTEGER, DIMENSION(:), POINTER                     :: atom_list, nshell, orbitals
       INTEGER, DIMENSION(:, :), POINTER                  :: first_sgf, l, last_sgf
-      LOGICAL                                            :: debug, dft_plus_u_atom, &
+      LOGICAL                                            :: debug, dft_plus_u_atom, do_kpoints, &
                                                             fm_work1_local_alloc, &
                                                             fm_work2_local_alloc, found, &
                                                             just_energy, smear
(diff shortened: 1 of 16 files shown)


oschuett:
@juerghutter, could you review this one?

garrettwx:
Dear devs,

If this enabling of DFT+U with k-points supports the PLUS_U_METHODS LOWDIN for &GEO_OPT and &CELL_OPT?

It seems that the previous version supports the LOWDIN method only for experimental forces calculation and not support for stress_tensor calculation?

Growl1234:
> Dear devs,
> 
> If this enabling of DFT+U with k-points supports the PLUS_U_METHODS LOWDIN for &GEO_OPT and &CELL_OPT?
> 
> It seems that the previous version supports the LOWDIN method only for experimental forces calculation and not support for stress_tensor calculation?

Unfortunately, this problem is not solved yet. This PR is just to solve problem directly about k-points, and stress tensor still cannot be calculated using Löwdin method.

juerghutter:
I'm very busy right now. If you can wait until next week, I can have a look.

garrettwx:
> Unfortunately, this problem is not solved yet. This PR is just to solve problem directly about k-points, and stress tensor still cannot be calculated using Löwdin method.

Thank u dev for your reply, and for the experimental version of LOWDIN for force calculations, this is also important for &GEO_OPT.

juerghutter:
I added PR #4855 as an alternative route to get this feature implemented.
The way proposed in the current PR seems to me not valid.

Growl1234:
> I added PR #4855 as an alternative route to get this feature implemented. The way proposed in the current PR seems to me not valid.

Hi Prof. Hutter,

Great to see this proper implementation! Also looking forward to seeing more of the remaining k-point–related issues being resolved. 

P.S. In my own previous tests, the approach seemed to work reasonably well for MnO₂ and CeO₂ (with correct band gaps and reasonable energy scaling, latter compared with Gamma-only supercell calculation), which are typical strongly correlated systems. However, it now appears that I had overlooked several important aspects. May I ask what fundamental or theoretical considerations, in the actual implementation, prevent this approach from working properly? (I have read the explanation in the new PR, but I would really appreciate some further clarification.)

PR #4819, opened 2026-02-11
Toolchain: revamp help message, extend --dry-run output in install_cp2k_toolchain.sh

he-zilong:
This PR proposes several revisions to the toolchain script including:

- rewritten and reorganized help message, with options ordered as in stages and more details clarified;
- extended printed output of --dry-run (see below);
- other general modifications so that most line fit in 80 characters and only three shellcheck items remain.

An example of current `--dry-run` output:
```shell
$ ./install_cp2k_toolchain.sh --dry-run -j 8 --with-gcc=system --mpi-mode=openmpi --with-sirius=no
No MPI installation detected. (Ignore this message if
    a fresh MPI installation is requested.)
WARNING: (./install_cp2k_toolchain.sh, line 926) CMake is used for installing one of the packages.
Toolchain script received the following options:
  --dry-run -j 8 --with-gcc=system --mpi-mode=openmpi --with-sirius=no
System specifications:
   -j                    = 8
  --target-cpu           = native
  --gpu-ver              = no
  --mpi-mode             = openmpi
  --math-mode            = openblas
List of effective settings after resolving package conflicts:
  --with-gcc             = __SYSTEM__
  --with-intel           = __DONTUSE__
  --with-amd             = __DONTUSE__
  --with-cmake           = __INSTALL__
[omitted]
  --with-mcl             = __DONTUSE__
Suggested cmake command if toolchain is built with your options: 
  cmake .. -DCMAKE_INSTALL_PREFIX=../install -DCP2K_DATA_DIR=/home/twilight/cp2k-dev/data -DCP2K_USE_MPI=ON -DCP2K_USE_FFTW3=ON -DCP2K_USE_LIBINT2=ON -DCP2K_USE_LIBXC=ON -DCP2K_USE_LIBXSMM=ON -DCP2K_USE_COSMA=ON -DCP2K_USE_ELPA=ON -DCP2K_USE_SPGLIB=ON -DCP2K_USE_VORI=ON
```

PR #4846, opened 2026-02-16
Toolchain: remove some legacy mentions

he-zilong:
Following #4819 where reference to `arch_base.tmpl` is removed due to commit 00026b9, two other legacy mentions are spotted:

- `checksums.sha256` is removed due to commit dc59bb0;
- `reflapack` is removed due to commit 907ab74.

Also, a link to the manual about building with Spack is added.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -30,9 +30,6 @@ or run in a subshell with:
 Note: it is recommended to install podman to take advantage of a spack cache. This will accelerate
 the build of the CP2K dependencies with Spack significantly.
 
-For more about building CP2K with Spack, see
-[manual](https://manual.cp2k.org/trunk/getting-started/build-with-spack.html).
-
 ### Method 2: Toolchain
 
 Alternatively, the [toolchain script](./tools/toolchain/install_cp2k_toolchain.sh) can also be run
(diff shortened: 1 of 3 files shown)


Growl1234:
That page, https://manual.cp2k.org/trunk/getting-started/build-with-spack.html, is actually not the same with what make_cp2k.sh does. The latter is still essentially building from scratch, with dependencies got from Spack rather than toolchain.

he-zilong:
> That page, https://manual.cp2k.org/trunk/getting-started/build-with-spack.html, is actually not the same with what make_cp2k.sh does. The latter is still essentially building from scratch, with dependencies got from Spack rather than toolchain.

Oops, thanks for the correction. Leave the job of mentioning "Build with Spack" and "Install from Distribution" in `INSTALL.md` for later.

Growl1234:
> Leave the job of mentioning "Build with Spack" and "Install from Distribution" in `INSTALL.md` for later.

For method 1 which uses `make_cp2k.sh`, I think we can simply add a below line in `INSTALL.md`: 
> Use `./make_cp2k.sh --help` to see available options and usage information.

PR #4848, opened 2026-02-17
Minor fixes about `make_cp2k.sh` and RPM-build

Growl1234:
Reset default libdir in `CMakeLists.txt`, following up to #4830; only set it as "lib" in configuration if CP2K is built with `make_cp2k.sh`.

Along with minor fixes to `make_cp2k.sh`: 
* Update help message
* Detect real physical CPU cores (avoid `nproc`, which reports logical cores and may be affected by `OMP_NUM_THREADS`)
* COSMA is not only used by SIRIUS; moving it to a different subgroup to prevent it from being unintentionally excluded from installation
* Fix inconsistency between `-df` / `-ef` feature flags and corresponding options in `CMakeLists.txt`
* Manually disabling one of `libvdwxc | spfft | spla | sirius` seems to at the same time disable other package of them in original script; moving them to a new group in order to actually meet this behavior for CMake step
* Run `cmake --build "${CMAKE_BUILD_PATH}" --target clean` after installation in order to save space


Growl1234:
There's another small question: when I run `make_cp2k.sh` with `-ue`, it fails due to conflict with system-installed Python 3.12 (while Spack seems to obtain its own one with the same version) unless I add another flag `-dlc`. I wonder if there's a solvation, or how we can prompt the user in the help message.

make_cp2k.sh
```diff
@@ -142,8 +142,8 @@ HAS_PODMAN="no"
 HELP="no"
 INSTALL_MESSAGE="NEVER"
 MPI_MODE="mpich"
-if command -v nproc &> /dev/null; then
-  MAX_PROCS=$(nproc)
+if command -v lscpu &> /dev/null; then
+  MAX_PROCS="$(lscpu -p=Core,Socket | grep -v '#' | sort -u | wc -l)"
```
mkrack: see above

make_cp2k.sh
```diff
@@ -265,18 +258,28 @@ while [[ $# -gt 0 ]]; do
...
                 ;;
-              smeagol | vori)
+              vori)
                 SED_PATTERN_LIST+=" -e '/\s*-\s+\"lib${2,,}@/ ${SUBST}"
                 ;;
             esac
             ;;
+          libvdwxc | spfft | spla | sirius)
+            echo "WARNING: You've enabled or disabled one of libvdwxc/spfft/spla/sirius, so"
+            echo "         the rest packages of them will also be enabled or disabled!"
```
mkrack: "the other packages will also be enabled or disabled!"

mkrack:
> There's another small question: when I run `make_cp2k.sh` with `-ue`, it fails due to conflict with system-installed Python 3.12 (while Spack seems to obtain its own one with the same version) unless I add another flag `-dlc`. I wonder if there's a solvation, or how we can prompt the user in the help message.

That's the reason why externals, i.e. the packages from the host system are ignored (so far possible) by default. Spack comes with its own Python. This worked for me so far when such problems occur. The need for `-dlc` which disables the installation of the python venv and thus of the boto3 module required for the spack cache disables caching. Your  host Python 3.12 should be new enough and I think it would the best if the user could fix the Python installation on the host system. Does your host system python provide python-venv and python-pip?

Growl1234:
> Does your host system python provide python-venv and python-pip?

Yes. My system is Rocky Linux 10.1. It provides python-pip that can be installed via `dnf install pip` or `dnf install python3-pip`. `Python-venv` is a built-in component included in the system-required `python3-libs` package.

mkrack:
> > Does your host system python provide python-venv and python-pip?
> 
> Yes. My system is Rocky Linux 10.1. It provides python-pip that can be installed via `dnf install pip` or `dnf install python3-pip`. `Python-venv` is a built-in component included in the system-required `python3-libs` package.

Maybe, it would make sense to include also a Rocky Linux tester in addition to ubuntu, fedora, and opensuse to understand if this is a problem caused by the base system or `make_cp2k.sh`.

Growl1234:
> Maybe, it would make sense to include also a Rocky Linux tester in addition to ubuntu, fedora, and opensuse to understand if this is a problem caused by the base system or `make_cp2k.sh`.

Maybe this is a good idea. In fact, Rocky Linux is a bug-for-bug compatible rebuild of RHEL, a downstream that's derived from Fedora. In principle, its behavior should therefore be quite similar to Fedora’s. Moreover, due to its near enterprise-grade stability, the likelihood of unexpected issues should be even lower.

Growl1234:
Another question: would it be possible to support `-ue` in combination with some other fundamental dependencies such as MPI and math libraries? This could be advantageous for some large-scale HPC systems.

mkrack:
> > Maybe, it would make sense to include also a Rocky Linux tester in addition to ubuntu, fedora, and opensuse to understand if this is a problem caused by the base system or `make_cp2k.sh`.
> 
> Maybe this is a good idea. In fact, Rocky Linux is a bug-for-bug compatible rebuild of RHEL, a downstream that's derived from Fedora. In principle, its behavior should therefore be quite similar to Fedora’s. Moreover, due to its near enterprise-grade stability, the likelihood of unexpected issues should be even lower.

I have added a Rocky Linux tester in PR #4852. I selected the rockylinux:9.3 base image from Dockerhub which is not that recent. Suggestions for a better choice are welcome.

mkrack:
> Another question: would it be possible to support `-ue` in combination with some other fundamental dependencies such as MPI and math libraries? This could be advantageous for some large-scale HPC systems.

Indeed, the `-ue` flag allows spack to detect with `spack external find` and `spack compiler find` libraries and binaries on the host system and spack will use them for the build. You can give spack hints in the configuration file if needed where to find these libraries or binaries.

oschuett:
The new test for Rocky Linux shows the same [4 failures](https://storage.googleapis.com/cp2k-ci/run-cp2k-spack-psmp-rockylinux-294899bb_report.txt) as on the [dashboard](https://dashboard.cp2k.org/archive/spack-psmp-rockylinux/index.html). So is this PR ready to be merged then?

Growl1234:
This regtest has exhibited these issues since its initial introduction. I am not sure about the exact cause, but it appears not to be related to this PR. 

Based on my own experience with Rocky Linux (9.5/9.6/10.0/10.1), might explicitly setting `ulimit -s unlimited` and increasing `OMP_STACKSIZE` (if not done) might help resolve them? (I set these options as defaults in my `.bashrc`, and I have never encountered this kind of error.)

From my side, the PR is ready to be merged.

mkrack:
The regression test is run with `ulimit -s unlimited` and `OMP_STACKSIZE=64M` (tried up to 256M without success). Did you ever try Rocky Linux 9.3?

mkrack:
> From my side, the PR is ready to be merged.

This PR is indeed not related to the test failure with Rocky Linux, but did you address my review comments?

Growl1234:
> The regression test is run with `ulimit -s unlimited` and `OMP_STACKSIZE=64M` (tried up to 256M without success). Did you ever try Rocky Linux 9.3?

I sincerely apologize for any inconvenience this may have caused. I am currently running tests on a Rocky Linux 9.3 system.

Based on the Rocky Linux release notes, the only potentially relevant changes in 9.5 appear to be minor updates to packages such as GCC and GDB, although this should not imply that the test suite would fail on 9.3.

PR #4853, opened 2026-02-18
Try to fix Misc error occurred in #4841

Growl1234:
This might be due to version updates of the packages installed in step 18, as well as the addition of some new packages.
I briefly had a look at the `test_misc.sh` script, and these actions seem to happen automatically.


oschuett:
Thanks a lot for looking into this! I guess, I should have pinned the versions of a [all packages](https://github.com/cp2k/cp2k/blob/master/tools/pao-ml/requirements.txt).

Growl1234:
> Thanks a lot for looking into this! I guess, I should have pinned the versions of a [all packages](https://github.com/cp2k/cp2k/blob/master/tools/pao-ml/requirements.txt).

In principle, this is indeed how it should work. At the very least, we should leave room to monitor version changes and update the related code accordingly. 
Interestingly, though, it seems that this particular update might turn out to be be beneficial? The extra comment, which was previously added to make things work with torch 2.6.0 (#3935), now appears to be removable, and the error message is generated actually due to this.

oschuett:
> In principle, this is indeed how it should work.

You're right. Then please go ahead add pin all the packages.

Growl1234:
> You're right. Then please go ahead add pin all the packages.

Done. It might be worth rerunning the Misc test in #4841 after merging this PR to see how it goes.

Issue #2050, opened 2022-04-01
CDFT (and LDA+U) k-point sampling support?


zc62:
CDFT and LDA+U seem not to have k-point sampling support. Is there any source that I can learn from to understand the challenge behind it? (i.e., the challenge to generalize the Gamma-point KS potential to each k-point KS potential.)

I am especially curious about the k-point implementation in CP2K, (for example, how is the real-space density matrix obtained?)
https://aip.scitation.org/doi/10.1063/5.0007045 says:
> An extension of Eq. (9) to include a k-point sampling within the first Brillouin zone is also available in CP2K.

but left without any references. Maybe there are some unpublished developers' notes?

juerghutter:
Hi

CP2K uses for k-points equivalent methods to the ones used by Gaussian and Crystal codes.
The only difference is the handling of the Coulomb terms (GPW).

See one of the many reviews by the Crystal people or the paper by Kudin and Scuseria:
PhysRevB.61.16440 (2000)

I don't know if the generalization of the CDFT method is feasible, but I guess it would be
rather hard. LDA+U should be doable without to much effort.

regards

Juerg Hutter

________________________________________
From: zc62 ***@***.***>
Sent: Saturday, April 2, 2022 12:02 AM
To: cp2k/cp2k
Cc: Subscribed
Subject: [cp2k/cp2k] CDFT (and LDA+U) k-point sampling support? (Issue #2050)

CDFT and LDA+U seem not to have k-point sampling support. Is there any source that I can learn from to understand the challenge behind it? (i.e., the challenge to generalize the Gamma-point KS potential to each k-point KS potential.)

I am especially curious about the k-point implementation in CP2K, (for example, how is the real-space density matrix obtained?)
https://aip.scitation.org/doi/10.1063/5.0007045 says:

An extension of Eq. (9) to include a k-point sampling within the first Brillouin zone is also available in CP2K.

but left without any references. Maybe there are some unpublished developers' notes?

he-zilong:
Let's retire this issue in favor of the recent dedicated #4854.

Issue #2076, opened 2022-04-22
force evalution for DFT+U method LOWDIN 


GengSS:
Hi,
I received an warning about the force calculations during DFT+U method LOWDIN

` 
*** WARNING in dft_plus_u.F:791 :: This is an experimental version of the ***

 *** forces calculation for the DFT+U method LOWDIN                        ***
`

So, my question is that should I trust the results from CP2K with this warning? I would like to do some tests, however the DFT+U is really dependent on the implementatin details  (basis sets, project etc), and it is not clear how to make the tests realiably.

Thanks
Best,
Geng

Growl1234:
Loewdin force should now work with PR #4855 applied.

PR #4855, opened 2026-02-18
DFT+U with k-points, only MULLIKEN option

juerghutter:
This PR shows a possible route to DFT+U with k-points.
It introduces the necessary generalizations in the basic matrices
and applys the loop over cell images for the MULLIKEN option.

For MULLIKEN_CHARGES a slighly more complex change is necessary with a first loop
to calclulate the charges and then a second one for the potential and forces.

For LOWDIN the case is even more complicated as it is required to apply the orthogonalization
in k space.


Growl1234:
Is the issue with Löwdin due to the calculation of `S^(1/2)`? I suspect the reason why it's not implemented for Löwdin method to calculate stress tensor is also this. 🤔

mkrack:
This PR looks much more reasonable, because it has a k point loop which calculates the orbital occupation matrix and thus the Hubbard potential contribution at each k point while #4841 includes only the gamma point. The latter might show the expected effects for strongly correlated materials with small k point grids given the dominance of the gamma point contribution in such cases, but PR #4841 is not correct in my opinion.
Nevertheless, this PR needs also some testing and validation for systems like NiO or UO2.

Growl1234:
It seems to be still blocked by `Method not implemented for k-points` abort message due to `qs_scf_initialization.F`. Maybe [such change](https://github.com/Growl1234/cp2k/blob/8084f870d269756a934b9798c1e20a1d922702a4/src/qs_scf_initialization.F#L1097-L1101) is still needed? (with `matrix_s_kp` included [here](https://github.com/Growl1234/cp2k/blob/8084f870d269756a934b9798c1e20a1d922702a4/src/qs_scf_initialization.F#L919) and [here](https://github.com/Growl1234/cp2k/blob/8084f870d269756a934b9798c1e20a1d922702a4/src/qs_scf_initialization.F#L930); since this is outside the `IF (dft_control%plus_u_method_id == plus_u_lowdin)` section, it does no harm with what Loewdin method behaviors)

Growl1234:
> Nevertheless, this PR needs also some testing and validation for systems like NiO or UO2.

When doing test with NiO, I received below warning:
```
 *** WARNING in dft_plus_u.F:1485 :: DFT+U energy contribution is negative ***
 *** possibly due to unphysical Mulliken charges!                          ***
```
and SCF is hard to converge. 

This is my input file: [NiO.inp.txt](https://github.com/user-attachments/files/25406470/NiO.inp.txt). The input structure is got [here](https://next-gen.materialsproject.org/materials/mp-19009). I also tried reducing U_eff to 1.5 eV, and the warning still appears during the first few steps, although the SCF eventually converges.

With UO2 it seems harder to make SCF converge in current diagonalization method. 

I tested CeO2 (with a unit cell size and stacking similar to UO2, lattice parameter 5.411 Å, but non-magnetic) using the literature-recommended Ueff of 4.08 eV and a 6×6×6 k-point grid. The results matched quite well.

P.S. The DFT+U correction energy differs from my previous test by roughly 3e-5, which is above the SCF convergence threshold (5e-6). This further confirms that my earlier approach was indeed not correct and may give a wrong result for much smaller cells.

garrettwx:
Hi, if you try NiOOH system, it may even get harder to converge by using standard diag method, as I have tested before. I have tried many settings. And, after some cycles of optimization (via BFGS), I even notice the strange location of adsorbed *H site. The same structure opted by VASP is safe. 

So here, this is why I guess, if cp2k can support more diag SCF methods beyond standard way, like DAVIDSON, LANCZOS....

Growl1234:
> Hi, if you try NiOOH system, it may even get harder to converge by using standard diag method, as I have tested before. I have tried many settings.

Yes. However, this is a separate issue, and it might not be appropriate to discuss it in detail here. The main focus here should be on ensuring that the current DFT+U with k-points is implemented correctly and accurately, which is sufficient for now.

That said, I am also curious about how challenging it would be to implement k-point support for non-standard diagonalization methods, which is, of course, more suitable for a separate thread, such as the one you opened.

garrettwx:
> > Hi, if you try NiOOH system, it may even get harder to converge by using standard diag method, as I have tested before. I have tried many settings.
> 
> Yes. However, this is a separate issue, and it might not be appropriate to discuss it in detail here. The main focus here should be on ensuring that the current DFT+U with k-points is implemented correctly and accurately, which is sufficient for now.
> 
> That said, I am also curious about how challenging it would be to implement k-point support for non-standard diagonalization methods, which is, of course, more suitable for a separate thread, such as the one you opened.

Yes, I mean, I have also add the DFT+U , and the use of MULLIKEN, and the same warning of yours occurs. So afterwards, I then echo for the completeness of LOWDIN, perhaps the mulliken_charge is more safe?

Growl1234:
> Yes, I mean, I have also add the DFT+U , and the use of MULLIKEN, and the same warning of yours occurs. So afterwards, I then echo for the completeness of LOWDIN, perhaps the mulliken_charge is more safe?

I still think Loewdin method is the safest, though (maybe) `S^(1/2)` makes it complicated for some issues. I haven't tried `MULLIKEN_CHARGES` method and am not sure if it performs better than Mulliken method.

juerghutter:
MULLIKEN_CHARGES option added to kpoints
Fixed some tests within the code.

garrettwx:
> Is the issue with Löwdin due to the calculation of `S^(1/2)`? I suspect the reason why it's not implemented for Löwdin method to calculate stress tensor is also this. 🤔

As I notice that for the LOWDIN tag in DFT+U, it says DFT+U methods based on Lowdin charges need S^(1/2). It seems that the spin contamination for k-points is yet to be solved. Maybe this is another question.

juerghutter:
Lowdin forces should now work. What is left is the k-point version for Lowdin energy+forces.

What do you mean by spin contamination for k-points is yet to be solved?

garrettwx:
> What do you mean by spin contamination for k-points is yet to be solved?

Sorry for my unclear description. I mean the :
*** WARNING in qs_scf_post_gpw.F :: Spin contamination estimate not ***
 *** implemented for k-points.

he-zilong:
Could this PR close #2050 and address #2076 once merged?

Growl1234:
> Could this PR close #2050 and address #2076 once merged?

I think #2076 could be addressed, but in #2050 there's also an issue about CDFT with k-points.

juerghutter:
Version merged:
- Kpoint enabled DFT+U Mulliken option
- Kpoint enabled DFT+U Milliken_charges option
- Forces for DFT+U Lowdin option

Still missing:  kpoint version of DFT+U Lowdin version

Issue #4818, opened 2026-02-11
TBLITE with CP2K v2026.1


mlacount91:
I was interested in using TBLITE to preform some MD calculations using GFN2. Some tests I did revealed 2 problems

1) TBLITE with GFN2 assigns 6 electrons per nitrogen rather than 5. I was not able to determine whether this is a problem with CP2K or TBLITE, a workaround I found was to set the net charge of the system to remove the extra electrons. I wasn't extensive with my testing and other elements may suffer the same issue.

2) SPIN_POLARIZED (UKS) with TBLITE is broken. Using N2 (un-optimized 1A bond length) as a test system with GFN1 I found the following energies:
CP2K-GFN1: -6.3307570163 Ha
CP2K-TBLITE-GFN1: -6.3307570728 Ha
CP2K-GFN1-UKS: -6.3307570163 Ha
CP2K-TBLITE-GFN1-UKS: -6.2912730795 Ha
The UKS TBLITE is clearly different than the other results. In addition, when performing a geometry optimization the force is wrong and the optimization simply increases the bond length by exactly 0.25A each step (each step going up in energy) until it eventually fails an SCF calculation.

Is UKS with TBLITE intentionally unsupported? If so I'd request a flag informing the user to be added to the code. I couldn't find any documentation suggesting UKS with TBLITE shouldn't work.

jpoto:
The occupation numbers are from tblite. I prepared a pull request with a simple fix.

DCM-Uni-Paderborn:
This appears to be addressed by #4804 and #4857:

- #4804 disables unsupported LSD/UKS and OT usage for TBLITE and enforces the orbital setup needed by TBLITE.
- #4857 fixes the partial occupation numbers from TBLITE and adds an NH3/GFN2 regression test for the nitrogen occupation issue.

Closing this as resolved. Please reopen if there is still a remaining TBLITE/UKS or occupation-number case not covered by these fixes.

PR #4857, opened 2026-02-18
fix for partial occupation numbers from tblite

jpoto:
fix for issue #4818


jpoto:
@juerghutter Any comments?

PR #4858, opened 2026-02-19
MO wavefunction text output for k-point calculations

Growl1234:
This is an attempt to enable CP2K to write wavefunction information of MOs to a text file `.mokp` when k-points are used, somehow similar to what `.molden` does for Gamma-point calculations. The generated `.mokp` file can give cell vectors and kpoint information at head, and then provide detailed MO information of each k-point: coefficients matrix with both real and imaginary parts, and detailed information (GTO basis definition or overlap matrix, can be selected by users).

If successful, this should make it possible to use CP2K calculations with k-points for a wide range of wavefunction analyses that could be done via post-processing (e.g. using a Python script, or dedicated programs such as Multiwfn), such as PDOS.

Here is a simple example of how it works: [test.zip](https://github.com/user-attachments/files/25482731/test.zip) *(Updated Feb 23)*

### DIFF
--- a/src/kpoint_mo_dump.F
+++ b/src/kpoint_mo_dump.F
@@ -240,15 +240,15 @@ CONTAINS
 
          ! Cell vectors: hmat(:,i) = i-th lattice vector, written as rows
          WRITE (iw, '(A)') "# CELL_VECTORS [Angstrom]"
-         WRITE (iw, '(3ES24.15)') &
+         WRITE (iw, '(3ES18.08)') &
             cell%hmat(1, 1)*angstrom, &
             cell%hmat(2, 1)*angstrom, &
             cell%hmat(3, 1)*angstrom
-         WRITE (iw, '(3ES24.15)') &
+         WRITE (iw, '(3ES18.08)') &
             cell%hmat(1, 2)*angstrom, &
             cell%hmat(2, 2)*angstrom, &
             cell%hmat(3, 2)*angstrom
-         WRITE (iw, '(3ES24.15)') &
+         WRITE (iw, '(3ES18.08)') &
             cell%hmat(1, 3)*angstrom, &
             cell%hmat(2, 3)*angstrom, &
             cell%hmat(3, 3)*angstrom
@@ -278,7 +278,7 @@ CONTAINS
          ! K-point list
          WRITE (iw, '(A)') "# KPOINT_LIST: ikp  kx  ky  kz  weight"
          DO ikp = 1, nkp
-            WRITE (iw, '(I6,4ES20.12)') ikp, xkp(1:3, ikp), wkp(ikp)
+            WRITE (iw, '(I6,4ES18.8)') ikp, xkp(1:3, ikp), wkp(ikp)
          END DO
       END IF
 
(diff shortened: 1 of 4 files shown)


src/kpoint_mo_dump.F
```diff
@@ -0,0 +1,513 @@
...
+      ! Open output file via print key
+      ! ================================================================
+      iw = cp_print_key_unit_nr(logger, print_section, "", &
+                                extension=".mokp", file_status="REPLACE")
+
+      ! ================================================================
+      ! WRITE HEADER
+      ! ================================================================
+      IF (iw > 0) THEN
+         WRITE (iw, '(A)') "# CP2K_KPOINT_MO_DUMP"
```
oschuett: Better add a format version to the header. This makes it easier for parsers  to deal with future changes.

oschuett (commented):
LGTM from my side. Is there anything else you want to discuss before I merge this?

Growl1234:
Thank you for your recognition! Initially, I added this label in hopes of seeking some preliminary opinions and suggestions. After subsequent refinements and testing, the current code seems ideal to me. However, I still have some concerns about whether there might be aspects that appear inappropriate from developer's perspectives, or if this could potentially be a redundant feature? 

(I am aware that CP2K can directly output the real parts of the orbital coefficients, and by combining the basis set information output when `PRINT_LEVEL` is set as `MEDIUM` with sufficient k-point sampling, one should be able to obtain the imaginary parts and the overlap matrix through further post-processing workflows. However, on one hand, relying entirely on this approach might, in my view, reduce the conciseness and readability of the output files; on the other hand, that feature currently has a bug, as mentioned in #4396. This is also why I had the idea of outputting an additional file.)

oschuett:
Ok, then I'll leave this PR open for another day to give people a chance to take a look.

> I am aware that CP2K can directly output the real parts of the orbital coefficients,...

I think, there are just different use-cases that land on different tradeoffs. For example, some folks opted for [parsing the binary wave function](https://github.com/nanotech-empa/cp2k-spm-tools/blob/main/cp2k_spm_tools/cp2k_wfn_file.py). Since your new output routine is well encapsulated, I don't see any harm in adding it.

garrettwx:
Maybe for post-treatments such as PDOS with k-points, the direct mark that distinguishes the different k-points with various projected states can be more advanced.

Growl1234:
> Maybe for post-treatments such as PDOS with k-points, the direct mark that distinguishes the different k-points with various projected states can be more advanced.

Sorry, I may not have fully understood your point. Could you please clarify what you mean by a more advanced marker for distinguishing k-points and projected states? For instance, a simple template or example snippet would be very helpful.

garrettwx:
> > Maybe for post-treatments such as PDOS with k-points, the direct mark that distinguishes the different k-points with various projected states can be more advanced.
> 
> Sorry, I may not have fully understood your point. Could you please clarify what you mean by a more advanced marker for distinguishing k-points and projected states? For instance, a simple template or example snippet would be very helpful.

you can provide the simple text for the output of pdos with k-points, so that the pdos with k-points can be obtained with convolution

Growl1234:
> You can provide the simple text for the output of pdos with k-points, so that the pdos with k-points can be obtained with convolution.

If I understand correctly, this functionality would more naturally belong to what `&PDOS` section does. The `.mokp` output introduced here is intended to provide general support for wavefunction analysis workflows, rather than to directly serve a specific analysis output such as PDOS. I may consider extending the built-in `&PDOS` functionality in CP2K to support k-points in the future.

PR #4864, opened 2026-02-21
Toolchain: Try update of libint

Growl1234:
This PR tries to update libint to the latest version and is mainly intended to test whether this update introduces any issues.

One question: 
Previously, when the old Makefiles were used, some modifications of them were required to address certain bugs. Now that libint has migrated to CMake, those previous adjustments seem no longer applicable. Are similar changes still needed?

### DIFF
--- a/tools/toolchain/install_requirements_fedora.sh
+++ b/tools/toolchain/install_requirements_fedora.sh
@@ -10,9 +10,11 @@ dnf -qy install \
   autoconf \
   autogen \
   automake \
+  boost-devel \
   bzip2 \
   ca-certificates \
   diffutils \
+  eigen3-devel \
   g++ \
   gcc \
   gfortran \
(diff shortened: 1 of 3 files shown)


Growl1234:
Given that Regtest pdbg has no responce, maybe it's better to open up a cleaner PR without too many force-pushes.

oschuett:
> Given that Regtest pdbg has no responce, maybe it's better to open up a cleaner PR without too many force-pushes.

Yeah, the CI is not bulletproof. I believe in part Github is to blame for dropping webhook events, but there are certainly also some race conditions on our side. Nevertheless, the CI works surprisingly well given its simple architecture.

One thing you can try when it gets stuck is to close and reopen a PR. It should trigger a fresh [processing](https://github.com/cp2k/cp2k-ci/blob/master/backend/backend.py#L229).

PR #4865, opened 2026-02-21
Toolchain: libint 2.6.0 -> 2.13.1

Growl1234:
This PR tries to update libint to the latest version and is mainly intended to test whether this update introduces any issues.

One question:
Previously, when the old Makefiles were used, some modifications of them were required to address certain bugs. Now that libint has migrated to CMake, those previous adjustments seem no longer applicable. Are similar changes still needed?

### DIFF
--- a/tools/toolchain/install_requirements_fedora.sh
+++ b/tools/toolchain/install_requirements_fedora.sh
@@ -10,9 +10,11 @@ dnf -qy install \
   autoconf \
   autogen \
   automake \
+  boost-devel \
   bzip2 \
   ca-certificates \
   diffutils \
+  eigen3-devel \
   g++ \
   gcc \
   gfortran \
--- a/tools/toolchain/install_requirements_ubuntu.sh
+++ b/tools/toolchain/install_requirements_ubuntu.sh
@@ -23,6 +23,8 @@ apt-get install -qq --no-install-recommends \
   gfortran \
   git \
   less \
+  libboost-dev \
+  libeigen3-dev \
   libtool \
   libtool-bin \
   make \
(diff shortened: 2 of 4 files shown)


oschuett:
So, what's the plan? Do you want to update the pre-build bundles at https://github.com/cp2k/libint-cp2k or start building everything in the toolchain?

Just in case you decide for the later, I've uploaded the latest tarball to our server: https://www.cp2k.org/static/downloads/libint-2.13.1.tar.gz

Growl1234:
I would prefer to update the pre-built bundles, since building everything from scratch in the toolchain would clearly add a significant amount of overhead.

Still have some questions, though:
1. At the moment, the build workflow in the cp2k/libint-cp2k repository appears to be broken. What is the main reason for this?
2. Assuming the workflow can be restored, the current CMake-based build system of libint has a drawback: it's not available to customize the name of the exported tarball via a command-line option, which means the generated tarball has to be renamed manually. However, we cannot be sure whether the naming convention of the tarball produced by `make export` will remain strictly the same format in future releases. Could this increase the maintenance burden? If the workflow cannot be restored, would we then have to upload the tarball to the Release page manually?

oschuett:
> At the moment, the build workflow in the cp2k/libint-cp2k repository appears to be broken. What is the main reason for this?

The people who built it finished their PhD and then left the project.

> Assuming the workflow can be restored,...

We shouldn't over-engineer this. I'd simply check in a separate Dockerfile that builds the bundles. I'll then run it locally and upload the bundles to cp2k.org. It's a somewhat manual process, but realistically we only have to do this 1-2 per year. And eventually we'll probably migrate to LibintX anyways.

Growl1234:
Hi, are these bundles ready?

PR #4867, opened 2026-02-21
Toolchain: fix tblite-0.5.0.tar.xz install behavior

he-zilong:
Ever since #4653, the toolchain has been installing tblite from
`tblite-0.5.0.tar.xz` which includes readily usable subprojects from other
repositories. However, attempts to perform an offline installation will still
reveal problems about dependencies not being located properly and requiring
duplicated downloads from github. I found a simple solution by creating two
soft links; see tblite/tblite#313 for the full story.

Also, the path to `install/tblite-0.5.0/bin` is added to the `$PATH` variable
in the `setup_tblite` file pointing to executables like `tblite`, `dftd4`, etc.

### DIFF
--- a/tools/toolchain/install_cp2k_toolchain.sh
+++ b/tools/toolchain/install_cp2k_toolchain.sh
@@ -295,13 +295,15 @@ Specific options:
   --with-spglib           Enable the spg library for symmetry groups detection.
                           This package depends on CMake.
                           Default = install
-  --with-dftd4            Enable the DFTD4 package by Grimme for the DFT-D4
-                          dispersion correction method.
+  --with-dftd4            Enable the standalone DFTD4 package by Grimme for the
+                          DFT-D4 dispersion correction method.
                           This package requires CMake.
-                          Default = install
+                          Default = no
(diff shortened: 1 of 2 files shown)


tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -1005,8 +1007,11 @@ fi
 
 # Since tblite includes dftd4, a separate dftd4 is not needed.
 if [ "${with_tblite}" != "__DONTUSE__" ]; then
-  report_warning ${LINENO} "When using tblite, do not use standalone dft-d4."
-  with_dftd4="__DONTUSE__"
+  if [ "${with_dftd4}" != "__DONTUSE__" ]; then
```
Growl1234, line 1010:
```
if [ "${with_tblite}" != "__DONTUSE__" ] && [ "${with_dftd4}" != "__DONTUSE__" ]
```
This maybe better.

he-zilong: I know it is functionally equivalent, but it is too long for a line and I
prefer more short lines (within 80 characters) to fewer long lines as a purely
stylistic choice. 😊

PR #4877, opened 2026-02-25
Fix behavior of `-ef/-df dlaf` and allow running as root with OpenMPI by default

Growl1234:
Original command `SED_PATTERN_LIST+=" -e '/\s*-\s+\"dla-future*@/ ${SUBST}"` is not completely valid because `*` here in `sed` actually males it detect things like "dla-future@", "dla-futur@", "dla-futureee@", etc. in the configuration file, but "dla-future-fortran@" cannot be detected (not the same with common bash script). It is therefore replaced with `.*`.


mkrack:
Thanks, LGTM.

PR #4881, opened 2026-02-26
Update documentation of building from source

Growl1234:


docs/getting-started/build-from-source.md
```diff
@@ -34,12 +40,14 @@ cmake -S . -B build \
...
 
-cmake --build build -j 32
+cmake --build build --target install -j 32
 ```
 
-## Others Build Options
+### Explanation of Above Build Options
 
 - `-GNinja` Generates [Ninja](https://ninja-build.org/) build files instead of GNU Makefiles.
+- `-DCMAKE_INSTALL_PREFIX=./install` Sets installation path to `./install`.
```
he-zilong: Does it mean the location of the CP2K binary executables (`cp2k.psmp` and the like)? If yes then please say so.

Growl1234:
The hint of `cmake -LH` is removed: it directly reads cached options from `CMakeCache.txt`, but we don't allow in-source build so the options listed would not be complete.

docs/getting-started/build-from-source.md
```diff
@@ -1,24 +1,52 @@
...
 
 At a minimum CP2K requirements a modern C and Fortran compiler,
 [DBCSR](https://github.com/cp2k/dbcsr/), BLAS, and LAPACK. For parallel builds it also needs at
-least MPI and ScaLAPACK. To list all available build options run `cmake -LH`. Detailed descriptions
-of most build options can be found in the technologies section:
+least MPI and ScaLAPACK.
+
+GCC is the most tested compiler from our side, while most of other compilers should work. It is not
+recommended to use Intel oneAPI for now; although it could work (up to version 2025.2.1), there are
+still some problems remaining unresolved.
```
he-zilong: By "GCC is the most tested compiler", do you mean `gfortran` most tested for compiling CP2K itself?
Speaking of compiler flavors (GNU with `gcc`/`g++`/`gfortran`, Intel OneAPI with `icc`/`icpc`/`ifx`, AMD with `clang`/`clang++`/`flang`), it would perhaps be a good idea to clarify that the toolchain and CP2K itself should be consistent (so one should not attempt e.g. loading some Intel-compiled OpenMPI or other library modules in toolchain but then compiling CP2K with GNU gfortran; this is just a made-up hypothetical scenario, but you never know how quirky some HPC system could possibly get).

docs/getting-started/build-from-source.md
```diff
@@ -1,24 +1,54 @@
...
-compilation process. The dependencies have to be installed in advance, either manually or through a
-package manager like [Spack](https://spack.readthedocs.io).
+compilation process. The dependencies have to be installed in advance.
 
-## Dependencies
+Currently, CP2K offers two convenient methods for installing CP2K and its dependencies. One is to
+use the classic toolchain scripts to customize and install the required dependencies in a single
+step, thereby preparing the environment for compiling CP2K. A more modern and automated approach is
+via `make_cp2k.sh`, which leverages [Spack](https://spack.readthedocs.io) to install the
+dependencies and subsequently build CP2K.
```
RMeli:
```suggestion
Currently, CP2K offers three convenient methods for installing CP2K and its dependencies. One is to
use the classic toolchain scripts to customize and install the required dependencies in a single
step, thereby preparing the environment for compiling CP2K. A more modern and automated approach
is via `make_cp2k.sh`, which leverages [Spack](https://spack.readthedocs.io) to install the
dependencies and subsequently build CP2K. Finally, CP2K and its dependencies can be installed directly using [Spack](https://spack.readthedocs.io).
```

Maybe worth mentioning that it's also possible to use Spack directly? I lost the plot a bit with all the recent toolchain and make_cp2k.sh changes, but my understanding is that the `make_cp2k.sh` is a convenience script for not using Spack directly, but is essentially doing

```
spack install --only-dependencies cp2k
spack build-env cp2k -- cmake ..
```

I still think that installing dependencies "manually" and then using bare CMake instead of relying on the Spack CP2K package is a missed opportunity for polishing and perfecting the CP2K Spack package, but we will try to keep the upstream CP2K Spack package as up-to-date as possible.

Growl1234: There is already another file named `build-with-spack.md` that explains how to install CP2K directly via Spack. Since this file is explicitly titled “Build from Source”, I’m wondering whether including similar instructions here might introduce some unnecessary duplication.

RMeli: That page was mostly focussed on building the whole of CP2K with Spack, while here I was pointing to using Spack directly to build dependencies and then to setup the build environment (which was not really reflected on the suggested change, but more on my comment).

Admittedly, I forgot I contributed a small section on that in the other page: [developer-workflow](https://manual.cp2k.org/trunk/getting-started/build-with-spack.html#developer-workflow). Maybe it is worth linking it from here then.

docs/getting-started/build-from-source.md
```diff
@@ -1,24 +1,54 @@
...
-compilation process. The dependencies have to be installed in advance, either manually or through a
-package manager like [Spack](https://spack.readthedocs.io).
+compilation process. The dependencies have to be installed in advance.
 
-## Dependencies
+Currently, CP2K offers two convenient methods for building CP2K from source with its dependencies.
+One is to use the classic toolchain scripts to customize and install the required dependencies in a
+single step, thereby preparing the environment for compiling CP2K. A more modern and automated
+approach is via `make_cp2k.sh`, which leverages [Spack](https://spack.readthedocs.io) to install the
+dependencies and subsequently build CP2K.
```
RMeli:
```suggestion
Currently, CP2K offers three convenient methods for building CP2K from source with its dependencies.
One is to use the classic toolchain scripts to customize and install the required dependencies in a
single step, thereby preparing the environment for compiling CP2K. A more modern and automated
approach is via `make_cp2k.sh`, which leverages [Spack](https://spack.readthedocs.io) to install the
dependencies and subsequently build CP2K. Finally, one can use Spack directly to install all the dependencies and setup the build environment, as described in [Building with Spack: Developer Workflow](https://manual.cp2k.org/trunk/getting-started/build-with-spack.html#developer-workflow).
```
Something like that is what I should have suggested in the first place.

docs/getting-started/build-from-source.md
```diff
@@ -1,24 +1,56 @@
...
+## Dependencies and build options
 
 At a minimum CP2K requirements a modern C and Fortran compiler,
 [DBCSR](https://github.com/cp2k/dbcsr/), BLAS, and LAPACK. For parallel builds it also needs at
-least MPI and ScaLAPACK. To list all available build options run `cmake -LH`. Detailed descriptions
-of most build options can be found in the technologies section:
+least MPI and ScaLAPACK.
+
+GCC is the most tested compiler from our side, while most of other compilers should work. It is not
+recommended to use Intel oneAPI for now; although it could work (up to version 2025.2.1), there are
```
oschuett: Instead of singling out Intel, I'd write something like _"your mileage may vary with other compilers"_ 
and then just link to https://www.cp2k.org/dev:compiler_support.

docs/getting-started/build-from-source.md
```diff
@@ -35,10 +67,9 @@ cmake -S . -B build \
     -DCP2K_USE_ACCEL=CUDA -DCP2K_WITH_GPU=A100
 
 cmake --build build -j 32
+cmake --install build
 ```
 
-## Others Build Options
```
oschuett, line 40: I think, this heading is still need.

docs/getting-started/build-from-source.md
```diff
@@ -1,24 +1,56 @@
...
+approach is via `make_cp2k.sh`, which leverages [Spack](https://spack.readthedocs.io) to install the
+dependencies and subsequently build CP2K. Alternatively, one can use Spack directly to install all
+the dependencies and setup the build environment, as described in
+[Building with Spack: Developer Workflow](https://manual.cp2k.org/trunk/getting-started/build-with-spack.html#developer-workflow).
+
+## Dependencies and build options
 
 At a minimum CP2K requirements a modern C and Fortran compiler,
 [DBCSR](https://github.com/cp2k/dbcsr/), BLAS, and LAPACK. For parallel builds it also needs at
-least MPI and ScaLAPACK. To list all available build options run `cmake -LH`. Detailed descriptions
```
oschuett, line 11: I realize that `cmake -LH` is currently broken, but instead of just removing it could we maybe salvage it?

oschuett (commented):
Thanks for improving the docs!

In case you haven't see it yet, the CI actually provides a [preview](https://ci.cp2k.org/artifacts/run-cp2k-manual-5db879bd/manual/getting-started/build-from-source.html):

"Checks" -> "Manual generation" -> "Browse Artifacts" -> "manual/" -> "index.html"

docs/getting-started/build-from-source.md
oschuett: And actually, most other compilers do not work because we're using a lot of modern Fortran features.

docs/getting-started/build-from-source.md
Growl1234: I think a better approach might be to introduce a dedicated option along with a CMake configuration file that simply displays the available CMake options.

On the one hand, this would avoid the readability issues of `cmake -LH`. On the other hand, `cmake -LH` only prints the options after the configuration step has completed, since it reads from the generated cache file. This does not fully align with the intention of conveniently listing available CMake options. With a custom option, we could place the corresponding check near the beginning of `CMakeLists.txt`, allowing the process to exit early and avoid executing the remaining configuration steps.

To be honest, I have previously experimented with this approach [here](https://github.com/Growl1234/cp2k/blob/patch-2/cmake/cp2kListOptions.cmake) and [here](https://github.com/Growl1234/cp2k/blob/patch-2/CMakeLists.txt#L338-L342). However, I encountered some issues related to options defined via `cmake_dependent_option`, which I have not yet resolved.

docs/getting-started/build-from-source.md
oschuett: Sorry for nagging, but I'd prefer if you keep the section heading as `## Others Build Options`.

This section is not related to the example above. It's just a collection of options that didn't fit into any of the technologies pages and are therefore documented here. Most of them are pretty esoteric.

docs/getting-started/build-from-source.md
oschuett: I underestimated the intricacies of `cmake_dependent_option`. Then I'm ok with removing the `cmake -LH`.

Ultimately, every CMake option will be documented properly anyways thanks to [this check](https://github.com/cp2k/cp2k/blob/07b1930b13b0d21eaec480d899a7b3860c2b6edb/tools/precommit/check_file_properties.py#L301).

docs/getting-started/build-from-source.md
Growl1234: Thank you for pointing this out. I will integrate it into the other options listed under “Dependencies and build options” above.

docs/getting-started/build-from-source.md
```diff
@@ -1,24 +1,66 @@
...
+compilation process. The dependencies have to be installed in advance.
 
-## Dependencies
+Currently, CP2K offers three convenient methods for building CP2K from source with its dependencies.
+One is to use the classic toolchain scripts to customize and install the required dependencies in a
+single step, thereby preparing the environment for compiling CP2K. A more modern and automated
+approach is via `make_cp2k.sh`, which leverages [Spack](https://spack.readthedocs.io) to install the
+dependencies and subsequently build CP2K. Alternatively, one can use Spack directly to install all
+the dependencies and setup the build environment, as described in
+[Building with Spack: Developer Workflow](https://manual.cp2k.org/trunk/getting-started/build-with-spack.html#developer-workflow).
```
oschuett: Here we can use a relative link that will also get checked by Sphinx:
```suggestion
[](./build-with-spack#developer-workflow).
```

docs/getting-started/build-from-source.md
```diff
@@ -1,24 +1,66 @@
...
 
 - **[](../technologies/eigensolvers/index)**
 - **[](../technologies/accelerators/index)**
 - **[](../technologies/libraries)**
 
-## Example
+There are some other important general options you may want to know:
+
+- `-G <Generator>` Specifies which type of build files would be generated. Default is
+  `Unix Makefiles`, which generates a GNU Makefile and allow you to build with running `make` in the
```
oschuett:
```suggestion
  `Unix Makefiles`, which generates a GNU Makefile and allows you to build with running `make` in the
```

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -78,86 +79,90 @@ $(basename "$SCRIPT_NAME") [options]
...
+  --no-check-certificate    Bypass verification of server's certificate while
+                            downloading anything from internet via wget command.
+                            In case wget errors about "certificate verification"
+                            or "common name doesn't match requested host name"
+                            occur while downloading tarballs, the recommended
+                            solution is to install the latest wget release;
+                            alternatively, this script can be rerun with this
+                            option in command line.
+                            Security wise this should still be okay as sha256
+                            checksums are checked after every tarball download.
```
oschuett:
```suggestion
                            checksums are validated after every tarball download.
```

docs/getting-started/build-from-source.md
```diff
@@ -1,24 +1,66 @@
...
+  `/path/to/installation/share/cp2k/data`.
+- `-DCP2K_ENABLE_CONSISTENCY_CHECKS` Only used for
+  [testing](https://dashboard.cp2k.org/archive/misc/index.html).
+- `-DCP2K_USE_CRAY_PM_ENERGY` Enables power monitoring on Cray systems.
+- `-DCP2K_USE_CRAY_PM_ACCEL_ENERGY` Enables power monitoring of accelerators on Cray systems.
+- `-DCP2K_USE_DBCSR_CONFIG` Make dbcsr cmake options (`DBCSR_USE_BLA`) available.
+
+Note that CMake is typically run *out-of-tree* in a seperate `build/` directory. We don't allow
+in-source build; if you run CMake with `-B .` (default) in the root directory with `CMakeList.txt`,
+it will give error.
```
oschuett:
```suggestion
in-source builds; if you run CMake in the root directory, it will give error.
```

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -405,6 +379,62 @@ explicitly written in the job submission script.
...
 
+show_help_hpc() {
+  cat << EOF
+
+For HPC users who wish to install toolchain dependencies and CP2K on public
+supercomputer clusters for oneself:
+
+As this is a complicated process, it is strongly advised to contact local
+system administrators or managers for timely, specific assistance. Here are
+some hints and observations may be useful:
```
oschuett:
```suggestion
some hints and observations that may be useful:
```

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -405,6 +379,62 @@ explicitly written in the job submission script.
...
+
+(5) An important common feature of clusters is the distinction of node types:
+    "login node", where users log in and perform tasks with low workload; and
+    "compute node", where resource-intensive computation jobs are carried out.
+    They may be hosted on separate machines, and their hardware specifications
+    (CPU, RAM, disk space, etc.) may be similar or different. Therefore, care
+    must be taken especially for the latter case; for instance, running
+    toolchain scripts on login node with "--target-cpu=native" (which is
+    default too if omitted) and executing CP2K on compute node afterwards may
+    result in unknown behaviors due to discrepancies in CPU architectures and
```
oschuett:
```suggestion
    result in poor performance or illegal instruction errors  due to discrepancies in CPU architectures and
```

INSTALL.md
```diff
@@ -27,6 +27,12 @@ or run in a subshell with:
 ./make_cp2k.sh
 ```
 
+You can run with following command to get information about how to use this script:
```
oschuett:
```suggestion
You can run the following command to get information about how to use this script:
```

oschuett (commented):
I did another read through and found a few more nit picks. Otherwise LGTM.

Good work!

docs/getting-started/build-from-source.md
Growl1234: It doesn't seem to work: 
```
/opt/cp2k/docs/getting-started/build-from-source.md:6: WARNING: 'myst' cross-reference target not found: './build-with-spack#developer-workflow' [myst.xref_missing]
```

图片

oschuett: I'm sorry! The correct syntax is `[](./build-with-spack.md#developer-workflow)` - so the file extension `.md` has to be included. Furthermore, you have to add the line `myst_heading_anchors = 3` to the [conf.py](https://github.com/cp2k/cp2k/blob/master/docs/conf.py). Apparently, we have not linked to a section before.

Growl1234: Thanks a lot! It works now.

PR #4884, opened 2026-02-27
ASPC (along with USE_PREV_WF) for k-points

Growl1234:
This is an attempt to implement ASPC and USE_PREV_WF for k-points. 

It performs a Fourier transform to get the current S(k), re-orthogonalizes the old wavefunctions using Cholesky decomposition, and properly rebuilds the density matrix P(R). For ASPC, the history accumulation and subspace projection are implemented for complex wavefunctions. All fallbacks for linear extrapolation methods in case the history buffer is insufficient can be safely set to `wfi_use_prev_wf_method_nr`.

Here are a set of tests that shows the effect. They include structural optimizations of four different systems. For each system, the initial guess for the subsequent optimization step is generated using `USE_PREV_P`, `USE_PREV_WF`, and `ASPC` method, respectively.

[tests.zip](https://github.com/user-attachments/files/25618038/tests.zip)

Among the four test systems mentioned above, two look good while the other two produce relatively poor initial guesses. It seems that the robustness of this approach still needs further discussion, or there may be other underlying issues.


Growl1234:
I would be very grateful if the developers could take a look at whether the modifications to some of the regtests (also explained in commit message) are correct and consistent with expectations, and to assess whether the approach adopted in this PR is valid.
If there are no major concerns — although fully reaching that point still feels quite difficult for me at this stage — I would greatly appreciate your guidance on whether further testing is still required.

fstein93:
Sorry for not answering to it. I was a few days off. It looks good to me but correct if I am wrong, I do not see any tests for the new feature.

Growl1234:
> I do not see any tests for the new feature.

My several own tests are here: https://github.com/user-attachments/files/25618038/tests.zip (as mentioned in main message)

I also enabled some regtests with k-points and extrapolation (`use_prev_p` and `linear_p` which are already supported but not actually tested in regtests; along with new `use_prev_wf` and `aspc`)

fstein93:
Thank you for the regtests. I have overseen that by default, `ASPC` is used as I rarely bother with extrapolation schemes.

PR #4887, opened 2026-02-27
precommit: Increase timeout for shellcheck to 30s

oschuett:


mkrack:
Increasing the timeout does not help. I will propose an alternative PR.

mkrack:
Please check [this commit](https://github.com/cp2k/cp2k/commit/264fe84f18e1d9320597c51fd405ad671f680389).

oschuett:
I realized that GCP also had a timeout that was still at 10s. I've now increased it to 35s because `cmakeformat` can take a bit longer too. So, that should account for at least some of the timeouts we were seeing.

For the future I'm changing the http code we're sending from 504 to 408 so that we can distinguish the two timeouts.

oschuett:
> Please check [this commit](https://github.com/cp2k/cp2k/commit/264fe84f18e1d9320597c51fd405ad671f680389).

Uff, so there are three timeouts! Maybe we just remove the one on the client side?

mkrack:
That's great. I think this is all mergeable with my changes. I will wait for your PR to be completed.

Issue #1807, opened 2021-12-15
toolchain: verify checksum on already downloaded packages


dev-zero:
The toolchain should (like most package managers do) verify the checksum of the tarballs not only after download but before unpacking.
Commit 7b3f17ae0526bb7351f43a2626058fa726785bec updated the Sirius tarball hash (because upstream decided to repackage with changes). With the current logic a user who has the tarball of SIRIUS v7.2.8 already will never get the new one automatically, nor get an error message that it changed.

oschuett:
Since we're planning to move to Spack, I'm going to close this issue.

PR #4895, opened 2026-03-02
Toolchain: check sha256sum for local packages

he-zilong:
**This issue is obsolete**

Attempt to address #1807 by explicitly checking whether sha256sum matches record in
addition to whether file exists for each package. The `checksum` wrapper
in `tool_kit.sh` is slightly rewritten with (1) the functionality of finding
sha256sum command refactored to `get_checksum_cmd` and (2) the positional
arguments swapped place for convenience.

### DIFF
--- a/tools/toolchain/scripts/tool_kit.sh
+++ b/tools/toolchain/scripts/tool_kit.sh
@@ -615,25 +615,25 @@ read_with() {
   esac
 }
 
-# helper routine to check integrity of downloaded files
-checksum() {
-  local __filename=$1
-  local __sha256=$2
+# check if we have sha256sum command, Mac OS X does not have
+# sha256sum, but has an equivalent with shasum -a 256
+get_checksum_cmd() {
   local __shasum_command='sha256sum'
-  # check if we have sha256sum command, Mac OS X does not have
-  # sha256sum, but has an equivalent with shasum -a 256
   if command -v "$__shasum_command" > /dev/null 2>&1 && ! ${__shasum_command} --version 2>&1 | grep -q 'Darwin'; then
     __shasum_command='sha256sum'
   else
     __shasum_command="shasum -a 256"
   fi
-  if echo "$__sha256  $__filename" | ${__shasum_command} --check; then
-    echo "Checksum of $__filename Ok"
-  else
-    rm -v ${__filename}
-    report_error "Checksum of $__filename could not be verified, abort."
-    return 1
-  fi
+  echo "$__shasum_command"
+}
+
+# helper routine to check integrity of files
+# usage: checksum sha256 filename
+checksum() {
+  local __sha256="$1"
+  local __filename="$2"
+  local __shasum_command=$(get_checksum_cmd)
+  echo "$__sha256  $__filename" | ${__shasum_command} --check
 }
 
 # downloader for the package tars, includes checksum
@@ -651,7 +651,12 @@ download_pkg_from_urlpath() {
     return 1
   fi
   # checksum
-  checksum "${__outfile}" "${__sha256}"
+  if checksum "${__sha256}" "${__outfile}"; then
+    echo "Checksum of $__filename Ok"
+  else
+    report_error "Checksum of $__filename could not be verified, abort."
+    return 1
+  fi
 }
 
 # download from CP2K.org
@@ -663,15 +668,7 @@ download_pkg_from_cp2k_org() {
 # verify the checksums inside the given checksum file
 verify_checksums() {
   local __checksum_file=$1
-  local __shasum_command='sha256sum'
-
-  # check if we have sha256sum command, Mac OS X does not have
-  # sha256sum, but has an equivalent with shasum -a 256
-  if command -v "$__shasum_command" > /dev/null 2>&1 && ! ${__shasum_command} --version 2>&1 | grep -q 'Darwin'; then
-    __shasum_command='sha256sum'
-  else
-    __shasum_command="shasum -a 256"
-  fi
+  local __shasum_command=$(get_checksum_cmd)
 
   ${__shasum_command} --check "${__checksum_file}" > /dev/null 2>&1
 }
@@ -680,15 +677,7 @@ verify_checksums() {
 write_checksums() {
   local __checksum_file=$1
   shift # remove output file from arguments to be able to pass them along properly quoted
-  local __shasum_command='sha256sum'
-
-  # check if we have sha256sum command, Mac OS X does not have
-  # sha256sum, but has an equivalent with shasum -a 256
-  if command -v "$__shasum_command" > /dev/null 2>&1 && ! ${__shasum_command} --version 2>&1 | grep -q 'Darwin'; then
-    __shasum_command='sha256sum'
-  else
-    __shasum_command="shasum -a 256"
-  fi
+  local __shasum_command=$(get_checksum_cmd)
 
   ${__shasum_command} "${VERSION_FILE}" "$@" > "${__checksum_file}"
 }
--- a/tools/toolchain/install_cp2k_toolchain.sh
+++ b/tools/toolchain/install_cp2k_toolchain.sh
@@ -401,12 +401,13 @@ some hints and observations that may be useful:
     recommended to check for available modules (with "module avail", "module
     show" or similar commands) beforehand, and activate desired packages when
     running the toolchain script with "--with-PKG=system" options so as to
-    avoid repeated labour. However, actual compatibility between modules and
+    avoid repeated labour. That said, actual compatibility between modules and
     CP2K to be built may still take rounds of trial-and-error to confirm, and
     resorting to "--with-PKG=install" can sometimes resolve problems if modules
     turn out to be outdated, or compiled inconsistently, or not registering
     complete variables and paths, or not built with GPU support on GPU machine,
-    etc. Please forward these complaints to whoever responsible for the server.
+    etc. Please forward complaints about faulty modules to whoever responsible
+    for the server first before submitting any bug report to program developer.
 
 (4) If no internet connection is available for downloading packages from public
     resources on the server, an offline installation of toolchain and CP2K may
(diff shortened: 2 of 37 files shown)


he-zilong:
Although "Regtest pdbg" passes, there is some error message from `install_cp2k_toolchain.sh --dry-run` early on:
```shell
[91msed: can't read //CMakeLists.txt: No such file or directory
[0m[91msed: can't read //CMakeLists.txt: No such file or directory
[0m[91msed: can't read //CMakeLists.txt: No such file or directory
[0m[91msed: can't read //CMakeLists.txt: No such file or directory
[0m[91msed: can't read //CMakeLists.txt: No such file or directory
[0mSuggested cmake command if toolchain is built with your options: 
  cmake .. -DCMAKE_INSTALL_PREFIX=../install -DCP2K_DATA_DIR=//data -DCP2K_USE_EVERYTHING=ON -DCP2K_USE_DLAF=OFF -DCP2K_USE_PEXSI=OFF
```

This is because the invoked script `./tools/toolchain/scripts/generate_cmake_options.sh` finds `./CMakeLists.txt` by using a relative path that assumes `install_cp2k_toolchain.sh` is in the directory `./tools/toolchain`, but during "Regtest pdbg" the scripts are copied outside before execution.
https://github.com/cp2k/cp2k/blob/223abcc0586bf87373b83aa4a783aedb3e3a30b9/tools/toolchain/scripts/generate_cmake_options.sh#L15
https://github.com/cp2k/cp2k/blob/223abcc0586bf87373b83aa4a783aedb3e3a30b9/tools/docker/Dockerfile.test_pdbg#L14-L29

tools/toolchain/scripts/tool_kit.sh
```diff
@@ -615,25 +615,25 @@ read_with() {
...
     __shasum_command='sha256sum'
   else
     __shasum_command="shasum -a 256"
   fi
-  if echo "$__sha256  $__filename" | ${__shasum_command} --check; then
-    echo "Checksum of $__filename Ok"
-  else
-    rm -v ${__filename}
-    report_error "Checksum of $__filename could not be verified, abort."
-    return 1
```
oschuett, line 635: I have a bad feeling about silently passing over a failed checksum.

So how about printing a warning message and for good measure also deleting the file in question?

he-zilong: I can see the concern here. Currently the `checksum` command has the same message to stdout/stderr and the same exit code as the original command found by `get_checksum_cmd`, and as far as I know `checksum` has been used solely in `if` conditions in `download_pkg_from_urlpath` of `tool_kit.sh` as well as all `install_*.sh` scripts, so post-processing logics for a failed checksum match is accounted for separately.

oschuett: Reading the old code more closely, I realized that it already behaved exactly as I'd liked.
I agree that `get_checksum_cmd()` is a nice dedup, but why do you need the other changes in `checksum()`?

he-zilong: The core logic of `checksum()` is checking sha256sum with command `echo "$__sha256  $__filename" | ${__shasum_command} --check`, which is retained in this PR.
The old code wraps up that core logic with an internal `if` condition statement for use in `download_pkg_from_urlpath()` only. So in this PR, the condition statement and subsequent actions of printing messages is moved to `download_pkg_from_urlpath()` to allow `checksum()` to be used in other `if` condition statements without extraneous message. The behavior of `download_pkg_from_urlpath()` is overall unchanged.
With the modified `checksum()` commandF, former constructs like `if [ -f "${pkg_name}" ]; then` in individual `install_*.sh` scripts are now changed to `if [ -f "${pkg_name}" ] && checksum "${pkg_sha256}" "${pkg_name}"; then`. Package that failes matching sha256sum cannot pass this `if` test and will be re-downloaded.
Maybe adding a comment on top of `checksum()` noting that it is only a wrapper function for sha256 command and is meant to be used in `if` conditions would help clarifying my point.

oschuett: > So in this PR, the condition statement and subsequent actions of printing messages is moved to download_pkg_from_urlpath() to allow checksum() to be used in other if condition statements without extraneous message.

I'm sorry, but I don't see this connection. Why do you have to remove the messaging? Why do you have to remove the `rm -v ${__filename}`?

I think, even for your new use-case, the `checksum()` function can remain to be like this:

```shell
checksum() {
  local __sha256="$1"
  local __filename="$2"
  local __shasum_command=$(get_checksum_cmd)
  if echo "$__sha256  $__filename" | ${__shasum_command} --check; then
    echo "Checksum of $__filename Ok"
  else
    rm -v ${__filename}
    echo "Checksum of $__filename could not be verified."
    return 1
  fi
}
```

he-zilong: As stated above, my intention is to make `checksum()` a wrapper function having the same message to stdout/stderr and the same exit code as the original command found by `get_checksum_cmd()`. This form is as straightforward as the `[ -f "${pkg_name}" ]` in an `if` statement as only the exit status is relevant to the condition logic.

Or, put it another way, consider the following hypothetical model:
```shell
checksum() {
  local __sha256="$1"
  local __filename="$2"
  local __shasum_command=$(get_checksum_cmd)
  echo "$__sha256  $__filename" | ${__shasum_command} --check
}

checksum_verbose_with_removal() {
  local __sha256="$1"
  local __filename="$2"
  if checksum "$__sha256" "$__filename" ; then
    echo "Checksum of $__filename Ok"
  else
    rm -v ${__filename}
    echo "Checksum of $__filename could not be verified."
    return 1
  fi
}
```
Although my `checksum_verbose_with_removal()` is equivalent to your `checksum()`, I would only use my `checksum()` as the condition in an `if` statement, not `checksum_verbose_with_removal()`.

oschuett: Note that `&&` is a short-circuited operator. So in your new use-case:
```
if [ -f foo.tgz ] && checksum "${foo_sha256}" "foo.tgz"; then
```
the `checksum` function is only called when `foo.tgz` exists and only then will it print additional messages.

> ... I would only use my `checksum()` as the condition in an if statement, not `checksum_verbose_with_removal()`.

Yes, but why? From where I stand, this seems to make things less secure for no good reason.

he-zilong: > Note that `&&` is a short-circuited operator. So in your new use-case:
> 
> ```
> if [ -f foo.tgz ] && checksum "${foo_sha256}" "foo.tgz"; then
> ```
> 
> the `checksum` function is only called when `foo.tgz` exists and only then will it print additional messages.

Exactly. If the `[ -f ]` test is skipped, sha256sum check on non-existent file will have exit code 1 with the following stderr message:
```shell
sha256sum: foo.tgz: No such file or directory
foo.tgz: FAILED open or read
sha256sum: WARNING: 1 listed file could not be read
```
This does not look pretty and reasonable when the user already requests packages to be downloaded. [Well, a new integrated verbose function that (1)checks file existence and sha256sum, (2)removes file with correct name but incorrect sha256sum, (3)downloads file from internet, (4)re-check file existence and sha256sum, and so on, may be designed in place of the current method.]

> Yes, but why? From where I stand, this seems to make things less secure for no good reason.

Which things, to be precise? I edited the `download_pkg_from_urlpath()` to keep its functionality, propagating to `download_pkg_from_cp2k_org()`so these should still be as secure as before. I did not find other shell scripts that source `tool_kit.sh` and rely on its `checksum()` command performing extra actions after checking sha256sum, so I decided to do decoupling. Had there be, surely new security measures can be added on demand.

oschuett: > Exactly. If the [ -f ] test is skipped, sha256sum check on non-existent file will have exit code 1 with the following stderr message:
> ```
> sha256sum: foo.tgz: No such file or directory
> foo.tgz: FAILED open or read
> sha256sum: WARNING: 1 listed file could not be read
> ```

Wait, this should not happen! When the `[ -f ]` tests evaluates to false, then the checksum should be skipped. That's why `&&` is called a [short-circuit operator](https://en.wikipedia.org/wiki/Short-circuit_evaluation).

> This does not look pretty and reasonable when the user already requests packages to be downloaded. 

I agree, such a message _would_ be very confusing for the user. However, again I don't think this can ever happen. I believe, you are making a wrong assumption.

he-zilong: An example context to demonstrate the intended logic:
```shell
if [ -f ${openmpi_pkg} ] && checksum "${openmpi_sha256}" "${openmpi_pkg}"; then
  echo "${openmpi_pkg} is found"
else
  download_pkg_from_cp2k_org "${openmpi_sha256}" "${openmpi_pkg}"
fi
echo "Installing from scratch into ${pkg_install_dir}"
[ -d openmpi-${openmpi_ver} ] && rm -rf openmpi-${openmpi_ver}
tar -xjf ${openmpi_pkg}
```
If the OpenMPI tarball does not exist, `[ -f ]` evaluates to false, `checksum()` is skipped as per short-circuit evaluation, condition is not satisfied and a download is requested;
if the tarball exists but sha256sum does not match record, `[ -f ]` evaluates to true, `checksum()` evaluates to false, condition is not satisfied and a download (overwriting the existing tarball) is requested;
finally, if the tarball exists and sha256sum matches record, both evaluates to true and a message saying it is found is printed and no download is necessary.

In my model, the first line cannot be replaced with
```shell
if checksum "${openmpi_sha256}" "${openmpi_pkg}"; then
```
as this will show the symptom I mentioned.

Growl1234: I think, the key question is whether introducing such an `if` condition in the script can robustly prevent checksum-related failures (with exit code 1) that can happen previously. If so, I see no problem for removing checksum report message.

Moreover, in my view, the changes introduced in this PR effectively avoid such a sporadic scenario: during the first run of the toolchain, the download of a package may fail (e.g., due to network issues), leaving behind an incomplete tarball. When the toolchain is executed a second time, the script detects the existing tarball but fails to decompress it and therefore exits immediately instead of re-downloading it, so the user must manually remove the corrupted tarball before the script can proceed normally.

oschuett: > An example context to demonstrate the intended logic: ...

As you've laid out, in all three cases the right thing happens.

> In my model, the first line cannot be replaced with `if checksum "${openmpi_sha256}" ...`

This is a new argument. And I'd reply that you should stick to the pattern you put forward in this PR, which is to only call `checksum` when the file actually exists.

> When the toolchain is executed a second time, the script detects the existing tarball but fails to decompress it and therefore exits immediately instead of re-downloading it, so the user must manually remove the corrupted tarball before the script can proceed normally.

From the way I understand the code, this is not true. Note that `checksum` calls `return 1` not `exit 1`. So the script will simply print a (useful) warning, remove the incomplete tarball, and then move on to downloading it again.

Growl1234: > From the way I understand the code, this is not true. Note that `checksum` calls `return 1` not `exit 1`. So the script will simply print a (useful) warning, remove the incomplete tarball, and then move on to downloading it again.

However, the question is that current toolchain scripts (without this PR applied) don't run checksum to check if the already existing tarballs meet their expectation. Originally the `checksum` is in downloading step, and the second run skips it.

oschuett: Yes, 99% of this PR are fine. All I'm asking for is that whenever a checksum fails that the file in question gets deleted immediately and a warning message is printed.

he-zilong: Thank you all for the reviews. I decided to create a new helper routine in `tool_kit.sh` with these functionalities; however, I also found new aspects that need discussion. Therefore, this PR will be abandoned in favor of a new one.

PR #4896, opened 2026-03-02
Add scripts for building libint bundles

oschuett:
This is a prerequisite for #4865. It will replace what we had at https://github.com/cp2k/libint-cp2k.

@Growl1234, please take a look.

### DIFF
--- a/tools/libint/build_bundles.sh
+++ b/tools/libint/build_bundles.sh
@@ -1,0 +1,41 @@
+#!/bin/bash -e
+
+libint_ver="2.13.1"
+libint_sha256="9651705c79f77418ef0230aafc0cf1b71b17c1c89e413ee0e5ee7818650ce978"
+
+mkdir /opt/libint-bundles
+
+wget -q "https://github.com/evaleev/libint/archive/refs/tags/v${libint_ver}.tar.gz" -O "libint-${libint_ver}.tar.gz"
+if ! echo "${libint_sha256} libint-${libint_ver}.tar.gz" | sha256sum --check; then
+  rm "libint-${libint_ver}.tar.gz"
+  exit 1
+fi
+
+tar -xzf "libint-${libint_ver}.tar.gz"
+cd "libint-${libint_ver}"
+
+for lmax in 4 5 6 7; do
+  echo "Working on lmax=${lmax}"
+  mkdir "build-lmax${lmax}"
+  cd "build-lmax${lmax}"
+  cmake .. \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+    -DLIBINT2_ENABLE_ERI=1 \
+    -DLIBINT2_ENABLE_ERI2=1 \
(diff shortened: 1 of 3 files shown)


tools/libint/build_bundles.sh
```diff
@@ -0,0 +1,34 @@
...
+  cmake .. \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+    -DLIBINT2_ENABLE_ERI=1 \
+    -DLIBINT2_ENABLE_ERI2=1 \
+    -DLIBINT2_ENABLE_ERI3=1 \
+    -DLIBINT2_MAX_AM="${lmax}" \
+    -DLIBINT2_ERI_MAX_AM="${lmax};${lmax-1}" \
+    -DLIBINT2_ERI2_MAX_AM="${lmax+2};${lmax+1}" \
+    -DLIBINT2_ERI3_MAX_AM="${lmax+2};${lmax+1}" \
```
Growl1234: This is not a valid arithmetic expression. I think it should be
```
    -DLIBINT2_ERI_MAX_AM="${lmax};$((lmax-1))" \
    -DLIBINT2_ERI2_MAX_AM="$((lmax+2));$((lmax+1))" \
    -DLIBINT2_ERI3_MAX_AM="$((lmax+2));$((lmax+1))" \
```

Growl1234 (commented):
Just one minor issue, otherwise looks great. Thanks a lot!

oschuett:
I've uploaded the bundles to our server:
- [libint-v2.13.1-cp2k-lmax-4.tgz](https://www.cp2k.org/static/downloads/libint-v2.13.1-cp2k-lmax-4.tgz)	 
- [libint-v2.13.1-cp2k-lmax-5.tgz](https://www.cp2k.org/static/downloads/libint-v2.13.1-cp2k-lmax-5.tgz)	 
- [libint-v2.13.1-cp2k-lmax-6.tgz](https://www.cp2k.org/static/downloads/libint-v2.13.1-cp2k-lmax-6.tgz)	 
- [libint-v2.13.1-cp2k-lmax-7.tgz](https://www.cp2k.org/static/downloads/libint-v2.13.1-cp2k-lmax-7.tgz)

Growl1234:
Oops, I forgot one thing: directly renaming the tarball will cause it to extract into a directory still with the original name, which does not meet the original expectation. Would it be acceptable to simply adjust the directory change command in toolchain script to `cd libint-${libint_ver}-post999` to match the actual condition, or I can submit a new PR to better deal with this?

oschuett:
It seems there is no official way to set the [LIBINT_BUILDID](https://github.com/evaleev/libint/blob/90838808e709ed985ae3982ce9613479cd3c2679/CMakeLists.txt#L42). So, personally I'd say let's just use the bundles as is and adopt the toolchain script.

But I'm also fine with you sending a PR and me building new bundles. However, it's gonna take a couple of hours because building the lmax=7 bundle takes foreeeeever!

PR #4904, opened 2026-03-04
Libint-cp2k: adjust build procedure

Growl1234:
Directly renaming the tarball will cause it to extract into a directory still with the original name, which does not meet the original expectation.

Since there would be a directory `libint-${libint_ver}-post999` whose contents are identical to those in the generated tarball, here we directly rename this directory and re-compress it to better handle this situation. The `.tar.xz` is used because it provides a higher compression ratio, thereby substantially reducing upload and download overhead.

### DIFF
--- a/tools/libint/Dockerfile
+++ b/tools/libint/Dockerfile
@@ -9,7 +9,8 @@ RUN apt-get install -qq --no-install-recommends \
     libgmp-dev \
     libboost-dev \
     make \
-    wget
+    wget \
+    xz-utils
 
 WORKDIR /opt
 COPY ./build_bundles.sh .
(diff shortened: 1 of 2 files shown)


oschuett:
I've removed the old bundles and uploaded the new one:
- [libint-v2.13.1-cp2k-lmax-4.tar.xz](https://www.cp2k.org/static/downloads/libint-v2.13.1-cp2k-lmax-4.tar.xz)
- [libint-v2.13.1-cp2k-lmax-5.tar.xz](https://www.cp2k.org/static/downloads/libint-v2.13.1-cp2k-lmax-5.tar.xz)
- [libint-v2.13.1-cp2k-lmax-6.tar.xz](https://www.cp2k.org/static/downloads/libint-v2.13.1-cp2k-lmax-6.tar.xz)
- [libint-v2.13.1-cp2k-lmax-7.tar.xz](https://www.cp2k.org/static/downloads/libint-v2.13.1-cp2k-lmax-7.tar.xz)

Issue #4843, opened 2026-02-16
RFC: The future of the Toolchain


oschuett:
Our original plan for this year was to replace the Toolchain with Spack and that work is already in [full swing](https://github.com/cp2k/cp2k/commits/master/make_cp2k.sh)

However, we recently also received a lot of great contributions for the Toolchain with the intention of keeping it around:

> For the record, I do NOT like the idea that toolchain is to be completely replaced by Spack + podman, and I want the toolchain system to be actively preserved and maintained alongside the Spack + podman system for building CP2K.

_Originally posted by @he-zilong in https://github.com/cp2k/cp2k/issues/4819#issuecomment-3905875596_

So let's discuss the future of the Toolchain.

alazzaro:
My 2 cents: the toolchain is used by all the projects I know (especially when installing at the user-level). However, it is a custom solution, hard to maintain. So, I will keep it, but likely I will move it to a different repo?

Growl1234:
After reviewing `make_cp2k.sh` and the recent changes (in particular #4840), I tried running the full build process on my own machine. Based on this experience, it now seems that the script can largely serve as a replacement for the traditional toolchain.

I still believe the toolchain retains some practical value as a fallback option, though, especially for users who prefer the traditional dependency-building workflow, or for older HPC systems where adapting to a newer build system may be more challenging. In my view, perhaps we could limit the scope of the toolchain to dependencies that are essential or particularly beneficial for typical CP2K use cases. Components that are far more oriented toward extensibility than practical necessity (for example, DLA-Future or PEXSI) might not need to be added. This would help reduce the maintenance burden of the toolchain — although, in my opinion, that burden is not especially high to begin with.

> So, I will keep it, but likely I will move it to a different repo?

I think this is a good idea, provided that the manual and the README offer sufficiently clear guidance for users.

oschuett:
I really like the idea of reducing the toolchain to the essentials. This will presumably serve the needs of 90%+ of our users.

> So, I will keep it, but likely I will move it to a different repo?

What's the advantage of moving it to another repo? Wouldn't that only slow down its development?

alazzaro:
To be clear: if you keep maintaining the toolchain, then leaving it in the main CP2K repo is the best choice. I propose to put it in a different repo ONLY if you decide to drop it in CP2K (but keep it alive somewhere else).

mkrack:
[...]

PR #4909, opened 2026-03-05
Toolchain: use new function to check filename and sha256sum of packages

### LINKED ISSUE #1807: toolchain: verify checksum on already downloaded packages

he-zilong:
Closes #1807

... superseding #4895. A new function `retrieve_package()` is created in the
`tool_kit.sh` that combines checks of file existence and sha256sum; in case
either fails, a download is requested. This function is applied to all of the
toolchain packages and should make the installation more secure even with
pre-existing tarball files (i.e. put by user offline or partially downloaded).


tools/toolchain/scripts/tool_kit.sh
```diff
@@ -651,7 +653,12 @@ download_pkg_from_urlpath() {
     return 1
   fi
   # checksum
-  checksum "${__outfile}" "${__sha256}"
+  if checksum "${__sha256}" "${__outfile}"; then
+    echo "Checksum of $__filename Ok"
+  else
```
oschuett, line 665:
```suggestion
  else
    rm -vf "${__outfile}"
```

oschuett (commented):
The `retrieve_package()` function is a nice simplification!

he-zilong:
Now, here are two new points worth discussion, though.

Dependencies of gcc
---
It is found that current installation `gcc` could not be completely offline
because some dependencies not in the `gcc` tarball are separately downloaded in
this part of `install_gcc.sh`:
```shell
# Download prerequisites from cp2k.org because gcc.gnu.org returns 403 when queried from GCP.
      sed -i 's|http://gcc.gnu.org/pub/gcc/infrastructure/|https://cp2k.org/static/downloads/|' ./contrib/download_prerequisites
      ./contrib/download_prerequisites > prereq.log 2>&1 || tail_excerpt prereq.log
```
In the corresponding file from `gcc-14.3.0.tar.gz`, some logics are similar to
toolchain (excerpt below):
  ```shell
gmp='gmp-6.2.1.tar.bz2'
mpfr='mpfr-4.1.0.tar.bz2'
mpc='mpc-1.2.1.tar.gz'
isl='isl-0.24.tar.bz2'
gettext='gettext-0.22.tar.gz'

base_url='http://gcc.gnu.org/pub/gcc/infrastructure/'

echo_archives() {
    echo "${gettext}"
    if "${only_gettext}"; then return; fi
    echo "${gmp}"
    echo "${mpfr}"
    echo "${mpc}"
    if [ ${graphite} -gt 0 ]; then echo "${isl}"; fi
}

if type wget > /dev/null ; then
  fetch='wget'
else
  fetch='curl -LO'
fi
chksum_extension='sha512'
directory='.'

for ar in $(echo_archives)
do
    if [ ${force} -gt 0 ]; then rm -f "${directory}/${ar}"; fi
    [ -e "${directory}/${ar}" ]                                               \
        || ( cd "${directory}" && ${fetch} --no-verbose "${base_url}${ar}" )  \
        || die "Cannot download ${ar} from ${base_url}"
done
unset ar
```
I think my `retrieve_package()` can also be applied to these `gcc` dependencies
in the toolchain script `install_gcc.sh` to place them at the correct directory,
after which running the original version of `./contrib/download_prerequisites`
should similarly find them and skip downloading from internet.

A potential improvement of `get_checksum_cmd()`
---
Also from the `./contrib/download_prerequisites` is a little piece
of code of switching sha512sum commands:
```shell
OS=$(uname)

case $chksum_extension in
  sha512)
    case $OS in
      "Darwin"|"FreeBSD"|"DragonFly"|"AIX")
        chksum='shasum -a 512 --check'
      ;;
      "OpenBSD")
        chksum='sha512 -c'
      ;;
      *)
        chksum='sha512sum -c'
      ;;
    esac
  ;;
```
Could this idea (ignoring -c or --check flags and replacing 512 with 256) apply
to our `get_checksum_cmd()` in `tool_kit.sh`? It looks more robust as it has
more explicit cases.

he-zilong:
I am implementing the new `get_checksum_cmd()` now
as it is more relevant to the original purpose of this PR.
Even if my idea about retrieving gcc dependencies is
accepted, it will be left for another PR sometime later.

oschuett:
I went ahead and merged this PR because I think we should stick to `sha256sum`.

Furthermore, I also don't think we should support the offline installation of GCC. Actually, the GCC is one of the packages I'd remove when we're going to slim down the Toolchain as discussed in #4843.

Growl1234:
> Furthermore, I also don't think we should support the offline installation of GCC. 

I agree with that. Extra offline installations of compilers often bring more issues rather than benefits. 

> Actually, the GCC is one of the packages I'd remove when we're going to slim down the Toolchain as discussed in #4843.

I'm sorry but, does it mean that there has been a plan to deal with toolchain?

PR #4910, opened 2026-03-05
Toolchain: Updates of COSMA and HDF5

Growl1234:


oschuett:
I've uploaded the new tarballs to our server:
- https://www.cp2k.org/static/downloads/COSMA-v2.8.1.tar.gz
- https://www.cp2k.org/static/downloads/COSTA-v2.3.0.tar.gz
- https://www.cp2k.org/static/downloads/Tiled-MM-v2.3.2.tar.gz
- https://www.cp2k.org/static/downloads/hdf5-2.1.0.tar.gz

PR #4914, opened 2026-03-05
Optimize resource usage during dependency builds with Spack

### LINKED ISSUE #4860: The step of building libint needs much more memory when buiding CP2K with Spack
... compared to building with toolchain. It may even lead to a crash on my machine due to memory exhaustion.

I also noticed the change in the libint version. Therefore, following the information provided [here](https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/libint/package.py#L177-L188), I generated a CP2K-compatible libint v2.11.2 package locally and replaced the original version in the toolchain. Unfortunately, the conclusion remains unchanged, and the time spent building libint within toolchain is even shorter than before.

Is this due to local spack cache or some other reasons? When I use `-dlc`, the memory usage is not that severe.

Growl1234:
Resolves #4860.

Use `spack install` instead of `make -j"${NUM_PROCS}" --file=spack_makefile`, as `-j"${NUM_PROCS}"` here refers to the number of Spack installation tasks and will results in a cumulation of several concurrent make steps each requesting all available cores.

### DIFF
--- a/make_cp2k.sh
+++ b/make_cp2k.sh
@@ -160,6 +160,7 @@ else
   MAX_PROCS=-1
   NUM_PROCS=${NUM_PROCS:-8}
 fi
+NUM_PACKAGES=1
 NVCC_VERSION=0
 REBUILD_CP2K="no"
 RUN_TEST="no"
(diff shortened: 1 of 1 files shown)


Growl1234:
Closing in favor of #4915.

PR #4921, opened 2026-03-06
ASPC with k-points: Move re-orthogonalization before the extrapolation step

Growl1234:


Growl1234:
That's an incorrect approach and make it perform same as `USE_PREV_WF`.

PR #4926, opened 2026-03-09
Add [Cell] and [Nval] sections to Molden output

Growl1234:


garrettwx:
This update is compatible with Multiwfn for parsing Molden file.

fstein93:
I am currently double-checking how other parsers behave. We need to be a bit cautious with adding additional fields to non-k-point-Molden files.

Growl1234:
> This update is compatible with Multiwfn for parsing Molden file.

Not completely in fact. There is one point: when writing the Molden file from scratch, unit consistency should be taken into account. Therefore, Bohr, as used in the Molden format specification, is adopted here instead of Å. [Nval] can be read correctly, though.

he-zilong:
> Not completely in fact. There is one point: when writing the Molden file from scratch, unit consistency should be taken into account. Therefore, Bohr, as used in the Molden format specification, is adopted here instead of Å.

How about adding an explicit keyword `(Angs|AU)` to the `[Cell]` and `[Atoms]` section headers, just as [molden format](https://www.theochem.ru.nl/molden/molden_format.html) specified, so that there is no ambiguity and data in both section is consistent?
The rest of work is to convince author of multiwfn to account for unit settings when reading `[Cell]` section instead of defaulting to Å.

Growl1234:
> How about adding an explicit keyword `(Angs|AU)` to the `[Cell]` and `[Atoms]` section headers, just as [molden format](https://www.theochem.ru.nl/molden/molden_format.html) specified, so that there is no ambiguity and data in both section is consistent? The rest of work is to convince author of multiwfn to account for unit settings when reading `[Cell]` section instead of defaulting to Å.

In that case, I think we could consider adding a keyword to `MO_MOLDEN` to control whether the cell and atomic positions are written in Angstrom or Bohr units.

he-zilong:
> I am currently double-checking how other parsers behave. We need to be a bit cautious with adding additional fields to non-k-point-Molden files.

What exactly are the "other parsers"? I think it will be beneficial to have such a list (in particular supported/compatible ones) stated clearly in the manual or documents.

fstein93:
What I mean is we should be careful when adding non-standard fields. It may be that unknown fields are ignored by parsers but we should keep in mind that it is not required, i.e. there should be an option to turn printing of non-standard fields on or off.

fstein93:
Another possible parser may be PySCF, and Google (CP2K+Molden) tells me [Google forum](https://groups.google.com/g/cp2k/c/UgYpbL89Hco), [Libra](https://compchem-cybertraining.github.io/Libra_Summer_School_2024/files/Mohammad_Shakiba/July11-nbra.pdf) and a variety of other smaller and larger projects. AFAIU, CP2K currently does not make use of non-standard fields but subtleties already may inhibit the use of CP2K molden files in some software as certain information may be rely on a certain format etc. and I am hesitant to break this behavior.

fstein93:
If you add an option to prevent printing of these fields, I am definitely willing to merge this PR.

Growl1234:
Keywords `WRITE_CELL` and `WRITE_NVAL` are added and default to F.

PR #4927, opened 2026-03-09
Follow-up to the fix in #4364

Growl1234:
After fix in PR #4364 (Fix bug part of issue #4353), the OT calculation has had the ability to output the MO eigenvalues. However, this fix is only valid when [FORCE_EVAL/DFT/SCF/PRINT/RESTART](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/SCF/PRINT/RESTART.html) is enabled; that means, if users use `&RESTART OFF` to prevent generation of binary wavefunction restart files, these eigenvalues would still be output as 0. This PR tries to fix the problem.

### DIFF
--- a/src/qs_scf_post_gpw.F
+++ b/src/qs_scf_post_gpw.F
@@ -2113,6 +2113,29 @@ CONTAINS
       ! k points
       CALL get_qs_env(qs_env, do_kpoints=do_kpoints)
 
+      ! For OT method, ensure MO eigenvalues are correctly loaded
+      IF (.NOT. do_kpoints .AND. PRESENT(scf_env)) THEN
+         IF (scf_env%method == ot_method_nr) THEN
+            CALL get_qs_env(qs_env, mos=mos, matrix_ks=ks_rmpv)
+            IF (ASSOCIATED(mos) .AND. ASSOCIATED(ks_rmpv)) THEN
+               DO ispin = 1, dft_control%nspins
+                  CALL get_mo_set(mo_set=mos(ispin), mo_coeff=mo_coeff, eigenvalues=mo_eigenvalues)
+                  IF (ASSOCIATED(qs_env%mo_derivs)) THEN
+                     mo_coeff_deriv => qs_env%mo_derivs(ispin)%matrix
+                  ELSE
+                     mo_coeff_deriv => NULL()
+                  END IF
+                  CALL calculate_subspace_eigenvalues(orbitals=mo_coeff, &
+                                                      ks_matrix=ks_rmpv(ispin)%matrix, &
+                                                      evals_arg=mo_eigenvalues, &
+                                                      do_rotation=.TRUE., &
+                                                      co_rotate_dbcsr=mo_coeff_deriv)
+                  CALL set_mo_occupation(mo_set=mos(ispin))
+               END DO
+            END IF
+         END IF
+      END IF
+
       ! Write last MO information to output file if requested
       dft_section => section_vals_get_subs_vals(input, "DFT")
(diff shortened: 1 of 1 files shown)


mkrack:
Would it not be easier and more elegant to fix that problem of the missing occupied orbital eigenvalues when using OT in the printout with `&RESTART off` in `qs_scf_output.F` directly?
```
diff --git a/src/qs_scf_output.F b/src/qs_scf_output.F
index 6ff0091ec4..0c300b00f4 100644
--- a/src/qs_scf_output.F
+++ b/src/qs_scf_output.F
@@ -332,6 +332,13 @@ CONTAINS
                                   nmo=nmo, &
                                   flexible_electron_count=flexible_electron_count)
 
+                  ! Ensure that the eigenvalues of the occupied orbitals are available
+                  CALL calculate_subspace_eigenvalues(orbitals=mo_coeff, &
+                                                      ks_matrix=matrix_ks, &
+                                                      evals_arg=mo_eigenvalues, &
+                                                      do_rotation=.TRUE.)
+                  CALL set_mo_occupation(mo_set=mo_set)
+
                   ! Retrieve the index of the last MO for which a printout is requested
                   mo_index_range => section_get_ivals(dft_section, "PRINT%MO%MO_INDEX_RANGE")
                   CPASSERT(ASSOCIATED(mo_index_range))
```

Growl1234:
> Would it not be easier and more elegant to fix that problem of the missing occupied orbital eigenvalues when using OT in the printout with `&RESTART off` in `qs_scf_output.F` directly?
> 
> ```
> diff --git a/src/qs_scf_output.F b/src/qs_scf_output.F
> index 6ff0091ec4..0c300b00f4 100644
> --- a/src/qs_scf_output.F
> +++ b/src/qs_scf_output.F
> @@ -332,6 +332,13 @@ CONTAINS
>                                    nmo=nmo, &
>                                    flexible_electron_count=flexible_electron_count)
>  
> +                  ! Ensure that the eigenvalues of the occupied orbitals are available
> +                  CALL calculate_subspace_eigenvalues(orbitals=mo_coeff, &
> +                                                      ks_matrix=matrix_ks, &
> +                                                      evals_arg=mo_eigenvalues, &
> +                                                      do_rotation=.TRUE.)
> +                  CALL set_mo_occupation(mo_set=mo_set)
> +
>                    ! Retrieve the index of the last MO for which a printout is requested
>                    mo_index_range => section_get_ivals(dft_section, "PRINT%MO%MO_INDEX_RANGE")
>                    CPASSERT(ASSOCIATED(mo_index_range))
> ```

I've tried but it then it seems the printed molden file (with `&MO_MOLDEN`) has non-zero eigenvalues only when `&MO` is on.

mkrack:
I think the problem is that the orbital dependent part of the `MOLDEN` output depends on the `MO` output, currently. It seems this is neither strictly reflected in the input nor in the code structure. The orbital information is stored in the `mo_set` and `MOLDEN` should pick up that information only when it has become available. The `umo_set` information of the unoccupied orbitals could be stored in the `mo_set` structure optionally, i.e. when needed. This would make `OT_NLUMO` from #4924 dispensable and the control is only done by the input keys `ADDED_MOS` and `MO_RANGE`. In principle, `ADDED_MOS` should be adjusted automatically based on the upper `MO_RANGE` index during the initialisation phase. It is better to inform the user that something has been fixed, e.g. the parameter `ADDED_MOS` has been increased, than to say sorry at the end of the run, because the eigenvalues were not computed and thus are not available for printout.

PR #4928, opened 2026-03-10
Toolchain: improve `generate_cmake_options.sh`

he-zilong:
Because `generate_cmake_options.sh` involves files and directories outside of
the `tools/toolchain` folder, it is necessary to validate their existence.
For instance, this problem may arise from the "regtest pdbg" pipeline where
toolchain scripts are copied elsewhere before executing:
https://github.com/cp2k/cp2k/blob/9782edaafd29042022628e50587f5e43aa572e14/tools/docker/Dockerfile.test_pdbg#L14-L29
Due to `--dry-run` option, regtest is not aborted but error message is present:
```shell
 [91msed: can't read //CMakeLists.txt: No such file or directory
 [0m [91msed: can't read //CMakeLists.txt: No such file or directory
 [0m [91msed: can't read //CMakeLists.txt: No such file or directory
 [0m [91msed: can't read //CMakeLists.txt: No such file or directory
 [0m [91msed: can't read //CMakeLists.txt: No such file or directory
 [0mSuggested cmake command if toolchain is built with your options: 
  cmake .. -DCMAKE_INSTALL_PREFIX=../install -DCP2K_DATA_DIR=//data -DCP2K_USE_EVERYTHING=ON -DCP2K_USE_DLAF=OFF -DCP2K_USE_PEXSI=OFF
```
This is now captured by validating existence and printing messages properly.

Meanwhile, analogues to the existing shell variable `CP2K_TOOLCHAIN_OPTIONS`
that is exported in the setup file, a new variable `CP2K_CMAKE_OPTIONS` has
been added to setup file to simplify command; in addition to the old way
```shell
cmake .. {a very long chain of cmake options copy-pasted from CLI message}
```
it is now easier to simply enter the command as
```shell
cmake .. ${CP2K_CMAKE_OPTIONS}
```
Finally, the message at the end of `generate_cmake_options.sh` (which should
conclude execution of `install_cp2k_toolchain.sh`) is redrafted.

### DIFF
--- a/tools/toolchain/scripts/generate_cmake_options.sh
+++ b/tools/toolchain/scripts/generate_cmake_options.sh
@@ -33,15 +33,15 @@ source "${INSTALLDIR}"/toolchain.env
 #
 # First, validate existence of relevant upper-level directory and file
 printf "\n========================== %s =========================\n" \
-"Generating CMake options for building CP2K"
-export CP2K_ROOT=$(real_path "${ROOTDIR}/../..")
+  "Generating CMake options for building CP2K"
+export CP2K_ROOT=$(cd "${ROOTDIR}/../.." && pwd)
 cat << EOF
 Root directory of CP2K is assumed to be ${CP2K_ROOT} (variable \${CP2K_ROOT}).
 Build directory will be ${CP2K_ROOT}/build for executing CMake.
 Install directory will be ${CP2K_ROOT}/install for CP2K binaries, libraries etc.
 EOF
 CMAKE_OPTIONS="-DCMAKE_INSTALL_PREFIX=${CP2K_ROOT}/install"
-if [ -d "${CP2K_ROOT}/data" ] ; then
+if [ -d "${CP2K_ROOT}/data" ]; then
   echo "Data directory ${CP2K_ROOT}/data is found and set as -DCP2K_DATA_DIR."
   CMAKE_OPTIONS="${CMAKE_OPTIONS} -DCP2K_DATA_DIR=${CP2K_ROOT}/data"
 else
@@ -111,8 +111,12 @@ fi
 
 # Export variable for CMake options to setup file
 echo "export CP2K_CMAKE_OPTIONS=\"${CMAKE_OPTIONS}\"" >> "${SETUPFILE}"
-echo "CMake options from parsing files are now collected in the variable
-\${CP2K_CMAKE_OPTIONS} that is exported at the end of ${SETUPFILE}."
+cat << EOF
+
+CMake options from parsing files are now collected in the variable
+\${CP2K_CMAKE_OPTIONS}, which is exported at the end of setup file:
+${SETUPFILE}
+EOF
 
 # -------------------------
 # print out user instructions
--- a/tools/toolchain/scripts/generate_cmake_options.sh
+++ b/tools/toolchain/scripts/generate_cmake_options.sh
@@ -46,7 +46,6 @@ EOF
   CMAKE_OPTIONS="-S ${CP2K_ROOT} -B ${CP2K_ROOT}/build"
   CMAKE_OPTIONS="${CMAKE_OPTIONS} -DCMAKE_INSTALL_PREFIX=${CP2K_ROOT}/install"
   CMAKE_OPTIONS="${CMAKE_OPTIONS} -DCMAKE_INSTALL_LIBDIR=lib"
-  CMAKE_OPTIONS="${CMAKE_OPTIONS} -DCMAKE_SKIP_RPATH=ON"
 else
   report_error ${LINENO} "\${CP2K_ROOT} does not have subdirectory src, so it
 cannot be set as source path for CMake options."
@@ -121,8 +120,17 @@ else
 fi
 
 # Export variable for CMake options to setup file
-printf "# ==================== Setup for CP2K ==================== #" >> "${SETUPFILE}"
-printf 'export CP2K_CMAKE_OPTIONS="%s"\n' "${CMAKE_OPTIONS}" >> "${SETUPFILE}"
+cat << EOF  >> "${SETUPFILE}"
+# ==================== Setup for CP2K ==================== #
+export CP2K_ROOT="${CP2K_ROOT}"
+export CP2K_CMAKE_OPTIONS="${CMAKE_OPTIONS}"
+prepend_path PATH "${CP2K_ROOT}/install/bin"
+prepend_path LD_LIBRARY_PATH "${CP2K_ROOT}/install/lib"
+prepend_path LD_RUN_PATH "${CP2K_ROOT}/install/lib"
+prepend_path LIBRARY_PATH "${CP2K_ROOT}/install/lib"
+prepend_path PKG_CONFIG_PATH "${CP2K_ROOT}/install/lib/pkgconfig"
+prepend_path CMAKE_PREFIX_PATH "${CP2K_ROOT}"
+EOF
 cat << EOF
 Suggested cmake command if toolchain is built with your options:
 
@@ -155,7 +163,9 @@ Toolchain is now ready for building CP2K! Instructions for next steps:
     suggested above:
       cd ${CP2K_ROOT}
       cmake ${CMAKE_OPTIONS}
-    Other commands from ${CP2K_ROOT}/CMakeLists.txt can also be added.
+    Other commands from ${CP2K_ROOT}/CMakeLists.txt can also be added. For more
+    information about available build options, see documentation:
+    https://manual.cp2k.org/trunk/getting-started/build-from-source.html.
     Alternative to copy-paste long lines in terminal is to use a variable from
     setup file for CMake options, which is not to be quoted so that whitespace
     delimiters allow it to expand to command options in shell:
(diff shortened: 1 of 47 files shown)


Growl1234:
If I were to address this issue in the regtests, I would [modify the Docker test files](https://github.com/cp2k/cp2k/blob/master/tools/docker/generate_dockerfiles.py) to remove the dedicated copy step and adjust the relative behaviors for the toolchain directory, rather than editing `generate_cmake_options.sh`. It may also be worth looking at the Docker test files related to `make_cp2k.sh`, since `make_cp2k.sh` also sets the default value of `CP2K_ROOT` to `${PWD}`.

Another point to note is that the regtests do not rely on the CMake options generated by the toolchain to build CP2K; instead, they use [cmake_cp2k.sh](https://github.com/cp2k/cp2k/blob/master/cmake/cmake_cp2k.sh).

he-zilong:
I see your point. Now, a new flag `--list-cmake-options` for toolchain is
introduced to control whether the `generate_cmake_options.sh` at the end is
called or not. By default, and for ordinary users working with CLI, it is set
to "yes"; for the automatic workflows in docker tests, it is set to "no".

tools/toolchain/scripts/generate_cmake_options.sh
```diff
@@ -145,13 +148,15 @@ Toolchain is now ready for building CP2K! Instructions for next steps:
 (3) Recommended - go to root directory of CP2K and build executable:
       cd ${CP2K_ROOT}
       mkdir build && cd build
```
Growl1234: If you use `-S ${CP2K_ROOT} -B ${CP2K_ROOT}/build` above, the command `mkdir build && cd build` here should be removed.

tools/toolchain/scripts/generate_cmake_options.sh
```diff
@@ -40,7 +40,10 @@ Root directory of CP2K is assumed to be ${CP2K_ROOT} (variable \${CP2K_ROOT}).
 Build directory will be ${CP2K_ROOT}/build for executing CMake.
 Install directory will be ${CP2K_ROOT}/install for CP2K binaries, libraries etc.
 EOF
-CMAKE_OPTIONS="-DCMAKE_INSTALL_PREFIX=${CP2K_ROOT}/install"
+CMAKE_OPTIONS="-S ${CP2K_ROOT} -B ${CP2K_ROOT}/build"
+CMAKE_OPTIONS="${CMAKE_OPTIONS} -DCMAKE_INSTALL_PREFIX=${CP2K_ROOT}/install"
+CMAKE_OPTIONS="${CMAKE_OPTIONS} -DCMAKE_INSTALL_LIBDIR=lib"
+CMAKE_OPTIONS="${CMAKE_OPTIONS} -DCMAKE_SKIP_RPATH=ON"
```
Growl1234: I don't know what `CMAKE_SKIP_RPATH` means and how it benefits for building CP2K here. And why would you like to force `CMAKE_INSTALL_LIBDIR` to `lib`?

tools/toolchain/scripts/generate_cmake_options.sh
```diff
@@ -145,13 +148,15 @@ Toolchain is now ready for building CP2K! Instructions for next steps:
 (3) Recommended - go to root directory of CP2K and build executable:
       cd ${CP2K_ROOT}
       mkdir build && cd build
-      cmake .. ${CMAKE_OPTIONS}
+      cmake ${CMAKE_OPTIONS}
       make install -j $(get_nprocs)
```
Growl1234: Same reason as above (Makefile is generated in `build` directory), `make install -j` should be changed to `cmake --build ${CP2K_ROOT}/build --target install -j`.

tools/toolchain/scripts/generate_cmake_options.sh
```diff
@@ -145,13 +148,15 @@ Toolchain is now ready for building CP2K! Instructions for next steps:
...
+    The CMake command shows the options from generate_cmake_options.sh. For more
+    CMake options, see ${CP2K_ROOT}/CMakeLists.txt.
     Alternatively the CMake command may take the same options from a variable as
     exported in the setup file, no need to copy-paste long lines in terminal:
-      cmake .. "\${CP2K_CMAKE_OPTIONS}"
+      cmake "\${CP2K_CMAKE_OPTIONS}"
     It may be helpful to also save a copy of command line messages to log files:
-      cmake .. "\${CP2K_CMAKE_OPTIONS}" 2>&1 | tee make.log
+      cmake "\${CP2K_CMAKE_OPTIONS}" 2>&1 | tee make.log
       make install -j $(get_nprocs) 2>&1 | tee install.log
```
Growl1234: Same.

tools/toolchain/scripts/generate_cmake_options.sh
Growl1234: `cmake --build ${CP2K_ROOT}/build --target clean`

tools/toolchain/scripts/generate_cmake_options.sh
he-zilong: So the `cmake` command can be executed at working directory `${CP2K_ROOT}` if the `-B` option is set? That's good.

tools/toolchain/scripts/generate_cmake_options.sh
he-zilong: Both `-DCMAKE_SKIP_RPATH=ON` and `-DCMAKE_INSTALL_LIBDIR=lib` are present in `make_cp2k.sh`, so I copied them here.
https://github.com/cp2k/cp2k/blob/f7bb64f48e842834c06c0e26df61b158a2627026/make_cp2k.sh#L1205-L1270

Growl1234: The `CMAKE_INSTALL_LIBDIR="lib"` set here is due to how `make_cp2k.sh` add `LD_LIBRARY_PATH`:
https://github.com/cp2k/cp2k/blob/f7bb64f48e842834c06c0e26df61b158a2627026/make_cp2k.sh#L1312
But here we make users set these environment variables by themselves (can be through environment modules for a workstation/HPC server or probably `~/.bashrc` for PC, etc.), so this makes no sense.

The `CMAKE_SKIP_RPATH` will entirely disable the generation of run-time search path. Note after CMake there's a hint for users to do tests through binaries that is stored in build directories, and in the past these binaries don't need to find `libcp2k.so` (until install step which removes the rpath by default). I'm afraid that this setting will make that hint nearly invalid (you must set `export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/path/to/cp2k/build/src`!), unless you modify that message.

he-zilong:
Running through the whole CP2K build workflow reveals two issues:
- Some path settings are duplicated for OpenBLAS in the setup file generated by `--with-openblas=install`; this is now fixed.
- The final message about regtests refers to the location of binary executable in `build/bin` instead of `install/bin`; this is not fixed yet.
```shell
After building CP2K the regtests can be run with the following command:
   ./tests/do_regtest.py /cp2k/build/bin psmp
```
Then, if it is set to `install/bin` where `cp2k.psmp` is located, a missing shared object file hinders regtest.
```shell
/cp2k/install/bin/cp2k.psmp: error while loading shared libraries: libcp2k.so.2026.1: cannot open shared object file: No such file or directory
```

Growl1234:
> Then, if it is set to `install/bin` where `cp2k.psmp` is located, a missing shared object file hinders regtest.

That's because the installed CP2K binary needs to explicitly link against `libcp2k.so` (through adding the path to library to `LD_LIBRARY_PATH`), whereas a build that has not been installed does not. This is exactly the default behavior of CMake regarding runtime paths (build-tree binaries retain runtime paths, while they are stripped upon installation). This is also why I did not recommend adding `CMAKE_SKIP_RPATH=ON` above, since that would remove the runtime paths from the build-tree binaries as well.

In `make_cp2k.sh`, a separate test step is provided at the end after everything has been installed, so adding this is not a problem, and it may also help save cache usage during the build (I guess).

If you still want that option, an alternative would be to add a note in the output of `generate_cp2k.sh` suggesting a new and proper testing command.

mkrack:
> In `make_cp2k.sh`, a separate test step is provided at the end after everything has been installed, so adding this is not a problem, and it may also help save cache usage during the build (I guess).

Although the use of `CMAKE_SKIP_RPATH=ON` is convenient, it can have side effects by pulling in unwanted (even conflicting) libraries from the host system. That is why, the libraries really needed by the generated CP2K binary are collected explicitly. By default, `make_cp2k.sh` tries to keep the CP2K build as much as possible local to the `CP2K_ROOT` folder to isolate the CP2K build from the host system.

he-zilong:
There are two variables `${cp2k_BINARY_DIR}` and `${PROJECT_BINARY_DIR}` in
`CMakeLists.txt` that looks very similar. What is their difference and how are
they assigned? Is the former intended to be `build` instead of `install` here?
https://github.com/cp2k/cp2k/blob/c6abaaea11d25e95494e6c5fa9298aec77cd1223/CMakeLists.txt#L1241-L1244

Growl1234:
I think the `PROJECT_BINARY_DIR` is equal to `cp2k_BINARY_DIR` according to some other CMake codes relevant to these variables.

Maybe here's the point:
https://github.com/cp2k/cp2k/blob/c6abaaea11d25e95494e6c5fa9298aec77cd1223/CMakeLists.txt#L25-L30

he-zilong:
I have corrected the variable name in the regtest message and replaced the two
other references of `${cp2k_BINARY_DIR}` to `${PROJECT_BINARY_DIR}` to reduce
confusion. I think this PR is ready to be merged.

he-zilong:
One day passed without seeing this PR merged...
I went ahead to remove more duplicated paths
exported by toolchain. These `${pkg_ROOT}` variables
were treated twice inside and outside of the
`if [ "$with_pkg" != "__SYSTEM__" ]` condition.

PR #4930, opened 2026-03-10
Include pexsi into pdbg tester

mkrack:


tools/spack/lsan.supp
```diff
@@ -5,4 +5,5 @@ leak:MPIR_Csel_prune
 leak:MPIDIU_avt_init
 leak:MPIR_Group_create
 leak:MPIDIU_alloc_lut
-leak:MPL_malloc
+leak:malloc
+leak:operator new(unsigned long)
```
oschuett: I think these patterns are too generic as they will exclude all allocations and thereby render the leak sanitizer ~~useful~~ useless.

PR #4906, opened 2026-03-04
Test openpmd with pdbg

mkrack:
Includes some minor fixes to get a pdbg version with openpmd compiled.

PR #4931, opened 2026-03-10
openPMD output routines: Fix indexing

franzpoeschel:
Since the first dimension of the output array is distributed among MPI processes, I had to adjust for uneven parallel decompositions, leading to slightly offset definitions for the array bounds along the first dimension.
While doing that, I apparently forgot adjusting the bounds back to the lower corner of the array (not at (0,0,0)). This fixes that.

TODO:

- [x] While testing this, I get a hangup in HDF5. Unsure yet if this is related to the issue at hand, need to look further.
    UPDATE: This issue seems to have come with HDF5 2.0, HDF5 1.14.3 works fine. Need to check what causes it.
    UPDATE \# 2: The error is unrelated, will be fixed separately
- [x] Use the chance to clarify some status outputs

### DIFF
--- a/src/pw/realspace_grid_openpmd.F
+++ b/src/pw/realspace_grid_openpmd.F
@@ -394,8 +394,9 @@ CONTAINS
 
       DO iat = 1, 3
          global_extent(iat) = (pw%pw_grid%npts(iat) + my_stride(iat) - 1)/my_stride(iat)
-         ! '+ 1' because upper end is inclusive, '- 1' for upper gaussian bracket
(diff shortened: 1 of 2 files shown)


franzpoeschel:
The HDF5 hangup is unrelated, check #4942 for a workaround and https://github.com/openPMD/openPMD-api/pull/1862 for a WIP attempt at really fixing this.

mkrack:
@franzpoeschel Thanks, that fixes #4906.

PR #4942, opened 2026-03-11
openPMD output: Fix hangup of parallel HDF5 in versions >= 2.0

franzpoeschel:
HDF5 seems to have become a lot stricter about collective parallel metadata handling as of version 2.0, leading to hanging file close operations. Ultimately, I need to fix this inside the openPMD-api (https://github.com/openPMD/openPMD-api/pull/1862), but this PR includes a workaround that at least makes parallel HDF5 output from CP2K to openPMD work again.

Also adds a call to explicitly open Iterations, good practise in parallel setups.

What the workaround does: `store_chunk` operations in the openPMD-api are intended to be non-collective. However, CP2K uses one kind of `store_chunk` operation that allocates memory in the backend, requiring some minimal structural setup of groups / datasets. 
Earlier, these setup operations had to be defined consistently, but not in the same order across MPI ranks. Now, the ordering has become decisive. 
As a workaround, this PR issues zero-sized `store_chunk` operations on ranks that do not contribute data. This runs the structural setup, while ignoring any actual data write.

### DIFF
--- a/src/openpmd_api.F
+++ b/src/openpmd_api.F
@@ -105,6 +105,7 @@ MODULE openpmd_api
       PROCEDURE, PUBLIC :: get_mesh => openpmd_iteration_get_mesh
       PROCEDURE, PUBLIC :: get_particle_species => openpmd_iteration_get_particle_species
       PROCEDURE, PUBLIC :: close => openpmd_iteration_close
+      PROCEDURE, PUBLIC :: open => openpmd_iteration_open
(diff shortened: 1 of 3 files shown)


mkrack:
`Regtest pdbg` does not include `openpmd`. Use the tester `Spack pdbg` with `openpmd` enabled instead.
I will merge this PR and enable again `openpmd` for the `Spack pdbg`.

Issue #4170, opened 2025-04-25
Problem about SYMMETRY RHOMBOHEDRAL


hyl0122:
[Mnse.txt](https://github.com/user-attachments/files/19911321/Mnse.txt)
I initially intended to perform structural optimization on MnSe while maintaining symmetry and the space group [Fm3̅m]. The structure is a = b = c and α = β = γ ≠ 90°. 
However, I found that when I add SYMMETRY RHOMBOHEDRAL in &CELL, CP2K forcibly triangularizes the basis matrix, resulting in the program recognizing the space group as P-1. This forced triangularization changes the lattice parameters, but the atomic coordinates remain unchanged. Has an error occurred? Why does the program forcibly triangularize the basis?

![Image](https://github.com/user-attachments/assets/eeecabb0-2c83-450b-8761-e9873e1f7916)
![Image](https://github.com/user-attachments/assets/7fcbad3e-2b5e-4b8a-becd-ec0f103055ef)

PR #4950, opened 2026-03-13
Clarify invalid/ambiguous cell definitions

fstein93:
@he-zilong Do you have better formulations in your mind?

### DIFF
--- a/src/cell_methods.F
+++ b/src/cell_methods.F
@@ -279,9 +279,8 @@ CONTAINS
 
       INTEGER                                            :: my_per
       INTEGER, DIMENSION(:), POINTER                     :: multiple_unit_cell
-      LOGICAL                                            :: cell_read_a, cell_read_abc, cell_read_b, &
-                                                            cell_read_c, cell_read_file, check, &
-                                                            my_check
+      LOGICAL :: cell_read_a, cell_read_abc, cell_read_alpha_beta_gamma, cell_read_b, cell_read_c, &
+         cell_read_file, check_read_abc, my_check
       REAL(KIND=dp), DIMENSION(:), POINTER               :: cell_angles, cell_par
       TYPE(section_vals_type), POINTER                   :: cell_ref_section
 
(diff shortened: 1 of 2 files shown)


he-zilong:
Thank you for the heads-up. It seems to me that the flow can be simpler:
```fortran
CALL section_vals_val_get(cell_section, "A", explicit=cell_read_a)
CALL section_vals_val_get(cell_section, "B", explicit=cell_read_b)
CALL section_vals_val_get(cell_section, "C", explicit=cell_read_c)
CALL section_vals_val_get(cell_section, "ABC", explicit=cell_read_abc)
CALL section_vals_val_get(cell_section, "ALPHA_BETA_GAMMA", explicit=cell_read_alpha_beta_gamma)

IF (cell_read_abc) THEN ! Presence of ABC overrides A, B, C
   IF (cell_read_a .OR. cell_read_b .OR. cell_read_c) &
      CALL cp_warn(__LOCATION__, &
                   "Cell Information provided through vectors A, B or C in conjunction with ABC."// &
                   " The definition of the ABC keyword will override the one provided by A, B and C.")
   CALL section_vals_val_get(cell_section, "ABC", r_vals=cell_par)
   IF (cell_read_alpha_beta_gamma) THEN ! Coexistence of ABC and ALPHA_BETA_GAMMA
      CALL section_vals_val_get(cell_section, "ALPHA_BETA_GAMMA", r_vals=cell_angles)
      CALL set_cell_param(cell, cell_par, cell_angles, do_init_cell=.FALSE.)
   ELSE
      ! ABC without ALPHA_BETA_GAMMA
   END IF
ELSE ! With no ABC, try to read A, B, C
   check_read_abc = ((cell_read_a .EQV. cell_read_b) .AND. (cell_read_b .EQV. cell_read_c))
   IF (check_read_abc) THEN
      CALL section_vals_val_get(cell_section, "A", r_vals=cell_par)
      cell%hmat(:, 1) = cell_par(:)
      CALL section_vals_val_get(cell_section, "B", r_vals=cell_par)
      cell%hmat(:, 2) = cell_par(:)
      CALL section_vals_val_get(cell_section, "C", r_vals=cell_par)
      cell%hmat(:, 3) = cell_par(:)
      IF (cell_read_alpha_beta_gamma) &
         CALL cp_warn(__LOCATION__, &
                      "The keyword ALPHA_BETA_GAMMA is ignored because it was used without the "// &
                      "keyword ABC.")
   ELSE
      CALL cp_abort(__LOCATION__, &
                   "Cell Information provided through vectors A, B and C. Not all three "// &
                   "vectors were provided! Cell setup may be incomplete!")
   END IF
END IF
```
Just to make sure if I get the logic right,
- why was it intended that `ABC` overrides `A`, `B`, `C`?
- if there is `ABC` but not `ALPHA_BETA_GAMMA`, is `CALL section_vals_val_get(cell_section, "ALPHA_BETA_GAMMA", r_vals=cell_angles)` safely returning three 90 deg values as the default value stated in manual?

fstein93:
>     * why was it intended that `ABC` overrides `A`, `B`, `C`?
I don't know for sure but I guess a user rather knows the lattice parameters correctly than the cell vectors.
>     * if there is `ABC` but not `ALPHA_BETA_GAMMA`, is `CALL section_vals_val_get(cell_section, "ALPHA_BETA_GAMMA", r_vals=cell_angles)` safely returning three 90 deg values as the default value stated in manual?
Yes. 90° is the default value and as long as the user does not set the keyword explicitly, the default is not overwritten.

he-zilong:
> Yes. 90° is the default value and as long as the user does not set the keyword explicitly, the default is not overwritten.

I see. From the last version of the code, it also looks fine to skip checking `cell_read_alpha_beta_gamma` in the first branch and proceed to get `cell_angles` directly. More lines shaved off!
```
IF (cell_read_abc) THEN ! Presence of ABC overrides A, B, C
   IF (cell_read_a .OR. cell_read_b .OR. cell_read_c) &
      CALL cp_warn(__LOCATION__, &
                   "Cell Information provided through vectors A, B or C in conjunction with ABC."// &
                   " The definition of the ABC keyword will override the one provided by A, B and C.")
   CALL section_vals_val_get(cell_section, "ABC", r_vals=cell_par)
   CALL section_vals_val_get(cell_section, "ALPHA_BETA_GAMMA", r_vals=cell_angles)
   CALL set_cell_param(cell, cell_par, cell_angles, do_init_cell=.FALSE.)
ELSE ...
```

By the way, I suggest the manual/documentation clarifying a couple of details:
- whether the file referred to by `&FORCE_EVAL/&SUBSYS/&TOPOLOGY/COORD_FILE_NAME` and `COORD_FILE_FORMAT` will be parsed for cell definitions in addition to atomic coordinates;
- in case either `A`, `B`, `C` is specified or a file is parsed for cell definitions, BUT the vectors do NOT conform to CP2K convention (where A lies along the X-axis, B is in the XY plane, making the cell matrix a lower triangular matrix) like in issue #4170, what is the behavior.

fstein93:
>     * whether the file referred to by `&FORCE_EVAL/&SUBSYS/&TOPOLOGY/COORD_FILE_NAME` and `COORD_FILE_FORMAT` will be parsed for cell definitions in addition to atomic coordinates;
There is already the keyword pair `CELL_FILE_FORMAT`/`CELL_FILE_NAME` in the `&CELL` section for it. I will make some adjustments to the manual.
>     * in case either `A`, `B`, `C` is specified or a file is parsed for cell definitions, BUT the vectors do NOT conform to CP2K convention (where A lies along the X-axis, B is in the XY plane, making the cell matrix a lower triangular matrix) like in issue [Problem about SYMMETRY RHOMBOHEDRAL #4170](https://github.com/cp2k/cp2k/issues/4170), what is the behavior.
This may vastly depend on the kind of calculations you are interested in. From my experience, it does not affect simple DFT calculations. But I cannot tell you how it is with more advanced calculations. Usually, the authors of the respective module need to ensure invariance wrt cyclic permutations of the cell vectors (acyclic permutations affect chiral properties).

he-zilong:
> This may vastly depends on the kind of calculations you are interested in. From my experience, it does not affect simple DFT calculations. But I cannot tell you how it is with more advanced calculations. Usually, the authors of the respective module need to ensure invariance wrt cyclic permutations of the cell vectors (acyclic permutations affect chiral properties).

Well... I mean, whatever the computation target is, _silently_ modifying cell vectors and/or atomic coordinates from user input is not good. In the scope of this PR, I would simply like to see (1) an explanation in user manual and (2) a warning message whenever it happens (like the triangularization mentioned in #4170) during execution.

fstein93:
The cell vectors and atomic positions are not silently modified. They are taken from the respective keywords or coordinate files. This PR adds warnings regarding where CP2K gets its cell definition from. #4170 is a different issue and needs a different solution beyond this PR.

he-zilong:
> The cell vectors and atomic positions are not silently modified.

Thank you for the clarification. Then, everything else (other than regtests with all-zero cell vectors, that is) looks good to me.

PR #4954, opened 2026-03-13
Toolchain: Update CMake version to 4.2.3

Growl1234:


oschuett:
I've uploaded the new bundles to our server. I'm not sure why weren't using the CMake tarballs on all architectures. I now included them, in case you want to streamline things:

- https://www.cp2k.org/static/downloads/cmake-4.2.3-linux-aarch64.sh
- https://www.cp2k.org/static/downloads/cmake-4.2.3-linux-aarch64.tar.gz
- https://www.cp2k.org/static/downloads/cmake-4.2.3-linux-x86_64.sh
- https://www.cp2k.org/static/downloads/cmake-4.2.3-linux-x86_64.tar.gz
- https://www.cp2k.org/static/downloads/cmake-4.2.3-macos-universal.tar.gz
- https://www.cp2k.org/static/downloads/trexio-2.6.1.tar.gz

Growl1234:
> I'm not sure why weren't using the CMake tarballs on all architectures. I now included them, in case you want to streamline things:

It seems that the behaviors of the two files are the same. Now I only need to use the `tar -xvf` command without configuring script options, which is indeed a bit more convenient. However, the packaging structures for the Linux and macOS versions are different, so they still need to be handled separately.

Another thing I am curious about is why the build-from-scratch approach was later abandoned (perhaps to significantly save time?).

oschuett:
> Another thing I am curious about is why the build-from-scratch approach was later abandoned (perhaps to significantly save time?).

IIRC, building CMake from scratch takes indeed a long time.

Growl1234:
Hi, is there any problem about this PR needing to be resolved?

tools/toolchain/scripts/stage4/install_cosma.sh
```diff
@@ -142,7 +142,15 @@ case "$with_cosma" in
       fi
       cd ..
 
-      cd COSMA-${cosma_ver} && mkdir build-cpu && cd build-cpu
+      cd COSMA-${cosma_ver}
+      # Fix two bugs in src/cosma/CMakeLists.txt:
```
oschuett, line 146: Are these bugs already reported or fixed upstream https://github.com/eth-cscs/COSMA?

Growl1234: I have submitted a PR https://github.com/eth-cscs/COSMA/pull/164.

Growl1234: It's now merged. We can remove this small patch when the next COSMA release is out.

Issue #1359, opened 2021-02-12
Adding Gaussian / Methfessel-Paxton / Marzari-Vanderbilt smearing to the toolbox


ltalirz:
Particularly in high-throughput studies, where one does not know a priori whether a given structure is metallic or insulating, it can be useful to use diagonalization with smearing as a fallback if orbital transformation does not converge.

Fermi-Dirac smearing is known to introduce rather strong deviations in e.g. the atomic forces compared to other smearing schemes (see e.g. Figure 7 of [this paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6424870/)).

To your knowledge, has there been any work in the direction of extending the [smearing schemes available in cp2k](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/SCF/SMEAR.html#list_METHOD) (forks, ...)?
If not, would it be possible to get some pointers on where/how to best get started; any details one should pay attention to?
I guess the place to start would be [here](https://github.com/cp2k/cp2k/blob/5a03381abe24376cf018f40887736cdc95d3b0a6/src/qs_mo_occupation.F#L557).

Apropos: @flavianojs from our group just [added to Quantum Espresso](https://gitlab.com/QEF/q-e/-/merge_requests/1294) an implementation for reliably computing Fermi energies for Methfessel-Paxton and Marzari-Vanderbilt smearing (which produce non-monotonous occupations) and might be able to help a bit with this as well.

mkrack:
Recently, I looked into that issue for the same reason, since Fermi-Dirac smearing is not optimal (e.g. long distribution tails) and it would be nice indeed to have MP smearing as an alternative (like in VASP with ISMEAR = 0, 1, or 2). Your starting point is fine, but with k points you should also consider [this spot](https://github.com/cp2k/cp2k/blob/5a03381abe24376cf018f40887736cdc95d3b0a6/src/kpoint_methods.F#L832). The module [fermi_utils](https://github.com/cp2k/cp2k/blob/master/src/fermi_utils.F) has either to be generalized to a smearing module or a new methfessel_paxton module has to be added.

Growl1234:
Hi @ltalirz. These methods should be added through #4958 in [this file](https://github.com/cp2k/cp2k/blob/master/src/smearing_utils.F). You can have a check on it.

ltalirz:
That's great news, congrats! I don't have the time now to test it, but I will close the issue. Thanks!

PR #4958, opened 2026-03-14
Implementation of alternative smearing methods

Growl1234:
This PR shows an attempt to implement Gaussian smearing method. 

The broading of it is much narrower than Fermi-Dirac distribution, and in most cases it should give a very resonable result. However, although Gaussian smearing is useful in practice, it does not have a clear physical meaning and therefore cannot be used in situations where electronic temperature or thermal excitations are of interest.

Most of the logic is adapted from how the Fermi–Dirac method works; the main differences lie in the mathematical formulas. In addition, an analytical form of df_i/dε_j is attempted here.

P.S. There are other two things I think worth considering: 
* [x] Maybe merging `gaussian_smear.F` and `fermi_utils.F` as one file `smearing_utils.F` will be better?
* [x] For better performance with metals, the Methfessel-Paxton method and Marzari-Vanderbilt method are also worth implementing (also seen from issue #1359).


Growl1234:
MP (first-order only) and MV should work. Here is a simple test using Cu cell: [tests.zip](https://github.com/user-attachments/files/26002357/tests.zip)

Similar to Gaussian smearing:

> Most of the logic is adapted from how the Fermi–Dirac method works; the main differences lie in the mathematical formulas.

src/input_cp2k_scf.F
```diff
@@ -1326,6 +1344,19 @@ SUBROUTINE create_smear_section(section)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
+      CALL keyword_create(keyword, __LOCATION__, &
+                          name="SIGMA", &
+                          description="Smearing width sigma for Gaussian broadening (in energy units). "// &
```
fstein93: Now, this parameter is also used for Methfessel-Paxton and Marzari-Vanderbilt

src/smearing_utils.F
```diff
@@ -0,0 +1,709 @@
...
+
+   ! Unified interface (method as parameter)
+   PUBLIC :: SmearOcc, SmearFixed, SmearFixedDeriv, SmearFixedDerivMV
+   PUBLIC :: Smearkp, Smearkp2
+
+   CHARACTER(len=*), PARAMETER, PRIVATE :: moduleN = 'smearing_utils'
+   INTEGER, PARAMETER, PRIVATE                           :: BISECT_MAX_ITER = 400
+   REAL(KIND=dp), PARAMETER, PRIVATE                     :: sqrtpi = 1.7724538509055159_dp
+   REAL(KIND=dp), PARAMETER, PRIVATE                     :: invsqrt2 = 0.70710678118654752_dp
+   REAL(KIND=dp), PARAMETER, PRIVATE                     :: sqrt2 = 1.4142135623730951_dp
```
fstein93: Please use the [mathconstants](https://github.com/cp2k/cp2k/blob/master/src/common/mathconstants.F) module for it

PR #4960, opened 2026-03-16
Parse cell information from extxyz and pdb formats

he-zilong:
The input keyword `&FORCE_EVAL/&SUBSYS/&CELL/CELL_FILE_FORMAT` allows periodic
boundary condition to be set via external file. This PR extends parser logic so
that `xyz` and `pdb` can also be parsed for cell information in addition to
`cp2k`, `cif` and `xsc`, specifically the `Lattice=` tag in the comments on the
second line of the extended xyz format and the `CRYST1` field of the pdb format.


src/input_cp2k_subsys.F
```diff
@@ -181,12 +182,14 @@ SUBROUTINE create_cell_section_low(section, periodic)
 
       CALL keyword_create(keyword, __LOCATION__, name="CELL_FILE_FORMAT", &
                           description="Specify the format of the cell file (if used)", &
-                          usage="CELL_FILE_FORMAT (CP2K|CIF|XSC)", &
-                          enum_c_vals=s2a("CP2K", "CIF", "XSC"), &
-                          enum_i_vals=[do_cell_cp2k, do_cell_cif, do_cell_xsc], &
+                          usage="CELL_FILE_FORMAT (CP2K|CIF|XSC|XYZ|PDB)", &
+                          enum_c_vals=s2a("CP2K", "CIF", "XSC", "XYZ", "PDB"), &
```
oschuett, line 186: Maybe we should call the option `EXTXYZ` to be consisted with the [trajectory format](https://github.com/cp2k/cp2k/blob/60bb2b342c237adb4008469f08e6c61bb16e1aa0/src/input_cp2k_motion_print.F#L280).

PR #4964, opened 2026-03-16
Add v4.0.x API of DFT-D4

Growl1234:
... while retaining compatibility with the older 3.x version.

### DIFF
--- a/tools/toolchain/scripts/stage8/install_dftd4.sh
+++ b/tools/toolchain/scripts/stage8/install_dftd4.sh
@@ -58,6 +58,7 @@ case "$with_dftd4" in
     echo "==================== Finding DFTD4 from system paths ===================="
     check_command pkg-config --modversion dftd4
     DFTD4_INCLUDE_PATH=$(pkg-config --cflags dftd4 | awk '{print $1}' | cut -dI -f2)
+    pkg_install_dir=$(dirname ${DFTD4_INCLUDE_PATH})
     add_include_from_paths DFTD4_CFLAGS "dftd4.h" $DFTD4_INCLUDE_PATH
     add_include_from_paths DFTD4_CFLAGS "dftd4.mod" $DFTD4_INCLUDE_PATH
     add_include_from_paths DFTD4_CFLAGS "mctc_io.mod" $DFTD4_INCLUDE_PATH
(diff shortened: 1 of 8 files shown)


tools/toolchain/scripts/stage8/install_dftd4.sh
```diff
@@ -32,18 +32,15 @@ case "$with_dftd4" in
     install_lock_file="${pkg_install_dir}/install_successful"
 
     if verify_checksums "${install_lock_file}"; then
-      echo "dftd4_dist-${dftd4_ver} is already installed, skipping it."
+      echo "dftd4-${dftd4_ver} is already installed, skipping it."
     else
-      retrieve_package "${dftd4_sha256}" "dftd4-${dftd4_ver}.tar.gz"
+      retrieve_package "${dftd4_sha256}" "dftd4-${dftd4_ver}-source.tar.xz"
```
Growl1234, line 37: Note: By default, `--install-all` skip the installation of a standalone DFT-D4 due to installation of tblite, so here I don't need to temporarily modify the downloading behavior. The download link of DFT-D4 v4.0.2 is here: https://github.com/dftd4/dftd4/releases/download/v4.0.2/dftd4-4.0.2-source.tar.xz

jpoto:
It looks good to me.

jpoto:
There is now an issue for the dft-d4 update in tblite:
https://github.com/tblite/tblite/issues/315

Growl1234:
> There is now an issue for the dft-d4 update in tblite: [tblite/tblite#315](https://github.com/tblite/tblite/issues/315)

Thanks! In fact we don't need to wait the newer version of tblite to be released because the compatibility of dft-d4 v3 is intentionally remained in case someone use dft-d4 package installed on their own (`--with-dftd4=system`). But does that mean the issue #4966 has to do with dft-d4 package itself?

jpoto:
No the issue is related to the initialization. Due to the changing errors it takes some time to pinpoint it.

jpoto:
I would suggest to merge it then I can work on the issue  #4966, which requires several small fixes.

Growl1234:
> I would suggest to merge it then I can work on the issue #4966, which requires several small fixes.

CC: @oschuett

Growl1234:
> I would suggest to merge it then I can work on the issue #4966, which requires several small fixes.

I did a final check and fixed some minor bugs related to `__SYSTEM__` option. Now it should be ready to merge, and the dftd4-source tarball could also be uploaded.

CC: @oschuett

PR #4970, opened 2026-03-17
Simplify CMake options output in toolchain

Growl1234:
Current output is too verbose for a terminal, so reverts back to original format. The detailed instruction or explanation is now written to a Markdown file without directly outputting, which is also convenient to expand. Also simplifies some output via preferring a relative directory in the CMake options.

Along with minor changes:
1. Because `generate_cmake_options.sh` uses relative path to find `CP2K_ROOT`, we must ensure users use a complete source tree. This detection is enhanced by changing warning message to error if `data` directory is not found in the source tree.
2. Makes the path to `do_regtests.py` shown at cmake step an absolute path in case someone who build CP2K on their own run `cmake` inside the `build` directory.
3. Still forces `CMAKE_INSTALL_LIBDIR` to `lib` to provide convenience for both current toolchain and `make_cp2k.sh`.


he-zilong:
> Current output is too verbose for a terminal, so reverts back to original format. The detailed instruction or explanation is now written to a Markdown file without directly outputting, which is also convenient to expand.

Do you intend this Markdown file to be read by user in raw form or in rendered form? The former is easy with Linux-native CLI utilities like `less` or `more`; the latter, not so much besides VSCode. (I know there are a few other Markdown files in the package, but I doubt how many an everyday user read them...)

Growl1234:
I think the raw form is enough; the structure of this Markdown is simple. If users really want a best readability they can also use `vim` (usually with colours).

he-zilong:
Should we also add some notes about how to write a job submission script for running regtests? IIRC there is an instruction for that elsewhere in the manual/documents.

Growl1234:
> IIRC there is an instruction for that elsewhere in the manual/documents.

It's here: https://www.cp2k.org/dev:regtesting
We can add this link to to the instruction file.

oschuett:
> It's here: https://www.cp2k.org/dev:regtesting

The wiki is kinda deprecated. The best place for curated documentation is the [manual](https://manual.cp2k.org).

We could e.g. add pages for _"Testing"_ and _"HPC"_ to the _"Technologies"_ section.

Growl1234:
It sounds great to add related pages to the manual! In my view, however, to do this effectively, we would need to accompany it with broader revisions and expansions to the existing manual. Personally, I feel that adding this type of documentation directly to the 'Technologies' section might not be the best fit.

Perhaps a more effective approach for now would be to expand the "Getting Started" section, or create a new section named like "For HPC clusters". We can also migrate the messages about HPC added in the toolchain help information to that page.

oschuett:
> It sounds great to add related pages to the manual! In my view, however, to do this effectively, we would need to accompany it with broader revisions and expansions to the existing manual.

Sure! Maybe you could start by opening an issue, where you outline your ideas for reorganizing the manual. Then we can briefly discuss your plans before we dive in.

> Perhaps a more effective approach for now would be to expand the "Getting Started" section,...

I think, it depends on the intended audience. Help for new users should indeed go into the _Getting Started_ section.

> ...or create a new section named like "For HPC clusters".

Maybe. Personally, it feels a bit too narrow for a top level section.

> We can also migrate the messages about HPC added in the toolchain help information to that page.

Yes. And we can also continue to move content from the wiki to the manual (I have a script for that).

Growl1234:
> Sure! Maybe you could start by opening an issue, where you outline your ideas for reorganizing the manual. Then we can briefly discuss your plans before we dive in.

I have opened an issue to outline my (long-term) expectations for the CP2K documentation. Sincerely hopes to see CP2K getting better and becoming a more perfect program.

CMakeLists.txt
```diff
@@ -106,6 +106,10 @@ foreach(__var ROCM_ROOT CRAY_ROCM_ROOT ORNL_ROCM_ROOT CRAY_ROCM_PREFIX
   endif()
 endforeach()
 
+set(CMAKE_INSTALL_LIBDIR
```
LecrisUT, line 109: The better change here would be to define `CP2K_INSTALL_*` with defaults from GNUInstallDirs variables. The calculation of `CMAKE_INSTALL_*` is quite complex it is best not to override it.

PR #4977, opened 2026-03-20
Revise build instructions for CP2K from source

mkrack:
Updated the documentation to reflect the removal of one build method and added detailed instructions for using the make_cp2k.sh script.


Growl1234:
I think it's duplicate of #4978.

PR #4979, opened 2026-03-20
remove projMO REFERENCE_TYPE

Anyborr:
Removes `REAL_TIME_PROPAGATION % PRINT % PROJECTION_MO % REFERENCE_TYPE` as an option. This is because one of the two options (`REFRERENCE_TYPE XAS_TDP`) is (1) superfluous; one can achieve the same results by using the `REFERENCE_TYPE SCF` option instead, and (2) it was not working properly if the reference .wfn was obtained from an XAS_TDP calculation where the source electron was not from the lowest MO of the system.

The only reason to keep the `REFERENCE_TYPE` option is to expand the types of MOs to project onto, e.g., an .rtpwfn file. But there does not seem to be a demand for this functionality at the moment.

### DIFF
--- a/src/emd/rt_projection_mo_utils.F
+++ b/src/emd/rt_projection_mo_utils.F
@@ -35,8 +35,6 @@ MODULE rt_projection_mo_utils
                                               cp_print_key_generate_filename,&
                                               cp_print_key_should_output,&
                                               cp_print_key_unit_nr
-   USE input_constants,                 ONLY: proj_mo_ref_scf,&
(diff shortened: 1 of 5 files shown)


oschuett:
@Anyborr, could you please update the documentation on [Real-Time Propagation](https://github.com/cp2k/cp2k/blob/master/docs/methods/sampling/ehrenfest.md). It still refers to the old `REFERENCE_TYPE` keyword and the `XAS_TDP` option.

PR #4985, opened 2026-03-22
Use Newton's minimization method for MP and MV smearing

Growl1234:
https://doi.org/10.1103/PhysRevB.107.195122


fstein93:
If your implementation is a replication of the paper you provided, can you add it to the [bibliography](https://github.com/cp2k/cp2k/blob/master/src/common/bibliography.F) and reference to it?

PR #4987, opened 2026-03-23
Fix test-farm failure of RPM build

Growl1234:
The original `do_regtests.sh` script contains a critical flaw: if a regtesting directory already exists from a previous run, the script will repeatedly create new subdirectories within it, leading to infinite recursion and nesting, eventually causing the testing farm to crash. This PR should fix this issue. Also dealt with some (probable) regressions in spec file.

The testing farm on Fedora Rawhide is currently not able to work, because LMod and its dependencies are not yet compatible with the Lua 5.5 version shipped in Rawhide. For robustness, I’m hesitating whether we should replace the extremely aggressive Rawhide with the latest **stable** release (`latest-stable` flag; currently Fedora 43, vs Fedora 44 as `latest` flag actually in development or beta status) in the configuration file; or, we can use `fedora-branched` instead.

### DIFF
--- a/tools/fedora/tests/do_regtest.sh
+++ b/tools/fedora/tests/do_regtest.sh
@@ -32,10 +32,18 @@ rlJournalStart
       rlRun "ls /usr/bin/cp2k.ssmp" 0 "Verify CP2K serial binary exists"
     fi
 
+    # Temporary change to try looking for why performance issue happen
+    if [[ "${CP2K_VARIANT}" == "openmpi" ]]; then
+      rlRun "cat /proc/sys/kernel/yama/ptrace_scope" 0 "Check ptrace scope (0=CMA works, 1+=CMA blocked)"
+      rlRun "df -h /dev/shm" 0 "Check shared memory size"
+    fi
+
     rlRun "./do_regtest.py $args" 0 "Run regression tests"
 
     if [[ -d "${workdir}" ]]; then
-      rlFileSubmit "${workdir}/TEST-"*/status.txt "regtest-${CP2K_VARIANT}-status.txt" 2>/dev/null || true
+      for file in ${workdir}/TEST-*/status.txt regtest-${CP2K_VARIANT}-status.txt; do
+        rlFileSubmit $file
+      done
     fi
 
     rlRun "popd"
--- a/tools/fedora/tests/do_regtest.sh
+++ b/tools/fedora/tests/do_regtest.sh
@@ -28,11 +28,6 @@ rlJournalStart
 
     if [[ "${CP2K_VARIANT}" != "serial" ]]; then
       rlRun "ls ${MPI_BIN}/cp2k.psmp" 0 "Verify CP2K MPI binary exists"
-      if [[ "${CP2K_VARIANT}" == "openmpi" ]]; then
-        rlRun "time mpiexec -n 2 cp2k.psmp --version > /dev/null" 0 "Check the PRTE startup overhead in OpenMPI 5.x."
-      else
-        rlRun "time mpiexec -n 2 cp2k.psmp --version > /dev/null"
-      fi
     else
       rlRun "ls /usr/bin/cp2k.ssmp" 0 "Verify CP2K serial binary exists"
     fi
--- a/tools/fedora/tests/do_regtest.sh
+++ b/tools/fedora/tests/do_regtest.sh
@@ -28,15 +28,15 @@ rlJournalStart
 
     if [[ "${CP2K_VARIANT}" != "serial" ]]; then
       rlRun "ls ${MPI_BIN}/cp2k.psmp" 0 "Verify CP2K MPI binary exists"
+      if [[ "${CP2K_VARIANT}" == "openmpi" ]]; then
+        rlRun "time mpiexec -n 2 cp2k.psmp --version > /dev/null" 0 "Check the PRTE startup overhead in OpenMPI 5.x."
+      else
+        rlRun "time mpiexec -n 2 cp2k.psmp --version > /dev/null"
+      fi
     else
       rlRun "ls /usr/bin/cp2k.ssmp" 0 "Verify CP2K serial binary exists"
     fi
 
-    # Temporary change to try looking for why performance issue happen
-    if [[ "${CP2K_VARIANT}" == "openmpi" ]]; then
-      rlRun "cat /proc/sys/kernel/yama/ptrace_scope" 0 "Check ptrace scope (0=CMA works, 1+=CMA blocked)"
-      rlRun "df -h /dev/shm" 0 "Check shared memory size"
-    fi
 
     rlRun "./do_regtest.py $args" 0 "Run regression tests"
 
--- a/tools/fedora/tests/do_regtest.sh
+++ b/tools/fedora/tests/do_regtest.sh
@@ -9,7 +9,10 @@ rlJournalStart
 
     rlRun "pushd tests" 0 "cd to cp2k tests folder"
 
-    rlRun "args=\"--maxtasks $(nproc) --ompthreads 2\"" 0 "Set base arguments"
+    workdir="/tmp/cp2k-regtest-${CP2K_VARIANT}"
+    rlRun "rm -rf ${workdir}" 0 "Clean previous workdir"
+    rlRun "args=\"--workbasedir ${workdir} --maxtasks $(nproc) --ompthreads 2\"" 0 "Set base arguments"
+
     [[ "${CP2K_SKIP_UNITTEST,,}" == "true" ]] && rlRun "args=\"\$args --skip_unittests\"" 0 "Skip unit tests"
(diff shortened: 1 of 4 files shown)


oschuett:
> the script will repeatedly create new subdirectories within it, leading to infinite recursion and nesting, 

I also stumbled across this problem in the past and I'm now adding a safeguard for it #4988.

> For robustness, I’m hesitating whether we should replace the extremely aggressive Rawhide with the latest stable release...

These packit tests are meant to catch packaging problems early to reduce the workload of [keeping CP2K in Fedora](https://github.com/cp2k/cp2k/pull/3256#issuecomment-1937890159). We do have a [separate test](https://dashboard.cp2k.org/archive/fedora-psmp/index.html) that target a recent release version of Fedora.
 
That being said, I also noticed that the packit tests are often broken. So, maybe there is something we could improve? What do you think @LecrisUT?

LecrisUT:
> That being said, I also noticed that the packit tests are often broken. So, maybe there is something we could improve? What do you think @LecrisUT?

Main issue is to get notifications on both sides when something goes wrong so we can help each other work together on them. Packit would help make pings when something goes wrong at a PR level, but if they are never run for PRs, then it is hard to pick them up.

> > For robustness, I’m hesitating whether we should replace the extremely aggressive Rawhide with the latest stable release...
> 
> These packit tests are meant to catch packaging problems early to reduce the workload of [keeping CP2K in Fedora](https://github.com/cp2k/cp2k/pull/3256#issuecomment-1937890159). We do have a [separate test](https://dashboard.cp2k.org/archive/fedora-psmp/index.html) that target a recent release version of Fedora.

Please do test `rawhide`. That is the only place where the latest `gfortran` changes will land and it is important to detect issues encountered there early so I could forward any such issues to `gcc` to be addressed. If you do find any issues with any components failing, do report to me and I can help and get them fixed asap.

tools/fedora/tests/do_regtest.sh
```diff
@@ -22,8 +25,19 @@ rlJournalStart
...
+    if [[ "${CP2K_VARIANT}" != "serial" ]]; then
+      rlRun "ls ${MPI_BIN}/cp2k.psmp" 0 "Verify CP2K MPI binary exists"
+    else
+      rlRun "ls /usr/bin/cp2k.ssmp" 0 "Verify CP2K serial binary exists"
+    fi
+
     rlRun "./do_regtest.py $args" 0 "Run regression tests"
 
+    if [[ -d "${workdir}" ]]; then
+      rlFileSubmit "${workdir}/TEST-"*/status.txt "regtest-${CP2K_VARIANT}-status.txt" 2>/dev/null || true
```
LecrisUT:
```suggestion
      for file in ${workdir}/TEST-*/status.txt regtest-${CP2K_VARIANT}-status.txt; do
        rlFileSubmit $file
      done
```

LecrisUT:
OpenMPI issues will be fixed in about 24h https://bugzilla.redhat.com/show_bug.cgi?id=2441371. Technically can be picked up faster, but long story short, that feature is still being worked on

Growl1234:
> OpenMPI issues will be fixed in about 24h https://bugzilla.redhat.com/show_bug.cgi?id=2441371. Technically can be picked up faster, but long story short, that feature is still being worked on

Thanks! Another problem is the OpenMPI version of rpm-built CP2K runs regtests much more slowly than the MPICH version. I suspect it has to do with how OpenMPI works and the limitations brought by the virtual machine or container.

Update 1: CMA and `/dev/shm` are both normal, so ruled out.

LecrisUT:
> > OpenMPI issues will be fixed in about 24h https://bugzilla.redhat.com/show_bug.cgi?id=2441371. Technically can be picked up faster, but long story short, that feature is still being worked on
> 
> Thanks! Another problem is the OpenMPI version of rpm-built CP2K runs regtests much more slowly than the MPICH version. I suspect it has to do with how OpenMPI works and the limitations brought by the virtual machine or container.

If you have any ideas on what can affect it, I can help you get the environment needed to test either cases. Afaict these are VMs right now. If you need access to an environment that you can ssh, you can have access to one if you run the `testing-farm` cli manually (with appropriate token from a FAS account).

RE OpenMPI bug, it seems more involved, but I am working with the people involved with it to get it fixed asap

Growl1234:
> If you have any ideas on what can affect it, I can help you get the environment needed to test either cases. Afaict these are VMs right now. If you need access to an environment that you can ssh, you can have access to one if you run the `testing-farm` cli manually (with appropriate token from a FAS account).

Thanks a lot! It's definitely benificial to have a better way to do related testings! My username of account on Fedora webpages is [growl](https://accounts.fedoraproject.org/user/growl/).

LecrisUT:
> > If you have any ideas on what can affect it, I can help you get the environment needed to test either cases. Afaict these are VMs right now. If you need access to an environment that you can ssh, you can have access to one if you run the `testing-farm` cli manually (with appropriate token from a FAS account).
> 
> Thanks a lot! It's definitely benificial to have a better way to do related testings! My username of account on Fedora webpages is [growl](https://accounts.fedoraproject.org/user/growl/).

So the place to get the testing-farm token is https://testing-farm.io/, from there you can use the `testing-farm` cli (simplest is from the the container `podman run quay.io/testing-farm/cli`). It should be under `testing-farm reserve`. I don't know if there is a shortcut for requesting from a previous run, I can ask them, but otherwise the full details on what is being set is in the [request](https://api.testing-farm.io/v0.1/requests/6bf97f04-0fa6-4db5-a829-ab5a1b99980a). Likely you would not need all of that if you know how to manually work with the copr there.

Btw [#fedora-ci](https://matrix.to/#/#fedora-ci:fedoraproject.org) and [#Fedora science](https://matrix.to/#/#scitech:fedoraproject.org) is where you can catch me in a sync chat if needed

Growl1234:
> So the place to get the testing-farm token is https://testing-farm.io/ ...

It seems that I need to "be part of an additional group in the Fedora Account System (either `fedora-contributor` or `testing-farm` group)."

LecrisUT:
> > So the place to get the testing-farm token is https://testing-farm.io/ ...
> 
> It seems that I need to "be part of an additional group in the Fedora Account System (either `fedora-contributor` or `testing-farm` group)."

Huh, I thought that is automatic after signing the FPCA. Can check the details on that.

Growl1234:
From [this log](https://artifacts.dev.testing-farm.io/b480c212-8571-43ae-bb33-ff8a68521089/work-smokee3b7udie/log.txt) I can see that the startup overhead of `mpiexec` (`mpirun`) in OpenMPI is already an issue.

**MPICH**
```
17:45:20                 stdout: real	0m0.702s
17:45:20                 stdout: user	0m0.621s
17:45:20                 stdout: sys	0m0.422s
```
**OpenMPI**
```
17:53:30                 stdout: real	0m1.772s
17:53:30                 stdout: user	0m0.457s
17:53:30                 stdout: sys	0m0.603s
```

Considering that CI issues related to OpenMPI are not limited to the Fedora testing farm, and problems can also be encountered in the [toolchain](https://dashboard.cp2k.org/archive/openmpi-psmp/index.html) and [Spack](https://dashboard.cp2k.org/archive/spack-openmpi-psmp/index.html) with a clear backtrace referring to the MPI behavior, I believe that to truly resolve them we may need to engage with the developers in the [OpenMPI repository](https://github.com/open-mpi/ompi).

Growl1234:
It seems [these variables added](https://github.com/Growl1234/cp2k/blob/b824defb2104bd7bdc3b26f2b4c8e49db4d6a038/tools/fedora/tests/regression.fmf#L22-L24) work for the VM environment in test-farm. That said, I still believe that OpenMPI has room for improvement in how it handles containerized or virtualized environments. It is also possible that the OpenMPI-related issues in CP2K CI could be mitigated by tuning certain environment variables (may or may not the ones added here).

Growl1234:
If the results of this test are also as expected, I personally think this PR can be ready to merge. As for the compatibility issues between Lmod (and its dependencies) and Lua 5.5 in Rawhide, we may have to wait (?) before they are resolved by upstream repos.

LecrisUT:
> If the results of this test are also as expected, I personally think this PR can be ready to merge. As for the compatibility issues between Lmod (and its dependencies) and Lua 5.5 in Rawhide, we may have to wait (?) before they are resolved by upstream repos.

Yeah, this one wound up to be quite a mess.

> It seems that I need to "be part of an additional group in the Fedora Account System (either `fedora-contributor` or `testing-farm` group)."

Yeah, got confirmation that it is more complicated. I can do the reservation on your behalf, but that would only be feasible if we do it in a sync chat. There does seem to be some issues with the reservation right now though.

> That said, I still believe that OpenMPI has room for improvement in how it handles containerized or virtualized environments. It is also possible that the OpenMPI-related issues in CP2K CI could be mitigated by tuning certain environment variables (may or may not the ones added here).

Yeah this requires quite a deep dive in the options that openmpi sets by default. One place to check what default options to set is in the [openmpi macro](https://src.fedoraproject.org/rpms/openmpi/blob/rawhide/f/macros.openmpi)

PR #4991, opened 2026-03-23
LS-SCF: Abort when analytical stress tensor is requested

oschuett:


juerghutter:
Does this apply to all of LS-SCF or only PAO?
Stress tensor calculation should not depend on how the density matrix has been calculated.

oschuett:
I was looking for the stress tensor implementation of LS-SCF in general and couldn't find it. Then I realized that there are no tests for it nor is it mentioned in the [paper](https://doi.org/10.1021/ct200897x). So, I concluded that it's simply not implemented. I'd love to be wrong on this. In that case, we should really add a test for it.

juerghutter:
I made a quick test and it seems to work. One should also check other properties like the polarizability.
LS-SCF/PAO is another story.

PR #4992, opened 2026-03-23
Activate Fedora rawhide tester (spack)

mkrack:
This tester is supposed to detect problems with the latest GCC version early and is therefore likely to fail in most cases.


mkrack:
The [COSMA compilation error](https://github.com/eth-cscs/COSMA/issues/160) is still present with COSMA 2.8.1.

PR #4995, opened 2026-03-23
Toolchain: OpenMPI 5.0.9 -> 5.0.10

Growl1234:


Growl1234:
Current CI use MPICH by default, so I didn't modify the downloading behavior temporarily. Could anyone please upload the [OpenMPI package](https://download.open-mpi.org/release/open-mpi/v5.0/openmpi-5.0.10.tar.bz2) to the CP2K server? Also, if possible, I want to then have a look again at what OpenMPI test would show after this update.

oschuett:
Voilà: https://www.cp2k.org/static/downloads/openmpi-5.0.10.tar.bz2

oschuett:
>  Also, if possible, I want to then have a look again at what OpenMPI test would show after this update.

I've triggered the OpenMPI test. It will keep re-starting when you update the branch as along as you don't force push.

PR #5001, opened 2026-03-25
Toolchain: Update CMake, OpenBLAS, HDF5, COSMA, and ScaLAPACK

Growl1234:


oschuett:
I've uploaded all the new tarballs.

Growl1234:
COSMA version 2.8.2 just released. We can now update it with removing the patches added previously.

mkrack:
The new COSMA v2.8.2 tar ball is uploaded.

PR #5006, opened 2026-03-25
Docker: Update to the latest HIP container

Growl1234:
For compatibility of CMake >=4.0

### DIFF
--- a/tools/docker/Dockerfile.test_spack_pdbg
+++ b/tools/docker/Dockerfile.test_spack_pdbg
@@ -43,13 +43,13 @@ COPY . cp2k/
 
 # Build CP2K dependencies
 WORKDIR /opt/cp2k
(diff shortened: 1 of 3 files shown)


oschuett:
Let's see how this goes. Last time I hit [compiler bugs](https://github.com/cp2k/cp2k/pull/4549#issuecomment-3617725638).

Growl1234:
It seems DBCSR couldn't find `hiprtc::hiprtc` target, which should be in path [/opt/rocm/lib/cmake/hiprtc](https://github.com/user-attachments/files/26245239/hiprtc.zip). I need to first update to DBCSR 2.9.1.

oschuett:
I've uploaded dbcsr-2.9.1.tar.gz

Growl1234:
Oops that's my fault. I'll deal with this mistake soon.

Issue #4966, opened 2026-03-16
Problems with `D4_REFERENCE_CODE F`


Growl1234:
Here are two examples from regtest `QS/regtest-dft-vdw-corr-4`:

[pbe_dftd4(_bsse).zip](https://github.com/user-attachments/files/26032868/pbe_dftd4._bsse.zip)

The first test `pbe_dftd4` gives a nearly wrong result: `Dispersion energy:   -0.00283209078164`, while the reference value is `-0.00283102230260`. This does not happen when `D4_REFERENCE_CODE` is set to T (by default).

The second test occurs segmentation fault, which maybe has to do with how the built-in d4 codes deal with ghost atoms. Sometimes it gives different error messages.

In most cases of 2D or 3D periodic systems (according to my own tests), the codes work correctly; however, the tests above suggest that there may indeed be some issues when dealing with (at least) molecules and systems with ghost atoms.

jpoto:
I will take a look. Do you use the current master branch?

Growl1234:
Yes. Thanks for your dedication!

P.S. What I wish to do is that after the problems (at least) in those regtests are solved then the default value of `D4_REFERENCE_CODE` can be safe to be set to F to take full advantage of CP2K's own codes to improve computational efficiency.

jpoto:
For the old version of DFT-D4(V3), I get a different error. 

For DFT-D4 (V4), I will need to use your pull request to get the segmentation fault.

Growl1234:
Yeah, I found it could produce various types of errors at each run. I just tried re-running the test `pbe_dftd4_bsse` several more times, and here are what I can collect at the moment:

[errors.zip](https://github.com/user-attachments/files/26056863/errors.zip)

Sometimes it seems the error is not directly from D4 codes, but all error messages disappear when I don't use `D4_REFERENCE_CODE F` ...

It seems such an abort message could also happen:
```
 *******************************************************************************
 *   ___                                                                       *
 *  /   \                                                                      *
 * [ABORT]                                                                     *
 *  \___/                      list_routinestat_get: pos < 1                   *
 *    |                                                                        *
 *  O/|                                                                        *
 * /| |                                                                        *
 * / \                                           common/list_routinestat.F:292 *
 *******************************************************************************
```

jpoto:
@juerghutter 
It seem that the problem is that eeq_charges accesses the charges of ghost atoms, which are not available in the vector that is provided to the function.

How do you suggest to solve this problem?

[...]

Issue #3961, opened 2025-02-10
[CP2K:21126] Counterpoise correction with Grimme D4


juerghutter:
D4 calculations ignore ghost atom setting resulting in wrong BSSE calculations.
Holger Sassnick

jpoto:
Should I take a look?

juerghutter:
You are welcome to have a look.
The atoms/coordinates setup will have to be screened for ghost atoms.
This has to apply (please check) for the energy calculation as well as
the EEQ charges.
The ghost atoms are tagged, have a look in the D3 code.

________________________________________

jpoto:
I have a draft under: https://github.com/cp2k/cp2k/pull/3985 but i am not sure it is correct.

juerghutter:
I haven't yet time to llok at it. A basic test would be to see that a BSSE calculation with and without D4 gives exactly the same result.

jpoto:
Hi Juerg,

I have set up a test and get:
1                Dispersion energy:                                           -0.00019977569183 
2                Dispersion energy:                                           -0.00019977558874
1+ghosts  Dispersion energy:                                           -0.00019977569183
2+ghosts  Dispersion energy:                                           -0.00019977558874
1+2            Dispersion energy:                                           -0.00087282454382

The first and second subsystem get the now the same energy, independent of the ghost atoms. 
Did you mean this?

You can find it under: https://github.com/cp2k/cp2k/pull/4005

juerghutter:
Hi

subsystems calculated with and without the ghost atom basis should have
the same dispersion energy. I think this is Energies 1 and 1+ghosts, as well
as 2 and 2+ghosts. That 1 and 2 have almost the same dispersion energy is
by chance.

What I meant was to do a full BSSE calculation of a system with
GGA + D4 and with GGA only.
Then the BSSE correction of the two calclulations should be the same.

This boiles down to the equivalence of the two energies in your example.

Juerg

________________________________________

jpoto:
It seems to work:
[pbe_dftd4_bsse.inp.out.txt](https://github.com/user-attachments/files/18887945/pbe_dftd4_bsse.inp.out.txt)
[pbe_dft_bsse.inp.out.txt](https://github.com/user-attachments/files/18887944/pbe_dft_bsse.inp.out.txt)

-34.358706+35.0553749=0.6966689
-34.357833+35.0545020=0.696669

Growl1234:
[...]

PR #5008, opened 2026-03-25
Jvp ghost d4

jpoto:
This should take care of the segmentation fault in #4966. It needs further testing for accuracy.

### DIFF
--- a/src/qs_dispersion_d4.F
+++ b/src/qs_dispersion_d4.F
@@ -21,6 +21,7 @@ MODULE qs_dispersion_d4
                          plane_distance, &
                          pbc, &
                          get_cell
+   USE qs_dispersion_types, ONLY: qs_atom_dispersion_type
    USE qs_environment_types, ONLY: get_qs_env, &
                                    qs_environment_type
    USE qs_force_types, ONLY: qs_force_type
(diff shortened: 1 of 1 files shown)


Growl1234:
Thanks for the fix! The segmentation fault has disappeared with this PR applied. However, with `D4_REFERENCE_CODE F` the result is `BSSE-free interaction energy:  0.006775` v.s. reference value 0.006697.

juerghutter:
I have doubts that this will fix D4 calculations with ghost atoms. Maybe it works for the reference (DFTD4) implementation,
but for the CP2K native implementation it will run into problems with eeq_charges. That routine is not enabled
for ghost atoms. Fixing it is non trivial. I would suggest to check for ghost atoms in eeq_charges and then throw an error.

jpoto:
@juerghutter 
Yes. I hoped you would know an easy way to restrict the atoms in eeq_charges, maybe even something more elegant for the changed part. 

Do you you think the speed-up is worth the effort to rewrite eeq_charges?

juerghutter:
I don't see a way to do this in eeq_charges without a major rewrite.
I don't understand what you mean by speed-up in this context. It's about functionality alone.

jpoto:
Is your eeq solver using MPI parallelization?

juerghutter:
Yes it is using FM matrices and the Ewald routines for PBC.

Growl1234:
Hi @jpoto,

Would you like to have a look at #5030 as a possible alternative, where I also add an optional section into `eeq_method` to force the exclusion of ghost atoms when doing D4 calculation? The result of `pbe_dftd4_bsse` now seems to be correct.

juerghutter:
Yes, this solution looks reasonable to me.
JH

he-zilong:
Is this draft still relevant now that #5030 has been merged? Has the related open issue #3961 been addressed?

PR #5026, opened 2026-03-30
Conventions: Stop reading PACKAGE files

oschuett:
Follow up to #4996.


he-zilong:
... Henceforth obsoleting the page [dev:packages](https://www.cp2k.org/dev:packages) and possibly `./tools/package_planner`?

PR #5030, opened 2026-03-30
Fix issues regarding native DFT-D4 codes

### LINKED ISSUE #4966: Problems with `D4_REFERENCE_CODE F`

Growl1234:
This should take care of below issues:
1. D4 correction for systems including ghost atoms encounters segmentation faults.
2. D4 correction energy from native codes is not the same as reference codes because smooth logistic CN cutoff in multicharge not included in native codes.

It can be an alternative approach of #5008 and fix #4966.

### DIFF
--- a/docs/technologies/libraries.md
+++ b/docs/technologies/libraries.md
@@ -207,6 +207,7 @@ by passing `-DCP2K_USE_ACE=ON` to CMake.
 ## DFTD4 (dispersion correction)
 
 - dftd4 - Generally Applicable Atomic-Charge Dependent London Dispersion Correction.
+- Please always use the CMake-built dftd4 package rather than the Meson-built one for CP2K.
 - For more information see <https://github.com/dftd4/dftd4>
 - Pass `-DCP2K_USE_DFTD4=ON` to CMake.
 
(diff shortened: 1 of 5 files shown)


oschuett:
@jpoto, I believe you are most familiar with the DFT-D4 code. Could you review this PR?

jpoto (approved):
It looks good and hopefully fixes the problems with the ghost atoms.

PR #5036, opened 2026-03-31
MPICH 4.3.2 -> 5.0.1

Growl1234:
This draft is to see if the coming MPICH release works fine with CP2K. [Unfortunately local testing of Spack fails due to network issue :-( ]

Previously MPICH 5.0.0 release has a [bug](https://github.com/pmodels/mpich/pull/7722) that makes it not work with some Fortran-based programs including CP2K and DBCSR. Now the [fix](https://github.com/pmodels/mpich/pull/7722) has been backported to 5.0.x branch and the upstream developers are planning a [new 5.0.1 release](https://github.com/pmodels/mpich/issues/7753#issuecomment-4143174167). Since this is a major version update of MPICH for CP2K, I think it's worth having a look before its final release and giving timely feedbacks (if there are issues).

Once MPICH 5.0.1 is finally released, this PR will be set ready for review; before that, please temporarily do not upload the tarball of RC version. The local recipe for Spack is added also temporily and will be dropped once 5.0.1 is provided via Spack packages.


Growl1234:
MPICH v5.0.1 is released now: https://github.com/pmodels/mpich/releases/tag/v5.0.1

Could anyone please upload the [tarball](https://github.com/pmodels/mpich/releases/download/v5.0.1/mpich-5.0.1.tar.gz) to the CP2K server?

PR #5043, opened 2026-04-02
GExt extrapolation in CP2K

mnottoli:
Added two new extrapolation methods for molecular dynamics simulations.

In both cases, the new coefficient matrix is obtained as an approximation of PS times the coefficient matrix at the previous step, where PS is the density matrix times the overlap matrix.
```math
C_{n+1}^* = (PS)^*_{n+1} C_{n}
```
The approximation of PS is obtained by doing a linear combination of (PS)_i at q previous steps of MD.
```math
(PS)^*_{n+1} \simeq \sum_i^q \alpha_i (PS)_{n-q+i}
```
The coefficients of the linear combination are obtained by approximating the current overlap as a linear combination of the q previous overlaps
```math
S_{n+1} \simeq \sum_{i=1}^q \alpha_i S_{n-q+i}
```

Practically, the minimization is slightly different to enforce useful properties:
- in the scheme `GEXT_PROJ`, the minimization takes the overlap S_n as a reference and ensures that the sum of the coefficients is 1 (it might be important in geometry optimizations)  
- In the scheme `GEXT_PROJ_QTR` a quasi time-reversibility is introduced by considering the overlaps in time symmetric pairs (it should help with the energy conservation in MD simulations)

This pull request adds two new keywords for `EXTRAPOLATION`, namely `GEXT_PROJ` and `GEXT_PROJ_QTR`.
Most of the work is done in `wfi_extrapolate`, two additional subroutines are used to solve the two least square problems: `dbcsr_diff_fitting`, `dbcsr_tr_fitting`. As they both work with the overlaps, they use the dbcsr interface for sparse linear algebra.

At the moment proper tests are missing, if given advice, I am willing to write the tests.


src/qs_wf_history_methods.F
```diff
@@ -1703,4 +1265,163 @@ SUBROUTINE wfi_purge_history(qs_env)
 
    END SUBROUTINE wfi_purge_history
 
+   SUBROUTINE dbcsr_diff_fitting(wf_history, current_overlap, coeffs, nvec)
```
Growl1234: Please add a header containing neccessary description like what other subroutines do. Moreover I'm hesitating if the name "dbcsr_..." is fit here.

src/qs_wf_history_methods.F
```diff
@@ -1065,460 +879,208 @@ SUBROUTINE wfi_extrapolate(wf_history, qs_env, dt, extrapolation_method_nr, &
...
-      END DO
+         CALL qs_rho_update_rho(rho, qs_env=qs_env)
+         CALL qs_ks_did_change(qs_env%ks_env, rho_changed=.TRUE.)
 
-      DO indx = 1, kplocal*nkp_groups
-         CALL cp_fm_cleanup_copy_general(info(indx, 1))
-         CALL cp_fm_cleanup_copy_general(info(indx, 2))
-      END DO
+      CASE (wfi_gext_proj_nr)
+         CPASSERT(.NOT. do_kpoints)
```
Growl1234, line 1150: Are you considering adding a k-points support of the new approach? The current master branch has contained a basic workflow to deal with WFN-based extrapolation with k-points.

Growl1234 (changes requested):
Please rebase your fork and don't use merge commits.

src/qs_wf_history_methods.F
```diff
@@ -1703,4 +1856,163 @@ SUBROUTINE wfi_purge_history(qs_env)
...
+            CALL dbcsr_add(tmp_j, ref_state%overlap, 1.0_dp, -1.0_dp)
+            CALL dbcsr_multiply("N", "N", 1.0_dp, tmp_i, tmp_j, 0.0_dp, tmp_k)
+            CALL dbcsr_trace(tmp_k, A(j-1, i-1))
+            A(i-1, j-1) = A(j-1, i-1)
+         END DO
+      END DO
+
+      ! add the Tikhonov regularization
+      DO i = 1, nvec-1
+         A(i, i) = A(i, i) + 1e-8_dp
```
fstein93: Please refactor this magic constant. Either as `REAL(KIND=dp), PARAMETER :: eps_tikhonov = 1e-8_dp` (or similar) or by introducing it as an input parameter

src/qs_wf_history_methods.F
```diff
@@ -1703,4 +1856,163 @@ SUBROUTINE wfi_purge_history(qs_env)
...
+            CALL dbcsr_add(tmp_j, state%overlap, 1.0_dp, 1.0_dp)
+            CALL dbcsr_multiply("N", "N", 1.0_dp, tmp_i, tmp_j, 0.0_dp, tmp_k)
+            CALL dbcsr_trace(tmp_k, A(j, i))
+            A(i, j) = A(j, i)
+         END DO
+      END DO
+
+      ! add the Tikhonov regularization
+      DO i = 1, ntr
+         A(i, i) = A(i, i) + 1e-8_dp
```
fstein93: Same

fstein93 (commented):
I agree to the comments by @Growl1234 . Please add some tests.

src/qs_wf_history_methods.F
mnottoli: Headers are done
I also removed dbcsr_ from the two routines, it was needed in an intermediate version in which I had the non dbcsr versions around.

src/qs_wf_history_methods.F
mnottoli: It is now an input parameter.

src/qs_wf_history_methods.F
mnottoli: While in principle it is possible, it would require at least a few days to not rush things. I don't know when you would like to merge this PR for the community paper.

Maybe we can do it with a second PR at a later point?

fstein93: I am in favor of a second PR.

Growl1234: I'm surely OK with and looking forward to the next PR!

mnottoli:
I added four tests using the template used for the other extrapolations: two with the "far" structure for GEXT_PROJ and GEXT_PROJ_QTR and two with the normal structure for GEXT_PROJ and GEXT_PROJ_QTR.

There are however a few differences with respect to other tests in the folder:
- The MD is extended to 6 steps to properly test the gext strategy (with order 5)
- I had to change slightly the input velocities to avoid some structures for which the SCF does not converge - it is not due to the extrapolation strategy, it is really a structural problem
- I think that testing only the final energy is not enough, slight changes in the extrapolation strategy result anyway in the same energy. I am also testing the error on approximating the current overlap as a linear combination of previous overlaps.
This required passing the printing options to the two fitting subroutines and adding a custom matcher in `matchers.py`.
Since computing the error on the extrapolated overlap is not necessary except for testing or to see if gext is doing well, it is only done if the print level is above the minimum. I don't know if there are better checks, like a debug option that is only passed to the tests.

mnottoli:
Thanks @Growl1234 and @fstein93 for your help!

src/qs_wf_history_methods.F
```diff
@@ -1703,4 +1857,230 @@ SUBROUTINE wfi_purge_history(qs_env)
...
+      ! is approximated by the linear combination of previous overlaps
+      IF (print_level > low_print_level) THEN
+         CALL dbcsr_copy(tmp_i, current_overlap)
+         DO i = 1, nvec
+            state => wfi_get_snapshot(wf_history, wf_index=i)
+            CALL dbcsr_add(tmp_i, state%overlap, 1.0_dp, -coeffs(i))
+         END DO
+         error = dbcsr_frobenius_norm(tmp_i)/dbcsr_frobenius_norm(state%overlap)
+         IF (io_unit > 0) THEN
+            WRITE (6, *) "GEXT overlap fitting error:", error
```
oschuett, line 1953: Could you please write to `io_unit` instead of hard-coding `6`.

src/qs_wf_history_methods.F
```diff
@@ -1703,4 +1857,230 @@ SUBROUTINE wfi_purge_history(qs_env)
...
```
oschuett, line 2073: Could you please write to `io_unit` instead of hard-coding `6`.

src/qs_wf_history_methods.F
mnottoli: Thanks for noticing, I am on it.

src/qs_wf_history_methods.F
Growl1234: I am currently working on fixing some issues related to the k-point WFN history path, and it is almost close to completion ***[Edit: See #5219]***. If your time and energy are limited, I can try taking over this work after the fix is merged.

PR #5053, opened 2026-04-10
Fix for gext extrapolation.

mnottoli:
I removed the hard coded output unit and formatted the output.


he-zilong:
I wonder if a format like `(/,T2,A,F20.10)` will result in the float-point data `F20.10` overlapping with the text-string string `A`... Could you copy-paste an example snippet of this line from a typical computation?

mnottoli:
Here
` GEXT overlap fitting error:        0.0782301061`

PR #5054, opened 2026-04-10
grid: Launch GPU kernels with 8x8x1 thread blocks

oschuett:
Towards #3925.

### DIFF
--- a/src/grid/gpu/grid_gpu_collint.h
+++ b/src/grid/gpu/grid_gpu_collint.h
@@ -430,7 +430,7 @@ __device__ static void cab_to_cxyz(const smem_task *task, const double *alpha,
             const orbital a = coset_inv[ico];
 #else
   // integrate
-  if (threadIdx.z == 0) { // TODO: How bad is this?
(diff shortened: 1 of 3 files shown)


oschuett:
I benchmarked this change against the new [QS_reference/w64ccGRB.inp](https://github.com/cp2k/cp2k/blob/master/benchmarks/QS_reference/w64ccGRB.inp) on two V100 with 32 MPI ranks:

|                          | Master: 4x4x4 cuda threads | This PR: 8x8x1 cuda threads |
| ------------------------ | ------------------------- | -------------------------- |
| grid_collocate_task_list | 460 seconds               | 479 seconds                |
| grid_integrate_task_list | 504 seconds               | 535 seconds                |




Apparently, the advantage of a 2D thread block for [block_to_cab()](https://github.com/cp2k/cp2k/blob/cb5d5fc1d43ddc38dff796d24324c402552ac14b/src/grid/gpu/grid_gpu_collocate.cu#L215) is less than the disadvantage for the rest of the kernel. So, I'm abandoning this PR.

PR #5058, opened 2026-04-13
benchmarks: Add QS_reference for water different functionals

oschuett:
Follow up to #2934.


oschuett:
Turns out the QS_reference benchmarks take in the CI [roughly the same time](https://storage.googleapis.com/cp2k-ci/run-cp2k-perf-openmp-3638c805_report.txt) as those stated in the [README.txt](https://github.com/cp2k/cp2k/blob/master/benchmarks/QS_reference/README.txt). Hence, I'm only including `w64PBE.inp` and `w64SCAN.inp` for now.

PR #5048, opened 2026-04-08
Refactor: when calling `cp_subsys_create()` do not call `read_cell()` again

he-zilong:
While working on the proposal in #4962 to parse cell information from topology file, I discovered that many environment initialization routines use `read_cell()` and `cp_subsys_create()` together.
https://github.com/cp2k/cp2k/blob/7a2a2325688a604b8fabcf9d9fe03426e2037a48/src/eip_environment.F#L104-L116
https://github.com/cp2k/cp2k/blob/7a2a2325688a604b8fabcf9d9fe03426e2037a48/src/embed_environment.F#L84-L100
https://github.com/cp2k/cp2k/blob/7a2a2325688a604b8fabcf9d9fe03426e2037a48/src/ipi_environment.F#L98-L109
https://github.com/cp2k/cp2k/blob/7a2a2325688a604b8fabcf9d9fe03426e2037a48/src/mixed_environment.F#L86-L102
https://github.com/cp2k/cp2k/blob/7a2a2325688a604b8fabcf9d9fe03426e2037a48/src/nnp_environment.F#L130-L144
This is even the case for long-estabilished modules like FIST...
https://github.com/cp2k/cp2k/blob/7a2a2325688a604b8fabcf9d9fe03426e2037a48/src/fist_environment.F#L143-L165
https://github.com/cp2k/cp2k/blob/7a2a2325688a604b8fabcf9d9fe03426e2037a48/src/fist_environment.F#L197-L201
and QS, lampshaded by a TODO remark.
https://github.com/cp2k/cp2k/blob/7a2a2325688a604b8fabcf9d9fe03426e2037a48/src/qs_subsys_methods.F#L99-L125
https://github.com/cp2k/cp2k/blob/7a2a2325688a604b8fabcf9d9fe03426e2037a48/src/qs_subsys_methods.F#L138-L145
Due to the fact that `cp_subsys_create()` calls `topology_control()` which calls `read_cell()` which already parses input for cell and cell_ref information, a separate call of `read_cell()` before or after calling `cp_subsys_create()` seems like a piece of tech debt that makes maintenance of `read_cell()` harder. In addition, this seems to result in outputs containing `CELL|`, `CELL_REF|` and `CELL_TOP|` printouts which can be confusing.

This PR would like to address this problem by taking cell and cell_ref information from the subsys as treated by `cp_subsys_create()`. This may be a risky move since behaviors in periodic boudary condition (PBC) setup vary from module to module, so substantial behavior-affecting edits are yet to be included in the first commit.

PR #5060, opened 2026-04-13
Refactor: drop `cell` input from `qs_init()` and `qs_subsys_create()`

he-zilong:
A follow-up to #5048 .

### DIFF
--- a/src/bsse.F
+++ b/src/bsse.F
@@ -13,7 +13,6 @@
 ! **************************************************************************************************
 MODULE bsse
    USE atomic_kind_types,               ONLY: get_atomic_kind
-   USE cell_types,                      ONLY: cell_type
    USE cp2k_info,                       ONLY: write_restart_header
    USE cp_external_control,             ONLY: external_control
    USE cp_log_handling,                 ONLY: cp_get_default_logger,&
@@ -236,7 +235,6 @@ CONTAINS
                                                             my_targ, n_rep, num_of_frag, old_size, &
                                                             present_charge, present_multpl
       INTEGER, DIMENSION(:), POINTER                     :: atom_index, atom_list, my_conf, tmplist
-      TYPE(cell_type), POINTER                           :: cell
       TYPE(cp_subsys_type), POINTER                      :: subsys
       TYPE(mp_para_env_type), POINTER                    :: para_env
       TYPE(particle_list_type), POINTER                  :: particles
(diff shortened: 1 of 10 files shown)


he-zilong:
That giant cell/cell_ref mess has finally been removed from `qs_subsys_create()` ! The fewer instances `CALL read_cell()` there are throughout CP2K source code, the easier it is to maintain the fallback behavior where cell is parsed from `&SUBSYS/&TOPOLOGY/&COORD_FILE_NAME` in case `&SUBSYS/&CELL` is not present (which requires passing topology_section to `read_cell()`).

```shell
$ grep "CALL read_cell(" ./src/ -nr
./src/qmmm_create.F:252:      CALL read_cell(qm_cell_small, qm_cell_small, cell_section=cell_section, &
./src/cell_methods.F:525:            CALL read_cell(cell_ref, cell_ref, use_ref_cell=use_ref_cell, &
./src/cp_ddapc_types.F:317:               CALL read_cell(mm_cell, dummy_cell, cell_section=cell_section, para_env=para_env)
./src/topology.F:193:      CALL read_cell(my_cell, my_cell_ref, cell_section=cell_section, &
```

A remaining call to `read_cell()` is in `SUBROUTINE cp_ddapc_ewald_create()`, which is called from `qs_init()`:
https://github.com/cp2k/cp2k/blob/25be7e468aca2dc0ff41adbea2eee2389c373f60/src/cp_ddapc_types.F#L295-L333
https://github.com/cp2k/cp2k/blob/25be7e468aca2dc0ff41adbea2eee2389c373f60/src/qs_environment.F#L365-L367
It seems that `cp_ddapc_ewald_create()` needs `qm_cell` and `subsys_section` as input parameters solely because of QM/MM calculation.

In `SUBROUTINE qmmm_env_create()`, the `CALL fist_init()` and `CALL qs_init()` use the same `subsys_section` which holds information of cell for MM region. The `fist_init()` has `CALL cp_subsys_create()` which has `CALL topology_control()` which has `CALL read_cell()` with `subsys_section` and `topology_section` as input, and `read_cell()` _modifies_ `subsys_section` to set `&SUBSYS/&CELL` by `&SUBSYS/&TOPOLOGY` in the fallback mentioned above, so the correct cell information in `subsys_section` after `fist_init()` should also be available in `qs_init()` and subsequently to `cp_ddapc_ewald_create()`.

The cell parser fallback may benefit from adding regtests, but I would like to leave this for other devs so as to get outside confirmations 😃

PR #5061, opened 2026-04-13
Add spin-channel-specific mixing parameters

Growl1234:
For (potentially) better convergence for some magnetic systems.

For example, users can set:
```
&MIXING
  METHOD BROYDEN_MIXING
  ALPHA 0.4
  ALPHA_MAG 1.6
  BETA 1.0
  BETA_MAG 0.0
  NBUFFER 8
&END MIXING
```

### DIFF
--- a/src/qs_density_mixing_types.F
+++ b/src/qs_density_mixing_types.F
@@ -63,8 +63,11 @@ MODULE qs_density_mixing_types
                                                            max_g2 = -1.0_dp, max_gvec_exp = -1.0_dp, pulay_alpha = -1.0_dp, &
                                                            pulay_beta = -1.0_dp, r_step = -1.0_dp, reg_par = -1.0_dp, &
                                                            sigma_max = -1.0_dp, wc = -1.0_dp, wmax = -1.0_dp
+      ! Spin-channel-specific mixing parameters (for nspin==2, ispin=2 is the magnetization channel)
+      REAL(KIND=dp)                                     :: alpha_mag = -1.0_dp, beta_mag = -1.0_dp
       REAL(KIND=dp), DIMENSION(:), POINTER              :: p_metric => NULL()
       REAL(KIND=dp), DIMENSION(:), POINTER              :: kerker_factor => NULL()
+      REAL(KIND=dp), DIMENSION(:), POINTER              :: kerker_factor_mag => NULL()
       REAL(KIND=dp), DIMENSION(:), POINTER              :: special_metric => NULL()
       REAL(KIND=dp), DIMENSION(:, :), POINTER           :: weight => NULL()
       REAL(KIND=dp), DIMENSION(:, :), POINTER           :: norm_res_buffer => NULL()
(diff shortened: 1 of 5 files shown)


gybyisichi-cmyk:
Thank the developers for their hard work. I'm really excited to see this feature. Does it mean that in the future, when using CP2K with diagonalization and k-points to calculate magnetic materials, we can also achieve good convergence?

Growl1234:
I think there are far more things to do in order to achieve this goal. The main purpose of this PR is to expose some more tunable options that may help improve convergence.





---Original---
From: ***@***.***&gt;
Date: Tue, Apr 14, 2026 01:36 AM
To: ***@***.***&gt;;
Cc: "SY ***@***.******@***.***&gt;;
Subject: Re: [cp2k/cp2k] Add spin-channel-specific mixing parameters (PR #5061)


gybyisichi-cmyk left a comment (cp2k/cp2k#5061)
 
Thank the developers for their hard work. I'm really excited to see this feature. Does it mean that in the future, when using CP2K with diagonalization and k-points to calculate magnetic materials, we can also achieve good convergence?

he-zilong:
Pardon my off-topic question, but is @gybyisichi-cmyk an alternative account to @yisichi-design ?

yisichi-design:
Yes, this is my home account. I will mainly use the account yisichi-design from now on. @he-zilong

oschuett:
Since this is a new feature, could you please add a test for it.

Growl1234:
Thanks for the reminders, I will check one or two regtests tomorrow.



---Original---
From: "Ole ***@***.***&gt;
Date: Wed, Apr 15, 2026 19:46 PM
To: ***@***.***&gt;;
Cc: "SY ***@***.******@***.***&gt;;
Subject: Re: [cp2k/cp2k] Add spin-channel-specific mixing parameters (PR #5061)


oschuett left a comment (cp2k/cp2k#5061)
 
Since this is a new feature, could you please add a test for it.

garrettwx:
@Growl1234 Hi, these days I have carefully checked the magnetic system calculation. I just refer to the work of CoN4 with axial Cl, where Co is the magnetic center. No doubt that the standard diag presents the hard convergence or no-convergence even switch on RELAX_MULTIPLICITY. 

Then I apply OT, still the very fast convergence by OT, and the restart from OT to diag with k-points is OK. 

But, if restart from OT into diag with different multiplicity, hard convergence and even no convergence < 600 cycles again. 

I may guess possible resolutions :1. necessary OT with k-points, 2. the combination between OT (first) and then the instant restart from OT with .wfn into diag with k-points. 

Since now Prof. Hutter states the big difficulty in OT with k-points, perhaps let cp2k itself performs the instant restart from OT into diag with k-points in one work flow or task will be more feasible currently ?

Growl1234:
@garrettwx

Actually, I'm currently looking into ABACUS (https://github.com/deepmodeling/abacus-develop), a program that features LCAO (Linear Combination of Atomic Orbitals; here based on numerical atomic orbitals) as one of its main approaches, and considering whether there are any specific algorithmic ideas that could be borrowed for CP2K.

I still maintain that the standard diagonalization itself is not the root of the problem (although an alternative approach like Block-Davidson can indeed be a better choice). Also, please note that OT is not compatible with smearing.

yisichi-design:
The thought of so many new features being added just gets me so excited—I can’t wait to try the new version of CP2K. Thank you to the developers for your hard work.

PR #5063, opened 2026-04-14
Reference benchmark 

juerghutter:
This MGGA benchmark includes basis functions up to l=4
resulting  in lp=10 for collocate and lp=11 for integrate/forces


oschuett:
Thanks a lot for this additional benchmark! It will be super valuable when optimizing the grid kernels. 
However, I'm hesitate to add it to [our daily benchmarks](https://dashboard.cp2k.org/archive/perf-cuda-volta/index.html) because it would take ~30 minutes on the V100.

PR #5071, opened 2026-04-17
Revert "Initialize matrices before parallel_gemm in make_basis_lowdin"

Growl1234:
Reverts cp2k/cp2k#5044


oschuett:
I thought this was a good defense in depth. Why do you want to revert it?

Issue #2554, opened 2023-02-06
PW_DFT/CONTROL keywords


hmhoseini:
It seems the following keywords in PW_DFT/CONTROL section have no effect on what is printed in the output: 
PRINT_FORCES
PRINT_STRESS
PRINT_TIMERS

In addition, keyword OUTPUT is not recognized in CONTROL section.

oschuett:
CC: @mtaillefumier

mtaillefumier:
they have been removed from the list of available options as they are only intended for debugging purpose. Forces and stress are not automatically calculated and an explicit call the cp2k dedicated functions should be made to retrieve the values.

oschuett:
Thanks! Indeed, these keywords no longer exist in the [FORCE_EVAL/PW_DFT/PRINT](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/PW_DFT/PRINT.html) section.

hmhoseini:
The keywords are in the PW_DFT/CONTROL section.

oschuett:
My bad. I guess, we should remove those keywords?

he-zilong:
This may or may not be relevant, but whose idea was it to make the entire [`&FORCE_EVAL/&PW_DFT`](https://manual.cp2k.org/cp2k-2026_1-branch/CP2K_INPUT/FORCE_EVAL/PW_DFT.html) section in the manual visible **only if** CP2K that provides material for manual has been built with `SIRIUS`??? As can be seen in the link to a release version of the online documentation, this mechanism makes a vacuous and useless page:

<img width="754" height="323" alt="Image" src="https://github.com/user-attachments/assets/72074913-f3bb-4d53-ade9-2659d95f7e58" />

This does NOT happen to other input options related to optional dependency packages, and is NOT user-friendly as it forces any curious user either to complete the non-trivial task of generating a local copy of manual or to comprehend the fortran source code.
https://github.com/cp2k/cp2k/blob/d5b0302fb90d77b374de586d6a752f98ce081e9e/src/input_cp2k_pwdft.F#L484-L502

I am not suggesting that the procedure for online manual generation should enable `SIRIUS` build by default (which is impractical given the complicated dependency chain of `SIRIUS`). My opinion is that the entirety of sections and keywords should _unconditionally_ be available (with some notes like "this section will only be parsed and become effective with this-and-this package enabled when building CP2K").

oschuett:
[...]

PR #5074, opened 2026-04-18
Docker: Switch manual to pdbg

oschuett:
Follow up to #4646, see also https://github.com/cp2k/cp2k/issues/2554#issuecomment-4272687540.


oschuett:
The `PW_DFT` section is [back](https://ci.cp2k.org/artifacts/run-cp2k-manual-3097eaae/manual/CP2K_INPUT/FORCE_EVAL/PW_DFT.html).

PR #5075, opened 2026-04-18
NEGF: xTB added

dryndyk:


he-zilong:
Just a little curious, does `xTB` mean CP2K native implementation only, or does it also encompass `tblite` interface (which has GFN2-xTB support)?

dryndyk:
> Just a little curious, does `xTB` mean CP2K native implementation only, or does it also encompass `tblite` interface (which has GFN2-xTB support)?

Works also for tblite, but tblite doesn't work with OT for some reasons and it is not good for big systems.

mkrack:
Is there a test input for NEGF with xTB?

dryndyk:
I will add one today.

mkrack:
Thanks for adding a test input. Are the files `tests/QS/regtest-negf-xtb-1/H28_k-device-0-1.dos‎` and `tests/QS/regtest-negf-xtb-1/H28_k-transm-0-1.transm‎` are really needed or did you commit accidentally some output files?

dryndyk:
Output files. To compare.

dryndyk:
I will make a new suite of test files for NEGF  soon. It takes some time.

mkrack:
I don't think the current test script supports a comparison to results in output files other than the CP2K standard output file. Therefore, I suggest to write key results to the standard output file and remove the other output files.

dryndyk:
OK, besides, the test is too slow.

dryndyk:
ready

fstein93:
There is problem with this PR as some tests are failing. Could you have a look at it ASAP?

dryndyk:
Yes, I have already tried to reproduce the problem. Unfortunately, I can't install cp2k.pdbg based on the toolkit (suddenly). I try spack now. To proceed, you can temporarily disable this test.

PR #3501, opened 2024-06-25
adding optional D4 support

jpoto:
This pull request contains the optional implementation of D4 from Grimme

For now it is deactivated per default. 

@oschuett 
Can you check the build system? 
I set it up that it downloads ninja and the D4 package from the public repository, not a copy in https://www.cp2k.org/static/downloads

@fstein93 
Maybe you can have a look?

PR #5080, opened 2026-04-20
Toolchain: Replace "scripts/generate_cmake_options.sh" with "build_cp2k.sh"

Growl1234:
Merge the behaviors of generating CMake options and building CP2K into one script so that there's no need for users to run several cmake steps manually.

The environment variables of CP2K itself are written into `cp2k/install/cp2k_env` rather `cp2k/toolchain/install/setup` to prevent probable duplication of CP2K environment variables caused by running `build_cp2k.sh` a second time. The command `source cp2k/toolchain/install/setup` is also written into  `cp2k/install/cp2k_env`, so users only need to source the `cp2k_env` file before running CP2K.

Also enables building static library of CP2K through `./build_cp2k.sh --build-static` which passes `-DBUILD_SHARED_LIBS=OFF` to CMake.


tools/toolchain/scripts/generate_cmake_options.sh
```diff
@@ -1,240 +0,0 @@
...
-
-### Other optional behaviors
-
-Both for toolchain buildings and CP2K building: once build is completed, the \`build\`
-directory can be safely deleted. However, you **MUST** keep the \`install\` directory
-as is.
-
-Especially, if you want to save disk space but at the same time keep the cached CMake
-files, you can run \`cmake --build build --target clean\` alternatively after installing
-successfully.
```
he-zilong, line 236: I think this part about cleaning up disk space should be preserved. Have you considered it as an option for the new `build_cp2k.sh` script?

tools/toolchain/scripts/generate_cmake_options.sh
```diff
@@ -1,240 +0,0 @@
...
-[ "${BASH_SOURCE[0]}" ] && SCRIPT_NAME="${BASH_SOURCE[0]}" || SCRIPT_NAME=$0
-SCRIPT_DIR="$(cd "$(dirname "$SCRIPT_NAME")" && pwd -P)"
-
-source "${SCRIPT_DIR}"/common_vars.sh
-source "${SCRIPT_DIR}"/tool_kit.sh
-source "${SCRIPT_DIR}"/signal_trap.sh
-source "${INSTALLDIR}"/toolchain.conf
-source "${INSTALLDIR}"/toolchain.env
-
-# This script assumes a working environment as follows:
```
he-zilong, line 15: I think the environment tree should be preserved as a comment in the new `build_cp2k.sh` script, so that anyone looking at the script can immediately get the idea that the script process involves upper-level directories.

tools/toolchain/build_cp2k.sh
```diff
@@ -0,0 +1,213 @@
+#!/bin/bash -e
+
+# shellcheck disable=all
+
+[ "${BASH_SOURCE[0]}" ] && SCRIPT_NAME="${BASH_SOURCE[0]}" || SCRIPT_NAME=$0
+TOOLCHAIN_ROOTDIR="${PWD}"
```
he-zilong, line 8: Due to the usage of `CP2K_ROOT=$(cd "${TOOLCHAIN_ROOTDIR}/../.." && pwd)`, an explicit check of the working directory is necessary like the main toolchain script so that it will not be executed elsewhere:
```suggestion
SCRIPT_DIR="$(cd "$(dirname "$SCRIPT_NAME")" && pwd -P)"

TOOLCHAIN_ROOTDIR="${PWD}"
# ------------------------------------------------------------------------
# Exit this script if it is not called from the ./tools/toolchain directory
# ------------------------------------------------------------------------
if [ "${TOOLCHAIN_ROOTDIR}" != "${SCRIPT_DIR}" ]; then
  cat << EOF
ERROR: (${SCRIPT_NAME}, line ${LINENO}) Incorrect execution location.
The absolute path of the build_cp2k.sh script is at:
  ${SCRIPT_DIR}
Actual working directory where it is currently called:
  ${TOOLCHAIN_ROOTDIR}
Please enter the absolute path above before executing the build_cp2k.sh script
so that subsequent scripts can be found and files can be placed correctly.
EOF
  exit 1
fi
```

tools/toolchain/build_cp2k.sh
```diff
@@ -0,0 +1,213 @@
...
+    CMAKE_OPTIONS+=" -DCP2K_USE_MPI=ON"
+    if [ -n "$(grep "MPI_F08" "${TOOLCHAIN_INSTALL_DIR}/toolchain.env")" ]; then
+      CMAKE_OPTIONS+=" -DCP2K_USE_MPI_F08=ON"
+    fi
+  fi
+  # If GPU acceleration is used, add the option about GPU acceleration
+  if [ "${ENABLE_CUDA}" = "__TRUE__" ]; then
+    CMAKE_OPTIONS+=" -DCP2K_USE_ACCEL=CUDA -DCP2K_WITH_GPU=${GPUVER}"
+  elif [ "${ENABLE_HIP}" = "__TRUE__" ]; then
+    CMAKE_OPTIONS+=" -DCP2K_USE_ACCEL=HIP -DCP2K_WITH_GPU=${GPUVER}"
```
he-zilong: Since `build_cp2k.sh` is run independently outside of `install_cp2k_toolchain.sh`, please check if intermediate variables determined in `install_cp2k_toolchain.sh` like `${GPUVER}` here are also available by sourcing the `${TOOLCHAIN_INSTALL_DIR}/setup` file.

he-zilong:
Apart from the suggestions above, it would be nice if this `build_cp2k.sh` is mentioned in `INSTALL.md` and/or `docs/getting-started/build-from-source.md`.

Growl1234:
> Apart from the suggestions above, it would be nice if this `build_cp2k.sh` is mentioned in `INSTALL.md` and/or `docs/getting-started/build-from-source.md`.

The documentation is updated. However, I have not seen the "suggestions above"; maybe you forgot to submit your review comments?

Growl1234:
It is also available now to install the toolchain and CP2K itself outside the source tree. The building steps are still required to be under the source tree as before.

Users can run:
```bash
./install_cp2k_toolchain.sh --install-dir=/opt/cp2k/toolchain
./build_cp2k.sh --prefix /opt/cp2k
```
And here's the result:
```bash
root@linux:/opt/cp2k# tree -L 2 -F
./
├── bin/
├── cp2k_env
├── include/
├── lib/
├── share/
└── toolchain/
    ...
    ├── COSMA-2.8.4/
    ...
    ├── setup
    ├── tool_kit.sh
    ...
```

he-zilong (commented):
>  However, I have not seen the "suggestions above"; maybe you forgot to submit your review comments?

Oops, my bad, submitted.

tools/toolchain/scripts/generate_cmake_options.sh
Growl1234: I didn't do it, but I can add it to the hints printed at the end.

tools/toolchain/build_cp2k.sh
Growl1234: Oops, that was my fault, and it seems actually the problem is not only here. I'll have a further look later.

tools/toolchain/scripts/generate_cmake_options.sh
Growl1234: Added.

tools/toolchain/build_cp2k.sh
Growl1234: Added.

tools/toolchain/build_cp2k.sh
Growl1234: I'm planning to add these variables to `toolchain.conf` and adjust above behaviors slightly. I didn't do test in the condition without MPI (original logic `if [ ${MPI_MODE} != "no" ]` will be treated as true no matter if MPI is configred), so I ignored the problem by mistake.

Growl1234: This should be resolved. Please have a look.

he-zilong:
I doubt if the wording about how the "toolchain is deprecated and will possibly be removed or slimmed down in a future release" is suitable for inclusion in `build-from-source.md‎` at this moment, as it is still a matter under discussion in #4843 for which we have _not_ collected sufficiently many user opinions.

Some remarks can be made about how the toolchain does not cover all optional advanced utility libraries (in particular GPU-oriented ones), how Spack is preferred over toolchain for managing environment and dependencies, and how the future of toolchain maintenance is, broadly speaking, under discussion. But these are all different from outright stating that toolchain is deprecated.

he-zilong:
As for my original reviews, the environment variables look good. But I still could not find the note about clean up disk space in the build directory. (Edit: sorry, just saw it. On that matter, I wondered how many files `cmake --target clean` can delete compared with brute-force `rm -rf ./build`...)

Growl1234:
> On that matter, I wondered how many files `cmake --target clean` can delete compared with brute-force `rm -rf ./build`...

With `cmake --target clean` rather than directly deleting `build` directory, the cached CMake files are reserved, and then when you want to rebuild CP2K without willing to change CMake configuration, simply go to the `build` directory and run `make install`. Both approaches are practically the same for saving space, as most of the space is taken up by generated code and object files anyways.

P.S. The `build_cp2k.sh` will force an almost entire rebuild anyway because the CMake step is performed before building CP2K. The `--clean` flag is to avoid a case that you drop off an option because the cached CMake files still assume it is valid.

oschuett:
This might be too much for this PR, but is there a chance we could eventually replace [cmake/cmake_cp2k.sh](https://github.com/cp2k/cp2k/blob/master/cmake/cmake_cp2k.sh) with this new `toolchain/build_cp2k.sh` script?

Growl1234:
I think there is, and it will be a chance to further test the current logic of generating CMake options from toolchain configuration. But this will require a bit further modification to make it fit for various types of testers; moreover Ninja generator was not considered in this script (actually I'm thinking about dropping Ninja from toolchain because it was once added only to build dftd4 package with `-GNinja`).

oschuett:
> I'm thinking about dropping Ninja from toolchain ...

I believe we started using Ninja because of [this issue](https://gitlab.kitware.com/cmake/cmake/issues/18188). I'm not sure if it's still relevant though.

Growl1234:
> > I'm thinking about dropping Ninja from toolchain ...
> 
> I believe we started using Ninja because of [this issue](https://gitlab.kitware.com/cmake/cmake/issues/18188). I'm not sure if it's still relevant though.

Ninja installation in toolchain is introduced through #3501.

Growl1234:
I've had a bit deeper look at those issues related to CMake generators, and perhaps we can give a proper hint in documentation (though I didn't encounter problem building with Makefile generated by CMake). I'm fine with keeping Ninja in the toolchain; that said, I'm still curios about why Ninja became a toolchain dependency together with and only for dftd4.

As of replacing `cmake/cmake_cp2k.sh`, there are some more things to do (as said above), which I believe is beyond the scope of this current PR but I will take care of in the future.

oschuett:
> (though I didn't encounter problem building with Makefile generated by CMake). 

Maybe CMake fixed the problem in the meantime and simply forgot to close [the issue](https://gitlab.kitware.com/cmake/cmake/-/work_items/18188). I think, we can take the risk and drop Ninja from the toolchain to see what happens.

> I'm still curios about why Ninja became a toolchain dependency together with and only for dftd4.

Don't ponder this too long. Some things just get lost to history ;-)

Growl1234:
I would like to suggest merging this PR if there's no other problems, for convenience to further deal with things with CI testers (as mentioned above) and other toolchain dependencies like SIRIUS (in #5098).

PR #5084, opened 2026-04-20
Skala functional via GauXC for XC energy evaluation using deep learning models

franzpoeschel:
This adds [GauXC](https://microsoft.github.io/skala/gauxc/fortran-library.html), which describes itself as `a modern, modular C++ library for the evaluation of quantities related to the exchange-correlation (XC) and exact-exchange (K) energy (e.g. potential, etc) in the Gaussian basis set discretization of Kohn-Sham density function theory (KS-DFT)` for evaluating exchange correlation energies. The main interest here is to integrate the Skala deep learning model for XC functionals into CP2K.

GauXC needs to be built from [this branch](https://github.com/wavefunction91/GauXC/tree/skala) for 1. the Skala functional and 2. Fortran bindings to be available.

The code logically differs a bit from the other existing XC functionals as GauXC's interface is more high-level with automated build of the Grid.

Status:

- [ ] Wait for https://github.com/wavefunction91/GauXC/pull/202, then remove the patch from the toolchain builder
- [x] MPI awareness
- [ ] CI integration
- [x] Rename "SKALA" keyword to "GAUXC" in the input file, GauXC is not restricted to only Skala
- [x] Remove the C binding, I couldnt get this to work in Fortran and then didnt bother further for the time being

Deferred to a later PR:

- [ ] k-point integration support (currently fails for `nimages > 1`)
- [ ] Optimization: Can initialized data structures of GauXC be kept alive across calls?

Things to consider in a review:

- [x] Correctness: The simulation converges, but comparing GauXC's PBE functional with CP2K's PBE functional currently gives different results by a factor of roughly 2.
- [x] Correctness: Do the bindings in `src/xc/xc_gauxc_interface.F` correctly transform between CP2K data structures and GauXC datatypes (consider also units?)
- [x] Correctness: Do the conversions from sparse DBCSR matrixes to regular n-dim Fortran arrays and back make sense?
- [x] Correctness: Does the integration of GauXC into `qs_ks_build_kohn_sham_matrix` make sense?

Generally, as I am new to CP2K, some verification of how much these things make sense would be appreciated.

Code structure:

* `src/xc/xc_gauxc_interface.F` replicates the GauXC interface for CP2K types and data structures
* `src/xc/xc_gauxc_functional.F` parses the data from the input file and applies the GauXC functional
* `src/qs_ks_methods.F`: `qs_ks_build_kohn_sham_matrix` calls the GauXC functional. It's rather difficult to understand the different code paths that this subroutine can take, so this should be verified by someone who knows what is going on in there.
* Other edited files: Mostly boilerplate


franzpoeschel:
~~Currently requires running GauXC from [this branch](https://github.com/franzpoeschel/GauXC/tree/cp2k-rks-density-fix) to avoid double-counting of the occupation factor in the RKS setup~~

awvwgk:
> Currently requires running GauXC from [this branch](https://github.com/franzpoeschel/GauXC/tree/cp2k-rks-density-fix) to avoid double-counting of the occupation factor in the RKS setup

Just checked https://github.com/franzpoeschel/GauXC/commit/328064fc21a795c86d9234f095e7c8cfa6fbd739. I believe it would be better to do the rescaling on the CP2K side rather than requiring a patch to GauXC. Alternatively, you can always call the unrestricted path and pass the spin-polarization density as a zero matrix.

DCM-Uni-Paderborn:
Lieber Franz,

many thanks for your PR and for continuing the GauXC/SKALA integration. The earlier patch note from DCM-Uni-Paderborn is outdated; please use the current links below instead.

The updated patch files are available from the DCM-Uni-Paderborn fork:

- Review-fixes patch, meant to be applied on top of the current PR branch: https://raw.githubusercontent.com/DCM-Uni-Paderborn/cp2k/patches/pr-5084-gauxc/pr-5084-gauxc-review-fixes.patch
- Combined patch against current CP2K master: https://raw.githubusercontent.com/DCM-Uni-Paderborn/cp2k/patches/pr-5084-gauxc/pr-5084-gauxc-combined-master.patch
- Corresponding work branch: https://github.com/DCM-Uni-Paderborn/cp2k/tree/pr-5084-gauxc-skala-mpi

The review-fixes patch contains the fixes on top of this PR: MPI-aware GauXC/SKALA force handling, the CP2K-side density/VXC handling fixes, GAPW_XC PBE handling, OneDFT/SKALA grid defaults, model-key handling, extra regression coverage for RKS/UKS/GAPW/f-shells/finite-difference forces, and explicit aborts for response/kernel paths requiring higher XC derivatives.

The combined-master patch contains the full GauXC/SKALA integration plus those fixes, for anyone starting from current CP2K master instead of this PR branch.

I checked that both patches apply cleanly to their intended bases. The current work branch builds, passes `make_pretty.sh`, and the `QS/regtest-gauxc` subset passes with 1 MPI rank and with 2 MPI ranks on my local setup. OpenMP with multiple threads is not claimed as validated from my local macOS build, because the native CP2K reference path also fails there independently of GauXC.

Current functional scope: GauXC/SKALA covers energy, VXC, and nuclear gradients for the tested QS cases. Response/kernel properties needing higher XC derivatives are not implemented yet and now abort explicitly instead of entering a partially meaningful path.

franzpoeschel:
@DCM-Uni-Paderborn Thank you for investigating the remaining issues. I have added your patch to this branch now, and also updated the PR to the latest master commit.
As suggested above by @awvwgk, I have reapplied the xmat-factor fix into CP2K, so the normal `skala` branch of GauXC yields correct results now, and not my patched version that I linked above.

With the minimal H2 example working correctly now, the next goal is to bundle this PR into a minimal reference implementation of GauXC in CP2K. Further things such as sparsity or MPI awareness can come at a later point. 
So my most immediate task is to continue cleanup things up.

Question to @awvwgk: For CI integration, the usual path would be to add GauXC as a Spack dependency. However, this PR currently works only for the WIP `skala` branch. Are Spack releases planned for GauXC? Are there plans for a release with a minimal working subset of the Fortran bindings available?

awvwgk:
> Question to @awvwgk: For CI integration, the usual path would be to add GauXC as a Spack dependency. However, this PR currently works only for the WIP `skala` branch. Are Spack releases planned for GauXC? Are there plans for a release with a minimal working subset of the Fortran bindings available?

See https://github.com/spack/spack-packages/pull/4814

tools/toolchain/scripts/stage6/install_stage6.sh
Growl1234, line 1: Why `install_gauxc.sh` is placed in `stage3` directory but is installed at stage 6?

DCM-Uni-Paderborn: Good catch, thanks. GauXC depends on libtorch and is installed from stage 6, so I moved `install_gauxc.sh` and the GauXC patch file into `tools/toolchain/scripts/stage6` and updated `install_stage6.sh` accordingly. The refreshed DCM patches include this cleanup.

DCM-Uni-Paderborn (approved):
I will merge it now, so that future changes can add to this effort by separate PRs. Well done!

franzpoeschel:
> I will merge it now, so that future changes can add to this effort by separate PRs. Well done!

Nice, thank you!

The CI currently installs GauXC, but does not yet actually compile CP2K against it yet, so I will open a follow-up PR to fix that bit

Growl1234:
> The CI currently installs GauXC, but does not yet actually compile CP2K against it yet, so I will open a follow-up PR to fix that bit

This should be resolved as part of #5242 via making CMake option `CP2K_USE_GAUXC` follow the setting of `CP2K_USE_EVERYTHING` by default.

DCM-Uni-Paderborn:
Dear SY Wang,

Ciao Franz,

thanks for the follow-up discussion on the GauXC CI/toolchain side. I opened #5244 as the separate follow-up after this PR was merged. It includes the stage6 GauXC toolchain move from #5242, makes `CP2K_USE_GAUXC` follow `CP2K_USE_EVERYTHING`, updates the GauXC/SKALA toolchain state after wavefunction91/GauXC#202, and keeps the remaining CP2K-side changes independent of the draft status of #5242.

One important detail: the conventional GauXC RKS PBE path still needs the CP2K/GauXC density-convention issue handled consistently. CP2K passes a spin-summed RKS density matrix, while the GauXC RKS path applies its own occupation factor when building the X matrix. The follow-up keeps this aligned and I locally checked `QS/regtest-gauxc` with 1 and 2 MPI ranks; both runs were 9/9 correct.

So #5242 does not have to be merged first. If #5244 is merged first, the main stage6/CMake part of #5242 is effectively covered there; if #5242 is merged first, #5244 can simply be rebased.

Greeting,
Thomas

PR #5088, opened 2026-04-22
Test-Farm: Revert to Lmod environment

Growl1234:


Growl1234:
The test-farm failure of "full" test is with 2 "wrong results", and there are several other testers on dashboard encountering the same problem. This PR should be safe to merge in my view.

Issue #4962, opened 2026-03-16
RFC: Make extended XYZ format the default


oschuett:
The [extended XYZ format](https://www.ovito.org/manual/reference/file_formats/input/xyz.html) stores the cell information alongside the atom positions. This allows for more streamlined and robust workflows. 

Since we now have support for [writing](https://github.com/cp2k/cp2k/pull/4601) and [reading](https://github.com/cp2k/cp2k/pull/4960) extXYZ , I think we should make it the default [FORMAT](https://manual.cp2k.org/trunk/CP2K_INPUT/MOTION/PRINT/TRAJECTORY.html#CP2K_INPUT.MOTION.PRINT.TRAJECTORY.FORMAT).

Furthermore, if the cell is not provided explicitly, then we should start to fall back to reading it from the [coordinate file](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/SUBSYS/TOPOLOGY.html#CP2K_INPUT.FORCE_EVAL.SUBSYS.TOPOLOGY.COORD_FILE_NAME).

While these changes are mostly backwards compatible, they will affect every CP2K run. Hence, I wanted to briefly discuss them upfront.

he-zilong:
> Furthermore, if the cell is not provided explicitly, then we should start to fall back to reading it from the [coordinate file](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/SUBSYS/TOPOLOGY.html#CP2K_INPUT.FORCE_EVAL.SUBSYS.TOPOLOGY.COORD_FILE_NAME).

I can take care of this now, but does it also mean that parsers for cell information in `G96` and `XTL` are desirable?

> Since we now have support for [writing](https://github.com/cp2k/cp2k/pull/4601) and [reading](https://github.com/cp2k/cp2k/pull/4960) extXYZ , I think we should make it the default [FORMAT](https://manual.cp2k.org/trunk/CP2K_INPUT/MOTION/PRINT/TRAJECTORY.html#CP2K_INPUT.MOTION.PRINT.TRAJECTORY.FORMAT).

I am in favor of this, though that also depends on how many MD post-processing programs can read cell information dynamically from EXTXYZ (especially trajectory from NPT simulations).

Growl1234:
I share the same concern as @he-zilong; the main obstacle is that extxyz is not yet very widely supported (as far as I know, VMD doesn't have extended support for extxyz, though OVITO and ASE may do).

[...]

Issue #4349, opened 2025-08-04
Feature request: write simulation box sizes into XYZ trajectory files


cihe13375:
If I understand it correctly, currently CP2K only supports DCD, PDB, and XYZ formats for trajectory output (https://manual.cp2k.org/trunk/CP2K_INPUT/MOTION/PRINT/TRAJECTORY.html#CP2K_INPUT.MOTION.PRINT.TRAJECTORY.FORMAT). DCD is a binary format (which means it cannot be easily parsed without dependencies), while PDB is not really a thing for non-biological simulations, so I often find myself using the XYZ format. However, currently the XYZ trajectory does not include the simulation box size information.

I know that the original XYZ format is not designed to record box sizes, but several extensions to include it are available (e.g. https://github.com/libAtoms/extxyz) so I hope that CP2K could support one of them. For the extxyz format linked before it's essentially adding the cell parameters into the comment line (e.g. `Lattice="10.0 0.0 0.0 0.0 10.0 0.0 0.0 0.0 10.0" Properties=species:S:1:pos:R:3`), which is not a difficult task I think.

mkrack:
Maybe the following information about existing tools provided with CP2K could be helpful for you. The trajectory files in `dcd` format printed by CP2K can be dumped in human readable form with `dumpdcd`. A `dumpdcd.psmp` or `dumpdcd.ssmp` binary is created during CP2K's installation in the same folder as the CP2K binary. `dumpdcd` provides a variety of options:
```
  -cell_file, -cell              : Input file with cell information in CP2K format (.cell)
  -debug, -d                     : Print debug information
  -ekin                          : Dump just the "temperature" of each atom
  -eformat                       : Print coordinates in scientific format
  -eo                            : Write standard output and standard error to the same logical unit
  -first_frame, -ff <int>        : Number of the first frame which is dumped
  -help, -h                      : Print this information
  -ignore_warnings               : Do not print warning messages, e.g. about inconsistencies
  -info, -i                      : Print additional information for each frame (see also -debug flag)
  -last_frame, -lf <int>         : Number of the last frame which is dumped
  -md_time_step <real>           : Use this MD time step instead of the one from the DCD header
  -output, -o <file_name>        : Name of the output file (default is stdout)
  -output_format, -of <DCD|XMOL> : Output format for dump
  -pbc                           : Apply the periodic boundary conditions (PBC) to each frame before it is dumped
                                   (origin at lower left)
[...]

PR #5089, opened 2026-04-23
Increase precision of cell info in extended XYZ format

oschuett:
Follow up to #4349 and #4601. 

@he-zilong, thanks for the hint https://github.com/cp2k/cp2k/issues/4962#issuecomment-4297965575.

### DIFF
--- a/src/motion_utils.F
+++ b/src/motion_utils.F
@@ -413,7 +413,7 @@ CONTAINS
             END IF
          CASE (dump_extxyz)
             CALL section_vals_val_get(root_section, "MOTION%PRINT%TRAJECTORY%PRINT_ATOM_KIND", l_val=print_kind)
-            WRITE (UNIT=cell_str, FMT="(9(1X,F10.3))") cell%hmat(:, 1)*angstrom, cell%hmat(:, 2)*angstrom, cell%hmat(:, 3)*angstrom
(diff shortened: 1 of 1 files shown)


mkrack:
Why not using a scientific format like `ES19.12` to keep more significant digits which would possibly allow for a shorter string while keeping the accuracy? Do you expect parsing issues for post-processing tools?

he-zilong:
To be clear, `ES19.12` is totally okay for me (my wording was "at least" in that comment). Please make sure that the format of `.cell` files is consistent with whatever the choice is here.

oschuett:
> Why not using a scientific format like ES19.12 to keep more significant digits which would possibly allow for a shorter string while keeping the accuracy?

Since the cell is only written once per frame, I don't think those few extra bytes really matter.

According to [these docs](https://github.com/libAtoms/extxyz#floating-point-number) scientific notation should be ok, but none of the examples I saw used it. So, I'd rather stick with the safe option. Furthermore, we're already using `F19.10` throughout [motion_utils.F](https://github.com/cp2k/cp2k/blob/master/src/motion_utils.F).

he-zilong:
By the way I propose making another little change here: the double quotes surrounding `Properties` string should probably be removed, e.g. `Properties=species:S:1:pos:R:3`, not `Properties="species:S:1:pos:R:3"`, because no other example I could find has the double quotes and OVITO (3.15.3) could not recognize the latter form which is generated from CP2K.

Issue #5055, opened 2026-04-12
Spack docker tests slow with `--network=host` flag


Growl1234:
When running `./spack_cache_start.sh; podman build --network=host --shm-size=1g -f ./Dockerfile.test_spack_psmp-gcc15 ../../` on my local machine, the regtest step is very slow with CPU usage near 0. 

With a detailed look via `podman run -it`, it seems that has to do with the flag `--network=host`; and after removing that option in `podman build` command, everything works fine. However, without this option, it seems I would have to rebuild all Spack dependencies at every run. Therefore, I wish devs can have a further look on this behavior.

I've also noted that Spack-pdbg test on the CP2K-CI also runs very slowly. Would it be possible that this also has to do with that `--network=host` option?

oschuett:
Hmm, this is strange. It should actually be slightly faster to start a container with `--network=host` that the default where podman has to create a virtual network. In any case, this should only matter when the container is created.

One thing you could try is to run the regtests with [`--keepalive`](https://github.com/cp2k/cp2k/blob/5fb5ea801f6f821b8fdae209002bda40e74ae568/tests/do_regtest.py#L76) to reduce the number of spawned child processes.

Another hack you could use is to interrupt the `podman build` after the Spack installation and then restart it without `--network=host`. With any luck podman's layer-cache will just continue where it left off.

mkrack:
> With a detailed look via `podman run -it`, it seems that has to do with the flag `--network=host`; and after removing that option in `podman build` command, everything works fine. However, without this option, it seems I would have to rebuild all Spack dependencies at every run. Therefore, I wish devs can have a further look on this behavior.

Without the flag `--network=host`, the spack cache `s3://spack-cache --s3-endpoint-url=http://localhost:9000` is not reachable during the `podman build` run and thus the build will start from scratch and build all dependencies each time.
 
> I've also noted that Spack-pdbg test on the CP2K-CI also runs very slowly. Would it be possible that this also has to do with that `--network=host` option?

I don't observe any significant impact on the timing running with and without `--network=host` locally on my desktop PC. Like for #5090, the slow-down occurs only with CI cloud tester.

Growl1234:
I’ve also noticed that it serves to store the Spack cache, but … it’s quite strange. This issue does exist locally, and I currently have no idea indeed what is causing it.

DCM-Uni-Paderborn:
[...]

PR #5090, opened 2026-04-23
Drop keepalive for spack dbg tester

mkrack:
The LeakSanitizer (LSAN) does not work properly in that mode

### DIFF
--- a/make_cp2k.sh
+++ b/make_cp2k.sh
@@ -1457,7 +1457,8 @@ export LAUNCH_SCRIPT
 cat << *** > "${LAUNCH_SCRIPT}"
 #!/bin/bash
 ulimit -c 0 -s unlimited
-export LSAN_OPTIONS=suppressions=${INSTALL_PREFIX}/bin/lsan.supp
+export ASAN_OPTIONS="detect_leaks=1"
+export LSAN_OPTIONS="suppressions=${INSTALL_PREFIX}/bin/lsan.supp"
 export PATH=${INSTALL_PREFIX}/bin:${PATH}
 export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}
 export OMP_NUM_THREADS=\${OMP_NUM_THREADS:-2}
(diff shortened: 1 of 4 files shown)


Growl1234:
Hi, I wonder is there any other way to make Spack pdbg tester not that slow?

he-zilong:
Apologies for the failing regtests `quartz-kv.inp` introduced by me; I want to look into it in #5091 .

he-zilong:
Does the original purpose of this PR have anything to do with the problems in [some comments in #4843](https://github.com/cp2k/cp2k/issues/4843#issuecomment-4304791289)?

mkrack:
> Does the original purpose of this PR have anything to do with the problems in [some comments in #4843](https://github.com/cp2k/cp2k/issues/4843#issuecomment-4304791289)?

I don't think so, but the issue #4843 is quite long and I might have overlooked something. This PR reverts partially #5066, i.e. only for the testers `spack pdbg` and `spack sdbg` because the `keepalive` mode combined with the LeakSanitizer (LSAN) can cause random instabilities during runtime. In #4843 build attempts with spack were not successful and thus the runtime errors discussed in that issue refer most likely to builds performed with the `toolchain`.

mkrack:
> Hi, I wonder is there any other way to make Spack pdbg tester not that slow?

The CI cloud tester for `spack pdbg` runs is significantly slower than local execution. I do not observe this behavior locally. For example, in the CI environment, each test case in ATOM/regtest-1 (which consists of small, typically fast-running tests) takes approximately 7–8 seconds to complete. In contrast, running the same `spack pdbg` tests on my desktop PC takes about 0.5 seconds per test in `ATOM/regtest-1`. This substantial performance gap strongly suggests that the extended runtimes are caused by the CI setup or the environment used for the cloud-based tests.

oschuett:
> The CI cloud tester for spack pdbg runs is significantly slower than local execution.

Maybe this is another occurrence of #5055?

> In contrast, running the same spack pdbg tests on my desktop PC takes about 0.5 seconds per test ...

Do you also run them in a container, ie. `podman build -f Dockerfile.test_spack_pdbg...`? 

For comparison, how fast are those tests with the [toolchain pdbg](https://dashboard.cp2k.org/archive/current-pdbg/index.html) on your machine?

mkrack:
Rebase and merge inspite of the failing test `quartz-kv.inp`) which is not related to this PR (see #5091).

oschuett:
If this slowdown really only occurs on the CI, then I'll do an upgrade of all its components next week.

mkrack:
> Do you also run them in a container, ie. `podman build -f Dockerfile.test_spack_pdbg...`?
> 
> For comparison, how fast are those tests with the [toolchain pdbg](https://dashboard.cp2k.org/archive/current-pdbg/index.html) on your machine?

I run the dockerfiles with podman locally as specified after `Usage:` in the header of the actual dockerfile in `tools/docker`. The locally obtained timings with `Dockerfile.test_spack_pdbg` o are very similar to the timing of the `toolchain pdbg` CI cloud tester.

Growl1234:
Unfortunately it seems I am able to reproduce the slowdown locally:

[1.txt](https://github.com/user-attachments/files/29535085/1.txt)

oschuett:
> Unfortunately it seems I am able to reproduce the slowdown locally:

On the contrary. Reproducing a problem is 90% of solving it :-)

Could you maybe open a new issue with your new findings.

Growl1234:
It seems disabling SIRIUS in Dockerfile (`-df sirius`) makes the issue disappear on my machine.

Issue #4416, opened 2025-09-01
grid_unittest fails on macOS


oschuett:
The grid unit test on macOS [has been failing](https://dashboard.cp2k.org/archive/darwin-gnu-arm64-psmp/index_full.html) ever since we upgraded GCC from 14.2.0 to 14.3.0.

In #4414 I tried to work around the issue by avoiding the exotic `vsscanf` function. Unfortunately, the test [still fails](https://dashboard.cp2k.org/archive/darwin-gnu-arm64-psmp/commit_97501a3e348e96b61bda85567e6572efcbfe2d54.txt) although we are now only relying on the much more popular `sscanf` function.

While macOS on ARM64 uses a different [calling convention](https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Update-code-that-passes-arguments-to-variadic-functions) for variadic functions, I find it hard to believe that `sscanf` is broken.

oschuett:
This problem went away after rebuilding the stack with Spack ¯\_(ツ)_/¯.

PR #5094, opened 2026-04-24
Allocate grid replay work arrays on the heap

DCM-Uni-Paderborn:
grid_replay currently allocates replay buffers as variable-length stack arrays. Some replay tasks can require sizeable buffers, and this also makes compiler behavior around large VLAs more fragile.

Move the pab/hab replay buffers to explicit heap allocations with allocation checks and cleanup.

Tested locally as part of a CP2K 2026.1 CP2K-GROMACS build on Apple Silicon; the grid_unittest passed and the full CP2K regtest suite passed: 4129 / 4129 correct.


oschuett:
This might be a fix for #4416.

PR #5099, opened 2026-04-25
grid: improve OpenMP use in task setup and memory loops

DCM-Uni-Paderborn:
This PR adds conservative OpenMP/SIMD improvements to the grid CPU, REF, and DGEMM task-list setup paths.

The changes are intentionally limited to independent writes and element-wise memory loops:

- add `GRID_PRAGMA_SIMD_LOOP` as an architecture-neutral `omp simd` helper
- add `GRID_OMP_MIN_ITERATIONS` to avoid OpenMP overhead on small loops
- parallelize CPU/REF task-array initialization
- parallelize CPU/REF first/last level-block task initialization
- add SIMD hints to CPU/REF thread-local grid reduction and final grid copy loops
- parallelize DGEMM task-structure initialization by computing `update_block_` directly from adjacent input entries instead of carrying serial loop state

The PR does not change the grid kernel math or the reduction order of the main physical accumulations. The SIMD helper deliberately does not specify a fixed `simdlen`, so the compiler can choose an appropriate vector width for the target architecture.

Testing:

- built `cp2k` with the psmp toolchain
- built `grid_unittest`
- ran `QS/regtest-grid` with `1 MPI rank x 2 OpenMP threads`: `16 / 16` correct
- ran `QS/regtest-gapw` with `1 MPI rank x 2 OpenMP threads`: `26 / 26` correct
- ran short DGEMM H2O-32 and H2O-64 comparisons with `1 MPI rank x 4 OpenMP threads`, `OPENBLAS_NUM_THREADS=1`
  - energies matched to printed precision
  - H2O-32 showed a small improvement
  - H2O-64 was within timing noise

Performance note:

The benchmark results show no regression and small improvements in some runs, but this PR should be viewed as a conservative cleanup/improvement rather than a major performance change.

### DIFF
--- a/src/grid/common/grid_common.h
+++ b/src/grid/common/grid_common.h
@@ -13,14 +13,19 @@
 #if defined(__GNUC__) && !defined(__INTEL_COMPILER) &&                         \
     !defined(__INTEL_LLVM_COMPILER) && __GNUC__ < 6
 #define GRID_PRAGMA_SIMD(OBJS, N)
+#define GRID_PRAGMA_SIMD_LOOP
 // Intel added the simd pragma with version 19.00.
 #elif defined(__INTEL_COMPILER) && __INTEL_COMPILER < 1900
 #define GRID_PRAGMA_SIMD(OBJS, N)
+#define GRID_PRAGMA_SIMD_LOOP
 // All compilers support the same syntax defined by the OpenMP standard.
 #else
 #define GRID_PRAGMA_SIMD(OBJS, N)                                              \
   _Pragma(GRID_STRINGIFY(omp simd linear OBJS simdlen(N)))
+#define GRID_PRAGMA_SIMD_LOOP _Pragma(GRID_STRINGIFY(omp simd))
 #endif
+
+#define GRID_OMP_MIN_ITERATIONS 1024
 
 // GCC added the unroll pragma with version 8 and...
 #if defined(__GNUC__) && !defined(__INTEL_COMPILER) &&                         \
--- a/src/grid/cpu/grid_cpu_task_list.c
+++ b/src/grid/cpu/grid_cpu_task_list.c
@@ -100,6 +100,7 @@ void grid_cpu_create_task_list(
   size = ntasks * sizeof(grid_cpu_task);
   task_list->tasks = malloc(size);
   assert(task_list->tasks != NULL || size == 0);
+#pragma omp parallel for schedule(static) if (ntasks > GRID_OMP_MIN_ITERATIONS)
   for (int i = 0; i < ntasks; i++) {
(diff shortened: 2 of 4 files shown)


oschuett:
I'm not sure about this one: The `REF` backend is intentionally kept simple as it's only used for validation. The `DGEMM` backend is essentially abandoned. And the `grid_cpu_create_task_list` routine is not relevant for performance.

mtaillefumier:
> I'm not sure about this one: The `REF` backend is intentionally kept simple as it's only used for validation. The `DGEMM` backend is essentially abandoned. And the `grid_cpu_create_task_list` routine is not relevant for performance.

I am not sure I understand how the `DGEMM` backend is “abandoned”. It is only used if a user explicitly selects it (GLOBAL / GRID / BACKEND DGEMM), which to me just shows that grid‑engine parameters are almost never changed from their defaults, not that the backend itself is dead. If the intention is to deprecate and remove `DGEMM`, that’s a separate discussion — but as long as it remains a documented, user‑selectable backend, calling it “abandoned” is not the right statement.

oschuett:
> I am not sure I understand how the DGEMM backend is “abandoned”

I'm sorry, I didn't mean to offend you. The DGEMM backend was a great idea that unfortunately did not pan out - as so many things in research. On the bright side, you did discover this [really neat trick](https://github.com/cp2k/cp2k/blob/1d63da52da2deadb90adee8be6bda0055cbafbd7/src/grid/cpu/grid_cpu_collint.h#L532).

mtaillefumier:
> > I am not sure I understand how the DGEMM backend is “abandoned”
> 
> I'm sorry, I didn't mean to offend you. The DGEMM backend was a great idea that unfortunately did not pan out - as so many things in research. On the bright side, you did discover this [really neat trick](https://github.com/cp2k/cp2k/blob/1d63da52da2deadb90adee8be6bda0055cbafbd7/src/grid/cpu/grid_cpu_collint.h#L532).

No offense taken; I’m just trying to be precise about backend status and design. My point about DGEMM is simple: as long as it is documented as a valid (and tested) option and shipped in releases, calling it “abandoned” is at best ambiguous for users. If the project position is that DGEMM is a failed experiment, the basis for that should be stated explicitly (e.g. benchmarks, lack of maintenance), and the backend should be clearly marked as deprecated and scheduled for removal, rather than left as a visible option and described informally as dead.

oschuett:
> If the project position is that DGEMM is a failed experiment, the basis for that should be stated explicitly (e.g. benchmarks, lack of maintenance)...

It's been a while, but IIRC the performance advantage of the DGEMM backend disappeared when you added the spherical masking. Without this masking the regtests won't pass and presumably for good reason. So I agree that removing the backend is the logical conclusion 😟.

PR #5102, opened 2026-04-26
Toolchain: Updates of FFTW and OpenBLAS

Growl1234:


oschuett:
Voilà:
- https://www.cp2k.org/static/downloads/OpenBLAS-0.3.33.tar.gz
- https://www.cp2k.org/static/downloads/fftw-3.3.11.tar.gz

PR #5104, opened 2026-04-27
Add spack OpenMPI debug tester (pdbg)

mkrack:


oschuett:
I'm really curious if this new OpenMPI tester shows the same [slowdown](https://github.com/cp2k/cp2k/pull/5090#issuecomment-4312497166) as the one with MPICH.

I'm starting to wonder if the MPI traffic might get routed thought the TCP/IP stack. That would indeed be slower within a Kubernetes cluster with its additional network virtualization.


Unfortunately, I can't readily upgrade the CI containers as [promissed](https://github.com/cp2k/cp2k/pull/5090#issuecomment-4313361190) because Nvidia dropped support for Volta with CUDA 13. This leaves us with only bad options.

mkrack:
> I'm really curious if this new OpenMPI tester shows the same https://github.com/cp2k/cp2k/pull/5090#issuecomment-4312497166 as the one with MPICH.

The OpenMPI pdbg tester appears to be equally slow with the current settings.

> Unfortunately, I can't readily upgrade the CI containers as https://github.com/cp2k/cp2k/pull/5090#issuecomment-4313361190 because Nvidia dropped support for Volta with CUDA 13. This leaves us with only bad options.

For the same reason we stay with CUDA 12.9 because newer versions [do not support](https://github.com/cp2k/cp2k/blob/master/make_cp2k.sh#L841-L847) a `sm_code` < 75.

mkrack:
The two timeouts do not occur running the Spack OpenMPI pdbg tester locally. Therefore, it might be again an issue of the CI environment.

PR #5108, opened 2026-04-27
Diagonalization: add opt-in direct generalized eigensolvers

DCM-Uni-Paderborn:
This PR adds an opt-in `DIRECT_GENERALIZED_DIAGONALIZATION` global keyword for dense generalized eigenvalue problems. When enabled, supported backends solve the generalized problem directly instead of first reducing it through a CP2K-side Cholesky transformation.

Implemented paths:

- cuSolverMP direct generalized solver for real symmetric and complex Hermitian matrices, guarded by the existing small-matrix fallback threshold (`n >= 64`).
- ScaLAPACK direct generalized solvers for real and complex matrices via `PDSYGVX` and `PZHEGVX`.
- DLAF remains compatible with `DIRECT_GENERALIZED_DIAGONALIZATION T/F`; the DLAF CMake integration is adjusted so the Fortran module include path is exported correctly.
- The default remains conservative: `DIRECT_GENERALIZED_DIAGONALIZATION F`.

The PR also improves cuSolverMP/NCCL diagnostics. On single-GPU nodes, multiple local MPI ranks cannot safely share the same visible GPU in this path; CP2K now aborts early with a clear message instead of exposing a generic NCCL invalid-usage error.

Build and toolchain updates:

- Add `GB10` / `sm_121` CUDA architecture handling.
- Allow DBCSR 2.9.1 CUDA builds with CUDA 13 deprecation warnings.
- Add temporary DBCSR GB10 parameter handling based on the closest available H100 parameters.
- Accept the modern CMake `NVHPC` compiler ID in addition to the legacy `PGI` ID.
- Add `CP2K_SCALAPACK_VENDOR=NVPL` using NVIDIA's official `nvpl::scalapack_*` and `nvpl::blacs_*` CMake targets, with automatic or explicit BLACS MPI-interface selection.
- Add toolchain support for locating system or prefix-provided cuSolverMP installations and their NCCL/CAL dependencies.

Testing:

- Built CP2K with CUDA, ScaLAPACK, cuSolverMP and NCCL on NVIDIA GB10.
- Verified `cp2k.psmp --version` reports `cusolvermp cusolvermp_nccl`.
- New cuSolverMP negative guard test: `8 MPI x 1 OMP` on one visible GPU aborts with the intended CP2K diagnostic.
- Targeted `DIRECT_GENERALIZED_DIAGONALIZATION T/F` comparisons for cuSolverMP, ScaLAPACK real, ScaLAPACK complex and DLAF paths showed matching energies to roundoff.
- DLAF CMake build with CUDA and Libint2: `QS/regtest-dlaf` and `QS/regtest-dlaf-2` passed for both `1 MPI x 8 OMP` and `8 MPI x 1 OMP`.
- NVPL ScaLAPACK CMake build against NVHPC 26.3: CMake selected `nvpl::scalapack_lp64;nvpl::blacs_lp64_openmpi4`, `ldd` confirmed `libnvpl_scalapack_lp64.so.0` and `libnvpl_blacs_lp64_openmpi4.so.0`, and `QS/regtest-scalapack-generalized` passed for both `1 MPI x 8 OMP` and `8 MPI x 1 OMP`.
- Full parallel regtest suite for the original keyword/generalized implementation after the keyword rename: `4395 / 4395 correct`, `Status: OK`, about 38 min. The later backend/build commits were covered by the targeted tests listed above.
- Prettify/precommit checks passed for the follow-up commits.

Si64 timing comparison, DZVP-GTH, 64 Si atoms, 832 x 832 dense eigensolver matrices. Times are CP2K total timer values.

Real Si64 `DIRECT_GENERALIZED_DIAGONALIZATION` timing matrix:

| backend | layout | DIRECT_GENERALIZED_DIAGONALIZATION | total energy [Ha] | total time [s] | eigensolver path |
|---|---:|---:|---:|---:|---|
| cuSolverMP | 1 MPI x 8 OMP | T | -253.460954826209274 | 9.970 | `cp_cfm_general_cusolver` |
| cuSolverMP | 1 MPI x 8 OMP | F | -253.460954826209360 | 258.785 | Cholesky fallback + standard eigensolver |
| ScaLAPACK | 1 MPI x 8 OMP | T | -253.460954826209445 | 1011.247 | `cp_cfm_geeig_scalapack` |
| ScaLAPACK | 1 MPI x 8 OMP | F | -253.460954826209445 | 985.640 | Cholesky fallback + standard eigensolver |
| ScaLAPACK | 8 MPI x 1 OMP | T | -253.460954826209388 | 3.277 | `cp_cfm_geeig_scalapack` |
| ScaLAPACK | 8 MPI x 1 OMP | F | -253.460954826209331 | 3.286 | Cholesky fallback + standard eigensolver |

Complex-k-point cases use the same Si64 cell with `MONKHORST-PACK 2 1 1` and `WAVEFUNCTIONS COMPLEX`.

Complex Si64 `DIRECT_GENERALIZED_DIAGONALIZATION` timing matrix:

| backend | layout | DIRECT_GENERALIZED_DIAGONALIZATION | total energy [Ha] | total time [s] | eigensolver path |
|---|---:|---:|---:|---:|---|
| cuSolverMP | 1 MPI x 8 OMP | T | -253.567156633445904 | 7.586 | `cp_cfm_general_cusolver` |
| cuSolverMP | 1 MPI x 8 OMP | F | -253.567156633446075 | 5.781 | Cholesky fallback + `cp_cfm_heevd_base` |
| ScaLAPACK | 1 MPI x 8 OMP | T | -253.567156633446416 | 6.019 | `cp_cfm_geeig_scalapack` |
| ScaLAPACK | 1 MPI x 8 OMP | F | -253.567156633446160 | 5.807 | Cholesky fallback + `cp_cfm_heevd_base` |
| ScaLAPACK | 8 MPI x 1 OMP | T | -253.567156633446302 | 3.981 | `cp_cfm_geeig_scalapack` |
| ScaLAPACK | 8 MPI x 1 OMP | F | -253.567156633446160 | 3.899 | Cholesky fallback + `cp_cfm_heevd_base` |

The real-path energies agree to roundoff, and the complex-path energies agree to about `1e-12 Ha` across `DIRECT T/F` and supported backends. For complex cuSolverMP, only `DIRECT T` exercises the cuSolverMP generalized path; `DIRECT F` performs the CP2K-side Cholesky fallback and then the complex standard eigensolver. ScaLAPACK benefits strongly from `8 MPI x 1 OMP`.


DCM-Uni-Paderborn:
Please manually trigger CUDA Pascal Regtest, Performance CUDA Volta, Spack (ssmp, CUDA P100) and Spack (psmp, CUDA P100), respectively.

RMeli:
Very nice! Thank you for checking DLAF as well. I can run additional tests on ours side, if needed.

DCM-Uni-Paderborn:
> Very nice! Thank you for checking DLAF as well. I can run additional tests on ours side, if needed.

Yes if possible that would be greatly appreciated since DLAF is often not explicitly tested. E.g. testing all different combinations of PREFERRED_DIAG_LIBRARY and PREFERRED_CHOLESKY_LIBRARY, with DIRECT_GENERALIZED_DIAGONALIZATION T/F for the complex/real cases, minor inconsistencies of the previous implementation has appeared ...

RMeli:
> since DLAF is often not explicitly tested

Yes, ideally I'd like to add DLA-Future to the tests on Alps a well, but there seem to be some issues with MPICH. I started working on using Cray MPICH directly, but ended up a bit messy with the current CI setup, and temporarily abandoned the attempt.

I'll do some tests, and report back.

DCM-Uni-Paderborn:
Dear Ole, the branch has been rebased onto current master (including the recent Spack package updates). Please manually re-trigger the previously failing Spack (ssmp, CUDA P100) and Spack (psmp, CUDA P100) checks. CUDA Pascal Regtest and Performance CUDA Volta can also be re-triggered if desired, but the old failures looked confined to the P100 Spack runs.

DCM-Uni-Paderborn:
Manual generation failed while downloading https://www.cp2k.org/static/downloads/SIRIUS-7.11.1.tar.gz. The URL is reachable now (HTTP 200), so this looks transient. Please re-trigger Manual generation.

DCM-Uni-Paderborn:
I had another look at the currently failing checks on the rebased branch.

From the reports, I do not see a failure that looks specific to this PR:

- `Conventions`, `Precommit`, `Manual generation`, `Regtest pdbg`, the regular Spack jobs, `Spack ssmp-static`, `Spack OpenMPI pdbg`, and `Performance CUDA Volta` pass.
- The new `QS/regtest-scalapack-generalized` tests pass in the failing MPI/Spack/ARM64 jobs where they are enabled.
- The `QS/regtest-cusolver` and `QS/regtest-dlaf` tests are skipped in the failing P100 Spack jobs because their requirements are not satisfied there.
- `CUDA Pascal Regtest` reports `Something in the CI system went wrong :-(` and only provides a partial report.
- The remaining failures look unrelated to this PR: repeated FFTW/PINT/RTBSE/CUDA/COSMA/DBCSR segfaults, the recurring `Spack (psmp, 4x2)` empty `Runtime failure with code 7` block, and one ARM64 `Si_bulk8-5.inp` abort in Pulay mixing due to a bad condition number.

Could someone please re-trigger or classify these failing jobs? I am happy to investigate further if any of them are suspected to be related to the generalized diagonalization changes.

mkrack:
I think the current failures are not related (specific) to this PR.

src/fm/cp_cfm_diag.F
```diff
@@ -197,16 +314,34 @@ SUBROUTINE cp_cfm_geeig(amatrix, bmatrix, eigenvectors, eigenvalues, work)
...
+         END IF
+         CALL cp_cfm_general_cusolver(amatrix, bmatrix, work, evals)
+         IF (check_eigenvectors) THEN
+            CALL check_generalized_diag(overlap_check, work, scratch_check, nmo)
+            CALL cp_cfm_release(scratch_check)
+            CALL cp_cfm_release(overlap_check)
+         END IF
 #if defined(__DLAF)
-      IF (diag_type == FM_DIAG_TYPE_DLAF .AND. amatrix%matrix_struct%nrow_global >= dlaf_neigvec_min) THEN
+      ELSE IF (diag_type == FM_DIAG_TYPE_DLAF .AND. amatrix%matrix_struct%nrow_global >= dlaf_neigvec_min) THEN
```
RMeli:
```suggestion
      ELSE IF (diag_type == FM_DIAG_TYPE_DLAF .AND. nao >= dlaf_neigvec_min) THEN
```
For consistency with the cuSolverMP implementation? It might also be worth aligning the keywords (i.e. `dlaf_neigvec_min` vs `cosolver_n_min`)?

RMeli:
```suggestion
      ELSE IF (diag_type == FM_DIAG_TYPE_DLAF .AND. direct_generalized_diagonalization .AND. amatrix%matrix_struct%nrow_global >= dlaf_neigvec_min) THEN
```
Also for consistency?

src/fm/cp_cfm_diag.F
```diff
@@ -197,16 +314,34 @@ SUBROUTINE cp_cfm_geeig(amatrix, bmatrix, eigenvectors, eigenvalues, work)
...
 
       CALL timeset(routineN, handle)
 
       CALL cp_cfm_get_info(amatrix, nrow_global=nao)
       ALLOCATE (evals(nao))
       nmo = SIZE(eigenvalues)
-
+      check_eigenvectors = diag_check_requested()
+
+      IF (diag_type == FM_DIAG_TYPE_CUSOLVER .AND. direct_generalized_diagonalization .AND. &
```
RMeli, line 328: If no direct diagonalization is enabled, this will fall back to CP2K's Cholesky decomposition and call `cp_cfm_heevd`. However, as far as I can tell, `cp_cfm_heevd` does not support cuSolverMP. Is that on purpose? For DLA-Future, the standard DLAF eigensolver would still be used in `cp_cfm_heevd`.

src/fm/cp_fm_diag.F
```diff
@@ -1365,21 +1491,64 @@ SUBROUTINE cp_fm_geeig(amatrix, bmatrix, eigenvectors, eigenvalues, work)
...
+         CALL cp_fm_geeig_scalapack(amatrix, bmatrix, work, eigenvalues)
+         IF (check_eigenvectors) THEN
+            CALL check_generalized_diag(overlap_check, work, scratch_check, nmo)
+            CALL cp_fm_release(scratch_check)
+            CALL cp_fm_release(overlap_check)
+         END IF
          CALL cp_fm_to_fm(work, eigenvectors, nmo)
+#endif
 #if defined(__DLAF)
       ELSE IF (diag_type == FM_DIAG_TYPE_DLAF .AND. amatrix%matrix_struct%nrow_global >= dlaf_neigvec_min) THEN
```
RMeli:
```suggestion
      ELSE IF (diag_type == FM_DIAG_TYPE_DLAF .AND. direct_generalized_diagonalization .AND. amatrix%matrix_struct%nrow_global >= dlaf_neigvec_min) THEN
```
Again for consistency? Otherwise it can be a bit surprising that `direct_generalized_diagonalization=.false.` still uses the generalized eigensolver for DLA-Future. `choose_eigv_solver` supports DLAF, so it will behave as expected.

src/environment.F
```diff
@@ -830,6 +832,18 @@ SUBROUTINE read_global_section(root_section, para_env, globenv)
                globenv%dlaf_neigvec_min
          END IF
 
+         IF (globenv%diag_library == "cuSOLVER" .OR. globenv%diag_library == "ScaLAPACK") THEN
```
RMeli: DLAF too?

CMakeLists.txt
```diff
@@ -255,7 +255,7 @@ set_property(
            "NVHPCBlas"
            "CUSTOM")
 
-set(CP2K_SCALAPACK_VENDOR_LIST "MKL" "SCI" "GENERIC" "auto")
+set(CP2K_SCALAPACK_VENDOR_LIST "MKL" "NVPL" "SCI" "GENERIC" "auto")
```
RMeli, line 258: Should NVPL be supported also as a BLAS/LAPACK vendor? I did start working on that (and trying to exploit CMake support for BLAS), but never finished it off.

RMeli (commented):
Thanks. I had a look and left a few comments, mainly about trying to align the behavior as much as possible for all the libraries (at least DLAF/cuSolverMP).

src/fm/cp_cfm_diag.F
DCM-Uni-Paderborn: Thanks, agreed on the consistency point. I pushed d901a5b558, which now gates the DLAF generalized path with DIRECT_GENERALIZED_DIAGONALIZATION and uses the local `nao` value for the size check.

I kept `DLAF_NEIGVEC_MIN` as-is for now because it is already an exposed keyword and also controls the standard DLAF eigensolver threshold. The cuSolverMP threshold is currently a fixed implementation guard. Renaming/alignment of these threshold controls would be a broader input-interface cleanup and probably better handled separately.

DCM-Uni-Paderborn: Applied in d901a5b558. The DLAF branch now requires `DIRECT_GENERALIZED_DIAGONALIZATION` as well, matching the cuSolverMP and ScaLAPACK behavior.

src/fm/cp_cfm_diag.F
DCM-Uni-Paderborn: Good point. For the real path, `DIRECT_GENERALIZED_DIAGONALIZATION=false` falls back to the CP2K-side Cholesky reduction and then `choose_eigv_solver`, so cuSolverMP is still used for the resulting standard eigenproblem when applicable.

For the complex path, however, `cp_cfm_heevd` currently does not have a cuSolverMP standard Hermitian eigensolver wrapper; it only has the existing DLAF special case and otherwise uses the base implementation. I did not add that here because it would be a new feature beyond the generalized eigensolver changes, but I can follow up separately for complex cuSolverMP standard diagonalization.

src/fm/cp_fm_diag.F
DCM-Uni-Paderborn: Agreed, applied the same guard here in d901a5b558. With `DIRECT_GENERALIZED_DIAGONALIZATION=false`, DLAF can still be selected by `choose_eigv_solver` after the CP2K-side Cholesky reduction, which matches the non-direct behavior.

src/environment.F
DCM-Uni-Paderborn: Yes, added DLAF to this header output in d901a5b558 so the selected value of `DIRECT_GENERALIZED_DIAGONALIZATION` is visible for all backends that can use the direct generalized path.

CMakeLists.txt
DCM-Uni-Paderborn: I agree this would be useful. This PR currently only adds NVPL as a ScaLAPACK vendor using the NVIDIA CMake targets, and I verified that configuration on the Grace/Hopper test system.

Adding NVPL as a BLAS/LAPACK vendor would touch the more general BLAS/LAPACK discovery path, so I would prefer to keep it as a focused follow-up PR rather than expanding this diagonalization PR further.

DCM-Uni-Paderborn:
Please re-trigger Manual generation.

Issue #4854, opened 2026-02-18
Requested features regarding k-points


fstein93:
This is a collection of ideas and an extension of an older threads #225 , #4811 
* [ ] OT
* [ ] SCCS
* [x] DFT+U
* [x] PDOS
* [ ] k-point symmetry reduction
* [x] Wannier functions, interface to Wannier90
* [ ] TDDFT
* [ ] alternative diagonalization schemes for k-points
* [x] #496
* [x] restart k-point calculation from Gamma-point calculation
* [ ] Alternative HFX (for instance two-center RI as used in FHI-aims or (partially) reciprocal-space scheme as used in PySCF, Head-Gordon group)
* [ ] post-SCF-methods (MP2, RPA, GW)
* [x] Extrapolation schemes of wavefunction in MD or opt jobs, like PS and ASPC as specified by [`&FORCE_EVAL/&DFT/&QS/EXTRAPOLATION`](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/QS.html#CP2K_INPUT.FORCE_EVAL.DFT.QS.EXTRAPOLATION)
* [ ] Periodic electric field as requested by `&FORCE_EVAL/&DFT/&PERIODIC_EFIELD` (which may or may not be a separate task from OT)
* [ ] Linear response properties, like NMR, as specified by [`&FORCE_EVAL/&PROPERTIES/&LINRES`](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/PROPERTIES/LINRES.html)
* [x] Löwdin/Loewdin population analysis as requested by `&FORCE_EVAL/&DFT/&PRINT/&LOWDIN`
  https://github.com/cp2k/cp2k/blob/e4d53b9f1ccef4809321ad919bddccd42e4beec4/src/population_analyses.F#L143
* [ ] wavefunction stability analysis/optimization
* [ ] constrained DFT
* [x] Harris-functional
* [ ] (LS_SCF)

Feel free to add more points by editing this post (if you can, else post it below and we will add it later)

Growl1234:
Are “DFT+U” and “DFT+U with k-points” redundant?

And, for HFX, there has already been RI-HFXk, although its applicability is rather limited. We could also appropriately incorporate Prof. Hutter’s reply from the discussion as clarification:

[...]

PR #5117, opened 2026-04-30
Closed PR

DCM-Uni-Paderborn:


he-zilong:
Great job! This is the sixth action point in #4854 and some literature reference for K290 Stabilization Work in the manual may be beneficial (even if it is "in preparation").

I have a little question: does the current Monkhorst-Pack grid, with or without symmetry reduction and having an even or odd number of subdivisions (e.g. `2 x 2 x 2` vs `3 x 3 x 3`), always contain the gamma-point? [Some other implementations](https://vasp.at/wiki/KPOINTS#Regular_k-point_mesh) have a shifted Monkhorst-Pack mesh called "gamma-centered mesh" to address that.

src/cryssym.F
```diff
@@ -451,44 +474,393 @@ SUBROUTINE kp_symmetry(csym, xkp, wkp, kpop)
...
+
+   END FUNCTION reciprocal_rotation
+
+! **************************************************************************************************
+!> \brief Invert a 3 x 3 matrix
+!> \param a matrix to invert
+!> \return inverse matrix
+! **************************************************************************************************
+   FUNCTION inverse_3x3(a) RESULT(ainv)
+      REAL(KIND=dp), DIMENSION(3, 3), INTENT(IN)         :: a
```
fstein93: We have that already in the `mathlib` module

src/cryssym.F
```diff
@@ -451,44 +474,393 @@ SUBROUTINE kp_symmetry(csym, xkp, wkp, kpop)
...
+
+      INTEGER                                            :: i, ibsign, iop, j, kr, nkpts
+      LOGICAL                                            :: add_inversion
+      REAL(KIND=dp)                                      :: eps
+      REAL(KIND=dp), DIMENSION(3)                        :: diff, rr, wcart
+
+      nkpts = SIZE(wkp)
+      eps = MAX(1.e-12_dp, 10.0_dp*csym%delta)
+      add_inversion = .TRUE.
+      MARK_USED(li)
```
fstein93: I would remove this variable then

DCM-Uni-Paderborn: Done. The temporary add_inversion variable and the now-unused li argument were removed from reduce_kpoint_mesh; the loop now directly covers both signs.

src/cryssym.F
DCM-Uni-Paderborn: Dear Frederik,

Thanks, good point. I replaced the local inverse_3x3 helper with mathlib:inv_3x3.

DCM-Uni-Paderborn:
Thanks! I added the Worlton/Warren CPC 1972 reference for the K290 backend and made sure SPGLIB backend use cites Togo/Tanaka.

Regarding Gamma: no, the current CP2K Monkhorst-Pack generator does not always include Gamma. It uses fractional coordinates `(2*i - n - 1)/(2*n)`, so odd grids contain Gamma, while even grids are half-step shifted, e.g. `2 x 2 x 2` gives `+/-1/4` in each direction. Symmetry reduction only reduces the parent mesh and does not add Gamma. A Gamma-centered mesh would be a separate grid convention and is not changed by this PR.

I also fixed the pdbg failure locally. The failing tests aborted because reduced k-point grids with atom permutations/phases need local DBCSR matrices in that configuration; the follow-up commits include the corresponding handling and pass the affected `regtest-kp-1` inputs locally with 2 MPI ranks.

Issue #4319, opened 2025-07-19
Floating-point exception in Wavelet Poisson solver


oschuett:
During #4313 I encountered a floating-point exception in [regtest-gpw-4/H2O_wavelet_XZ.inp](https://github.com/cp2k/cp2k/blob/master/tests/QS/regtest-gpw-4/H2O_wavelet_XZ.inp):

```
 SCF WAVEFUNCTION OPTIMIZATION

  Step     Update method      Time    Convergence         Total energy    Change
  ------------------------------------------------------------------------------

Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x12aefa6c6e59 in ???
#1  0x12aefa6c5e75 in ???
#2  0x12aefa4d632f in ???
#3  0x12af12d78c56 in __pw_methods_MOD_pw_derive._omp_fn.0
	at /opt/cp2k/src/pw/pw_methods.F:2014
#4  0x12aefa9d9976 in ???
#5  0x12af12dab780 in __pw_methods_MOD_pw_derive
	at /opt/cp2k/src/pw/pw_methods.F:2012
#6  0x12af12dcb7e5 in calc_stress_and_gradient_c1d_gs
	at /opt/cp2k/src/pw/pw_poisson_methods.F:1149
#7  0x12af12dd29b9 in __pw_poisson_methods_MOD_pw_poisson_solve_v_nodv_c1d_gs_c1d_gs
	at /opt/cp2k/src/pw/pw_poisson_methods.F:698
#8  0x12af11930fc9 in __qs_ks_methods_MOD_qs_ks_build_kohn_sham_matrix
	at /opt/cp2k/src/qs_ks_methods.F:395
#9  0x12af11935cd3 in rebuild_ks_matrix
	at /opt/cp2k/src/qs_ks_methods.F:1286
#10  0x12af11935cd3 in __qs_ks_methods_MOD_qs_ks_update_qs_env
	at /opt/cp2k/src/qs_ks_methods.F:1132
#11  0x12af11c283ff in __qs_scf_MOD_scf_env_do_scf
	at /opt/cp2k/src/qs_scf.F:497
#12  0x12af11c32b95 in __qs_scf_MOD_scf
	at /opt/cp2k/src/qs_scf.F:247
#13  0x12af117d2034 in __qs_energy_MOD_qs_energies
	at /opt/cp2k/src/qs_energy.F:126
[...]

PR #5122, opened 2026-05-01
Support all 2D periodicities for the Wavelet Poisson solver

DCM-Uni-Paderborn:
This PR extends `POISSON_SOLVER WAVELET` so that all 2D periodicity combinations are supported: `PERIODIC XY`, `PERIODIC XZ`, and `PERIODIC YZ`.

Previously, the Wavelet solver only accepted `PERIODIC XZ`. The implementation now maps the requested 2D periodicity internally onto the solver axis convention, while keeping the existing optimized `XZ` path unchanged.

Main changes:
- Accept `POISSON%PERIODIC XY`, `XZ`, and `YZ` for `POISSON_SOLVER WAVELET`.
- Add internal Wavelet axis permutation support for rotated 2D cases.
- Fix density transfer between CP2K real-space grids and Wavelet z-sliced grids for permuted axes.
- Print the actual Wavelet periodicity instead of hardcoding `XZ`.
- Keep explicit mismatches allowed, but warn when `CELL%PERIODIC` and `POISSON%PERIODIC` differ, including for `WAVELET`.
- Update input documentation accordingly.

Regtests added:
- `H2O_wavelet_XY.inp`
- `H2O_wavelet_YZ.inp`

Validation performed:
- `make_pretty.sh`: OK
- `build-serial`: OK
- `build-parallel`: OK
- Rotated `XZ`, `XY`, and `YZ` Wavelet H2O tests give the same total energy: `-16.594332766163539 Ha`.
- Verified with both `ssmp` and `psmp -n 2`.
- Verified that explicit `CELL`/`POISSON` periodicity mismatch still runs but now emits a warning.


he-zilong:
The regtest pdbg reports
```
------------------------------ Slow Tests ------------------------------
Re-running QS/regtest-gpw-4 to avoid false positives.
Duration threshold (2x 95th %ile): 21.42 sec
Found 2 slow tests (16 suppressed):
    QS/regtest-gpw-4/H2O_wavelet_XY.inp                                              (  56.25 ±2.25 sec)
    QS/regtest-gpw-4/H2O_wavelet_YZ.inp                                              (  56.76 ±2.30 sec)
```
because the original `QS/regtest-gpw-4/H2O_wavelet_XZ.inp` is registered as a slow test after commit 18ccc3c to address the issue #4319 :
https://github.com/cp2k/cp2k/blob/d78ff4c8626e850fa08c58610234992c7c221bd1/tests/SLOW_TESTS_SUPPRESSIONS#L97-L98
Will restoring the original parameters of `H2O_wavelet_XZ.inp` still reproduce the crash in #4319 today, and would this also happen to the two new regtests?

DCM-Uni-Paderborn:
Dear Zilong,

thanks a lot for catching this. I checked the original smaller parameters again with the current PR code, and the previous XZ case no longer reproduces the crash locally. The two new rotated cases, XY and YZ, also run cleanly with the same smaller setup and give the same energy within the existing tolerance.

I therefore pushed a small follow-up commit that restores the Wavelet periodicity tests to CUTOFF 100 and ABC 8.0, updates the references, and removes the old H2O_wavelet_XZ slow-test suppression instead of adding new suppressions. This should keep the tests meaningful while avoiding the slow-test failure.

Thanks again for pointing to the old workaround and issue #4319.

PR #4971, opened 2026-03-18
Revamp cell parser and allow for fallback to parse from coordinate file

he-zilong:
Following recent developments of cell-parsing logic in #4950 and #4960, this PR
drafts implementation of an idea in #4962 as follows:

> Furthermore, if the cell is not provided explicitly, then we should start to fall back to reading it from the [coordinate file](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/SUBSYS/TOPOLOGY.html#CP2K_INPUT.FORCE_EVAL.SUBSYS.TOPOLOGY.COORD_FILE_NAME).

With this in mind, a note about precedence order of cell parsing logic is added
to the manual: firstly `CELL_FILE_NAME` and `CELL_FILE_FORMAT`, secondly `ABC`
and `ALPHA_BETA_GAMMA`, thirdly `A`, `B`, `C`, and finally `COORD_FILE_NAME`
under the `TOPOLOGY` section.

A major consequence is that `SUBROUTINE read_cell` having access to the `CELL`
section in input file as before is not enough now: it has to know the contents
of `TOPOLOGY` section with an optional `topology_section` pointer as input
argument. As I noted in [this comment](https://github.com/cp2k/cp2k/issues/4962#issuecomment-4068402866), `SUBROUTINE read_cell` is called in
several different contexts; a task yet to be done is to verify that each of
them do have access to `topology_section` (e.g. if they are already working
with the upper-level `subsys_section`) and that fallback to parsing coordinate
file for cell information is desirable, before adjusting anything beyond the
`cell_methods.F` code to enable this behavior. The first commit has no such
adjustments yet.

Also, `SUBROUTINE read_cell` in this PR has undergone a major overhaul due to
the former logic using auxiliary `SUBROUTINE read_cell_from_external_file` a
bit contrived to work with. In the current proof-of-concept, all cell data are
read from external file into `read_mat` or `read_ang` and `read_len` and has
units converted before assigning to the cell setup. Whether this may introduce
numeric noise and leads to noticeable variation of computational results is to
be tested.

Issue #3384, opened 2024-05-01
Potential Bug: Space group not maintained during cell optimization despite KEEP_SPACE_GROUP=T


zhangbw20:
Dear developers of cp2k, Greetings
### Description
I encountered an issue where the space group was not maintained during lattice optimization despite setting "KEEP_SPACE_GROUP" to true. This issue was observed while optimizing the lattice structures of iodine (No.64,Cmce) and silicon (No.227,Fd-3m) in their primitive cell forms, where the non-zero elements of three upper-triangular elements in lattice were left unchanged during the whole process of optimization.
```
$ grep -A2 "Vector a" I2_N01.out
 CELL| Vector a [angstrom]:       3.763    -2.281     0.000   |a| =     4.400496
 CELL| Vector b [angstrom]:       3.763     2.281     0.000   |b| =     4.400496
 CELL| Vector c [angstrom]:       0.000     0.000     9.706   |c| =     9.706146
--
 CELL| Vector a [angstrom]:       3.802    -2.281     0.000   |a| =     4.433490
 CELL| Vector b [angstrom]:       3.802     2.325     0.000   |b| =     4.456413
 CELL| Vector c [angstrom]:       0.000     0.000     9.693   |c| =     9.692778
--
 CELL| Vector a [angstrom]:       4.028    -2.281     0.000   |a| =     4.629319
 CELL| Vector b [angstrom]:       4.028     2.575     0.000   |b| =     4.781154
 CELL| Vector c [angstrom]:       0.000    -0.000     9.604   |c| =     9.603716
--
 CELL| Vector a [angstrom]:       4.269    -2.281     0.000   |a| =     4.840494
 CELL| Vector b [angstrom]:       4.269     2.825     0.000   |b| =     5.119503
 CELL| Vector c [angstrom]:       0.000    -0.000     9.495   |c| =     9.494987
--
 CELL| Vector a [angstrom]:       4.487    -2.281     0.000   |a| =     5.033592
 CELL| Vector b [angstrom]:       4.487     3.034     0.000   |b| =     5.416581
 CELL| Vector c [angstrom]:       0.000    -0.000     9.384   |c| =     9.383704
--
...
```

### Relevant code
Line 129~137 of gopt_f_methods.f:
```
            ALLOCATE (x0(3*nparticle + 6))
            CALL pack_subsys_particles(subsys=subsys, r=x0)
            idg = 3*nparticle
            DO i = 1, 3
               DO j = 1, i
[...]

PR #5123, opened 2026-05-01
Add k-point symmetry reduction methods

DCM-Uni-Paderborn:
## Summary

This PR adds symmetry reduction for Monkhorst-Pack k-point grids, with the
existing K290 machinery as the default backend and an optional SPGLIB-backed
reduction path for comparison and future extension.

- Add `KPOINTS%SYMMETRY_REDUCTION_METHOD K290|SPGLIB`.
- Keep `K290` as the default reduction backend.
- Add `KPOINTS%INVERSION_SYMMETRY_ONLY` for a conservative inversion/time-reversal
  reduction mode.
- Add `KPOINTS%EPS_SYMMETRY` as the canonical symmetry tolerance keyword while
  keeping `EPS_GEO` as a backwards-compatible alias.
- Add SPGLIB-based k-point orbit reduction using the symmetry operations and atom
  mappings accepted by CP2K.
- Keep restart handling and k-point group parallelization working with the
  reduced k-point sets.
- Add guards for method combinations where full atomic k-point symmetry is not
  yet validated.

## What Works

Full atomic k-point symmetry reduction is enabled for plain GPW k-point
calculations that do not use method features requiring additional, method-specific
phase or basis transformations.

The following parts are working and covered by local tests:

- K290 and SPGLIB both reduce equivalent Monkhorst-Pack grids and give matching
  energies for the tested GPW cases.
- `INVERSION_SYMMETRY_ONLY` reduces the grid to inversion/time-reversal pairs and
  avoids printing a backend name because no full atomic backend is active.
- K-point restart files can be written and read again for both K290 and SPGLIB.
- Reduced k-point sets work with PSMP k-point grouping.
- `MULTIPLE_UNIT_CELL 2 2 2` keeps the high-symmetry 64-atom benchmark reduced to
  the same 56 irreducible k-points as the primitive 8-atom cell.
- A symmetry-broken variant of the same 64-atom benchmark is reduced
  consistently to 252 irreducible k-points by both K290 and SPGLIB.

For HFX, RPA/GW exact exchange, GW, GAPW, GAPW_XC, LRIGPW, RIGPW and DFTB, the
code currently restricts requested full atomic k-point symmetry to the
conservative inversion/time-reversal path.  This keeps those calculations
running correctly while avoiding unvalidated atom/basis/phase transformations.

## What Was Explored

The implementation was tested along two paths.  First, the existing K290-based
CP2K symmetry infrastructure was extended and stabilized because it already
matches CP2K's internal atom mapping and phase conventions.  Second, SPGLIB was
added as an alternative reduction backend under `SYMMETRY_REDUCTION_METHOD
SPGLIB`, initially using the accepted symmetry operations conservatively.

Several broader method combinations were tried during development, including
GAPW, HFX, GW/RPA-related inputs, DFTB, LRIGPW and RIGPW.  These exposed
method-specific transformation questions outside plain GPW.  Instead of enabling
unsafe full atomic symmetry for those paths, this PR now guards them and falls
back to inversion/time-reversal reduction.

## K290 Stabilization Work

The first implementation attempts exposed two distinct issues in the K290 path:
the reduced k-point list itself was correct, but the density-matrix
reconstruction over the full Monkhorst-Pack mesh needed the right atom
permutations, cell shifts and phase factors for every symmetry-related k-point.
Several intermediate variants were tested to isolate this.

Steps that were tried and rejected:

- Reducing the k-point list without carrying enough atom/cell-shift information
  into the density reconstruction.  This worked for simple inversion cases but
  failed for lower-symmetry atomic mappings.
- Applying only the rotation part of the symmetry operation.  This missed the
  lattice translations needed when atoms are mapped across periodic cell
  boundaries.
- Changing the complex phase convention globally.  This fixed some trial cases
  but broke others, so the issue was narrowed down to using the correct
  operation-dependent shift instead of changing the global convention.
- Enabling full atomic symmetry for GAPW/HFX/GW/DFTB-style methods before the
  method-specific basis and phase transformations were validated.  Those paths
  were therefore guarded instead of being enabled unsafely.

The changes that ultimately made K290 stable for the GPW path were:

- Store and reuse the accepted atom permutations and cell shifts for the
  symmetry operations used in the k-point reduction.
- Use those operation-dependent mappings during density-matrix reconstruction
  rather than relying only on the reduced irreducible k-point list.
- Apply the phase factors consistently when reconstructing contributions from
  symmetry-related k-points.
- Keep full atomic symmetry active only for method combinations where those
  transformations are currently validated; fall back to inversion/time-reversal
  for the guarded methods.
- Cross-check the resulting K290 reduction against SPGLIB for the same accepted
  operations.

With these fixes, K290 gives the same one-step energies as the full grid for the
tested GPW cases, including the 64-atom `MULTIPLE_UNIT_CELL 2 2 2` benchmark,
and agrees with SPGLIB for both the high-symmetry 56-k-point case and the
symmetry-broken 252-k-point case.

## 64-Atom Benchmark

The benchmark uses a diamond-like H test cell, `MULTIPLE_UNIT_CELL 2 2 2`, a
`12 x 12 x 12` Monkhorst-Pack grid, the local serial build and
`OMP_NUM_THREADS=1`.  To keep this as a compact symmetry/timing check, each
input used `MAX_SCF 1` with `IGNORE_CONVERGENCE_FAILURE`; the relevant check is
that the one-step energies agree between full and reduced grids.

Fully symmetric structure:

| case | atoms | k-points | total energy [Ha] | wall time [s] |
|---|---:|---:|---:|---:|
| `1x1x1`, K290 | 8 | 56 | -3.82277282964690 | 0.84 |
| `1x1x1`, SPGLIB | 8 | 56 | -3.82277282964690 | 0.71 |
| `2x2x2`, full grid | 64 | 1728 | -30.58211574509667 | 27.93 |
| `2x2x2`, inversion only | 64 | 864 | -30.58211574509667 | 14.09 |
| `2x2x2`, K290 | 64 | 56 | -30.58211574509667 | 6.21 |
| `2x2x2`, SPGLIB | 64 | 56 | -30.58211574509667 | 6.24 |

Symmetry-broken variant with one displaced atom:

| case | atoms | k-points | total energy [Ha] | wall time [s] |
|---|---:|---:|---:|---:|
| `2x2x2`, full grid | 64 | 1728 | -30.669250091315476 | 27.52 |
| `2x2x2`, inversion only | 64 | 864 | -30.669250091315476 | 14.18 |
| `2x2x2`, K290 | 64 | 252 | -30.669250091315476 | 8.60 |
| `2x2x2`, SPGLIB | 64 | 252 | -30.669250091315476 | 8.61 |

The 56-k-point run is about 4.5x faster than the full 1728-point run in this
small one-step benchmark.  It is not expected to scale as `1728/56`, because the
density reconstruction and symmetry transformations still traverse the full
mesh; for example `transform_dmat` and `symtrans_phase` are still called for the
full grid.

## Tests Added

New regtests cover:

- GPW full atomic k-point symmetry with K290 and SPGLIB.
- Orthorhombic, monoclinic, hexagonal C3 and tetragonal C4 symmetry cases.
- Inversion-only reduction.
- K-point restart for K290 and SPGLIB.
- Guarded compatibility cases for GAPW, GAPW_XC, LRIGPW, RIGPW, HFX, RI-HFX/GW
  and DFTB.

Representative new inputs include:

- `tests/QS/regtest-kp-1/h_sym_red.inp`
- `tests/QS/regtest-kp-1/h_spglib_sym_red.inp`
- `tests/QS/regtest-kp-1/h_inv_red.inp`
- `tests/QS/regtest-kp-1/h_sym_red_restart.inp`
- `tests/QS/regtest-kp-1/h_spglib_sym_red_restart.inp`
- `tests/QS/regtest-kp-1/h_hex_c3_sym_red.inp`
- `tests/QS/regtest-kp-1/h_tetra_c4_sym_red.inp`
- `tests/QS/regtest-kp-1/c_gapw_kpsym.inp`
- `tests/QS/regtest-kp-1/c_gapwxc_kpsym.inp`
- `tests/QS/regtest-kp-2/cc2_lri_kpsym.inp`
- `tests/QS/regtest-kp-hfx-ri/hBN_gpw_pbe0_kpsym.inp`
- `tests/QS/regtest-kp-hfx-ri/diamond_gpw_extRI_kpsym.inp`
- `tests/QS/regtest-scalable-gw/07_G0W0_periodic_H2Te_kp_sampling_SCF_kpsym.inp`
- `tests/DFTB/regtest-nonscc/si_kp1_kpsym.inp`
- `tests/DFTB/regtest-scc/c_kp1_kpsym.inp`

## Local Testing

- Built the serial target with CMake/Ninja.
- Built the serial Libint-enabled target for HFX coverage.
- Built and tested the PSMP target for k-point grouping.
- Checked K290, SPGLIB and inversion-only GPW energies against full-grid or
  matching-backend references.
- Verified K290/SPGLIB restart reads from written `.kp` files.
- Verified guarded HFX/GAPW/GW/DFTB-related inputs continue to run and do not
  claim a full atomic symmetry backend when only inversion/time-reversal is used.


juerghutter:
Merge now. This will be best explored and further tested in real life applications.

JWilhelm:
Great!! We will check the k-point symmetry on our crystal benchmark set and will let you know about the result!

he-zilong:
> Great!! We will check the k-point symmetry on our crystal benchmark set and will let you know about the result!

A crystal benchmark set sounds very interesting! Does it cover each of the 230 space groups? Computationally how expensive the test cases are in general?

samalbibek:
> > Great!! We will check the k-point symmetry on our crystal benchmark set and will let you know about the result!
> 
> A crystal benchmark set sounds very interesting! Does it cover each of the 230 space groups? Computationally how expensive the test cases are in general?

Currently, we have  the dataset for 3d crystals, which consists of 103 materials. For these systems, the computational cost is relatively modest—test cases typically run in a few minutes when using smaller basis sets such as DZVP-MOLOPT-GTH and without k-point symmetry. We will update the timings both with and without k-point symmetry, which we will share soon.

DCM-Uni-Paderborn:
Dear Zilong,

For this PR, we also performed an extensive local validation sweep covering all 230 space groups. That sweep was mainly used as a diagnostic benchmark for the symmetry detection and k-point reduction machinery, including primitive and non-primitive cells, centrosymmetric and non-centrosymmetric cases, and symmorphic as well as non-symmorphic operations.

For the CP2K regtests we intentionally kept only a representative subset. Running all 230 space groups in CI would be too heavy and also somewhat redundant for every commit. The committed regtests therefore focus on relevant edge cases, such as different lattice centerings, inversion-only versus full symmetry, the K290/SPGLIB paths, fractional translations/non-symmorphic cases, and comparison against unreduced/full-grid references.

The individual local space-group test cases were designed to be small, typically using compact cells, modest basis sets, and small k-point meshes, so a single case is usually cheap. The cost becomes significant only because the full sweep multiplies this by 230 space groups and, in some checks, by several symmetry backends/settings. That is why the exhaustive sweep is better suited as a local/pre-PR validation benchmark, while CI keeps selected representative cases.

DCM-Uni-Paderborn:
Dear Bibek,

that sounds very useful, thanks a lot. We would be very interested in any practical experience from this benchmark set, in particular timings with and without k-point symmetry, the achieved reduction factors, and any cases where the symmetry reduction is unexpectedly conservative or fails.

Please also let us know about any bugs, suspicious energies, problematic space groups, or cases where K290 and SPGLIB behave differently. Such real-material benchmarks are exactly what will help us improve the implementation beyond the current regtest coverage.

Best,
Thomas

samalbibek:
> Dear Bibek,
> 
> that sounds very useful, thanks a lot. We would be very interested in any practical experience from this benchmark set, in particular timings with and without k-point symmetry, the achieved reduction factors, and any cases where the symmetry reduction is unexpectedly conservative or fails.
> 
> Please also let us know about any bugs, suspicious energies, problematic space groups, or cases where K290 and SPGLIB behave differently. Such real-material benchmarks are exactly what will help us improve the implementation beyond the current regtest coverage.
> 
> Best, Thomas

Dear Thomas,

The newer version of CP2K seems to have an issue during the SCF convergence step. 

For your reference, I have attached the input and output files for the bulk AlAs. I have also included the output file obtained using an older CP2K binary, which completes successfully.  We are currently figuring out in which precise commit this SCF convergence issue is introduced.

Thanks,
Bibek
[cp2k_oldbin_out.txt](https://github.com/user-attachments/files/27604331/cp2k_oldbin_out.txt)
[cp2k_out.txt](https://github.com/user-attachments/files/27604332/cp2k_out.txt)
[cp2k_inp.txt](https://github.com/user-attachments/files/27604333/cp2k_inp.txt)

he-zilong:
> Dear Thomas,
> 
> The newer version of CP2K seems to have an issue during the SCF convergence step.
> 
> For your reference, I have attached the input and output files for the bulk AlAs. I have also included the output file obtained using an older CP2K binary, which completes successfully. We are currently figuring out in which precise commit this SCF convergence issue is introduced.
> 
> Thanks, Bibek 

Dear Bibek,

The primitive FCC lattice specified in this input file may deviate from the program convention as mentioned in #3384 :
```
    &CELL
      A   0.0000   2.8300   2.8300
      B   2.8300   0.0000   2.8300
      C   2.8300   2.8300   0.0000
      PERIODIC XYZ
    &END CELL
```
Instead of giving cell vectors `A`, `B`, `C`, could you try specifying the lengths and angled with `ABC` (which would be three 2.83*sqrt(2) values) and `ALPHA_BETA_GAMMA` (three 60 degrees values) and see if the old and new versions of CP2K have identical behaviors?
A new warning that you can see early in the new output was added by me, but I am also curious how SCF cycles in a static calculation where the cell is not varying could be affected...
```
 *** WARNING in cell_methods.F:457 :: Cell vectors are read but cell   ***
 *** matrix is not a lower triangle, not conforming to the program     ***
 *** convention that A lies along the X-axis and B is in the XY plane. ***
```
Regards,
Zilong

JWilhelm:
Dear Zilong,

thanks a lot for checking this. That was super helpful, and I also think this is the issue.

Still, I am quite concerned about the implications. If I understand correctly, this means that some formerly working k-point inputs may now give wrong results when the cell matrix is not in the newly adopted CP2K input convention. What worries me is that users often take lattice vectors directly from databases or other codes and paste the A/B/C vectors into CP2K. If the corresponding cell matrix is not lower triangular in the convention expected by CP2K, then they may get wrong results with only a warning. In practice, many users will not notice such a warning.

Would it be possible to make CP2K transform a general valid unit cell internally into the appropriate CP2K format? If that is not straightforward, could we at least make this a hard error instead of only a warning?

Do you also know which PR or commit introduced the change after which these formerly working inputs no longer work as expected?

Thanks again for finding this. I think it is important that we understand the implications of the issue and decide on a robust strategy, because this could otherwise be quite dangerous for users.

DCM-Uni-Paderborn:
Thanks for looking into this. I agree that the attached AlAs input points to a cell-convention issue rather than to the k-point symmetry reduction introduced in #5123. In both the old and new outputs the k-point point-group symmetrization is OFF and the explicit k-point list contains 2048 points, so the new symmetry-reduction machinery is not active here.

However, the case is still quite concerning. The new run only emits a warning for the non-lower-triangular primitive FCC cell and then continues, but the SCF behavior and the resulting density/energy are clearly wrong compared to the older binary. Since users often paste valid lattice vectors directly from databases or other codes, I agree that a warning is probably not sufficient.

As an immediate robust fix, CP2K should probably abort for such non-conforming A/B/C cell matrices in k-point calculations, at least until the full internal canonicalization is implemented. The more user-friendly long-term solution would be to transform a general valid input cell internally into CP2K's canonical convention, but that needs to consistently transform scaled coordinates, k-points, band paths, stress/restart conventions, etc.

The warning itself appears to have been introduced in #4971 / commit 5557c943c6c, not in #5123. A bisection around that change would be useful to confirm whether the behavioral regression starts there.

he-zilong:
> Still, I am quite concerned about the implications. If I understand correctly, this means that some formerly working k-point inputs may now give wrong results when the cell matrix is not in the newly adopted CP2K input convention. What worries me is that users often take lattice vectors directly from databases or other codes and paste the A/B/C vectors into CP2K. If the corresponding cell matrix is not lower triangular in the convention expected by CP2K, then they may get wrong results with only a warning. In practice, many users will not notice such a warning.

I think the program convention has been there for a while (since the assumption is taken as granted in some tasks that changes cell like MD or OPT), and only recently it is realized that it should be conveyed with more clarity in the manual... I have not heard of a database with non-lower-triangular cell matrix, but I do agree that the discrepancy in a static calculation is not desirable.

> Would it be possible to make CP2K transform a general valid unit cell internally into the appropriate CP2K format? 

That is presumably already how the keywords `ABC` and `ALPHA_BETA_GAMMA` work under the hood (and I recommend using them with `SCALED` coordinates rather than Cartesian coordinates). Alternatively, an external `CIF` can be similarly parsed for the coordinate (`COORD_FILE_NAME`) and cell (`CELL_FILE_NAME`, though this may be omitted due to a recently introduced fallback) information.

> Do you also know which PR or commit introduced the change after which these formerly working inputs no longer work as expected?

**_(Edit: updated)_** My guess is some revision after my PR #5140 that got merged five days ago; from a local CP2K compiled at commit 223025a, the converged result from two kinds of cell specifications still agree well. I prepared two input files from `cp2k_inp.txt` above with the following difference:
```
80,82c80,81
<       A   0.0000   2.8300   2.8300
<       B   2.8300   0.0000   2.8300
<       C   2.8300   2.8300   0.0000
---
>       ABC 2.83*SQRT(2) 2.83*SQRT(2) 2.83*SQRT(2)
>       ALPHA_BETA_GAMMA 60.0 60.0 60.0
```
The difference between output logs starting from the warning is attached below.
<details>
<summary>diff ./cell_1.out ./cell_2.out</summary>

```
121,125d120
<  *** WARNING in cell_methods.F:457 :: Cell vectors are read but cell   ***
<  *** matrix is not a lower triangle, not conforming to the program     ***
<  *** convention that A lies along the X-axis and B is in the XY plane. ***
<
<
127,129c122,124
<  CELL| Vector a [angstrom]:       0.000     2.830     2.830   |a| =     4.002224
<  CELL| Vector b [angstrom]:       2.830     0.000     2.830   |b| =     4.002224
<  CELL| Vector c [angstrom]:       2.830     2.830     0.000   |c| =     4.002224
---
>  CELL| Vector a [angstrom]:       4.002     0.000     0.000   |a| =     4.002224
>  CELL| Vector b [angstrom]:       2.001     3.466     0.000   |b| =     4.002224
>  CELL| Vector c [angstrom]:       2.001     1.155     3.268   |c| =     4.002224
137,139c132,134
<  CELL_REF| Vector a [angstrom     0.000     2.830     2.830   |a| =     4.002224
<  CELL_REF| Vector b [angstrom     2.830     0.000     2.830   |b| =     4.002224
<  CELL_REF| Vector c [angstrom     2.830     2.830     0.000   |c| =     4.002224
---
>  CELL_REF| Vector a [angstrom     4.002     0.000     0.000   |a| =     4.002224
>  CELL_REF| Vector b [angstrom     2.001     3.466     0.000   |b| =     4.002224
>  CELL_REF| Vector c [angstrom     2.001     1.155     3.268   |c| =     4.002224
2509c2504
<       2    2 As   33      1.415000      1.415000      1.415000   5.0000  74.9216
---
>       2    2 As   33      2.001112      1.155343      0.816951   5.0000  74.9216
2741,2756c2736,2751
<      1 NoMix/Diag. 0.10E+00   10.0     1.16941532        -8.5065309738 -8.51E+00
<      2 Broy./Diag. 0.10E+00   11.3     0.00660009        -8.2950195408  2.12E-01
<      3 Broy./Diag. 0.10E+00   11.5     0.09612458        -8.2703179989  2.47E-02
<      4 Broy./Diag. 0.10E+00   11.5     0.00302944        -8.3851221001 -1.15E-01
<      5 Broy./Diag. 0.10E+00   11.5     0.00940119        -8.4014253640 -1.63E-02
<      6 Broy./Diag. 0.10E+00   11.3     0.00037135        -8.4022565702 -8.31E-04
<      7 Broy./Diag. 0.10E+00   11.6     0.00316822        -8.4135751507 -1.13E-02
<      8 Broy./Diag. 0.10E+00   11.4     0.00011731        -8.4148860900 -1.31E-03
<      9 Broy./Diag. 0.10E+00   11.6     0.00030949        -8.4180912117 -3.21E-03
<     10 Broy./Diag. 0.10E+00   11.8     0.00004854        -8.4182975330 -2.06E-04
<     11 Broy./Diag. 0.10E+00   11.7     0.00003396        -8.4186619625 -3.64E-04
<     12 Broy./Diag. 0.10E+00   11.8     0.00001166        -8.4187682298 -1.06E-04
<     13 Broy./Diag. 0.10E+00   12.1     0.00000848        -8.4187673746  8.55E-07
<     14 Broy./Diag. 0.10E+00   11.9     0.00000165        -8.4187648454  2.53E-06
<     15 Broy./Diag. 0.10E+00   11.9     0.00000412        -8.4187483933  1.65E-05
<     16 Broy./Diag. 0.10E+00   11.7     0.00000087        -8.4187427112  5.68E-06
---
>      1 NoMix/Diag. 0.10E+00   10.3     1.16941553        -8.5065309738 -8.51E+00
>      2 Broy./Diag. 0.10E+00   11.3     0.00660012        -8.2950195055  2.12E-01
>      3 Broy./Diag. 0.10E+00   11.1     0.09612513        -8.2703179953  2.47E-02
>      4 Broy./Diag. 0.10E+00   10.9     0.00302944        -8.3851220922 -1.15E-01
>      5 Broy./Diag. 0.10E+00   11.3     0.00940119        -8.4014253634 -1.63E-02
>      6 Broy./Diag. 0.10E+00   11.2     0.00037135        -8.4022565695 -8.31E-04
>      7 Broy./Diag. 0.10E+00   11.7     0.00316822        -8.4135751482 -1.13E-02
>      8 Broy./Diag. 0.10E+00   11.6     0.00011731        -8.4148860867 -1.31E-03
>      9 Broy./Diag. 0.10E+00   11.8     0.00030949        -8.4180912073 -3.21E-03
>     10 Broy./Diag. 0.10E+00   11.8     0.00004854        -8.4182975285 -2.06E-04
>     11 Broy./Diag. 0.10E+00   11.7     0.00003396        -8.4186619580 -3.64E-04
>     12 Broy./Diag. 0.10E+00   11.7     0.00001166        -8.4187682253 -1.06E-04
>     13 Broy./Diag. 0.10E+00   12.1     0.00000848        -8.4187673700  8.55E-07
>     14 Broy./Diag. 0.10E+00   12.6     0.00000165        -8.4187648408  2.53E-06
>     15 Broy./Diag. 0.10E+00   12.8     0.00000412        -8.4187483886  1.65E-05
>     16 Broy./Diag. 0.10E+00   13.2     0.00000087        -8.4187427065  5.68E-06
2768,2770c2763,2765
<   Core Hamiltonian energy:                                      4.27199145749173
<   Hartree energy:                                               8.92108386725779
<   Exchange-correlation energy:                                 -2.40767259621379
---
>   Core Hamiltonian energy:                                      4.27199150291847
>   Hartree energy:                                               8.92108382038538
>   Exchange-correlation energy:                                 -2.40767259004288
2772c2767
<   Fermi energy:                                                 0.16421175422957
---
>   Fermi energy:                                                 0.16421204653989
2774,2775c2769,2770
<   Total energy:                                                -8.41874271124270
<   Total energy (extrapolated to T->0):                         -8.41874271124270
---
>   Total energy:                                                -8.41874270651745
>   Total energy (extrapolated to T->0):                         -8.41874270651745
2781,2782c2776,2777
<        1     Al       1          2.535388                              0.464612
<        2     As       2          5.464612                             -0.464612
---
>        1     Al       1          2.535385                              0.464615
>        2     As       2          5.464615                             -0.464615
2814c2809
<  KPOINTS| Time for k-point line                                            0.926
---
>  KPOINTS| Time for k-point line                                            0.984
2816c2811
<  ENERGY| Total FORCE_EVAL ( QS ) energy [hartree]             -8.418742711242697
---
>  ENERGY| Total FORCE_EVAL ( QS ) energy [hartree]             -8.418742706517447
```

</details>

he-zilong:
Dear Bibek @samalbibek ,

Unfortunately (or not), I could _not_ reproduce the failure of SCF convergence in your `cp2k_out.txt`. My toolchain-based CP2K behaves consistently from 223025a to the very recent 49c4bc4 (covering a range before and after 59927b9) for both kinds of cell specifications, converging within numerical precision to the same result as in the collapsed excerpt above. Probably this means that I jumped to the conclusion too quickly and in fact nothing behaves erroneously with a k-point band calculation on a non-conventional cell matrix. I will shortly afterwards leave a message at #5170 for clarification.

There are really a lot of reasons for SCF going wild, and maybe more scrutiny is required on your side, for instance about the old and new environments. A case in point,I use ELPA as the diagonalization library like your `cp2k_oldbin_out.txt` for the results above; using `PREFERRED_DIAG_LIBRARY SCALAPACK` for a similar setup to `cp2k_out.txt` does not have any anomalous effect on SCF cycles and the outcome.

Apologies for my help attempt ending up being misleading...

Regards,
Zilong

samalbibek:
> Dear Bibek @samalbibek ,
> 
> Unfortunately (or not), I could _not_ reproduce the failure of SCF convergence in your `cp2k_out.txt`. My toolchain-based CP2K behaves consistently from [223025a](https://github.com/cp2k/cp2k/commit/223025a8cec493916b4ea895ddc77de1ea0dc56a) to the very recent [49c4bc4](https://github.com/cp2k/cp2k/commit/49c4bc4af337803662974fd9d753d10067d1a4aa) (covering a range before and after [59927b9](https://github.com/cp2k/cp2k/commit/59927b92295ba76b543f0f0c33dbb8907e9998c0)) for both kinds of cell specifications, converging within numerical precision to the same result as in the collapsed excerpt above. Probably this means that I jumped to the conclusion too quickly and in fact nothing behaves erroneously with a k-point band calculation on a non-conventional cell matrix. I will shortly afterwards leave a message at #5170 for clarification.
> 
> There are really a lot of reasons for SCF going wild, and maybe more scrutiny is required on your side, for instance about the old and new environments. A case in point,I use ELPA as the diagonalization library like your `cp2k_oldbin_out.txt` for the results above; using `PREFERRED_DIAG_LIBRARY SCALAPACK` for a similar setup to `cp2k_out.txt` does not have any anomalous effect on SCF cycles and the outcome.
> 
> Apologies for my help attempt ending up being misleading...
> 
> Regards, Zilong

Dear Zilong,

Thank you very much for taking the time to investigate this issue.

We will look into the problem more thoroughly on our side to identify the underlying cause.

Thank you again for your help and for the detailed investigation.

Best regards,
Bibek

samalbibek:
Dear Thomas,

The earlier issue has been resolved; it was due to some missing libraries during the compilation of the latest version of CP2K.

I have now run the k-point symmetry calculations for several elemental materials. For a few cases, the calculations terminated with warnings, while for others they completed successfully. I have summarised the results in the attached PDF
[kpt-sym test - elemental materials.pdf](https://github.com/user-attachments/files/28163033/kpt-sym.test.-.elemental.materials.pdf).

However, I am not fully confident about the correctness of the k-point symmetry calculations based on the results, particularly when comparing total energies and wall times with and without kpt symmetry calculations. I have also included all relevant input and output files for your reference (https://github.com/samalbibek/CP2K). The only modifications I made in the input files (compared to the non–k-point symmetry runs) are listed below; everything else was kept unchanged.

    &KPOINTS
      EPS_SYMMETRY 1.0E-8
      FULL_GRID OFF
      SCHEME MONKHORST-PACK 16 16 16
      PARALLEL_GROUP_SIZE   -1
      SYMMETRY ON
    &END KPOINTS

Could you kindly let me know if there might be anything missing or incorrectly specified in the input that could be leading to these unexpected results?

Kind regards, 
Bibek
[kpt-sym test - elemental materials.pdf](https://github.com/user-attachments/files/28163107/kpt-sym.test.-.elemental.materials.pdf)

DCM-Uni-Paderborn:
Dear Bibek,

I checked the Fe case with a current CP2K master. The very large discrepancy in your k-point symmetry result is not consistent with ordinary library/build differences: in your output the integrated electronic density is wrong by about 34 electrons, so the reduced-k-point result is physically invalid.

Please note that a master cloned on Friday 22 May predates PR #5275, merged on 23 May, which fixes K290 symmetry for nonstandard cells and large k-point meshes. With current master the catastrophic Fe discrepancy is no longer reproduced on my side, but I still see a residual non-neutral integrated density of order 1e-2 e for the K290-reduced Fe test. I will have a look ... 

Best,
Thomas

PR #5124, opened 2026-05-01
Add k-point symmetry backends and guards

DCM-Uni-Paderborn:
This PR extends CP2K k-point symmetry reduction with a selectable backend/reduction interface and adds conservative guards for methods where full atomic k-point symmetry is not yet validated.

### Summary

- Add `DFT%KPOINTS%SYMMETRY_BACKEND K290|SPGLIB` in addition to `SYMMETRY_REDUCTION_METHOD K290|SPGLIB`.
- Keep K290 as the default backend and reduction method.
- `SYMMETRY_BACKEND` selects the source/application of atomic symmetry operations, whereas `SYMMETRY_REDUCTION_METHOD` selects the algorithm used to reduce Monkhorst-Pack and MacDonald k-point meshes.
- If `SYMMETRY_BACKEND SPGLIB` is set without an explicit reduction method, the reduction method follows and becomes `SPGLIB`.
- `SYMMETRY_REDUCTION_METHOD SPGLIB` with `SYMMETRY_BACKEND K290` is supported as a comparison/diagnostic mode.
- `SYMMETRY_BACKEND SPGLIB` with `SYMMETRY_REDUCTION_METHOD K290` is rejected because SPGLIB backend operation requires SPGLIB reduction mappings.
- Add `EPS_SYMMETRY` as the preferred k-point symmetry tolerance keyword, while keeping `EPS_GEO` as an alias.
- Fix k-point symmetry phase handling and atom/basis transformations for the enabled paths.
- Enable/cover symmetry-reduced k-points for GPW, GAPW, GAPW_XC, HFX/RI-HFX smoke cases, DFTB, xTB, and gamma-only ACTIVE_SPACE.
- Add guards for response/correlation paths where full k-point symmetry is not yet production-ready.

### Guarded behavior

The following cases are conservatively guarded or restricted:

- `DFT%KPOINTS` with MP2/SOS-MP2, GW/BSE, TDDFPT/TDDFT-style response, and linear-response paths aborts early with method-specific messages.
- LOW_SCALING RPA k-point symmetry is restricted to inversion/time-reversal.
- ACTIVE_SPACE supports gamma k-points only; multi-k-point ACTIVE_SPACE remains guarded.
- GW should use its own `WF_CORRELATION%LOW_SCALING%KPOINTS` / `RI_RPA%GW%KPOINTS_SELF_ENERGY` paths, not `DFT%KPOINTS`.
- `WAVEFUNCTIONS REAL` is supported for gamma/special k-point cases without nontrivial Bloch phases. General atomic k-point symmetries with phase factors require `WAVEFUNCTIONS COMPLEX`.
- Nonzero shifted `MACDONALD` meshes remain under the existing CP2K restriction and are not symmetry-reduced by this PR.

### Tests added

New/updated regtests cover:

- K290 and SPGLIB k-point symmetry reduction for several cell symmetries.
- SPGLIB backend/reduction combinations.
- Explicit `WAVEFUNCTIONS COMPLEX` for a nontrivial SPGLIB backend case.
- Explicit `WAVEFUNCTIONS REAL` for a gamma/special-k-point SPGLIB backend case.
- MacDonald mesh coverage for the unshifted symmetry-reduced SPGLIB backend path.
- GAPW and GAPW_XC.
- HFX/RI-HFX smoke tests.
- DFTB and xTB k-point symmetry smoke tests.
- LOW_SCALING RPA inversion/time-reversal guard behavior.
- Gamma k-points for ACTIVE_SPACE.
- Restart coverage for symmetry-reduced k-point inputs.

### Local validation

- Built `cp2k.ssmp`.
- Ran `make_pretty.sh` on the PR files in a temporary ASCII-path worktree.
- `git diff --check` passes.
- `check_inputs.py` reports 0 errors for the main touched test directories.
- Relevant local regtests were run, including `QS/regtest-kp-1`, `QS/regtest-kp-2`, `QS/regtest-kp-hfx-ri-admm-2`, DFTB, and xTB/tblite smoke directories.


src/cryssym.F
```diff
@@ -441,54 +507,633 @@ SUBROUTINE kp_symmetry(csym, xkp, wkp, kpop)
...
+                  kr = csym%ibrot(iop)
                END IF
+
+               DO j = 1, nkpts
+                  diff(1:3) = xkp(1:3, j) - rr(1:3)
+                  diff(1:3) = diff(1:3) - ANINT(diff(1:3))
+                  IF (ALL(ABS(diff(1:3)) < eps)) THEN
+                     IF (csym%kplink(1, j) == 0) THEN
+                        csym%kplink(1, j) = i
+                        wkp(i) = wkp(i) + 1.0_dp
```
JWilhelm, line 830: Can you briefly explain why we add 1.0 to the k-point weight? I thought the sum of all weights needs to be 1? I couldn't find a division of the k-points weights by the number of k-points.

JWilhelm (approved):
Thanks a lot, Thomas. We will try the k-point symmetry for our 2d materials and 3d bulk materials benchmark set and will let you know about the performance. 

There is one minor confusion to me concerning k-point weights, see comment below.

marci73 (approved):

src/cryssym.F
juerghutter: See kpoint_methods line 184

JWilhelm: Ah, got it, thanks, Jürg!

PR #5126, opened 2026-05-02
Make spin-channel representation consistent in reciprocal-space mixing

Growl1234:
Follow up to #5061.

In the current implementation, `gspace_mixing` already transforms the two spin densities from the spin-up/spin-down representation to the charge/magnetization representation before calling the individual mixing routines. However, the Pulay and Broyden mixers then performed an additional internal spin-channel transformation, changing the second channel from the intended magnetization density into a different linear combination of the spin densities. As a result, `ALPHA_MAG` and `BETA_MAG` might not have acted on the magnetization channel as intended.

This PR removes the redundant spin-channel transformations inside `pulay_mixing` and `broyden_mixing`, and makes the spin-channel representation in reciprocal-space mixing explicit and consistent across grid densities, GAPW compensation charges, and mixing history.

### DIFF
--- a/src/qs_gspace_mixing.F
+++ b/src/qs_gspace_mixing.F
@@ -73,7 +73,7 @@ CONTAINS
       CHARACTER(len=*), PARAMETER                        :: routineN = 'gspace_mixing'
 
       INTEGER                                            :: handle, iatom, ig, ispin, natom, ng, &
-                                                            nimg, nspin
+                                                            nspin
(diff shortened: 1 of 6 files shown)


Growl1234:
It seems not that straight-forward, and there are still more things have to be considered...

oschuett:
It seem, a merge conflict has appeared over the weekend.

PR #5131, opened 2026-05-05
Fix TB gradients, virials, and k-point symmetry

DCM-Uni-Paderborn:
# Fix TB gradients, virials, tblite reference checks, and k-point symmetry

## Summary

This PR continues the CP2K/tblite interface work originally started by Johann Pototschnig
(`jpoto`) and relies in part on Sebastian Ehlert's `tblite` library. It makes the CP2K tight
binding paths substantially more consistent for nuclear gradients, virials/stress tensors, and
k-point symmetry, especially for DFTB, CP2K-native GFN0/GFN1, and `&XTB &TBLITE` GFN1/GFN2/IPEA1.

Main changes:

- fix several CP2K/tblite force and virial terms for periodic systems,
- add a native `tblite` CLI reference path under `&XTB &TBLITE &REFERENCE_CLI`,
- enable and test k-point symmetry reduction for CP2K-native TB and CP2K/tblite TB paths,
- improve DFTB SCC/Ewald virials and update affected DFTB references,
- add/extend regtests for molecular and truly periodic systems, including force/stress DEBUG checks,
- update the CP2K toolchain patches for `tblite`/embedded DFTD4 and standalone DFTD4.

## User-visible input changes

- `&DFT &XTB` can now select the library implementation via
  `GFN_TYPE TBLITE`; in that case an `&XTB &TBLITE` section is required.
- `&XTB &TBLITE` supports `METHOD GFN1`, `METHOD GFN2`, and `METHOD IPEA1`.
- `&XTB &TBLITE` has an optional `REFERENCE_CLI` keyword and a lone
  `&REFERENCE_CLI` subsection. If present/enabled, CP2K writes a `.gen` geometry, runs native
  `tblite run --grad`, and compares energy, gradient, and virial.
- `SCC_MIXER TBLITE` is documented as a modified Broyden mixer; the associated controls are named
  `TBLITE_MIXER_DAMPING` and `TBLITE_MIXER_MAX_SCF`.
- `DO_EWALD` remains available as a deprecated/debug override; the manual now documents the required
  periodic `POISSON` setup for CP2K-native DFTB/xTB and the `TBLITE` exception.
- CP2K warns if `CELL%PERIODIC` and `POISSON%PERIODIC` differ, but still runs.

### CP2K/tblite GFN1/GFN2/IPEA1

- Periodic image-cell matrix construction now fills all image cells instead of only image cell 1.
- H0, overlap/potential, GFN2 multipole, CN-chain-rule, and D4-related virial paths were checked
  term by term against finite differences.
- Periodic self-image virial signs/factors were corrected in the CP2K/tblite interface.
- The previously bad strongly distorted Si8/GFN2 Gamma stress case was traced to a hard D4 two-body
  cutoff discontinuity and fixed by adding a smooth switching region in the D4 path.
- `RUN_TYPE DEBUG` force/stress checks now behave consistently for molecular and periodic
  CP2K/tblite GFN1/GFN2 cases; isolated stress is allowed for tblite with a warning because it is a
  diagnostic virial, not a physical isolated-system stress.

### Native tblite CLI reference

The `REFERENCE_CLI` path initially showed good molecular agreement but poor periodic virials for
Ar-fcc:

- GFN1 periodic virial diff was about `1.15e-2`,
- GFN2 periodic virial diff was about `2.95e-2`.

The parser, units, row order, and sign convention were checked and were correct. The remaining error
was in native `tblite`: periodic H0 self-image pairs (`iat == jat`, non-zero translation) used the
normal pair sigma factor `0.5`. The correct symmetrized cell-derivative factor for these self-image
pairs is `0.25`.

Updated `REFERENCE_CLI` comparison after the native tblite H0 sigma fix:

| Case | SCF | Energy diff | Gradient diff sum/max | Virial diff sum/max |
| --- | ---: | ---: | ---: | ---: |
| CH2O/GFN1, molecular | 12 | `6.775398020409e-10` | `3.296429618975e-08` / `1.577689361792e-08` | `2.884140131919e-08` / `2.625531908551e-08` |
| CH2O/GFN2, molecular | 15 | `6.651852402229e-10` | `1.437665769393e-07` / `6.050656990025e-08` | `1.240845506761e-07` / `8.358997352631e-08` |
| Ar-fcc/GFN1, periodic | 7 | `3.595097552989e-09` | `1.125627343711e-09` / `3.688615600028e-10` | `5.441065574157e-09` / `1.653417298474e-09` |
| Ar-fcc/GFN2, periodic | 20 | `8.975906240494e-09` | `2.140900031075e-08` / `4.424293700931e-09` | `3.094086449720e-07` / `8.057020835922e-08` |

`REFERENCE_CLI` is intentionally skipped/guarded for CP2K multi-k-point runs: native `tblite` CLI is
a useful Gamma reference, but it is not a CP2K-equivalent k-point/symmetry implementation.

### DFTB and CP2K-native xTB/GFN0/GFN1

- DFTB SCC/Ewald background handling was fixed so the homogeneous background contributes to energy
  and atomic properties without entering the SCC overlap virial as a constant potential shift.
- The TB real-space Ewald self-image virial prefactor was fixed.
- CP2K-native DFTB/GFN0/GFN1 k-point symmetry support was extended and tested with the CP2K atomic
  backend and SPGLIB backend.
- Full atomic k-point symmetry is used for high-symmetry production geometries; for displaced DEBUG
  geometries the reduced symmetry is recomputed instead of reusing an invalid high-symmetry map.
- DFTB3 forces, Gamma stress, k-point forces, and k-point stress were tested and covered by new
  regtests.

## What was tried

The debugging was deliberately systematic rather than a single-code-path patch:

- molecular and truly periodic crystals were checked separately; large vacuum cells were not used as
  periodic substitutes,
- GFN1 and GFN2 were tested with tight `EPS_DEFAULT`/`EPS_SCF`, DX scans, and high `MAX_SCF`,
- CP2K/tblite results were compared to native `tblite` CLI for energy, gradient, and virial,
- H0, overlap/potential, Coulomb/multipole, CN-chain-rule, D4, D3/s-D3, DFTD4 and SCC/Ewald pieces
  were isolated where practical and checked against finite differences,
- CP2K-native DFTB, DFTB3, GFN0, GFN1, CP2K/tblite GFN1/GFN2/IPEA1, and native `tblite` CLI were
  tested separately,
- k-point full grids, CP2K atomic symmetry, SPGLIB symmetry, displaced atoms, skew cells, and
  MPI-rank reproducibility were checked,
- 1D/2D periodicity, charged systems, and open-shell/spin behavior were probed.

What helped:

- restoring periodic H0/CN self-image contributions,
- fixing self-image virial signs/factors,
- fixing the native tblite H0 self-image sigma factor,
- fixing the DFTB SCC/Ewald background gauge contribution,
- recomputing k-point symmetry for lowered-symmetry DEBUG geometries,
- smoothing the hard D4 two-body cutoff in the stress-sensitive GFN2/Si8 case,
- updating the external library toolchain patches and DFTB reference values.

What did not help or was not the root cause:

- simply tightening `EPS_SCF`, `EPS_DEFAULT`, `DX`, or `MAX_SCF`,
- changing only SCC damping/mixer parameters for the hard GFN2 stress cases,
- treating the native `tblite` CLI periodic virial as authoritative before fixing its own H0 sigma
  self-image term,
- using large vacuum boxes as stand-ins for true periodic crystals,
- assuming all heteronuclear crystals are good regtests: several are dominated by overlap
  conditioning/SCF-state sensitivity rather than a clean force/stress bug.

## What works now

- CP2K/tblite GFN1/GFN2 molecular force DEBUG checks are at `~1e-8`.
- CP2K/tblite GFN2 isolated diagnostic stress is `~4e-9`.
- CP2K/tblite GFN1/GFN2 Ar-fcc k-point force/stress DEBUG runs with `64 -> 4` k-point reduction.
- CP2K/tblite GFN2 `PERIODIC XZ` Ar-layer k-point stress DEBUG is reduced from the reproduced `5.83e-4` remainder to `2.66e-5`; forces are `2.29e-6`.
- CP2K/tblite GFN1/GFN2 skew/asym Si8 Gamma stress is now in the `~1e-6` range rather than
  `~1e-3`.
- CP2K/tblite IPEA1 runs with k-point reduction and force/stress DEBUG coverage.
- CP2K-native DFTB non-SCC/SCC and GFN0/GFN1 k-point symmetry paths pass the new tests.
- DFTB3 force, Gamma stress, k-point force, and k-point stress tests pass.
- DFTD4/DFTD3/D3(BJ) CP2K tests pass with the updated build; external DFTD4/tblite patches are ready
  for the toolchain.

Representative final regtest values:

| Path | Representative result |
| --- | ---: |
| CP2K/tblite GFN2 CH2O force DEBUG | `M042 = 1e-8` |
| CP2K/tblite GFN2 CH2O isolated stress DEBUG | `M042 = 4.067e-9` |
| CP2K/tblite GFN2 Si8 Gamma stress DEBUG | `M042 = 4.76966e-7` |
| CP2K/tblite GFN2 Ar-fcc k-point stress DEBUG | `M042 = 3.33e-7`, `N_special_kpoints = 4` |
| CP2K/tblite GFN2 Ar-layer XZ k-point stress DEBUG | `M042 = 2.66e-5`, `N_special_kpoints = 2` |
| CP2K/tblite GFN1 Ar-fcc k-point stress DEBUG | `M042 = 1.6417e-8`, `N_special_kpoints = 4` |
| CP2K/tblite IPEA1 Ar-fcc k-point stress DEBUG | `M042 = 2.4656e-8`, `N_special_kpoints = 4` |
| CP2K-native GFN0 Si k-point symmetry | `N_special_kpoints = 4`; displaced geometry `12` |
| DFTB non-SCC Si k-point stress | `M042 = 3.807e-9`, `N_special_kpoints = 4` |
| DFTB SCC C k-point stress | `M042 = 5.409e-9` |
| DFTB SCC ZnS k-point stress | `M042 = 2.18113e-7`, `N_special_kpoints = 4` |
| DFTB3 H2O k-point stress | `M042 = 6.89e-10`, `N_special_kpoints = 4` |

## Remaining limitations

- The reproduced `PERIODIC XZ` CP2K/tblite GFN2 Ar-layer k-point virial remainder is now much smaller (`2.66e-5` instead of `5.83e-4`), but not yet in the `1e-6` range; forces are good.
- `PERIODIC X` full stress DEBUG is limited by the general CP2K cell/Poisson/stress machinery, not by
  a TB-specific quick fix.
- CP2K/tblite open-shell/LSD still aborts intentionally:
  `LSD option not compatible with tblite library`.
- Native `tblite` CLI remains a serial external Gamma-reference path; it is not a k-point/symmetry
  reference implementation.
- Some heteronuclear CP2K/tblite crystals with k-points are sensitive to overlap conditioning and
  SCF state; Ar/Si remain the robust PR regtest choices.

## Verification

- formatting:
  - `tools/precommit/format_fortran.py` on changed Fortran files,
  - `tools/precommit/format_input_file.py` on changed/new CP2K inputs,
  - `tools/precommit/precommit.py -m` on non-Fortran/non-input files.
- patch checks:
  - `tblite-0.5.0-gradient-fixes.patch`: `patch -p1 --dry-run` against fresh `tblite-0.5.0`,
  - `dftd4-4.1.0-gradient-fixes.patch`: `patch -p1 --dry-run` against fresh `dftd4-4.1.0`.
- build:
  - `cmake --build build-tblite-mpi-check --target cp2k-bin -- -j 8`
- regtests with `mpiranks=2`, `ompthreads=1`:
  - `xTB/regtest-tblite-gfn2`, `xTB/regtest-tblite-gfn1-grad`,
    `xTB/regtest-tblite-ipea1`, `DFTB/regtest-scc-2`: `49 / 49` correct,
  - `DFTB/regtest-scc`, `DFTB/regtest-nonscc`: after updating affected references,
    `89 / 89` correct,
  - additionally checked in the same PR-ready run: `xTB/regtest-1`, `xTB/regtest-3`,
    `xTB/regtest-gfn0`, `QS/regtest-dft-vdw-corr-4`; those were green, with only the old DFTB
    references requiring the follow-up update above.


Growl1234:
Would you like to share the patches regarding dftd4 and tblite with the upstream repositories?

TBLITE_GFN_GRADIENT_NOTES.md
Growl1234, line 1: And could you please use English for documentation?

DCM-Uni-Paderborn:
Yes. We are in close contact with the upstream developers and plan to make these changes available to the respective upstream projects through that route. In part this has already happened, for example for the related simple-dftd3 changes.

tests/matchers.py
```diff
@@ -72,6 +72,53 @@ def run(self, output: str, **kwargs: Any) -> MatchResult:
         return MatchResult("OK", error=None, value=value)  # passed
 
 
+# ======================================================================================
+class RegexValueMatcher(Matcher):
```
oschuett: This is a lot of code duplication. How about just adding a `regex` flag to the `GenericMatcher`:

```
class GenericMatcher(Matcher):
    def __init__(self, pattern: str, col: int, regex: bool=False):
        ...
```

tests/matchers.py
```diff
@@ -72,6 +72,53 @@ def run(self, output: str, **kwargs: Any) -> MatchResult:
...
+class TextPresenceMatcher(Matcher):
+    def __init__(self, text: str):
+        self.text = text
+
+    def run(self, output: str, **kwargs: Any) -> MatchResult:
+        if self.text not in output:
+            return MatchResult(
+                "WRONG RESULT", f"Text not found: '{self.text}'.\n", value=None
+            )
+        return MatchResult("OK", error=None, value=1.0)
```
oschuett:
```suggestion
        return MatchResult("OK", error=None, value=None)
```

tests/matchers.py
oschuett: ...or alternatively make the `RegexMatcher` the base class and the `GenericMatcher` a derived class:

```python
class GenericMatcher(RegexMatcher):
    def __init__(self, pattern: str, col: int):
        regex = escape_special_chars(pattern) + r'(\s+\S+){' + str(col-1) + r'}\s+(\S+)'
        super().__init__(regex)
```

DCM-Uni-Paderborn: Good point. I addressed this by extending `GenericMatcher` with an optional `regex` flag and removed the separate `RegexValueMatcher`. The two DEBUG matchers now use `GenericMatcher(..., regex=True)`.

tests/matchers.py
DCM-Uni-Paderborn: Agreed, thanks. This matcher only checks for text presence and does not extract a numerical value, so returning `None` is clearer. I changed it to return `MatchResult("OK", error=None, value=None)`.

Growl1234:
@oschuett, could you please upload [dftd4-4.1.1.tar.xz](https://github.com/dftd4/dftd4/releases/download/v4.1.1/dftd4-4.1.1.tar.xz) to CP2K server? Thanks!

mkrack:
https://www.cp2k.org/static/downloads/dftd4-4.1.1.tar.xz

sha256: c8e6388d7d7d748dbcf91117f35aa50108492d4fd2266d60782cf85a16651887

awvwgk:
I am in the process of including the patches upstream:

- [x] https://github.com/dftd3/simple-dftd3/issues/182 & https://github.com/dftd3/simple-dftd3/pull/183
- [x] https://github.com/dftd4/dftd4/issues/361 & https://github.com/dftd4/dftd4/pull/362
- [x] https://github.com/tblite/tblite/pull/326
- [ ] https://github.com/tblite/tblite/issues/325

src/qs_charge_mixing.F
```diff
@@ -104,6 +149,98 @@ SUBROUTINE charge_mixing(mixing_method, mixing_store, charges, para_env, iter_co
 
    END SUBROUTINE charge_mixing
 
+! **************************************************************************************************
+!> \brief Return the tblite SCC-mixer residual on the CP2K EPS_SCF scale.
```
jpoto, line 153: The function re-scaling the error is duplicated. 
It might be better to use / define a different EPS.

src/tblite_interface.F
```diff
@@ -611,9 +754,15 @@ SUBROUTINE build_tblite_matrices(qs_env, calculate_forces)
 
       CALL timeset(routineN, handle)
 
+      mp_test_mode = 0
+#if defined(__TBLITE_DEBUG_DIAGNOSTICS)
```
jpoto, line 758: It seems that there are about 200 different variants for debugging. 
I think it should either be removed from the master branch or some kind of explanation/key for this different debug modes.

src/tblite_interface.F
```diff
@@ -913,44 +1129,115 @@ SUBROUTINE tb_update_charges(qs_env, dft_control, tb, calculate_forces, use_rho)
...
       TYPE(qs_scf_env_type), POINTER                     :: scf_env
+      TYPE(scf_control_type), POINTER                    :: scf_control
       TYPE(xtb_atom_type), POINTER                       :: xtb_kind
 
       ! also compute multipoles needed by GFN2
       do_dipole = .FALSE.
       do_quadrupole = .FALSE.
+      mp_test_mode = 0
+#if defined(__TBLITE_DEBUG_DIAGNOSTICS)
+      CALL GET_ENVIRONMENT_VARIABLE("CP2K_TBLITE_MULTIPOLE_TEST", debug_value, STATUS=debug_status)
```
jpoto, line 1178: The skip_scf_dispersion... is also used only for debugging. 
It does not make the code as unreadable as  mp_test_mode, but it could be an idea to reduce it to one debug variable.

DCM-Uni-Paderborn: Addressed in #5841: the component-specific energy, potential, and gradient switches are consolidated into the single developer-only `CP2K_TBLITE_DEBUG_SKIP_SCF_DISPERSION` switch.

src/tblite_interface.F
DCM-Uni-Paderborn: The numbered `mp_test_mode` infrastructure was subsequently removed from `master` in #5169.

src/qs_charge_mixing.F
DCM-Uni-Paderborn: Addressed in #5841: both tblite residual paths now use one shared conversion helper and one shared convergence reference. I retained the common CP2K `EPS_SCF` scale because both residuals feed the same `iter_delta` convergence measure.

TBLITE_GFN_GRADIENT_NOTES.md
DCM-Uni-Paderborn: The temporary notes file was removed before merge and is not present on `master`.

PR #5139, opened 2026-05-06
Closed PR

DCM-Uni-Paderborn:


oschuett:
This looks weird! What's going on with the new `gxtb` directory???

DCM-Uni-Paderborn:
Closing this PR because it includes g-XTB follow-up work. The intended follow-up should be scoped to GFN2 only; I will prepare a separate clean GFN2-only branch/PR from current master if there is a remaining GFN2-only delta.

PR #5142, opened 2026-05-06
Closed PR

DCM-Uni-Paderborn:

PR #5143, opened 2026-05-06
Update DFTD4/simple-dftd3 toolchain patches

DCM-Uni-Paderborn:
## Summary
This PR refreshes the external dispersion-library patches used by the CP2K toolchain and adapts the CP2K DFTD4 interface to the DFTD4 4.2 API.

The changes are part of the follow-up work on CP2K/tblite GFN2 force and stress consistency. They build on the CP2K/tblite interface work by Johann Pototschnig and on fixes developed in close contact with the tblite/DFTD4/simple-dftd3 upstream developers, including Sebastian Ehlert.

## Changes
- Update standalone DFTD4 in the toolchain from 4.1.1 to 4.2.0.
- Update the simple-dftd3 dependency used by tblite from 1.3.2 to 1.4.0.
- Replace the old DFTD4 4.1.1 and simple-dftd3 1.3.2 gradient/stress patches with patches against the new upstream releases.
- Add the __DFTD4_V4_2 toolchain define and expose it in cp2k_info.
- Adapt the CP2K DFTD4 interface to pass the new smooth-cutoff widths introduced by DFTD4 4.2.
- Add a non-enabled Si8/GFN2/tblite low-symmetry stress DEBUG input as a diagnostic case for future follow-up work.

## Notes
- DFTD4 4.2.0 and simple-dftd3 1.4.0 already contain upstream smooth-cutoff infrastructure. The CP2K toolchain patches keep the intended narrow two-body smoothing default and carry the OpenMP reduction cleanups.
- Upstreaming of the external-library patches is planned separately and is being coordinated with the respective upstream projects.
- This replaces closed PR #5142 with a clean branch containing only the original toolchain/library-patch state, without the unrelated g-xTB commits.

## Checks
- git diff --check master...HEAD


Growl1234:
Is there any way to not use environment variables? It's definitely best to set it directly when doing calculation (for example, input keywords), and it seems the upstream developer of dftd4 and tblite doesn't like this approach either.

oschuett:
Agreed, we really should not rely on environment variables for input. Hopefully, we can [remove this](https://github.com/cp2k/cp2k/blob/ce370e297b8ae825b8b0c764ddc9ca52aa607b42/src/qs_dispersion_d4.F#L290-L299) once all the patches have been up streamed and the dust settled.

src/qs_dispersion_d4.F
```diff
@@ -254,6 +287,16 @@ SUBROUTINE calculate_dispersion_d4_pairpot(qs_env, dispersion_env, evdw, calcula
       IF (cutoff%disp3 > cutoff%disp2) THEN
          CPABORT("D4: Three-body cutoff should be smaller than two-body cutoff")
       END IF
+#if defined(__DFTD4_V4_2)
+      ! DFTD4 4.2 introduced explicit smooth-cutoff widths.
+      ! Keep the same narrow two-body smoothing used by the toolchain patch.
+      cutoff%width2 = 0.05_dp
+      cutoff%width3 = 0.0_dp
```
awvwgk, line 294: Should both of these be 0.05?

DCM-Uni-Paderborn: No, the different defaults are intentional. The stress discontinuity investigated in #5131 was isolated to the two-body D4 term, so its narrow `0.05` bohr smoothing remains enabled. Three-body smoothing stays disabled by default to preserve the previous behavior, but #5842 exposes both widths as regular CP2K input keywords.

DCM-Uni-Paderborn:
Following up on the environment-variable discussion: now that the smooth-cutoff API is upstream and the toolchain patches are gone, I opened #5842 to remove those overrides from the CP2K D4 interface. The widths are instead controlled by `D4_CUTOFF_WIDTH` / `D4_2B_CUTOFF_WIDTH` and `D4_3B_CUTOFF_WIDTH` in `&PAIR_POTENTIAL`, with the current numerical defaults preserved.

PR #5144, opened 2026-05-07
Improve DFTB/xTB stress handling, k-point reductions, and UKS coverage

DCM-Uni-Paderborn:
## Summary

This PR improves force, stress/virial, k-point, and spin-polarized coverage for the DFTB/xTB/tblite paths.

The main focus is on making stress handling more robust for molecular, fully periodic, and reduced-periodic systems, including k-point calculations with symmetry reduction. It also adds UKS/LSD force support for tblite-backed GFN1/GFN2 calculations and extends regression coverage across the related DFTB/xTB methods.

## Changes

- Fix stress/virial handling in direct TB and tblite-backed paths.
- Improve tblite k-point stress contributions, including self-image handling.
- Refine k-point symmetry reduction and add checks for the number of special k-points.
- Add stress/force debug coverage for 1D, 2D, and 3D periodic systems.
- Add deformed Si8 GFN2/tblite finite-difference and k-point stress diagnostics.
- Enable tblite LSD/UKS force support for GFN1/GFN2 by handling charge and magnetization channels consistently.
- Add RKS/UKS force regression coverage for DFTB, DFTB3, GFN0, internal GFN1, tblite GFN1/GFN2, and IPEA1.

Spin-polarized tblite support is enabled for the native tblite SCC path. The CP2K-side tblite SCC mixing path remains restricted for spin-polarized calculations.

## Testing

Built CP2K successfully after merging current `master`.

Focused tblite regression tests:

- `xTB/regtest-tblite-gfn1-grad`: OK
- `xTB/regtest-tblite-gfn1-periodic`: OK
- `xTB/regtest-tblite-gfn2`: OK
- `xTB/regtest-tblite-ipea1`: OK

Combined focused tblite run:

- `82 / 82` tests passed

Additional branch verification:

- `DFTB/regtest-scc`: `64 / 64` OK
- `DFTB/regtest-scc-2`: `22 / 22` OK
- `xTB/regtest-gfn0`: `43 / 43` OK
- `xTB/regtest-gfn1-d`: `16 / 16` OK
- `xTB/regtest-tblite-gfn2`: `42 / 42` OK
- `xTB/regtest-tblite-gfn1-grad`: `21 / 21` OK


oschuett:
Since this is another major change, should we maybe first fix the issues found by the [dashboard](https://dashboard.cp2k.org)?

src/tblite_interface.F
```diff
@@ -229,41 +294,101 @@ END SUBROUTINE tb_update_geometry
...
+      INTEGER                                            :: nSpin
 
       TYPE(scf_info)                                     :: info
 
+      nSpin = dft_control%nspins
+      IF (nSpin /= 1 .AND. nSpin /= 2) CPABORT("tblite supports only one or two spin channels")
+
+      tb%mol%charge = dft_control%charge
+      tb%mol%uhf = MAX(0, dft_control%multiplicity - 1)
+      IF (nSpin == 1) tb%mol%uhf = 0
```
awvwgk: tblite can still use the number of unpaired electrons in a restricted open-shell calculation even if spin-polarization / unrestricted calculation is not enabled.

DCM-Uni-Paderborn: Dear Sebastian,

Good point, thanks. I removed the `nSpin == 1` reset so `tb%mol%uhf` now preserves the number of unpaired electrons from `MULTIPLICITY` also for restricted open-shell calculations.

The spin-polarization interaction is still only added for the unrestricted/two-spin-channel path (`nSpin == 2`), so restricted open-shell and spin-polarized calculations remain distinct.

I checked this with a restricted O2/GFN2 smoke test (`MULTIPLICITY 3` without `LSD`) and reran `xTB/regtest-tblite-gfn2` locally: 42/42 OK.

DCM-Uni-Paderborn:
Ciao Ole,

thanks for pointing me to the dashboard. I traced this back through the branch history: the failing case was not the original CP2K-internal GFN1 ch2o.inp path, which still converges and matches the reference. The problematic case was the new ch2o_scc_mixer_tblite.inp test introduced in this branch, which forced SCC_MIXER TBLITE for the CP2K-internal GFN1 implementation. I removed this unstable experimental regtest from the regular suite and kept the stable default/CP2K-mixer CH2O coverage.

Growl1234:
Does this solve the problem shown in my below tests (they are run without this PR applied)?
[tests.zip](https://github.com/user-attachments/files/27476810/tests.zip)

DCM-Uni-Paderborn:
> Does this solve the problem shown in my below tests (they are run without this PR applied)? [tests.zip](https://github.com/user-attachments/files/27476810/tests.zip)

Hi Growl1234?

thanks, these tests are very useful. I looked at them, and the short answer is: no this is an unrelated issue. 

In fact, the attached case seems to expose two separate issues. For the SPGLIB backend, the initial fixed-geometry energy agrees with the non-symmetry reference to numerical precision, so the SPGLIB atomic k-point operations themselves look fine for a static calculation. The failure happens during CELL_OPT: the input has KEEP_SYMMETRY F and KEEP_SPACE_GROUP F, so after the first cell/geometry update the initially detected atomic symmetry is no longer guaranteed to be valid, but the k-point symmetry mapping is still being used.

For the K290 backend there is an additional problem: in this conventional diamond/non-primitive case the static starting energy is already wrong compared to the full/reference grid, despite generating the same irreducible k-point set. So this needs a separate guard or fix for K290.

The proper fix should therefore be to restrict full atomic k-point symmetry automatically for geometry-changing runs unless the symmetry is actively preserved and revalidated. In the conservative path CP2K should fall back to inversion/time-reversal symmetry rather than silently producing wrong stresses/energies. I verified that this inversion-only path lets your CELL_OPT case run through consistently with the non-symmetry reference.

I will add a targeted guard/test for this failure mode. Full SPGLIB symmetry during CELL_OPT can then be enabled more generally once the symmetry is preserved/revalidated at each updated structure.

Growl1234:
@DCM-Uni-Paderborn Thanks for your analysis!

PR #5157, opened 2026-05-08
Real Space Resolution of Identity for GW Calculations

RitajTyagi:
This PR implements the Real Space Resolution of Identity (RI-RS) scheme for $GW$ based on the separable RI implementation of Duchemin and Blase \[JCP 150, 174120 (2019)\]. Benchmarks show high accuracy, with deviations within 1--2 meV for IP and EA across the GW100 and ACC24 datasets. For systems with ~3000 basis functions, we observe significant speedups: $\Sigma^C$ and $\Sigma^X$ calculation times are reduced by factors of more than $100\times$ compared to the existing resolution-of-identity implementation with a truncated Coulomb operator. The current implementation scales efficiently to systems with thousands of basis functions. The implementation is currently validated for systems containing atoms up to the 4th row of the periodic table.

### Future and Ongoing Work

- Generation of optimized grid points for additional atoms and basis functions.
- Implementation of periodic boundary conditions (PBC) and full k-point sampling.
- Further refinements for computational efficiency.

### Other Changes

- Resolve a bug in the use of contracted auxiliary basis sets for $GW$ calculations.
- Modify the write statement for $GW$ band-structure output in spin-polarized systems.


Growl1234:
It seems this has caused [compilation failure](https://storage.googleapis.com/cp2k-ci/dashboard_sdbg_report.txt) when building serial versions of CP2K (ssmp/sdbg):
```
/opt/cp2k/src/gw_large_cell_gamma_ri_rs.F:50:8:

   50 |    USE mpi,                             ONLY: MPI_COMM_RANK,&
      |        1
Fatal Error: Cannot open module file 'mpi.mod' for reading at (1): No such file or directory
```
I think we should use the mpiwrapper in CP2K if possible instead of directly using MPI, and it should be well encapsulated.

RitajTyagi:
Thanks for catching that, @Growl1234 , and thanks for the quick fix, @oschuett ! I'll keep the preference for para_env / mpiwrapper in mind for future updates.

PR #5159, opened 2026-05-09
Use para_env instead of MPI in gw_large_cell_gamma_ri_rs.F

oschuett:
Follow up to #5157


JWilhelm:
thanks for spotting this, Ole.

PR #5161, opened 2026-05-09
Add specific Gamma-centered shifting in Monkhorst-Pack scheme

Growl1234:


src/input_cp2k_kpoints.F
```diff
@@ -99,6 +99,14 @@ SUBROUTINE create_kpoints_section(section)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
+      CALL keyword_create(keyword, __LOCATION__, name="GAMMA_CENTERED", &
+                          description="Generate a gamma-centered variant of the "// &
```
he-zilong, line 103: This needs more explanation, that it always include gamma-point in the mesh by introducing a shift at the same number of subdivisions, that it is essentially effective only for when an even number of subdivisions is used (in one or more directions?), etc.

he-zilong (commented):
Thank you for implementing this idea from #4854 .

he-zilong:
Some follow-up works may be necessary once this PR is merged. @DCM-Uni-Paderborn could you please make a plan to test whether this new `GAMMA_CENTERED` option works well with the k-point symmetry reduction backends `K290`/`SPGLIB` on hexagonal/face-centered cubic/face-centered orthorhombic lattices?

DCM-Uni-Paderborn:
Dear Zilong,

I tested this locally against full-grid reference calculations using the PR branch on top of current master.

The short answer is: SPGLIB works well in the tested cases, while K290 still has a problem for one gamma-centered hexagonal case. I checked:

- hexagonal lattice, 4x4x2, `GAMMA_CENTERED T`
- face-centered cubic lattice, 4x4x4, `GAMMA_CENTERED T`
- face-centered orthorhombic lattice, 4x4x4, `GAMMA_CENTERED T`

For fcc and face-centered orthorhombic lattices, both K290 and SPGLIB reproduced the full-grid energy to numerical precision. For the hexagonal case, SPGLIB also reproduced the full-grid result, but K290 did not: it reduced the gamma-centered 4x4x2 mesh to 8 special k-points instead of the 20 found by SPGLIB, and the total energy differed from the full-grid reference by about 0.21 Ha.

As a control, the same hexagonal 4x4x2 case without `GAMMA_CENTERED` works correctly with both K290 and SPGLIB. So the issue appears specific to the combination of K290 + hexagonal lattice + even gamma-centered Monkhorst-Pack mesh.

I will try to fix this K290-specific issue separately. Since SPGLIB behaves correctly in these tests and the problem seems to be in the older K290 reduction path rather than in the new option itself, I do not want to block this PR on that K290 follow-up.

DCM-Uni-Paderborn (approved):

DCM-Uni-Paderborn:
I found a fix, I will soon merge this PR and create a follow-up PR in due course including my fixes of this issue and other ones ...

PR #5165, opened 2026-05-10
Improve k-point symmetry for Gamma-centered meshes and moving geometries

DCM-Uni-Paderborn:
## Summary

This PR improves the k-point symmetry infrastructure in several related areas:

- enable `GAMMA_CENTERED` also for `SCHEME MACDONALD`
- make Gamma-centered Monkhorst-Pack and MacDonald meshes work with both `SYMMETRY_BACKEND K290` and `SYMMETRY_BACKEND SPGLIB`
- resolve the K290-side issue observed in Growl1234's `GAMMA_CENTERED` PR (#5161)
- fix the AO basis rotation convention used for atomic k-point symmetry operations
- refresh atomic k-point symmetry dynamically for moving geometries such as `GEO_OPT`, `CELL_OPT`, `MD`, `DRIVER`, and `EHRENFEST`, addressing the moving-geometry issue reported by Growl1234 in PR #5144
- keep unsupported method combinations guarded conservatively
- add expert keyword `DEBUG_FULL_KPOINT_SYMMETRY` for using full atomic k-point symmetry in DEBUG finite-difference points
- improve handling/documentation around `GENERAL` k-point input and symmetry reduction
- add regression tests for Gamma-centered K290/SPGLIB, MacDonald meshes, GEO_OPT, DEBUG, and HFX RI symmetry coverage

## Details

The K290 path previously had trouble with Gamma-centered meshes, especially for non-trivial rotations and basis functions beyond s-only cases. The fix makes the phase/cell-shift convention and AO basis rotation consistent, and allows the shifted mesh to be reduced after the K290 symmetry operations have been constructed.

This also addresses the K290-side issue observed during review/testing of Growl1234's `GAMMA_CENTERED` PR (#5161): Gamma-centered meshes can now be reduced consistently with K290 for the tested Monkhorst-Pack and MacDonald cases, with energies matching the corresponding SPGLIB/backend references.

For moving geometries, CP2K no longer reuses atomic k-point symmetry from a previous geometry. Instead, the k-point symmetry information is rebuilt from the current cell and atomic positions whenever full atomic symmetry is active. If the selected method is not yet safe for full atomic k-point symmetry, the existing conservative guards still apply.

This dynamic refresh addresses the moving-geometry problem reported by Growl1234 in [PR #5144](https://github.com/cp2k/cp2k/pull/5144#issuecomment-4396591099). The attached tests compared no symmetry against `SYMMETRY_BACKEND K290` and `SYMMETRY_BACKEND SPGLIB` for a `RUN_TYPE CELL_OPT` / `TYPE DIRECT_CELL_OPT` calculation with `KEEP_SYMMETRY F`, where the cell changes during the run. Rebuilding the k-point symmetry from the current cell and atomic positions avoids carrying over an irreducible k-point mapping from an earlier geometry. The same refresh path covers `GEO_OPT` and the other moving-geometry run types listed above.

For `DEBUG`, the default remains conservative: finite-difference energy points use inversion/time-reversal only, avoiding comparisons between slightly different irreducible k-point sets. The new expert keyword

```text
DEBUG_FULL_KPOINT_SYMMETRY T
```

allows full dynamic symmetry in DEBUG finite-difference points when explicitly requested.

## Testing

Local checks:

```text
./make_pretty.sh
git diff --check
tools/regtesting/check_inputs.py build-serial-libint/bin/cp2k.ssmp tests/QS/regtest-kp-1
cmake --build build-serial-libint --target cp2k.ssmp -j 8
```

Selected local smoke/regression tests:

```text
h_hex_c3_gamma_k290.inp
  Number of Special K-points: 8
  E = -9.413339673529373

h_hex_c3_gamma_spglib.inp
  Number of Special K-points: 8
  E = -9.413339673529396

h_hex_c3_gamma_macdonald_k290.inp
  Number of Special K-points: 12
  E = -9.413752006495693

h_hex_c3_gamma_macdonald_spglib.inp
  Number of Special K-points: 12
  E = -9.413752006495718

c_debug_kpsym_default.inp
  DEBUG finite-difference points remain conservative
  final Number of Special K-points: 4

c_debug_kpsym_full.inp
  DEBUG_FULL_KPOINT_SYMMETRY T enables full symmetry
  final Number of Special K-points: 3
```

Additional tested cases include GEO_OPT symmetry refresh, GAPW/SPGLIB coverage, HFX RI k-point symmetry, and guarded DFTB cases.


oschuett:
There has been again a lot of fall out on the [dashboard](https://dashboard.cp2k.org).

To recover all the Spack tests, the [patches](https://github.com/cp2k/cp2k/tree/master/tools/toolchain/scripts/stage8) can be added to our [dev repository](https://github.com/cp2k/cp2k/blob/master/tools/spack/spack_repo/cp2k_dev/) while they are being upstreamed.

Growl1234:
> There has been again a lot of fall out on the [dashboard](https://dashboard.cp2k.org).
> 
> To recover all the Spack tests, the [patches](https://github.com/cp2k/cp2k/tree/master/tools/toolchain/scripts/stage8) can be added to our [dev repository](https://github.com/cp2k/cp2k/blob/master/tools/spack/spack_repo/cp2k_dev/) while they are being upstreamed.

I think, a bit more regtests should be added as "Required", such as "Regtest sdbg" and "Spack psmp" (not pdbg because it's too slow).

oschuett:
> I think, a bit more regtests should be added as "Required", such as "Regtest sdbg" and "Spack psmp" 

Yes, requiring "Spack psmp" and "Conventions" would have helped, but our cloud bill is already 3x compared to last year. So, I'll first need to run the numbers to ensure that we're not breaking the bank.

I think, there is also a more broader discussion to be had about how to scale our development process to better accommodate new contributors and LLMs. Perhaps we should open an RFC about that topic?

Growl1234:
Looking forward to the discussion.

Issue #5129, opened 2026-05-04
undefined reference functional in D4 leads to error


jpoto:
If the reference functional is not in the list available by there should be some kind of check/message. 

Right now there is an error:
https://groups.google.com/g/cp2k/c/tGahetE-f50

he-zilong:
Observation: CP2K uses the `get_rational_dumping` interface from DFT-D4
https://github.com/cp2k/cp2k/blob/4733e3f172294034c9e0fac46f15fd3f12116880/src/qs_dispersion_d4.F#L226-L246
which is defined in `dftd4/src/dftd4/param.f90` and eventually utilizes [`get_functional_id`](https://github.com/dftd4/dftd4/blob/c7a0062c1d72661c7e3b6a28378eaf9caf67cbd9/src/dftd4/param.f90#L825) for parsing names. Presumably this should be consistent with the [`dftd4/assets/parameters.toml`](https://github.com/dftd4/dftd4/blob/main/assets/parameters.toml) file where parameters are stored.

It is worth noting that the [ORCA 6.1.1 documentation on Dispersion Corrections](https://www.faccts.de/docs/orca/6.1/manual/contents/modelchemistries/dispersioncorrections.html) has several comprehensive lists for the available functional and parameters, and in the spirit of #4972 this could be done for the CP2K manual as well.

@Growl1234 Could you share your thoughts on this matter?

Growl1234:
IIRC, whether `D4_REFERENCE_CODE` is T or F should share the same logic of finally calculating D4 energy correction, so I'm somewhat curious of why `r2SCAN_LIBXC` works for `D4_REFERENCE_CODE F`. Anyways, a proper abort message is indeed missing.

@he-zilong you're right, it should be considered to add them to the manual.

awvwgk:
CP2K specific functional names can be added as alias in D4 if needed.

jpoto:
The error message is now in the master. (#5168)

PR #5168, opened 2026-05-11
Jvp terok build

jpoto:
Adressing #5129


jpoto:
I have introduced tests to catch the error, but not a conversion list.

DCM-Uni-Paderborn:
How about enabling LibXC IDs as a unique identifier of XC functionals (see https://libxc.gitlab.io/functionals/) in addition/beside parsing the param.f90 file?

jpoto:
I would merge this first. 

The dftd4 function can handle some standard libxc names. 
https://github.com/dftd4/dftd4/blob/873ffd3b5ba8436d4b7a011a5d4cf6cc7d103ff8/test/unit/test_param.f90#L163

I started looking into converting CP2K functional names to one of them.

PR #5170, opened 2026-05-11
Closed PR

DCM-Uni-Paderborn:


he-zilong:
The case is very intriguing! Apparently [several regtests](https://storage.googleapis.com/cp2k-ci/run-cp2k-pdbg-8f0feea7_report.txt) perform k-point calculation on a non-conventional cell matrix; if they have been introduced much earlier and was working at that time, then they should catch any offending revisions in some recent PR...

> The intended user-facing workaround is to rotate the cell into CP2K convention or specify the cell through `ABC` and `ALPHA_BETA_GAMMA`.

Or use an external `CIF`. 

Edit: some failing regtests uses only the keyword `ABC` for cell parameters, which should generate a diagonal cell matrix ("numerically orthorhombic" as CP2K says; the implicit keyword `ALPHA_BETA_GAMMA` is by default three 90.0 degrees) and should of course be lower-triangular. For example, `tests/xTB/regtest-tblite-gfn1-grad/Ar_fcc_gfn1_kp_spglib_backend_debug.inp` has
```
    &CELL
      ABC 5.2600000000 5.2600000000 5.2600000000
      PERIODIC XYZ
    &END CELL
```

The parameter `REAL(KIND=dp), PARAMETER :: eps_cell = 100.0_dp*EPSILON(1.0_dp)` introduced in 8ff5174 may be smaller than the existing one for parsing and initializing cell in `cell_methods.F`:
https://github.com/cp2k/cp2k/blob/bec0900f9466404e2e80a01ae430705a70231d65/src/cell_methods.F#L129

Growl1234:
I still don't get the point why we must ban k-points calculation with non-conventional cell? A non-canonical cell orientation is not unusual in practice. Users do not always construct cells through the `ABC` / `ALPHA_BETA_GAMMA` input form, and external tools that generate CP2K inputs, such as Multiwfn or structure-conversion workflows, may provide explicit `A`, `B`, and `C` vectors that are physically valid but do not follow CP2K's internal canonical convention. From the user's perspective, a merely rotated primitive cell should not make k-point sampling fundamentally invalid. The real issue should be identified more precisely and fixed or done a fallback.

DCM-Uni-Paderborn:
> I still don't get the point why we must ban k-points calculation with non-conventional cell? A non-canonical cell orientation is not unusual in practice. Users do not always construct cells through the `ABC` / `ALPHA_BETA_GAMMA` input form, and external tools that generate CP2K inputs, such as Multiwfn or structure-conversion workflows, may provide explicit `A`, `B`, and `C` vectors that are physically valid but do not follow CP2K's internal canonical convention. From the user's perspective, a merely rotated primitive cell should not make k-point sampling fundamentally invalid. The real issue should be identified more precisely and fixed or done a fallback.

I know my intention was to address this in two steps ...

DCM-Uni-Paderborn:
Thanks for the comments. I reworked this PR away from the initial hard-abort approach: non-canonical but physically valid cells are now canonicalized internally for DFT k-point calculations, with the coordinates transformed consistently, instead of asking users to rotate the input cell. I also aligned the new tolerance with the existing cell-initialization scale, which should avoid the spurious `ABC` false positives raised above. The updated PR description has the details and the local tests.

he-zilong:
Whoa, whoa, this is getting a too far. Despite the program convention, I do not think there was a problem with a static k-point band calculation previously, as I have compared [in a comment of #5123](https://github.com/cp2k/cp2k/pull/5123#issuecomment-4426411880) with a standard and a non-standard cell setup. I agree with Growl1234: the situation is more likely that some problematic code has been introduced these days and should be identified and fixed.

Out of the context of a discovered program inconsistency, and even regardless of k-points and symmetries, the cell and coordinate transformation alone from non-standard to standard is a major change worthy of an independent PR for discussion. Same for the new writeup in `docs/methods/dft/k-points.md`.

DCM-Uni-Paderborn:
> Whoa, whoa, this is getting a too far. Despite the program convention, I do not think there was a problem with a static k-point band calculation previously, as I have compared [in a comment of #5123](https://github.com/cp2k/cp2k/pull/5123#issuecomment-4426411880) with a standard and a non-standard cell setup. I agree with Growl1234: the situation is more likely that some problematic code has been introduced these days and should be identified and fixed.
> 
> Out of the context of a discovered program inconsistency, and even regardless of k-points and symmetries, the cell and coordinate transformation alone from non-standard to standard is a major change worthy of an independent PR for discussion. Same for the new writeup in `docs/methods/dft/k-points.md`.



> Whoa, whoa, this is getting a too far. Despite the program convention, I do not think there was a problem with a static k-point band calculation previously, as I have compared [in a comment of #5123](https://github.com/cp2k/cp2k/pull/5123#issuecomment-4426411880) with a standard and a non-standard cell setup. I agree with Growl1234: the situation is more likely that some problematic code has been introduced these days and should be identified and fixed.
> 
> Out of the context of a discovered program inconsistency, and even regardless of k-points and symmetries, the cell and coordinate transformation alone from non-standard to standard is a major change worthy of an independent PR for discussion. Same for the new writeup in `docs/methods/dft/k-points.md`.

Dear He Zilong, or is it Zilong He? 

I agree! I will now cut this PR to solely address the issue raised in PR #5123. As I have seen only now there is a reference to your PR #5140, which I try to consider for you to review and test. All the remaining modifications, in particular SCHEME GENERAL, will then be placed in a separate follow-up PR.

DCM-Uni-Paderborn:
Thanks again for the comments. I have now cut this PR back to the focused non-canonical-cell issue from the #5123 discussion. The GENERAL k-point symmetry work, K290/SPGLIB combinations, and the longer k-point documentation are kept out of this PR and will be handled separately. The updated version adds an AlAs-like FCC standard-vs-explicit-A/B/C regression pair; the two local runs agree to 3.9e-11 Ha. I also adjusted the symmetry setup so that KPOINTS%SYMMETRY OFF does not probe SPGLIB, which keeps this path usable when SPGLIB is not linked.

DCM-Uni-Paderborn:
I checked the CI history carefully. The earlier failure on the broader intermediate version (`a696e72`) was from `h_noncanonical_cell_spglib_sym_red.inp`, which belonged to the removed GENERAL/SPGLIB scope and is no longer part of this PR.

The current focused version then exposed a separate pdbg LeakSanitizer issue: `topology%cell_muc` can alias `topology%cell` through `cell_retain`, and my canonicalization path was nullifying that alias instead of releasing its reference before replacing the topology cell. I fixed that reference-counting bug in the latest push (`8db92be`). The AlAs standard/non-canonical local checks still agree to 3.9e-11 Ha.

Growl1234 (approved):

JWilhelm:
Hi Thomas, just for me to understand: Do you think Bibek can run his 3d bulk crystal test set again with arbitrary cells on your latest PR?

he-zilong:
Sorry if this PR would get suspended for a while, but I am trying to do `git bisect` between 223025a and 59927b9 to pinpoint when the behavior diverges for the different cell setup as mentioned in #5123, among other real-life affairs. I believe that this PR has not addressed the root issue yet if the solution is to "rotate explicit non-canonical cells into CP2K's internal convention".
Edit 1: 55d01f6 is still as good as 223025a
Edit 2: same as 495eafe

he-zilong:
Dear Thomas @DCM-Uni-Paderborn ,

After many unsuccessful attempts at reproducing the problem reported by Bibek in #5123 as I mentioned [in a comment](https://github.com/cp2k/cp2k/pull/5123#issuecomment-4433028404), I have now become somewhat doubtful about this PR. I see its entire purpose is to prevent non-conventional cells from entering a k-point band calculation, but this may be a moot point if no problem between them exists in the first place. We need more convincing evidence like reproducibly erroneous cases and/or scientific arguments (that the former normal-termination results are unreliable) to justify impactful changes like in the current PR, which unfortunately is lacking at the moment. I blame myself for coming to a hasty conclusion that inspired this idea.

Please reconsider if this PR is necessary still, and sorry for wasting your time.

Regards,
Zilong (this is pinyin of my given name, by the way)

DCM-Uni-Paderborn:
> Dear Thomas @DCM-Uni-Paderborn ,
> 
> After many unsuccessful attempts at reproducing the problem reported by Bibek in #5123 as I mentioned [in a comment](https://github.com/cp2k/cp2k/pull/5123#issuecomment-4433028404), I have now become somewhat doubtful about this PR. I see its entire purpose is to prevent non-conventional cells from entering a k-point band calculation, but this may be a moot point if no problem between them exists in the first place. We need more convincing evidence like reproducibly erroneous cases and/or scientific arguments (that the former normal-termination results are unreliable) to justify impactful changes like in the current PR, which unfortunately is lacking at the moment. I blame myself for coming to a hasty conclusion that inspired this idea.
> 
> Please reconsider if this PR is necessary still, and sorry for wasting your time.
> 
> Regards, Zilong (this is pinyin of my given name, by the way)

Dear Zilong, 

I will then close this PR and add the little what is left (crys_sym_gen(..., use_spglib=...) plus use_spglib=kpoint%symmetry) to #5173 and rebase it ... 

Greetings, 
Thomas

PR #5174, opened 2026-05-12
try to autodetect functional for d4

jpoto:


jpoto:
I would need to test how many functionals are covered. 
See #5129

PR #5178, opened 2026-05-12
automatic detection of functional for d4

jpoto:


he-zilong:
Dear Johann,

Could this auto-detection mechanism be generalized to other types in `&VDW_POTENTIAL/&PAIR_POTENTIAL/TYPE`, and explained clearly in the manual with a list of the recognized functionals?

Regards,
Zilong

jpoto:
It should be no problem to extent it. 

I would leave it as it is right now. If it can pick up the functional it uses it, otherwise an error is reminding the user to set it.

src/qs_dispersion_utils.F
```diff
@@ -43,11 +45,157 @@ MODULE qs_dispersion_utils
...
+      END DO
+
+      IF (has_becke88 .AND. has_lyp .AND. has_vwn .AND. has_xalpha) THEN
+         xc_functional_name = "B3LYP"
+      ELSE IF (has_becke88 .AND. has_lyp) THEN
+         xc_functional_name = "BLYP"
+      ELSE IF (has_becke88 .AND. has_p86c) THEN
+         xc_functional_name = "BP"
+      ELSE IF (has_optx .AND. has_lyp) THEN
+         xc_functional_name = "OLYP"
```
awvwgk, line 184: D4 knows about olyp:
https://github.com/dftd4/dftd4/blob/873ffd3b5ba8436d4b7a011a5d4cf6cc7d103ff8/assets/parameters.toml#L254-L256

src/qs_dispersion_utils.F
```diff
@@ -43,11 +45,157 @@ MODULE qs_dispersion_utils
...
+      CASE ("BLYP")
+         d4_functional_name = "blyp"
+      CASE ("BP")
+         d4_functional_name = "bp"
+      CASE ("TPSS")
+         d4_functional_name = "tpss"
+      CASE ("SCAN")
+         d4_functional_name = "scan"
+      CASE ("R2SCAN")
+         d4_functional_name = "r2scan"
```
awvwgk, line 77: For all those functionals the name used in D4 is identical, to avoid typos why not copy over the `xc_functional_name` directly?


```suggestion
      CASE ("PBE", "PBE0", "B3LYP", "BLYP", "BP", "TPSS", "SCAN", "R2SCAN")
         d4_functional_name = TRIM(ADJUSTL(xc_functional_name))
```

jpoto: I make a new merge request

src/qs_dispersion_utils.F
jpoto: yes. it is to identify it from some CP2K parameter

PR #5192, opened 2026-05-14
Regtests for TDDFPT Oscillator Strength

juerghutter:
This commit prepares for a major refactoring of TDDFPT oscillator strength planned in a follow up PR.

Mathers for TDDFPT excitation energy and oscillator strength check sums have been added.


oschuett:
There are already two matchers with very similar names. Could they maybe be unified?

https://github.com/cp2k/cp2k/blob/91e51cab28f301bb6a63c03862597245ce4e647e/tests/matchers.py#L180-L181

juerghutter:
I consolidated the markers for TDDFPT energies and oscillator strength.

PR #5198, opened 2026-05-14
Closed PR

DCM-Uni-Paderborn:


juerghutter:
Can you hold back this PR for some days. I'm working on a refactoring of the moments routine.
I would like to rework it in order to avoid references to virtual states and bring the used formulas
in line what is used in other codes and possibly fix an inconsistency in the Berry formula.

PR #5199, opened 2026-05-14
Add SCF_MOMENT dipoles for k-point TDDFPT KERNEL NONE

DCM-Uni-Paderborn:
This PR adds a small, production-oriented follow-up to the existing k-point TDDFPT KERNEL NONE support.

Summary:
- add `DIPOLE_FORM SCF_MOMENT` for k-point TDDFPT,
- compute interband transition dipoles directly in the SCF k-point MO basis,
- use the overlap-derivative / velocity-energy-gap relation for the off-diagonal occupied-virtual matrix elements,
- keep the existing velocity-form path unchanged,
- add a H2 k-point KERNEL NONE regtest using the UZH MOLOPT basis and potentials.

The new path avoids a separate re-diagonalization for the transition moments and uses the SCF MOs already produced for the k-point calculation. Degenerate subspaces remain gauge-dependent at the level of individual transitions, so the meaningful comparison is over folded/degenerate transition blocks rather than isolated states.

KERNEL FULL status:
K-point TDDFPT `KERNEL FULL` is deliberately not enabled or included in this PR. This PR is restricted to the production-oriented `KERNEL NONE` oscillator-strength path.

Testing:
- `./make_pretty.sh`
- `ninja -C build-serial cp2k.ssmp`
- direct runs of:
  - `H2_kp_tddfpt_none_scf_moment.inp`
  - `H2_kp_tddfpt_none_mp.inp`
  - `H2_kp_tddfpt_none_general.inp`
  - `H2_kp_tddfpt_none_sym.inp`


JWilhelm:
Hi Thomas, 

really nice to have k-points in TDDFT! 

By chance, are you planning to put your computation of dipole moments (currently in SUBROUTINE tddfpt_kpoint_independent_particle in qs_tddfpt2_methods.F) to qs_moments.F ? Then also other code parts could use your k-point dipoles. Or do you think this might yield a conflict with the already existing subroutine qs_moment_kpoints ?

DCM-Uni-Paderborn:
Dear Jan,

after having looked into this now, yes, I think moving the reusable part into qs_moments.F
makes sense as a follow-up. It should also be more straightforward than addressing KERNEL
FULL and extending all the k-point machinery there, which I would then leave to the
specialists ... ;)

However, I would not replace qs_moment_kpoints directly, since that routine currently
supports the general moment-printing/Berry-curvature path and re-diagonalizes H(k),
whereas the new TDDFPT SCF_MOMENT path deliberately uses the existing SCF k-point MO
coefficients.

A good compromise would be to add a new helper in qs_moments.F which builds the complex
MO-basis k-point dipole/overlap-derivative matrices from the SCF MOs. TDDFPT could then
call that helper and keep only the TDDFPT-specific transition sorting, occupation logic,
oscillator-strength weighting, and output.

This should avoid a conflict with qs_moment_kpoints while still making the reusable
SCF k-point dipole matrix available to other code paths.

Best,
Thomas

PR #21, opened 2018-10-05
S-ALMO patch

oschuett:
Fermi-Dirac smearing within the ALMO formalism, allowing an ALMO description of metals.
(contains also a quick fix for molecular guess with the diagonalisation algorithm, within ALMO)

Originally submitted as PR #18 .

PR #5206, opened 2026-05-15
Grid gpu engines merge

mtaillefumier:
- Use hip backend for both Nvidia and AMD platforms
- Small performances improvements for large value of l


oschuett:
Thanks a lot for putting this together!

Unfortunately, our GPU tests are in shambles at the moment. I'm working on it...

mtaillefumier:
Large l calculations are dominated by the calculation of the blocks. This operation is a complex version of the scatter-reduce function which in the case at hand can be implemented deterministically because the mapping 
$$
\sum_ij c(b_ij)
$$ is known in advance. 

The original `GPU` and `HIP` backends are on par for small values of `l` and start to diverge significantly at larger angular momentum because of extensive use of `atomicAdd` operations in the integrate function in the GPU backend. The lines https://github.com/cp2k/cp2k/blob/2256cec57518a66cedf6d7165fb4dd3f977fff17/src/grid/gpu/grid_gpu_integrate.cu#L369 or https://github.com/cp2k/cp2k/blob/2256cec57518a66cedf6d7165fb4dd3f977fff17/src/grid/gpu/grid_gpu_integrate.cu#L329 are the most limiting factors in term of performance because the execution time does not scale linearly with the number of atomic operations and the memory controller serialize the operation in a unspecified order. It becomes a serious issue in many cases.

Others constraints such as shared memory occupation or register spills limit the performances at large l as it reduces occupancy. 

Below are some tests using `H2O-64.inp` and the following basis sets and potentials with `MPI=4` `OMP=4` and `GPU=4` (lmax = 7)
```
BASIS_SET_FILE_NAME BASIS_MOLOPT_UZH
POTENTIAL_FILE_NAME POTENTIAL_UZH

&KIND H
      BASIS_SET TZV2P-MOLOPT-SCAN-GTH-q1
      POTENTIAL GTH-SCAN-q1
    &END KIND
    &KIND O
      BASIS_SET TZV2P-MOLOPT-SCAN-GTH-q6
      POTENTIAL GTH-SCAN-q6
&END KIND
```

We get for 

- mi300 

```
HIP/CUDA
CP2K                                 1  1.0    0.537    0.537  106.888  106.888
grid_integrate_task_list           117 12.3    8.036    8.634    8.036    8.634

GPU
CP2K                                 1  1.0    0.457    0.458  146.967  146.967
grid_integrate_task_list           117 12.3   39.418   44.880   39.418   44.880

```

- gh200

```
CUDA
CP2K                                 1  1.0    0.157    0.157   73.388   73.388
grid_integrate_task_list           117 12.3    3.804    3.941    3.804    3.941
grid_collocate_task_list           117  9.6    2.842    2.918    2.842    2.918

HIP/CUDA
CP2K                                 1  1.0    0.249    0.249   73.849   73.849
grid_integrate_task_list           117 12.3    7.166    7.439    7.166    7.439
grid_collocate_task_list  not printed

```

H2O-64.inp is not the best example to show how the two backends behave though. Taking the example from issue #3925. In that case `lmax = 11`. 

```
CPU: 
grid_integrate_task_list            22 10.7   36.338   37.404   36.338   37.404
grid_collocate_task_list            22  8.0   15.482   15.915   15.482   15.915
HIP
grid_integrate_task_list            22 10.7    9.232    9.553    9.232    9.553
grid_collocate_task_list            22  8.0    2.388    2.498    2.388    2.498
GPU
grid_integrate_task_list            22 10.7   46.300   48.451   46.300   48.451
grid_collocate_task_list            22  8.0    3.107    3.247    3.107    3.247

```
The CPU backend will scale almost linearly with the number of MPI ranks which is not the case for the HIP or GPU backends (they are limited by the # of GPU not the number of MPI rank / GPU). In practice large l cases might run faster on CPU (with a lot of mpi ranks) than on GPU for this reason.

Although the PR replaces the GPU backend with the hip backend I am not entirely convinced this is the best option. We could benefit from both worlds on both architectures if the right backend is selected automatically depending on the angular momentum. The limiting factor in both cases is the 2 matrix-matrix multiplications needed to calculate the blocks NOT the calculation of the integrals. Coalescing memory accesses would require atomic operations as I do not think there are any other way to do this without atomic and coaslesced memory accesses.

mtaillefumier:
> Thanks a lot for putting this together!
> 
> Unfortunately, our GPU tests are in shambles at the moment. I'm working on it...

I ran the regtests locally. Only two tests failfor some curious reasons they are all about k-points.

oschuett:
> We could benefit from both worlds on both architectures if the right backend is selected automatically depending on the angular momentum. 

Yes, it would be good to have a separate kernel optimized for small `lp`, since that's the more common use case.

Still, I think unifying the CUDA and HIP backends is the right next step.

mtaillefumier:
OK. I will work on this.

oschuett:
The [CUDA performance test](https://dashboard.cp2k.org/archive/perf-cuda-volta/index.html) on the dashboard has found [an issue](https://dashboard.cp2k.org/archive/perf-cuda-volta/commit_3d3abc6804f6646a88c0bf1d0dfcd6fe047d8284.txt) that's probably caused by this PR.

mtaillefumier:
> The [CUDA performance test](https://dashboard.cp2k.org/archive/perf-cuda-volta/index.html) on the dashboard has found [an issue](https://dashboard.cp2k.org/archive/perf-cuda-volta/commit_3d3abc6804f6646a88c0bf1d0dfcd6fe047d8284.txt) that's probably caused by this PR.

I’ve found the root cause: this input requires an extremely large amount of scratch memory for the coefficient calculations, more than the GPU can provide. The implementation is behaving as designed; this case just exposes a scaling limitation. I’m working on drastically reducing the memory footprint, and that may improve performance as well.

LStuber:
I have started getting errors 700 in the grid code since this PR was merged, all my GPU runs suddenly segfault unless I disable GPU backend for grid. I don't really have time to investigate now, but sharing the stack trace here in case it makes more sense to you @mtaillefumier . The error is probably in a CUDA kernel that is not shown in the stack trace, but I notice it shows "rocmbackend" even though I'm on a x86 + H100 system, not sure if that's intended:

> 1: ERROR: "cudaErrorIllegalAddress" at /opt/cp2k/src/grid/gpu/../../offload/offload_runtime.h:340
> 1:
> 1: Program received signal SIGABRT: Process abort signal.
> 1:
> 1: Backtrace for this error:
> 1: #0  0x4000000bd09f in ???
> 1: #1  0x4000000bc057 in ???
> 1: #2  0x4000000608f7 in ???
> 1: #3  0x400000352008 in __pthread_kill_implementation
> 1:    at ./nptl/pthread_kill.c:44
> 1: #4  0x40000030a83b in __GI_raise
> 1:    at ../sysdeps/posix/raise.c:26
> 1: #5  0x4000002f7133 in __GI_abort
> 1:    at ./stdlib/abort.c:79
> 1: #6  0xaaaaab1a80f3 in offloadFree
> 1:    at /opt/cp2k/src/grid/gpu/../../offload/offload_runtime.h:404
> 1: #7  0xaaaaab1a80f3 in _ZN12rocm_backend10gpu_vectorIdE9associateEPvS2_m
> 1:    at /opt/cp2k/src/grid/gpu/grid_gpu_context.h:109
> 1: #8  0xaaaaab1a80f3 in grid_gpu_collocate_task_list
> 1:    at /opt/cp2k/src/grid/gpu/grid_gpu_context.cu:397
> 1: #9  0xaaaaab1932fb in grid_collocate_task_list
> 1:    at /opt/cp2k/src/grid/grid_task_list.c:189
> 1: #10  0xaaaaaab4f61f in __grid_api_MOD_grid_collocate_task_list
> 1:    at /opt/cp2k/src/grid/grid_api.F:1017
> 1: #11  0xaaaaaac5f9bb in __qs_collocate_density_MOD_calculate_rho_elec
> 1:    at /opt/cp2k/src/qs_collocate_density.F:1620
> 1: #12  0xaaaaab793b23 in qs_rho_update_rho_low
> 1:    at /opt/cp2k/src/qs_rho_methods.F:548
> 1: #13  0xaaaaab793b23 in __qs_rho_methods_MOD_qs_rho_update_rho
> 1:    at /opt/cp2k/src/qs_rho_methods.F:448
> 1: #14  0xaaaaab7b36db in scf_env_initial_rho_setup
> 1:    at /opt/cp2k/src/qs_scf_initialization.F:1281
> 1: #15  0xaaaaab7b36db in init_scf_run
> 1:    at /opt/cp2k/src/qs_scf_initialization.F:1153
> 1: #16  0xaaaaab7b504b in __qs_scf_initialization_MOD_qs_scf_env_initialize
> 1:    at /opt/cp2k/src/qs_scf_initialization.F:218
> 1: #17  0xaaaaab7ae153 in __qs_scf_MOD_scf
> 1:    at /opt/cp2k/src/qs_scf.F:237
> 1: #18  0xaaaaab6b762f in __qs_energy_MOD_qs_energies
> 1:    at /opt/cp2k/src/qs_energy.F:126
> 1: #19  0xaaaaaab34bfb in __force_env_methods_MOD_force_env_calc_energy_force
> 1:    at /opt/cp2k/src/force_env_methods.F:283
> 1: #20  0xaaaaab02bb1f in cp2k_run
> 1:    at /opt/cp2k/src/start/cp2k_runs.F:334
> 1: #21  0xaaaaab02eaab in __cp2k_runs_MOD_run_input
> 1:    at /opt/cp2k/src/start/cp2k_runs.F:938
> 1: #22  0xaaaaaab2ed7b in cp2k
> 1:    at /opt/cp2k/src/start/cp2k.F:379
> 1: #23  0xaaaaaab1fd23 in main
> 1:    at /opt/cp2k/src/start/cp2k.F:44

mtaillefumier:
I know that. I opened a PR fixing it. Don't worry about `rocm_backend`. I did not  change the namespace and not really planning to change any time soon. It really does not matter at all.

LStuber:
> I know that. I opened a PR fixing it. Don't worry about `rocm_backend`. I did not change the namespace and not really planning to change any time soon. It really does not matter at all.

Hi Mathieu, any update on this 2nd PR? (the one that fixes the error 700). I'm still getting the error 700 with CP2K master compiled on July 1st.

mtaillefumier:
> > I know that. I opened a PR fixing it. Don't worry about `rocm_backend`. I did not change the namespace and not really planning to change any time soon. It really does not matter at all.
> 
> Hi Mathieu, any update on this 2nd PR? (the one that fixes the error 700). I'm still getting the error 700 with CP2K master compiled on July 1st.

It would certainly help me if the input file is provided because it is not simple to debug an issue without the input triggering it.

LStuber:
> > > I know that. I opened a PR fixing it. Don't worry about `rocm_backend`. I did not change the namespace and not really planning to change any time soon. It really does not matter at all.
> > 
> > 
> > Hi Mathieu, any update on this 2nd PR? (the one that fixes the error 700). I'm still getting the error 700 with CP2K master compiled on July 1st.
> 
> It would certainly help me if the input file is provided because it is not simple to debug an issue without the input triggering it.

Hi @mtaillefumier ,

I managed to narrow down the problem to this small input, however, you need to run it with 8 MPI ranks, for some reason it works when I run it with only 1-4 ranks. If you only have 1 - 2 GPUs available, you can spawn multiple ranks per GPU until you have 8.

[buggedinput.tgz](https://github.com/user-attachments/files/30128735/buggedinput.tgz)

Can you try, see if you can reproduce? It should give you the  abort error I reported here: https://github.com/cp2k/cp2k/pull/5206#issuecomment-4576455333

mtaillefumier:
Hi @LStuber, i can not reproduce the bug on my machine. The entire simulation fit a V100 with memory to spare and the 8 rank case works without any issue. I can not reproduce the issue so far with MPI = 8 and one GPU.

oschuett:
I don't think this is an out-of-memory problem because it crashes in `cudaFree`. So my guess would be that's a _double free_.
https://github.com/cp2k/cp2k/blob/3d5583cc723f17a3686d51f8306896ec5e6667ea/src/grid/gpu/grid_gpu_context.cu#L498
https://github.com/cp2k/cp2k/blob/3d5583cc723f17a3686d51f8306896ec5e6667ea/src/grid/gpu/grid_gpu_context.h#L152
https://github.com/cp2k/cp2k/blob/3d5583cc723f17a3686d51f8306896ec5e6667ea/src/grid/gpu/grid_gpu_context.h#L202

> Don't worry about rocm_backend. I did not change the namespace and not really planning to change any time soon.

Maybe we should still change it to `gpu_backend` as it does show up in stack traces and then confuses people.

mtaillefumier:
> I don't think this is an out-of-memory problem because it crashes in `cudaFree`. So my guess would be that's a _double free_.
> 
> https://github.com/cp2k/cp2k/blob/3d5583cc723f17a3686d51f8306896ec5e6667ea/src/grid/gpu/grid_gpu_context.cu#L498
> 
> https://github.com/cp2k/cp2k/blob/3d5583cc723f17a3686d51f8306896ec5e6667ea/src/grid/gpu/grid_gpu_context.h#L152

I know what the code does I wrote it. There can not be double free for that variable as it is not allocated in the first place. the reset function does this

```
inline void reset() {
    if (internal_allocation_ && (device_ptr_ != nullptr)) {
      offloadFree(device_ptr_);
    }

    allocated_size_ = 0;
    current_size_ = 0;
    device_ptr_ = nullptr;
    host_ptr_ = nullptr;
    internal_allocation_ = false;
  }
```

the `pab_block_` is only initialized to empty when the context is created. That's all. Then the code only assign device_ptr_ to the buffer that is allocated outside the grid backend so the cuda_free can not happen since the buffer is external to the grid backend. My suspicion is that the context gets corrupted by something else switch change internal_allocation_  from false to true and the associate function only set internal_allocation_ to false. 

> https://github.com/cp2k/cp2k/blob/3d5583cc723f17a3686d51f8306896ec5e6667ea/src/grid/gpu/grid_gpu_context.h#L202
> 
> > Don't worry about rocm_backend. I did not change the namespace and not really planning to change any time soon.
> 
> Maybe we should still change it to `gpu_backend` as it does show up in stack traces and then confuses people.

PR #5210, opened 2026-05-15
Docker: Use Ubuntu 26.04 for toolchain testings

Growl1234:


Growl1234:
Could someone trigger Ubuntu GCC 11 and 16 testers? Thanks!

Growl1234:
It seems Ubuntu 26.04 has dropped `python3-reentry`. I'll try to deal with it soon.

PR #5211, opened 2026-05-15
spack* added to .gitignore

dryndyk:
Many spack folders are necessary, and files started with spack!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,6 +100,7 @@ build-*
 .spack_patched
 /spack
 /spack_cache
+spack*
 
 ### CMake Patch ###
 # External projects


mkrack:
That's a quite general git ignore pattern. Adding `spack*` to `.gitignore` will ignore all files and folders that start with "spack" anywhere in the cp2k project tree including for instance new files in `tools/spack`. What do you want to achieve?

dryndyk:
I do not want to see spack_psmp, spack-sdbg, etc., always listed as untracked.
Maybe /spack*/ is enough.

mkrack:
In that case I would suggest to replace 
```
/spack_cache
```
by the more general
```
/spack_*/
/spack-*/
```
in `.gitignore`

dryndyk:
Yes, please do it.

mkrack:
Thanks, adopted your proposal via PR #5212.

PR #5215, opened 2026-05-16
Toolchain: tblite 0.5.0 -> 0.6.0

Growl1234:
@DCM-Uni-Paderborn, could you please have a look at how to deal with the added patches since the version tblite and its dependencies has been updated? Currently I've temporarily removed the previous patching behaviors.


DCM-Uni-Paderborn:
Hi @Growl1234,

I had a closer look at the tblite 0.6.0 update and the old CP2K-side patching.

My current understanding is that we should not simply reintroduce the old `tblite-0.5.0-gradient-fixes.patch` for tblite 0.6.0. The most important parts of that patch are already upstream in tblite 0.6.0, in particular the H0 zero-distance/CN self-image handling and the multipole/Ewald stress fixes.

However, the separate CP2K patches for the vendored `s-dftd3` and `dftd4` subprojects still look relevant for the toolchain build and should stay applied for tblite 0.6.0. I would also add these patch files to the `write_checksums` call in `install_tblite.sh`, otherwise changing one of those patches would not trigger a clean reinstall.

One important point: PR #5215 currently changes the toolchain to tblite 0.6.0, but the corresponding xTB/tblite regtest reference and tolerance updates are in PR #5214. So I think the cleanest way forward is either:

1. merge/rebase #5214 first and then rebase #5215 on top of it, or
2. combine both into one tblite-0.6.0 compatibility PR.

In either case the combined state should contain:
- tblite 0.6.0 for Spack/toolchain,
- no old `tblite-0.5.0-gradient-fixes.patch`,
- continued application of the `s-dftd3` and `dftd4` gradient/stress patches,
- checksum tracking for those remaining patches,
- the regtest reference/tolerance updates from #5214.

There is one subtle remaining scientific question: the old CP2K toolchain patch also had a small H0 self-image virial branch that is not fully present in tblite 0.6.0. I would not reintroduce that blindly as a new patch. I think we should first check the targeted GFN1/GFN2 stress-debug cases with native tblite 0.6.0 and CP2K/tblite 0.6.0. If there is still a reproducible virial/stress discrepancy, I can add a very small dedicated follow-up patch.

So my recommendation would be: use tblite 0.6.0 as the new baseline, drop the obsolete tblite patch, keep the D3/D4 patches, bring the regtest references from #5214 into the same baseline, and only add a new H0-virial patch if the stress-debug comparison proves that it is still needed.

Greeting, Thomas

Growl1234:
@DCM-Uni-Paderborn Thank you for your reply! I'll then remove the tblite-0.5.0 patch and add back the checksums of other two patches. I also hope #5214 to be merged first.

PR #5219, opened 2026-05-16
Fix PBC image phase for k-point WFN extrapolation

Growl1234:
Follow up to the implementation of WFN-based extrapolation methods with k-points (#4884, #4943, #4949).

For k-point calculations, the Bloch AO basis depends on the internal PBC image convention used for atom positions in the overlap neighbor lists. During geometry or cell optimization, this convention may change for atoms close to the PBC wrapping boundary. Saved MO coefficients then need an atom-wise Bloch phase correction before being reused with the current S(k). This PR applies this correction centrally in the k-point WFN-history path, so that k-point versions of `USE_PREV_WF`, `PS`, and `ASPC` share the same conversion before reorthogonalization and density reconstruction.

There's still a remaining issue with WFN-based extrapolation related to k-point symmetry reduction, which is suitable to be handled separately.


Growl1234:
The slow test `QS/regtest-tdwf/co2_4-tdpm.inp` does not seem to be related to this PR.

DCM-Uni-Paderborn (approved):
Good to go, relation to GExt needs to be checked ...

Growl1234:
> Good to go, relation to GExt needs to be checked ...

I'm preparing a GExt approach for k-points, which seems to me rather straightforward because the current infrastructure is already sufficiently well-developed (especially with this PR applied). If further tests meets the expectation, it will be prepared as a next PR.

oschuett:
Now we've got ourselves a merge conflict...

DCM-Uni-Paderborn:
> Now we've got ourselves a merge conflict...

I have further reduced the computational effort of the tdwf regtest, which often is marked as slow ...

Growl1234:
> Now we've got ourselves a merge conflict...

I've resolved the merge conflict.

Issue #125, opened 2018-12-12
 PERIODIC_EFIELD does not work with Mixing


juerghutter:
A related problem is that the Berry phase dipole only works with
equal occupation for all states (same for OT). This means the
smearing option is not possible (metals!), but simple diagonalization
schemes without smearing should work.

juerghutter:
Periodic Efield needs wavefunction gradients and direct optimization. This is only possible with OT.
This is now tested in the input (commit 9f65b72b56b351e86ec3d6b55edbaa3bee0448db)

PR #5222, opened 2026-05-16
Move SCF k-point dipoles to qs_moments

DCM-Uni-Paderborn:
## Context
This is the follow-up to Jan Wilhelm's question in #5199 about moving the reusable k-point dipole computation from TDDFPT into `qs_moments.F` while avoiding a conflict with `qs_moment_kpoints`: https://github.com/cp2k/cp2k/pull/5199#issuecomment-4466865578

It implements the compromise described in the reply there: https://github.com/cp2k/cp2k/pull/5199#issuecomment-4467082204

## Summary
- add `qs_moment_kpoints_scf_mos` to build reusable SCF k-point dipole matrices in `qs_moments.F`
- use the helper from the TDDFPT `SCF_MOMENT` path while keeping TDDFPT-specific transition sorting, occupation handling, oscillator-strength weighting, and output local to TDDFPT
- leave `qs_moment_kpoints` and `KERNEL FULL` unchanged

## Tests
- `./make_pretty.sh`
- `git diff --check`
- `ninja -C build-serial cp2k.ssmp`
- `OMPI_CC=/opt/homebrew/bin/gcc-15 OMPI_CXX=/opt/homebrew/bin/g++-15 ninja -C build-parallel cp2k.psmp`
- `CP2K_DATA_DIR=../../../data ../../../build-serial/bin/cp2k.ssmp -i H2_kp_tddfpt_none_scf_moment.inp -o H2_kp_tddfpt_none_scf_moment.out`
- `OMP_NUM_THREADS=1 CP2K_DATA_DIR=../../../data mpiexec -n 2 --bind-to none ../../../build-parallel/bin/cp2k.psmp -i H2_kp_tddfpt_none_scf_moment.inp -o H2_kp_tddfpt_none_scf_moment.p2.out`


he-zilong:
Dear Thomas,

Would this PR enable dipole calculation with k-points as requested by [`&FORCE_EVAL/&DFT/&PRINT/&MOMENTS`](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/PRINT/MOMENTS.html) with `MAX_MOMENT 1` and `PERIODIC T` (for the Berry-phase formulation)? Would k-point dipole calculation work with electronic smearing, or would it be still limited to uniform occupation as in the first post of #125 ?

Potentially, this move could enable more functionalities than TDDFPT DIPOLE_MOMENTS, such as the IR intensities from a vibrational analysis job which requires dipole derivatives and is not available with k-points previously. (Another frequent user request is periodic electric field with k-points, but Berry-phase dipole is not its only prerequisite, see #125 again.) These applications need some follow-up tests...

Regards,
Zilong

DCM-Uni-Paderborn:
Dear Zilong,

thanks for raising this. I would view this PR in a narrower sense for now.

It does not change the existing `&FORCE_EVAL/&DFT/&PRINT/&MOMENTS` k-point printing path: that path still goes through `qs_moment_kpoints` / `qs_moment_kpoints_deep`. The new helper in `qs_moments.F` is currently only wired into the TDDFPT `DIPOLE_FORM SCF_MOMENT` path, where it builds the complex MO-basis k-point dipole matrices from the already available SCF k-point MOs.

So this PR should not by itself enable additional Berry-phase `MOMENTS` functionality, nor does it remove the existing occupation limitation. K-point TDDFPT still aborts for smeared occupations, and the existing Berry-phase moment code still has the non-uniform occupation limitation.

I agree that moving the reusable matrix construction into `qs_moments.F` makes follow-up work more natural. IR intensities from vibrational analysis, periodic electric fields with k-points, and broader `MOMENTS` integration would still need separate wiring and dedicated tests, though. I would prefer to keep this PR limited to the TDDFPT `SCF_MOMENT` use case and handle those applications in a potential follow-up PR ...

Greetings,
Thomas

PR #5225, opened 2026-05-17
Reuse symmetry-reduced SCF MOs in the Wannier90 k-point export (H. Mirhosseini)

DCM-Uni-Paderborn:
This PR extends the CP2K/Wannier90 k-point interface so that Wannier90 can use the full SCF k-point mesh directly and, when possible, reuse already available SCF MO coefficients from symmetry-reduced k-point calculations.

Main changes:

- Add `DFT%PRINT%WANNIER90%KPOINTS_SOURCE SCF` while keeping the historical `MP_GRID` path as the default.
- Add `REUSE_SCF_MOS`, `USE_BLOCH_PHASES`, and `VALIDATE_REUSE_SCF_MOS` for the SCF-sourced Wannier90 export.
- Reconstruct full-mesh Wannier90 MOs from symmetry-reduced SCF k-points using the stored atomic k-point symmetry mappings.
- Cover K290 and SPGLIB symmetry paths, including MONKHORST-PACK, MACDONALD, and GENERAL k-point schemes.
- Add Bloch-phase handling, metric/eigenvalue validation, and Ritz stabilization for degenerate reconstructed subspaces.
- Fall back to the existing full-mesh diagonalization path when SCF MO reuse is not safe, for example when required symmetry data are unavailable or distributed over incompatible k-point parallel groups.
- Harden MPI wrapper reductions used by the psmp validation path.
- Add focused regtests and a small `.mmn` comparison helper for gauge-invariant overlap validation.

The implementation validates the reconstructed SCF MOs against an internal full-mesh diagonalization reference. For the H/FCC 4x4x4 K290/SPGLIB tests, the reconstructed subspaces agree with the reference at about `1e-15`, and eigenvalues agree at about `1e-12`.

The external Wannier90 end-to-end path has also been checked. Simple non-degenerate one-band cases are quantitatively identical between `REUSE_SCF_MOS T` and `REUSE_SCF_MOS F`. For exactly degenerate multiband cases, the gauge-invariant `.mmn` singular values and `Omega I` agree, but the final MLWF minimization may still choose a different unitary gauge and therefore a different `Omega Total`. This PR therefore guarantees the reconstructed physical subspace, not bitwise/gauge-identical Wannier90 minimization trajectories for degenerate multiband spaces.

Testing:

- `./make_pretty.sh`
- `build-serial-libint/bin/cp2k.ssmp --version`
  - flags: `omp libint no_statm_access spglib`
- `./tests/do_regtest.py --mpiranks 1 --ompthreads 1 --maxtasks 4 --timeout 1200 --maxerrors 10 --restrictdir QS/regtest-kp-1 --cp2kdatadir /Users/tkuehne/cp2k-cell-kpoints/data /Users/tkuehne/cp2k-cell-kpoints/build-serial-libint/bin ssmp`
  - `66 / 66`, OK
- `./tests/do_regtest.py --mpiranks 1 --ompthreads 1 --maxtasks 4 --timeout 1200 --maxerrors 10 --restrictdir QS/regtest-kp-1-spglib --cp2kdatadir /Users/tkuehne/cp2k-cell-kpoints/data /Users/tkuehne/cp2k-cell-kpoints/build-serial-libint/bin ssmp`
  - `41 / 41`, OK
- local psmp smoke tests for K290/SPGLIB 4x4x4 reuse and fallback behavior
- external `wannier90.x` checks for one-band and degenerate H/FCC multiband cases


tools/regtesting/compare_wannier90_mmn.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python3
```
oschuett, line 1: For Python scripts it's good practice to add types annotations and have them checked with [mypy](https://mypy.readthedocs.io).

To run mypy in the CI can add your script here:

https://github.com/cp2k/cp2k/blob/52dc25883c417cc11c600f78b99989524fd9dfb9/tools/docker/scripts/test_misc.sh#L46

DCM-Uni-Paderborn: Thanks, fixed. I added strict type annotations to the helper and included it in `tools/docker/scripts/test_misc.sh` with `mypy --strict`.

PR #5226, opened 2026-05-17
Molden: Replace [Nval] with [Pseudo] and enable marking ghost atoms

### LINKED ISSUE #4353: Feature request: molden output with cell/ECP/ghost/orbital(OT) information

Growl1234:
Finally resolves #4353.

Note: The ghost atoms are marked by setting their atomic number to zero.

### DIFF
--- a/src/input_cp2k_print_dft.F
+++ b/src/input_cp2k_print_dft.F
@@ -622,6 +622,13 @@ CONTAINS
                           default_l_val=.FALSE., lone_keyword_l_val=.TRUE.)
       CALL section_add_keyword(print_key, keyword)
       CALL keyword_release(keyword)
+      CALL keyword_create(keyword, __LOCATION__, name="MARK_GHOST", &
+                          description="Controls whether ghost atoms are marked in the [Atoms] block by "// &
+                          "setting their atomic number to zero.", &
+                          usage="MARK_GHOST T", &
+                          default_l_val=.FALSE., lone_keyword_l_val=.TRUE.)
+      CALL section_add_keyword(print_key, keyword)
+      CALL keyword_release(keyword)
       CALL keyword_create(keyword, __LOCATION__, name="NDIGITS", &
                           description="Specifies the number of significant digits retained. 3 is OK for visualization.", &
                           usage="NDIGITS {int}", &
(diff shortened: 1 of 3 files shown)


Growl1234:
It also seems there's a notable point: The `zeff` in `[Pseudo]` already treats the effective nuclear charge as zero.

he-zilong (approved):
Nice! I suppose the work on CP2K's side is done, and the next step would be leaving a message [in the Chinese thread to the author of multiwfn](http://bbs.keinsci.com/thread-31827-1-1.html) and clarify that the use of `[Pseudo]` tag similar to ORCA is better than `[Nval]` (my Mg13 cluster with two kinds of element Mg could be an example).

PR #5231, opened 2026-05-17
Test installed CP2K CMake package config

### LINKED ISSUE #5059: Add tests for cp2kConfig.cmake
We're trying to migrate [Gromacs-CP2K](https://manual.gromacs.org/current/install-guide/index.html#building-with-cp2k-qm-mm-support) to CMake. So, I'm wondering if  [cp2kConfig.cmake.in](https://github.com/cp2k/cp2k/blob/master/cmake/cp2kConfig.cmake.in) is actually covered by any tests?

@mtaillefumier

DCM-Uni-Paderborn:
Closes #5059.

This adds explicit coverage for the installed `cp2kConfig.cmake` package file in the existing GROMACS Docker test. That test already installs CP2K before configuring GROMACS, so it is a natural place to validate the downstream CMake package interface and to prepare the planned GROMACS-side migration from `gmxManageCP2K.cmake` to CP2K's installed package config.

The smoke test:

- creates a tiny downstream CMake project
- calls `find_package(cp2k CONFIG REQUIRED)` with `CMAKE_PREFIX_PATH=/opt/cp2k`
- verifies that the imported target `cp2k::cp2k` exists
- verifies that the imported CP2K library location exists on disk
- stores `cp2k_config_cmake.out` as an artifact if configuration fails

The PR also fixes issues exposed by the new test in the installed package config:

- uses the same dependency package names as the main CP2K build (`Libxc`, `Spglib`, `DeePMD`, etc.)
- restores missing downstream dependencies for enabled features such as TBLITE, TREXIO, MiMiC, and openPMD
- recreates generated interface targets needed by exported link interfaces, e.g. `cp2k::tblite`, `cp2k::greenx`, and `cp2k::cosma`

## Testing

- `bash -n tools/docker/scripts/test_gromacs.sh`
- `./make_pretty.sh cmake/cp2kConfig.cmake.in tools/docker/scripts/test_gromacs.sh`
- `git diff --check origin/master...HEAD`
- `git diff --check`
- CP2K pre-push hook checked 2 files successfully

### DIFF
--- a/cmake/cp2kConfig.cmake.in
+++ b/cmake/cp2kConfig.cmake.in
@@ -61,17 +61,22 @@ if(NOT TARGET cp2k::cp2k)
 
   if(@CP2K_USE_DLAF@)
     set(CP2K_USE_DLAF @CP2K_USE_DLAF@)
-    find_dependency(DLAF REQUIRED)
+    find_dependency(DLAFFortran 0.4.0 REQUIRED)
   endif()
 
   if(@CP2K_USE_LIBXC@)
     set(CP2K_USE_LIBXC @CP2K_USE_LIBXC@)
-    find_dependency(LibXC 7 REQUIRED CONFIG)
+    find_dependency(Libxc 7 REQUIRED CONFIG)
   endif()
 
   if(@CP2K_USE_COSMA@)
     set(CP2K_USE_COSMA @CP2K_USE_COSMA@)
     find_dependency(cosma REQUIRED)
+    if(NOT TARGET cp2k::cosma)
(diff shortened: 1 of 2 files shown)


oschuett:
Thanks a lot for taking this on!

To give some more context, the plan is to migrate Gromacs' [gmxManageCP2K.cmake](https://gitlab.com/gromacs/gromacs/-/blob/main/cmake/gmxManageCP2K.cmake) onto our `cp2kConfig.cmake`.

CC: @dmmorozo

PR #5236, opened 2026-05-18
Spack docker tester: Restore cache setup and access via host endpoint

### LINKED ISSUE #5234: Using a dedicated cache network instead of host returns an error
Running the usage examples in the Spack docker files like 
```
./spack_cache_start.sh; podman build --network=cp2k-spack-cache-net --shm-size=1g -t spack_psmp-gcc14 -f ./Dockerfile.test_spack_psmp-gcc14 ../../
```
return the error
```
setup network: cannot use networks as rootless
```
The error occurs after a dedicated cache network instead of host has been introduced in #5218.

@Growl1234 Please, could you have a look.

Growl1234:
An alternative workflow of #5218, and hopefully resolves #5234.

This reverts the changes in `spack_cache_start.sh` but uses a different approach to access the local cache in Dockerfiles, thus not using custom network either exposing the regtests to host networking completely. There would also be no need to pass `--network=host` to `podman build` command.


mkrack:
LGTM, the error from #5234 disappeared and the spack output includes `Pushed to build cache: 'local-cache'` lines while the MinIO container on the host system is being populated.

PR #5238, opened 2026-05-18
Add Adaptively Compressed Exchange (ACE) option for hybrid DFT ground-state calculations

ritama1996:
This PR adds an initial implementation of the adaptively compressed exchange (ACE) operator for hybrid DFT ground-state calculations.

The implementation follows the ACE construction of Lin, J. Chem. Theory Comput. 2016, where the full HFX operator is periodically rebuilt and then represented by a low-rank ACE projector between rebuild steps.

Main changes:

- Add ACE support in the HFX/ADMM hybrid DFT path.
- Add an input keyword to control ACE usage/rebuild behavior.
- Build the ACE projector from a full HFX/ADMM calculation.
- Apply the stored ACE projector in subsequent SCF steps.
- Add support for force calculations.
- Add ground-state regression tests for ACE calculations.

New regression tests are added in:

```text
tests/QS/regtest-hybrid-ace-gs/

The tests check both total energy and total atomic force for small H2O systems, including GPW/GAPW and ADMM with different mixing variants.

### CODE REVIEW

File: tools/toolchain/scripts/stage7/install_stage7.sh [hunk 1]
```diff
@@ -3,7 +3,7 @@
 # TODO: Review and if possible fix shellcheck errors.
 # shellcheck disable=all
 
-./scripts/stage7/install_hdf5.sh
+#./scripts/stage7/install_hdf5.sh
```
Growl1234: Why do you comment the HDF5 installing, which causes installing failure?

src/common/bibliography.F
```diff
@@ -2135,27 +2141,7 @@ SUBROUTINE add_all_references()
...
-                         "Orbitals", &
-                         source="J. Chem. Theory Comput.", volume="20", pages="3894", &
-                         year=2024, doi="10.1021/acs.jctc.4c00235")
-
-      CALL add_reference(key=Schreder2024_2, &
-                         authors=s2a("L. Schreder, S. Luber"), &
-                         title="Propagated (fragment) Pipek–Mezey Wannier functions in "// &
-                         "real-time time-dependent density functional theory", &
-                         source="J. Chem. Phys.", volume="160", pages="214117", &
-                         year=2024, doi="10.1063/5.0203442")
```
Growl1234, line 2165: Are those lines removed by mistake?

Growl1234 (commented):
Very excited to see the implementation of ACE approach!

To fix precommit error, simply run `./make_cp2k.sh`. There are also some other comments on some code that might not be correct and/or could cause regtest failure.

src/common/bibliography.F
ritama1996: Yes, it was showing some merge conflicts. I guess it should be fine now.

tools/toolchain/scripts/stage7/install_stage7.sh
ritama1996: Yes, it was having errors

ritama1996:
> Very excited to see the implementation of ACE approach!
> 
> To fix precommit error, simply run `./make_cp2k.sh`. There are also some other comments which might not be correct and/or could cause regtest failure.

Thank you so much!

tests/TEST_DIRS
```diff
@@ -425,4 +425,3 @@ QS/regtest-trexio-2                                         trexio
 QS/regtest-rtbse-gxac                                       libint greenx !ifx
 QS/regtest-cneo
 QS/regtest-tddfpt-bse                                       libint libxc
-QS/regtest-tdwf
```
Growl1234, line 429: Here the same. Another problem is that, you added a regtest directory but have not included it in this `TEST_DIRS` file.

ritama1996:
The precommit issues have been fixed and the checks are now passing. I have also restored the accidental toolchain change and addressed the bibliography conflict from the rebase. Please let me know if further changes are needed.

PR #5239, opened 2026-05-18
Rename spack psmp tester and add new gcc 16 tester

mkrack:


Growl1234:
@DCM-Uni-Paderborn Could you please have a look at the failed test?

Issue #4972, opened 2026-03-19
RFC: The expansions to the CP2K manual


Growl1234:
From my perspective, the current CP2K documentation could be further improved in terms of usability, especially for new users. A key issue is that many keywords in the Input Reference are described only very briefly, without sufficient explanation of their practical meaning and usage in real calculations. Important aspects such as typical use cases, parameter sensitivity, and common pitfalls are often not covered. As a result, users frequently need to consult the original literature to understand the underlying methods and implementation details, and rely on repeated trial and error to use certain features effectively. This leads to a fairly steep learning curve for CP2K.

I'm aware that the current Input Reference is directly generated from source code, and one effective approach would be to extend these descriptions in the Fortran source. However, this may not always be the most suitable solution, especially when more detailed, tutorial-style explanations are needed.

As an alternative (or complementary) approach, my thought is that, **while preserving the current Input Reference structure, we could significantly expand the sections on theoretical background and computational methodology. These sections could include more detailed and practically useful explanations, and be cross-linked with the keyword descriptions in the Input Reference.**

As a reference, the documentation of ORCA — including its user manual and tutorials — demonstrates how detailed explanations and practical guidance can be integrated effectively. While reaching a similar level of completeness may take so much time, it could be a reasonable long-term direction to work towards incrementally.

If possible, it might also be valuable to encourage contributions from a wide range of users who are familiar with the CP2K codebase.

I would be very appreciated to hear the community’s thoughts on this.

oschuett:
Thanks a lot for taking on the documentation!

On approach that always appealed to me is the [The Documentation System](https://docs.divio.com/documentation-system/) by Daniele Procida ([video](https://www.youtube.com/watch?v=t4vKPhjcMZg)).

Looking at our manual through the lens of those four categories: 

- We covered **References** well. 

- Our _Method_ section is currently a mix of **Explanations** and **Howto**. While we could try to separate these two, I think the more pressing issue here is completeness. Since the Sphinx-based manual is only 3 years old, many of the older - but popular - methods are not covered well, e.g. [GAPW](https://manual.cp2k.org/trunk/methods/dft/gapw.html) and [ADMM](https://manual.cp2k.org/trunk/methods/dft/hartree-fock/admm.html).

[...]

Issue #1233, opened 2020-12-18
Clean up gopt_f77_methods.f90


oschuett:
The [gopt_f77_methods.fypp](https://github.com/cp2k/cp2k/blob/master/src/motion/gopt_f77_methods.fypp) is a hack to implement a circular dependency between modules. It's needed for the `GEO_OPT` and `MD` type of [cell optimization](https://manual.cp2k.org/trunk/CP2K_INPUT/MOTION/CELL_OPT.html#TYPE). They are both obsolete and we should simply remove them to eliminate that circular dependency.

dev-zero:
see also #1231

oschuett:
Upps, I overlook that. Should I close this issue or shall we specialize them and make  #1231 about `f77_interface.F` only?

dev-zero:
Probably specializing them makes sense, this way we can (hopefully) close issues directly with PRs and we get a better overview rather than having subtasks in issues. I wasn't aware that the f77 interface in `motion/` is actually more a workaround rather than an API. Thanks!

he-zilong:
A major distinction between options of `&MOTION/&CELL_OPT/TYPE` is how the pressure tensor is calculated which is handled by several `BLOCK`s in `cp_eval_at()` of `gopt_f77_methods.F`. Why this design? Would a refactor where these `BLOCK`s are converted to proper `SUBROUTINE`s in a separate `MODULE` help with this issue?

oschuett:
I think, we should just remove the [&MOTION/&CELL_OPT/TYPE](https://manual.cp2k.org/trunk/CP2K_INPUT/MOTION/CELL_OPT.html#TYPE) keyword. AFAIK, it's not used and without it we don't have the recursive call path that leads to the circular dependency between modules.

he-zilong:
> I think, we should just remove the [&MOTION/&CELL_OPT/TYPE](https://manual.cp2k.org/trunk/CP2K_INPUT/MOTION/CELL_OPT.html#TYPE) keyword. AFAIK, it's not used and without it we don't have the recursive call path that leads to the circular dependency between modules.

[...]

PR #5240, opened 2026-05-18
Rewrite documentation of molecular dynamics and structure optimization

Growl1234:


docs/methods/optimization/geometry_and_cell_opt.md
```diff
@@ -0,0 +1,362 @@
...
+only internal atomic coordinates should relax.
+
+Use `CELL_OPT` when the equilibrium cell volume, shape, pressure, or residual stress is part of the
+question. It is most common for bulk crystals, strained solids, two-dimensional materials with
+relaxed in-plane lattice constants, and structures prepared for fixed-cell molecular dynamics. A
+cell optimization is not a replacement for finite-temperature pressure sampling; if the desired
+quantity is a thermal average at finite temperature, use an appropriate NPT molecular-dynamics
+workflow instead.
+
+`CELL_OPT` has different operating modes selected by [`TYPE`](#CP2K_INPUT.MOTION.CELL_OPT.TYPE):
```
he-zilong: Please note that in #1233 there is a proposed removal of this `TYPE` keyword so that we always use `DIRECT_CELL_OPT` in practice. Type `GEO_OPT` is somewhat obsolete, and type `MD` is expensive and hard to converge.

docs/methods/optimization/geometry_and_cell_opt.md
```diff
@@ -0,0 +1,362 @@
...
+optimization iteration can require more than one force evaluation, depending on the optimizer, line
+search, and the selected `CELL_OPT` mode.
+
+## Starting structure and cell
+
+Start from a chemically and physically sensible structure. Severe close contacts, unrealistic
+coordination environments, or a badly prepared cell can lead to unstable forces, SCF failures,
+unphysical atom displacements, or optimizer steps that are difficult to recover from.
+
+For `CELL_OPT`, the initial cell matters as much as the initial coordinates. The starting volume and
```
he-zilong, line 117: Please put emphasis on the observation in #3384 that the cell optimization job expects a cell matrix conforming to the convention that A is along the X-axis and B is on the XY plane.

docs/methods/optimization/geometry_and_cell_opt.md
```diff
@@ -0,0 +1,362 @@
...
+The most important atomic convergence criteria are
+[`MAX_FORCE`](#CP2K_INPUT.MOTION.GEO_OPT.MAX_FORCE),
+[`RMS_FORCE`](#CP2K_INPUT.MOTION.GEO_OPT.RMS_FORCE), [`MAX_DR`](#CP2K_INPUT.MOTION.GEO_OPT.MAX_DR),
+and [`RMS_DR`](#CP2K_INPUT.MOTION.GEO_OPT.RMS_DR). Force criteria are given in Hartree/Bohr, while
+displacement criteria are given in Bohr. For `CELL_OPT`, the pressure target and pressure
+convergence are controlled by [`EXTERNAL_PRESSURE`](#CP2K_INPUT.MOTION.CELL_OPT.EXTERNAL_PRESSURE)
+and [`PRESSURE_TOLERANCE`](#CP2K_INPUT.MOTION.CELL_OPT.PRESSURE_TOLERANCE). The optimization is
+considered converged only when all active criteria are satisfied.
+
+When `CELL_OPT` uses `TYPE GEO_OPT`, both the inner `GEO_OPT` criteria and the outer `CELL_OPT`
```
he-zilong: See the notes for `TYPE` above

docs/methods/optimization/geometry_and_cell_opt.md
```diff
@@ -0,0 +1,362 @@
...
+structure or cell and the force/stress quality. Then consider reducing the `BFGS`/`LBFGS` trust
+radius, constraining problematic cell degrees of freedom, switching to `CG`, or using a short
+preliminary optimization with looser settings before tightening the convergence criteria.
+
+## Convergence criteria
+
+The most important atomic convergence criteria are
+[`MAX_FORCE`](#CP2K_INPUT.MOTION.GEO_OPT.MAX_FORCE),
+[`RMS_FORCE`](#CP2K_INPUT.MOTION.GEO_OPT.RMS_FORCE), [`MAX_DR`](#CP2K_INPUT.MOTION.GEO_OPT.MAX_DR),
+and [`RMS_DR`](#CP2K_INPUT.MOTION.GEO_OPT.RMS_DR). Force criteria are given in Hartree/Bohr, while
```
he-zilong: Please mention that the convergence criteria of the LBFGS optimizer are different from that of the BFGS and CG optimizers. It seems common for users to question why LBFGS is easily and loosely converged...

docs/methods/optimization/geometry_and_cell_opt.md
```diff
@@ -0,0 +1,362 @@
...
+on defaults:
+
+```none
+&GLOBAL
+  PROJECT my_cell_opt
+  RUN_TYPE CELL_OPT
+&END GLOBAL
+
+&FORCE_EVAL
+  ! Define the method that provides energies, forces, and stress.
```
he-zilong, line 83: The definition of the `STRESS_TENSOR` keyword should be explicit. Maybe setting `STRESS_TENSOR ANALYTICAL` as default is okay, and only fall back to numerical stress tensor when no analytical stress tensor is available.

docs/methods/optimization/geometry_and_cell_opt.md
```diff
@@ -0,0 +1,362 @@
...
+For ab-initio geometry or cell optimization, every optimization step requires an
+electronic-structure calculation. The force quality depends on the SCF convergence, grid settings,
+basis set, pseudopotentials, and method-specific thresholds. For `CELL_OPT`, the stress tensor and
+pressure must also be accurate enough for reliable cell updates. Important Quickstep settings
+include [`EPS_DEFAULT`](#CP2K_INPUT.FORCE_EVAL.DFT.QS.EPS_DEFAULT),
+[`CUTOFF`](#CP2K_INPUT.FORCE_EVAL.DFT.MGRID.CUTOFF),
+[`REL_CUTOFF`](#CP2K_INPUT.FORCE_EVAL.DFT.MGRID.REL_CUTOFF), and
+[`EPS_SCF`](#CP2K_INPUT.FORCE_EVAL.DFT.SCF.EPS_SCF).
+
+Optimization is driven by forces, and `CELL_OPT` is additionally driven by stress, rather than by
```
he-zilong, line 294: Not a proposed revision, but just a little note that I would like to write a separate document about checking `CUTOFF` with forces and stress tensors sometime later.

docs/methods/optimization/geometry_and_cell_opt.md
```diff
@@ -0,0 +1,362 @@
...
+```
+
+This indicates that the previous density matrix may not provide a high-quality initial guess. If
+this warning appears together with abnormal changes in the geometry, cell, total energy, forces,
+stress, or SCF behaviour, stop using that density-matrix-based method and switch to a safer
+alternative.
+
+## Output files and restarts
+
+The main output file contains the optimization progress and convergence checks. CP2K also writes
```
he-zilong, line 331: Please mention that the latest CP2K version (after my #5118 and #5140) there is a `cif` and an extended `xyz` output for the final structure. The name (excluding extensions) is both `project-FINAL-1_{cycles}` where `{cycles}` is the number of iterations at convergence or at `MAX_ITER`.

docs/methods/optimization/geometry_and_cell_opt.md
```diff
@@ -0,0 +1,362 @@
...
+
+[`MAX_ITER`](#CP2K_INPUT.MOTION.GEO_OPT.MAX_ITER) limits the number of optimization iterations. One
+optimization iteration can require more than one force evaluation, depending on the optimizer, line
+search, and the selected `CELL_OPT` mode.
+
+## Starting structure and cell
+
+Start from a chemically and physically sensible structure. Severe close contacts, unrealistic
+coordination environments, or a badly prepared cell can lead to unstable forces, SCF failures,
+unphysical atom displacements, or optimizer steps that are difficult to recover from.
```
he-zilong, line 115: It is worth mentioning the problems with experimental cif.
```suggestion
unphysical atom displacements, or optimizer steps that are difficult to recover from. In particular, experimental `cif` should not be used as the initial structure without validation; refinement and careful revision is necessary to make it "computation-ready" during pre-processing.
- Watch out for crystallographic disorder and fractional occupation of atoms: using the superposition of all atoms as if every occupancy is 1.00 is a frequent pitfall as it introduces contacting or even overlapping atoms.
- Beware of composition: the atomic structure may not match the intended charge-neutral chemical formula, especially hydrogen atoms, which may be missing or duplicated due to disorder.
```

docs/methods/sampling/molecular_dynamics.md
```diff
@@ -26,78 +17,268 @@ privacy_mode:
...
+```
+
+For thermostatted or barostatted simulations, also inspect the corresponding conserved quantity or
+cell output when available. For production runs, write restart files frequently enough that the
+trajectory can be continued without losing significant sampling time.
+
+## Validation checklist
+
+Before running a long production trajectory, it's suggested to check that:
+
```
he-zilong, line 273:
```suggestion

- The affordable length of trajectory is sufficient for observing phenomena, for instance chemical reactions need to be kinetically favorable so as to happen on a picosecond timescale that AIMD usually manages to cover.
```

he-zilong (commented):
Another step for #4972 !

docs/methods/optimization/geometry_and_cell_opt.md
Growl1234: Should I add a note to this part or better directly remove this?

he-zilong: I would remove it altogether.

docs/methods/optimization/geometry_and_cell_opt.md
oschuett: We could link to this existing page: https://manual.cp2k.org/trunk/methods/dft/cutoff.html

docs/methods/optimization/index.md
```diff
@@ -5,6 +5,6 @@
 titlesonly:
 maxdepth: 2
 ---
-geometry
```
oschuett, line 8: Please also add a redirect: 

https://github.com/cp2k/cp2k/blob/e068a9b9a79ff801c506cc5bc38f62940e03fd5f/docs/conf.py#L68

docs/methods/optimization/geometry_and_cell_opt.md
he-zilong: > We could link to this existing page: https://manual.cp2k.org/trunk/methods/dft/cutoff.html

The text says "do not choose `EPS_SCF`, `CUTOFF`, `REL_CUTOFF`, or `EPS_DEFAULT` only from a final static-energy test." Which is what the existing "How to Converge the CUTOFF and REL_CUTOFF" page you linked is doing (and more precisely, it uses the "final static-energy" after `MAX_SCF 1`, i.e. not at full convergence).

I think the existing "How to Converge the CUTOFF and REL_CUTOFF" page is already long enough that I do not intend to add anything to it. My intention is to describe a distinct, slightly more sophisticated method of testing `CUTOFF` by force and/or stress on a separate page. However, I could not work on this matter until a few days later...

(While I am on this topic: what external plotting program do you suggest I use for the documentation, e.g. gnuplot and/or python+matplotlib? This is mostly about the style and preference of dependency.)

oschuett: > My intention is to describe a distinct, slightly more sophisticated method of testing CUTOFF ....

Sounds good!

> what external plotting program do you suggest I use for the documentation, ...

Please use python+matplotlib and ideally also check in the script itself so that other can build on it.

Growl1234:
Thanks for all your suggestions! I am a bit busy these days, and I'll hopefully get back to this PR by this weekend at the latest.

docs/methods/optimization/geometry_and_cell_opt.md
Growl1234: The non-convention cell seems also work with `CELL_OPT`. I've seen the reason why a convention cell is expected but maybe it's better to add a code that converts the cell to make it meet expectation?

he-zilong: For this PR's purpose of updating the document page, I think clarifying the convention and the current program behavior is crucial.
Potential revisions to the code (either converting the input cell and coordinates to match the convention, or estabilishing the MOTION support for now-unconventional cells) are better discussed elsewhere later, and there would be more to do for the actual coding work than updating this document again.

docs/methods/sampling/molecular_dynamics.md
```diff
@@ -26,78 +17,271 @@ privacy_mode:
...
-of the highest oscillation of the system (typically H atoms, that is the reason why sometimes D
-atoms are used).
-
-Conserving the total energy well is not enough if the system is very heterogeneous and the
-interesting part is small. In that case there is the risk that even a large error on it might pass
-unnoticed. If this is to be excepted (for example the atom with the sharpest gaussian basis set is
-in the interesting part) then checking that part of the system (oscillations, kinetic energy,...) is
-advisable.
+The [`CSVR / TIMECON`](#CP2K_INPUT.MOTION.MD.THERMOSTAT.CSVR.TIMECON) value controls the coupling
+strength. Useful starting values are `50 fs` for rapid heating or cooling, `100 fs` as a simple
```
he-zilong: The default unit of `TIMECON` is already `fs`, so a time constant of 100.0 fs can be simply written as `TIMECON 100.0` in the input file (just like the setup above).

For those who prefer explicitly specifying the unit, a time constant of 100.0 fs should be written as `TIMECON [fs] 100.0` in the input, with the unit in square brackets and put before the value. See for example the `./tests/Fist/regtest-water/H2O-TIP3P-6x6x6.inp` file.

Please correct the occurrences of `TIMECON .. fs` below by either omitting `fs` or using the preceding `[fs]` format.

he-zilong (commented):
Apart from the two subtle details about cell convention and unit format, everything else LGTM 😀

docs/methods/sampling/molecular_dynamics.md
Growl1234: Fixed.

docs/methods/optimization/geometry_and_cell_opt.md
Growl1234: Thanks for the suggestion, then there has been a hint that we expect a convention cell for `CELL_OPT` at the next paragraph.

PR #5241, opened 2026-05-18
Reuse SCF k-point dipoles in MOMENTS output

DCM-Uni-Paderborn:
This is a small follow-up to #5222 and the discussion about making the reusable SCF k-point dipole matrix available from `qs_moments.F`.

The PR wires the existing SCF-MO k-point dipole helper into the default `DFT%PRINT%MOMENTS` k-point path when no explicit `MOMENTS%KPOINTS` or `MOMENTS%KPOINT_SET` is requested. Explicit MOMENTS k-point inputs continue to use the existing `qs_moment_kpoints_deep` path, so the general moment-printing/Berry-curvature machinery remains unchanged for those cases.

The SCF-MO path now also returns the number of available SCF MOs per spin, so the MOMENTS printer only prints the actually available SCF-MO block instead of carrying AO-dimensioned unused entries.

This deliberately does not try to address IR intensities from vibrational analysis or periodic electric fields with k-points; those require separate Berry-phase / force-response wiring and dedicated tests.

## Testing

- `cmake --build build-serial --target cp2k.ssmp -- -j 4`
- Ran the MOMENTS k-point inputs manually with `cp2k.ssmp`:
  - `C2_pbe_scf_kp.inp`
  - `C2_pbe_moment_kp.inp`
  - `C2_pbe_moment_kpset.inp`
  - `CrSBr_open_shell.inp`
- `git diff --check`

### DIFF
--- a/src/qs_moments.F
+++ b/src/qs_moments.F
@@ -3830,12 +3830,15 @@ CONTAINS
 !> \param qs_env ...
 !> \param dipole ...
 !> \param rcc retained for interface compatibility; interband dipoles are origin independent
+!> \param nmo_spin_out number of SCF MOs available for each spin
 ! **************************************************************************************************
-   SUBROUTINE qs_moment_kpoints_scf_mos(qs_env, dipole, rcc)
+   SUBROUTINE qs_moment_kpoints_scf_mos(qs_env, dipole, rcc, nmo_spin_out)
       TYPE(qs_environment_type), POINTER                 :: qs_env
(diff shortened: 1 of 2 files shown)


JWilhelm:
Dear Thomas, 

after having a close look, it looks good to me! I asked Shridhar, who has been coding the k-dependent dipoles to also have a look until tomorrow 11 am. Then I will finalize the review.

JWilhelm:
Thanks for waiting, Thomas. Looks good also to Shridhar!

JWilhelm (approved):

PR #5242, opened 2026-05-18
Move GauXC toolchain script from stage3 to stage6

Growl1234:
See https://github.com/cp2k/cp2k/pull/5084#discussion_r3241948271.


franzpoeschel:
The GauXC test will probably still fail. The install scripts add a patch to GauXC that is (partially) no longer up to date (the patch fixes a behavior that is now also fixed within CP2K, so the fix is currently duplicated, likely giving wrong results).

The relevant part of the patch is currently open as a PR https://github.com/wavefunction91/GauXC/pull/202. I'll suggest waiting for that to be merged, and then using GauXC without patches in the toolchain.

Growl1234:
It seems, a problem at CMake step is occurring, which is more likely related to GauXC's CMake configuration files ...

DCM-Uni-Paderborn:
Dear SY Wang,

Ciao Franz,

thanks for the comments here. Franz' concern about the old GauXC patch was correct: after the GauXC-side changes in wavefunction91/GauXC#202, the earlier toolchain patch needed to be refreshed instead of being carried forward unchanged.

I opened #5244 as the follow-up PR. It includes the stage6 move and the `CP2K_USE_GAUXC` / `CP2K_USE_EVERYTHING` CMake change from this draft PR, but also updates the GauXC/SKALA toolchain integration and keeps only the still-needed compatibility fixes. The GauXC patch application against the pinned GauXC revision was checked locally, and a CP2K `psmp` build with GauXC/libtorch/SKALA enabled passed `QS/regtest-gauxc` with both 1 and 2 MPI ranks.

This means #5242 does not have to be merged first. If #5244 lands first, the main content of this draft PR is covered there. If you prefer to merge #5242 first, #5244 can be rebased afterwards.

Greeting,
Thomas

Issue #202, opened 2019-02-13
[toolchain] inconsistent setup script and arch files when reusing build directory


dev-zero:
When starting to build a toolchain which fails somewhere in the middle and then continue to build it with different flags or when changing the flags after a successful build are the `setup` script and the `arch/` files sometimes inconsistent.

Example when starting with a toolchain-provided OpenMPI and then restart/continue with the system-provided one (linking against the toolchain-provided OpenMPI, but not adding `mpirun` to the `$PATH` in the `setup`):

```setup
#!/bin/bash
source "/data/******/GITHUB/toolchain/scripts/tool_kit.sh"
export CP2K_TOOLCHAIN_OPTIONS="--with-quip"
export VALGRIND_OPTIONS="--suppressions=/data/******/GITHUB/toolchain/install/valgrind.supp --max-stackframe=2168152 --error-exitcode=42"
export LSAN_OPTIONS=suppressions=/data/******/GITHUB/toolchain/install/lsan.supp
export TSAN_OPTIONS=suppressions=/data/******/GITHUB/toolchain/install/tsan.supp
prepend_path LD_LIBRARY_PATH "/data/******/GITHUB/toolchain/install/lapack-3.8.0/lib"
prepend_path LD_RUN_PATH "/data/******/GITHUB/toolchain/install/lapack-3.8.0/lib"
prepend_path LIBRARY_PATH "/data/******/GITHUB/toolchain/install/lapack-3.8.0/lib"
prepend_path LD_LIBRARY_PATH "/data/******/GITHUB/toolchain/install/openblas-0.3.3/lib"
prepend_path LD_RUN_PATH "/data/******/GITHUB/toolchain/install/openblas-0.3.3/lib"
prepend_path LIBRARY_PATH "/data/******/GITHUB/toolchain/install/openblas-0.3.3/lib"
prepend_path CPATH "/data/******/GITHUB/toolchain/install/openblas-0.3.3/include"
prepend_path LD_LIBRARY_PATH "/data/******/GITHUB/toolchain/install/fftw-3.3.7/lib"
prepend_path LD_RUN_PATH "/data/******/GITHUB/toolchain/install/fftw-3.3.7/lib"
prepend_path LIBRARY_PATH "/data/******/GITHUB/toolchain/install/fftw-3.3.7/lib"
prepend_path CPATH "/data/******/GITHUB/toolchain/install/fftw-3.3.7/include"
[...]

PR #5244, opened 2026-05-18
Follow up GauXC/SKALA integration after #5084

DCM-Uni-Paderborn:
This PR follows up on the merged GauXC/SKALA integration from #5084.

The main changes are:

- update the GauXC toolchain integration to use the current `wavefunction91/GauXC:skala` state after GauXC #202
- move the GauXC toolchain script and patch to stage6
- make `CP2K_USE_GAUXC` follow `CP2K_USE_EVERYTHING`
- align GauXC RKS density-matrix scaling with CP2K's spin-summed RKS density convention
- add additional guards for unsupported GauXC/SKALA paths
- extend GauXC/SKALA regression coverage, including PBE baselines, UKS, GAPW, and GAPW_XC cases
- keep SKALA/OneDFT response-kernel paths guarded where higher XC derivatives are not implemented

This also keeps the remaining CP2K-side changes independent of draft PR #5242.

Validation done locally:

- `make_pretty.sh tools/toolchain/scripts/stage6/gauxc-1.1-skala-cp2k-fixes.patch`
- `git diff --check`
- GauXC patch application test against the pinned GauXC revision
- CP2K `psmp` build with GauXC/libtorch/SKALA enabled
- `QS/regtest-gauxc` with 1 MPI rank: 9/9 correct
- `QS/regtest-gauxc` with 2 MPI ranks: 9/9 correct

I did not run the complete CP2K regression suite locally.


franzpoeschel:
Some comments on the GauXC patch:

1. This seems to reintroduce the xmat factor fix in GauXC. I did not realize until now that this patch https://github.com/cp2k/cp2k/pull/5084/changes/35c86bf2c45d1574a999f6d478d24e87219727a9 reverted the CP2K-side fix. If we want to keep the fix inside CP2K, we should undo the changes in GauXC again and reapply this patch https://github.com/cp2k/cp2k/pull/5084/changes/7033dc64c01edef077b63dd935523f4b1cefa6c6 (maybe additionally in other places added since then?).
2. The gau2grid malloc fix would ideally go into gau2grid directly, but that seems to be unmaintained? Ref. https://github.com/psi4/gau2grid/pull/76
3. Diverging on the CMake scripts is probably fine, considering this is about making things run in the CI / toolchain.
4. The fixed path lookup seems like it should go upstream into GauXC?
5. For the other changes, I don't really understand what makes them necessary (guarding some MPI calls) or what they do. We should probably try to not give users diverging behaviors depending how they installed GauXC though?

DCM-Uni-Paderborn:
I will merge this now, but a follow-up PR with cheaper regtests with smaller size and cutoff would be appreciated.

Issue #3524, opened 2024-07-02
Fortran runtime error: Bad value during floating point read using cubecruncher.x


jakobcasa:
Dear all

I did install the tool cubecruncher according to the instructions
(e.g. $ git clone https://github.com/cp2k/cp2k.git
$ cd cp2k/tools/cubecruncher
$ make
)

but using ./cubecruncher.x  -i ../../../m1/1_opt_S0_MeCN_CAM-B3LYP_def2-TZVP.eldens.cube -subtract ../../../m1/1_opt_S1_MeCN_CAM-B3LYP_def2-TZVP.eldens.cube -o test.cube

I get the following Error

 Reading from cube          ../../../m1/1_opt_S0_MeCN_CAM-B3LYP_def2-TZVP.eldens.cube
 Subtracting cube           ../../../m1/1_opt_S1_MeCN_CAM-B3LYP_def2-TZVP.eldens.cube
 Writing to   cube          test.cube
Reading cube ... At line 157 of file cubecruncher.f90 (unit = 117, file = '../../../m1/1_opt_S0_MeCN_CAM-B3LYP_def2-TZVP.eldens.cube')
Fortran runtime error: Bad value during floating point read

Error termination. Backtrace:
#0  0x1504aea681a9 in formatted_transfer_scalar_read
	at ../../../libgfortran/io/transfer.c:1790
#1  0x1504aea69875 in formatted_transfer
	at ../../../libgfortran/io/transfer.c:2491
#2  0x4088b5 in ???
#3  0x40fb2a in ???
#4  0x40107c in ???
#5  0x1504ae43feaf in ???
#6  0x1504ae43ff5f in ???
#7  0x4010b4 in ???
#8  0xffffffffffffffff in ???

What could be the issue here?

Thank you for the help

best
Jakobcasa

niklashoelter:
Any updates on this?

he-zilong:
The error points to this line of fortran code:
https://github.com/cp2k/cp2k/blob/993cc6f9025760d3120bb65f5af964521e61d467/tools/cubecruncher/cubecruncher.f90#L157
It demands that each line in cube file should record grid data in the format `6E13.5`, which is intended to work with cubes produced by CP2K Quickstep and as far as I know adheres to official Gaussian specification at [cubegen](https://gaussian.com/cubegen/).
[...]

PR #5245, opened 2026-05-18
Write cube values with explicit three-digit exponents

### LINKED ISSUE #850: CDFT Becke weight function breaks for small ATOMIC_RADII
Dear CP2K developers,

I have found that when ionic radii are used for a bulk MgO crystal the Becke weight function appears to break ([see image](https://user-images.githubusercontent.com/10100029/78054986-c9f98a00-737a-11ea-9e08-174ac58f349b.png)). When using covalent radii however the weight function is calculated correctly ([see image](https://user-images.githubusercontent.com/10100029/78055109-f3b2b100-737a-11ea-9a4b-1c332c702852.png)).

More generally, it appears that for certain systems the Becke weight function breaks when the constrained atoms have ATOMIC_RADII smaller than the unconstrained atoms. I have confirmed this behaviour for two different structures on two different machines using CP2K 7.1.

I include an example input and output file for the system with ionic and covalent radii.

Ionic radii: Mg 0.72 Å, O 1.40 Å  (Shannon, R. D. (1976))
[Input](https://github.com/cp2k/cp2k/files/4410567/ionic_input.txt)
[Output](https://github.com/cp2k/cp2k/files/4410568/ionic_output.txt)
[Geometry](https://github.com/cp2k/cp2k/files/4410583/input.txt)

Covalent radii: Mg 1.39 Å, O 0.63 Å (Pyykkö, P., & Atsumi, M. (2009)
[Input](https://github.com/cp2k/cp2k/files/4410574/covalent_input.txt)
[Output](https://github.com/cp2k/cp2k/files/4410575/covalent_output.txt)

(Originally posted on CP2K user forum, moved here as I believe this to be a bug: https://groups.google.com/forum/#!searchin/cp2k/ahart|sort:date/cp2k/1yTNsYN6nRQ/S9HSmHXBAQAJ)

DCM-Uni-Paderborn:
This PR directly addresses #850.

CP2K's cube writer used `E13.5`. For values with three-digit exponents, some Fortran runtimes drop the `E` to stay within the 13-character field, e.g. `0.33004-101`. These values are valid for Fortran reads but can be misparsed by common cube viewers, which creates artificial isosurfaces in CDFT Becke weight cubes when many values are extremely small.

Use `E13.5E3` for cube output and cubecruncher output so values keep the same 13-character field width but always include an explicit exponent marker, e.g. `0.33004E-101`. The fixed width preserves the MPI-IO layout while making the files more portable.

Fixes #850.

Verification performed:
- Built `cp2k.psmp` and `realspace_grid_cube_unittest.psmp`.
- Ran `realspace_grid_cube_unittest.psmp`.
- Ran `make_pretty.sh` on the changed files.
- Ran `git diff --check`.
- Reproduced the issue input: old ion-radii Becke cube had 1163 no-`E` three-digit exponent values; patched output has none.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1865,6 +1865,7 @@ list(
   "gemm_square_unittest"
   "memory_utilities_unittest"
   "parallel_rng_types_unittest"
+  "realspace_grid_cube_unittest"
   "graph"
   "dumpdcd"
   "xyz2dcd"
@@ -1880,6 +1881,7 @@ add_executable(cp2k-bin start/cp2k.F)
 add_executable(gemm_square_unittest common/gemm_square_unittest.F)
 add_executable(memory_utilities_unittest common/memory_utilities_unittest.F)
 add_executable(parallel_rng_types_unittest common/parallel_rng_types_unittest.F)
+add_executable(realspace_grid_cube_unittest pw/realspace_grid_cube_unittest.F)
 add_executable(graph metadyn_tools/graph.F)
 add_executable(dumpdcd motion/dumpdcd.F)
 add_executable(xyz2dcd motion/xyz2dcd.F)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1519,6 +1519,7 @@ set(CP2K_PROGS_F
     common/gemm_square_unittest.F
     common/memory_utilities_unittest.F
     common/parallel_rng_types_unittest.F
+    pw/realspace_grid_cube_unittest.F
     metadyn_tools/graph.F
     motion/dumpdcd.F
     motion/xyz2dcd.F
--- a/src/input_cp2k_external.F
+++ b/src/input_cp2k_external.F
@@ -116,7 +116,10 @@ CONTAINS
       !keyword for reading the external potential from cube file
       CALL keyword_create(keyword, __LOCATION__, name="READ_FROM_CUBE", &
                           description="Switch for reading the external potential from file pot.cube. The values "// &
-                          "of the potential must be on the grid points of the realspace grid.", &
+                          "of the potential must be on the grid points of the realspace grid. "// &
+                          "Whitespace-separated cube values are accepted. If adjacent values are written "// &
+                          "without whitespace, each value must occupy a 13-character E13.5 field, as in "// &
+                          "CP2K-generated cube files.", &
                           usage="READ_FROM_CUBE T", default_l_val=.FALSE., lone_keyword_l_val=.TRUE.)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
--- a/src/input_cp2k_opt.F
+++ b/src/input_cp2k_opt.F
@@ -270,7 +270,9 @@ CONTAINS
 
       CALL keyword_create(keyword, __LOCATION__, name="READ_EMBED_POT_CUBE", &
                           description="Read the embedding potential  "// &
-                          "(restart) from the cube file.", &
+                          "(restart) from the cube file. Whitespace-separated cube values are accepted. If "// &
+                          "adjacent values are written without whitespace, each value must occupy a "// &
+                          "13-character E13.5 field, as in CP2K-generated cube files.", &
                           usage="READ_EMBED_POT_CUBE .FALSE.", default_l_val=.FALSE.)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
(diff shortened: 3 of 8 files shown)


he-zilong:
As shown in the now-closed #3524 there are users attempting to read cube files from third-party programs. I see that we restrict the float-point format of `READ` as `E13.5` for the new `SUBROUTINE cube_read_values` in `src/pw/realspace_grid_cube.F`, and similarly as `6E13.5` for the existing `SUBROUTINE read_cube` in `tools/cubecruncher/cubecruncher.f90`. Maybe we should clarify this strict format requirement in the documentation.

DCM-Uni-Paderborn:
Thanks, good point. I clarified the cube input descriptions in the generated input reference: whitespace-separated values are accepted, while adjacent values without whitespace must use the 13-character E13.5-style fields written by CP2K.

he-zilong:
[A Chinese discussion](http://bbs.keinsci.com/thread-60030-1-1.html) points out that the new cube file format in the 2026.2 version does not leave a whitespace delimiter before negative values, which unfortunately breaks downstream file parsers. The proposed revision is to use `6(1PE14.5E3)` which ensures the presence of a whitespace can be recognized by VESTA, ChimeraX, gview, etc.

mkrack:
> [A Chinese discussion](http://bbs.keinsci.com/thread-60030-1-1.html) points out that the new cube file format in the 2026.2 version does not leave a whitespace delimiter before negative values, which unfortunately breaks downstream file parsers. The proposed revision is to use `6(1PE14.5E3)` which ensures the presence of a whitespace can be recognized by VESTA, ChimeraX, gview, etc.

I think `6(1X,ES13.5E3)` is probably cleaner, safer, adheres to modern Fortran standards, and prevents accidental formatting bugs downstream in the code, because `1P` is an old legacy modifier that changes the global state of the Fortran compiler runtime.

he-zilong:
> I think `6(1X,ES13.5E3)` is probably cleaner, safer, adheres to modern Fortran standards, and prevents accidental formatting bugs downstream in the code, because `1P` is an old legacy modifier that changes the global state of the Fortran compiler runtime.

Makes sense. The only problem about `6(1X,ES13.5E3)` I could see is that it is actually 14-character wide rather than 13. I do not know if there are still cube file parsers expecting a strict 13-character width or if they are more commonly expecting at least one whitespace as a delimiter.

If possible, I think the READ commands on the CP2K's side (the main program or the cubecruncher) should not have restriction on the width of each value.

mkrack:
`6(1PE14.5E3)` consumes also 14 characters for each value. When 13 characters are a strict limit, one digit has to be dropped in the mantissa `6(1X,ES12.4E3)` or exponent 6(1X,ES12.5E2)`.

he-zilong:
Sure. The motivation of this PR was to make room for one more digit on the exponent to accommodate the more extreme values from #850. I think most of the time 2 digits on the exponent suffices, and one digit of mantissa can be sacrificed to accommodate a rare 3-digit exponent. I am not sure how much complexity a format switch would introduce, though.

Growl1234:
I'd prefer `6(1X,ES13.5E3)`. It does no harm, and most program such as VESTA, GaussView etc. can be compatible with it.

One of my confusion: This PR deliberately preserved the 13-character field width to retain the existing MPI-IO layout. However, this turned an internal MPI-IO implementation detail into a constraint on the external cube representation and introduced hard-coded assumptions such as `cube_entry_len = 13`. I do not think the design should have been this.

Growl1234:
I've seen from forums that more and more users are being confused about post-processing tools reading cube files produced by v2026.2. Maybe we should backport #5698 and add an v2026.2.1?

he-zilong:
> Maybe we should backport #5698 and add an v2026.2.1?

Good idea! The version number can simply be 2026.3, with an precedent in 2024.

(And a tangential request - could we include #6019 in this minor release as well? It may help users whose NEB task seems to stuck.)

Growl1234:
If I was to backport, I would only carry #5633 (typo), #5646 (fixing a problem already exists before v2026.2 but not fixed then), and #5698. I didn't have a look at #6019, but maybe it's acceptable as it only added one line. Let's leave decision to maintainers.

Edit: #5639 should also be added.

Growl1234:
@oschuett In case you didn't see it, would you be OK with backporting several fixes as a "patch release"?

oschuett:
Backports are tricky because we don't have a good process for them. In particular the CI can currently only run on the master branch. That being said, we did make small backports in the past. 

Since you are aiming for a larger backport, perhaps we should just make a regular release?

Growl1234:
Oh CI is a relevant concern. Then we could only backport #5698.

PR #5251, opened 2026-05-19
Bug fix for TDDFPT oscillator strength for unrestricted calculations

juerghutter:
This fixes a long standing bug in the calculation of oscillator strengths for unrestricted systems.
Transition dipoles were correct, but the calcullation of oscillator strengths used a wrong formula.


DCM-Uni-Paderborn:
I see, this is to say that the cross terms were missing?!?

juerghutter:
Thanks to Nanchen Dongfang for pointing out the problem and for running test calculations of the fix.

- (old)              trans_dipoles(:, :, 1) = SQRT(trans_dipoles(:, :, 1)**2 + trans_dipoles(:, :, 2)**2)

+ (new)           trans_dipoles(:, :, 1) = trans_dipoles(:, :, 1) + trans_dipoles(:, :, 2)

PR #5257, opened 2026-05-20
Motion: Remove &CELL_OPT/TYPE keyword

### LINKED ISSUE #4426: Stress tensor is computed at every step when using CELL_OPT type GEO_OPT
Hello,

I'm trying to do cell optimization (using CP2K v2025.1 currently), and analytical stress tensor is unavailable for my configuration, so I've decided to try `GEO_OPT` type of `CELL_OPT` for time saving instead of `DIRECT_CELL_OPT`. Documentation says "The stress tensor is computed at the optimized geometry", however, it seems not to be the case, and stress tensor is recomputed at every step of geometry optimization (which requires 20 points, and is the reason why I wanted to avoid  `DIRECT_CELL_OPT`).

I've looked into restart file of internal `GEO_OPT` run (PMDA_r2scan_d4_opt_2-GEO_OPT-0-1.restart), and it contained unmodified `STRESS_TENSOR NUMERICAL` statement from my original input, which seems to be a bug. But I couldn't find a code which is responsible for generation of intermediate inputs to check the logic.

I've attached my input and output files.

[PMDA_r2scan_d4_opt_2.tar.gz](https://github.com/user-attachments/files/22227824/PMDA_r2scan_d4_opt_2.tar.gz)

oschuett:
Towards #1233. Closes #4426.


he-zilong:
Mmm, maybe my message muddled more than myself: I mentioned my motivation for managing the _initial_ dimer vector, which would be irrelevant to the deprecation of `&MOTION/&CELL_OPT/TYPE` keyword, or the `gopt_f77_methods.F` workaround for circular dependency in general. So, sorry, this PR is not "in preparation of" my planned work.

oschuett:
No worries. I've updated the description.

I'm still convinced  this is a step in the right direction, which will enable future improvements to the motion code.

he-zilong:
Certainly! That said, although I agree with the removal of functionality, I wonder if it is desirable to make all formerly working input files with the `TYPE` keyword suddenly aborting due to it being unrecongnized now; should we perhaps leave a deprecation notice like the `&MOTION/&MD/&REFTRAJ/EVAL_ENERGY_FORCES` keyword? (But again, I am afraid a deprecation notice would delay severely the actual removal of the keyword. We have been using `&MOTION/&MD/&REFTRAJ/EVAL` long enough that it is time to discard `EVAL_ENERGY_FORCES`.)

Also something I did not spare time to figure out about the #4426 case: what exactly made analytical stress tensor unavailable there? Is it the pure meta-GGA functional r2SCAN, or the DFT-D4 correction, or the GAPW method with the "6-31Gxx" basis set, or other factors? The author of that issue may want to hear about upcoming enhancements or confirming supports for analytical stress tensor.

oschuett:
> should we perhaps leave a deprecation notice 

Good point. I've marked the keyword as _removed_ so users now get this helpful message:

```
 * [ABORT]  The specified keyword 'TYPE' is not available anymore: The keyword *
 *  \___/    MOTION/CELL_OPT/TYPE has been removed because cell optimizations  *
 *    |                       now always use DIRECT_CELL_OPT.                  *
 * / \                                               input/input_parsing.F:262 *
```

he-zilong:
Thanks, I did not know about the `removed=.TRUE.` option. Perhaps this should be applied to other deprecated keywords.

It is weird how the regtests `Fist/regtest-12/argon_opt_cell.inp` and `Fist/regtest-12/argon_opt_cell_ks.inp` formerly with `TYPE GEO_OPT` are now failing with `GEOMETRY wrong or EMAX_SPLINE too small!` due to the switch to `DIRECT_CELL_OPT`. The length of cell vector B jumps from 12.96 to 50.09 at the first step, and some atoms are displaced very far along the y-axis in the trajectory.

oschuett:
> It is weird how the regtests Fist/regtest-12/argon_opt_cell.inp and ... argon_opt_cell_ks.inp ... are now failing

Yeah, turns out the two-point line search was too aggressive for a starting structure so far off.

PR #5261, opened 2026-05-21
Revise static build

mkrack:


oschuett:
Btw, I noticed that many of the Spack tests on the dashboard, e.g. [here](
https://storage.googleapis.com/cp2k-ci/dashboard_spack-psmp-gcc11_report.txt), are currently showing `UNKNOWN` because they exited without printing a `Status:` line.

Growl1234:
I’m a bit hesitant about whether this is the right fix, since both fully static builds of `dftd4` and `tblite`, as well as static linking against them, work fine on their own. The issue seems to lie somewhere in the CMake/Ninja workflow, and may possibly originate from Spack or the standalone `dftd4` package.

Personally, I would suggest bringing back the local recipe that does not depend on other Spack packages, thereby forcing `tblite`’s CMake build to use its own `subprojects` path.

mkrack:
> Btw, I noticed that many of the Spack tests on the dashboard, e.g. [here](https://storage.googleapis.com/cp2k-ci/dashboard_spack-psmp-gcc11_report.txt), are currently showing `UNKNOWN` because they exited without printing a `Status:` line.

It seems these test runs got terminated prematurely during the py-torch build. Maybe, there was a kind of time-out or lack of resources. Are there any log files which could provide more information?

Growl1234:
Or, simply adding
```
    dftd4:
      require:
        - "~shared"
```
could be enough to resolve the problem?

The 
```
    tblite:
      require:
        - "~shared"
```
does not seem to make sense because the Spack recipe of tblite does not include the `shared` flag.

mkrack:
> I’m a bit hesitant about whether this is the right fix, since both fully static builds of `dftd4` and `tblite`, as well as static linking against them, work fine on their own. The issue seems to lie somewhere in the CMake/Ninja workflow, and may possibly originate from Spack or the standalone `dftd4` package.
> 
> Personally, I would suggest bringing back the local recipe that does not depend on other Spack packages, thereby forcing `tblite`’s CMake build to use its own `subprojects` path.

Indeed, this is the price to pay for getting rid of the local spack recipe for `tblite`. I have a preference here for minimizing the number of local spack recipes. The static build is already highly "hacked" and I am not sure how sustainable it will be.

mkrack:
> Or, simply adding
> 
> ```
>     dftd4:
>       require:
>         - "~shared"
> ```
> 
> could be enough to resolve the problem?

That block is already present, isn't it?
> 
> The
> 
> ```
>     tblite:
>       require:
>         - "~shared"
> ```
> 
> does not seem to make sense because the Spack recipe of tblite does not include the `shared` flag.

Sorry, that's a left-over from my tests which could be removed, though it should not have an  effect with `tblite` commented, currently. I tried the `meson` builds with `- "default_library=static"` but I could not get that resolved by `clingo` even with `unify: false`. I always got mismatches in the `meson` and `cmake` dependency trees.

Growl1234:
> That block is already present, isn't it?

Yes, what I mean is — please correct me if I’m wrong — that we may be able to fix the failure by keeping tblite included and only adding these lines.

mkrack:
> > That block is already present, isn't it?
> 
> Yes, what I mean is — please correct me if I’m wrong — that we may be able to fix the failure by keeping tblite included and only adding these lines.

Thanks, maybe, I haven't tried that. I will check.

mkrack:
> Btw, I noticed that many of the Spack tests on the dashboard, e.g. [here](https://storage.googleapis.com/cp2k-ci/dashboard_spack-psmp-gcc11_report.txt), are currently showing `UNKNOWN` because they exited without printing a `Status:` line.

Started that GCC11 test for this PR to see if the behaviour is reproduced.

oschuett:
> Started that GCC11 test for this PR to see if the behaviour is reproduced.

Unfortunately, it's tricky to retain the logs of failed pods in Kubernetes, while still taking advantage of its retry logic.

So, I'm not 100% sure what's going on, but my guess would be that it runs out of memory 😕

mkrack:
> > Started that GCC11 test for this PR to see if the behaviour is reproduced.
> 
> Unfortunately, it's tricky to retain the logs of failed pods in Kubernetes, while still taking advantage of its retry logic.
> 
> So, I'm not 100% sure what's going on, but my guess would be that it runs out of memory 😕

The GCC11 run here made it to the regression tests. When I checked the next time, however, it was again building py-torch. It seems that the runs are often restarted automatically in an unpredictable manner.

Update: Finally, the GCC11 test run finished.

PR #5262, opened 2026-05-21
Allow FIT line search mode for cell optimization

### LINKED ISSUE #5243: FIT line search mode in CG not allowed for cell optimization
As below:
https://github.com/cp2k/cp2k/blob/1dd88fb0fa33a3b47ed2bfb8c043716672359e11/src/motion/cg_utils.F#L103-L114

However, when I simply add the `CASE (ls_fit) CALL linmin_fit(...)` section to these lines as geometry optimization approach does, it seems to me work as expected. Here's an example: [TiO2-cell_opt-cg-fit.zip](https://github.com/user-attachments/files/27972064/TiO2-cell_opt-cg-fit.zip)

Therefore, I'm wondering what prevents `FIT` to be supported as a line search mode for cell optimization

Growl1234:
Closes #5243.

### DIFF
--- a/src/motion/cg_utils.F
+++ b/src/motion/cg_utils.F
@@ -105,6 +105,9 @@ CONTAINS
          CASE (ls_2pnt)
             CALL linmin_2pnt(gopt_env, xvec, xi, g, opt_energy, gopt_param, use_only_grad=.TRUE., &
                              output_unit=output_unit)
+         CASE (ls_fit)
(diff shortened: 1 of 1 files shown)


Growl1234:
It seems the regtest has unintentionally restarted two times, both when it's nearly completed ...

oschuett:
We're using [spot instances](https://docs.cloud.google.com/compute/docs/instances/spot) for the CI. These are much cheaper, but get occasionally preempted during high demand.

PR #5270, opened 2026-05-22
GauXC: HDF5 debug output

franzpoeschel:
Adds optional GauXC HDF5 debug output via DFT%XC%XC_FUNCTIONAL%GAUXC%OUTPUT_PATH. When set, CP2K writes the GauXC molecule and basis-set records through GauXC's native HDF5 writer. The output path is expected to point to an existing directory.

The HDF5 interface is guarded by GAUXC_HAS_HDF5 from gauxc/gauxc_config.f, so the default CP2K toolchain build with GauXC HDF5 disabled still compiles cleanly and aborts explicitly if OUTPUT_PATH is requested without HDF5 support.

Validation performed locally:
- make_pretty.sh -m src/CMakeLists.txt src/cp2k_info.F src/input_cp2k_xc.F src/xc/xc_gauxc_functional.F src/xc/xc_gauxc_interface.F
- git diff --check

### DIFF
--- a/src/xc/xc_gauxc_interface.F
+++ b/src/xc/xc_gauxc_interface.F
@@ -117,6 +117,10 @@ MODULE xc_gauxc_interface
       omp_get_max_threads, &
       omp_set_num_threads
 #endif
+#ifdef GAUXC_HAS_HDF5
+   USE gauxc_external_hdf5_write, ONLY: &
+      gauxc_write_hdf5_record
+#endif
    USE string_utilities, ONLY: &
       uppercase
 #endif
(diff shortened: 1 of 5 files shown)


src/xc/xc_gauxc_interface.F
```diff
@@ -1084,4 +1072,67 @@ END FUNCTION read_pruning_scheme
 
 #endif
 
+! TODO
+#define GAUXC_HAS_HDF5
```
franzpoeschel: TODO

PR #5276, opened 2026-05-23
Correct author name in BASIS_MOLOPT_LnPP1

kirk0830:
Dear CP2K developers,

I just found several small typos in the reference information of this file. 
After checking the original paper, I think the extra "a" and "b" might be superscripts. This PR will remove those superscripts.

Regards,
Yike


oschuett:
I think, you're right and it even happened twice: with _David C. Cantu_ and _Jun Li_. Good catch!

So, if you reopen this PR then I'd be happy to merge it.

https://github.com/cp2k/cp2k/blob/0622d442e4c67e9316425ec80a903993996b2865/data/BASIS_MOLOPT_LnPP1#L6-L10

 ---
image

kirk0830:
@oschuett Thanks, I have reopened it~

PR #5277, opened 2026-05-23
Grid improvements

mtaillefumier:
This PR fixes few issues 

- The previous implementation was allocating scratch memory for each task instead of block which wasted a large amount GPU memory for no performance gain. The new implementation restricts memory allocation to blocks (hab, pab) instead. It also fixes the dashboard issue observed after merging the GPU and HIP backends.

- Fix API encapsulation: make grid_task_list opaque

The top-level `grid_task_list` was exposing internal backend pointers in the public header, breaking encapsulation and forcing all backend headers into the public API namespace.

`GPU` and `DGEMM` backends already use opaque handles. This change makes the top-level wrapper and the remaining backends consistent with GPU and DGEMM.

Benefits:
- Callers cannot depend on internal layout
- Backend changes don't break the API
- Reduced compile-time coupling
- Standard C library design pattern

Any code that directly access `task_list->cpu`, `task_list->gpu`, etc. is violating the abstraction and must be fixed to use the public API instead.

- Minor performance improvements

### DIFF
--- a/src/grid/cpu/grid_cpu_task_list.c
+++ b/src/grid/cpu/grid_cpu_task_list.c
@@ -17,6 +17,7 @@
 #include "grid_cpu_collocate.h"
 #include "grid_cpu_integrate.h"
 #include "grid_cpu_task_list.h"
+#include "grid_cpu_task_list_internal.h"
 
 /*******************************************************************************
  * \brief Comperator passed to qsort to compare two tasks.
@@ -59,7 +60,8 @@ void grid_cpu_create_task_list(
     grid_cpu_free_task_list(*task_list_out);
   }
 
-  grid_cpu_task_list *task_list = malloc(sizeof(grid_cpu_task_list));
+  grid_cpu_task_list_internal *task_list =
+      malloc(sizeof(grid_cpu_task_list_internal));
   assert(task_list != NULL);
 
   task_list->orthorhombic = orthorhombic;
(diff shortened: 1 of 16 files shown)


oschuett:
I just noticed that the CUDA Pascal Regtest is still broken due to #5084. I'll see what I can do about that.

mtaillefumier:
> I just noticed that the CUDA Pascal Regtest is still broken due to #5084. I'll see what I can do about that.

I may simply turn it off for the time being so that I check if the tests failing locally also fail in the CI/CD.

oschuett:
I fixed the build via #5316. There are still 4 wrong results, which we'll just ignore for this PR. Can you please rebase.

mtaillefumier:
@oschuett: I updated the HEAD with the latest master branch. I am puzzled about these failing tests and would be curious to know which tests fail on your side because it is difficult to know where the problem comes from. It depends on MPI and OMP. OMP is not used in the GPU bakcend and MPi will only affect the task and grid partitionings. 
The failing daskboard tests can be turned on again. They work locally.

It is probably wise to open an issue for these failing tests so that I have trace of this problem. I will work on the namespace renaming in a separate PR and hopefully have a very simple memory management system on GPU so that the  wrokspace memory footprint  is minimal.

oschuett:
> I am puzzled about these failing tests and would be curious to know which tests fail on your side because it is difficult to know where the problem comes from. 

I think, the failing tests are fallout from other PRs. Unfortunately, we're currently struggling a bit to keep the dashboard green.

PR #5280, opened 2026-05-24
Allow k-point USE_PREV_WF to survive symmetry refresh when safe

Growl1234:
Dynamic full k-point symmetry refresh currently clears the WFN history unconditionally. This makes USE_PREV_WF fall back to USE_GUESS at every geometry/cell step when atomic k-point symmetry is active.

This PR keeps the previous k-point WFN snapshot only when it can be safely reused. For `USE_PREV_WF`, the snapshot k-points are matched against the current irreducible k-point layout, and the previous WFN is reloaded when the representation is compatible. If compatibility checks fail, the code falls back to the initial guess. Higher-order extrapolation methods such as `ASPC`/`PS` are kept conservative and still require an unchanged k-point layout.

An independent issue is found when try further dealing with the extrapolation: full atomic k-point symmetry with changed PBC image conventions can already produce an incorrect density even with `USE_GUESS`. In such cases `USE_PREV_WF` is conservatively rejected. This is planned to be resolved in a seperated PR in the near future.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -325,6 +325,7 @@ list(
   kpoint_k_r_trafo_simple.F
   kpoint_methods.F
   kpoint_mo_dump.F
+  kpoint_mo_symmetry_methods.F
   kpoint_transitional.F
   kpoint_types.F
   kpsym.F
--- a/src/force_env_methods.F
+++ b/src/force_env_methods.F
@@ -169,8 +169,8 @@ MODULE force_env_methods
    USE qs_mo_types,                     ONLY: mo_set_type
    USE qs_rho_types,                    ONLY: qs_rho_get,&
                                               qs_rho_type
-   USE qs_wf_history_types,             ONLY: qs_wf_history_type,&
-                                              wfi_clear
+   USE qs_wf_history_methods,           ONLY: wfi_handle_kpoint_symmetry_refresh
+   USE qs_wf_history_types,             ONLY: qs_wf_history_type
    USE restraint,                       ONLY: restraint_control
    USE scine_utils,                     ONLY: write_scine
    USE string_utilities,                ONLY: compress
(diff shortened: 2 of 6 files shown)


DCM-Uni-Paderborn:
Sorry, simply too fast for me to review it. Die you experience a problem or another shortcoming?

Growl1234:
Sorry for the inconvenience. On the one hand the tblite path encounters problem, so it seems there are still other issues; on the other hand I would prefer the problem directy with k-point symmetry to be resolved first.

Maybe you could give some more ideas? I would really appreciate it if so.

DCM-Uni-Paderborn:
Dear Shiyang,

the general direction of this PR looks valuable. In particular, moving the k-point MO symmetry transformation out of the Wannier90-specific path and making it available as a shared utility seems like the right abstraction. This logic is not intrinsically tied to one output interface, and it could be useful for other post-SCF or export paths that need a consistent full-grid or symmetry-remapped k-point representation.

The more delicate part is the WFN-history reuse across a dynamic k-point symmetry refresh. That touches several pieces at once: the irreducible k-point layout, atom mappings, basis rotations, PBC image conventions, phases, and the consistency of historical overlap data. For CP2K this should probably be handled very conservatively, because an incorrect reuse path would be much worse than falling back to a fresh initial guess.

It may therefore be better to separate the work into two layers:

1. First, introduce and test the shared k-point MO symmetry transformation utility.
2. Then, in a separate step, extend WFN-history reuse once the compatibility conditions are fully understood and covered by regression tests.

The PR description also mentions that changed PBC image conventions with full atomic k-point symmetry can already lead to an incorrect density even with `USE_GUESS`. That sounds like a more fundamental issue and should probably be resolved independently before relying on WFN-history remapping in those cases.

So overall: the idea is useful and worth pursuing, but the WFN-history part needs a very careful, well-tested implementation path before it should be merged into CP2K.

Best,
Thomas

Growl1234:
Thank you for the reply. As a conservative step forward, I will first submit the change to “Move SCF MO k-point symmetry transform helper out of qs_wannier90” later, and drop the remaining changes until a better solution is found and sufficiently validated.

PR #5281, opened 2026-05-25
Avoid tblite TDDFPT OECORR COULOMB neighborlist crash

marisaschneider75-commits:
## Summary

Avoid calling `build_xtb_coulomb()` in the tblite TDDFPT/OE_CORR path.

## Problem

TDDFPT calculations with orbital energy corrections (`OE_CORR`) and tblite-enabled xTB crash with a segmentation fault in:

```text
tddfpt_oecorr
→ build_xtb_ks_matrix
→ build_xtb_coulomb
→ neighbor_list_iterate
```

The issue originates from the `sab_xtbe` neighbor list not being initialized for the tblite branch. In `qs_neighbor_lists.F`, `sab_xtbe` is only built when:

```fortran id="ztz5ix"
xtb .AND. (.NOT. do_tblite)
```

However, `build_xtb_coulomb()` was still called from `xtb_ks_matrix.F` during the TDDFPT OE_CORR path.

## Current Fix

The patch skips the additional CP2K-internal xTB Coulomb build when `do_tblite` is enabled:

```fortran id="sq2b8n"
IF (xtb_control%coulomb_interaction .AND. &
    (.NOT. xtb_control%do_tblite)) THEN
```

This avoids the segmentation fault and allows the TDDFPT/OE_CORR tblite calculations to complete.

## Notes

This should currently be considered a workaround rather than a fully validated physical fix. It still needs to be investigated whether the corresponding Coulomb/SCC contributions are already fully handled internally by tblite in this TDDFPT/OE_CORR context.

### DIFF
--- a/src/qs_tddfpt2_utils.F
+++ b/src/qs_tddfpt2_utils.F
@@ -73,6 +73,7 @@ MODULE qs_tddfpt2_utils
    USE util,                            ONLY: sort
    USE xc_pot_saop,                     ONLY: add_saop_pot
    USE xtb_ks_matrix,                   ONLY: build_xtb_ks_matrix
+   USE qs_neighbor_lists,               ONLY: build_qs_neighbor_lists  
 #include "./base/base_uses.f90"
 
    IMPLICIT NONE
@@ -635,6 +636,8 @@ CONTAINS
       TYPE(section_vals_type), POINTER                   :: hfx_section, xc_fun_empty, &
                                                             xc_fun_original
       TYPE(tddfpt2_control_type), POINTER                :: tddfpt_control
+      TYPE(mp_para_env_type), POINTER                    :: para_env  
+
 
       CALL timeset(routineN, handle)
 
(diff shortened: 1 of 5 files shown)


DCM-Uni-Paderborn:
Dear Marisa,  

thanks for tracking this down. I looked through the crash path and tested a local follow-up patch.

My read is that the immediate segfault comes from entering the CP2K-internal xTB Coulomb path in a tblite run. `sab_xtbe` is intentionally not built for tblite in `qs_neighbor_lists.F` (`xtb .AND. (.NOT. do_tblite)`), so `build_xtb_coulomb()` must not assume that this neighbor list exists.

I would suggest making the TDDFPT/OE_CORR KS rebuild mirror the normal KS rebuild logic:

1. In `src/qs_tddfpt2_utils.F`, import `build_tblite_ks_matrix`.
2. In `tddfpt_oecorr`, change the xTB branch to:
   - call `build_tblite_ks_matrix(..., ext_ks_matrix=matrix_ks_oep)` when `xtb_control%do_tblite` is true;
   - call `build_xtb_ks_matrix(..., ext_ks_matrix=matrix_ks_oep)` otherwise.
3. Remove the added `build_qs_neighbor_lists(...)` call and the extra `para_env` local from `tddfpt_oecorr`. Rebuilding QS neighbor lists there does not create `sab_xtbe` for tblite anyway.
4. Keep the `build_xtb_coulomb()` association check after `get_qs_env(..., sab_xtbe=n_list)` so any remaining misuse fails with a clear CPABORT instead of a segmentation fault.
5. Keep the `do_tblite` guard around `build_xtb_coulomb()` in `xtb_ks_matrix.F` as an additional defensive guard.
6. Add a small run-only regression input under `tests/xTB/regtest-stda/`, e.g. CH2O with `GFN_TYPE TBLITE`, `TBLITE METHOD GFN1`, `TDDFPT`, and `OE_CORR SHIFT`, registered in `TEST_FILES.toml` with `[]`.

I tested this locally with:
- `git diff --check`
- `./make_pretty.sh -m` on the touched files
- a local CMake build of `cp2k.psmp` without tblite enabled

Greetings, 
Thomas

PR #5282, opened 2026-05-25
Enable analytical periodic-subspace stress for 2D systems with ANALYTIC/MT

Growl1234:
This PR enables analytical stress tensor calculations for 2D periodic systems using the `ANALYTIC` and `MT` Poisson solvers. `WAVELET` remains unsupported because it uses a separate real-space Poisson solver path and requires independent stress validation.

Only the periodic-subspace components are supported and tested. For example, with `PERIODIC XY`, only the xx, xy, yx, and yy components are considered physically meaningful; components involving the non-periodic direction are projected out.

For `CELL_OPT`, the non-periodic direction must be constrained. If no `CONSTRAINT` is set, CP2K sets the matching one automatically, e.g. `CONSTRAINT Z` for `PERIODIC XY`. Explicit mismatches are rejected (e.g. `PERIODIC XY` with `CONSTRAINT` as `NONE` or not including `Z`, while `CONSTRAINT XZ` or `CONSTRAINT YZ` is acceptable).


he-zilong (commented):
It seems to me that you have enabled stress for 2D periodic QS calculations in general. The numerical one is more apparent given the major renovation to `force_env_calc_num_pressure()` in `src/force_env_methods.F`.
Meanwhile, the fundamental infrastrcture of `ANALYTIC` and `MT` Poisson solvers in `src/pw/pw_poisson_methods.F` does not seem to change much. I think the case is arguably better described as "the algorithm was already capable of computing raw data for analytical stress, but the lack of post-treatments such as the appropriate projection rendered it 'not tested' previously." Am I right?

Growl1234:
I think it is essentially right. The ANALYTIC2D/MT2D Poisson stress machinery was already largely capable of producing the periodic-subspace analytical stress. This PR mainly adds the missing projection, validation, and CELL_OPT/debug handling needed to make that result well-defined and tested for 2D systems. The larger change in `force_env_calc_num_pressure` is for the numerical finite-difference check, which must use symmetric strains in the periodic subspace for 2D QS.

he-zilong:
In that case, please remove the word "analytical" from the title of this PR as it may be misleading and/or confusing.

Any plans for stress for 1D periodic QS calculations? Maybe the algorithm of `PERIODIC` and `ANALYTIC` Poisson solvers are already capable too ...?

Growl1234:
> In that case, please remove the word "analytical" from the title of this PR as it may be misleading and/or confusing.

OK I'll consider that, though my main final purpose is to remove the abort message when calculating analytical stress tensor in 2D systems because numerical stress was already available anyways.

> Any plans for stress for 1D periodic QS calculations? Maybe the algorithm of `PERIODIC` and `ANALYTICAL` Poisson solvers are already capable too ...?

In my view it needs seperate validation, so I wouldn't include it in this PR.

src/force_env_methods.F
```diff
@@ -769,56 +776,130 @@ SUBROUTINE force_env_calc_num_pressure(force_env, dx)
...
-               END DO
-               DO i = 1, ncore
-                  CALL real_to_scaled(s, ref_pos_core(i, 1:3), cell_local)
-                  CALL scaled_to_real(core_particles%els(i)%r, s, cell)
-               END DO
-               DO i = 1, nshell
-                  CALL real_to_scaled(s, ref_pos_shell(i, 1:3), cell_local)
-                  CALL scaled_to_real(shell_particles%els(i)%r, s, cell)
+      IF (use_sym_strain_2d) THEN
+         DO ip = 1, 3
```
oschuett: This part looks like a lot of code duplication. Is there maybe another way?

Growl1234: I'll have a look, maybe I could refactor the duplicated code as a seperate subroutine (edit: Adding a subroutine increases rather than reduces lines... so I tried another way).

PR #5286, opened 2026-05-25
Motion: Add citation for BFGS model Hessian

oschuett:


he-zilong:
I think refactoring the subroutine `construct_initial_hess` out of `bfgs_optimizer.F` and into a standalone module would be nice! It may also be expanded with other algorithms for generating a model Hessian as used by [ORCA](https://www.faccts.de/docs/orca/6.1/manual/contents/structurereactivity/optimizations.html#initial-hessian-for-minimization).

oschuett:
Yes, I think this is a worthwhile area to invest in. While I don't have first-hand experience, it's my understanding that already the simple model by Lindh et al. can lead to significant speedups. And given the current AI hype, I'd expect the topic to attract re-newed attention.

PR #5287, opened 2026-05-25
Add broadened DOS/PDOS output with shared OT LUMO generation

Growl1234:
This PR adds broadened DOS/PDOS output that can be plotted directly from CP2K output files.

Previously, `PRINT%DOS` wrote a discrete histogram-like DOS, while `PRINT%PDOS` wrote MO-resolved projection weights. This PR adds Gaussian, Lorentzian, and pseudo-Voigt broadening for DOS and PDOS.

Main changes:

- Add `BROADEN_TYPE` (Gaussian, Lorentzian, or pseudo-Voigt), `BROADEN_WIDTH`, and `VOIGT_MIXING` (Lorentzian fraction of pseudo-Voigt broadening) to `PRINT%DOS` and `PRINT%PDOS`.
- Write broadened DOS curves to `*.dos`.
- Write broadened PDOS curves to `*.pdos` and move the previous MO-resolved PDOS projection-weight output to `*.pdos_raw`.
- Add `NDIGITS` to `PRINT%PDOS`; DOS/PDOS curve output now defaults to 6 digits.
- Share the broadening implementation between DOS and PDOS through `qs_dos_utils`.

Also improves virtual-orbital handling:

- When DOS, PDOS, and Molden request OT LUMOs simultaneously, the maximum requested number is generated once and shared, avoiding repeated calls to the LUMO eigensolver.
- For diagonalization runs, if `NLUMO` in DOS/PDOS setting is larger than `ADDED_MOS`, then `ADDED_MOS` would be increased to the value of `NLUMO`.

P.S. K-point PDOS is left for future work, as its implementation requires the Lowdin-transformed MO coefficients, \(S(k)^{1/2} C(k)\). This is closely related to the key ingredient needed for Lowdin DFT+U with k-points, though the PDOS implementation is simpler because it is only a post-SCF projection and does not feed back into the SCF cycle.


he-zilong:
> Previously, `PRINT%DOS` wrote a discrete histogram-like DOS, while `PRINT%PDOS` wrote MO-resolved projection weights. This PR adds Gaussian, Lorentzian, and pseudo-Voigt broadening for DOS and PDOS.
> ...
> * Write broadened DOS curves to `*.dos`.
> * Write broadened PDOS curves to `*.pdos`.
> * Move the previous MO-resolved PDOS projection-weight output to `*.pdos_raw`.
> ...

Despite in favor of the general idea, I do not like this implementation due to potentially breaking workflows already adapted to the format of existing outputs (`*.dos`, `*.pdos`, etc, however inconvenient they may feel). Please keep them intact and write new data to new files (for instance, with suffixes `*-broadened.dos` and `*-broadened.pdos`).

> For OT calculations, this PR also improves virtual-orbital handling and contains breaking changes:
> 
> * Rename `PRINT%PDOS%NLUMO` to `PRINT%PDOS%OT_NLUMO`.
> * Add `OT_NLUMO` to `PRINT%DOS`.
> * When DOS, PDOS, and Molden request OT LUMOs simultaneously, the maximum requested number is generated once and shared, avoiding repeated calls to the LUMO eigensolver.
> * For diagonalization runs, DOS/PDOS use the orbitals already present in the converged MO set, e.g. from `ADDED_MOS`, no longer `NLUMO`.

Wouldn't it be more user-friendly and backwards-compatible if OT and diagonalization are distinguished only in the code backend, and not in the interface of input keyword `NLUMO`? Given the treatment of OT LUMO, it may also feel that it is expected to see the maximum requested number of `ADDED_MOS` and `NLUMO` taken for diagonalization...

Growl1234:
Thank you, I will re-consider that. It seems I'll first need to revert some changes, and moreover the `OT_NLUMO` in `MO_MOLDEN` can also be dealt with like what you suggested. In this case, `make_lumo_gpw` that using OT eigensolver will work not only for OT calculations, and I think we should also give a warning message along with the diagonalization task setting `NLUMO` > `ADDED_MOS`.

As of filename, I would prefer to keep the idea of replacing the major output and printing legacy output to a `.dos_raw` or `.pdos_raw` file (I can add back printing unbroadened DOS file).

PR #5294, opened 2026-05-26
toolchain/build_cp2k.sh: Improve logging and add "--rebuild-only" flag 

### LINKED ISSUE #5293: build log write / rebuild
@Growl1234 

Would you be opposed to removing the piping in 
https://github.com/cp2k/cp2k/blob/ba924cc32cd9d0b6b85799a4bea446b7374db93d/tools/toolchain/build_cp2k.sh#L252
so that the output and errors are written to standard output and can be observed during build?

Growl1234:
Closes #5293.

CC: @jpoto


Growl1234:
The `--rebuild-only` flag is added. With this flag one can rebuild CP2K without rerunning CMake configuration, so that only modified file(s) and possibly some other files affected by it/them will be re-compiled. It requires a successful build before and is not compatible with flags that affect CMake configuration.

PR #5295, opened 2026-05-26
NNP: near-linear CPU scaling with cell-list cache and OpenMP

DhruvSkyy:
Optimises CP2K's native NNP implementation via a linked-cell neighbour cache, a Verlet skin, spline-tabulated ACSF, a touched-atom force scatter, and OpenMP coverage of the per-atom loops.

Builds on the original NNP code by Prof. Christoph Schran (@cschran). The user-facing interface and input files are unchanged. The path-integral helium-NNP coupling has been updated to call the same new helpers so the two code paths stay aligned, and all existing NNP regtests continue to pass at machine precision.

Adds a bit-exact regression test (`H2O-256_C-NNP_MD-bitexact.inp`) to catch integrator and neighbour-cache drift.

Writes `docs/methods/machine_learning/nnp.md`.

The figure below shows single-rank wall time per MD step on bulk water from 192 to 3072 atoms. The upstream path grows super-linearly with system size while the optimised path scales near-linearly and tracks LAMMPS+n2p2 closely.

Single-rank wall time per MD step on bulk water versus atom count, comparing CP2K master, the optimised path and LAMMPS+n2p2.

**API note:** `nnp_calc_acsf`'s third argument changed from `dsymdxyz` (optional `REAL(:,:,:)`) to `calc_forces` (`LOGICAL`). In-tree callers are all updated; external/fork code calling it directly will hit a compile-time type-mismatch error (not a silent change) and needs its call updated.

cc @ClaudioRMalvino


src/nnp_environment.F
```diff
@@ -689,18 +689,144 @@ SUBROUTINE nnp_init_model(nnp_env, printtag)
...
+      ! Bias the standard deviation of committee disagreement
+      NULLIFY (bias_section)
+      explicit = .FALSE.
+      !HELIUM NNP does atm not allow for bias (not even defined)
+      bias_section => section_vals_get_subs_vals(nnp_env%nnp_input, "BIAS", can_return_null=.TRUE.)
+      IF (ASSOCIATED(bias_section)) CALL section_vals_get(bias_section, explicit=explicit)
+      nnp_env%bias = .FALSE.
+      IF (explicit) THEN
+         IF (nnp_env%n_committee > 1) THEN
+            IF (logger%para_env%is_source()) THEN
```
fstein93: Please check `unit_nr>0` instead. You may need to move fetching the unit_nr in line 258 outside of the branch and use the same check for the left-over branch.

src/nnp_environment.F
```diff
@@ -689,18 +689,144 @@ SUBROUTINE nnp_init_model(nnp_env, printtag)
...
+            CALL section_vals_val_get(bias_section, "ALIGN_NNP_ENERGIES", explicit=explicit)
+            nnp_env%bias_align = explicit
+            IF (explicit) THEN
+               NULLIFY (work)
+               CALL section_vals_val_get(bias_section, "ALIGN_NNP_ENERGIES", r_vals=work)
+               IF (SIZE(work) /= nnp_env%n_committee) THEN
+                  CPABORT("ALIGN_NNP_ENERGIES size mismatch wrt committee size.")
+               END IF
+               nnp_env%bias_e_avrg(:) = work
+               IF (logger%para_env%is_source()) THEN
```
fstein93: Same

src/nnp_environment.F
```diff
@@ -689,18 +689,144 @@ SUBROUTINE nnp_init_model(nnp_env, printtag)
...
+                  WRITE (unit_nr, *) TRIM(printtag)//"| Biasing is aligned by shifting the energy prediction of the C-NNP members"
+               END IF
+            END IF
+         ELSE
+            CPWARN("NNP committee size is 1, BIAS section is ignored.")
+         END IF
+      END IF
+
+      IF (logger%para_env%is_source()) THEN
```
fstein93: Same

tests/matchers.py
```diff
@@ -78,6 +78,54 @@ def run(self, output: str, **kwargs: Any) -> MatchResult:
         return MatchResult("OK", error=None, value=value)  # passed
 
 
+# ======================================================================================
+class FirstMatcher(Matcher):
+    """Like GenericMatcher but reads the FIRST matching line instead of the last.
```
oschuett: Instead of duplicating the entire class, could you please just add a flag `first: bool = False`, like we did for regular expressions.

DhruvSkyy: Hi Ole, I think I have done that now.

DCM-Uni-Paderborn:
Is there still work in progress? The PR is still in draft mode ... 

Greetings, 
Thomas

cschran:
## Code review: correctness, memory, and style issues

Detailed review of the full diff. Issues are labelled to match the summary table in the accompanying review document.

---

### M1 — Spurious `pack_eta` guard on radial symfgrp release

`nnp_environment_types.F`, `nnp_env_release`, radial loop body.  
`pack_eta` is only allocated for **angular** groups (in `nnp_init_acsf_groups`). The guard for it in the radial release block is always false and misleads readers into thinking radial groups might have pack arrays. Either remove it or add a comment.

```fortran
! Current (nnp_environment_types.F ~line 453):
            IF (ALLOCATED(nnp_env%rad(i)%symfgrp(j)%pack_eta)) &
               DEALLOCATE (nnp_env%rad(i)%symfgrp(j)%pack_eta)
```

```fortran
! Suggested: remove the dead guard entirely. pack_eta is angular-only.
! The IF (ALLOCATED(spline_y/spline_dy)) guards below are sufficient.
```

---

### M2 — Missing `IF (ALLOCATED)` guards on base symfgrp arrays

`nnp_environment_types.F`, `nnp_env_release` ~line 450.  
`symf`, `ele`, and `ele_ind` are deallocated unconditionally for every symfgrp. If `nnp_init_acsf_groups` was never called (early error exit from `nnp_init_model`), this crashes. All three are allocated together, so one guard suffices.

```fortran
! Current (crashes if nnp_init_acsf_groups never ran):
            DEALLOCATE (nnp_env%rad(i)%symfgrp(j)%symf, &
                        nnp_env%rad(i)%symfgrp(j)%ele, &
                        nnp_env%rad(i)%symfgrp(j)%ele_ind)
```

```fortran
! Suggested:
            IF (ALLOCATED(nnp_env%rad(i)%symfgrp(j)%symf)) &
               DEALLOCATE (nnp_env%rad(i)%symfgrp(j)%symf, &
                           nnp_env%rad(i)%symfgrp(j)%ele, &
                           nnp_env%rad(i)%symfgrp(j)%ele_ind)
```

Same change for the angular block (`nnp_env%ang(i)%symfgrp(j)%symf/ele/ele_ind`).

---

### M3 — `nnp_workspace_grow_caches` discards old values without comment

`nnp_neighbor_interface.F` ~line 445.  
The four cutoff caches are deallocated and reallocated without preserving old content, unlike `nnp_neigh_grp_grow` which uses `MOVE_ALLOC`. This is **correct** — `nnp_fill_fc_dfc_cache` overwrites them completely before any read — but the inconsistency is confusing without a brief note.

```fortran
! Current (no explanation):
      new_cap = MAX(MAX(8, n_needed), INT(workspace%cache_cap*1.5_dp) + 8)
      IF (ALLOCATED(workspace%fc_cache1)) DEALLOCATE (workspace%fc_cache1)
      ...
```

```fortran
! Suggested: add a single-line comment explaining the discard is safe.
      new_cap = MAX(MAX(8, n_needed), INT(workspace%cache_cap*1.5_dp) + 8)
      ! Contents not preserved: nnp_fill_fc_dfc_cache overwrites fully before any read.
      IF (ALLOCATED(workspace%fc_cache1)) DEALLOCATE (workspace%fc_cache1)
      ...
```

---

### D1/C1 — Missing justification for standalone Hermite spline implementation

`nnp_acsf.F`, `nnp_build_radial_splines` `\brief` ~line 1222.  
The custom cubic Hermite tables (`spline_y(n_symf, n_grid)`) deliberately bypass CP2K's `splines_methods` module. The performance reason (sf-first layout enables `!$OMP SIMD` across SFs) is valid but absent from the docstring. Without it, a future reviewer will try to "clean this up" by switching to `splines_methods`.

```fortran
! Current \brief (8-line wall of prose, missing the key constraint):
!> \brief Build per-radial-group Hermite cubic spline tables.
!>        Each radial group s carries its own dense (n_symf, n_grid) tables
!>        spline_y and spline_dy. Within a group all SFs share the cutoff
!>        => same uniform grid, so the inner SF loop in nnp_calc_rad can
!>        compute the Hermite basis params ONCE and stream the per-SF y/dy
!>        values contiguously (sf-first layout).
!>
!>        Each table stores y(r) = EXP(-eta_sf*(r-rs_sf)^2)*fcut(r) ...
```

```fortran
! Suggested: condense to 2 lines and add the key constraint note.
!> \brief Build sf-first (n_symf, n_grid) Hermite cubic spline tables for radial SFs.
!>        Uses a custom sf-first layout (not splines_methods) so nnp_calc_rad can
!>        stream contiguous SF values under !$OMP SIMD without indirect addressing.
!> \param nnp ...
```

---

### D2 — Inlined angular kernel lacks cross-reference to `nnp_calc_ang`

`nnp_acsf.F`, serial ELSE branch of the angular loop, ~line 226.  
The serial path inlines the full angular force kernel to eliminate call overhead. The OMP path calls `nnp_calc_ang`. These must stay in sync. There is currently no comment pointing readers to `nnp_calc_ang` as the reference implementation.

```fortran
! Current: no cross-reference (appears at the start of the inline kernel ~line 268):
                  IF (do_forces) THEN
                     ...
                     pref_lam_il = zeta_il*tmpzeta_il*lam_il*inv_g2_il
```

```fortran
! Suggested: add one line before the inlined kernel begins.
                  IF (do_forces) THEN
                     ! Inlined equivalent of nnp_calc_ang; OMP path calls it directly.
                     ...
                     pref_lam_il = zeta_il*tmpzeta_il*lam_il*inv_g2_il
```

---

### F1 — `DEFAULT(SHARED)` in OpenMP region

`nnp_acsf.F` ~line 748.  
The existing comment at lines 747–748 correctly documents why `DEFAULT(SHARED)` is used instead of `DEFAULT(NONE)`: listing an absent OPTIONAL dummy (`stress`) in an OpenMP clause is non-conforming on some compilers. The workaround is sound. However, the comment can be made more explicit, and an alternative that enables `DEFAULT(NONE)` is shown below.

```fortran
! Current comment (correct but terse):
         ! DEFAULT(SHARED) (not NONE) keeps the OPTIONAL `stress` dummy out of
         ! the data-sharing clauses; the PRESENT() guard alone gates its use.
!$OMP PARALLEL DEFAULT(SHARED) &
```

```fortran
! Option A: improve the comment to state the Fortran standard reason.
         ! DEFAULT(NONE) is avoided because listing an absent OPTIONAL dummy in any
         ! OpenMP clause is undefined behavior (F2018 §15.5.2.13). PRESENT() gates
         ! the actual use, so SHARED is safe here.
!$OMP PARALLEL DEFAULT(SHARED) &
```

```fortran
! Option B: capture PRESENT() before the region and use DEFAULT(NONE).
! Add before the ASSOCIATE block:
      LOGICAL :: do_stress
      REAL(KIND=dp), DIMENSION(:, :, :), POINTER :: stress_ptr
      do_stress = PRESENT(stress)
      NULLIFY (stress_ptr)
      IF (do_stress) stress_ptr => stress

! Then in the OMP clause:
!$OMP PARALLEL DEFAULT(NONE) &
!$OMP          SHARED(nnp, ind, self_dGdr, off, workspace, neighbor, &
!$OMP                 cache_cap_loc, max_ang_symf_loc, do_stress, stress_ptr) &
!$OMP          PRIVATE(...)
         ...
         IF (do_stress) THEN
            stress_ptr(:, l, m) = stress_ptr(:, l, m) - ...
```

Option A is the lower-risk change; Option B is more conforming if `DEFAULT(NONE)` is a hard project requirement.

---

### F3 — Missing `timeset`/`timestop` in non-trivial subroutines

Two subroutines have non-trivial work and no timing instrumentation, contrary to CP2K convention.

**`nnp_build_radial_splines`** (called once per NNP init, but loops over all elements × groups × grid points):

```fortran
! Current: no timing
   SUBROUTINE nnp_build_radial_splines(nnp)
      TYPE(nnp_type), INTENT(INOUT), POINTER             :: nnp
      INTEGER                                            :: ind, k, n_symf, p, s, sf
```

```fortran
! Suggested:
   SUBROUTINE nnp_build_radial_splines(nnp)
      TYPE(nnp_type), INTENT(INOUT), POINTER             :: nnp
      CHARACTER(len=*), PARAMETER :: routineN = 'nnp_build_radial_splines'
      INTEGER                                            :: handle, ind, k, n_symf, p, s, sf
      ! ... existing body ...
      CALL timestop(handle)
   END SUBROUTINE nnp_build_radial_splines
```

**`nnp_neighbor_interface_prepare`** (called once per force eval from `nnp_prepare_neighbor_cache`):

```fortran
! Current: no timing
   SUBROUTINE nnp_neighbor_interface_prepare(nnp)
      TYPE(nnp_type), INTENT(INOUT)                      :: nnp
      LOGICAL                                            :: rebuild
```

```fortran
! Suggested:
   SUBROUTINE nnp_neighbor_interface_prepare(nnp)
      TYPE(nnp_type), INTENT(INOUT)                      :: nnp
      CHARACTER(len=*), PARAMETER :: routineN = 'nnp_neighbor_interface_prepare'
      INTEGER                                            :: handle
      LOGICAL                                            :: rebuild
      ! ... existing body ...
      CALL timestop(handle)
   END SUBROUTINE nnp_neighbor_interface_prepare
```

---

### F4 — Spline lazy-build sentinel checks current-element group 1

`nnp_acsf.F` ~line 127.  
`nnp_build_radial_splines` builds **all** elements at once, but the rebuild trigger checks only element `ind`'s first group. If `ind` happens to have `n_symfgrp == 0` for all atoms in an MPI rank's partition, the build is never triggered even though other elements may need it. (In practice this can't happen because the rank would then compute no ACSF for that element, but the invariant isn't obvious.)

A more robust sentinel is element 1, group 1, checked unconditionally:

```fortran
! Current:
      IF (nnp%rad(ind)%n_symfgrp > 0) THEN
         IF (.NOT. nnp%rad(ind)%symfgrp(1)%spline_built) CALL nnp_build_radial_splines(nnp)
      END IF
```

```fortran
! Suggested: use element 1 as the global sentinel (build is all-or-nothing).
      IF (nnp%n_ele > 0 .AND. nnp%rad(1)%n_symfgrp > 0) THEN
         IF (.NOT. nnp%rad(1)%symfgrp(1)%spline_built) CALL nnp_build_radial_splines(nnp)
      END IF
```

---

### F5 — Precondition for `nnp_scatter_dgdr_to_forces` should be a `CPASSERT`

`nnp_force.F` ~line 707.  
The docstring correctly states that `nnp_calc_acsf(..., calc_forces=.TRUE.)` must have run for atom `i` before `nnp_scatter_dgdr_to_forces` is called. There is no runtime check. A new caller that gets the ordering wrong will read uninitialised `self_dGdr`/`dGdr_rad`/`dGdr_ang_jj`/`dGdr_ang_kk` silently.

A minimal guard using the existing `initialized` flag:

```fortran
! Suggested: add at the top of nnp_scatter_dgdr_to_forces after the ASSOCIATE.
      CPASSERT(nnp%neighbor_interface_state%initialized)
```

This doesn't verify that `calc_forces=.TRUE.` was used (no flag exists for that), but it at least catches the case where the workspace was never built at all. If a `forces_computed` flag on `nnp_neighbor_workspace_type` is too heavy, document the constraint more visibly with `CPABORT` text:

```fortran
! Alternative: document with an explicit abort path.
      IF (.NOT. nnp%neighbor_interface_state%initialized) &
         CPABORT("nnp_scatter_dgdr_to_forces: call nnp_prepare_neighbor_cache first")
```

---

### S1–S5 — Comment style: multi-line inline blocks

CP2K convention is one-line inline comments for non-obvious WHY only. Several new subroutines use multi-line prose blocks in the code body. A representative sample with suggested condensations:

**`nnp_acsf.F` ~line 47–56 (module-level block, no doxygen):**
```fortran
! Current: 9-line design-rationale prose block at module scope
! (between the USE statements and IMPLICIT NONE)
```
```fortran
! Suggested: remove. The rationale belongs in nnp_build_radial_splines's \brief.
```

**`nnp_calc_ang`, pass labels ~line 1431–1456:**
```fortran
! Current (3–5 line explanation per pass):
      ! ---- Pass 1: angular base + tmpzeta (scalar tight loop) ----
      ! Branchless cusp handling: when tmp <= 0 we set tmpzeta_arr = 0, which
      ! propagates to both angular_arr and pref_lam in pass 3, giving
      ! sym = 0 and force = 0 without an explicit CYCLE branch.
```
```fortran
! Suggested (one line each):
      ! Pass 1: branchless cusp clamp -- zero tmpzeta propagates to sym and pref_lam.
```

**`nnp_force.F` ~line 148–150 (inline within `nnp_calc_energy_force`):**
```fortran
! Current:
         ! Per-neighbour dGdr lives in the per-element workspace; the force scatter
         ! walks the per-(ind) neighbour lists directly, with no global
         ! (3, n_sf, num_atoms) slab.
```
```fortran
! Suggested:
         ! dGdr lives in per-element workspace; no global (3, n_sf, num_atoms) slab.
```

---

### B1 — API break: `nnp_calc_acsf` signature change

The third argument changed from `dsymdxyz OPTIONAL REAL(:,:,:)` to `calc_forces LOGICAL`. All internal callers are updated (confirmed by grep). Please note this in the PR description so downstream forks that call `nnp_calc_acsf` directly know to update their call sites — it will fail at compile time with a type mismatch, not silently.

cschran:
**Bug fix: stale neighbor chain with `VERLET_SKIN 0`**

When `VERLET_SKIN` is 0, the Verlet displacement check never triggers a full rebuild, so `nnp_cell_list_rebin_chain` was also never called — leaving the head/next chain stale. Fix: remove the `skin > 0` guard so the rebin always runs on an initialised cache. `src/nnp_cell_list.F`, line 112:

```fortran
-ELSE IF (skin > 0.0_dp .AND. cache%initialized) THEN
+ELSE IF (cache%initialized) THEN
```

**Additional validation**

Benchmarked against master across four systems (H2O-64, H2O-256, NaCl-H2O unit cell, NaCl-H2O 2×2×2 supercell): speedups of 1.6–4.3× on `nnp_calc_energy_force`, with results numerically identical to master. Also ran NPT for NaCl-H2O against the RPA C-NNP model from [arXiv:2311.01527](https://arxiv.org/abs/2311.01527), confirming the published density.

This is otherwise ready to be moved out of draft.

src/nnp_environment_types.F
```diff
@@ -221,25 +259,154 @@ MODULE nnp_environment_types
...
+!> \author Dhruv Sharma (ds2173@cam.ac.uk)
+! **************************************************************************************************
+   TYPE nnp_neighbor_workspace_type
+      INTEGER                                             :: max_rad_symf = 0
+      INTEGER                                             :: max_ang_symf = 0
+      INTEGER                                             :: n_input_nodes = 0
+      ! High-water mark of the per-element angular-cache slab, tracked so the 1D
+      ! cutoff caches settle at the true per-element peak.
+      INTEGER                                             :: cache_cap = 0
+      TYPE(nnp_neighbor_type)                             :: neighbor
```
fstein93: You can default-initialize it to `nnp_neighbor_type()`. For the time being, you need to apply the workaround similar to [here](https://github.com/cp2k/cp2k/blob/f6e87dc039703dfc5ff5fb1c3ce37e6ee177b07c/src/mp2_types.F#L354).

fstein93 (changes requested):

fstein93 (approved):

oschuett:
Since this is a pretty big addition and we're now in the feature freeze, I'd say we merge this after the release.

oschuett:
@hforbert, could you please take a look at this PR.

hforbert:
> @hforbert, could you please take a look at this PR.

I am, but it's a pretty big change so it takes a bit. :)

src/nnp_acsf.F
```diff
@@ -500,70 +1115,203 @@ END SUBROUTINE nnp_scale_acsf
...
+!>      Vector libm hides EXP latency across the SF loop without needing
+!>      an eta-dedup pre-pass.
+!>   3. SIMD sym scatter: sym(sf) = prefzeta*angular*symtmp*ftot.
+!>   4. Force scatter (only when forces requested). The clamp in pass 1
+!>      forces tmpzeta_arr(sf) = 0 at the cusp, which makes both angular
+!>      AND pref_lam vanish naturally -- no per-iter branch.
+!>
+!> Caller passes precomputed cutoff values for the j/k legs; the j-k leg
+!> is computed inline. Geometry-only derivative bases are hoisted outside
+!> the SF loops as before.
```
hforbert, line 1314: One should also mention here, that this subroutine is inlined as well (in nnp_calc_acsf i believe) and
changes here therefore also need to be considered there. A comment at the inlining site already exists.

hforbert (changes requested):
Other than one comment on an inlined routine, I had not time to really look at the changes. I did however run some additional tests and things look fine. Timing does of course not always improve (one of my test cases it got worse by 20%, but that's an extreme case with lots of tiny systems, in reasonably large systems this change improves speed noticeably. I intend to
do some additions related to this code soon, so its better this is in already when I do... :)

oschuett:
>  I intend to do some additions related to this code soon, so its better this is in already when I do... :)

I read this as approval and will therefore merge this PR once the requested comment has been added.

hforbert (approved):

hforbert:
> > I intend to do some additions related to this code soon, so its better this is in already when I do... :)
> 
> I read this as approval and will therefore merge this PR once the requested comment has been added.

correct, but I wanted to wait for the comment change :)

DCM-Uni-Paderborn:
Update on the merge-readiness work:

- Rebased the nine PR commits onto the latest `master` and resolved the `tests/matchers.py` conflict by retaining both the upstream MTLR matcher and the new NNP matchers.
- Ran the current CP2K formatter. It produced only style-only changes in three Fortran files, committed separately as `8473533904`.
- Verified complete serial and MPI/OpenMP builds with GCC/GFortran 16, and separately tested the new matcher first/last-occurrence behavior.
- Compared the concurrently rebased contributor branch with the local series using `range-diff`; all nine PR commits were patch-equivalent before the guarded force-push.
- The PR is now conflict-free and GitHub reports it as mergeable. All review threads remain resolved and both approvals are still present.

Current CI status: Git History, Precommit, Misc, and Manual generation have passed. Only `Regtest pdbg` is still running. Once that succeeds, the PR should be ready to merge.

PR #5296, opened 2026-05-26
autodetection reference functional

jpoto:
This should enable some common functionals, otherwise raise an error. 

extension of #5178

Any comments?


src/qs_dispersion_utils.F
```diff
@@ -93,12 +219,20 @@ SUBROUTINE qs_dispersion_env_set(dispersion_env, xc_section)
...
-                  CALL cp_abort(__LOCATION__, "D2 vdW REFERENCE_FUNCTIONAL expected a second parameter "// &
-                                "(example REFERENCE_FUNCTIONAL PBE). "// &
-                                "Go to https://www.chemie.uni-bonn.de/grimme/de/software/dft-d3/ "// &
-                                "for a full list of supported functionals")
+                  CALL xc_functional_to_d_name(xc_functional_name, dispersion_env%ref_functional)
+                  IF (dispersion_env%ref_functional == "none") THEN
+                     CALL cp_abort(__LOCATION__, &
+                                   "D2 vdW: XC_FUNCTIONAL not supported. "// &
+                                   "Set REFERENCE_FUNCTIONAL explicitly. "// &
+                                   "Go to https://www.chemie.uni-bonn.de/grimme/de/software/dft-d3/ "// &
```
fstein93, line 227: Is this supposed to be D2 or D3? The link points to D3.

jpoto: I didn't create it, but it is in the D2 branch and as far as I understand it points to the D3 list, because they use the same (or similar) functionals.

PR #5297, opened 2026-05-26
TREXIO: Export full k-point grids from symmetry-reduced SCF (S. Azadi)

DCM-Uni-Paderborn:
This adds an optional TREXIO export path for calculations that use k-point
symmetry reduction in the SCF.

By default, TREXIO keeps the current behavior and writes the SCF k-point object
as-is. With the new `DFT%PRINT%TREXIO%FULL_KPOINT_GRID` keyword, CP2K instead
builds a temporary unreduced export grid and writes the full k-point mesh to
TREXIO.

The full-grid MO coefficients are reconstructed from the symmetry-reduced SCF
orbitals using the existing Wannier90 SCF-MO reuse machinery. If reconstruction
is not possible, CP2K falls back to diagonalizing the full export grid.

Summary:
- add `DFT%PRINT%TREXIO%FULL_KPOINT_GRID`
- add `DFT%PRINT%TREXIO%REUSE_SCF_MOS`
- reuse the Wannier90 SCF-MO reconstruction logic for TREXIO full-grid export
- keep default TREXIO behavior unchanged
- add a TREXIO regtest input for symmetry-reduced k-point SCF with full-grid export

Testing:
- `git diff --check HEAD^ HEAD`
- CP2K file property checks
- CP2K Fortran/input prettifiers
- `cmake --build build-trexio-pr --target cp2k-bin -j 4`
- local non-TREXIO runtime parser check for the new input

Note:
The local machine does not currently have TREXIO/HDF5 installed, so the
`__TREXIO` path and the new TREXIO regtest still need to be exercised by a
TREXIO-enabled build/CI.

### DIFF
--- a/src/input_cp2k_print_dft.F
+++ b/src/input_cp2k_print_dft.F
@@ -729,7 +729,7 @@ CONTAINS
 
       CALL section_create(subsection, __LOCATION__, name="TREXIO", &
                           description="Write a TREXIO file to disk.", &
-                          n_keywords=1, n_subsections=0, repeats=.FALSE.)
+                          n_keywords=4, n_subsections=0, repeats=.FALSE.)
       CALL keyword_create(keyword, __LOCATION__, name="FILENAME", &
                           description="Body of Filename for the trexio file.", &
                           usage="FILENAME {name}", default_c_val="TREXIO", &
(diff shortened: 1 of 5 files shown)


Growl1234:
Dear Thomas,

I'd like to ask that would the "full-grid MO coefficients reconstructed from the symmetry-reduced SCF orbitals" likely be useful for not only trexio but also some other features? I am really interested in the general idea provided in this PR.

Best,
Shiyang

DCM-Uni-Paderborn:
Dear Shiyang,

yes, I think the general idea should be useful beyond TREXIO. Conceptually it is not TREXIO-specific: if the SCF was done on a symmetry-reduced k-point set, we can reconstruct the MO coefficients on the corresponding full k-point grid and hand that full-grid representation to a consumer that expects unreduced k-points.

At the moment this PR keeps the implementation deliberately narrow and reuses the machinery that was already developed for the Wannier90 path. For TREXIO this is enough, because it only needs a temporary full-grid representation for export. If other CP2K features would benefit from the same object, I agree it would make sense to factor this into a more general helper/API rather than keeping it tied to TREXIO or Wannier90.

The main cases I have in mind are post-SCF/export interfaces that need full-grid wavefunctions but should still profit from symmetry reduction during the SCF. For anything that changes the Hamiltonian or requires a genuinely new calculation on the full grid, the fallback full-grid diagonalization path is still the safer route.

So my preference would be: keep this PR focused on TREXIO, but treat it as a first consumer of a more general k-point/MO reconstruction utility. If there is another concrete feature that needs it, we can refactor the shared logic out cleanly in a follow-up.

Best,
Thomas

PR #5299, opened 2026-05-26
Implement k-point PDOS with Lowdin projections

Growl1234:
The implementation uses Lowdin-transformed k-point MO coefficients, $S(k)^{0.5} C(k)$, to compute atomic-kind and angular-momentum projected weights. Contributions are accumulated with k-point weights and reduced across k-point groups before writing the broadened PDOS curves.

`*.pdos_raw` and `LDOS`/`R_LDOS` are still left unavailable.


oschuett:
Looks good! How about adding a test?

Growl1234:
[QS/regtest-kp-3/cc9.inp](https://github.com/Growl1234/cp2k/blob/dos/tests/QS/regtest-kp-3/cc9.inp) is modified as a test to see whether PDOS can be generated without error.

PR #5301, opened 2026-05-27
Avoid full k-point symmetry for noncanonical cells

DCM-Uni-Paderborn:
## Summary

This PR makes k-point symmetry reduction conservative for cells that do not follow CP2K's lower-triangular cell convention.

- fall back to `KPOINTS%INVERSION_SYMMETRY_ONLY` when full atomic k-point symmetry is requested for a `MONKHORST-PACK` or `MACDONALD` grid with a non-lower-triangular cell matrix
- add a regression test with a noncanonical cell that verifies the fallback path and the inversion-reduced k-point count
- rewrite the existing `Zn_fcc_gfn2_kp_k290.inp` xTB test using `ABC`/`ALPHA_BETA_GAMMA`, so it continues to exercise the intended full K290 point-group reduction

## Motivation

The full atomic k-point reduction currently assumes that the cell matrix follows CP2K's internal convention, namely `A` along the x-axis and `B` in the xy-plane. Users often paste valid primitive lattice vectors from databases or other codes directly into `A`/`B`/`C`; these vectors can be crystallographically valid but noncanonical for CP2K.

For such inputs, full point-group k-point reduction can be unsafe. In the elemental benchmark discussed in #5123, most inputs use noncanonical primitive cells. Some cases abort during group determination, while others complete with visibly wrong reduced-k-point results, including non-neutral integrated electronic densities.

Canonicalizing general input cells would still be useful as a separate, larger input-handling improvement. However, it is not sufficient to fix the present issue by itself: for the Fe benchmark, rewriting the cell in canonical `ABC`/`ALPHA_BETA_GAMMA` form still leaves the full K290-reduced calculation with a non-neutral integrated density. This PR therefore keeps the affected calculations on the robust inversion-only reduction until the full point-group reduction is safe for these cases.

## Validation

Built and tested with current master plus this patch.

- New regression input `h_noncanonical_kpsym.inp`
  - emits the new fallback warning
  - `Number of Special K-points: 4`
  - `ENERGY| Total FORCE_EVAL ( QS ) energy [hartree] = -4.349166468900473`

- Spark Fe benchmark from the #5123 discussion
  - current master, noncanonical full atomic symmetry: `E = -123.535092735502715 Ha`, integrated density error `-0.0186632691 e`
  - current master, canonical `ABC`/`ALPHA_BETA_GAMMA` full atomic symmetry: `E = -123.532789635614549 Ha`, integrated density error `-0.0272115182 e`
  - full/reference run: `E = -123.542545786079799 Ha`, neutral integrated density
  - patched fallback run: `E = -123.542545786079799 Ha`, neutral integrated density, 2048 k-points

- Representative inputs from the external elemental benchmark
  - `ag`: previously aborts in full group determination; patched run falls back to inversion and reaches the energy step with neutral integrated density
  - `nb`: patched smoke test reaches the energy step with neutral integrated density
  - `fe`: patched fallback matches the full/reference result
  - `co`: patched HCP-class smoke test reaches the energy step with neutral integrated density
  - `po`: canonical control case remains on full k-point symmetry reduction

I also scanned the CP2K test tree for `SYMMETRY ON`, `FULL_GRID OFF`, regular MP/MacDonald grids, and noncanonical `A`/`B`/`C` cells. After converting the existing Zn xTB test to `ABC`/`ALPHA_BETA_GAMMA`, the only remaining match is the new intentional fallback regression test.

Related to #5123 and #5275.


he-zilong:
> [...] for the Fe benchmark, rewriting the cell in canonical `ABC`/`ALPHA_BETA_GAMMA` form still leaves the full K290-reduced calculation with a non-neutral integrated density. 
> [...]
>   * current master, noncanonical full atomic symmetry: `E = -123.535092735502715 Ha`, integrated density error `-0.0186632691 e`
>   * current master, canonical `ABC`/`ALPHA_BETA_GAMMA` full atomic symmetry: `E = -123.532789635614549 Ha`, integrated density error `-0.0272115182 e`
>   * full/reference run: `E = -123.542545786079799 Ha`, neutral integrated density
>   * patched fallback run: `E = -123.542545786079799 Ha`, neutral integrated density, 2048 k-points

Frankly I am a little suspicious... If even the canonical cell requested by `ABC`/`ALPHA_BETA_GAMMA` gives rise to problematic integrated density then what needs a deeper look into is the symmetry reduction methods rather than the cell conventions.

Let's forget about the non-canonical cell for a moment and stick to the input with `ABC`/`ALPHA_BETA_GAMMA`: how does it behave with different k-point symmetry reduction options compared with the full reference, and do K290 and spglib backends show consistent (whether correct or not) results? Does number of k-points matter? How are the SCF cycles converging?

DCM-Uni-Paderborn:
@/tmp/pr5301_comment_4559707512.md

PR #5303, opened 2026-05-27
Document design for general-cell canonicalization

DCM-Uni-Paderborn:
## Summary

This is a design/infrastructure PR for the long-term general-cell canonicalization work. It adds a developer note that defines the required transform policy, invariants, guard rails, and regression-test coverage before CP2K can safely canonicalize arbitrary input cells internally.

The scope is deliberately broader than `CELL` alone. A correct implementation also has to handle `COORD`, `SHELL_COORD`, `CORE_COORD`, external coordinate files, `KPOINTS GENERAL`, band paths, electric fields, constraints, velocities, stress, and restarts.

## Motivation

CP2K currently warns when explicit cell vectors do not follow the internal convention where `A` lies on the Cartesian x axis and `B` lies in the xy plane. Simply rewriting such a cell into canonical form is unsafe unless all dependent Cartesian and reciprocal quantities are transformed at the same time. Otherwise the calculation can silently change the physical setup.

This note separates that broader infrastructure work from narrow bug fixes that avoid unsafe behavior for specific symmetry paths.

## Main points

- Define `H_in`, `H_can`, the Cartesian transform `T = H_can inv(H_in)`, and the reciprocal transform `transpose(inv(T))`.
- State which quantities keep their numeric values, for example fractional coordinates and reciprocal `B_VECTOR` k-points.
- State which quantities must be transformed, for example Cartesian coordinates, fields, velocities, constraints, Cartesian k-points, band paths, and stress.
- Require guard rails for any Cartesian input section that is not yet supported.
- Require paired regression coverage across coordinates, external files, k-points, fields, constraints, velocities, stress, and restarts before automatic canonicalization becomes the default.

## Test plan

- `make -C docs html`

This generated `docs/_build/html/developer/cell-canonicalization.html` locally. In a checkout
without the generated input reference pages, the command still exits nonzero because the existing
`docs/index.md` references `CP2K_INPUT`, `bibliography`, and `units`.

### DIFF
--- a/docs/developer/cell-canonicalization.md
+++ b/docs/developer/cell-canonicalization.md
@@ -1,0 +1,108 @@
+# General Cell Canonicalization
+
+This note describes the requirements for canonicalizing general input cell vectors in CP2K. It is a
+design checklist for future implementation work, not a user-facing promise that all cases are
+handled already.
+
+CP2K accepts cell vectors through `CELL%A`, `CELL%B`, `CELL%C`, `CELL%ABC`, `CELL%ALPHA_BETA_GAMMA`,
+and several external cell or coordinate file formats. Internally, many algorithms assume the
+conventional orientation where vector `A` lies on the Cartesian x axis and vector `B` lies in the xy
+plane. If a user provides a rotated but otherwise equivalent cell, merely rewriting `CELL` into that
+canonical orientation is not enough. Every input quantity whose meaning depends on the Cartesian
+frame must be transformed consistently.
+
+## Invariants
+
+A complete implementation must preserve the following invariants.
+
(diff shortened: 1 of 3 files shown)


oschuett:
I agree that we need a space for design work and planning.

However, I'm not sure if the manual is the right place because its audience are our users.

Maybe the [Github Wiki](https://github.com/cp2k/cp2k/wiki) or [Github Discussions](https://github.com/cp2k/cp2k/discussions) are more suitable?

Growl1234:
I would suggest using GitHub wiki.

DCM-Uni-Paderborn:
Closing this draft for now. The canonicalization work should come back as a real infrastructure PR with implementation and broad regression coverage rather than as a standalone design note.

PR #4974, opened 2026-03-19
Add explanation and error message about CELL/SYMMETRY

he-zilong:
The `&FORCE_EVAL/&SUBSYS/&CELL/SYMMETRY` keyword was not explained very well
either in the manual or when program is running. This PR would like to clarify
with a warning message and some additional notes, but does not change program
behavior otherwise. It is intended to close an old issue; if anybody would like
to discuss modifying the procedure of enforcing cell symmetry, or in general
what the program behavior should be like with a non-lower-triangular cell
matrix in the input, feel free to create new issue(s).

Closes #4170

PR #5307, opened 2026-05-27
Canonicalize general input cells consistently

DCM-Uni-Paderborn:
## Summary
- Add `CELL%CANONICALIZE` (default `T`) to transform general input cells to CP2K's internal convention while storing the original input cell and input-to-canonical transforms.
- Apply the transform to cell-dependent Cartesian input paths: `COORD`, `SHELL_COORD`, `CORE_COORD`, external coordinate files, binary restarts, velocities, general Cartesian k-points, band paths, DFT/FIST/tblite electric fields, colvar fixed points/normals, dimer vectors, and fixed-atom restart coordinates.
- Keep the #5301 inversion-only fallback available with `CANONICALIZE F`; partial `FIXED_ATOMS` component masks now abort under canonicalization because a rotated partial Cartesian mask cannot be represented safely.
- Stress and restart data follow the canonical internal cell after read, avoiding mixed input/canonical frames in downstream cell derivatives and restart files.
- Update affected non-orthogonal-cell regtest references to the canonicalized-frame results. These shifts are expected frame/cell-canonicalization effects, not grid-backend drift.

## Tests
- `OMPI_CC=/opt/homebrew/bin/gcc-15 OMPI_CXX=/opt/homebrew/bin/g++-15 ninja -C build-canonicalization-gcc cp2k.psmp -j8`
- `tools/precommit/precommit.py -m $(git diff --name-only origin/master..HEAD)`
- `OMP_NUM_THREADS=1 OMPI_CC=/opt/homebrew/bin/gcc-15 OMPI_CXX=/opt/homebrew/bin/g++-15 tests/do_regtest.py --mpiranks 1 --ompthreads 1 --maxtasks 4 --timeout 300 --maxerrors 20 --restrictdir QS/regtest-kp-1 --workbasedir /tmp/cp2k-regtest-canonicalization-final --cp2kdatadir /tmp/cp2k-current-master/data /tmp/cp2k-current-master/build-canonicalization-gcc/bin psmp` -> `Summary: correct: 72 / 72; Status: OK`
- Reran the locally available regtest directories affected by the `pdbg` reference update: `QS/regtest-grid`, `QS/regtest-gpw-3`, `QS/regtest-libxc`, `QS/regtest-nonortho`, `QS/regtest-gpw-6-2`, `QS/regtest-gpw-9`, `QS/regtest-slab`, `QS/regtest-rtp-5`, `QS/regtest-ri-mp2`, `QS/regtest-mp2-grad-2`, `Fist/regtest-1-2`, `Fist/regtest-3`, `Fist/regtest-7-1`, `Fist/regtest-7-2`, `Fist/regtest-13`, and `Fist/regtest-pol` -> all `Status: OK`.
- Full-grid sanity check for `simple_non-ortho_grid` with `NGRIDS 1`: `AUTO`, `CPU`, and `REFERENCE` all give the same SCF total energy, `-19.99169079407152`, confirming that the updated reference is not caused by a grid-backend discrepancy.
- Invariance check against an explicit canonical input: original input with automatic canonicalization gives `-19.99169079407152`; manually canonicalized cell/coordinates with `CANONICALIZE F` gives `-19.99169079408560` (difference about `1.4e-11`, from input rounding). Keeping the original non-canonical frame with `CANONICALIZE F` gives `-20.18797860872477`, matching the old energy family and confirming that the reference shift is the intended frame change.


DCM-Uni-Paderborn:
Dear Bibek,

PR #5307 has now been merged into CP2K master.

The remaining numerical differences were traced to the new internal canonicalization of general non-orthogonal input cells, not to a library or grid-backend issue. We also checked this explicitly with a full-grid reference calculation: for the `simple_non-ortho_grid` case with `NGRIDS 1`, the `AUTO`, `CPU`, and `REFERENCE` grid backends all give the same SCF total energy. In addition, the automatically canonicalized input agrees with a manually canonicalized cell/coordinate input with `CANONICALIZE F` to about `1e-11`.

So the updated references should now reflect the intended canonicalized-frame behavior.

Could you please rebase/update to current CP2K master and rerun the Elemental-Materials set once more? If anything still differs, please send me the failing inputs and outputs and I will check them directly.

Many thanks again for testing this so carefully.

All the best,
Thomas

he-zilong:
Dear Thomas,

This is a major _status quo_ change that, while potentially improving user experience by implementing the features to handle non-canonical cells, may lead to confusion due to inconsistent default behavior compared with former versions. Any modification to input atomic coordinates without adequate number of notifications and discussions will be unfavorable (as with other QC programs) and I have mixed feelings about the `&CELL/CANONICALIZE` keyword default to `.TRUE.` abruptly.

Maybe we can take some gentle, incremental steps in several future releases. For the next 2026.2 version, I think the `&CELL/CANONICALIZE` keyword should default to `.FALSE.` but:
- If the atoms are represented by relative fractional coordinates as requested by `&COORD/SCALED T`, a non-canonical cell should trigger a warning for being non-canonical and then be canonicalized by default (because this happens before constructing atomic Cartesian coordinates from fractional coordinates), with explicit messages saying this is an experimental feature and printing the rotation matrix for canonicalization;
- If the atoms are represented by absolute Cartesian coordinates, a non-canonical cell should trigger a warning for being non-canonical but does NOT get canonicalized by default (so atomic coordinates and other orientation-dependent values are not transformed), with the warning message suggesting that either the structure should be edited in a modelling program or the `&CELL/CANONICALIZE` keyword should be explicitly set to `.TRUE.`.

Then, only after gathering sufficiently many user feedbacks (they surely will have responses about which databases provide non-canonical cell with Cartesian coordinates, whether a default canonicalization is desirable, whether the impact on e.g. periodic electric field or non-hydrostatic pressure calculation is good, whether output files should use input orientation or standard orientation, ..., that developers want to hear) could we reconsider the move to make canonicalization as default.

Thank you for the extensive programming work and please think about my proposal.

Regards,
Zilong

DCM-Uni-Paderborn:
Dear Zilong,

thank you, I think this is a very fair concern.

The intention of #5307 was not to change the physical input, but to put general non-canonical cells into CP2K's internal canonical frame while transforming the dependent Cartesian quantities consistently. The tests and full-grid references indicate that this is internally consistent.

However, I agree that consistency does not automatically make this the least surprising default. Changing the default behavior for Cartesian-coordinate inputs is a real user-interface and backwards-compatibility issue, especially for direction-dependent quantities such as external fields, non-hydrostatic stress, constraints, velocities, restarts, and output orientation.

I would therefore separate two issues:

1. The safety fix in #5309 is still needed independently of the canonicalization default, because the full point-group k-point density-matrix symmetrization is not charge-conserving for skew cells even in canonical `ABC`/`ALPHA_BETA_GAMMA` form.

2. The default policy for `CELL/CANONICALIZE` deserves a separate discussion. I agree that making canonicalization opt-in, or at least using a more conservative default for Cartesian-coordinate inputs, may be better for the next release.

Your `SCALED`-coordinate proposal makes sense conceptually, since fractional coordinates can be converted after canonicalizing the cell. I would still be careful implementing this as an implicit context-dependent default without broad coverage, because CP2K also has `SHELL_COORD`, `CORE_COORD`, external coordinate files, restarts, velocities, constraints, fields, k-points, and band paths. A clean solution may be an explicit `AUTO`/`TRUE`/`FALSE` policy plus clear warnings and printing of the transformation matrix.

So I am happy to prepare a follow-up design/implementation PR for the default policy, but I agree it should be discussed explicitly rather than silently changing long-standing behavior.

Best,
Thomas

DCM-Uni-Paderborn:
I opened a broader RFC discussion for this topic here:

https://github.com/cp2k/cp2k/discussions/5310

I tried to summarize the related history (#4170, #4950, #4971/#4962, #4974, #5303, #5307, #5309) and frame the main policy question separately from the current numerical/safety fixes: should `CELL/CANONICALIZE` remain a boolean default, become opt-in, or become an explicit `AUTO`/`TRUE`/`FALSE` policy with clear warnings and transformation-matrix output?

PR #5309, opened 2026-05-28
Avoid full k-point symmetry for skew cells

DCM-Uni-Paderborn:
## Summary

This PR keeps reduced k-point calculations on a charge-conserving path for non-orthogonal cells. If `MONKHORST-PACK` or `MACDONALD` k-points request full atomic k-point symmetry for a skew cell, CP2K now falls back to `KPOINTS%INVERSION_SYMMETRY_ONLY` and emits a warning. The existing non-lower-triangular-cell fallback is preserved.

The regtest references are updated where the safe fallback increases the number of special k-points:

- `h_hex_c3_gamma_k290`: 8 -> 20
- `h_hex_c3_gamma_spglib`: 8 -> 20
- `h_hex_c3_gamma_macdonald_spglib`: 12 -> 32
- `diamond_gpw_extRI_kpsym`: 2 -> 4
- `Zn_fcc_gfn2_kp_k290`: 4 -> 14

## Motivation

This completes the follow-up investigation to #5301/#5307. On current master after #5307, rewriting the Fe benchmark cell into canonical `ABC`/`ALPHA_BETA_GAMMA` form does not fix the full point-group k-point reduction: both K290 and spglib reduce the 16x16x16 grid to 140 k-points, but the integrated electronic density is not neutral and the total energy is shifted relative to the full-grid reference.

Representative current-master results before this PR:

| case | k-points | integrated density error | total energy [Ha] |
| --- | ---: | ---: | ---: |
| canonical full grid | 4096 | `1e-10` | `-123.54254579485098` |
| canonical inversion only | 2048 | `1e-10` | `-123.54254579449282` |
| canonical K290 full symmetry | 140 | `-0.0272115182` | `-123.53278963531541` |
| canonical spglib backend | 140 | `-0.0272115182` | `-123.53278963521340` |

The same behaviour appears for the non-canonical Fe cell. K290 and spglib are therefore consistent with each other, but both exercise the same unsafe density-matrix symmetrization path for skew cells. Mesh sweeps over 8/12/16 keep the same qualitative problem, so this is not a simple k-point convergence issue. I also checked the obvious internal suspects in the transform path (reciprocal folding phase, little-group averaging via `nwred`, and the AO block rotation convention); none restored charge conservation.

Until the full point-group density-matrix transform is made robust for non-orthogonal cells, using inversion-only reduction is the safe user-visible behaviour.

## Validation

Built and tested on the NVIDIA Spark host with current `master` at `48c63364e3` plus this branch.

Final Fe 16x16x16 check after this PR:

The 2048-point K290/spglib rows are intentional: these are user requests that now trigger the new inversion-only fallback, not full point-group reductions to 140 k-points.

| case | k-points | integrated density error | total energy [Ha] |
| --- | ---: | ---: | ---: |
| canonical full grid | 4096 | `1e-10` | `-123.54254579436157` |
| canonical inversion only | 2048 | `1e-10` | `-123.54254579479700` |
| canonical K290 request -> inversion fallback | 2048 | `1e-10` | `-123.54254579473296` |
| canonical spglib request -> inversion fallback | 2048 | `1e-10` | `-123.54254579481585` |
| non-canonical full grid | 4096 | `1e-10` | `-123.54254578592314` |
| non-canonical inversion only | 2048 | `1e-10` | `-123.54254578564385` |
| non-canonical K290 request -> inversion fallback -> inversion fallback | 2048 | `1e-10` | `-123.54254578579067` |
| non-canonical spglib request -> inversion fallback -> inversion fallback | 2048 | `1e-10` | `-123.54254578544521` |

Regtests:

- `QS/regtest-kp-1-spg` + `QS/regtest-kp-hfx-ri`: `23 / 23` correct after the final patch.
- Broader affected-directory sweep also passed all non-Wannier90 checks: `DFTB/regtest-scc`, `DFTB/regtest-scc-spglib`, `DFTB/regtest-nonscc`, `QS/regtest-kp-2`, `QS/regtest-kp-3`, `QS/regtest-kp-spglib`, `QS/regtest-harris-kp`. The local Spark build lacks the Wannier90 output path and tblite support, so Wannier90 text matchers in `QS/regtest-kp-1`/`QS/regtest-kp-1-spglib` and the tblite-only `xTB/regtest-tblite-gfn2` case could not all be validated locally. The CI `Regtest pdbg` run supplied the missing tblite k-point reference for `Zn_fcc_gfn2_kp_k290` (`4 -> 14`); its energy matcher was already OK.

Manual final check:

- `QS/regtest-kp-1/h_hex_c3_gamma_k290.inp`: `N_special_kpoints = 20`, `E_total = -9.413339735078168 Ha`, with the new non-orthogonal-cell fallback warning.

### DIFF
--- a/src/kpoint_types.F
+++ b/src/kpoint_types.F
@@ -737,7 +737,8 @@ CONTAINS
          DIMENSION(:), POINTER                           :: tmpstringlist
       INTEGER                                            :: i, n_rep, nval, wfntype
       LOGICAL                                            :: available, backend_explicit, &
-                                                            reduction_explicit
+                                                            non_lower_triangular_cell, &
+                                                            non_orthogonal_cell, reduction_explicit
       REAL(KIND=dp)                                      :: ff
       REAL(KIND=dp), DIMENSION(3, 3)                     :: cart_a_vec
       REAL(KIND=dp), DIMENSION(:), POINTER               :: reallist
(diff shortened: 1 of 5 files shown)


he-zilong (commented):
Dear Thomas,

Is my understanding correct that the incorrect behavior (non-neutral integrated density) observed in #5301 has been attributed to a skewed/non-orthogonal cell? It is somewhat worrying that this conclusion sounds a bit too general as if hexagonal and monoclinic cells with only one non-orthogonal cell angle are also affected. Could this be confirmed by reproducing the behavior in (BCC and non-BCC) systems other than Fe as reported in #5123?

The fallback introduced in this PR looks fine to me otherwise because it is essentially just disabling a new, yet-to-be-extensively-validated feature introduced after the last release version.

Regards,
Zilong

P.S. I have some question that may or may not be relevant to the situation here. Do symmetry reduction algorithms have special treatments for real-space rhombohedral cells ($a = b = c, \alpha = \beta = \gamma$) with particular angles? While it is well-known that a primitive rhombohedral cell with $\alpha = \beta = \gamma = 60 \deg$ corresponds to a face-centered cubic lattice, one with $\alpha = \beta = \gamma = 90 \deg$ corresponds to a simple-cubic lattice, and one with $\alpha = \beta = \gamma = 104.5 \deg$ (or $\arccos \frac{-1}{3}$ to be precise) corresponds to a body-centered cubic lattice, I am not sure if these particular cases influence the symmetry reduction in the reciprocal space. Consider a series of rhombohedral cells with the angle varying continuously around 60 degrees, say from 59.9 to 60.0 to 60.1; is it expected theoretically that some sort of "jump discontinuity" for the number and distribution of k-points happens at 60.0 degrees?

DCM-Uni-Paderborn:
Dear Zilong,

thank you, this is a good point. I agree that the wording should not imply that we have proven a charge-conservation failure for every possible non-orthogonal Bravais cell.

What I have explicitly reproduced is the failure for the Fe benchmark from #5123/#5301, both in the original non-canonical representation and after rewriting the cell into canonical `ABC`/`ALPHA_BETA_GAMMA` form. In that case K290 and spglib consistently reduce the 16x16x16 grid to 140 k-points, but the density integration is no longer neutral, whereas the full grid and inversion-only reduction remain neutral.

The PR therefore uses “non-orthogonal cell” as a conservative risk criterion for the shared full point-group density-matrix symmetrization path, not as a statement that every hexagonal, monoclinic, or rhombohedral case has already been shown to fail numerically. The affected reference updates show that this path is reached by several existing non-orthogonal-cell tests, but they are not meant to prove that all such cells are wrong. I can soften the PR text to make that distinction clearer.

About the rhombohedral question: yes, I would expect discontinuous changes in the number and distribution of irreducible k-points at special metric points such as 60 deg, 90 deg, or arccos(-1/3), because the lattice point group itself changes discretely there. In practice this is also tolerance-dependent. Such a jump in the symmetry reduction is not by itself suspicious. What would be suspicious is if the reduced-grid calculation no longer reproduces the full-grid charge/energy within the expected tolerance. That is the issue observed here.

So I think the safe current interpretation is: the fallback is deliberately conservative, and we should only relax it later if we have a broader validation set showing which non-orthogonal subclasses are safe for full point-group density-matrix symmetrization.

Best,
Thomas

P.S. Or, in other words, both k290 and spglib are finding the correct reduction, it's the reconstruction of the density that is the issue - I am working on that ...

PR #5312, opened 2026-05-28
Dimer: add control of initialization method and implement read from molden

he-zilong:
The dimer method for searching transition state requires an initial dimer vector. Previously it is read from the `&TRANSITION_STATE/&DIMER/&DIMER_VECTOR` section, or if not available, generated randomly. In practice, the latter makes a poor guess that may distort the structure.

This PR adds a keyword `INITIALIZATION_METHOD` to control the initialization, which for backwards compatibility defaults to `RANDOM` and is ignored if the `&DIMER_VECTOR` section is present (e.g. in restart files). A new option `MOLDEN` is created which allows reading a molden file from a vibrational analysis of the input structure and forming an initial dimer vector by a linear combination of one or more vibrational normal modes with indices `VIB_INDEX` and weights `VIB_WEIGHT`. For instance:

```
&DIMER
  INITIALIZATION_METHOD MOLDEN
  VIB_INDEX 1 2
  VIB_WEIGHT 0.4 0.6
  VIB_MOLDEN_NAME TS_guess_freq-VIBRATIONS-1.mol
```

The default value for `VIB_INDEX` and `VIB_WEIGHT` takes the very first vibrational normal mode directly as the initial dimer vector, which may be a much better guess than the randomly generated one. The reason to parse the molden file is because it is plaintext-readable and likely have been visualized by user in order to pick the vibrational normal mode with appropriate atomic motion.

### DIFF
--- a/src/input_constants.F
+++ b/src/input_constants.F
@@ -551,6 +551,9 @@ MODULE input_constants
    INTEGER, PARAMETER, PUBLIC               :: none_ts_method_id = 0, &
                                                default_dimer_method_id = 1
 
+   INTEGER, PARAMETER, PUBLIC               :: dimer_init_random = 0, &
+                                               dimer_init_molden = 1
+
    INTEGER, PARAMETER, PUBLIC               :: do_first_rotation_step = 0, &
                                                do_second_rotation_step = 1, &
                                                do_third_rotation_step = 2
--- a/src/motion/dimer_types.F
+++ b/src/motion/dimer_types.F
@@ -16,16 +16,27 @@ MODULE dimer_types
    USE cell_types,                      ONLY: &
         cell_transform_input_cartesian, cell_type, use_perd_x, use_perd_xy, use_perd_xyz, &
         use_perd_xz, use_perd_y, use_perd_yz, use_perd_z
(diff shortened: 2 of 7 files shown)


fstein93:
Could you prepare a test for the new feature?

he-zilong:
> Could you prepare a test for the new feature?

I found that even the existing functionality of printing molden via `&VIBRATIONAL_ANALYSIS/&PRINT/&MOLDEN_VIB` has not been triggered by any regtest input yet. So I may need a while to come up with a suite of fast test inputs.

he-zilong:
I have added two regtest files: `argon_vib_molden_write.inp` performs a vibrational analysis to produce `argon_vib_molden_write-VIBRATIONS-1.mol`, which then is read by `argon_dimer_molden_read.inp` to initialize the dimer vector.

The `INITIALIZATION_METHOD` keyword can be expanded in the future with new options. For instance, it may be possible to create a model Hessian and obtain vibrational force constants and normal modes for the input structure, as mentioned [in a slightly unrelated issue](https://github.com/cp2k/cp2k/issues/1233#issuecomment-4509743436).

PR #5313, opened 2026-05-28
Fix high-l real spherical harmonic rotations

DCM-Uni-Paderborn:
This fixes the real spherical harmonic rotation recursion used for AO basis rotations.

The previous implementation had two problems in the Ivanic/Ruedenberg recursion for `l >= 3`:

- the edge cases in `p_func` addressed the `l-1` rotation matrix with `m = +/-l`, which is outside the physical band and therefore reads the zero-padded part of the work array instead of the `+/-(l-1)` edge columns;
- the `dd` factor in `r_recursion` was initialized as one for all positive/negative `m`, rather than acting as the intended Kronecker delta for `m = +/-1`.

For Fe/DZVP-MOLOPT-SR-GTH this produced a broken f-shell AO rotation matrix: even the identity rotation had two zero rows/columns in the 26x26 basis rotation block. In k-point symmetry this corrupted full point-group density-matrix reconstruction and made the 140-k-point Fe reduction lose charge relative to the full-grid/inversion-only reference.

The patch corrects the recursion and adds a standalone unit test that checks identity rotations and orthogonality for real spherical harmonic rotation matrices up to `l = 5`.

Validation performed locally:

- `python3 tools/precommit/precommit.py src/aobasis/orbital_transformation_matrices_unittest.F src/aobasis/orbital_transformation_matrices.F src/CMakeLists.txt`
- `cmake --build /tmp/cp2k-kpoint-phase/build-phase-gcc --target orbital_transformation_matrices_unittest -j 8`
- `/tmp/cp2k-kpoint-phase/build-phase-gcc/bin/orbital_transformation_matrices_unittest.psmp`
- `cmake --build /tmp/cp2k-kpoint-phase/build-phase-gcc --target cp2k.psmp -j 8`
- Fe 2x2x2 MP k-points:
  - full grid: 4 k-points, charge `-15.9999999999`, energy `-123.533772912037591` Ha
  - inversion-only: 4 k-points, charge `-15.9999999999`, energy `-123.533772912037591` Ha
  - full point group: 2 k-points, charge `-15.9999999999`, energy `-123.533772909851109` Ha
- Fe 4x4x4 MP k-points:
  - full grid: 32 k-points, charge `-15.9999999999`, energy `-123.539807690564800` Ha
  - inversion-only: 32 k-points, charge `-15.9999999999`, energy `-123.539807690564800` Ha
  - full point group: 6 k-points, charge `-15.9999999999`, energy `-123.539807289227610` Ha
- Fe 16x16x16 MP k-points with full point-group reduction:
  - 140 k-points, charge `-15.9999999999`, energy `-123.542545803670521` Ha
  - before this fix, the same 140-k-point path gave charge `-16.0272115150` and energy `-123.532789630038252` Ha.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1517,6 +1517,7 @@ set(CP2K_PROGS_C
     grid/grid_miniapp.c;grid/grid_unittest.c;dbm/dbm_miniapp.c;start/libcp2k_unittest.c
 )
 set(CP2K_PROGS_F
+    aobasis/orbital_transformation_matrices_unittest.F
     dbt/dbt_unittest.F
     dbt/tas/dbt_tas_unittest.F
     start/cp2k.F
@@ -1868,6 +1869,7 @@ add_library(cp2k::cp2k ALIAS cp2k)
 list(
   APPEND
   __CP2K_APPS
+  "orbital_transformation_matrices_unittest"
   "gemm_square_unittest"
   "memory_utilities_unittest"
   "parallel_rng_types_unittest"
@@ -1884,6 +1886,8 @@ list(
 add_executable(cp2k-bin start/cp2k.F)
 # for linking
 
+add_executable(orbital_transformation_matrices_unittest
+               aobasis/orbital_transformation_matrices_unittest.F)
 add_executable(gemm_square_unittest common/gemm_square_unittest.F)
 add_executable(memory_utilities_unittest common/memory_utilities_unittest.F)
 add_executable(parallel_rng_types_unittest common/parallel_rng_types_unittest.F)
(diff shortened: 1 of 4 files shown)


src/CMakeLists.txt
```diff
@@ -1884,6 +1886,8 @@ list(
 add_executable(cp2k-bin start/cp2k.F)
 # for linking
 
+add_executable(orbital_transformation_matrices_unittest
```
oschuett, line 1889: New unit tests also have to be added to [tests/UNIT_TESTS](https://github.com/cp2k/cp2k/blob/master/tests/UNIT_TESTS).

DCM-Uni-Paderborn: Thanks, fixed in the latest push: `orbital_transformation_matrices_unittest` is now listed in `tests/UNIT_TESTS`.

he-zilong (commented):
Now this is an important fix! I do not know how widely used this AO rotation routine is utilized in general, but this can potentially help many more k-point calculations with Fe atoms described with `DZVP-MOLOPT-SR-GTH-q16` and `GTH-PBE-q16`, and even with other transition metal elements. Let's wait and see...

PR #5317, opened 2026-05-29
CMake: Fix libint2 Fortran module finding

Growl1234:
Libint2 in RPM build has enabled Fortran, but is still disabled in rpmbuild process of CP2K. This draft tries to test the case that libint2 is enabled to see how it goes,


Growl1234:
The issues observed in Test Farm (full) actually look similar to many dashboard failures. Fortunately, the tests relevant to this PR, namely those related to the `libint` dependency, have passed.

Therefore, this PR has already achieved its purpose, and I do not plan to continue testing the RPM build or Test Farm after relaxing the tolerance...

Growl1234:
And in the longer run, it may also be worth reconsidering whether CP2K still needs to maintain its own `FindLibint2.cmake`, because CMake is now the only supported way to build Libint libraries, which means CP2K could rely on the upstream-installed CMake package configuration files, such as `lib/cmake/libint2/libint2-config*.cmake`, instead of duplicating Libint2 detection logic in a custom find module...

oschuett:
> CP2K could rely on the upstream-installed CMake package configuration files

If that has become an option, then we should definitely take it.

PR #5318, opened 2026-05-29
Add conservative AUTO default for cell canonicalization

DCM-Uni-Paderborn:
## Summary

- change `CELL/CANONICALIZE` from a default-true logical keyword to an `AUTO`/`TRUE`/`FALSE` policy keyword
- make the new default `AUTO` conservative: for general input-cell vectors it keeps the user-provided orientation and warns instead of silently rotating Cartesian or direction-dependent input
- keep compatibility aliases for explicit opt-in/out (`T`, `F`, `.TRUE.`, `.FALSE.`), with a bare `CANONICALIZE` keyword still meaning `TRUE`
- update the k-point symmetry regression coverage so explicit `TRUE` preserves the canonicalization/reduction path while default `AUTO` exercises the robust noncanonical path
- make legacy regtests that depend on the previous implicit canonicalization default opt in explicitly with `CANONICALIZE T`

## Rationale

This implements the conservative compromise discussed in https://github.com/cp2k/cp2k/discussions/5310. Default-true canonicalization is user-visible for workflows that treat Cartesian coordinates, directions, band paths, fields, constraints, velocities, stress, or restarts as being in a laboratory/input frame. Until CP2K has complete internal canonicalization coverage for all such cell-dependent input, `AUTO` should avoid unsafe silent transformations.

The explicit `TRUE` mode keeps the existing canonicalization behavior for users and tests that intentionally want the canonical internal frame and the associated k-point symmetry reduction.

## Regtest report follow-up

The first `pdbg` CI report showed 42 wrong-result regressions, while AiiDA and Misc were OK. The failures were old tests with noncanonical/general cells whose references were produced under the old implicit `CANONICALIZE T` default. This update makes that dependency explicit in those regression inputs instead of changing their references to the new conservative default behavior. Restart chains needed the producing input as well as the consuming input to opt in, otherwise later values could differ at the 1e-9 level.

## Tests

- `./tools/precommit/precommit.py -m ...`
- `cmake --build /tmp/cp2k-canonicalize-auto/build-gcc --target cp2k.psmp -j 8`
- `QS/regtest-kp-1`: 74/74 OK
- affected report directories run locally with the available `psmp` feature set; all runnable affected directories passed after the explicit opt-ins, including `Fist/regtest-7-1` and `Fist/regtest-7-2` at 114/114 OK
- manual parser compatibility checks for `.TRUE.`, `FALSE`, and `.FALSE.` aliases


Growl1234 (approved):

he-zilong (commented):
I noticed that the following old regtest inputs have been edited to specify `CANONICALIZE T` even though the other keyword present under the `&CELL` section is `ABC` in all cases:
```
tests/Fist/regtest-7-1/uo2_shell_npe_2.inp
tests/Fist/regtest-7-1/uo2_shell_npe_2res.inp
tests/Fist/regtest-7-2/UO2-2x2x2-binary_restart-cs-1.inp
tests/Fist/regtest-7-2/UO2-2x2x2-binary_restart-cs-2.inp
tests/Fist/regtest-7-2/UO2-2x2x2-binary_restart-cs-3.inp
tests/Fist/regtest-7-2/UO2-2x2x2-binary_restart-cs-4.inp
tests/Fist/regtest-7-2/UO2-2x2x2-binary_restart-cs-5.inp
tests/Fist/regtest-7-2/UO2-2x2x2-binary_restart-cs-6.inp
tests/Fist/regtest-7-2/UO2-2x2x2-binary_restart-cs-7.inp
tests/Fist/regtest-7-2/UO2-2x2x2-binary_restart-cs-8.inp
tests/Fist/regtest-7-2/UO2-2x2x2-nocs2cs-npt-1.inp
tests/Fist/regtest-7-2/UO2-2x2x2-nocs2cs-npt-2.inp
tests/Fist/regtest-7-2/UO2-2x2x2-nocs2cs-npt-3.inp
```
This is not expected because they should not be affected by the recent revisions related to cell canonicalization in the first place. The program design has been that a single `ABC` keyword like `ABC 10.936 10.936 10.936` (meaning that the implicit `ALPHA_BETA_GAMMA` keyword uses a default value of three 90 degrees) should create a cell matrix that is precisely diagonal which is of course lower triangular and thus already canonical.

DCM-Uni-Paderborn:
Thanks, good catch. I agree that a plain `ABC ...` cell with the implicit `90 90 90` angles should be diagonal and therefore already canonical.

I checked this locally by removing these entries. The direct ABC cases are indeed unaffected, but the NPE/NPT restart chains are a bit subtler: some of these inputs generate restart files after the cell has evolved away from the initial diagonal ABC cell, and the downstream tests then read that deformed cell through `EXT_RESTART`. In that situation the explicit legacy canonicalization is needed on the restart-producing runs to preserve the old restart trajectory at the very tight regtest tolerances.

I pushed `ad7cfdb772` to clean this up: the redundant `CANONICALIZE T` lines were removed from the pure restart consumers, while keeping them on the restart-producing NPE/NPT inputs where they affect the generated restart state. With that reduced set, local checks are green:

- `Fist/regtest-7-1`: 54/54 OK
- `Fist/regtest-7-2`: 60/60 OK

CI is rerunning now.

PR #5329, opened 2026-05-31
Add a few DFT-D3 functional parameters

### LINKED ISSUE #5264: Feature request: built-in and auto-detected DFT-D3 parameters for new functionals
The functional-specific DFT-D3 correction parameters `s6`, `s8`, `a1`, `a2` etc as recorded here
https://github.com/cp2k/cp2k/blob/919a8b9f6199214e0ecc8ec7c67ca39d4e9e827b/src/qs_dispersion_utils.F#L416
https://github.com/cp2k/cp2k/blob/919a8b9f6199214e0ecc8ec7c67ca39d4e9e827b/src/qs_dispersion_utils.F#L790
are last updated at least 2 years ago, missing some data for e.g. [r2SCAN](https://github.com/dftd3/simple-dftd3/blob/55b886fff57ad6df9208b26c96b797e9b09d7315/assets/parameters.toml#L280-L281) and its derivatives.

I propose that the new parameters from the regularly maintained [`simple-dftd3/assets/parameters.toml`](https://github.com/dftd3/simple-dftd3/blob/main/assets/parameters.toml) be incorporated, so that there is not need to specify custom parameters via `D3BJ_SCALING` or `D3_SCALING` keywords under the `&FORCE_EVAL/&DFT/&XC/&VDW_POTENTIAL/&PAIR_POTENTIAL` section.

In addition, an automatic detection of functional like in #5178 for DFT-D3 would be nice. (Edit: #5178 is superseded by #5296 which has implemented such a detection.)

he-zilong:
This PR resolves #5264 partly, with some functional-specific DFT-D3 parameters from the [`simple-dftd3/assets/parameters.toml`](https://github.com/dftd3/simple-dftd3/blob/main/assets/parameters.toml) added to the `qs_dispersion_utils.F` file, in particular the ones for PW1PW[1] and the r2SCAN family[2][3]. Additionally, the links to parameters have been updated.

[1] _Phys. Chem. Chem. Phys._, 19, 32184-32215 (2017). DOI: [10.1039/c7cp04913g](https://doi.org/10.1039/c7cp04913g)
[2] _J. Chem. Phys._ 154, 061101 (2021). DOI: [10.1063/5.0041008](https://doi.org/10.1063/5.0041008)
[3] _J. Chem. Phys._ 156, 134105 (2022). DOI: [10.1063/5.0086040](https://doi.org/10.1063/5.0086040)

### DIFF
--- a/src/qs_dispersion_utils.F
+++ b/src/qs_dispersion_utils.F
@@ -561,8 +561,10 @@ CONTAINS
       functional = ref_functional
       CALL uppercase(functional)
       ! values for different functionals from:
-      ! https://www.chemie.uni-bonn.de/grimme/de/software/dft-d3
+      ! https://www.chemie.uni-bonn.de/grimme/de/software/dft-d3/zero_damping
       ! L. Goerigk et al. PCCP 2017, 32147-32744, SI
+      ! alternatively see the parameter file for the s-dftd3 program:
+      ! https://github.com/dftd3/simple-dftd3/blob/main/assets/parameters.toml
       SELECT CASE (TRIM(functional))
       CASE DEFAULT
          ! unknown functional
@@ -877,7 +879,7 @@ CONTAINS
          s6 = 1.000_dp
          sr6 = 1.021_dp
          s8 = 0.862_dp
-      CASE ("REVTPSSh")
+      CASE ("REVTPSSH")
          s6 = 1.000_dp
          sr6 = 1.3224_dp
          s8 = 1.2504_dp
@@ -935,8 +937,10 @@ CONTAINS
       CALL uppercase(functional)
 
       ! values for different functionals from:
-      ! http://www.thch.uni-bonn.de/tc/downloads/DFT-D3/functionalsbj.html
+      ! https://www.chemie.uni-bonn.de/grimme/de/software/dft-d3/bj_damping
       ! L. Goerigk et al. PCCP 2017, 32147-32744, SI
+      ! alternatively see the parameter file for the s-dftd3 program:
+      ! https://github.com/dftd3/simple-dftd3/blob/main/assets/parameters.toml
       SELECT CASE (TRIM(functional))
       CASE DEFAULT
          ! unknown functional
@@ -1006,7 +1010,7 @@ CONTAINS
          a1 = 0.3919_dp
          s8 = 1.8541_dp
          a2 = 5.0897_dp
-      CASE ("LCWhPBE")
+      CASE ("LCWHPBE")
          s6 = 1.0000_dp
          a1 = 0.2746_dp
          s8 = 1.1908_dp
@@ -1051,7 +1055,7 @@ CONTAINS
          a1 = 0.3995_dp
          s8 = 1.4623_dp
          a2 = 5.1405_dp
-      CASE ("PBEsol")
+      CASE ("PBESOL")
          s6 = 1.0000_dp
          a1 = 0.4466_dp
          s8 = 2.9491_dp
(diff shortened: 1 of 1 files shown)


Growl1234:
Just a little note: using “resolve” or “close” before linking an issue could mark that issue to be closed automatically when the PR is merged.

he-zilong:
> Just a little note: using “resolve” or “close” before linking an issue could mark that issue to be closed automatically when the PR is merged.

I know, and the link is intentional here. Judging from the [github documents](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue) it is not yet possible to link a PR to an issue in a way that does not close the issue once the PR is merged. But linking multiple PRs to the same issue is possible with the syntax.

In this case, although this PR does not fully synchronize the parameters available, I myself am no longer interested in keeping the original issue open seeing that there is not much outside interest. It is not clear to me which functionals besides r2SCAN are nowadays popular enough for first-principles calculations that many users expect built-in DFT-D3 parameters. (By the way, a separate feature where the DFT-D2 implementation can switch between the common formulation and the wB97XD-specific formulation may be of interest.)

fstein93:
In general, I would add all available parametrizations. Unfortunately, we never got around to prepare a list of available functionals for each correction or even use an internal integer instead of a string to represent the requested parametrization.

he-zilong:
> In general, I would add all available parametrizations. Unfortunately, we never got around to prepare a list of available functionals for each correction or even use an internal integer instead of a string to represent the requested parametrization.

Maybe it is possible just to drop the built-in list of parameters in `src/qs_dispersion_utils.F`, copy the `simple-dftd3/assets/parameters.toml` to our `data` directory, and then rewrite our fortran code to parse it for parameters. As a bonus, I envision that the same parameter file may also be parsed to create a list of available functionals in the documentation. (Provided that the `simple-dftd3` authors agree and the licenses are compatible, that is.)

For me this PR is good enough at the moment; having the increasingly popular r2SCAN supported is already a convenience for many users.

oschuett:
> Maybe it is possible just to drop the built-in list of parameters in src/qs_dispersion_utils.F, copy the simple-dftd3/assets/parameters.toml to our data directory, and then rewrite our fortran code to parse it for parameters. 

Alternatively, we could write a Python script that generates the Fortran code, like e.g. the [minimax_tools](https://github.com/cp2k/cp2k/tree/master/tools/minimax_tools).

he-zilong:
I understand that the approach of getting parameters from `simple-dftd3` package in #5346 is better, but I think having the built-in set of r2SCAN parameters independent of external libraries is more convenient for cases like a minimal build. Thus, I see no reason to abandon this PR; please get it merged...

Growl1234:
I think a Python script might be the right way forward, and I have some concern that directly introducing an `s-dftd3` dependency might complicate what is otherwise a fairly simple workflow...

@jpoto, might you want to share some thoughts?

jpoto:
The python script would require regular runs and maintenance, but does not introduce a dependency. 
So it would always be available right away for the script. If we use the script then the parameters would be depend on the CP2K version, otherwise on the version of tblite/simple-dftd3. If someone installs tblite themselves, they would need to keep track of it.

oschuett:
> The python script would require regular runs and maintenance, but does not introduce a dependency.

A new direct dependency would also increase our maintenance load. Hence, I'd still prefer the script solution as it would provide the superior user experience.

jpoto:
> A new direct dependency would also increase our maintenance load. Hence, I'd still prefer the script solution as it would provide the superior user experience.

 If you look at #5346 it uses simple-dftd3 from tblite, which we need anyways. I would keep it this way to avoid a new dependency.

he-zilong:
This PR also fixes some names of functionals that contain lowercase letters and cannot be matched against the result of `CALL uppercase(functional)`, so I still want to see it merged as a short-term fix. For the long-term implementation we can use #5264 or #5346 to discuss.

PR #5335, opened 2026-06-01
Add documentation of DOS (Density of States)

Growl1234:


he-zilong:
Given the breaking changes in #5287 and #5326, it is better to clarify that the broadened and plottable `.dos` and `.pdos` files are only available after May 2026 (or release version 2026.2), and that the former `.dos` and `.pdos` files have been renamed to `.dos_raw` and `.pdos_raw` now.

Meanwhile, I have other works on the documentation...

Growl1234:
> Given the breaking changes in #5287 and #5326, it is better to clarify that the broadened and plottable `.dos` and `.pdos` files are only available after May 2026 (or release version 2026.2)...

I've added an "important" section to clarify this.

PR #5340, opened 2026-06-02
Gauxc object caching

franzpoeschel:
CP2K currently re-initializes the GauXC data structures (including grid and integrator) for every SCF step. This introduces caching for these objects inside the QS environment, including verification that the parameters stay the same.

TODO:

- [x] more correctness verification
- [ ] Merge https://github.com/cp2k/cp2k/pull/5533 first, this is now based on that


franzpoeschel:
Temporarily closing this as I will be off work until July

Growl1234:
> Temporarily closing this as I will be off work until July

Actually you can [convert it to draft](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#draft-pull-requests) and keep it open...

图片

franzpoeschel:
The only remaining failures are 8 tests, all with the same message:

```
MODEL SKALA requires the GAUXC_SKALA_MODEL environment variable
```

I do not understand why this would come up with this PR. These 8 tests all require the GAUXC_SKALA_MODEL environment variable already on master, this PR does not change the visible behavior of the GauXC implementation, and the CI configuration is not touched at all.

If GAUXC_SKALA_MODEL environment variable is really not set in the CI, then these tests should already fail on the master branch before this PR (and testing them locally, they do).

So, the failures are correct, but they should occur already before this PR.

mkrack:
> The only remaining failures are 8 tests, all with the same message:
> 
> ```
> MODEL SKALA requires the GAUXC_SKALA_MODEL environment variable
> ```
> 
> I do not understand why this would come up with this PR. These 8 tests all require the GAUXC_SKALA_MODEL environment variable already on master, this PR does not change the visible behavior of the GauXC implementation, and the CI configuration is not touched at all.
> 
> If GAUXC_SKALA_MODEL environment variable is really not set in the CI, then these tests should already fail on the master branch before this PR (and testing them locally, they do).
> 
> So, the failures are correct, but they should occur already before this PR.

Maybe, you have to re-run the `Regtest pdbg` tester, because the `spack pdbg` tester run does not show any failures.

franzpoeschel:
> Maybe, you have to re-run the `Regtest pdbg` tester, because the `spack pdbg` tester run does not show any failures.

I can certainly try once more, but I am not very hopeful, I have seen these failures repeatedly already

mkrack:
> I can certainly try once more, but I am not very hopeful, I have seen these failures repeatedly already

Hmm, I have also restarted the `spack pdbg` tester to see if any random failures show up.

mkrack:
That's weird. The failures are indeed reproducible with the `Regtest pdbg` tester while all GauXC tests pass again with the `Spack pdbg` tester.

franzpoeschel:
Which files define the Regtest pdbg CI workflow? I did not write the CI integration for GauXC, so I don't really know where to look for the setup.

Grepping for the missing environment variable inside the source tree yields these files:
```
docs/methods/dft/gauxc.md
make_cp2k.sh
tools/toolchain/scripts/stage6/install_gauxc.sh
src/input_cp2k_xc.F
src/skala_gpw_functional.F
src/xc/xc_gauxc_functional.F
```

Does the model need to be installed and made available at some other place as well?

DCM-Uni-Paderborn:
Dear Franz,

the `Regtest pdbg` target is defined in `tools/docker/cp2k-ci.conf` (section `[pdbg]`), which points to `tools/docker/Dockerfile.test_pdbg`. That Dockerfile installs the CP2K toolchain, builds CP2K, and invokes `tools/docker/scripts/test_regtest.sh`. The latter sources `/opt/cp2k-toolchain/install/setup` before running `tests/do_regtest.py`.

The SKALA model is already handled by `tools/toolchain/scripts/stage6/install_gauxc.sh`: it downloads `skala-1.1.fun`, installs it below `<GauXC prefix>/share/gauxc/onedft_models/`, and writes `GAUXC_SKALA_MODEL` to the toolchain setup file. Thus, no additional CI installation location is required.

The environment variable was actually present. The caching PR moved the resolved model path into `cp_gauxc_cache_params%model_name`, but that field used `default_string_length` (80 characters). The toolchain path is 94 characters long. Fortran therefore returned `STATUS=-1` from `GET_ENVIRONMENT_VARIABLE` because the value was truncated, and the existing check reported this as if the variable were missing. The Spack path happened not to trigger the same truncation.

I changed the cached model field to `default_path_length`, fixed the non-GauXC fallback compilation of the new cache module, and rebased the branch onto current `master`. A direct reproducer gives status `-1` with the old 80-character buffer and status `0` with the path-sized buffer. Precommit and a local debug compilation without GauXC pass; the restarted `Regtest pdbg` should now verify the toolchain/GauXC path.

I also checked the still-unmerged PR #5533 separately. It is not folded into this branch. The combined code needs to cache the effective GauXC model name (`NONE` or the resolved SKALA path) and preserve #5533's replicated runtime for conventional GauXC; a synthetic integration with those semantics passes precommit and local compilation.

franzpoeschel:
> The caching PR moved the resolved model path into `cp_gauxc_cache_params%model_name`, but that field used `default_string_length` (80 characters). The toolchain path is 94 characters long.

Ah, that explains why I could not reproduce this locally, the 80 chars were enough for the path I was using

> I also checked the still-unmerged PR #5533 separately. It is not folded into this branch. The combined code needs to cache the effective GauXC model name (`NONE` or the resolved SKALA path) and preserve #5533's replicated runtime for conventional GauXC; a synthetic integration with those semantics passes precommit and local compilation.

Yup, there were bound to be some conflicts with pulling parts of the GauXC implementation to a cache module. I will try creating a branch of #5533 that uses this PR #5340 as a basis. I think this is the easier ordering.

jpoto (approved):
looks good, with exception of questions about earlier code

oschuett (approved):

franzpoeschel:
@oschuett This implicitly merged #5533 now, too. I rebased this PR on #5533 yesterday, to anticipate conflicts that will be introduced by it
I just noticed that I had linked the wrong PR in the description, updated now

Growl1234:
Looks like GauXC failures are happening with [Spack pdbg](https://storage.googleapis.com/cp2k-ci/run-cp2k-spack-pdbg-2b9dde00_report.txt) tester in #5594, which might be related to this PR.

franzpoeschel:
> Looks like GauXC failures are happening with [Spack pdbg](https://storage.googleapis.com/cp2k-ci/run-cp2k-spack-pdbg-2b9dde00_report.txt) tester, which might be related to this PR.

I will check

src/xc/xc_gauxc_cache.F
Growl1234, line 1: Can you remove the `GO TO`s appear in this file?

src/qs_environment_types.F
```diff
@@ -322,6 +324,7 @@ MODULE qs_environment_types
       TYPE(mo_set_type), DIMENSION(:), POINTER              :: mos_last_converged => NULL()
       ! tblite
       TYPE(tblite_type), POINTER                            :: tb_tblite => Null()
+      TYPE(cp_gauxc_cache_type)                             :: gauxc_cache
```
Growl1234, line 327: The new `TYPE` needs to be initialized.

src/xc/xc_gauxc_cache.F
```diff
@@ -0,0 +1,430 @@
...
+      TYPE(cp_gauxc_integrator_type)               :: gradient_integrator
+      INTEGER                                      :: mpi_comm
+      LOGICAL                                      :: is_init = .FALSE.
+      REAL(c_double), DIMENSION(:, :), ALLOCATABLE :: last_positions
+
+   CONTAINS
+
+      PROCEDURE, PUBLIC                            :: needs_gradient_grid => gauxc_cache_type_needs_gradient_grid
+      PROCEDURE, PUBLIC                            :: max_l => gauxc_cache_type_max_l
+   END TYPE cp_gauxc_cache_type
```
Growl1234, line 67: The new variables need to be initialized.

Growl1234 (commented):
And regarding [coding conventions](https://storage.googleapis.com/cp2k-ci/dashboard_conventions_report.txt) issue.

src/xc/xc_gauxc_cache.F
franzpoeschel: That would imply removing error handling
I guess I can do that, but that would impair the code quality rather than improve it

Growl1234: I think, there should be ways to properly deal with error-handing you mentioned... @fstein93 What do you think?

franzpoeschel: Admittedly, I am new to Fortran and I have transferred this pattern from C programming, where *scoped resource cleanup on error* is widely considered to be one of the two remaining justified uses of `goto` (the other being *break from inner loop*). The alternatives in both cases are error-prone `if then else` constructions with runtime flags.
Might be that Fortran has a better solution of which I am unaware. I can alternatively rewrite this in the mentioned `if then else` style, but the result will be less readable than the current state.

franzpoeschel:
> Looks like GauXC failures are happening with [Spack pdbg](https://storage.googleapis.com/cp2k-ci/run-cp2k-spack-pdbg-2b9dde00_report.txt) tester in #5594, which might be related to this PR.

These failures were introduced by #5533, not by this PR. As written above, #5533 was just accidentally merged along with this PR.

#5533 [patches the behavior of GauXC](https://github.com/cp2k/cp2k/pull/5533/changes#diff-0a08fc7ab333df2cbe36d94c3acce5dfe5e07752229d6f5ff1a11c20bf462352), and uses that patch to remove the old xmat factor correction [from CP2K](https://github.com/cp2k/cp2k/pull/5533/changes#diff-5ccf377192a45ff7f4e8fa9713ca16dafe3a359457a64a62a4cc7c251a6ccfbe). 
This of course breaks any test run not using the patched version of GauXC. Reapplying the xmat factor fix also gives the correct results again.

I am not too happy about maintaining a divergent version of GauXC as part of CP2K's toolchain, and the previous agreement some weeks ago on this had been to keep the xmat factor correction inside CP2K (discussion with @awvwgk). So I am not sure why #5533 moves this back into GauXC.

@DCM-Uni-Paderborn We should discuss this. Patching GauXC is fine *temporarily* if we plan to contribute the changes upstream to GauXC. Otherwise, relying on patched behavior breaks all installations that do not use this particular patched setup, including several CI runs of CP2K.

src/xc/xc_gauxc_cache.F
fstein93: I have already stumbled across it while working on #5597 . I do get the idea and find it quite readable although there is probably a missing `GOTO 500` [here](https://github.com/franzpoeschel/cp2k/blob/745a15c6e08864a4607a7292701e2bb91902a44e/src/xc/xc_gauxc_cache.F#L298). A potentially lower performance from missing compiler optimizations is also probably not the concern here.
An alternative to `GOTO` is a sequence of subroutines, each calling the other one. You can restrict their scope to `gauxc_cache_init` by binding them to `gauxc_cache_init` via `CONTAINS` (see for instance [here](https://github.com/cp2k/cp2k/blob/39fe4c47c6ffbb58ee35c86b875c1acad04cd37b/src/start/cp2k_runs.F#L792)).

franzpoeschel: Will try that. The cleanup section only requires `cache` and `status`, so creating new functions should not be too cumbersome.

Growl1234:
> I am not too happy about maintaining a divergent version of GauXC as part of CP2K's toolchain, and the previous agreement some weeks ago on this had been to keep the xmat factor correction inside CP2K (discussion with @awvwgk).

Fully agreed. We should ensure the differenct dependency building path consistent.

Growl1234:
I think, anyways we need to resolve the dashboard failure. So, shall we just roll back the #5533 part?

franzpoeschel:
> So, shall we just roll back the #5533 part?

Likely not necessary, but let me check tomorrow. It might be fine to roll back only the factor correction.

PR #5341, opened 2026-06-03
Reorganize electronic-structure analysis docs

Growl1234:
This PR moves the GW band-structure page and RESP charge page from the properties section into the electronic-structure analysis section, which provides a clearer structure as a basis for future documentation expansion on related electronic-structure analysis topics.


JWilhelm (approved):
Looks good to me!

Issue #4828, opened 2026-02-12
CP2K build including LIBXSMM using spack


mkrack:
A successful build of CP2K including LIBXSMM using Spack requires currently a [local spack recipe for LIBXSMM](https://github.com/cp2k/cp2k/blob/master/tools/spack/spack_repo/cp2k_dev/packages/libxsmm/package.py) which differs from the upstream spack-packages recipe
```
31c31
<     version("1.17-cp2k", commit="e0c4a2389afba36c453233ad7de07bd92c715bec")
---
>     version("1.17-cp2k", commit="6f883620f58afdeebab28039fc9cf580e76a5ec6")
90,94d89
<     variant(
<         "wrap",
<         default="0",
<         description="Set WRAP to determines the kind of routine called for intercepted GEMMs",
<     )
105c100
<     # depends_on("binutils+ld+gas@2.33:", type="build")
---
>     depends_on("binutils+ld+gas@2.33:", type="build")
109c104
<     # requires("target=x86_64:", when="@:1")
---
>     requires("target=x86_64:", when="@:1")
128,131c123,126
<         # if spec.target.family == "aarch64":
<         #   make_args += ["PLATFORM=1"]
<         # else:
<         #   make_args += ["SYM=1"]
---
>         if spec.target.family == "aarch64":
>             make_args += ["PLATFORM=1"]
>         else:
>             make_args += ["SYM=1"]
150,152d144
< 
<         wrap_val = spec.variants["wrap"].value
<         make_args += ["WRAP={0}".format(wrap_val)]
```
@hfp Could you review these changes? Although these changes work with CP2K, they might break the build for other codes or platforms.

Growl1234:
The libxsmm team have plan to release 2.0 around June 30th (or at latest around July 13). Currently the main branch has had a reliable CMake system to consume (at least for libXS/DBCSR/CP2K), so once the new release is out it will be possible to update Spack official recipe of libxsmm, use its CMake system in CP2K Spack workflow, and drop the local recipe.

Growl1234:
The diff here regarding specific `1.17-cp2k` has been synced in https://github.com/spack/spack-packages/pull/5354. I'm afraid that Spack official build of DBCSR and CP2K never used the `1.17-cp2k` version commit before...

mkrack:
[...]

Issue #4122, opened 2025-04-04
libxsmm: grid_unittest fails using Spack/CMake


mkrack:
After updating the `libxsmm` version for the CP2K build using Spack/CMake from version `1.17` to `main-2024-09` - the latter one is now the same `libxsmm` version (git hash `e0c4a2389afba36c453233ad7de07bd92c715bec`) currently used by the toolchain build - the `grid_unittest` is always failing with the Spack build while it passes with the toolchain build. A few (2-4) other tests are also failing randomly. The toolchain build sets explicitly `WRAP=0` while the Spack build only sets `SYM=1`. Adding `WRAP=0` and/or setting `SYM=0` in the Spack build doesn't solve the problem with `grid_unittest`.

@hfp Do you have any idea why this happens?

Growl1234:
Since libxsmm/libxs has been re-integrated and switched to CMake build for CP2K, I think we can consider this issue as resolved.

PR #5343, opened 2026-06-03
Reintegrate with LIBXS, LIBXSTREAM, and LIBXSMM

hfp:
* LIBXS (https://libxs.readthedocs.io/) foundational library.
  - Implements SMMs with different JIT providers, e.g., MKL-JIT.
  - LIBXSMM only provides kernels (reg. upstream version).
* LIBXSTREAM (https://libxstream.readthedocs.io/).
  - Revised OpenCL backend now formally separate.
* CMake and Toolchain
  - Eliminated mkl_sequential/mkl_gnu_thread mismatch.
  - Toolchain adjusted for libxs/libxstream/libxsmm.
  - Case-insensitive CP2K_USE_ACCEL.  
    DBCSR originally wanted lower case.  
    This is to harmonize with DBCSR.
  - Allow CP2K's FYPP to be used for DBCSR.
    This enables DBCSR Tarball versions.
* Regenerated all Docker recipes.
  - This make FYPP available for DBCSR build.
  - This enables DBCSR Tarball versions.


tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -489,6 +496,8 @@ with_dbcsr="__INSTALL__"
 with_fftw="__INSTALL__"
 with_libint="__INSTALL__"
 with_libxsmm="__INSTALL__"
+with_libxs="__INSTALL__"
+with_libxstream="__INSTALL__"
```
Growl1234: Libxstream requires OpenCL headers to be installed first, otherwise it gives errors. Therefore, it's better not to enable it by default unless user explicitly set `--enable-opencl`?
```suggestion
with_libxstream="__DONTUSE__"
```

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -1083,6 +1098,22 @@ if [ "${ENABLE_OPENCL}" = "__TRUE__" ]; then
...
   fi
+  if [ "${with_libxs}" = "__DONTUSE__" ]; then
+    report_warning ${LINENO} "When enabling OpenCL, libxs is needed."
+    with_libxs="__INSTALL__"
+  fi
+  if [ "${with_libxstream}" = "__DONTUSE__" ]; then
+    report_warning ${LINENO} "When enabling OpenCL, libxstream is needed."
+    with_libxstream="__INSTALL__"
+  fi
+fi
```
Growl1234, line 1108: Moreover, there could be a check:
```suggestion
else
  if [ "${with_libxstream}" != "__DONTUSE__" ]; then
    report_warning ${LINENO} "Libxstream cannot be used without OpenCL enabled; disabling."
    with_libxstream="__DONTUSE__"
  fi
fi
```

Growl1234 (changes requested):

Growl1234:
Besides above comments, I also have concerns if this would break Ubuntu GCC tests and Fedora RPM-build, as release versions of libxs, libxstream, and new release of DBCSR are not yet available. Of course, though, I would be very happy if their releases are (nearly) ready and there are plans to provide prebuilt packages through package managers.

tools/toolchain/scripts/stage4/install_libxs.sh
```diff
@@ -0,0 +1,121 @@
...
+        FC=$FC \
+        FORTRAN=1 \
+        PREFIX=${pkg_install_dir} \
+        install > install.log 2>&1 || tail_excerpt install.log
+      cd ..
+      write_checksums "${install_lock_file}" "${SCRIPT_DIR}/stage4/$(basename ${SCRIPT_NAME})"
+      if [ ! -d ${pkg_install_dir}/lib/pkgconfig ]; then
+        mkdir -p ${pkg_install_dir}/lib/pkgconfig
+      fi
+      cp ${pkg_install_dir}/lib/*.pc ${pkg_install_dir}/lib/pkgconfig/
```
Growl1234: IIRC, the installation of libxs already puts `*.pc` into `lib/pkgconfig`.
```suggestion
```

tools/toolchain/scripts/stage4/install_libxstream.sh
```diff
@@ -0,0 +1,130 @@
...
+        CC=$CC \
+        LIBXSROOT="${LIBXSROOT}" \
+        PREFIX=${pkg_install_dir} \
+        install > install.log 2>&1 || tail_excerpt install.log
+      cd ..
+      write_checksums "${install_lock_file}" "${SCRIPT_DIR}/stage4/$(basename ${SCRIPT_NAME})"
+      if [ ! -d ${pkg_install_dir}/lib/pkgconfig ]; then
+        mkdir -p ${pkg_install_dir}/lib/pkgconfig
+      fi
+      cp ${pkg_install_dir}/lib/*.pc ${pkg_install_dir}/lib/pkgconfig/
```
Growl1234: Here the same
```suggestion
```

tools/toolchain/install_cp2k_toolchain.sh
hfp: Indeed.

hfp:
> Besides above comments, I also have concerns if this would break Ubuntu GCC tests and Fedora RPM-build, as release versions of libxs, libxstream, and new release of DBCSR are not yet available. Of course, though, I would be very happy if their releases are (nearly) ready and there are plans to provide prebuilt packages through package managers.

This is indeed the idea, get this passing/merged and have some followup PR.

hfp:
Since this change requires DBCSR to have received the same treatment, it needs to build a SHA-based DBCSR for time being. The generic Tarball from GH is (unlike a hand-crafted releases) without Submodules sourced-in, hence FYPP is not found when DBCSR is build. The plus side is, we can now refer DBCSR by SHA (or release) and the coupling is effectively reduced. CP2K now does not need an official DBCSR release anymore but can freely chose and regularly test with any DBCSR revision.

alazzaro:
I can make a new DBCSR RC next week....

hfp:
> I can make a new DBCSR RC next week....

Thank you! That's indeed what is also expected for LIBXS/LIBXSTREAM like official releases. The good things is now, I can settle this PR and for whatever release we already tested it and a follow-up PR can put the official releases.

Growl1234 (approved):
Really nice refactoring! Looking forward to the official releases.

hfp:
This PR will receive:

* Official releases from LIBXS, LIBXSTREAM, and DBCSR.
* Finalized OpenCL GPU backend support.
* Removed boilerplate to feed LIBXS.

The latter means, LIBXS currently does not rely on preprocessed Fortran code and hence call-sites detect "kernel providers" such as MKL-JIT and LIBXSMM. LIBXSMM can now be taken from upstream (there will be no releases but rolling/stable main branch). LIBXSMM fully supports ARM64 target variety and other architectures (beside of x86-64).

mtaillefumier:
Unfortunately this PR breaks quite a few things and should be revisited
- A bare cp2k build does not work anymore. Using this command for instance 
```
cmake -DCP2K_USE_MPI=ON -DCP2K_USE_LIBINT2=ON -DCP2K_USE_LIBXC=ON -DCP2K_BLAS_VENDOR=OpenBLAS -GNinja -DCP2K_ENABLE_PW_GPU=OFF -DCMAKE_BUILD_TYPE=Release -DCP2K_USE_ACCEL=CUDA -DCMAKE_CUDA_ARCHITECTURES=70 -DCP2K_USE_LIBXS=ON ..
```
leads to this error
```
/opt/spack/linux-cascadelake/gcc-14.3.0-jujr3kbnr5h4rwlnjjzh6tn3nwy272zl/bin/gcc -D__BLAS -D__LIBXS -D__OFFLOAD_CUDA -I/opt/spack/linux-cascadelake/libint-2.11.2-kuhcnzfka4o4mgruf6bnsiba3uhn6rlz/include/libint2 -I/opt/spack/linux-cascadelake/libint-2.11.2-kuhcnzfka4o4mgruf6bnsiba3uhn6rlz/include -I/home/mathieut/Projects/codes/cp2k/src -I/home/mathieut/Projects/codes/cp2k/build_cuda/src -I/home/mathieut/Projects/codes/cp2k/src/base -I/home/mathieut/Projects/codes/cp2k/src/common -I/home/mathieut/Projects/codes/cp2k/src/motion -I/home/mathieut/Projects/codes/cp2k/src/dbm -I/home/mathieut/Projects/codes/cp2k/build_cuda/src/mod_files -isystem /opt/spack/linux-cascadelake/cuda-12.9.1-2za2urd3iw3kvz3lipkqde2qqlpp3mks/targets/x86_64-linux/include -isystem /opt/spack/linux-cascadelake/openblas-0.3.32-r4xwn46qu6zyeiotvuepiaan74blfsik/include -isystem /usr/libxs -isystem /opt/spack/linux-cascadelake/openmpi-5.0.10-u3p4ztsdvc7ihvv7cfe6zkouldscb25l/include -isystem /opt/spack/linux-cascadelake/openmpi-5.0.10-u3p4ztsdvc7ihvv7cfe6zkouldscb25l/lib -O3  -std=gnu11 -fPIE -g -fno-omit-frame-pointer -Wno-deprecated-declarations -Wno-vla-parameter -O3 -march=native -mtune=native -funroll-loops -fopenmp -MD -MT src/CMakeFiles/dbm_miniapp.dir/dbm/dbm_multiply_cpu.c.o -MF src/CMakeFiles/dbm_miniapp.dir/dbm/dbm_multiply_cpu.c.o.d -o src/CMakeFiles/dbm_miniapp.dir/dbm/dbm_multiply_cpu.c.o -c /home/mathieut/Projects/codes/cp2k/src/dbm/dbm_multiply_cpu.c
/home/mathieut/Projects/codes/cp2k/src/dbm/dbm_multiply_cpu.c:17:10: fatal error: libxs_gemm.h: No such file or directory
   17 | #include <libxs_gemm.h>
```
when `libxs` and `libxsmm` are properly installed.
CMake also print this at the end which is inconsistent with the real installation path.
```
  - LIBXS
    - include directories: /usr/libxs
    - libraries: /usr/lib64/libxs.a
```
- The library detection should **not** rely on variables to be defined (`LIBXSTREAMROOT`, LIBXSROOT, LIBXSMMROOT)
- The cp2k::LIBXS target is currently never referenced, so it is either unnecessary and should be dropped, or its intended use should be wired up explicitly; in its present form it only adds dead configuration.
- The option `NO_DEFAULT_PATH` in find_{path,library} in libxsmm.cmake should also be avoided because  /usr, /usr/local or CMAKE_INSTALL_PATH may be completely ignored (breaking Spack).
- The Spack build does not work either because CP2K_USE_LIBXS is on by default. 
- It breaks the documented policy taken to set all dependencies OFF by default (they are few exceptions). 
- The compilation is likely to fail even when `CP2K_USE_LIBXS=OFF`
- The option `CP2K_USE_LIBXSMM` was deleted for no apparent reason. 
- If `LIBXSMM` is a hard dependency of `libxs`, it should be clearly reflected in the `CMakeLists.txt`
- the `cp2kConfig.cmake.in` should be updated to reflect transitive dependencies as this block 
```
  if(@CP2K_USE_LIBXS@)
    set(CP2K_USE_LIBXS @CP2K_USE_LIBXS@)
    find_dependency(LIBXS REQUIRED)
  endif()
```
is not aligned with the corresponding block in the CMakeLists.txt

```
if(CP2K_USE_LIBXS)
  find_package(LIBXS REQUIRED)
  find_package(LIBXSMM)
  if(LIBXSMM_FOUND)
    message(STATUS "Using LIBXS + LIBXSMM for Small Matrix Multiplication")
  else()
    message(STATUS "Using LIBXS for Small Matrix Multiplication")
  endif()
endif()
```

- The detection fails when LIBXS is installed in the root. CMake print this 
```
-- Found LIBXS: /usr/libxs
```

The libxstream detection is also problematic 

```
  # Locate LIBXSTREAM: user override > DBCSR hint > pkg-config > path probing
  set(LIBXSTREAMROOT
      ""
      CACHE PATH "Root directory of LIBXSTREAM")
  if(NOT LIBXSTREAMROOT AND DBCSR_LIBXSTREAMROOT)
    set(LIBXSTREAMROOT "${DBCSR_LIBXSTREAMROOT}")
  endif()
  if(NOT LIBXSTREAMROOT AND PkgConfig_FOUND)
    pkg_check_modules(_LIBXSTREAM QUIET libxstream)
    if(_LIBXSTREAM_FOUND)
      pkg_get_variable(LIBXSTREAMROOT libxstream prefix)
    endif()
  endif()
  if(NOT LIBXSTREAMROOT)
    foreach(_dir "${CMAKE_SOURCE_DIR}/../libxstream" "$ENV{HOME}/libxstream")
      if(EXISTS "${_dir}/include/libxstream.h")
        set(LIBXSTREAMROOT "${_dir}")
        break()
      endif()
    endforeach()
  endif()
  if(NOT LIBXSTREAMROOT OR NOT EXISTS "${LIBXSTREAMROOT}/include/libxstream.h")
    message(
      FATAL_ERROR
        "LIBXSTREAMROOT not found. Set -DLIBXSTREAMROOT=/path/to/libxstream")
  endif()
  message(STATUS "Using LIBXSTREAM from ${LIBXSTREAMROOT}")
  set(_libxstream_suffixes_save ${CMAKE_FIND_LIBRARY_SUFFIXES})
  if(BUILD_SHARED_LIBS)
    set(CMAKE_FIND_LIBRARY_SUFFIXES .so .dylib)
  else()
    set(CMAKE_FIND_LIBRARY_SUFFIXES .a)
  endif()
  find_library(
    CP2K_LIBXSTREAM_LIBRARY xstream
    PATHS "${LIBXSTREAMROOT}/lib"
    NO_DEFAULT_PATH)
  if(NOT CP2K_LIBXSTREAM_LIBRARY)
    set(CMAKE_FIND_LIBRARY_SUFFIXES ${_libxstream_suffixes_save})
    find_library(
      CP2K_LIBXSTREAM_LIBRARY xstream
      PATHS "${LIBXSTREAMROOT}/lib"
      NO_DEFAULT_PATH)
  endif()
  set(CMAKE_FIND_LIBRARY_SUFFIXES ${_libxstream_suffixes_save})
  if(NOT CP2K_LIBXSTREAM_LIBRARY)
    message(FATAL_ERROR "libxstream library not found in ${LIBXSTREAMROOT}/lib")
  endif()
``` 

should be moved in a dedicated file because it currently breaks `find_package(cp2k)`.

Growl1234:
Regarding Spack, I think we could wait for the official releases along with Spack recipes (hope it could come soon; libxs, libxsmm, libxstream and dbcsr are all affected), but of course we can temporarily use local recipes. @mkrack, what do you think of this?

Regarding CMake configuration, I would agree with @mtaillefumier's points.

> fatal error: libxs_gemm.h: No such file or directory
>   17 | #include <libxs_gemm.h>

Maybe environment variable `CPATH` is missing? If it works with `CPATH`, then I'd suggest to revise current CMake configuration because `CPATH` should not be manually added if a proper CMake configuration is present.

mkrack:
> Regarding Spack, I think we could wait for the official releases along with Spack recipes (hope it could come soon; libxs, libxsmm, libxstream and dbcsr are all affected), but of course we can temporarily use local recipes. @mkrack, what do you think of this?

Indeed, we should develop local spack recipes for the new packages which can then be uploaded after successful testing with cp2k.

@hfp I was surprised that this pull request was simply merged even though it was clear that it would break all spack-based testers.

mtaillefumier:
This PR creates many pain points and `Spack` is a very serious one as cp2k is used in many spack build pipelines. 
- CP2K_USE_LIBXSMM was not mentioned as deprecated, it was removed. 
- The `libxsmm` and `libxs` detection logic are very brittle and several changes are not implemented correctly. This will break every package using `cp2k` as a `cmake` dependency (`find_package(cp2k)`). 
- Many recent changes made to the CMake build system are brittle (the libxc detection for instance), although it is a separate issue.

I want to raise same process issue very clearly: unreviewed PRs like this cause a lot of avoidable work for centers, people maintaining Spack or downstream packages, and cp2k developers maintaining build scripts. In my opinion, PRs should not be merged by their author; there should be at least one independent reviewer approval before merging.

mtaillefumier:
> Regarding Spack, I think we could wait for the official releases along with Spack recipes (hope it could come soon; libxs, libxsmm, libxstream and dbcsr are all affected), but of course we can temporarily use local recipes. @mkrack, what do you think of this?
> 
> Regarding CMake configuration, I would agree with @mtaillefumier's points.
> 
> > fatal error: libxs_gemm.h: No such file or directory
> > 17 | #include <libxs_gemm.h>
> 
> Maybe environment variable `CPATH` is missing? If it works with `CPATH`, then I'd suggest to revise current CMake configuration because `CPATH` should not be manually added if a proper CMake configuration is present.

It is not good practice to rely on CPATH for include directories. It works with C and C++ but is not supported in Fortran for instance.

Growl1234:
Hmm, I thought some follow-ups would be done before the next release, with release versions of related packages and additional cleanups, though we might have to wait for a week or two. That said, encountering this situation right before an official release is indeed awkward.

Perhaps a reasonable approach would be to first revert these changes, and then integrate them in the next development cycle — there is sufficient time before 2027.1 to fully resolve the related issues within one cycle? Although I am still very positive about the motivation behind this change and look forward to seeing its practical effects.

@oschuett I also want to hear your thoughts...

mtaillefumier:
Do not misunderstand me: I am not against integrating these libraries. Their inclusion itself is reasonable and I agree with the motivation. What I am questioning is how they have been integrated in this PR, because the current implementation breaks several basic `CMake `idioms at once and causes regressions in bare builds, `Spack `builds, and `find_package(cp2k)` usage.

To be clear, my concern here is also about ownership and capacity. There are only a handful of developers (roughly three or four) who are familiar enough with the `CMake `build system to safely make non‑trivial changes.

As a result, the regressions introduced here effectively translate into additional pressure on the same small group of people to debug and repair issues that could have been caught up front. This PR can absolutely be cleaned up, but the way it was merged concentrated the maintenance burden on the few build‑system maintainers without involving them in the design or review.

hfp:
ACK

mtaillefumier:
Thanks for the ACK. Given the current regressions (bare builds, Spack builds, and find_package(cp2k)), I think we still need a concrete plan to get back to a working state.

My proposal would be:

- Revert this PR before the upcoming release to unblock existing pipelines.
- Re‑introduce the changes in smaller, reviewed PRs in the next development cycle, with the CMake maintainers involved from the start.
- Or support @Growl1234 current PR #5365 to solve the issues reported here.

My preference is to contribute to @Growl1234 PR #5365 

If you prefer a different approach to restore working builds and config packages, could you outline it and a rough timeline? I’m happy to help with review or cleanup; I just want to avoid leaving the build system broken for downstream users.

hfp:
I have fixed this already, but I cannot send the PR over the weekend. It's not the broken picture drawn here. I suggest to give this two days into the week.

Growl1234:
> I have fixed this already, but I cannot send the PR over the weekend. It's not the broken picture drawn here. I suggest to give this two days into the week.

Thanks for the quick response! Looking forward to your fixes, though I still think the changes regarding CMake in #5365 is the right way forward. I've also submitted a similar PR on the DBCSR side.

oschuett:
> @oschuett I also want to hear your thoughts...

First of all, I do want to thank Hans for putting in the effort of creating libxs to resolve the stalemate around libxsmm!

Of course, it would have been better to let more people review this change before merging it, but hindsight is 20/20.

> To be clear, my concern here is also about ownership and capacity. 

I've also been thinking about this a lot lately and was actually planning to open an RFC about [code owners](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) next week.

> That said, encountering this situation right before an official release is indeed awkward.

In the past the announced target date  was more like a soft deadline for the start of a feature freeze period. The actual release always happened a few weeks later when all the (major) issues were fixed. As the project grows, we might also have to formalize our release process a bit more...

he-zilong:
Is it correct that the deprecation of `libxsmm` also renders the unresolved issues #4122 and #4828 as moot now?

Growl1234:
> Is it correct that the deprecation of `libxsmm` also renders the unresolved issues #4122 and #4828 as moot now?

No, libxsmm is still there but its role has changed.

hfp:
> Is it correct that the deprecation of `libxsmm` also renders the unresolved issues #4122 and #4828 as moot now?

It's indeed a JIT-kernel provider now. Only a single function is essentially used from LIBXSMM which is part of the core API going forward. Though, LIBXSMM removed some API used to integrate with DBCSR and CP2K plus the confusion about rolling development with no releases (think of Spack and how this confused). LIBXS, among other functionality, now provides the integration for DBCSR and CP2K. LIBXSMM remains a valid "provider" also targeting non-x86 kernels.

mtaillefumier:
My two cents. 

- Are both libxsmm and libxs optional dependencies?
- Does libxs requires libxsmm ? is this dependence soft or hard?
- Does it make sense to build cp2k with libxsmm only if libxs is eventually replacing it?

This matters because it changes the way the dependencies are treated in cmake. If cp2k support for `libxsmm` and `libxs` is decoupled then we should use `option(CP2K_USE_LIBXSMM)` and `option(CP2K_USE_LIBXS)`. If libxs requires libxsmm in some cases then we should use `cmake_dependent_option(CP2K_USE_LIBXSMM)`. 

My current understanding is that libxsmm and libxs can be used independently for now?

Growl1234:
I think, libxsmm can no longer work independently for CMake because it now just provide a JIT-kernel for `libxs`. I'd be in favor of switching to `cmake_dependent_option(CP2K_USE_LIBXSMM)`.

mtaillefumier:
that's exactly the kind of ambiguity I would like to have an answer to because I can build `cp2k+libxsmm` right now.

cmake/CompilerConfiguration.cmake
```diff
@@ -224,3 +224,8 @@ if(CMAKE_C_COMPILER_ID STREQUAL "Cray" AND CMAKE_C_COMPILER_VERSION
   list(FILTER CMAKE_C_IMPLICIT_LINK_LIBRARIES EXCLUDE REGEX "tcmalloc")
   list(FILTER CMAKE_Fortran_IMPLICIT_LINK_LIBRARIES EXCLUDE REGEX "tcmalloc")
 endif()
+
+# Suppress GFortran runtime warnings when LIBXS provides the wrapper
```
oschuett, line 228: I know you already explained this to me, but unfortunately I can't quite remember. Can you remind me why we need to suppress all runtime warnings?

hfp:
[LIBXS 1.0](https://github.com/hfp/libxs/releases/tag/1.0.0) is released. LIBXSTREAM will follow asap.

Growl1234:
Looks like the version changing in files like `CMakeLists.txt` is done after cutting the release, so it still remains `0.0.0`...

hfp:
> Looks like the version changing in files like `CMakeLists.txt` is done after cutting the release, so it still remains `0.0.0`...

Indeed, the RPM stuff did not pickup the tag. Let me see.

hfp:
> > Looks like the version changing in files like `CMakeLists.txt` is done after cutting the release, so it still remains `0.0.0`...
> 
> Indeed, the RPM stuff did not pickup the tag. Let me see.

Ok, the version stamp is "1.0.0" but the artifact is called "1.0". I need to retag.

hfp:
> > > Looks like the version changing in files like `CMakeLists.txt` is done after cutting the release, so it still remains `0.0.0`...
> > 
> > 
> > Indeed, the RPM stuff did not pickup the tag. Let me see.
> 
> Ok, the version stamp is "1.0.0" but the artifact is called "1.0". I need to retag.

No idea, only the RPM/PackIt does not make it. Looks actually bad like "red".

hfp:
> [LIBXS 1.0](https://github.com/hfp/libxs/releases/tag/1.0.0) is released.

[LIBXSTREAM 1.0](https://github.com/hfp/libxstream/releases/tag/1.0.0) is released.

PR #5344, opened 2026-06-03
Reorganize dashboard and reduce Spack testing cadence

oschuett:
Since we had a cloud bill of 673 USD last month, I'm reducing the cadence of some tests:

- spack-psmp-gcc14: `daily` -> `weekly`
- spack-pdbg: `asap` -> `daily`

Furthermore, I also synced the order of tests between `dashboard.conf` and `cp2k-ci.conf`.

### DIFF
--- a/tools/dashboard/README.md
+++ b/tools/dashboard/README.md
@@ -52,7 +52,6 @@ typical entry looks like this:
 
 ```
 [mkrack-pdbg]
-sortkey:     100
 name:        Linux-x86-64-gfortran.pdbg
 host:        PSI, merlinl03
 notify:      off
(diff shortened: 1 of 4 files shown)


Growl1234:
Would it be acceptable to reduce the frequency of some Ubuntu GCC tests? We could keep a few representative tests running daily, e.g. Ubuntu GCC 9, 12, and 15.

oschuett:
I recently started to collect detailed [usage stats](https://storage.googleapis.com/cp2k-ci/usage_stats.txt). Note that for May only the last two weeks were included.

Based on those numbers, moving these three GCC tests from daily to weekly would save only 1.4 %. 

Ultimately we'll need to find a way to have our compute budget match the growth of the project.

PR #5346, opened 2026-06-03
Jvp d3 library

jpoto:
Enable getting parameters for D3 from simple-dftd3. requires tblite install. 
Would enable for example r2scan d3 correction.

PR #5347, opened 2026-06-03
Toolchain: Revise default behaviors

Growl1234:
With this PR tblite would be installed by default, and SIRIUS would have to be requested explicitly (`--with-sirius`).

The change is based on the expected practical value and likely usage frequency of these optional dependencies for users: tblite, along with its subprojects, provides high practical value — not only supporting GFN2-xTB, but also, more importantly, D4 corrections (and possibly extra D3 functional parameters; see #5346) — and is more likely to be used frequently. SIRIUS provides the plane-wave module, while CP2K is most commonly used through Quickstep's efficient GPW/GAPW methods.

### DIFF
--- a/tools/toolchain/install_cp2k_toolchain.sh
+++ b/tools/toolchain/install_cp2k_toolchain.sh
@@ -289,6 +289,7 @@ Specific options of --with-PKG:
                           Default = install
   --with-fmt              Enable the formatting C/C++ library.
                           This package is required for SIRIUS.
+                          Default = no
   --with-libtorch         Enable libtorch as a machine learning framework.
                           This package is required for NequIP and Allegro, and
                           also for installing DeePMD-kit.
@@ -299,7 +300,7 @@ Specific options of --with-PKG:
                           Default = no
   --with-hdf5             Enable the hdf5 library for file format support.
                           This package is used by sirius and trexio.
-                          Default = install
+                          Default = no
   --with-libsmeagol       Enable interface to SMEAGOL NEGF library.
                           This package requires MPI.
                           Default = no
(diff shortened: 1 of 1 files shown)


he-zilong (approved):
From a more conservative angle, making `SIRIUS` not installed by default reduces the chance that its complicated dependency preparation leads to hard-to-diagnose toolchain errors on some old machines; such inconvenience unfortunately outweighs the benefit of having a copy of `SIRIUS` installed by default when the necessity in usage is unclear. After all, users with demand of plane-wave codes are probably installing alternatives in parallel to the CP2K package.

oschuett:
While it's true that CP2K is primarily a GTO code, there a good PW use-cases as well.
So instead of disabling SIRIUS by default, maybe we could try to make its installation more robust?

he-zilong:
> While it's true that CP2K is primarily a GTO code, there a good PW use-cases as well. So instead of disabling SIRIUS by default, maybe we could try to make its installation more robust?

I expect some response to "make its installation more robust" along the lines of "just drop toolchain and use spack already!" But we have gone through that previously and let's not distract this PR...

My point is more focused on the benefit part: currently a few "good PW use-cases" (such as the examples for [how to run calculations like Quantum ESPRESSO](https://www.cp2k.org/howto:running_qe_computation)) do not seem to justify SIRIUS sufficiently as a core dependency installed by default. If, say, generally wavefunction post-processing tools can handle CP2K's gaussian-type basis sets and SIRIUS's plane-wave basis sets equally well, or that SIRIUS gets so fast and robust that what is CP2K's nightmare is just a piece of cake for SIRIUS, then there would be stronger reasons to get SIRIUS installed by default and persuade users to resolve installation issues exclusive to SIRIUS rather than simply skipping it.

oschuett:
>  "just drop toolchain and use spack already!"

Yeah, that's fair. Dropping SIRIUS entirely from the toolchain would actually be more in line with what we have discussed in #4843. And presumably the SIRIUS developers would [be ok with that](https://github.com/cp2k/cp2k/pull/5098#issuecomment-4318974440).

Growl1234:
I think this is mostly a trade-off for the default setup. The default toolchain path should focus on functionality that is either broadly used in practice or closely related to CP2K's core strengths. SIRIUS/PW support is valuable, but it seems more specialized and less clearly justified as a default dependency.

Improving the SIRIUS installation is certainly useful, and #5098 was already a step in that direction. Actually I would not like to discuss further about SIRIUS installation itself here, but making it as robust as a Spack-based installation is not realistic within the toolchain given the number of external dependencies involved in SIRIUS; and therefore, dropping SIRIUS from the toolchain could also significantly reduce maintenance burden of toolchain itself... those discussion is a broader cleanup topic and should be left in next development cycles.

PR #5349, opened 2026-06-03
DOS section explicit required and added new unit for electric field

ShridharShanbhag:
Adding a new unit kind Electric Field with the units:
volt per meter (Vm-1), volt per nm (Vnm-1) and volt per angstrom (VA-1).

When including &PROPERTIES, the Density of states is currently always printed with default values even without the &DOS section being explicit. This pull request will ensure that the DOS section must be explicit in order to print the DOS_PDOS.out file.

The EFIELD section gets a new AMPLITUDE option where the electric field strength can be included in volt per meter (default) or volt per nm or volt per angstrom instead of needing to specify INTENSITY.

If BOTH options AMPLITUDE and INTENSITY are used, raise error "Both INTENSITY and AMPLITUDE provided in EFIELD section."
Backward compatibility is maintained such that all existing input files that only contain "INTENSITY" proceed with the calculations as before and yield the same result.

### DIFF
--- a/src/common/cp_units.F
+++ b/src/common/cp_units.F
@@ -1059,7 +1059,7 @@ CONTAINS
          CASE (cp_units_au)
             res = "au_efield"
          CASE (cp_units_none)
-            res = "electric field strength"
+            res = "electric field"
             IF (.NOT. my_accept_undefined) &
                CALL cp_abort(__LOCATION__, &
                              "unit not yet fully specified, unit of kind "// &
--- a/src/common/cp_units.F
+++ b/src/common/cp_units.F
@@ -372,6 +372,9 @@ CONTAINS
          CASE ("AU_EFIELD")
             unit_id(i_unit) = cp_units_au
             kind_id(i_unit) = cp_ukind_efield
+         CASE ("EFIELD")
+            unit_id(i_unit) = cp_units_none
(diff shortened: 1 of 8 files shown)


JWilhelm:
Please do not merge yet, I would need to check further.

JWilhelm:
The PR looks good for me now and could be accepted in my opinion.

src/common/cp_units.F
```diff
@@ -58,7 +58,8 @@ MODULE cp_units
                                  cp_ukind_undef = 8, &
                                  cp_ukind_potential = 9, &
                                  cp_ukind_force = 10, &
-                                 cp_ukind_max = 10
+                                 cp_ukind_efield = 11, &
```
oschuett, line 61: Uh, I just noticed that `11` is assigned twice. That probably not intentional, right?

Unfortunately, I already hit _merge_.

ShridharShanbhag: It is intentional. cp_ukind_max is the maximum number that cp_ukind can be. When we had 10 kinds, it was 10.
Now we have 11 kinds, so it is 11.

cp_ukind_max serves as the size of units (at line 156 in cp_units.F). It is the number of kinds of physical quantities defined. (line 1297, 1301 in the same file).

PR #5352, opened 2026-06-04
OFFLOAD: slightly revised init/deinit

hfp:
- Better incorporate __OFFLOAD_UNIFIED_MEMORY.


hfp:
This was separated from #5343.

hfp:
All mandatory checks passed, but canceling the never starting P100-test makes this PR look like failed. I will merge this if there are no objections.

PR #5357, opened 2026-06-05
Adjust regtests and ignore tblite SCC mixer residual for OT/LS_SCF

Growl1234:
- `tests/QS/regtest-gpw-2-3/Au*`: Uses diagonalization with smearing instead (metal system should not use OT) and makes it fast
- `xTB/regtest-tblite-gfn2/O2_gfn2_uks_ot.inp`: Removes `SCF_GUESS MOPAC` for much better convergence and fixes an error caused from source code.
- `xTB/regtest-3-spglib/si_uks_kp_tblite_mixer_spglib.inp`: Uses `BROYDEN_MIXING` for much better convergence, also removes `SCF_GUESS MOPAC` and disables smearing
- `QS/regtest-gpw-6-2/si8_noort_broy`: make SCF converge and speed up
- Resolve slow tests regarding k-point symmetry
- `QS/regtest-moments-kpoints`: Use magnitudes instead of signed components due to k-point MO phases
- `DFTB/regtest-scc/ch2o_tblite_mixer.inp`: Uses `BROYDEN_MIXING` for much better convergence

This PR does not consider the runtime failures regarding GauXC.


Growl1234:
The test, `QS/regtest-moments-kpoints/C2_pbe_scf_kp.inp`, shows such weird difference:
```
/opt/cp2k/regtesting/TEST-2026-06-03_15-50-25/QS/regtest-moments-kpoints/C2_pbe_scf_kp.inp.out
Spec: {'matcher': 'Dipole_at_kp_1', 'tol': 1e-06, 'ref': -0.544}
Difference too large: 2.00e+00 > 1e-06, value: 0.544.
```
Anyone has idea why this happens? I have a feeling that this difference in sign might not be abnormal, potentially related to k-point MO phases issue. ***[Update: I decided to switch to a different matcher, using magnitudes instead of the real/imaginary components]***

Growl1234:
Ready to merge if all triggered tests pass.

oschuett:
Btw, the CI not only processes new comments but also listens to edits. So, you can just re-use an existing comment. Inserting even a single space character will trigger a re-eval.

tests/matchers.py
```diff
@@ -78,6 +78,54 @@ def run(self, output: str, **kwargs: Any) -> MatchResult:
         return MatchResult("OK", error=None, value=value)  # passed
 
 
+# ======================================================================================
+class ComplexAbsMatcher(Matcher):
```
oschuett, line 82: After pondering this a bit more, I don't think we should do computations in the matchers. If a quantity is relevant enough to be used in a test, then it should appear directly in CP2K's output.

Growl1234: That makes sense. I'll try relacing it with a present and robust reference.

Growl1234: The only existing output that can be read directly while avoiding the phase issue is the Berry curvature. However, that quantity is already covered by another test, and I assume the original author had a reason to use a different kind of reference value here.

Doing derived computations in the matcher is indeed not ideal and I'm going to remove it, but the sign/phase flip of an individual complex k-point dipole matrix element is physically allowed, so testing a signed real/imaginary component directly is also not robust. At minimum, I think the check should be made insensitive to this arbitrary sign/phase, e.g. by comparing the absolute value.

It should also be available to enforce a phase convention in the CP2K source just to stabilize this printed component, but it is not worth the extra complexity to me.

oschuett: Thanks for #5364. 

> The only existing output that can be read directly while avoiding the phase issue is the Berry curvature. 

I was thinking, we could add another column to the output for the abs value.

Alternatively, we could print a [checksum just for testing](https://github.com/search?q=repo%3Acp2k%2Fcp2k%20path%3Amatchers.py%20checksum&type=code).

PR #5338, opened 2026-06-02
grid: Clean up GPU backend and fix race condition 

mtaillefumier:
- Fix a compilation error when using rocm 6.3 as `__shfh_down_sync` is supported inconsistently on rocm
- Use less confusing notations for the grid parameters (use x,y,z instead of indices)
- Integrate and collocate are agnostic to the lattice representation and ordering the lattice vectors

PR #5358, opened 2026-06-05
Revert PR 5351

DCM-Uni-Paderborn:
This reverts #5351, which was merged after I had misread the intended scope discussion - sorry! The merged change kept only the `grid_gpu_context.h` change, whereas the follow-up discussion intended to remove that controversial part and keep the other non-controversial changes separately.

I will prepare the corrected follow-up without the `grid_gpu_context.h` change and without duplicating the regtest work now covered by #5357.

### DIFF
--- a/src/grid/gpu/grid_gpu_context.h
+++ b/src/grid/gpu/grid_gpu_context.h
@@ -280,10 +280,34 @@ public:
   }
 
   void check_orthorhombicity(const bool ortho) {
-    // Keep the PW grid classification. The GPU orthorhombic kernels assume the
-    // axis-aligned grid layout selected by cell2grid; mutually orthogonal
-    // lattice vectors alone are not enough.
-    orthorhombic_ = ortho;
+    if (ortho) {
+      orthorhombic_ = true;
+      return;
+    }
+    T norm1, norm2, norm3;
+    bool orthogonal[3] = {false, false, false};
+    norm1 = dh_[0] * dh_[0] + dh_[1] * dh_[1] + dh_[2] * dh_[2];
+    norm2 = dh_[3] * dh_[3] + dh_[4] * dh_[4] + dh_[5] * dh_[5];
+    norm3 = dh_[6] * dh_[6] + dh_[7] * dh_[7] + dh_[8] * dh_[8];
+
+    norm1 = 1.0 / sqrt(norm1);
+    norm2 = 1.0 / sqrt(norm2);
+    norm3 = 1.0 / sqrt(norm3);
+
+    /* x z */
+    orthogonal[0] =
+        ((fabs(dh_[0] * dh_[6] + dh_[1] * dh_[7] + dh_[2] * dh_[8]) * norm1 *
+          norm3) < 1e-12);
+    /* y z */
+    orthogonal[1] =
+        ((fabs(dh_[3] * dh_[6] + dh_[4] * dh_[7] + dh_[5] * dh_[8]) * norm2 *
+          norm3) < 1e-12);
+    /* x y */
+    orthogonal[2] =
+        ((fabs(dh_[0] * dh_[3] + dh_[1] * dh_[4] + dh_[2] * dh_[5]) * norm1 *
+          norm2) < 1e-12);
+
+    orthorhombic_ = orthogonal[0] && orthogonal[1] && orthogonal[2];
   }
 
   inline void copy_to_host(T *data__, offloadStream_t &stream) {


oschuett:
Yeah, my idea was to have separate PRs with dedicated discussions on all the thorny topics.

Since `check_orthorhombicity()` is one of those topics, we  might as well have that discussion here.

To repeat @mtaillefumier's point from https://github.com/cp2k/cp2k/pull/5351#discussion_r3362116680:

> – The GPU backend should be completely agnostic to the lattice representation, since the mathematical properties of the lattice do not depend on the particular ordering of its lattice vectors. Removing the orthogonality tests does not solve the underlying issue, it only masks it.

While your argument is mathematically sound, the fact that some tests are sensitive to this, shows that numerically there is a difference between the two code paths. And I think, this is a strong reason to not switch dynamically, but to make a decision once and then stick to it throughout the entire simulation. Hence, I'd be in favor of keeping #5351.

mtaillefumier:
> Yeah, my idea was to have separate PRs with dedicated discussions on all the thorny topics.
> 
> Since `check_orthorhombicity()` is one of those topics, we might as well have that discussion here.
> 
> To repeat @mtaillefumier's point from [#5351 (comment)](https://github.com/cp2k/cp2k/pull/5351#discussion_r3362116680):
> 
> > – The GPU backend should be completely agnostic to the lattice representation, since the mathematical properties of the lattice do not depend on the particular ordering of its lattice vectors. Removing the orthogonality tests does not solve the underlying issue, it only masks it.
> 
> While your argument is mathematically sound, the fact that some tests are sensitive to this, shows that numerically there is a difference between the two code paths. And I think, this is a strong reason to not switch dynamically, but to make a decision once and then stick to it throughout the entire simulation. Hence, I'd be in favor of keeping #5351.

The fact that tests are sensitive to these parameters only indicate that there is a bug because the results for collocate and integrate should be agnostic to the lattice **representation**. Hiding the problem by removing the test or freezing a flag is not a solution; it just makes the underlying inconsistency harder to see.  Cell optimization is a typical example where orthogonality of the lattice vectors can be lost or restored during the simulation. The data structure takes this into account every time it is updated and will use the generic kernel when orthogonality is lost (the criteria is at 10^-12) so leaving it dynamical is not an issue. 

The generic and orthogonal paths should give the same answer for an orthorhombic lattice, regardless of the canonical parameter. In practice, they differ because the summation ordering still depends on how the spherical cutoff is applied. As described in PR #5338, both the CPU and GPU implementations of the orthogonal kernel contain the same inconsistency: they mix a discrete cutoff radius expressed in lattice points with quantities computed in real space; the true real‑space cutoff radius should be used instead. Doing so reveals the flaw but also breaks some existing unit tests, which is why this inconsistency persists.

The solution, implemented in PR #5338, effectively removes an assumption I made when writing the GPU backend. The CPU backend has same issue although it is never triggered because the orthogonality parameter is set to true only when the canonical representation of a orthorhombic lattice is used. 

The error is in line https://github.com/cp2k/cp2k/blob/446a0bc27ea643677b7ac1b23eb15b76e79862d5/src/grid/gpu/grid_gpu_collocate.cu#L220 and below. Both the collocate and integrate kernels have the same issue. This line of code assumes that the *matrix representation of the lattice vectors* is diagonal which is wrong because only the metric is diagonal in a normal or orthogonal space. 

The norm of the lattice vectors should be used to calculate the sphere boundaries instead of the vectors coordinates. Doing so remove the sensitivity to the canonical parameter (it might be relevant elsewhere) and restore almost all failing tests exempt 

```
regtesting/TEST-2026-06-05_16-21-52/QS/regtest-gpw-2-3/Au13ico_mtd.inp.out
Spec: {'matcher': 'E_total', 'tol': 5e-13, 'ref': -434.3956435445884}
Difference too large: 8.24e-09 > 5e-13, value: -434.3956399672304.

regtesting/TEST-2026-06-05_16-21-52/QS/regtest-gpw-2-3/Au12_rmsd_AB_mtd.inp.out
Spec: {'matcher': 'E_total', 'tol': 3e-11, 'ref': -398.91130074125454}
Difference too large: 9.42e-10 > 3e-11, value: -398.9113011168657.

regtesting/TEST-2026-06-05_16-21-52/QS/regtest-gpw-2-3/Au12_rmsd_A_mtd.inp.out
Spec: {'matcher': 'E_total', 'tol': 2e-11, 'ref': -398.90196058523196}
Difference too large: 6.70e-07 > 2e-11, value: -398.902227910262.

regtesting/TEST-2026-06-05_16-21-52/QS/regtest-kp-2/cc2.inp.out
Spec: {'matcher': 'E_total', 'tol': 1e-12, 'ref': -46.0362556979842}
Difference too large: 7.19e-07 > 1e-12, value: -46.03622257740221.

regtesting/TEST-2026-06-05_16-21-52/QS/regtest-kp-1-spg/c_cell_opt_kpsym_guard.inp.out
Spec: {'matcher': 'N_special_kpoints', 'tol': 0.0, 'ref': 1}
Difference too large: 3.00e+00 > 0.0, value: 4.0.

regtesting/TEST-2026-06-05_16-21-52/QS/regtest-rtbse/h2_ft.inp.out
Spec: {'matcher': 'M118', 'tol': 0.001, 'ref': 8.4430117}
Difference too large: 1.00e+00 > 0.001, value: 0.0.

```

The second-to-last failure (N_special_kpoints) is a quantity computed by spglib and is completely independent of the grid backend. I do not yet know why the last test fails;  however the total energy and density (using the validation parameter of the grid backend) are within 10^-16 between the CPU and GPU backends for that case.

Side note on reproducibility: 

- We should not expect the grid backend to produce bitwise‑identical results across different collocate/integrate implementations (even GPUs). For instance, the calculations of the `hab` coefficients are deterministic, but the order of operations differs between the REF and GPU backends, so small differences are expected. 
- The GPU implementation of the collocate function uses atomic operations extensively and will always be noisier than the CPU implementation because the Gaussian summation order is random and is hardware specific; differences of order 10^-12 (but not 10^-8) are possible and unavoidable.

oschuett:
> both the CPU and GPU implementations of the orthogonal kernel contain the same inconsistency: they mix a discrete cutoff radius expressed in lattice points with quantities computed in real space; 


This discretization of the radius is a very mild approximation that allows to cache the sphere bounds, which eliminates a ton of `sqrt()` calls. For the CPU backend this yields a speedup of ~10%. Perhaps the GPU backend does not benefit as much, but I think we should still aim to keep the two backends as consistent as possible. Otherwise we'll hurt reproducibility, make testing harder, and confuse the users.

> The fact that tests are sensitive to these parameters only indicate that there is a bug because the results for collocate and integrate should be agnostic to the lattice representation. 

If there additionally is a bug then we're really discussion two separate issues, right? What kind of bug do you suspect?

mtaillefumier:
I think we’re talking past each other. I fully accept that caching sphere bounds via a discretized radius can give up to ~10% speedup, and I’m not arguing against that optimization. My point is different: combining a radius defined in grid points with quantities in Cartesian coordinates is not a “mild approximation,” it is mathematically inconsistent, because it makes the result depend on the lattice representation (canonical vs non‑canonical, small rotations) rather than only on the physical cell.

When tests move, there are only a few generic explanations:
– Randomness in floating‑point order (atomics, parallel reductions, etc.),
– A code change that sends the program down a different, but equivalent, path, or
– A genuine bug, i.e. a wrong or inconsistent formula or buggy dependency.

The Au13 MTD test (regtest-gpw-2-3/Au13ico_mtd.inp) already gives different answers for MPI = 1, 2, or 4, with or without my changes.

oschuett:
Ok, got it. Thanks for the explanation. Then let's merge this and then #5338 to fix it.

I'm still not convinced that the GPU backend should be allowed to ignore the [orthorhombic flag](https://github.com/cp2k/cp2k/blob/06f838dc4a894b1dda983365af91778f112c2c87/src/grid/grid_task_list.h#L21). While the dynamic approach might give a speedup under rare circumstances, IMHO that's not worth adding a moving part. However, I realize now that this is a separate discussion to be had.

PR #5361, opened 2026-06-06
Fix dashboard code regressions

DCM-Uni-Paderborn:
This keeps the PR focused on the non-controversial dashboard fixes, while adding one targeted regtest input fix for the current `ssmp` GauXC/SKALA dashboard failure. It deliberately excludes the `grid_gpu_context.h` change, dependency/test suppressions, and broad tolerance loosening.

Summary:
- initialize the CP L-BFGS optimizer scratch arrays explicitly and only copy optional convergence targets when they are present
- add a FFTW fast path for plans with zero or one partial FFT, avoiding the OpenMP region when there is nothing to distribute
- mask and restore IEEE halting mode around Torch/OneDFT calls that may otherwise trip dashboard floating-point exception settings
- make GauXC status printing use explicit units and preserve stderr handling
- fix XAS TP-SCF core-state vector selection by copying the identified state and aborting if none is found
- restore `H2_SKALA_CDFT_CI.inp` to converged CDFT/SCF settings instead of comparing an intentionally unconverged one-step state, and update the `M077` reference from the CI-equivalent `ssmp` GauXC/SKALA container

Testing:
- `./make_pretty.sh`
- `git diff --check origin/master..HEAD`
- Terok `Dockerfile.test_ssmp` GauXC subset before the regtest fix: reproduced `H2_SKALA_CDFT_CI.inp = -0.7026160817` wrong vs `-0.7037517464`
- Terok `Dockerfile.test_ssmp` GauXC subset with the regtest fix: `QS/regtest-gauxc` `76 / 76`, `H2_SKALA_CDFT_CI.inp = -0.7043868633` OK


DCM-Uni-Paderborn:
> /cp2kci ssmp

OK looks like a ssmp-/OpenMP-specific Bug, I will have a look - thanks!

PR #5362, opened 2026-06-06
Improve native SKALA CUDA routing and Torch execution

DCM-Uni-Paderborn:
This PR improves the native-grid SKALA CUDA/Torch execution path by reducing repeated routing, Tensor bookkeeping, and payload construction around atom chunks and subchunks.

## Changes

- Improve native SKALA CUDA multi-GPU handling.
- Cache atom-chunk return routing to reduce repeated MPI/Tensor bookkeeping.
- Add atom subchunk inference for native SKALA, including routed gradient accumulation.
- Use Torch views for atom subchunks instead of rebuilding full payloads.
- Collapse RKS dynamic routing payloads and gradient routing overhead.
- Read SKALA Torch metadata through the existing Torch metadata API after model load.
- Call the SKALA Torch energy method directly where possible.
- Allow atom chunks on a single rank.
- Add CUDA-oriented auto-tuning for native SKALA atom subchunk sizes.
- Fix the GauXC keyword count for the new native SKALA CUDA controls.

## Scope

This PR is limited to native-grid SKALA GPU/Torch execution plumbing.

Periodic SKALA stress/k-points/NLCC support and GauXC CUTLASS/LWD controls are intentionally kept separate.

## Validation

- CP2K pre-push hook on the changed files: OK
- `git diff --check`: OK
- Native SKALA CUDA path tested in the prepared branch context with serial/MPI/OpenMP/CUDA smoke coverage.


src/torch_api.F
```diff
@@ -636,6 +714,58 @@ END SUBROUTINE torch_c_model_load
 #endif
    END SUBROUTINE torch_model_load
 
+! **************************************************************************************************
+!> \brief Loads a Torch model and reads two metadata entries.
+! **************************************************************************************************
+   SUBROUTINE torch_model_load_with_metadata(model, filename, key1, value1, key2, value2)
```
oschuett: As discussed in a previous PR, the same functionality can be archived by calling `torch_model_load()` once and `torch_model_read_metada()` twice.

Btw, I think the `extra_files` feature in Torch is broken by design and should not be used. Instead, metadata can be conveniently stored in model attributes, which can be read from a loaded model via `torch_model_get_attr()`. 

See the PAO code for an example:

https://github.com/cp2k/cp2k/blob/e0f0c179264e9b32184eb80bae544c589135706a/tools/pao-ml/pao/model.py#L27-L29

https://github.com/cp2k/cp2k/blob/e0f0c179264e9b32184eb80bae544c589135706a/tools/pao-ml/pao/model.py#L63

https://github.com/cp2k/cp2k/blob/e0f0c179264e9b32184eb80bae544c589135706a/src/pao_model.F#L89

DCM-Uni-Paderborn: Dear Ole, 

thanks and done in 76b60884c6. I removed the new `torch_model_load_with_metadata()` / C++ fused metadata-loading wrapper and switched the SKALA path back to `torch_model_load()` plus two `torch_model_read_metadata()` calls.

I agree that model attributes are the cleaner long-term representation. I kept that out of this PR for now because the current SKALA TorchScript artifacts still provide these entries as metadata; we can move the SKALA exporter and CP2K reader to `torch_model_get_attr()` together once the model artifacts carry the attributes.

src/skala_gpw_functional.F
```diff
@@ -274,27 +285,56 @@ SUBROUTINE skala_gpw_eval(vxc_rho, vxc_tau, exc, rho_r, rho_g, tau, xc_section,
...
+         IF (native_grid_use_cuda) THEN
+            native_grid_atom_chunk_max_rows = auto_atom_chunk_max_rows(features, &
+                                                                       rho_r(1)%pw_grid%para%group)
+         ELSE
+            native_grid_atom_chunk_max_rows = 0
+         END IF
+      END IF
+      IF (native_grid_diagnostics .AND. features%uses_atom_chunks .AND. &
+          rho_r(1)%pw_grid%para%group%mepos == 0) THEN
+         WRITE (UNIT=cp_logger_get_default_io_unit(), FMT="(T2,A,1X,I0)") &
```
fstein93: I think you need to check whether the output unit is positive.

DCM-Uni-Paderborn: Ciao Frederick,

thanks, fixed in 9bb625adbd. The native SKALA diagnostics now cache `cp_logger_get_default_io_unit()` in `iw` and only write for positive units: the atom-chunk max-row diagnostic is guarded with `iw > 0`, and `print_native_grid_diagnostics` plus the CUDA-device diagnostics return early for `iw <= 0`.

PR #5372, opened 2026-06-07
Docs: Add fragment links and note about citation keys to syntax cheat sheet

oschuett:


he-zilong:
Thank you!

By the way, could the link to "Publication" also say where to create or look up the tags in bibliography? It is somewhat confusing that it is not always the same as the key in `src/common/bibliography.F`...

PR #5373, opened 2026-06-07
RPM-build: Temporarily disable SMM backend until LibXS is provided in Fedora repo

Growl1234:


Growl1234:
It seems this PR has not got attention...

And I also have some questions about how to handle the SMM backend in RPM builds. I'm afraid we can no more rely on `libxsmm`, since our plan is to remove `FindLIBXSMM.cmake`, and there will be no new releases of `libxsmm` — even if there will be, it won't choose CMake as the build method in the RPM build recipe. `Libxs` has a BLAS fallback path, and if in the future it adopts a CMake-based RPM-build with official releases, then this issue would be much easier to handle.

oschuett:
> It seems this PR has not got attention...

Sorry, is it ready to be merged?

Growl1234:
Yes

PR #5374, opened 2026-06-08
Enable optional GauXC CUTLASS toolchain builds

DCM-Uni-Paderborn:
This adds a default-off CP2K toolchain switch for building the bundled GauXC with CUDA/CUTLASS local work driver support.

Scope:
- add `--enable-gauxc-cutlass[=yes|no]` to the CP2K toolchain
- require `--enable-cuda=yes`, `--with-gauxc=install`, and a CUTLASS-capable CUDA architecture (`A100`, `A40`, `H100`, `GB10`)
- pass `GAUXC_ENABLE_CUDA=ON`, `GAUXC_ENABLE_CUTLASS=ON`, and `CMAKE_CUDA_ARCHITECTURES` only for that opt-in path
- keep the default GauXC toolchain build unchanged (`GAUXC_ENABLE_CUDA=OFF`, `GAUXC_ENABLE_CUTLASS=OFF`)
- carry the needed GauXC compatibility fixes in CP2K's embedded GauXC patch so this does not depend on external GauXC PRs being merged first

This is intentionally limited to toolchain compatibility. It does not change CP2K input semantics, SKALA/native-grid behavior, periodic SKALA code, benchmarks, or paper/documentation results.

Validation:
- `git diff --check`
- GauXC embedded patch dry-run on toolchain GauXC revision with `patch -l -p1 --dry-run`
- `python3 tools/precommit/precommit.py ...` on the four touched files from an ASCII temp worktree: OK
- toolchain dry-runs:
  - `--dry-run --with-gauxc=install`: `--enable-gauxc-cutlass = __FALSE__`
  - `--dry-run --with-gauxc=install --enable-cuda=yes --gpu-ver=A100 --enable-gauxc-cutlass=yes`: `--enable-gauxc-cutlass = __TRUE__`
  - missing CUDA rejects `--enable-gauxc-cutlass`
  - `--with-gauxc=system` rejects `--enable-gauxc-cutlass`
  - `--gpu-ver=V100` rejects `--enable-gauxc-cutlass`


Growl1234:
I’m a bit hesitant about further integrating additional GauXC-related features into the toolchain. The current toolchain installation of GauXC is not ideal, since GauXC still downloads quite a few additional components through CMake’s `FetchContent`. However, it's clear that adding all of these dependencies to the toolchain would  increase its maintenance burden. If we consider slimming down the toolchain in the future, GauXC may be one of the first candidates for removal.

DCM-Uni-Paderborn:
> I’m a bit hesitant about further integrating additional GauXC-related features into the toolchain. The current toolchain installation of GauXC is not ideal, since GauXC still downloads quite a few additional components through CMake’s `FetchContent`. However, it's clear that adding all of these dependencies to the toolchain would increase its maintenance burden. If we consider slimming down the toolchain in the future, GauXC may be one of the first candidates for removal.

I see what you say, but I hope GauXC is here to stay. If my upstream GauXC PRs will be eventually merged, life will start to get slightly easier, but I am not too optimistic since so far no reaction ...

oschuett:
> However, it's clear that adding all of these dependencies to the toolchain would increase its maintenance burden.

> I see what you say, but I hope GauXC is here to stay. 

I think, both are valid points. The problem is that, GauXC and the toolchain are maintained by different people. Hence there needs to be a mechanism for the two teams to push back at each other and negotiate. So, I think this again points in the direction of code owners.

tools/toolchain/scripts/stage6/install_gauxc.sh
```diff
@@ -113,14 +122,15 @@ case "${with_gauxc}" in
         -DGAUXC_ENABLE_MPI="${gauxc_enable_mpi}" \
         -DGAUXC_ENABLE_OPENMP="${gauxc_enable_openmp}" \
         -DGAUXC_ENABLE_TESTS=OFF \
-        -DGAUXC_ENABLE_CUDA=OFF \
+        -DGAUXC_ENABLE_CUDA="${gauxc_enable_cuda}" \
```
oschuett, line 125: Note that we're currently not testing GauXC with CUDA. So, we're really flying blind here.
https://github.com/cp2k/cp2k/blob/be2ee82e432e7819c92ac14c11fb96c5075acdbf/cmake/cmake_cp2k.sh#L172

DCM-Uni-Paderborn: Dear Ole,

Yes, agreed. This PR intentionally keeps the path default-off because CP2K CI does not currently exercise GauXC with CUDA. The change only adds an opt-in toolchain build mode and keeps the existing default GauXC build at `GAUXC_ENABLE_CUDA=OFF` / `GAUXC_ENABLE_CUTLASS=OFF`.

I have validated the toolchain option handling and the embedded GauXC patch application locally, but I agree that this is not a substitute for a real CP2K+GauXC+CUDA build/run. I can add that limitation explicitly to the PR description and will provide external validation logs from a CUDA machine before claiming runtime support.

DCM-Uni-Paderborn: Having said that, I nevertheless think the PR has merit, or?

oschuett:
> Having said that, I nevertheless think the PR has merit, or?

Yes, this PR certainly has merit and I'm generally excited about GauXC.

The thing is that, while many folks would like to drop the toolchain in favor of Spack, @Growl1234 and @he-zilong have stepped up to continue the toolchain as a robust installation path for offline environments. So, I'm just looking for a way to keep everyone happy.

I guess the two questions are:
- @DCM-Uni-Paderborn, why can't you use Spack?
- @Growl1234, what would it take to make you comfortable including GauXC into the toolchain?

Growl1234:
The most important thing for me is to have a robust workaround for offline installation, and this is even not only for user convenience. Based on my inspection, installing GauXC takes almost 10 minutes in CI, and my tests on my own PC show that the time spent in `FetchContent` is not negligible.

DCM-Uni-Paderborn:
> > Having said that, I nevertheless think the PR has merit, or?
> 
> Yes, this PR certainly has merit and I'm generally excited about GauXC.
> 
> The thing is that, while many folks would like to drop the toolchain in favor of Spack, @Growl1234 and @he-zilong have stepped up to continue the toolchain as a robust installation path for offline environments. So, I'm just looking for a way to keep everyone happy.
> 
> I guess the two questions are:
> 
> * @DCM-Uni-Paderborn, why can't you use Spack?
> * @Growl1234, what would it take to make you comfortable including GauXC into the toolchain?

Dear Ole, dear Zilong,

well the embarrassing truth is that I'm not used to Spack. However, all my students love it, I assume that I can use Spack for my own builds and validation.

However, the reason I still care about the CP2K toolchain here, except my own ignorance, is that it remains an important installation path for CP2K users, especially in offline or restricted HPC environments. So I would prefer GauXC/SKALA not to become usable only through Spack.

Greetings, 
Thomas

DCM-Uni-Paderborn:
Wrt. the maintenance concern, a compromise can definitely be found s.o. simply has to explain to me what is the best way ...

DCM-Uni-Paderborn:
For this PR I would suggest keeping the scope as-is: default-off, opt-in only, and no change to the default GauXC/toolchain path.

I agree that the broader offline-installation concern is real. My proposal would be to handle that in a follow-up PR ... That should address the robustness concern without expanding this CUDA/CUTLASS enabling PR.

So unless you prefer otherwise, I would keep #5374 focused and treat offline GauXC dependency staging as the next toolchain cleanup step.

Growl1234:
> Wrt. the maintenance concern, a compromise can definitely be found s.o. simply has to explain to me what is the best way ...

I can think of two possible approaches.

If this is to be fixed only on the CP2K toolchain side, the only reasonable — though quite cumbersome — solution would be to identify all packages that are necessarily downloaded through `FetchContent` during the GauXC toolchain installation, and then provide separate toolchain installation scripts for them. To avoid confusion, these packages, together with GauXC itself, could be placed in a separate directory, for example `stage10`.

Unfortunately, such an approach would still require careful inspection. In particular, we would need to check whether `FetchContent` is still triggered even when the corresponding dependencies have already been installed and correctly added to the environment. If it is still triggered, then the problem would likely lie in the CMake/build configuration of GauXC or one of its dependencies, and would not be something that the CP2K toolchain can solve on its own.

Another, more ideal approach would be to add a `subprojects` directory directly in the upstream repository, containing the required Git submodules, and then use a complete release tarball that includes these submodule packages. In that case, nothing special would need to be done on the toolchain side. However, this would require careful discussion with the upstream developers, and would ultimately depend on whether they are willing to adopt such an approach.

In either case, I tend to think that this is not a very easy problem to solve.



> For this PR I would suggest keeping the scope as-is: default-off, opt-in only, and no change to the default GauXC/toolchain path.
> 
> I agree that the broader offline-installation concern is real. My proposal would be to handle that in a follow-up PR ... That should address the robustness concern without expanding this CUDA/CUTLASS enabling PR.
> 
> So unless you prefer otherwise, I would keep #5374 focused and treat offline GauXC dependency staging as the next toolchain cleanup step.

Yes, the offline-installation issues should indeed to be treated seperately.

PR #5376, opened 2026-06-08
Update UZH basis sets and pseudopotentials

DCM-Uni-Paderborn:
This PR updates the CP2K UZH data files with the improved UZH basis-set and GTH pseudopotential entries.

## Changes

- Updates selected entries in `data/BASIS_MOLOPT_UZH` from the improved UZH basis-set data.
- Updates selected entries in `data/POTENTIAL_UZH` from the improved UZH GTH pseudopotential data.
- Keeps the existing data-file organization unchanged; entries are replaced by primary data key rather than adding new files.
- Preserves backward-compatible GGA aliases for the updated Ba basis sets and Ca q10 pseudopotential.

## Local checks

- Replaced 48 `BASIS_MOLOPT_UZH` entries and 6 `POTENTIAL_UZH` entries.
- Checked that primary entry keys and aliases remain unique in both files.
- Checked that q-values in updated basis-set names match available PBE/SCAN/PBE0 pseudopotentials.
- Checked that occupations in the updated pseudopotentials sum to the q-values in their names.
- Verified the restored Ba basis-set and Ca `GTH-GGA-q10` pseudopotential aliases with CP2K smoke tests.
- Checked that no q-less `GTH-PBE` alias remains in `data/POTENTIAL_UZH`.
- Ran `git diff --check` on the changed data files.

The corresponding release repository, `DCM-Uni-Paderborn/UZH-protocol`, has been updated with matching data files.


juerghutter:
I can live with this, but we have to put up a warning that PP and Basis sets for some
rather common elements have changed. People who want to stay with the old version
(e.g. for consistency within a project) should download an old version of these files.

he-zilong (commented):
There was a similar scenario two years ago, where PR #3343 updated the Na pseudopotentials by reusing the existing names for new data and renaming the old data with an "_old" suffix. I only learned about this after someone else asked why the same reftraj run produced significantly different (~0.6 a.u.) energies on 2025.1 and 2024.1 versions.

Is it possible to, say, print the git commit that last updated the data when outputting the name of basis sets and pseudopotentials?

mkrack:
LGTM, as far as that is possible just by visual inspection.

mkrack:
> There was a similar scenario two years ago, where PR #3343 updated the Na pseudopotentials by reusing the existing names for new data and renaming the old data with an "_old" suffix. I only learned about this after someone else asked why the same reftraj run produced significantly different (~0.6 a.u.) energies on 2025.1 and 2024.1 versions.
> 
> Is it possible to, say, print the git commit that last updated the data when outputting the name of basis sets and pseudopotentials?

That's indeed a problem. We miss a proper versioning of potentials and basis sets. Appending `_old` is a weak (poor) workaround (btw, if I am not mistaken that happened also for U once.)

Introducing a unique hash value for each atomic potential and basis set dataset which is attached to or even included explicitly into the dataset for each element could be an appropriate solution. The hash value is stored in the restart file and can be printed in the output file, too. On restart, CP2K can detect a mismatch and notify the user about the mismatch or even stop the run.

DCM-Uni-Paderborn:
I agree, we see the problem here again since I am using the UZH versions as standard for everything we do for quite a while already, which is why the Wannier90 retests fail. It's an easy fix to print the commit in the output without detailed git knowledge if it the commit is written inside the file. Maybe s.o. can do that ... ;)

tgmaxson:
First time directly interacting with the project here.  I only incidentally discovered this PR / change by chance after a colleague pointed it out to me.

This is a major issue for people using MLIPs / multiple clusters as if a potential is referred to by name, they will get inconsistent results.  This has previously been an issue in VASP / Materials Project that caused some grief by my understanding (there were silent updates within the versioning system if I understand correctly).

As a recently converted user of CP2K for some projects, this is very alarming as this would be easy to miss.  I strongly suggest just releasing a new basis set file with a version appended to it.  No need to call the old one `_old`, just call this new one `_v2` and set the old one to `_v1`.  By removing the non versioned file, you will force an obvious breaking change to users to pick which is correct.  Optionally just retain the old file as a symbolic link to `_v1`.

PR #5379, opened 2026-06-09
Fix truncation warning in __shfl_down_sync mask

mkrack:
Replace non-standard __CUDA_CC__ macro with __activemask(), eliminating the need for architecture-specific masking and avoiding integer truncation warnings.

### DIFF
--- a/src/grid/gpu/grid_gpu_integrate.cu
+++ b/src/grid/gpu/grid_gpu_integrate.cu
@@ -683,11 +683,7 @@ __launch_bounds__(64) void integrate_kernel(const kernel_params dev_) {
         // Warp-level reduction (32 lanes)
         // After this loop, lane 0 of warp 0 holds the result
         for (int offset = 16; offset > 0; offset >>= 1) {
-#if defined(__CUDA_CC__)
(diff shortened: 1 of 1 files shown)


mtaillefumier:
@mkrack: There is a mistake in the current implementation. I already fixed it in PR #5333. the `#ifdef` statement is necessary because not every version of hip support `__shfl_down_sync()`.

The real fix is 

```
#if defined(__CUDACC__)
__shfl_down_sync(0xffffffff, ...)
#else
__shfl_down(,)
#endif
```

using `__activemask` is also incorrect because all threads need to call the instruction to get the right answer. It works here because all threads are guarantied to reach that point.

mkrack:
@mtaillefumier Thanks for checking. Feel free to revert this commit and apply the proper fix.

mtaillefumier:
I rebased on it and fixed it in pr #5338. I want to run the regtests locally to be sure that I did not miss anything.

mtaillefumier:
I do not have merge rights and I won't merge a PR without someone else reviewing it either. 

I have more serious concern about cosma failing some tests

PR #5380, opened 2026-06-09
Floquet physics

ShridharShanbhag:
This pull request adds a method to evaluate Floquet Bandstructures in CP2K.

We add a new section &FLOQUET in the &BANDSTRUCTURE section under PROPERTIES.
One can provide the amplitude and frequency of an input monochromatic light pulse along with the polarisation and phase offsets. This can be used to define a light pulse with circular, elliptical or linear polarisation along any arbitrary direction.

The code will then produce two files: FLOQUET_DOS.out and QUASI_ENERGIES.bs. These will contain the Floquet density of states and the quasi_energies for each kpoint in the k-path provided in the BANDSTRUCTURE_PATH.


src/floquet_utils.F
```diff
@@ -0,0 +1,657 @@
...
+      REAL(KIND=dp)                                      :: fbz_max, fbz_min, omega, qe
+      REAL(KIND=dp), ALLOCATABLE, DIMENSION(:)           :: quasi_energies
+      TYPE(mo_set_type), DIMENSION(:), POINTER           :: mos
+
+      CALL timeset(routineN, handle)
+
+      CALL get_qs_env(qs_env, mos=mos)
+      CALL get_mo_set(mo_set=mos(1), nao=nao)
+
+      fname = "QUASI_ENERGIES.bs"
```
fstein93: Would it make sense to have the file name as an input parameter or make it dependent on the project title?

src/floquet_utils.F
```diff
@@ -0,0 +1,657 @@
...
+      CALL timeset(routineN, handle)
+
+      CALL get_qs_env(qs_env, mos=mos)
+
+      CALL get_mo_set(mo_set=mos(1), mu=mu)
+
+      energy_step = bs_env%energy_step_floquet
+      E_min = mu - bs_env%energy_window_floquet
+      n_E = SIZE(a_k)
+      fname = "FLOQUET_DOS.out"
```
fstein93: Same here

src/floquet_utils.F
```diff
@@ -0,0 +1,657 @@
...
+      REAL(KIND=dp)                                      :: amplitude, omega
+      REAL(KIND=dp), DIMENSION(3)                        :: e_vec, phi, polarisation
+      TYPE(mo_set_type), DIMENSION(:), POINTER           :: mos
+
+      CALL timeset(routineN, handle)
+
+      ! Builds the matrix that occupies the off diagonal blocks in the Floquet Matrix H_F
+
+      polarisation(:) = bs_env%floquet_polarisation(:)
+      IF (SQRT(SUM(polarisation**2.0_dp)) < EPSILON(0.0_dp)) THEN
```
fstein93: Please use `SQRT(SUM(polarisation**2))`

ShridharShanbhag: Thank you. I made this change in the latest version.

src/floquet_utils.F
ShridharShanbhag: Thank you for your comment. I added file name for both cases as an input.

PR #5384, opened 2026-06-09
Add cuda_arch propagation to GauXC spack recipe

mkrack:


tools/spack/spack_repo/cp2k_dev/packages/gauxc/package.py
```diff
@@ -115,6 +115,8 @@ def cmake_args(self):
             self.define("GAUXC_ENABLE_CUTLASS", spec.satisfies("+cutlass")),
             self.define_from_variant("CMAKE_POSITION_INDEPENDENT_CODE", "pic"),
         ]
+        if "+cuda" in spec:
+            args.append(self.define_cuda_architectures(self))
```
mtaillefumier: If my reading is correct this package uses cmake as build system. Is there any particular reason to call a method that I never seeing before instead of 

-DCMAKE_CUDA_CARCHITECTURES=spec.variants["cuda_arch"].value

or does this function is a Spack funtionality that returns the right option automatically?

mtaillefumier:
Would it be possible to open an issue containing a list of new packages to add to the cp2k package.py recipe?

mtaillefumier:
@mkrack: if you prefer to keep the function call then the correct syntax is
```
if self.spec.satisfies("+cuda"):
            args.append(CMakeBuilder.define_cuda_architectures(self))
```

mtaillefumier (approved):
Thanks for fixing the typo. No objections to merge this

tools/spack/spack_repo/cp2k_dev/packages/gauxc/package.py
mkrack: As I understood that CMake function from Spack's `CudaPackage` automatically detects and maps the Spack `cuda_arch` values directly to the `CMAKE_CUDA_ARCHITECTURES` variable. That appears to me convenient.

PR #4922, opened 2026-03-07
Vibrational analysis: extend width of data field

he-zilong:
The old format for printing out IR and Raman intensities from a vibrational
analysis job is `3(1X,F12.6,8X)`, whose fixed 12-character width for float
type data may be faced with overflow in practice; this PR proposes sacrificing
width of trailing blank and extending width of data field with a new format
`3(1X,ES18.6E3,2X)` that is less likely to become a string of asterisks (****)
and obscure the data result. Note that this does not imply that an abnormally
high intensity from vibrational analysis is free from numeric error and/or
chemically sensible; the user is still responsible of verifying data.

For downstream output file parsers, this should be fine unless a restriction
of data type `F12.6` is expected.

Closes #4219.

PR #5387, opened 2026-06-12
Vibrational analysis: reduce width of data field (rework #4922)

he-zilong:
Three months ago I created PR #4922 where the format of vibrational analysis printout was revised from `3(1X,F12.6,8X)` to `3(1X,ES18.10E3,2X)` so that the wider data fields would allow for potentially large values (in particular the IR intensities and Raman activities) to be printed.

However, during tests for #5312 I noticed that the format becomes unaligned and overflows 80 characters per line like this (some surrounding texts included for comparison):
```
 REPLICA| layout of the replica grid, number of groups                         8
 REPLICA| layout of the replica grid, size of each group                       1
 REPLICA| MPI process to grid (group,rank) correspondence:
  (   0 :    0,   0)  (   1 :    1,   0)  (   2 :    2,   0)  (   3 :    3,   0)
  (   4 :    4,   0)  (   5 :    5,   0)  (   6 :    6,   0)  (   7 :    7,   0)

 VIB|                        NORMAL MODES - CARTESIAN DISPLACEMENTS
 VIB|
 VIB|                         1                    2                    3
 VIB|Frequency (cm^-1) -5.8075433121E+000    1.9526543945E+001    2.2835157787E+001
 VIB|Red.Masses (a.u.)  3.9948000000E+001    3.9948000000E+001    3.9948000000E+001
 VIB|Frc consts (a.u.) -5.0988401965E-005    5.7641762593E-004    7.8830543305E-004
  ATOM  EL             X     Y     Z        X     Y     Z        X     Y     Z
     1  Ar            0.44 -0.24 -0.02    -0.13 -0.26  0.41     0.21  0.28  0.27
     2  Ar            0.01  0.50  0.02    -0.29  0.02 -0.41     0.02 -0.31  0.14
```

As a remedy, this PR modifies the format to `3(1X,ES17.10E2,2X)` and adjusts the alignment of the data accordingly, which renders output like this:

```
 REPLICA| layout of the replica grid, number of groups                         8
 REPLICA| layout of the replica grid, size of each group                       1
 REPLICA| MPI process to grid (group,rank) correspondence:
  (   0 :    0,   0)  (   1 :    1,   0)  (   2 :    2,   0)  (   3 :    3,   0)
  (   4 :    4,   0)  (   5 :    5,   0)  (   6 :    6,   0)  (   7 :    7,   0)

 VIB|                        NORMAL MODES - CARTESIAN DISPLACEMENTS
 VIB|
 VIB|                         1                    2                    3
 VIB|Frequency (cm^-1) -5.8075433121E+00    1.9526543945E+01    2.2835157787E+01
 VIB|Red.Masses (a.u.)  3.9948000000E+01    3.9948000000E+01    3.9948000000E+01
 VIB|Frc consts (a.u.) -5.0988401965E-05    5.7641762593E-04    7.8830543305E-04
  ATOM  EL               X     Y     Z       X     Y     Z       X     Y     Z
     1  Ar              0.44 -0.24 -0.02   -0.13 -0.26  0.41    0.21  0.28  0.27
     2  Ar              0.01  0.50  0.02   -0.29  0.02 -0.41    0.02 -0.31  0.14
```

While the new format has fewer digits than the one in #4922, it is still better than `3(1X,F12.6,8X)` in the last _stable release_ (2026.1) version and permits values up to +/-9.9999999999E+99 to be displayed properly rather than a string of asterisks. It is right-justified with the atom displacement data below.

Although this PR does not touch the underlying mathematical algorithm, it may still be necessary to fix the reference value of a few regtests due to output parsing. I apologize for the inconvenience.

(Edit: the previous version used a slightly different format with fewer significant digits.)

### DIFF
--- a/src/motion/vibrational_analysis.F
+++ b/src/motion/vibrational_analysis.F
@@ -864,23 +864,23 @@ CONTAINS
          to = MIN(from + 2, nvib)
          WRITE (UNIT=iw, FMT="(T2,'VIB|',13X,3(8X,I5,8X))") &
             (icol, icol=from, to)
-         WRITE (UNIT=iw, FMT="(T2,'VIB|Frequency (cm^-1)',3(1X,ES18.10E3,2X))") &
+         WRITE (UNIT=iw, FMT="(T2,'VIB|Frequency (cm^-1)',3(1X,ES15.8E2,5X))") &
             (freq(icol), icol=from, to)
          IF (ASSOCIATED(intensities_d)) THEN
-            WRITE (UNIT=iw, FMT="(T2,'VIB|IR int (KM/Mole) ',3(1X,ES18.10E3,2X))") &
+            WRITE (UNIT=iw, FMT="(T2,'VIB|IR int (KM/Mole) ',3(1X,ES15.8E2,5X))") &
                (fint*intensities_d(icol)**2, icol=from, to)
          END IF
          IF (ASSOCIATED(intensities_p)) THEN
-            WRITE (UNIT=iw, FMT="(T2,'VIB|Raman (A^4/amu)  ',3(1X,ES18.10E3,2X))") &
+            WRITE (UNIT=iw, FMT="(T2,'VIB|Raman (A^4/amu)  ',3(1X,ES15.8E2,5X))") &
                (pint*intensities_p(icol), icol=from, to)
-            WRITE (UNIT=iw, FMT="(T2,'VIB|Depol Ratio (P)  ',3(1X,ES18.10E3,2X))") &
+            WRITE (UNIT=iw, FMT="(T2,'VIB|Depol Ratio (P)  ',3(1X,ES15.8E2,5X))") &
                (depol_p(icol), icol=from, to)
-            WRITE (UNIT=iw, FMT="(T2,'VIB|Depol Ratio (U)  ',3(1X,ES18.10E3,2X))") &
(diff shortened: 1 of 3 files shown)


fstein93:
In generally approved. But I wonder whether it is possible to display one significant digit more (9 instead of 8).

he-zilong:
> In generally approved. But I wonder whether it is possible to display one significant digit more (9 instead of 8).

It seems that this will either overflow 80 characters again (as `3(1X,ES16.9E2,4X)`), make the data conjoined with the unit at the start of the line (as `3(ES16.9E2,5X)`; I do not want to touch the labels like `VIB|IR int (KM/Mole)` because they are already condensed), or lose one digit on the exponential part (as `3(1X,ES15.9E1,5X)`).

fstein93:
And what about using 3 spaces instead of 4? (If I understand the format identifier correctly?)

he-zilong:
> And what about using 3 spaces instead of 4? (If I understand the format identifier correctly?)

Wait, that is indeed another possible idea; there is a gap of 4 spaces between the per-atom components of different modes, which may be compressed to allow for one or two more spaces for the data above. Let me try...

he-zilong:
And done! The net effect of this PR is taking one digit off from the exponent part of `3(1X,ES18.10E3,2X)`; now `3(1X,ES17.10E2,2X)` has the same width of mantissa and allows for a tight alignment which has been updated in the first post.

PR #5388, opened 2026-06-12
Update conventions suppression

Growl1234:
Follow up to #5346 and #5343

### DIFF
--- a/tools/conventions/conventions.supp
+++ b/tools/conventions/conventions.supp
@@ -1,4 +1,3 @@
-ai_contraction_sphi.F: Module "libxsmm" USEd without ONLY clause or not PRIVATE https://cp2k.org/conv#c002
 almo_scf.F: Found WRITE statement with hardcoded unit in "almo_scf_init" https://cp2k.org/conv#c012
 al_system_dynamics.F: Found WRITE statement with hardcoded unit in "dump_vel" https://cp2k.org/conv#c012
 base_hooks.F: Found CALL m_abort in procedure "cp_abort" https://cp2k.org/conv#c102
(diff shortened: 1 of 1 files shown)


oschuett (commented):
LGTM

fstein93:
Approved. It requires fixes on the library site because of missing symbols in the interface and/or missing `private` declarations in their module files.
see grimme-lab/mctc-lib#112, dftd3/simple-dftd3#192

Growl1234:
I just ran a local conventions test (via Dockerfile) based on #5365, and it returns:
```
Found 5 unsuppressed issues:

  libxs_jit.F: Found type libxs_cpuid_t without initializer https://cp2k.org/conv#c016
  libxs_jit.F: Found type libxs_fprint_t without initializer https://cp2k.org/conv#c016
  libxs_jit.F: Found type libxs_matdiff_t without initializer https://cp2k.org/conv#c016
  libxs_jit.F: Found type libxs_registry_info_t without initializer https://cp2k.org/conv#c016
  libxs_jit.F: Found type libxs_timer_info_t without initializer https://cp2k.org/conv#c016
```
@hfp Would you prefer to fix these warnings properly, or simply suppress them for now? If we choose the latter, I would add the suppressions in this PR rather than that already wide-ranging one.

jpoto:
When the libraries are fixed we then also need: #5393

fstein93:
mctc-lib fixed the visibility of symbols in their master branch.

Growl1234:
Thank you all to report issues to the upstream! The suppressions regarding `mctc-lib` and `s-dftd3` could be removed once the next release of `tblite` is out.

> Would you prefer to fix these warnings properly, or simply suppress them for now? If we choose the latter, I would add the suppressions in this PR rather than that already wide-ranging one.

@hfp I've added those conventions issues above to suppressions.

PR #5391, opened 2026-06-12
RIRS: fix bounds of grids for geometries with mixed atom kinds

MGraml:


RitajTyagi:
@MGraml Thanks for pointing out and fixing the bug.

PR #5397, opened 2026-06-13
Toolchain: Print categorized configuration summary

Growl1234:
Replaces the raw list of resolved `--with-*` options with a clearer, categorized report that is printed in both `dry-run` and normal execution modes. Packages are grouped by whether they will be installed, detected from the system, linked from user-provided paths, or not used, while package lists are formatted compactly in columns to keep the output readable.

Also prevents `--install-all` from incorrectly marking MKL/ACML as an installable package, since they could only be detected from the system or linked from an existing installation.

### DIFF
--- a/tools/toolchain/install_cp2k_toolchain.sh
+++ b/tools/toolchain/install_cp2k_toolchain.sh
@@ -32,7 +32,6 @@ export ROOTDIR="${PWD}"
 export SCRIPTDIR="${ROOTDIR}/scripts"
 export BUILDDIR="${ROOTDIR}/build"
 export INSTALLDIR="${ROOTDIR}/install"
-export SETUPFILE="${INSTALLDIR}/setup"
 export TOOLKIT_SCRIPT="${SCRIPTDIR}/tool_kit.sh"
 
 # ------------------------------------------------------------------------
@@ -649,9 +648,6 @@ while [ $# -ge 1 ]; do
         exit 1
       fi
       export INSTALLDIR="${1#--install-dir=}"
-      export SETUPFILE="${INSTALLDIR}/setup"
-      cp "${SCRIPTDIR}"/tool_kit.sh "${INSTALLDIR}"/
-      export TOOLKIT_SCRIPT="${INSTALLDIR}/tool_kit.sh"
       ;;
     --no-check-certificate)
       export DOWNLOADER_FLAGS="--no-check-certificate"
@@ -1261,6 +1257,12 @@ fi
 # ------------------------------------------------------------------------
 
 mkdir -p "${INSTALLDIR}"
+export SETUPFILE="${INSTALLDIR}/setup"
+
+if [ "${INSTALLDIR}" != "${ROOTDIR}/install" ]; then
+  cp "${SCRIPTDIR}"/tool_kit.sh "${INSTALLDIR}"/
+  export TOOLKIT_SCRIPT="${INSTALLDIR}/tool_kit.sh"
+fi
 
 # Select the correct compute number based on the GPU architecture
 case ${GPUVER} in
--- a/tools/toolchain/install_cp2k_toolchain.sh
+++ b/tools/toolchain/install_cp2k_toolchain.sh
@@ -662,6 +662,8 @@ while [ $# -ge 1 ]; do
         if [ "${ii}" != "intel" ] &&
           [ "${ii}" != "intelmpi" ] &&
           [ "${ii}" != "amd" ] &&
+          [ "${ii}" != "mkl" ] &&
+          [ "${ii}" != "acml" ] &&
           [ "${ii}" != "cusolvermp" ]; then
           eval "with_${ii}=__INSTALL__"
         fi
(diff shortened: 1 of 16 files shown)


Growl1234:
Known issue to be resolved: When using MKL to provide BLAS/FFTW/ScaLAPACK and not installing libvdwxc, `fftw` should appear in `Packages not used:` list, but actually this is in `Packages to be installed:` list. The logic deciding whether to use MKL's own FFTW should be moved to the main script.

he-zilong:
Could you please include an example snippet of the new printout format?

To be honest and nitpicky, despite `install_cp2k_toolchain.sh` already requiring bash in the shebang and using some bash features (like [`${var[i]}` array reference](https://www.shellcheck.net/wiki/SC3054) and [`source` aliase of `.`](https://www.shellcheck.net/wiki/SC3046)), I still feel uncomfortable about new instances here like [`[[]]` conditioning replacing `[]`](https://www.shellcheck.net/wiki/SC3010), [`==` comparison replacing `=`](https://www.shellcheck.net/wiki/SC3014), [`local` variable declaration](https://www.shellcheck.net/wiki/SC3043), [C-style arithmetic `for (())` loops](https://www.shellcheck.net/wiki/SC3005), and `printf -v var ...` replacing `var=$(printf ...)` for writing to variable instead of stdout. I personally stubbornly prefer the old POSIX-compatible style and do not see bash-only syntax as convenient...

Growl1234:
> Could you please include an example snippet of the new printout format?

You can see from regtest report:

```text
Toolchain configuration summary
-------------------------------
System specifications:
   -j                    = 32
  --target-cpu           = native
  --gpu-ver              = no
  --mpi-mode             = mpich
  --math-mode            = openblas

Enabled features:
  --enable-tsan          = no
  --enable-cuda          = no
  --enable-gauxc-cutlass = no
  --enable-hip           = no
  --enable-opencl        = no
  --enable-cray          = no

Packages to be installed:
  - cmake        - ninja        - mpich        - openblas     - fftw
  - libint       - libxc        - gauxc        - libxsmm      - libxs
  - cosma        - scalapack    - elpa         - dbcsr        - plumed
  - spfft        - spla         - gsl          - spglib       - hdf5
  - libvdwxc     - sirius       - libvori      - libtorch     - deepmd
  - ace          - tblite       - pugixml      - libsmeagol   - fmt
  - trexio       - libfci       - greenx       - gmp          - mcl

Packages to be detected from system:
  - gcc

Packages not used:
  - intel        - amd          - openmpi      - intelmpi     - mkl
  - acml         - libxstream   - cusolvermp   - dftd4
```

Growl1234:
- I did not see `[[ ]]` introduced in my patch (I don't like this style either).
- "`==` instead of `=`"  seems to be widely used in toolchain scripts.
- I believe `local a` is safer than `_myfunc_a` because it reduces the risk of temporary helper variables leaking into or colliding with the global script state.
- C style loops and `printf -v` might be worth revising from my side, though I think `printf -v` is cleaner.

Growl1234:
The arithmetic `for` loops are dropped.

Growl1234:
I also fixed the issue that `tool_kit.sh` was copied before the custom install directory was created.

https://github.com/cp2k/cp2k/blob/85b8a9b49c10d557fd7a980df5c112ee8f438efe/tools/toolchain/install_cp2k_toolchain.sh#L647

Using `[[ "${dir}" != /* ]]` instead of `[ ... ]` is necessary here because `/*` is a Bash pattern and may be expanded by the shell with `[ ... ]`.

Growl1234:
I also changed `tail_excerpt` to make it exit with 1 after printing the log excerpt, so failed sub-builds are not silently ignored. This is important for installation scripts that may process multiple bundled components, such as COSMA, or multiple build variants, such as ELPA and DBCSR when GPU acceleration is enabled.

The OpenBLAS native-target build is the only intentional non-fatal case I found, because it falls back to the dynamic-arch build, so that call now prints the log directly instead of using `tail_excerpt`.

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -1404,30 +1434,190 @@ for ii in ${package_list}; do
...
+  printf '\n'
+}
+
+# ------------------------------------------------------------------------
+# Build packages unless dry-run mode is enabled.
+# ------------------------------------------------------------------------
+if [ "${dry_run}" = "__TRUE__" ]; then
+  print_toolchain_summary
+  printf "With --dry-run option, this script concludes with above report.\n"
+  printf "The setup, toolchain env and conf files are written to ./install.\n"
```
he-zilong: Is this meant to be `[...] written to ${INSTALLDIR}`?

he-zilong (commented):
Now this looks better. Let's hear what others say about this.

tools/toolchain/install_cp2k_toolchain.sh
Growl1234: Yes. This is adjusted in the next commit. Thank you for pointing it out.

tools/toolchain/scripts/tool_kit.sh
```diff
@@ -39,12 +39,12 @@ report_error() {
     local __message="$1"
   fi
   echo "ERROR: (${SCRIPT_NAME}${__lineno}) $__message" >&2
+  exit 1
```
oschuett, line 42: This breaks constructs like the following:

```
 if ! download_pkg_from_cp2k_org "${libxs_sha256}" "libxs-${libxs_ver}.tar.gz" 2> /dev/null; then
          download_pkg_from_urlpath "${libxs_sha256}" "${libxs_ver}" \
```

oschuett (commented):
>  I personally stubbornly prefer the old POSIX-compatible style and do not see bash-only syntax as convenient...

Maybe we should just rewrite the whole thing in Python?

tools/toolchain/scripts/tool_kit.sh
Growl1234: Sorry I’m currently away from my computer and cannot handle this right now. Maybe for now just directly use `download_pkg_from_urlpath`? I believe this construct should finally be replaced by `retrieve_package` once the release of libxs* and dbcsr is present.

he-zilong:
> > I personally stubbornly prefer the old POSIX-compatible style and do not see bash-only syntax as convenient...
> 
> Maybe we should just rewrite the whole thing in Python?

And then, once again, whoever porting toolchain to Python gets to realize sooner or later, "isn't this slipping into the spack-with-package-manager territory just with different download sources and dependency resolution?" Given the success of `make_cp2k.sh` (which nonetheless has utilized much more non-POSIX bash features than toolchain), I would say a shell-based interactive script still has its merit considering that the bash environment and interpreter is likely more consistent across a variety of systems and machines than [python](https://xkcd.com/1987/)...

oschuett:
> I would say a shell-based interactive script still has its merit considering that the bash environment and interpreter is likely more consistent across a variety of systems and machines than [python](https://xkcd.com/1987/)...

That XKCD actually makes a very good point. And indeed even TOP-10 HPC systems still run Python 3.6 by default.

Growl1234:
Looks like this breaks another construction...
https://github.com/cp2k/cp2k/blob/4d701726f9b1d78eef2c3b2a7eca32c49a9f2371/tools/toolchain/scripts/stage2/install_openblas.sh#L103-L107

Growl1234:
> Looks like this breaks another construction...

... and maybe changing `exit 1` in `report_error` to `return 1` helps?

PR #5400, opened 2026-06-15
Improve CP2K CMake package metadata exports

Growl1234:
Export `CP2K_USE_*` feature metadata unconditionally in `cp2kConfig.cmake` and change optional dependency handling from configure-time conditional blocks such as:
  ```cmake
  if(@CP2K_USE_DFTD4@)
    set(CP2K_USE_DFTD4 @CP2K_USE_DFTD4@)
    find_dependency(dftd4 REQUIRED)
  endif()
  ```
  to:
  ```cmake
  set(CP2K_USE_DFTD4 @CP2K_USE_DFTD4@)
  if(CP2K_USE_DFTD4)
    find_dependency(dftd4 REQUIRED)
  endif()
  ```
Slight clean up:
- Consolidate a few closely related dependency blocks e.g. ScaLAPACK requires MPI, libXSMM as an optional dependency of LibXS but not as a standalone dependency of CP2K
- Compiler ID fallback check in the exported config file.


mtaillefumier (approved):
This can be merged.

Issue #721, opened 2020-01-09
Inconsistencies when reading reference trajectories in MD


abussy:
I have been made aware of problems regarding the parsing of reference trajectories in `MOTION%MD%REFTRAJ`. I had a look hoping for a quick fix but the issue is actually quite complex.

In `*.xyz` files, the second line of each frame is reserved for a description. In principle anything could be written there, in any format. As we can see at line 1538 of `motion/integrator.F` (development version of CP2K), the program expects a well defined format, `FMT="(T6,I8,T23,F12.3,T41,F20.10)"`, from which it extracts frame index, time and potential energy. If this returns an error, then the frame index is read from the `md_env` and time and energy are set to zero.

This may fail in different ways:
- If the frame description starts with `#` or `!` the line is simply ignored and the first atomic position is interpreted as the description instead (general property of `cp_parser_type`)
- Because of `T6` in the expected format, any description of 5 characters or less will not cause an error at reading time. Hence CP2K interprets this as a frame with index, time and energy zero
-  The `STRIDE` keyword combined with non CP2K-format descriptions results in frames indices with increments of one, and not the value of stride

This can be problematic for users analyzing trajectories coming from different programs or homemade scripts.

I am not sure how this should be tackled. Issuing a warning in the manual could be a start.

I played around with `tests/QS/regtest-sparsity/Ar-ref-1.inp` to produce these errors.

dev-zero:
I expect a lot more and similar problems in all our file parsers since they are usually not unittested and we have almost no negative tests (tests which are supposed to fail).

One way to tackle this systematically would be to factor out facility routines (like parsers) into well designed objects and start unittesting them (similar to my `parallel_rng_types` refactoring).
This probably requires untying the parsing functions from our internal objects since those are usually difficult to setup without the rest of the CP2K environment/facilities.

For parsers reading large data blocks it might be beneficial to use an "emitting parser" scheme (for example call a user-supplied callback after successfully parsing a single MD frame and/or coordinate lines) since Fortran is missing [coroutines](https://j3-fortran.org/doc/year/19/19-169.pdf), to avoid allocating big chunks of memory just to copy them between structures.

PR #5401, opened 2026-06-15
Reftraj: parse extxyz for cell and deprecate `EVAL_ENERGY_FORCES`/`EVAL_FORCES`

he-zilong:
This PR introduces a new functionality and a breaking change regarding the REFTRAJ-type MD run, which is intended to be included in the upcoming 2026.2 release.

- [x] As discussed in #4962, the extended XYZ has been parsed for the initial cell in `&FORCE_EVAL/&SUBSYS/&CELL` but not the per-frame (possibly varying) cell in `&MOTION/&MD/&REFTRAJ`. This PR plans to unify the parsing logic of comment line for cell information as well as step, time and energy values.
- [x] The legacy keywords `EVAL_ENERGY_FORCES`/`EVAL_FORCES` are finally removed in favor of the `EVAL` setting after more than 2 years. It is safe to assume that the deprecation message has done its job in notifying users to revise the ancient input files for execution in newer CP2K versions.


he-zilong:
Finally, the extended XYZ parser is available for reftraj runs! I have made the logic more lenient regarding the format of comment line compared with the old implementation mentioned in #721.

A revision to the manual would follow afterwards.

he-zilong:
Taking the opportunity to elaborate on the REFTRAJ functionality, I have made minor revisions to some of my other past documentations as well. 

The new subroutine `read_xyz_comment()` has been put in the `src/cell_methods.F` so that `src/motion/integrator.F` does not pick up additional dependency; it can parse step, time and energy information in addition to cell specification, although those values are optional and are not requested when setting up the `&FORCE_EVAL/&SUBSYS/&CELL` by XYZ file. The full support for extended XYZ with extra columns beyond `Properties=species:S:1:pos:R:3` in the future would be better put here:
https://github.com/cp2k/cp2k/blob/561f47539b1bf1b38497acb340ae8cf50844f843/src/topology_xyz.F#L107-L108

This PR should retain the support for existing XYZ comments as generated here:
https://github.com/cp2k/cp2k/blob/561f47539b1bf1b38497acb340ae8cf50844f843/src/motion_utils.F#L412-L418
https://github.com/cp2k/cp2k/blob/561f47539b1bf1b38497acb340ae8cf50844f843/src/motion/dumpdcd.F#L873-L895

PR #5406, opened 2026-06-17
OFFLOAD: Restore internal host mempool for LIBXS builds

Growl1234:
This PR restores CP2K's internal host mempool for regular LIBXS builds. LIBXS is still used for SMM/JIT dispatch, and the LIBXSTREAM-specific allocation path is unchanged.

Using LIBXS for SMM/JIT should not implicitly make its default pool responsible for CP2K's generic host allocations. Those allocations are expected to follow CP2K's own mempool lifecycle, including cleanup between calculations. This becomes visible with `--keepalive`, where several calculations are executed in the same `cp2k --shell` process: the original code reproduces the DBM allocation failure, while restoring the internal host mempool makes the same tests pass.

### DIFF
--- a/src/offload/offload_mempool.c
+++ b/src/offload/offload_mempool.c
@@ -55,27 +55,19 @@ typedef struct offload_mempool {
  * \brief Private pools for host and device memory.
  * \author Ole Schuett
  ******************************************************************************/
-#if !defined(__LIBXS)
 static offload_mempool_t mempool_host = {0};
-#endif
-#if !defined(__LIBXSTREAM)
 static offload_mempool_t mempool_device = {0};
-#endif
 
 /*******************************************************************************
  * \brief Private counters for statistics.
  * \author Hans Pabst
  ******************************************************************************/
-#if !defined(__LIBXS)
 static struct {
   uint64_t mallocs, mempeak;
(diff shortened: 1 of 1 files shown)


Growl1234:
/cp2kci spack-psmp-gcc15 spack-ssmp spack-ssmp-static spack-openmpi-psmp

Growl1234:
Regarding this issue, there is also another possibility, namely that the problem lies on the libxs side...

Growl1234:
I've attempted some fixes on the LIBXS side; however, the CP2K-side problem is still present. The DBM tests consistently fail only when run with `--keepalive`, while all of them pass when each test is executed in a fresh process.

This suggests that the remaining issue is caused by CP2K retaining the process-wide LIBXS default pool across calculations. Therefore, I still propose restoring CP2K's own host memory pool for the non-LIBXSTREAM path. The LIBXS-side fixes should still be useful independently, but they do not address this lifecycle mismatch.

oschuett:
> Using LIBXS for SMM/JIT should not implicitly make its default pool responsible for CP2K's generic host allocations.

I agree that this is surprising. I think, instead we should try to improve the memory pool implemented in CP2K.

PR #5407, opened 2026-06-18
add new Pulay density mixing for grand canonical calculations

ZiweiChai:
Allow mixing the G=0 density components in Pulay mixing scheme to support grand canonical SCF optimizations.
Generalized the Kerker preconditioner and Pulay metric to improve the convergence behaviors of Pulay mixing for metallic systems.
Added input parameters for the new Pulay mixing.


Growl1234 (commented):
My main concern: CP2K stores the Kerker-preconditioned residual $Kr$ in the Pulay history. Applying the new metric $M$ to that history therefore gives an effective overlap $\langle r_i|KMK|r_j\rangle$, which may strongly underweight the $G=0$ and low- $G$ components the scheme is meant to control. I think the clarification regarding the derivation for CP2K’s residual definition is needed.

The spin treatment is also unclear: the new Kerker parameters affect only the charge channel, while the same metric is applied to both charge and magnetization residuals. This should be documented and covered by grand-canonical, metallic, and spin-polarized tests. Have you noticed the `ALPHA_MAG` and `BETA_MAG` before?

ZiweiChai:
> My main concern: CP2K stores the Kerker-preconditioned residual K r in the Pulay history. Applying the new metric M to that history therefore gives an effective overlap ⟨ r i | K M K | r j ⟩ , which may strongly underweight the G = 0 and low- G components the scheme is meant to control. I think the clarification regarding the derivation for CP2K’s residual definition is needed.
> 
> The spin treatment is also unclear: the new Kerker parameters affect only the charge channel, while the same metric is applied to both charge and magnetization residuals. This should be documented and covered by grand-canonical, metallic, and spin-polarized tests. Have you noticed the `ALPHA_MAG` and `BETA_MAG` before?

Many thanks for the comments. We didn't modify the old algorithm structure of how Pulay mixing is performed in CP2K, but only changed the way of calculating mixing_store%kerker_factor(:) and mixing_store%special_metric(:). I'm not sure whether there is any practical purpose in using ⟨ri|KMK|rj⟩ instead of ⟨ri|M|rj⟩ in constructing the Pulay matrix, or how much better the former can be than the latter. Conceptually, the latter seems to be the more reasonable choice. But in the current implementation, the tests showed that the convergence behavior for metallic slabs was really good with proper parameters. Concerning the spin treatment, I may have overlooked this, or there may have been other developments that I did not follow closely, since the current code was transferred from an old implementation. I will try to fix this issue in the next steps.

Growl1234:
Thanks very much for the clarification! Let's look forward to how this approach performs across a broader range of practical applications :-)

src/qs_density_mixing_types.F
```diff
@@ -542,18 +551,19 @@ SUBROUTINE create_mixing_section(section, ls_scf)
...
+                                       multisec_mix, new_pulay_mix], &
                           enum_desc=s2a("No mixing is applied", &
                                         "Direct mixing of new and old density matrices", &
                                         "Mixing of the potential in reciprocal space using the Kerker damping", &
                                         "Pulay mixing", &
                                         "Original CP2K Broyden mixing with a constant BROY_W0 regularization", &
                                         "Modified Broyden mixing with dynamic residual weights controlled by BROY_W0, "// &
                                         "BROY_WREF, and BROY_WMAX", &
-                                        "Multisecant scheme for mixing"))
+                                        "Multisecant scheme for mixing", "New Pulay mixing"))
```
oschuett: It would be good if there were some explanation about what the `NEW_PULAY_MIXING` exactly is and how it compares the the good ol' `PULAY_MIXING`. Maybe you could elaborate the input description or add something to the [method section](https://github.com/cp2k/cp2k/tree/master/docs/methods) of the manual?

oschuett (commented):
Generally LGTM.

Since this is a new method, could you please add tests?

ZiweiChai:
> Generally LGTM.
> 
> Since this is a new method, could you please add tests?

Thanks for pointing this out. I have added a regression test under QS/regtest-new-pulay-mixing and registered it in TEST_FILES.toml. I tested the input manually and used the resulting total energy as the reference value.

src/qs_density_mixing_types.F
ZiweiChai: Many thanks. The following changes has been submitted:
A description of the new Pualy mixing functionality in qs_density_mixing_types.F
The citation of the new Pulay mixing method in qs_mixing_utils.F

PR #5409, opened 2026-06-18
Docker [Spack]: Default to system GCC and enable Rawhide ssmp testing

Growl1234:
- Let the Fedora and Rawhide Spack testers use the distribution-provided GCC instead of pinning versions that may become outdated as the base images advance. The Spack Fedora test fails because cuurent system `glibc-2.43` has incompatibility with `libgomp` in `gcc-15` while Fedora 44 uses `gcc-16`.
- Add the existing Rawhide ssmp Spack tester to the dashboard.
- Report error if cache Python packages installation failed. This probably helps locating why Rawhide test fails.

### DIFF
--- a/tools/docker/generate_dockerfiles.py
+++ b/tools/docker/generate_dockerfiles.py
@@ -95,7 +95,6 @@ def main() -> None:
                 version="ssmp",
                 mpi_mode="no",
                 base_image="fedora:rawhide",
-                gcc_version=None,
                 testopts=testopts,
                 image_tag=f.image_tag,
             )
@@ -107,7 +106,6 @@ def main() -> None:
                 version="psmp",
                 mpi_mode="mpich",
                 base_image="fedora:rawhide",
-                gcc_version=None,
                 feature_flags="-ef openpmd",
                 testopts=testopts,
                 image_tag=f.image_tag,
@@ -120,7 +118,6 @@ def main() -> None:
                 version="psmp",
                 mpi_mode="mpich",
                 base_image="fedora:latest",
-                gcc_version=None,
                 feature_flags="-ef openpmd",
                 testopts=testopts,
                 image_tag=f.image_tag,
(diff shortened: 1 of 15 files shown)


Growl1234:
/cp2kci spack-psmp-fedora spack-psmp-rawhide spack-ssmp-rawhide

Growl1234:
It seems that `cffi` must be built from source on Fedora Rawhide because no compatible wheel is available for Python 3.15. Therefore, `libffi-devel` is installed to provide the `ffi.h` header.

tools/docker/generate_dockerfiles.py
```diff
@@ -701,7 +701,7 @@ def install_cp2k_spack(
     version: str,
     mpi_mode: str,
     base_image: str = "ubuntu:26.04",
-    gcc_version: int = 14,
+    gcc_version: int | None = 14,
```
oschuett: A more common API would be the following:
```suggestion
    gcc_version: int | None = None,
```

Growl1234: Thanks for suggestions, that makes sense. However, I need to double-check in case something like `g++-None` is generated.

Growl1234:
There appears to be an existing startup-path bug rather than a Spack or GCC issue. The `sdbg`/`pdbg` image invokes the binary as `cp2k`, whose short name exposes an out-of-bounds substring check that is caught by the Debug runtime checks; the usual versioned executable names and Release builds do not expose it.

oschuett:
> There appears to be an existing startup-path bug rather than a Spack or GCC issue. 

This might be the same the bug described in https://github.com/conda-forge/cp2k-feedstock/pull/77.

mkrack:
> > There appears to be an existing startup-path bug rather than a Spack or GCC issue.
> 
> This might be the same the bug described in [conda-forge/cp2k-feedstock#77](https://github.com/conda-forge/cp2k-feedstock/pull/77).

The crash for `cp2k -v` occurs since conda-forge updated to libgfortran-ng 15.2. Though the packages are still built wth gcc 14.3, the gcc 15.2 runtime environment is now packaged. The 15.2 runtime enviroment is supposed to be fully compatible with the version 14, but it is possibly more strict. The string returned by `compiler_options()` is very long and assigned to an allocatable `CHARACTER` variable first which is then cut into chunks for printout. That `CHARACTER` variable is never allocated explicitly and it must be somehow implicitly allocated by the `compiler_options()` function. Not sure if that is robust Fortran. This issue should be fixed for the next release.

Growl1234:
> The 15.2 runtime enviroment ... is possibly more strict.

I think this is the point...

oschuett:
> The 15.2 runtime enviroment ... is possibly more strict.

So is there anything we can do from our side that would merit opening a tracking issue?

Growl1234:
After addressing `openPMD_get_default_extension()`, the same pattern was found in `openPMD_json_merge()` as well... Though it should be fixed in this follow-up commit(?), it may be worth having a look at the remaining openPMD bindings more thoroughly. GCC 15 also appears to be noticeably stricter about language conformance: the known fixes already extend beyond CP2K itself to issues in dependencies such as COSMA and SIRIUS...

> So is there anything we can do from our side that would merit opening a tracking issue?

For now, the Rawhide test, with its aggressive rolling updates, already provides valuable coverage, but it was still insufficient to expose this issue earlier, for example, it does not have Debug building test.

PR #5416, opened 2026-06-20
Grid: Fix CUDA build

oschuett:
I still don't quite understand why this ever worked and why it then stopped working - apparently after #5343 🤷


mtaillefumier (approved):
Yes. it would certainly fix the atomicAdd definition problem for anything below pascal. Please merge this change and I will rebase afterwards.

Growl1234:
I think this is likely a long-standing native-CUDA / pre-Pascal compatibility gap. #5343 or one another PR may only have changed some aspect of the build environment and dependency combination, thereby exposing it for now.

PR #5417, opened 2026-06-20
offload: Complete removal of libxstream from mempool

oschuett:
This completes the removal of libxstream from the mempool that was started by #5406.

~~This rolls back all the changes in `src/offload/` from #5343.~~

### DIFF
--- a/src/offload/offload_library.c
+++ b/src/offload/offload_library.c
@@ -98,15 +98,14 @@ int offload_get_chosen_device(void) { return chosen_device_id; }
  * \author Ole Schuett
  ******************************************************************************/
 void offload_activate_chosen_device(void) {
-  if (chosen_device_id < 0) {
-    return;
-  }
 #if defined(__OFFLOAD_CUDA)
   OFFLOAD_CHECK(cudaSetDevice(chosen_device_id));
 #elif defined(__OFFLOAD_HIP)
   OFFLOAD_CHECK(hipSetDevice(chosen_device_id));
 #elif defined(__OFFLOAD_OPENCL)
-  OFFLOAD_CHECK(libxstream_device_set_active(chosen_device_id));
+  if (chosen_device_id >= 0) {
+    OFFLOAD_CHECK(libxstream_device_set_active(chosen_device_id));
+  }
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


Growl1234:
CC: @hfp

Maybe, it's worth adding an OpenCL tester in the future.

oschuett:
Yes, if we had tests for everything that would be a totally different story. I'm just really worried that one of the various changes to the memory management will break us on LUMI. I also contemplated de-risking the code, but I concluded that we'd just end up in an even more ill-defined stated than we are already due to #5406.

But maybe I'm overreacting here? @mtaillefumier, what do you think?

mtaillefumier:
I will run the regtests on AMD GPU for consistency.

mtaillefumier:
> Yes, if we had tests for everything that would be a totally different story. I'm just really worried that one of the various changes to the memory management will break us on LUMI. I also contemplated de-risking the code, but I concluded that we'd just end up in an even more ill-defined stated than we are already due to #5406.
> 
> But maybe I'm overreacting here? @mtaillefumier, what do you think?

That's a safe approach. Let me compile cp2k on AMD systems and run the regtests. Doing this should answer some question. AFAIK the gpu backend for collocate/integrate does not use memory pool at all internally (only the offload buffers may).

mtaillefumier:
No failure on mi300 platform when cp2k is compiled without libxs.

hfp:
I agree, if this fixes things for yet unknown reasons, it's safer to do so prior to release. It only motivates me to find the root cause.

hfp:
Do we have a list (not necessary complete) of CI-tests that failed with the original code?

oschuett:
> Do we have a list (not necessary complete) of CI-tests that failed with the original code?

The problem showed up when running the regtests with `--keepalive`. So it was probably an issue with re-initializing the libxstream. However, I'd really prefer to improve the mempool implementation in CP2K instead of maintaining two code paths.

hfp:
> > Do we have a list (not necessary complete) of CI-tests that failed with the original code?
> 
> The problem showed up when running the regtests with `--keepalive`. So it was probably an issue with re-initializing the libxstream. However, I'd really prefer to improve the mempool implementation in CP2K instead of maintaining two code paths.

Yes, the idea was to use libxstream as a vehicle and (selectively) adopt changes going forward.

oschuett:
> Yes, the idea was to use libxstream as a vehicle and (selectively) adopt changes going forward.

I think, for experimentation a separate git branch would be more appropriate. Otherwise we a) have to maintain two code paths and b) your libraries start to feature creep beyond their original scope:

- libxsmm + libxs = multiply small matrices
- libxstream = OpenCL runtime

PR #5418, opened 2026-06-21
Fix MKL target selection for Intel-threaded BLAS

Growl1234:
PR #5343 made `intel-thread` the default BLAS threading mode for Intel compilers. However, FindMKL only creates `seq`, `omp`, and `tbb` target variants, while `intel-thread` was incorrectly mapped to the non-existent `intel` suffix. This PR maps all OpenMP-backed MKL modes, including `intel-thread`, to the existing `omp` target variant.


Growl1234:
The regtest errors in Intel tester is expected to me, though I don't know why they are not working now...

PR #5419, opened 2026-06-21
Use pdbg for Rawhide tester

Growl1234:
Switch Rawhide testers to `pdbg` so more problems regarding newest GCC could be exposed and tracked. All errors on Rawhide testers could be expected failure.


mkrack:
> * Rename `spack-pamp-gcc15` to `spack-psmp` in consistent with existing `spack-ssmp`, `spack-ssmp-static`, `spack-openmpi-psmp` etc. because GCC-15 is the default compiler of Ubuntu 26.04.

I suggest keeping the GCC version in the tester name. I recently added it to the `spack_psmp` tester on purpose after the Ubuntu 24.04 to 26.04 update. Because the default GCC version changes with each Ubuntu release, explicitly including it prevents it from becoming a moving target.

Growl1234:
With Rawhide debug testers, many warnings are treated as errors, such as `-Werror=deprecated-openmp` and `-Werror=character-truncation`. Moreover `liblsan` needs to be installed seperately through `dnf`.

cmake/CompilerConfiguration.cmake
```diff
@@ -82,19 +82,27 @@ add_compile_options(
 )
 
 # Debug
+
+# TODO: Replace deprecated MASTER constructs with MASKED (only available since
```
oschuett, line 86: Should we open tracking issues for this TODO and the FIXME below?

Growl1234: Opened: #5426

Issue #5422, opened 2026-06-22
Incompatibility of pre-built libtorch with MKL causes runtime failure


Growl1234:
All GauXC regtests currently fail in my MPI build.

The failure is not a reference mismatch: the jobs crash during the initial SCF setup, in the ScaLAPACK `pdpotrf` call from `cp_fm_cholesky.F`. This also affects the `*_REFERENCE` cases, so it does not appear to be tied to one particular GauXC feature or input.

The common part of the backtrace is:

```text
__cp_fm_cholesky_MOD_cp_fm_cholesky_decompose
  at src/fm/cp_fm_cholesky.F:108
  -> pdpotrf(...)

init_scf_run
qs_scf_env_initialize
scf
qs_energies
force_env_calc_energy_force
```

and the MPI job eventually terminates with signal 11.

The build uses GauXC with MPI, OpenMP, host backend, and OneDFT enabled. Since the failure is in a general ScaLAPACK path rather than inside a GauXC routine, it is not clear yet whether this comes from GauXC itself, one of its dependencies, or the resulting link configuration.

I am attaching the full regtest log and toolchain scripts used for the build. I build CP2K on my Rocky 10.2 system, with `module load intel-onemkl-2026.0.0  mpich-5.0.1`, `./install_cp2k_toolchain.sh --with-cmake=system --with-gauxc -j24` and `./build_cp2k.sh -j 24`.

[regtest.log](https://github.com/user-attachments/files/29191095/regtest.log)
[regtest-gauxc.zip](https://github.com/user-attachments/files/29191103/regtest-gauxc.zip)

Growl1234:
With FFTW enabled (sorry, my fault not writing `math_mode` to `toolchain.conf`), it shows another error:
```text
Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x14d41c423d05 in ???
#1  0x14d41c422d65 in ???
#2  0x14d4148681bf in ???
#3  0x14d41ced49d5 in ???
#4  0x14d41b8b3d74 in ???
#5  0x14d41b8bc625 in ???
#6  0x14d41b8bbaeb in ???
#7  0x14d41b8bbc01 in ???
#8  0x14d421853452 in __fftw3_lib_MOD_fftw3_create_plan_3d
	at /opt/test/cp2k-dev/src/pw/fft/fftw3_lib.F:715
[...]

PR #5424, opened 2026-06-22
Toolchain: Block installation of pre-built libtorch for MKL path

Growl1234:
Short-term workaround for #5422.


Growl1234:
The Intel tests failed after this PR was merged because `--with-gauxc=no` is not added after `--install-all` and thus triggers libtorch installation even if `--with-libtorch=no` is passed.

Issue #5377, opened 2026-06-08
RFC: Code Owners


oschuett:
I'm starting to think that, we should formally introduce [code owners](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners).

Code owners are teams of people that possess a good mental model of their part of the code base and feel responsible for keeping it healthy. **To that end, they get to review every change to their code.**

In practice, we always had code owners, but the informal system broke down as the dev team grew. Nowadays, owners are forced to closely watch every PR that flies by, which is very stressful.

Of course, adding bureaucracy carries the risk of slowing down our development pace. However, I expect that people will mostly stay in their lanes and only few PRs need someone else's review. On the plus side, we'll probably catch some architectural mistakes earlier on.

What do you think?

mtaillefumier:
Thanks for starting the discussion. That's a very good first step. My main areas are 

- grid backend (with you)
- cmake build system
- dependencies handling
- spack related issues
- sirius

My contributions to the scientific part will be limited because most contributions are outside my area of scientific expertise.

- One action should be put in place very soon because recent events have shown that we can not continue with the current model. **No developer should be allowed to merge their own PR without the explicit approval of at least one, and preferably two, other people.**.
- The status of the `master/develop/main` branch should be stated explicitly as master is a not dumpster of ideas, unfinished or experimental work that breaks user experience. Any PR breaking the build system or any fundamental parts of the code should only be merged after proper approval and testing.
- If a develop branch is needed then we should create it, commit new changes to develop, and only merge into master after proper testing. Documentation and simple bug fixes may be excluded from develop, provided they receive appropriate review.
- The CI/CD covers many conventional devs mistakes (my own included) and I think we should tighten this even more.
- These Areas need extra attention. They should include the build system, spack recipe, CI/CD, toolchain. Any PR touching them without review should be explicitly rejected.

Anyone reviewing scientific papers or research proposals understand these simple rules and most if not 99.9% of software projects follow similar standards. 

Any change that breaks user experience or existing workflows without prior warning and documentation should be subject to automatic revert. The burden is on the author to communicate breaking changes up front and ensure they are properly reviewed and tested.

The goal of these rules is to reduce stress on everyone and avoid introducing breaking changes without explicit warning and testing because some changes can have severe consequences for the entire community. We should learn from the transition to `CMake` that took many years of testing and avoid repeating the inconveniences for the past days.

dryndyk:
[...]

PR #5431, opened 2026-06-22
Add CODEOWNERS

oschuett:
Following our discussion in #5377, I've put together an initial CODEOWNERS file.

I also did some spring cleaning and removed write access from folks who haven't contributed in over a year.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,69 @@
+# Top-level files
+/*                           @cp2k/cp2k-admins
+
+# Data
+/data/                       @mkrack @juerghutter
+
+# CMake
+/cmake/                      @mkrack @mtaillefumier @RMeli
+CMakeLists.txt               @mkrack @mtaillefumier @RMeli
+
+# Spack
+/tools/spack/                @mkrack @mtaillefumier @RMeli
+
+# Toolchain
+/tools/toolchain/            @mkrack @oschuett
+
+# Testing + Conventions
+/tools/conventions/          @mkrack @oschuett
+/tools/precommit/            @mkrack @oschuett
+/tools/dashboard/            @mkrack @oschuett
+/tools/docker/               @mkrack @oschuett
+
+# Load bearing APIs and files
+/src/fm/                     @fstein93 @oschuett
+/src/pw/                     @fstein93 @oschuett
+/src/pw_env/                 @fstein93 @oschuett
+/src/mpiwrap/                @fstein93 @oschuett
+/src/dbm/                    @fstein93 @oschuett
+/src/dbt/                    @fstein93 @oschuett
+/src/dbx/                    @fstein93 @oschuett
+/src/base/                   @fstein93 @oschuett
+/src/start/                  @fstein93 @oschuett
+/src/f77_interface.F         @fstein93 @oschuett
+/tests/do_regtest.py         @fstein93 @oschuett
+/tests/matcher_classes.py    @fstein93 @oschuett
+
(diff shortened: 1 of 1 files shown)


Growl1234:
Thanks for adding this. I have two comments on the current mapping.

1. The entries for `/fm/`, `/pw/`, `/pw_env/`, and `/mpiwrap/` are rooted at the repository top level, whereas the corresponding directories are under `src/`. As written, these patterns therefore do not match the intended files. They should presumably be `/src/fm/`, `/src/pw/`, `/src/pw_env/`, and `/src/mpiwrap/`.

2. The current mapping also leaves a substantial number of files without an assigned owner. In particular, `/*` only matches paths directly below the repository root; it is not a recursive fallback for the rest of the repository. Therefore, much of `src/` outside the explicitly listed patterns still has no "code owner".

Is this intended to be a deliberately partial initial mapping? Otherwise, it may be worth considering either additional entries for the remaining areas or a fallback rule, with the more specific rules below overriding it.

oschuett:
> 1. ...They should presumably be /src/fm/, /src/pw/, /src/pw_env/, and /src/mpiwrap/.

Good catch, I fixed them.

> 2. ... Is this intended to be a deliberately partial initial mapping?

Yes, most of the code-base is up for grabs because the original authors have left the project.

CODEOWNERS
```diff
@@ -0,0 +1,53 @@
+# Top-level files
+/*                           @cp2k/cp2k-admins
+
+# Data
+/data/                       @mkrack @juerghutter
+
+# CMake + Spack
+/cmake/                      @mkrack @mtaillefumier
```
RMeli: Add `CMakeLists.txt` as well? I'm happy to keep an eye on CMake and Spack, if needed.

CODEOWNERS
```diff
@@ -0,0 +1,53 @@
...
+/tools/toolchain/            @mkrack @oschuett
+
+# Testing + Conventions
+/tools/conventions/          @mkrack @oschuett
+/tools/precommit/            @mkrack @oschuett
+/tools/dashboard/            @mkrack @oschuett
+/tools/docker/               @mkrack @oschuett
+
+# Load bearing APIs and files
+/src/fm/                     @fstein93 @oschuett
```
RMeli, line 24:
```suggestion
/src/fm/                     @fstein93 @oschuett @RMeli
```
To keep an eye on DLA-Future.

fstein93:
Can you add me to `src/mp2*` and together with @JWilhelm `src/rpa_*`? @JWilhelm should be also responsible for `src/gw_*`.

CODEOWNERS
oschuett: I added a separate rule for DLA-Future.

RMeli: Thanks. I wasn't sure which one was best, because DLAF integration touches files in `fm/` that are not only DLAF-specific. But let's go with that, and I can change it to be broader in the future.

CODEOWNERS
```diff
@@ -0,0 +1,67 @@
+# Top-level files
+/*                           @cp2k/cp2k-admins
+
+# Data
+/data/                       @mkrack @juerghutter
+
+# CMake
+/cmake/                      @mkrack @mtaillefumier @RMeli
+/CMakeLists.txt              @mkrack @mtaillefumier @RMeli
```
RMeli:
```suggestion
CMakeLists.txt              @mkrack @mtaillefumier @RMeli
```
? This should match all of them.

PR #5434, opened 2026-06-23
benchmarks: Add IGNORE_CONVERGENCE_FAILURE to w64PBE.inp

oschuett:
...to restore the [CUDA performance test](https://dashboard.cp2k.org/archive/perf-cuda-volta/index.html).

### DIFF
--- a/benchmarks/QS_reference/w64PBE.inp
+++ b/benchmarks/QS_reference/w64PBE.inp
@@ -31,6 +31,7 @@
     &END QS
     &SCF
       EPS_SCF 1.E-7
+      IGNORE_CONVERGENCE_FAILURE
       MAX_SCF 10
       SCF_GUESS ATOMIC
       &OT ON


juerghutter:
According to the reference output, there is no convergence problem with this run.
Why should a CUDA or any other hardware setup behave differently?

oschuett:
Ok, then this benchmark managed to find a bug in the GPU grid code that all the other regtests missed - scary. 

@mtaillefumier

juerghutter:
What is the signature of the convergence failure?

oschuett:
The outer loop exhausts its 10 iterations. You can download the full output from [here](https://storage.googleapis.com/cp2k-ci/dashboard_perf-cuda-volta_artifacts.zip).

mtaillefumier:
To be clear, there may or may not be a bug in the grid backend, and this benchmark does not answer that question at all. It only shows that the particular arithmetic here triggers an extra SCF iteration.

I also need to correct some assumptions that were made:

- Floating‑point arithmetic is non‑associative, so small differences in the order of operations at this level of convergence can absolutely lead to one or two additional iterations.
- Different hardware and backends can (and do) behave differently for exactly this reason, even when the high‑level algorithm and input are the same.

mtaillefumier:
I’ll need to examine the detailed SCF/OT logs (iteration history and residuals). The information that the outer loop exhausts its 10 iterations, by itself, is not sufficient to conclude there’s a bug rather than just slightly slower convergence.

juerghutter:
The problem is not one or two additional iterations. The convergence shows small differences up to ca. step 17 when
an accuracy of about 10^-5 is reached. The reference then goes on and converges to 10^-7 in additional 6 steps,
whereas the "CUDA" run starts to oscillate between 10^-5 and 10^-6 and never converges for 100 steps.
This is not necessarily due to the CUDA grid code. It could be the use of inconsistent screening in DBCSR or diagonalization.

mtaillefumier:
> The problem is not one or two additional iterations. The convergence shows small differences up to ca. step 17 when
> an accuracy of about 10^-5 is reached. The reference then goes on and converges to 10^-7 in additional 6 steps,
> whereas the "CUDA" run starts to oscillate between 10^-5 and 10^-6 and never converges for 100 steps.
> This is not necessarily due to the CUDA grid code. It could be the use of inconsistent screening in DBCSR or diagonalization.

Yes. That was the missing information. I do not think that hardware difference would introduce that much fluctuations though. 

Turning Validation option in grid section leads to this. 

```
Error: Validation failure in grid integrate
   hab diff:     4.665091e-12
   hab rel_diff: 4.665091e-12
   hab value:    3.147889e-04
   hab i:        216203
```

mtaillefumier:
there is a race condition somewhere in the coefficients calculations.

Growl1234:
Looks like not only one example encounters the convergence issue... surprised of that.

mtaillefumier:
I can think of two potential problems:
- A race condition in the hab calculations / forces
- Or an wrong offset that add or remove quantities.
The problem is likely located in the integrate part of the code. The collocate part works ok according to the validation mode. 

The other curious behavior is that the validation option triggers an error for H2O-32.inp only when calculating the forces while it agrees during the SCF steps. So in one case `w64PBE.inp` the validation test fails during the SCF while in the other case it fails during the forces calculations.  

To illustrate the issue

```
Validated grid collocate, max rel. diff: 1.665335e-15
Validated grid collocate, max rel. diff: 4.551914e-15
Validated grid collocate, max rel. diff: 4.551914e-15
Validated grid collocate, max rel. diff: 4.551914e-15

 *** WARNING in qs_scf.F:684 :: SCF run NOT converged ***

Error: Validation failure in grid integrate
   forces diff:     2.498088e-08
   forces rel_diff: 2.498088e-08
   forces value:    5.843177e-01
   forces atom:     2
   forces dir:      1

```

mtaillefumier:
I found the issue. It’s a race between two warps in the same thread block accessing and modifying the task information in shared memory. One warp can start updating the task while the other is still in the middle of computations that assume the task data is invariant. The bug is latent on AMD GPUs because warpSize = 64 (the whole block is a single warp), but it is triggered on NVIDIA GPUs where warpSize = 32 (the block spans two warps).

I have a patch ready for this. @oschuett do you want me to create PR on this branch so we can test the theory. I checked with w64PBE.inp and 

```
&GRID
VALIDATE TRUE
&END GRID
```

Total runtime without the check is 185.495 s on my v100.

oschuett:
> I found the issue.

Great! Then yes please open a PR with the fix.

mtaillefumier:
PR opened. https://github.com/cp2k/cp2k/pull/5437.

PR #5435, opened 2026-06-23
Revise k-points documentation

Growl1234:
@DCM-Uni-Paderborn Would you be object to removing `Wannier90` section from `k-points` page and finally adding documentation of Wannier90 in a new page?

### DIFF
--- a/docs/methods/dft/k-points.md
+++ b/docs/methods/dft/k-points.md
@@ -46,6 +46,36 @@ Complex wavefunctions are the default for k-point calculations; see
 for Gamma and special k-points whose Bloch phases can be represented as real. Use complex
 wavefunctions for a general mesh or for atomic k-point symmetry reduction.
 
+#### Feature compatibility
+
+Compatibility with k-point sampling depends on the selected CP2K feature and implementation path.
+The following table summarizes selected common cases for `DFT%KPOINTS`; it is not exhaustive. Refer
+to the documentation of the relevant feature for detailed requirements and limitations or see issue
+[#4854](https://github.com/cp2k/cp2k/issues/4854) which tracks the current status of k-points
+support.
+
+| Feature                       | Compatibility and limitations                                           |
+| ----------------------------- | ----------------------------------------------------------------------- |
+| Standard diagonalization      | **Supported.**                                                          |
+| Orbital transformation (OT)   | **Unsupported.** No k-point path available.                             |
+| Atomic symmetry reduction     | **Experimental.** Validate against an equivalent unreduced calculation. |
+| WFN extrapolation             | **Supported.**                                                          |
+| Other diagonalization methods | **Unsupported.** No k-point path available.                             |
+| Hybrid functionals            | **Supported via [RI-HFXk](hartree-fock/ri_kpoints).**                   |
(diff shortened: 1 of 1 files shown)


he-zilong:
This is related to both #4854 and #4972. I think a key point that users may want to know, but the documentation does not cover yet, is how to confirm whether some feature is compatible with k-point sampling or not (both on the theory/algorithm and program implementation side, and in practice where a supercell vs. k-point comparison may be necessary).

By the way, would it be a good idea to remove the very outdated www.cp2k.org/faq:kpoints entirely? @oschuett

oschuett:
> By the way, would it be a good idea to remove the very outdated [www.cp2k.org/faq:kpoints](http://www.cp2k.org/faq:kpoints) entirely? @oschuett

I've replaced the old wiki page with a bunch of links.

Growl1234:
> I think a key point that users may want to know, but the documentation does not cover yet, is how to confirm whether some feature is compatible with k-point sampling or not (both on the theory/algorithm and program implementation side, and in practice where a supercell vs. k-point comparison may be necessary).

I am still considering the best way to document this, since the scope is quite broad. Linking only to issue #4854 may also be insufficient, as it would not explain the underlying limitations or rationale.

Growl1234:
I added a first compact table for this. It is likely incomplete, especially for features I am less familiar with, so suggestions for important missing entries or corrections to the listed status would be welcome. Also linked with #4854.

I am still not sure whether maintaining a comprehensive compatibility list in this documentation page would be realistic; perhaps there is a better approach?

he-zilong:
The list misses the commonly requested periodic electric field (based on Berry-phase operator; the latest relevant development is #5241 IIRC). I do not have a better idea as to where the list of features should be put; it feels reasonable to have both a dedicated section of one-liner summary on one documentation page, and elaborations regarding specific features on separate pages.

garrettwx:
I think we could first put a small section that contains the current functions that support k-point version. And attached in its page is the link here. And by the way, I notice some important functions are missing in the main page, e.g. blue moon ensemble, slow growth method for AIMD.

DCM-Uni-Paderborn:
> @DCM-Uni-Paderborn Would you be object to removing `Wannier90` section from `k-points` page and finally adding documentation of Wannier90 in a new page?

Fine with me.

Growl1234:
Thanks, I agree that a compact one-line summary here, with detailed limitations on the respective feature pages, is a reasonable structure. However, I would prefer not to substantially broaden this PR. Several relevant features currently have no narrative documentation beyond the input reference, so a comprehensive and reliable compatibility matrix would require a separate documentation effort. I'd prefer keeping the table non-exhaustive for now; if you think there are other commonly requested k-point support missing in this table, feel free to let me know or directly adding them.

he-zilong (approved):
> I'd prefer keeping the table non-exhaustive for now; if you think there are other commonly requested k-point support missing in this table, feel free to let me know or directly adding them.

Agreed, let's wait and see what concerns other users may have. The list, along with the rest of the documentation, looks good to me now.

> I notice some important functions are missing in the main page, e.g. blue moon ensemble, slow growth method for AIMD.

@garrettwx this is the topic for #4972, not here.

PR #5436, opened 2026-06-23
Docker: Remove Spack profile from cmake_cp2k.sh

Growl1234:


mkrack:
Not sure, if this change has an impact on any spack tester, because `make_cp2k.sh` does not use `cmake_cp2k.sh`. Maybe, the if clauses
```
if [[ "${PROFILE}" == "spack" ]] && [[ "${VERSION}" == "psmp" ]]; then
```
and
```
elif [[ "${PROFILE}" == "spack" ]] && [[ "${VERSION}" == "ssmp" ]]; then
```
could be dropped?

Growl1234:
Thanks for reminder, I forgot that this path is not used anymore...

Growl1234:
So again the regtest failure due to GauXC slow tests...

PR #5437, opened 2026-06-23
Fix a race condition in the grid backend

mtaillefumier:
The kernel calculating the hab/forces/stress components has a race condition that allowed one warp to update the task information in shared memory while the other warp could potentially still do some calculation.

The bug is latent on AMD GPU (because only one warp runs) while it is triggered on NVidia GPU because two warps are needed.


mtaillefumier:
i think we need a H2O-32 test with GRID validation on to check "real" cases. We only need one or two SCF iterations and forces calculations.

Issue #5439, opened 2026-06-24
Poor performance of CPU MPI calculations with native SKALA GPW


Growl1234:
While testing the CPU MPI behavior of the experimental native SKALA GPW path with a modified regtest input `H2_NATIVE_SKALA_GPW_PBC_FORCE_DEBUG`, I encountered two related issues.

First, without `NATIVE_GRID_ATOM_CHUNKS`, increasing the MPI count from 2 to 8 ranks made a small H2 test slower rather than faster:

```text
MPI ranks: 2
CP2K                             6.954 s
qs_vxc_create                    6.649 s
torch_model_forward_mol_tensor   1.985 s
skala_gpw_backward               4.026 s

MPI ranks: 8
CP2K                             10.000 s
qs_vxc_create                    9.667 s
torch_model_forward_mol_tensor   3.061 s
skala_gpw_backward               5.880 s
```

The Torch forward and backward work dominates the calculation and becomes slower at 8 ranks. This appears consistent with the current default path, where dynamic features are collected into a global feature block and the Torch evaluation is then effectively performed on every rank, rather than being distributed across ranks.

Second, enabling

```text
NATIVE_GRID_ATOM_CHUNKS T
```

causes the same H2 calculation to hang with 8 MPI ranks.

For H2, only two non-empty atom chunks can be created, so six ranks receive no chunk. The chunk path then appears to be selected locally: ranks with a non-empty chunk enter the atom-chunk communication path, while ranks without one fall back to the global-feature path. This likely results in different MPI collectives being called on the same communicator, causing a deadlock.

The H2 case is intentionally too small to be a meaningful scaling benchmark. Nevertheless, ranks without assigned atom chunks should not make the calculation hang. The atom-chunk implementation should either support idle ranks while keeping all ranks on a consistent communication path, or terminate collectively with a clear error when the number of MPI ranks exceeds the number of available atom chunks.

The testing files are here:
- [test.zip](https://github.com/user-attachments/files/29279199/test.zip)
- [test-atomchunk.zip](https://github.com/user-attachments/files/29279245/test-atomchunk.zip)

I built CP2K with self-compiled PyTorch (compatible with MKL) and toolchain-installed GauXC.

Growl1234:
The second problem should be fixed by #5440.

DCM-Uni-Paderborn:
[...]

PR #5440, opened 2026-06-24
GauXC/SKALA: Handle empty atom chunks in MPI runs

Growl1234:
When MPI ranks outnumber available atom chunks, ranks without a local chunk fell back to the non-chunk path while the remaining ranks entered atom-chunk routing, resulting in mismatched MPI collectives and a hang. (#5439)

This PR keeps all ranks on the atom-chunk communication path instead. Ranks without a local chunk skip Torch tensor construction and model evaluation, but still participate in feature/gradient routing and reductions. It does not bring performance improvement for very small system such as H2, but avoids the hang.


oschuett:
@DCM-Uni-Paderborn what do you think?

Growl1234:
At least, the solution for regtests should not be delayed in my view, as it affects every CP2K regtest runs. I'm going to push them to a new PR (#5465) and rebase this one after #5465 is merged if this PR still doesn't get a review.

DCM-Uni-Paderborn (approved):
As far as I see looks ok

PR #5443, opened 2026-06-24
MAO output

juerghutter:
Output MAO in ML-PAO form for learning

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -363,6 +363,7 @@ list(
   mao_methods.F
   mao_optimizer.F
   mao_wfn_analysis.F
+  mao_io.F
   matrix_exp.F
   maxwell_solver_interface.F
   mdctrl_types.F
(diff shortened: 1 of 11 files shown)


oschuett:
Thanks a lot for building the MAO-PAO bridge.

This PR is also the first to trigger the new CODEOWNER mechanism. Since you are adding a new file, it's asking for a review from the CMake team. This is not what I indented. I'll look into this.

RMeli:
> Since you are adding a new file, it's asking for a review from the CMake team. This is not what I indented. I'll look into this.

@oschuett this works as expected. The `CMakesLists.txt` file has been modified, hence the trigger. Personally, I generally prefer a bit more noise than missing important things. In situations like this, I think it's easy to agree that there is no need at all to wait for approval from the "owners".

oschuett:
Adding a new file is pretty common and does not affect other developers. Hence, I don't think this should trigger a review. So, I've just opened #5444.

mtaillefumier (approved):
LTGM

juerghutter:
Any idea how this PR can be merged?

PR #5444, opened 2026-06-24
CODEOWNERS: Allow everyone to add new source files

oschuett:
While there is also some delicate logic in [src/CMakeLists.txt](https://github.com/cp2k/cp2k/blob/master/src/CMakeLists.txt), I'll leave it to the CMake team to migrate this as they see fit.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,7 +6,7 @@
 
 # CMake
 /cmake/                      @mkrack @mtaillefumier @RMeli
-CMakeLists.txt               @mkrack @mtaillefumier @RMeli
+/CMakeLists.txt              @mkrack @mtaillefumier @RMeli
 
 # Spack
 /tools/spack/                @mkrack @mtaillefumier @RMeli


Growl1234:
I feel that the lower part of `src/CMakeLists.txt` is not particularly suitable to be moved elsewhere, and the current structure already matches what `src/CMakeLists.txt` is expected to handle. Is there any possible workaround for this?

~Moreover, this should only be an automatic request for code review, and would not block the PR from being merged in principle.~

oschuett:
> Is there any possible workaround for this?

I know this is controversial, but we could replace the hard-coded list with a discovery mechanism (glob). The old Makefile did this and personally I really liked it.

> Moreover, this should only be an automatic request for code review, and would not block the PR from being merged in principle.

If the reviews were optional then we'd be back at a soft rule, which didn't work. Also, AFAIK Github doesn't support it.

Growl1234:
> we could replace the hard-coded list with a discovery mechanism (glob).

If there is discovery mechanism available in CMake, I'll be very happy about applying it. @mtaillefumier What do you think?

Growl1234:
One another way maybe to use `src/cp2k_source_lists.cmake`, although I personally don't quite like it.

PR #5445, opened 2026-06-24
Update GauXC to dev20260608

mkrack:
- Drop variants "onedft" and "gau2grid"
- Prepare merge with upstream spack recipe

### DIFF
--- a/tools/spack/spack_repo/cp2k_dev/packages/gauxc/package.py
+++ b/tools/spack/spack_repo/cp2k_dev/packages/gauxc/package.py
@@ -23,7 +23,7 @@ class Gauxc(CMakePackage, CudaPackage):
     license("BSD-3-Clause")
 
     version("master", branch="master")
-    version("dev20260522", branch="skala", commit="69ee054054c642bce0d9b5e8c9c0c13afa97b774")
+    version("dev20260608", branch="skala", commit="36aeab80bf7b2245087bdfeffd8e61b11206b3bd")
     version(
         "1.2.dev2",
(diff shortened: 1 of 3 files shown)


tools/spack/spack_repo/cp2k_dev/packages/gauxc/package.py
```diff
@@ -23,25 +23,25 @@ class Gauxc(CMakePackage, CudaPackage):
...
     version(
         "1.2.dev2",
         sha256="6659d00522b443f2557a960a9ba0217449d2ee25df809634d6becca23eb0f1ff",
         url="https://github.com/microsoft/skala/releases/download/v1.1.1/gauxc-skala-r2.tar.gz",
     )
     version("1.1", sha256="17de077fb23e44d03b0ed14dcd8117c01e5b3431fbefa2352d751639ada7f91c")
 
-    variant("c", default=True, description="Build with C API support")
-    variant("fortran", default=True, description="Build with Fortran support")
+    variant("c", default=True, description="Build with C API support", when="@master")
```
mtaillefumier, line 34: minor change. if "@version:" is used, `spack `should automatically include master or main in the supported versions.

mtaillefumier (approved):

tools/spack/spack_repo/cp2k_dev/packages/gauxc/package.py
mkrack: Thanks for the suggestion, but I would like to keep definitions explicit instead of relying on any automatism, because, unfortunately, things are quite fuzzy. Usually, `master` is the most recent code version which is not the case here. Hope that we can improve/resolve that in the future.

PR #5449, opened 2026-06-24
Integrated (absolute) spin density: correct output for GAPW

juerghutter:


juerghutter:
Is there any way I can still merge PRs?

oschuett:
> Is there any way I can still merge PRs?

I just added you to the cp2k-admins team, which allows you to overwrite these checks.

Of course, we still need a solution for the slow tests that have been introduced by #5421 and single-handedly increased our total regtest time by 50%.

@DCM-Uni-Paderborn

Growl1234:
> Of course, we still need a solution for the slow tests that have been introduced by #5421 and single-handedly increased our total regtest time by 50%.

I tried, but it's hard to resolve them by using lower-level basis sets or adjusting other SCF settings. It seems to come from the GauXC/SKALA implementation itself, and ironically, issue #5439 is exactly what I encountered when trying to deal with those tests...

Currently I add them to suppression list in #5440 and then the time spent is about 26min... Maybe this is because it skips the re-run step for slow tests?

oschuett:
Thanks for looking into this @Growl1234!

Besides adding them temporarily(!) to the suppression list, we should split `tests/QS/regtest-gauxc` into ~16 directories to take advantage of the available parallelism.

Growl1234:
> we should split tests/QS/regtest-gauxc into ~16 directories to take advantage of the available parallelism.

Agreed.

Update: I will do it along with #5440.

PR #5454, opened 2026-06-24
Add hint regarding missing FFTW3

fstein93:
As discussed at the develepers' meeting, FFTW3 may be turned into a hard dependency in the future (new FFT backend WIP). FFTW3 is precompiled on many clusters or directly available in the math libraries of hardware developers (MKL, AOCL, cuFFT). Already now, some features are not available without FFTW3 (or at least only in case of very large unit cells).
So, I add a warning if a user decides to compile CP2K with FFTW3 support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1358,7 +1358,9 @@ if(NOT CP2K_USE_VORI)
 endif()
 
 if(NOT CP2K_USE_FFTW3)
-  message("   - FFTW3")
+  message(
+    "   - FFTW3: FFTW3 may become mandatory in future releases of CP2K. Please consider installing and enabling it. Some features of CP2K may not work properly without FFTW3."
+  )
 endif()
 
 if(NOT CP2K_USE_LIBTORCH)


Growl1234:
Would it be better to use warning message?

he-zilong:
I suggest adding a note in the [libraries](https://manual.cp2k.org/trunk/technologies/libraries.html#fftw-improved-performance-of-ffts) documentation for how to check if an existing FFTW library is FFTW3. (For some reason, there are broken warnings with incorrect format; see[docs/README.md](https://github.com/cp2k/cp2k/blob/master/docs/README.md).)

Growl1234:
> I suggest adding a note in the [libraries](https://manual.cp2k.org/trunk/technologies/libraries.html#fftw-improved-performance-of-ffts) documentation for how to check if an existing FFTW library is FFTW3. (For some reason, there are broken warnings with incorrect format; see[docs/README.md](https://github.com/cp2k/cp2k/blob/master/docs/README.md).)

I'll revise this page in a seperate PR.

mtaillefumier (approved):

mtaillefumier:
From what I understand of the short discussion, the question may rather be do we impose fftw3 as mandatory now.

All tests pass, only four tests are slower than expected that's why the regtests are indicated as failed.

mkrack (approved):

oschuett:
I think, making FFTW a hard dependency, just like BLAS and ScaLAPACK, is the right move.

fstein93:
I just wanted to announce it to the users such that we can make FFTW mandatory with the next release.

Growl1234:
> I just wanted to announce it to the users such that we can make FFTW mandatory with the next release.

If FFTW is going to become a hard dependency in the future, this message should be made more prominent so that users are more likely to notice it.

PR #5456, opened 2026-06-25
toolchain/build_cp2k.sh: Write module environment to cp2k_env if exists

Growl1234:
For convenience of direct `source cp2k_env` and `module sh-to-mod`.


mkrack:
LGTM, but did the tests really check/validate the changes, because the `toolchain pdbg` test run was fullly cached?

Growl1234:
> LGTM, but did the tests really check/validate the changes, because the `toolchain pdbg` test run was fullly cached?

`build_cp2k.sh` is currently not covered by the CI tests. I considered integrating this script into CI before, but the required setup is relatively involved and may exceed the intended scope of this helper script, so that work was postponed. Still, I think it would be useful to improve the script so that it can eventually be used by at least a subset of CI builds that compile CP2K entirely from the toolchain environment.

mkrack (approved):

PR #5442, opened 2026-06-24
Toolchain: Move libFCI installation to stage7

Growl1234:

PR #5458, opened 2026-06-25
Sync CMake configuration to cp2kConfig.cmake

Growl1234:
Follow up to #5442.


oschuett:
Thanks for the quick follow up. It's really unfortunate, that we have to maintain the `cp2kConfig.cmake` manually...

mkrack (approved):

mtaillefumier (approved):

PR #5461, opened 2026-06-25
Add the hint regarding FFTW to the description of libraries

fstein93:
Follow-up to #5454 
@Growl1234 Is this what you had in mind?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1369,8 +1369,10 @@ if(NOT CP2K_USE_VORI)
 endif()
 
 if(NOT CP2K_USE_FFTW3)
+  message("   - FFTW3")
   message(
-    "   - FFTW3: FFTW3 may become mandatory in future releases of CP2K. Please consider installing and enabling it. Some features of CP2K may not work properly without FFTW3."
+    WARNING
+      "FFTW3 may become mandatory in future releases of CP2K. Calculations requiring large grid sizes may not work properly without FFTW3. Please consider installing and enabling it."
   )
 endif()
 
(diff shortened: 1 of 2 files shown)


oschuett:
While you're at it, could you also fix the other `:warning:` blocks in `libraries.md`?

Growl1234:
I think, what I'm referring to is at `CMakeLists.txt`:
```cmake
if(NOT CP2K_USE_FFTW3)
  message("   - FFTW3")
  message(
    WARNING
      "FFTW3 may become mandatory in future releases of CP2K. Please consider installing and enabling it. Some features of CP2K may not work properly without FFTW3."
  )
endif()
```

fstein93:
> I think, what I'm referring to is at `CMakeLists.txt`:
> 
> ```cmake
> if(NOT CP2K_USE_FFTW3)
>   message("   - FFTW3")
>   message(
>     WARNING
>       "FFTW3 may become mandatory in future releases of CP2K. Please consider installing and enabling it. Some features of CP2K may not work properly without FFTW3."
>   )
> endif()
> ```

Could you elaborate please?

Growl1234:
It's here:
https://github.com/cp2k/cp2k/blob/6ab90febadc0a35594f983fe2d07f082d3cc8bc7/CMakeLists.txt#L1371-L1375

With above changes applied the CMake will print as below:
图片

mtaillefumier (approved):

mtaillefumier:
Side comment: The message itself already concedes that fftw3 is not really optional. `Some features of CP2K may not work properly without FFTW3` is in practice a hard dependency for any user.

PR #5468, opened 2026-06-26
Toolchain cleanup: Follow-up fix and enhancement

Growl1234:


oschuett (approved):

Growl1234:
This PR breaks CUDA dashboard most likely because `DBCSR_CUDA_ROOT` and `DBCSR_HIP_ROOT` are removed. However, I'm quite strange why CMake would find DBCSR through these `_ROOT` variable...

Growl1234:
Found it: `cmake_cp2k.sh` directly write `DBCSR_DIR` with those variables. This should be dropped.

Issue #5460, opened 2026-06-25
Performance regression of grid_integrate_task_list on CUDA


oschuett:
Looking at the [CUDA Performance test](https://dashboard.cp2k.org/archive/perf-cuda-volta/index_full.html), it seems there was a severe performance regression of `grid_integrate_task_list`  in [diag_cu144_broy.inp](https://dashboard.cp2k.org/archive/perf-cuda-volta/diag_cu144_broy_timings_6cpu_1gpu_full.txt) and [w64PBE.inp](https://dashboard.cp2k.org/archive/perf-cuda-volta/w64PBE_timings_6cpu_1gpu_full.txt) after #5206.

On the bright side, there was also a very nice speedup of `grid_collocate_task_list` :-)

mtaillefumier:
FYI it is very unlikely I will work on this before the release.

oschuett:
Could you perhaps venture a guess as to the nature of regressions, i.e. which calculations are affected?

mtaillefumier:
I opened a PR solving the issue hopefully.

oschuett:
@mtaillefumier, thanks a lot for #5471 and #5478!

The [timings](https://dashboard.cp2k.org/archive/perf-cuda-volta/diag_cu144_broy_timings_6cpu_1gpu_full.txt) for `diag_cu144_broy.inp` are reasonable again. On balance its win some, lose some. I guess, we're walking the Pareto front at this point.

|              | `grid_collocate_task_list` | `grid_integrate_task_list` |
| ------------ | ---- | ---- |
| before #5206 | 27 s | 28 s |
| after #5478  | 17 s | 37 s |

PR #5471, opened 2026-06-26
Fix performance regression in the grid backend

mtaillefumier:
The performance cost comes from the number of iterations in the multipass algorithm. The original CUDA kernel was calculating 20 coefficients in a single pass while the hip version only 10 which requires twice as much time for the same calculation. 

The new implementation now calculate 20 coefficients in a single pass at the cost of shared memory that double. Occupancy will likely decrease because of this. 

I also tried a different implementation that would calculate all coefficients in a single pass up to lp = 5 but it is slower than the multipass implementation by a very large margin.

### DIFF
--- a/src/grid/gpu/grid_gpu_collocate.cu
+++ b/src/grid/gpu/grid_gpu_collocate.cu
@@ -89,7 +89,8 @@ __launch_bounds__(64) void calculate_coefficients(const kernel_params dev_) {
   T *smem_alpha = &shared_memory[0];
   const int tid = thread_global_index();
   const int offset = dev_.sorted_blocks_offset_dev[block_index()];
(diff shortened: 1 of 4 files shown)


oschuett:
For `diag_cu144_broy.inp` the time spend in `grid_integrate_task_list` drops from 92 to 67 seconds.

However, [it used to be](https://dashboard.cp2k.org/archive/perf-cuda-volta/diag_cu144_broy_timings_6cpu_1gpu_full.txt) 28 seconds. So, this is good progress towards #5460, but I wouldn't declare victory just yet.


Should I already merge this PR or do you have time to take another look?

mtaillefumier:
Merge it. I won't have time to look into this much further next week.

oschuett (approved):

PR #5472, opened 2026-06-27
Toolchain: Update libxs* and DBCSR; fix CUDA/HIP package discovery

Growl1234:

PR #5478, opened 2026-06-28
Reordering of coefficients calculations

mtaillefumier:
- As `hab` and forces do not require atomic operations, hab elements can be calculated in parallel.
- Calculations of `hab` is deterministic.
- Atomic operations for calculating the forces and stress coefficients was limited to the minimum possible. They can be removed at the price of calling a separate kernel that would implement a gather instead of a scatter.
- Moved CUDA/HIP architectures setup before calling `find_package(dbcsr)` (author @Growl1234)

### DIFF
--- a/src/grid/gpu/grid_gpu_integrate.cu
+++ b/src/grid/gpu/grid_gpu_integrate.cu
@@ -61,7 +61,7 @@ __device__ __inline__ T block_reduce_64(T *table, const T val_, const int tid) {
 }
 
 template <typename T, bool COMPUTE_TAU>
-__global__ __launch_bounds__(64) void compute_hab_v4(const kernel_params dev_) {
+__global__ __launch_bounds__(64) void compute_hab(const kernel_params dev_) {
   // Copy task from global to shared memory and precompute some stuff.
   extern __shared__ T shared_memory[];
   // T *smem_cab = &shared_memory[dev_.smem_cab_offset];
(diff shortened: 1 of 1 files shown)


mtaillefumier:
@oschuett : shared memory is used in the `hab` calculations to store the alpha coefficients. I could use it to store the cab coefficients as well but the cost is occupation. Profiling the code also reveals that calculating `hab` and `forces` is compute bound 80% (30~40 % memory) and using shared memory only improve the cu total runtime by 0.3 s on my V100 runs. 

The cu test is particularly stringent. The current version of integrate is slower around 5s compared to the previous backend but collocate is 10 s faster so the total runtime is still smaller with the current PR.

mtaillefumier:
`/usr/bin/ld: cannot find -lcudart: No such file or directory` is present again.

Growl1234:
I found the issue: [cmake_cp2k.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/cmake_cp2k.sh) directly write `DBCSR_DIR` with `DBCSR_HIP/CUDA_ROOT` variable which has been removed in toolchain cleanups. This should be dropped, as CMake finding logic should now directly detect `CMAKE_PREFIX_PATH`:

```patch
diff --git a/tools/docker/scripts/cmake_cp2k.sh b/tools/docker/scripts/cmake_cp2k.sh
index d27c83c4ae..4bab46154c 100644
--- a/tools/docker/scripts/cmake_cp2k.sh
+++ b/tools/docker/scripts/cmake_cp2k.sh
@@ -122,7 +122,6 @@ elif [[ "${PROFILE}" == "toolchain_cuda_"* ]] && [[ "${VERSION}" == "psmp" ]]; t
   cmake \
     -GNinja \
     -DCMAKE_INSTALL_PREFIX="${INSTALL_PREFIX}" \
-    -DDBCSR_DIR="${DBCSR_CUDA_ROOT}/lib/cmake/dbcsr" \
     -DSPLA_ROOT="${SPLA_ROOT}-cuda" \
     -DCP2K_WITH_GPU="${PROFILE:15}" \
     -DCP2K_USE_ACCEL=CUDA \
@@ -150,7 +149,6 @@ elif [[ "${PROFILE}" == "toolchain_hip_"* ]] && [[ "${VERSION}" == "psmp" ]]; th
   cmake \
     -GNinja \
     -DCMAKE_INSTALL_PREFIX="${INSTALL_PREFIX}" \
-    -DDBCSR_DIR="${DBCSR_HIP_ROOT}/lib/cmake/dbcsr" \
     -DCP2K_WITH_GPU="${PROFILE:14}" \
     -DCP2K_USE_ACCEL=HIP \
     -DCP2K_USE_MPI=ON \
```

Also, the `DBCSR_ROOT` in toolchain scripts need to be avoided.

mtaillefumier:
Nice finding. I can solve the issue in this PR. Depending on `CMAKE_PREFIX_PATH` is the best way to handle this.

Growl1234:
I am willing to resolve this together with #5472. And I found a similar issue regarding CUDA/HIP package discovery on the COSMA side...

Growl1234:
Another thing that might affect the effiency is, GPU architectures is forced to be 52 via DBCSR's `enable_language(CUDA)` config if CUDA version of DBCSR is found, but the right number of V100 should be 70 (defined in `CMakeLists.txt` on the CP2K side). I am not quite sure whether this is an issue that should be revised.

mtaillefumier:
@Growl1234 : Sorry I edited your previous message by error. Corrected now

mtaillefumier:
> Another thing that might affect the effiency is, GPU architectures is forced to be 52 via DBCSR's `enable_language(CUDA)` config if CUDA version of DBCSR is found, but the right number of V100 should be 70 (defined in `CMakeLists.txt` on the CP2K side). I am not quite sure whether this is an issue that should be revised.

yes we should solve this because using 52 trigger software version of atomic operations which is slow on any modern enough GPU

Growl1234:
Maybe moving the `CP2K_WITH_GPU` to `CMAKE_CUDA_ARCHITECTURES` mapping before dependency discovery would help here. This would ensure that a CUDA-enabled dependency such as DBCSR does not initialize CUDA with the compiler default architecture before CP2K processes the requested GPU target.

I am also considering a few ordering adjustments that might make the CMake configuration output a bit clearer.

[1.patch](https://github.com/user-attachments/files/29436342/1.patch)

mtaillefumier:
> Maybe moving the `CP2K_WITH_GPU` to `CMAKE_CUDA_ARCHITECTURES` mapping before dependency discovery would help here. This would ensure that a CUDA-enabled dependency such as DBCSR does not initialize CUDA with the compiler default architecture before CP2K processes the requested GPU target.
> 
> I am also considering a few ordering adjustments that might make the CMake configuration output a bit clearer.
> 
> [1.patch](https://github.com/user-attachments/files/29435794/1.patch)

it should work. I can review these changes quickly when you consider them ready. I never face this issue because I use CMAKE_CUDA_ARCHITECTURES. spack does the same if I remember correctly.

Growl1234:
> it should work. I can review these changes quickly when you consider them ready. I never face this issue because I use CMAKE_CUDA_ARCHITECTURES. spack does the same if I remember correctly.

Thank you. It passed compilation on my local test, so I think it's ready.

Growl1234:
Would you like to include this CMake patch to your PR?

mtaillefumier:
I certainly can. 
Option 1: You can open a PR on the branch so that authorship is not lost. 
Option 2: I merge the patch directly and make sure to add your name as author.

Let me know what option is more convenient for you.

Growl1234:
I think, just add it to this PR, as it also resolves things about CUDA build configuration.

mtaillefumier:
Results for `diag_cu144_broy.inp` 

This PR
```
 CP2K                                 1  1.0    0.135    0.135  241.259  241.259
 grid_integrate_task_list            15 10.0   31.090   31.090   31.090   31.090
 grid_collocate_task_list            16  7.0   16.389   16.389   16.389   16.389

```
cp2k 2026.1
```
 CP2K                                 1  1.0    0.161    0.161  245.138  245.138
 grid_collocate_task_list            16  7.0   26.119   26.119   26.119   26.119
 grid_integrate_task_list            15 10.0   25.553   25.553   25.553   25.553

```

Growl1234:
Looks like the CMake patch isn't here, but the new commit includes another patch that's applied elsewhere #5472 ...

mtaillefumier:
we are missing something. The error is still here `/usr/bin/ld: cannot find -lcudart: No such file or directory`.

no other option than waiting to merge the PR that solve the issue. That's OK

Growl1234:
I think, according to the small discussion above, what we want in the new commit here is this below patch file unless I uploaded a wrong file... but I saw just the two lines deleted in that commit?

> [1.patch](https://github.com/user-attachments/files/29436342/1.patch)

Regarding testing error, yes, we'd have to wait #5472 to be merged to make the status green because it also contains some other important fixes.

Growl1234:
And I think Pascal regtest might also fail due to the same reason...

Growl1234:
I see #5472 merged, so this PR can rebase.

oschuett (approved):

Issue #5184, opened 2026-05-13
Incompatible version of DBCSR in toolchains


yunai2384:
I encountered a CMake configuration error when trying to build CP2K (version 2026.1) after successfully installing the toolchain.
It appears there is a version string mismatch for the DBCSR library:

    CMakeLists.txt requests: Version 2.6
    Toolchain installed: Version 2026.1.v2026.1...-dirty (instead of 2.9.0)

CMake fails to recognize the installed DBCSR because the version string format is incompatible.

Error Log:
------------------------------------------------------------
-                          DBCSR                           -
------------------------------------------------------------

CMake Error at CMakeLists.txt:615 (find_package):
  Could not find a configuration file for package "DBCSR" that is compatible
  with requested version "2.6".

  The following configuration files were considered but not accepted:

    /home/yunai/E/SW/abinito/cp2k2026.1/tools/toolchain/install/dbcsr-2.9.0/lib/cmake/dbcsr/DBCSRConfig.cmake, version: 2026.1.v2026.1v2026.1-dirty

Growl1234:
Sorry, I cannot reproduce this issue with CP2K v2026.1. Could you please share some more information, such as your system environment, toolchain options, CMake options, and if there's any modifications of related files?

yunai2384:
> Sorry, I cannot reproduce this issue with CP2K v2026.1. Could you please share some more information, such as your system environment, toolchain options, CMake options, and if there's any modifications of related files?

My system is Rockylinux 9, I deleted all the environment variables, including mpich, mpich-devel, and the command was sourced "source tools/toolchain/install/setup"

In the version  of tag=v2026.1, I show the content of dbcsr version below:
1. tools/toolchain/scripts/stage9/install_dbcsr.sh, line 9
  #!/bin/bash -e

  # TODO: Review and if possible fix shellcheck errors.
  # shellcheck disable=all

  [ "${BASH_SOURCE[0]}" ] && SCRIPT_NAME="${BASH_SOURCE[0]}" || SCRIPT_NAME=$0
  SCRIPT_DIR="$(cd "$(dirname "${SCRIPT_NAME}")/.." && pwd -P)"

  dbcsr_ver="2.9.0"
  dbcsr_sha256="a04cacd2203bd97a31ac993f9ab84237a48191140bba29efadbc27db544bbcd6"
2. CMakeLists.txt, line 615
  
  message("\n------------------------------------------------------------")
  message("-                          DBCSR                           -")
[...]

PR #5481, opened 2026-06-29
Reorganize dependency and electronic-structure documentation

Growl1234:
This PR revises and reorganizes several documentation pages.

* Update the dependency documentation to reflect the current build and dependency setup.
* Move the GauXC usage and limitation notes from the general library overview to a dedicated DFT page.
* Add a standalone Wannier90 page containing the material previously included in the k-points documentation.

This keeps the library overview focused on installation and build requirements, while method-specific usage details are documented alongside the relevant electronic-structure functionality.

### DIFF
--- a/docs/getting-started/build-from-source.md
+++ b/docs/getting-started/build-from-source.md
@@ -1,42 +1,121 @@
 # Build from Source
 
 CP2K uses the [CMake](https://cmake.org) build system, which detects dependencies and controls the
-compilation process. The dependencies have to be installed in advance.
+compilation process. This page describes how to obtain a complete CP2K source tree and build it with
+its dependencies.
 
-Currently, CP2K offers two convenient ways for building CP2K from source with its dependencies:
+## Obtaining source code
 
-- The toolchain scripts, which customize and install the required dependencies in a single step,
-  thereby preparing the environment for compiling CP2K. Run `install_cp2k_toolchain.sh` under
-  `cp2k/tools/toolchain` to perform the toolchain steps; for detailed information, see
-  <https://github.com/cp2k/cp2k/blob/master/tools/toolchain/README.md>. After building toolchain,
-  you can use `build_cp2k.sh` to build and install CP2K with prepared toolchain environment.
-- The `make_cp2k.sh` script, a more modern and automated approach which leverages
-  [Spack](https://spack.readthedocs.io) to install the dependencies and subsequently build CP2K.
+CP2K is available as a versioned release tarball or through the Git repository. The directory
+produced by either method is the root of the CP2K source tree, referred to below as `CP2K_ROOT`.
 
-```{note}
-- The toolchain does not cover all optional dependencies due to their complicated dependencies
-  requirements, such as DLA-Future, PEXSI, and SIRIUS with complete support (including NLCG).
-- The latter approach with Spack is only available for the current
-[CP2K master branch](https://github.com/cp2k/cp2k) and CP2K release versions `2026.02` and newer.
+### Release tarballs
+
+For a stable released version, download the versioned `cp2k-<version>.tar.bz2` asset from the
+[CP2K releases](https://github.com/cp2k/cp2k/releases) page and unpack it:
+
+```shell
+tar -xjf cp2k-<version>.tar.bz2
+cd cp2k-<version>
 ```
 
-## make_cp2k.sh
+```{tip}
+It's strongly recommended to use the versioned release tarball rather than GitHub's automatically
+generated `Source code` archive, especially for version <=2025.2. The versioned tarball is the
+release artifact and contains any source components bundled for that release.
+```
 
-The `make_cp2k.sh` script supports and facilitates the build of CP2K and its dependencies from
-scratch using [Spack](https://spack.readthedocs.io) and
-[CMake](https://cmake.org/cmake/help/latest/index.html). The build is performed locally within the
-folder `CP2K_ROOT` which refers to the root directory of CP2K source tree.
+### Git checkout
 
-This script can be run using the default options with `./make_cp2k.sh`. The flags `-h` or `--help`
-print the available options:
+A Git checkout is appropriate for development builds or when a particular branch is required:
+
+```shell
+git clone --recursive https://github.com/cp2k/cp2k.git cp2k
+cd cp2k
+```
+
+To check out a supported release branch, replace the branch name as appropriate:
+
+```shell
+git clone --recursive -b support/v<version> https://github.com/cp2k/cp2k.git cp2k
+cd cp2k
+```
+
+The `--recursive` is important for versions \<=2025.2, since it includes the DBCSR submodule. If the
+repository was cloned without `--recursive`, initialize the required submodule before building:
+
+```shell
+git submodule update --init --recursive
+```
+
+```{warning}
+When configuring CP2K with CMake after building the toolchain inside a CP2K Git checkout with tags,
+DBCSR revision detection can use the enclosing CP2K repository and generate incorrect DBCSR version
+metadata, causing CMake report an error regarding the DBCSR package discovery; see
+[issue #5184](https://github.com/cp2k/cp2k/issues/5184). Therefore, use the versioned release
+artifact rather than a Git tag for released CP2K versions.
+```
+
+## Setting up dependencies and building CP2K
+
+At a minimum, CP2K requires a modern C and Fortran compiler,
+[DBCSR](https://github.com/cp2k/dbcsr/), BLAS, and LAPACK. MPI builds additionally require MPI and
+ScaLAPACK. For currently supported compilers, see the
+[GitHub Wiki page](https://github.com/cp2k/cp2k/wiki/Compiler-Support).
+
+Detailed descriptions of available dependencies can be found in the technologies section:
+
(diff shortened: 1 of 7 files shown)


Growl1234:
@DCM-Uni-Paderborn Please have a look at `gauxc.md` and `wannier90.md`.

docs/getting-started/build-from-source.md
```diff
@@ -26,26 +25,10 @@ Currently, CP2K offers two convenient ways for building CP2K from source with it
 The `make_cp2k.sh` script supports and facilitates the build of CP2K and its dependencies from
 scratch using [Spack](https://spack.readthedocs.io) and
 [CMake](https://cmake.org/cmake/help/latest/index.html). The build is performed locally within the
-folder `CP2K_ROOT` which defaults to the current working directory. This should be a `cp2k` folder
-containing the CP2K source tree which can be download with
+folder `CP2K_ROOT` which refers to the root directory of CP2K source tree.
 
-```
-> git clone https://github.com/cp2k/cp2k.git cp2k
```
he-zilong, line 33: While the directory does not have to be named `cp2k` now, I think we should still add some notes about obtaining the source code, in particular if there is any distinction between release tarballs and `git clone`d repository. (It is not just because of #5184; earlier versions where release tarballs was bundled with DBCSR code but repository code did not contain it had created many confusions.)

docs/technologies/libraries.md
```diff
@@ -269,6 +208,8 @@ SIRIUS is a domain specific library for electronic structure calculations with p
 
 ## Torch (Machine Learning Framework needed for NequIP)
```
he-zilong: As now GauXC is also partly dependent on libtorch, the "needed for NequIP" part may be moved from the section title to the sentence below.

Could this section describe the CPU and GPU hardware requirements for libtorch briefly?

docs/methods/dft/gauxc.md
```diff
@@ -0,0 +1,183 @@
...
+  &XC_FUNCTIONAL
+    &GAUXC
+      FUNCTIONAL PBE
+    &END GAUXC
+  &END XC_FUNCTIONAL
+&END XC
+```
+
+A non-`NONE` model selects a GauXC OneDFT model. The model can be a `.fun` file or an installed
+model name. The underlying functional is optional in this case and defaults to `PBE`:
```
he-zilong, line 33: A link to the format specification and fetch method for the `.fun` files is needed both here and in [the input reference](https://manual.cp2k.org/trunk/CP2K_INPUT/FORCE_EVAL/DFT/XC/XC_FUNCTIONAL.html#cp2k-input-force-eval-dft-xc-xc-functional-gauxc).

docs/methods/electronic_structure/wannier90.md
```diff
@@ -0,0 +1,148 @@
...
+Wannier90 requires a complete, uniformly weighted k-point mesh with its nearest-neighbour
+connectivity. A high-symmetry band path is not a suitable input mesh. See [](../dft/k-points) for
+k-point sampling and convergence.
+
+## Basic workflow
+
+1. Run a periodic, diagonalization-based SCF calculation with a converged k-point mesh and enough
+   bands for the intended Wannierization.
+1. Enable `&DFT%PRINT%WANNIER90`. CP2K writes the Wannier90 input and data files.
+1. Add the Wannier90 settings specific to the calculation, such as projections, disentanglement
```
he-zilong, line 17: Are these three distinct consecutive steps, or three keypoints for one step of calculation? The numbered list items are starting all with 1 here...

he-zilong (commented):
Here, have a link to #4972 and a bit of some very basic opinions...

docs/methods/electronic_structure/wannier90.md
Growl1234: This is the side effect of precommit workflow: it replaces every numbered list with `1.`:
```
+----------------------------------------------------------------------------------------------------------------------------------------------------------------+
|                                              /root/GitHub/cp2k-dev/docs/methods/electronic_structure/wannier90.md                                              |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------+
| File modified:                                                                                                                                                 |
| --- before                                                                                                                                                     |
| +++ after                                                                                                                                                      |
| @@ -13,8 +13,8 @@                                                                                                                                              |
|                                                                                                                                                                |
|  1. Run a periodic, diagonalization-based SCF calculation with a converged k-point mesh and enough                                                             |
|     bands for the intended Wannierization.                                                                                                                     |
| -2. Enable `&DFT%PRINT%WANNIER90`. CP2K writes the Wannier90 input and data files.                                                                             |
| -3. Add the Wannier90 settings specific to the calculation, such as projections, disentanglement                                                               |
| +1. Enable `&DFT%PRINT%WANNIER90`. CP2K writes the Wannier90 input and data files.                                                                             |
| +1. Add the Wannier90 settings specific to the calculation, such as projections, disentanglement                                                               |
|     windows, or post-processing options, to the generated `.win` file, then run Wannier90 with the                                                             |
|     same seed name.                                                                                                                                            |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------+


Summary: Found 1, skipped 0, checked 1, and failed 1 files.                                                  
Status: FAILED
```

This should be fixed, but I have to replace them with non-number list symbol for now...

docs/technologies/libraries.md
Growl1234: Honestly I'm not quite familiar with libtorch, but maybe we can collect it from the PyTorch source tree with the help of LLM?

docs/getting-started/build-from-source.md
Growl1234: Creating am `Obtaining source code` can resolve this, but this may incidates that the description of dependency handling and CMake options needs to be revised. I'll take a look.

docs/methods/electronic_structure/wannier90.md
oschuett: Don't worry. The 1., 1., 1. in the Markdown will still render properly in HTML as 1., 2., 3., ...

Growl1234:
@he-zilong Could you have a look in case I'm missing anything?

he-zilong:
The existing contents are good enough for me, so any further opinions are going to be additive...

The DBCSR section in `libraries.md` can have a link to the "Obtaining source code" section in `build-from-source.md` which can stress again the availability across historical releases.

The "MPI and ScaLAPACK" section in `libraries.md` may mention the difference of the default behaviors regarding CPU affinities via `--bind-to` option between OpenMPI and MPICH. Furthermore, a brief note about the UCX component in OpenMPI would be very nice.

Growl1234:
> The "MPI and ScaLAPACK" section in `libraries.md` may mention the difference of the default behaviors regarding CPU affinities via `--bind-to` option between OpenMPI and MPICH. Furthermore, a brief note about the UCX component in OpenMPI would be very nice.

Yeah, the default `--bind-to` option will have a significant affect on MPI+OpenMP hybrid parallelism. However, I don't quite see the point of adding UCX; maybe I'm missing something?

he-zilong:
> > The "MPI and ScaLAPACK" section in `libraries.md` may mention the difference of the default behaviors regarding CPU affinities via `--bind-to` option between OpenMPI and MPICH. Furthermore, a brief note about the UCX component in OpenMPI would be very nice.
> 
> Yeah, the default `--bind-to` option will have a significant affect on MPI+OpenMP hybrid parallelism. However, I don't quite see the point of adding UCX; maybe I'm missing something?

Thank you for clarifying the binding pattern. Maybe a note about how hyperthreading makes the number of logical cores different from the number of physical cores would also be helpful here.

Correct me if I am wrong, but the UCX component of OpenMPI appears to be (1) a new feature of version 5.x compared with 4.x, (2) installed by default from the `dnf` package manager but not from the custom build (as in toolchain), and (3) prone to emit lots of visible warnings and errors on some machines due to dependency issues. Even if its presence does not seem to enhance performance in practice, users may still be interested how to enable or disable UCX intentionally and how to check if an existing OpenMPI build utilizes UCX or not.

Growl1234:
OpenMPI will auto-decide the inclusion of UCX by default: if `ucx` package is detected from system, it is switched on, otherwise it is switched off. Manually controlling the setting of UCX (`./configure --with-ucx` or `./configure --with-ucx=no`) is on OpenMPI's side, and I don't see how CP2K will take use of this feature.

docs/methods/dft/index.md
```diff
@@ -8,14 +8,15 @@ maxdepth: 2
...
+cutoff
 local_ri
 constrained
 cneo
 linear_scaling
-k-points
-basis_sets
-pseudopotentials
-cutoff
+gauxc
```
oschuett, line 19: We could also place the GauXC page into the *Machine Learning* sections, where it might be more visible.

Growl1234: I considered this, but I would prefer to keep it in the DFT section. GauXC is primarily an external XC integration framework for Quickstep: the page also covers conventional functionals and the native-grid path, while OneDFT/SKALA is only one of its supported use cases.

Placing it under Machine Learning could therefore give the misleading impression that GauXC is itself an ML-specific feature. I placed it near the end of the DFT section as a DFT extension, where it should remain discoverable alongside the relevant functionality. A cross-reference from the Machine Learning section could be added later if needed.

PR #5489, opened 2026-06-29
Add local spack recipe for libxstream

mkrack:


make_cp2k.sh
```diff
@@ -1069,6 +1069,7 @@ if [[ ! -d "${SPACK_BUILD_PATH}" ]]; then
       -e "0,/~cuda/s//+cuda cuda_arch=${CUDA_SM_CODE}/" \
       -e 's/"~cuda\s+~gpu_direct"/"\+cuda \+gpu_direct"/' \
       -e '/\s*#\s*-\s+"fabrics=efa,ucx"/ s/#/ /' \
+      -e '/\s*#\s*-\s+"libxstream@/ s/#/ /' \
```
Growl1234: IIRC libxstream is used for OpenCL not CUDA.

mkrack: Correct. There is no `opencl` option or path yet in `make_cp2k.sh`, so this is only to test building the dependencies including `libxstream` for GPUs with spack.

mtaillefumier (approved):

PR #5446, opened 2026-06-24
PR for adding libGint to CP2K

marcelloPuligheddu:
# LibGint

LibGint is a CUDA library for the calculation of the Hartree-Fock exchange on GPUs.

This pull request enables CP2K to use libGint for the calculation of the Hartree-Fock exchange at Γ-Point.

 

# Integration into CP2K

Compared to a regular HF calculation, the changes needed in the input file are :

```

&FORCE_EVAL  &DFT  &XC  &HF    HFX_LIBRARY libGint

&FORCE_EVAL  &DFT  &XC  &HF    &MEMORY     MAX_MEMORY X

```

HFX_LIBRARY can be libint[default] or libGint

MAX_MEMORY X maintains it usual meaning if HFX_LIBRARY is libint.

If HFX_LIBRARY is libGint X is the maximum amount (in MB) of device memory each MPI rank is able to use during the exchange calculation.

 

E.g.

```

    &XC

      &HF

        HFX_LIBRARY libGint

        &INTERACTION_POTENTIAL

          CUTOFF_RADIUS 6

          POTENTIAL_TYPE TRUNCATED

          T_C_G_DATA ./t_c_g.dat

        &END INTERACTION_POTENTIAL

        &MEMORY

          MAX_MEMORY 7000

        &END MEMORY

        &SCREENING

          EPS_SCHWARZ 1.0E-11

        &END SCREENING

      &END HF

    &END XC

```

 

 

# Installation

 

The pull request includes a toolchain script that automatically downloads and installs libGint. Currently libGint is hosted on GitHub and downloaded as

```

git clone --depth=1 --single-branch --branch release_v1 https://github.com/marcelloPuligheddu/libGint.git libGint-${libGint_ver}

```

Spack build is not supported at present.

 

# Regression test

The PR adds two tests in tests/QS/regtest-libgint.

- H1 is more of a health check, running libint and libGint on H2 to check the computed KS matrix.

- H2O-HFX-6.inp is a quick 6-water-molecules-in-a-box test.

 

# Debugging

The code contains a "both" option for HFX_LIBRARY, in which case both libraries are used and the output KS matrix is compared.

Given the experimental nature of the code, we kept this option. Each rank will output the maximum absolute difference in the KS matrix.

 

# Limitations

The code should still be considered experimental.

 

Each MPI rank can only use one GPU (with CUDA device ID 0). A group of MPI ranks may be bound to different GPUs, for example, by using the following directive when submitting a job script to the Slurm scheduler.

```

[...]

Issue #1250, opened 2020-12-24
Replace CPWARN/CPABORT with Fypp macros


dev-zero:
The traditional CPP employed by Fortran has some limitations when it comes to handling newline/spaces in macros, making it necessary to explicitly use `CALL cp_abort` when using longer comments since it can't introduce newlines itself.
Fypp has facilities to introduce automatic line breaks which would resolve that.

Together with this change we could also introduce a number of restrictions/improvements:

* do not allow empty comments for `CPABORT`: while useful for debugging/development they may be difficult to understand years later as people also tend to not write documentation
* introduce 2-parameter version of `CPASSERT` (and make it the default): C/C++ has the idiom `ASSERT(foo && "my error message")` to add a runtime-visible error message to an assert, helping to document the assert at the same time. Some of our asserts are clear depending on the context, others are not. A mandatory `CPASSERT(<cond>, "message")` would add this context explicitly and help document the code, provided that the `CPASSERT` macro can handle Fortran line breaks in a portable way.

oschuett:
I believe our current [error reporting routines](https://www.cp2k.org/dev:error_handling) have proven to be sufficient.

While we could use Fypp to add a more convenience it would also make the code less self-explanatory, which I think is not worth it.

he-zilong:
> * do not allow empty comments for `CPABORT`

Just a note that many of these were added in commit 5b6a98f back in 2015. Eliminating them will surely be a systematic project that may use some ai help. After that it should be added to the convention check.

PR #5491, opened 2026-06-30
Precommit: check for `CPASSERT(.FALSE.)`, and empty `CPABORT()`

he-zilong:
> Speaking of this, we should probably declare `CPASSERT(.FALSE.)` or `CPABORT("")` as one of the violations to the [coding convention](https://www.cp2k.org/conv). Unfortunately there are [lots of legacy empty CPABORT commands](https://github.com/cp2k/cp2k/issues/1250#issuecomment-4277745791), many of which appear under the `CASE DEFAULT` branch in a `SELECT CASE` construct.

_Originally posted by @he-zilong in https://github.com/cp2k/cp2k/pull/5446#discussion_r3492304086_

This PR adds a new convention for this matter and cleans up the violations. (Edit: with about 90 files touched, the scope of this PR is already so large that 6 code owners are requested to review. Yet there are still remaining violations; I want to see if the next-round conventions check will discover them and take further actions...)

### DIFF
--- a/src/fm/cp_cfm_basic_linalg.F
+++ b/src/fm/cp_cfm_basic_linalg.F
@@ -326,11 +326,9 @@ CONTAINS
                END DO
             END IF
          ELSE
-#if defined(__parallel)
-            CPABORT("to do (pdscal,pdcopy,pdaxpy)")
-#else
-            CPABORT("to do")
-#endif
+            CALL cp_abort(__LOCATION__, &
+                          "cp_cfm_scale_and_add is not yet implemented for cases "// &
+                          "where input two matrix structures are not equivalent")
          END IF
       END IF
       CALL timestop(handle)
@@ -433,11 +431,9 @@ CONTAINS
                END DO
             END IF
          ELSE
-#if defined(__parallel)
-            CPABORT("to do (pdscal,pdcopy,pdaxpy)")
-#else
-            CPABORT("to do")
-#endif
+            CALL cp_abort(__LOCATION__, &
+                          "cp_cfm_scale_and_add_fm is not yet implemented for cases "// &
+                          "where two input matrix structures are not equivalent")
          END IF
       END IF
       CALL timestop(handle)
--- a/src/fm/cp_cfm_basic_linalg.F
+++ b/src/fm/cp_cfm_basic_linalg.F
@@ -329,7 +329,7 @@ CONTAINS
 #if defined(__parallel)
             CPABORT("to do (pdscal,pdcopy,pdaxpy)")
 #else
-            CPABORT("")
+            CPABORT("to do")
 #endif
          END IF
       END IF
@@ -436,7 +436,7 @@ CONTAINS
 #if defined(__parallel)
             CPABORT("to do (pdscal,pdcopy,pdaxpy)")
 #else
-            CPABORT("")
+            CPABORT("to do")
 #endif
          END IF
       END IF
--- a/src/fm/cp_fm_basic_linalg.F
+++ b/src/fm/cp_fm_basic_linalg.F
@@ -218,11 +218,9 @@ CONTAINS
             CALL daxpy(size_a, my_beta, b, 1, a, 1)
 
          ELSE
-#ifdef __parallel
-            CPABORT("to do (pdscal,pdcopy,pdaxpy)")
-#else
-            CPABORT("to do")
-#endif
+            CALL cp_abort(__LOCATION__, &
+                          "cp_fm_scale_and_add is not yet implemented for cases "// &
+                          "where two input matrix structures are not equivalent")
          END IF
 
       END IF
--- a/src/fm/cp_fm_basic_linalg.F
+++ b/src/fm/cp_fm_basic_linalg.F
@@ -221,7 +221,7 @@ CONTAINS
 #ifdef __parallel
             CPABORT("to do (pdscal,pdcopy,pdaxpy)")
 #else
-            CPABORT("")
+            CPABORT("to do")
 #endif
          END IF
 
--- a/tools/conventions/analyze_gfortran_ast.py
+++ b/tools/conventions/analyze_gfortran_ast.py
@@ -195,16 +195,6 @@ def process_log_file(fhandle: TextIO) -> None:
                     text = f'Found lossy conversion {m.group(1)} without KIND argument in "{cur_proc}"'
                     msg(text, 15)
 
-        elif (
-            "CPASSERT(.FALSE.)" in line
-            or 'CPABORT("")' in line
-            or "CPABORT('')" in line
-        ):
-            msg(
-                f'Found CPASSERT(.FALSE.) or empty CPABORT() message in "{cur_proc}"',
-                17,
-            )
-
     # check for run-away DT_END search
     assert skip_until_DT_END is False
 
--- a/tools/conventions/analyze_gfortran_ast.py
+++ b/tools/conventions/analyze_gfortran_ast.py
@@ -195,6 +195,16 @@ def process_log_file(fhandle: TextIO) -> None:
                     text = f'Found lossy conversion {m.group(1)} without KIND argument in "{cur_proc}"'
                     msg(text, 15)
 
+        elif (
+            "CPASSERT(.FALSE.)" in line
+            or 'CPABORT("")' in line
+            or "CPABORT('')" in line
+        ):
+            msg(
+                f'Found CPASSERT(.FALSE.) or empty CPABORT() message in "{cur_proc}"',
+                17,
+            )
+
     # check for run-away DT_END search
     assert skip_until_DT_END is False
 
--- a/tools/precommit/check_file_properties.py
+++ b/tools/precommit/check_file_properties.py
@@ -223,42 +223,13 @@ def check_file(path: pathlib.Path) -> List[str]:
     suppress_cpabort = ["semi_empirical_int_debug.F"]
     if fn_ext == ".F" and basefn not in suppress_cpabort:
         if "CPASSERT(.FALSE.)" in content:
-            warnings += [f"{path}: Forced unexplained abortion with CPASSERT(.FALSE.)"]
+            warnings += [
+                f"{path}: Found CPASSERT(.FALSE.) - please use CPABORT() with messages"
+            ]
         if "CPABORT('')" in content or 'CPABORT("")' in content:
-            warnings += [f"{path}: Unexplained CPABORT() with empty message"]
-
-    # Check IF(.FALSE.)
-    suppress_iffalse = [
-        "ps_implicit_methods.F",
-        "pexsi_interface.F",
-        "glbopt_mincrawl.F",
-        "dm_ls_chebyshev.F",
-        "xc_tpss.F",
-        "pao_ml_neuralnet.F",
-        "tmc_tree_acceptance.F",
-        "tmc_moves.F",
-        "tmc_dot_tree.F",
-        "almo_scf_optimizer.F",
-        "qs_initial_guess.F",
-        "qs_linres_current.F",
-        "atom_xc.F",
-        "hfx_energy_potential.F",
-        "kg_vertex_coloring_methods.F",
-        "fftw3_lib.F",
-        "vibrational_analysis.F",
-        "kg_environment.F",
-        "qs_scf_diagonalization.F",
-        "qs_ks_methods.F",
-        "hfx_load_balance_methods.F",
-        "preconditioner_makes.F",
-        "qs_ot_minimizer.F",
-        "rt_bse.F",
-        "qs_linres_op.F",
-        "molecular_states.F",
-    ]
-    if fn_ext == ".F" and basefn not in suppress_iffalse:
-        if "IF(.FALSE.)" in content or "IF (.FALSE.)" in content:
-            warnings += [f"{path}: Inactive code with IF (.FALSE.)"]
+            warnings += [
+                f"{path}: Found CPABORT() with empty message - please fill in the reason"
+            ]
 
     # check banner
     year = datetime.now(timezone.utc).year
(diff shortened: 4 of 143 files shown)


he-zilong:
Let's migrate www.cp2k.org/conv to github wiki and add the new convention as c017 there.

Growl1234:
`CPABORT(MSG)` expands to `CALL cp__b(__SHORT_FILE__, __LINE__, MSG)`, so the file and line arguments are appended to the same physical source line before the Fortran compiler checks its length. As a result, a `CPABORT` invocation that appears to fit within the 132-column limit can still exceed it after preprocessing.

For messages that are even moderately long, it would be safer to use `CALL cp_abort(__LOCATION__, & ...)` instead. This keeps the location information while allowing the message expression to be continued over subsequent Fortran lines, avoiding line-length failures from macro expansion.

fstein93 (approved):

fstein93:
Great job! As already mentioned, you need to use `CALL cp_abort(...)` in a few cases.

he-zilong:
Thank you for the suggestion. Meanwhile, the following warnings are irrelevant:
```
/opt/cp2k/src/nnp_acsf.F:74:61:

   74 |       TYPE(nnp_neighbor_type)                            :: neighbor
      |                                                             ^~~~~~~~
Warning: 'nnp_neighbor_type.383.dist_rad.dim[0].lbound' is used uninitialized [-Wuninitialized]
```
```
/opt/cp2k/src/topology_coordinate_util.F:175:66:

  175 |                   IF ((id_work(k) == atom_list(j)%id_name) .AND. (charge(k) == atom_info%atm_charge(first + j - 1))) THEN
      |                                                                  1
Warning: Equality comparison for REAL(8) at (1) [-Wcompare-reals]
/opt/cp2k/src/topology_coordinate_util.F:257:79:

  257 |             IF ((atom_info%id_atom_names(i) == atom_info%id_atmname(j)) .AND. (charge(i) == atom_info%atm_charge(j))) THEN
      |                                                                               1
Warning: Equality comparison for REAL(8) at (1) [-Wcompare-reals]
/opt/cp2k/src/topology_coordinate_util.F:728:71:

  728 |                   IF ((id2str(atom_list(j)%id_name) == atmname) .AND. (qeff == atom_info%atm_charge(first + j - 1))) THEN
      |                                                                       1
Warning: Equality comparison for REAL(8) at (1) [-Wcompare-reals]
```

fstein93:
Exactly, both warnings should have already appeared beforehand and can be ignored.

he-zilong:
Apparently the new rule is not yet active in the conventions test...

JWilhelm (approved):
Looks good to me concerning the mp2* and rpa* files!

tools/conventions/analyze_gfortran_ast.py
```diff
@@ -195,6 +195,16 @@ def msg(message: str, conv_num: int) -> None:
                     text = f'Found lossy conversion {m.group(1)} without KIND argument in "{cur_proc}"'
                     msg(text, 15)
 
+        elif (
```
oschuett: This actually parses the [AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree) produced by `gfortran -fdump-fortran-original`.

Since you want to check the source code, a better place would be [precommit/check_file_properties.py](https://github.com/cp2k/cp2k/blob/master/tools/precommit/check_file_properties.py).

he-zilong: Sorry I do not quite understand. I made up the new rule based on the existing convention checks, which I assumed to be looking at the source code (or something syntactically the same); do you mean that constructs like `CPASSERT(.FALSE.)` and `CPASSERT("")` will be converted to an AST that will not have these exact strings of characters, and thus will not  work? Could this be remedied by looking for `CPASSERT` and `CPABORT` commands followed by seeing if the argument is `.FALSE.` and empty string-type variables?

oschuett: I'm sorry. This is indeed a bit confusing. We have two places for _checks_. The [conventions](https://dashboard.cp2k.org/archive/conventions/index.html) are for checks that need to run a compiler. Check that only look a source code are in [check_file_properties.py](https://github.com/cp2k/cp2k/blob/master/tools/precommit/check_file_properties.py) and are run as part of `make_pretty.sh` and [precommit](https://dashboard.cp2k.org/archive/precommit/index.html).

For an example see [this](https://github.com/cp2k/cp2k/blob/23ff2abb531b087bba98bfa5ea37e018e4965359/tools/precommit/check_file_properties.py#L219-L225).

he-zilong: I see your point, thanks for explaining! I have wondered why the conventions need a standalone check instead of going as part of the precommit check. (But then, would it be feasible to integrate the conventions check into `regtest pdbg` which also runs a compiler and is required for every PR?)

After some consideration though, I think my idea of "looking for CPASSERT and CPABORT commands followed by seeing if the argument is .FALSE. and empty string-type variables" would be better modelled after the WRITE statement here:
https://github.com/cp2k/cp2k/blob/9fc3e74bfed7d2aeda081dfbe252eba3852af0b8/tools/conventions/analyze_gfortran_ast.py#L174-L177

oschuett: > But then, would it be feasible to integrate the conventions check into regtest pdbg which also runs a compiler and is required for every PR?

That's an interesting idea, but I'm afraid that would be too much of a burden. The convention uses different [compiler flags](https://github.com/cp2k/cp2k/blob/9fc3e74bfed7d2aeda081dfbe252eba3852af0b8/cmake/CompilerConfiguration.cmake#L116) and it [redirects](https://github.com/cp2k/cp2k/blob/9fc3e74bfed7d2aeda081dfbe252eba3852af0b8/tools/docker/scripts/cmake_cp2k.sh#L178) the compilers stdout/stderr using [redirect_gfortran_output.py](https://github.com/cp2k/cp2k/blob/master/tools/conventions/redirect_gfortran_output.py).

he-zilong:
The precommit check for `CPASSERT(.FALSE.)` is working as it catches ` src/mixed_cdft_methods.F` as expected.

The convention check for `CPABORT("")` is not working yet - Shouldn't its macro expansion as `CALL cp__b(__SHORT_FILE__,__LINE__,MSG)` make the `MSG` as `''`?

tools/precommit/check_file_properties.py
```diff
@@ -224,6 +224,11 @@ def check_file(path: pathlib.Path) -> List[str]:
             if STR_END_SPACE_RE.search(a) and STR_BEGIN_SPACE_RE.search(b):
                 warnings += [f"{path}:{i+1} Double space in multi-line string"]
 
+    # Check CPASSERT(.FALSE.)
+    if fn_ext == ".F":
+        if "CPASSERT(.FALSE.)" in content:
```
oschuett, line 225:
```suggestion
        if "CPASSERT(.FALSE.)" in content or 'CPABORT("")' in content:
```

tools/conventions/analyze_gfortran_ast.py
```diff
@@ -159,6 +159,12 @@ def msg(message: str, conv_num: int) -> None:
                 msg(f'Found CALL RANDOM_SEED in procedure "{cur_proc}"', 105)
             elif tokens[1].lower().startswith("_gfortran_execute_command_line"):
                 msg(f'Found CALL EXECUTE_COMMAND_LINE in procedure "{cur_proc}"', 106)
+            elif tokens[1].lower() == "cp__b":
+                if tokens[4] == "''":
+                    msg(
+                        f'Found CPABORT with empty message in procedure "{cur_proc}"',
+                        107,
+                    )
```
oschuett:
```suggestion
```

oschuett:
> The convention check for CPABORT("") is not working yet - Shouldn't its macro expansion as CALL cp__b(__SHORT_FILE__,__LINE__,MSG) make the MSG as ''?

Yes, but there is no need to take this into account. Just check for both patterns in the original source code, i.e. before the pre-processor is applied.

he-zilong:
> > The convention check for CPABORT("") is not working yet - Shouldn't its macro expansion as CALL cp__b(**SHORT_FILE**,**LINE**,MSG) make the MSG as ''?
> 
> Yes, but there is no need to take this into account. Just check for both patterns in the original source code, i.e. before the pre-processor is applied.

Essentially, the suggestions above are excluding the checks from the rigorous coding conventions then... I am puzzled if this means that new additions to the conventions system is not encouraged.

Do new precommit items require corresponding automatic fixes, since `make_pretty.sh` uses the `--allow-modifications` option of `precommit.py`?

oschuett:
> ... I am puzzled if this means that new additions to the conventions system is not encouraged.

Not at all, but for these particular new checks the `check_file_properties.py` is just the far better place to add them. 

It will still be rigorously enforced on every pull request.

> Do new precommit items require corresponding automatic fixes, since make_pretty.sh uses the --allow-modifications option of precommit.py?

No, it's not a requirement. Only some of the tools invoked by precommit have auto-fixes - mostly the formatters.

he-zilong:
> Not at all, but for these particular new checks the `check_file_properties.py` is just the far better place to add them.
> 
> It will still be rigorously enforced on every pull request.

Done; converting to draft now as I expect a few remaining cases of violations scattered around which would be addressed a while later.

```
src/minimax/minimax_exp_k15.F: Unexplained CPABORT() with empty message
src/mixed_cdft_methods.F: Forced unexplained abortion with CPASSERT(.FALSE.)
src/almo_scf_optimizer.F: Unexplained CPABORT() with empty message
src/pw/pw_spline_utils.F: Unexplained CPABORT() with empty message
src/qs_linres_current.F: Unexplained CPABORT() with empty message
src/qs_collocate_density.F: Unexplained CPABORT() with empty message
src/semi_empirical_int_utils.F: Unexplained CPABORT() with empty message
src/atom_utils.F: Unexplained CPABORT() with empty message
src/pw/ps_wavelet_fft3d.F: Unexplained CPABORT() with empty message
src/fm/cp_fm_basic_linalg.F: Unexplained CPABORT() with empty message
```

jpoto (approved):
Looks good to me.

src/fm/cp_cfm_basic_linalg.F
he-zilong, line 1: The purpose of the `if` constructs revised here are unclear; perhaps using a unified `CPABORT("Not yet implemented")` would be better?

src/fm/cp_fm_basic_linalg.F
he-zilong, line 1: Ditto

he-zilong (commented):
The final boss would be `./src/semi_empirical_int_debug.F` with **18** instances of `CPABORT("")` together with tons of `WRITE (*, *)` where the hardcoded asterisk write unit is also a convention violation...

he-zilong:
I have decided not to invest more time on enforcing the new rules. Therefore, I have added `./src/semi_empirical_int_debug.F` as an exception to suppress the detection.

Meanwhile, I have also requested a check for `IF (.FALSE.)` constructs, usually leftovers from debugging, but the existing instances have been left untouched.

fstein93:
> I have decided not to invest more time on enforcing the new rules. Therefore, I have added `./src/semi_empirical_int_debug.F` as an exception to suppress the detection.
> 
> Meanwhile, I have also requested a check for `IF (.FALSE.)` constructs, usually leftovers from debugging, but the existing instances have been left untouched.

When reviewing PRs, I ask people to either remove debug code entirely (prefered) or if necessary, wrap it in `IF (.FALSE.) THEN ... END IF`. This may be reasonable with experimental code. Usually, people remove it directly. But I am in favor of removing this kind of code after 1-2 years like I did in several occasions. Another problem is code hidden behind comments which becomes smelly very easily due to lack of compiler checks. That is something which is more difficult to cover in this checks due to a lack of common patterns (Developers may use equations/assignments as a comment or start a comment with `call` etc.)

tools/precommit/check_file_properties.py
```diff
@@ -224,6 +219,47 @@ def check_file(path: pathlib.Path) -> List[str]:
             if STR_END_SPACE_RE.search(a) and STR_BEGIN_SPACE_RE.search(b):
                 warnings += [f"{path}:{i+1} Double space in multi-line string"]
 
+    # Check CPASSERT(.FALSE.) and empty CPABORT() messages
+    suppress_cpabort = ["semi_empirical_int_debug.F"]
+    if fn_ext == ".F" and basefn not in suppress_cpabort:
+        if "CPASSERT(.FALSE.)" in content:
+            warnings += [f"{path}: Forced unexplained abortion with CPASSERT(.FALSE.)"]
```
oschuett: The phrase _"forced abortion"_ could be badly misunderstood.

```suggestion
            warnings += [f"{path}: Found CPASSERT(.FALSE.) - please use CPABORT instead."]
```

tools/precommit/check_file_properties.py
```diff
@@ -224,6 +219,47 @@ def check_file(path: pathlib.Path) -> List[str]:
...
+        if "CPABORT('')" in content or 'CPABORT("")' in content:
+            warnings += [f"{path}: Unexplained CPABORT() with empty message"]
```
oschuett:
```suggestion
            warnings += [f"{path}: Found CPABORT() with empty message"]
```

oschuett (commented):
I agree with @fstein93 that `IF (.FALSE.)` should be allowed. It doesn't hurt usability and helps development - admittedly while cluttering the code a bit.

oschuett:
Since this PR makes low-risk changes to a lot of files, I'm using my admin privilege to merge it despite some code owner approvals still missing.

he-zilong:
> Since this PR makes low-risk changes to a lot of files, I'm using my admin privilege to merge it despite some code owner approvals still missing.

Thank you! Let's leave the adjustment of the wording and contents in the error messages for future PRs by these code owners.

hforbert:
ah sorry, since it were trivial changes to "my" files and on the page said somewhere,
that only one approval was needed (which several already gave), I thought
it was not necessary for me anymore.

PR #5497, opened 2026-07-01
make_cp2k.sh: resume interrupted dependency builds

Growl1234:
This changes `make_cp2k.sh` so that an interrupted dependency build can be resumed by simply re-running the script with the same dependency-related options.

Previously, the presence of the local `spack/` directory caused the script to skip the dependency setup path entirely. If `BUILD_DEPENDENCIES_COMPLETED` was missing, it aborted and required either `-bd` or manual removal of the whole directory. As a result, the local Spack checkout, virtual environment, environment state, and partially completed installation could not be reused.

Now, the completion marker determines whether the dependency build needs to run:

* an unfinished local Spack setup is reused and `spack install` is run again;
* completed packages and Spack's existing installation state are retained;
* the Spack release archive is downloaded with resume support;
* an existing venv with the required cache dependencies is reused;
* `-bd` keeps its existing meaning and still removes the local dependency stack for a clean rebuild.

A resumed build should be run with the same dependency-related options as the interrupted invocation. Changing the MPI implementation, enabled features, compiler choice, GPU setup, or other dependency-selection options still requires `-bd` to avoid reusing an incompatible concretized environment.

### DIFF
--- a/make_cp2k.sh
+++ b/make_cp2k.sh
@@ -933,21 +933,39 @@ if [[ "${REBUILD_CP2K}" == "yes" ]]; then
   done
 fi
 
-if [[ ! -d "${SPACK_BUILD_PATH}" ]]; then
+# An incomplete build already contains the Spack installation, source cache,
+# environment, and every successfully installed dependency. Re-enter the
+# setup/install path until the completion marker is written.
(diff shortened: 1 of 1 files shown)


mtaillefumier (approved):
LGTM

mkrack (commented):
Thanks, looks good to me.

Issue #4663, opened 2026-01-04
Performance regressions of COSMA + OpenMP


oschuett:
The `GW_PBE_4benzene` OpenMP benchmark spends [57 additional seconds](https://dashboard.cp2k.org/archive/perf-openmp/index_full.html) in  `parallel_gemm_fm_cosma` since we [switched to CMake](#4483). 

The problem disappears when CP2K is build without SIRIUS. 

I suspect the root-cause a linking problem with COSMA's overloading of `pdgemm`.

```
From /workspace/artifacts/GW_PBE_4benzene_32omp.out:
-------------------------------------------------------------------------------
 -                                                                             -
 -                                T I M I N G                                  -
 -                                                                             -
 -------------------------------------------------------------------------------
 SUBROUTINE                       CALLS  ASD         SELF TIME        TOTAL TIME
                                MAXIMUM       AVERAGE  MAXIMUM  AVERAGE  MAXIMUM
 CP2K                                 1  1.0    0.014    0.014  128.669  128.669
 qs_energies                          1  2.0    0.000    0.000  128.308  128.308
 mp2_main                             1  3.0    0.000    0.000  125.280  125.280
 mp2_gpw_main                         1  4.0    0.000    0.000  125.173  125.173
 rpa_ri_compute_en                    1  5.0    0.000    0.000  120.751  120.751
 rpa_num_int                          1  6.0    0.001    0.001  120.742  120.742
 compute_mat_P_omega                  1  7.0    0.003    0.003   57.320   57.320
 compute_mat_P_omega_contract        10  8.0    8.259    8.259   56.778   56.778
 parallel_gemm_fm_cosma             105  8.4   56.776   56.776   56.776   56.776
 ```

@mtaillefumier @RMeli

RMeli:
@oschuett thanks for the ping, we will look into it.

mtaillefumier:
I cannot find any reason why excluding SIRIUS from linking changes the timings to such drastic extent. Several elements do not seem to add up:
- `SIRIUS` does not use `COSMA` and the linking order is `SIRIUS` , `COSMA`, `SCALAPACK`, `BLAS`.  
- `cp2k.psmp` appears to be executed on a single rank with a very large number of `OMP` threads. It reduces `COSMA` influence by a significant amount as there is no communication happening during the execution.
[...]

PR #5499, opened 2026-07-01
Spack: Remove `+dlaf` from SIRIUS requirement

Growl1234:
A draft to see if this is really relevant:

>> Unfortunately it seems I am able to reproduce the slowdown locally:
>> [1.txt](https://github.com/user-attachments/files/29535085/1.txt)

> It seems disabling SIRIUS in Dockerfile (`-df sirius`) makes the issue disappear on my machine.

_Originally posted by @Growl1234 in https://github.com/cp2k/cp2k/issues/5090#issuecomment-4851946869_


oschuett:
This might be related to the earlier performance regression #4663, which I also addressed by disabling SIRIUS #4646.

Growl1234:
Toolchain pdbg regtests and Spack psmp (even without `--keepalive` before) do not encounter this; it happens only with Spack pdbg. So I'm still remaining curious what would cause this...

Growl1234:
Looks like it's really relevant...

Before (the [dashboard report](https://dashboard.cp2k.org/archive/spack-pdbg/commit_0d88fc2a8ea56536466e8e4a348537a19445409b.txt)):
```
parallel_rng_types_unittest     - OK (   7.02 sec)
dbt_tas_unittest                - OK (   7.29 sec)
libcp2k_unittest                - OK (   7.78 sec)
dbt_unittest                    - OK (   8.62 sec)
```

After (current Spack regtest report; similar to toolchain pdbg regtest):
```
parallel_rng_types_unittest     - OK (   1.29 sec)
dbt_tas_unittest                - OK (   1.56 sec)
libcp2k_unittest                - OK (   2.09 sec)
dbt_unittest                    - OK (   2.96 sec)
```

mkrack:
> Toolchain pdbg regtests and Spack psmp (even without `--keepalive` before) do not encounter this; it happens only with Spack pdbg. So I'm still remaining curious what would cause this...

`spack pdbg` builds all dependencies consistently with `RelWithDebInfo`. How are dependencies built like SIRIUS and COSMA by `toolchain pdbg`, currently?

Growl1234:
> `spack pdbg` builds all dependencies consistently with `RelWithDebInfo`. How are dependencies built like SIRIUS and COSAM by `toolchain pdbg`, currently?

Both are built with `RelWithDebInfo` unconditionally.

Growl1234:
There are some SIRIUS FLAGS in toolchain:
https://github.com/cp2k/cp2k/blob/e6548e293017dc583cc97de7f96a5e9a5868611d/tools/toolchain/scripts/stage8/install_sirius.sh#L35-L49
which is passed:
https://github.com/cp2k/cp2k/blob/e6548e293017dc583cc97de7f96a5e9a5868611d/tools/toolchain/scripts/stage8/install_sirius.sh#L83-L84

Growl1234:
Another difference, SIRIUS in toolchain is built without NLCG and memory pool (umpire).

mkrack:
> Another difference, SIRIUS in toolchain is built without NLCG and memory pool (umpire).

I don't think we touch NLCG. We can try to disable umpire with `~memory_pool` to see if that works and has an impact.

Growl1234:
I'll have a try locally.

Growl1234:
Disabling Umpire not only fails to alleviate the slowdown, but also causes a substantial increase in memory consumption during regtests, which makes the situation even more puzzling [Update: memory consumption seems to have nothing to do with umpire].

Growl1234:
Removing `+dlaf` from SIRIUS configuration requirement works.

Growl1234:
Since DLAF is disabled by default in `make_cp2k.sh`, I'll also comment them in `specs:` section.

mkrack:
> Since DLAF is disabled by default in `make_cp2k.sh`, I'll also comment them in `specs:` section.

DLAF is disabled because the DLAF tests are failing. 

@RMeli Could the slowdown be a side effect of building SIRUS with DLAF support?

Growl1234:
> Could the slowdown be a side effect of building SIRUS with DLAF support?

The increased memory consumption seems also related to SIRIUS with DLAF.

mkrack:
> > Could the slowdown be a side effect of building SIRUS with DLAF support?
> 
> The increased memory consumption seems also related to SIRIUS with DLAF.

If this is true, we should consider to exclude DLAF  from [CP2K_USE_EVERYTHING](https://github.com/cp2k/cp2k/blob/master/CMakeLists.txt#L157-L158).
@RMeli @oschuett What do you think?

oschuett:
> If this is true, we should consider to exclude DLAF from [CP2K_USE_EVERYTHING](https://github.com/cp2k/cp2k/blob/master/CMakeLists.txt#L157-L158).

No, I would not introduced exceptions to `CP2K_USE_EVERYTHING`. It is not intended for end-user. It's to ensure that we're testing every available feature. So, if we need to temporarily disable a library then it's actually good that it [shows up explicitly](https://github.com/cp2k/cp2k/blob/9fc3e74bfed7d2aeda081dfbe252eba3852af0b8/tools/docker/scripts/cmake_cp2k.sh#L46-L48).

mkrack:
> No, I would not introduced exceptions to `CP2K_USE_EVERYTHING`. It is not intended for end-user. It's to ensure that we're testing every available feature. So, if we need to temporarily disable a library then it's actually good that it [shows up explicitly](https://github.com/cp2k/cp2k/blob/9fc3e74bfed7d2aeda081dfbe252eba3852af0b8/tools/docker/scripts/cmake_cp2k.sh#L46-L48).

If I am not mistaken, we have with `openPMD` already an exception?

Growl1234:
This behavior seems to be a mistake? OpenPMD is not included in the `Dependencies not included in this build` report either...

oschuett:
> If I am not mistaken, we have with openPMD already an exception?

Oh, I was not aware of that. That's not good, because as a result we have not been testing it.

mkrack:
> > If I am not mistaken, we have with openPMD already an exception?
> 
> Oh, I was not aware of that. That's not good, because as a result we have not been testing it.

It is tested by the spack testers where that feature is included explicitly by `-ef openpmd`. That flag could be dropped if it is part of EVERYTHING.

Growl1234:
And since OpenPMD is [required to have MPI](https://github.com/cp2k/cp2k/blob/9fc3e74bfed7d2aeda081dfbe252eba3852af0b8/CMakeLists.txt#L954-L962), it would be better to make it a `cmake_dependent_option`. But this seems out of scope for this PR.

tools/spack/cp2k_deps_p.yaml
```diff
@@ -166,7 +166,6 @@ spack:
       require:
         - "~apps"
         - "+dftd3"
-        - "+dlaf"
```
mkrack, line 169: I suggest to use `~dlaf` explicitly so that we can later activate DLAF conditionally again in `make_cp2k.sh` by uncommenting the specs `dla-future` and `dla-future-fortran` and switching to `+dlaf`.

Growl1234: I think, the SIRIUS DLAF variant should be kept consistent with CP2K: `~dlaf` by default (in fact already the default in Spack recipe), and switch both to `+dlaf` when DLAF is explicitly enabled.

mkrack: Sorry, I do not understand what you mean with "should be kept consistent with CP2K"?

RMeli:
> @RMeli Could the slowdown be a side effect of building SIRUS with DLAF support?

Is DLAF actually used by SIRIUS in these runs? If yes, it is possible that there is a slowdown in the tests (DLAF works well for large matrices and large block sizes, but has overhead for small ones).

> The increased memory consumption seems also related to SIRIUS with DLAF.

Likewise, with small block sizes (and depending on page size), DLAF memory consumption can be significant indeed.

Growl1234:
> Is DLAF actually used by SIRIUS in these runs? If yes, it is possible that there is a slowdown in the tests (DLAF works well for large matrices and large block sizes, but has overhead for small ones).

The fact is that debug type of CP2K built with DLAF-supported SIRIUS will run globally slow with >1 MPI ranks, even on `cp2k --version` and  `libcp2k_unittest`. So I'm afraid this is not the reason. Perhaps enabling DLA-Future in SIRIUS causes a substantial LeakSanitizer-related teardown overhead for multi-rank pdbg executables?

tools/spack/cp2k_deps_p.yaml
Growl1234: Assuming the DLAF issue is fixed in the future and the script enable DLAF by default: When `make_cp2k.sh` is run with `-df dlaf`, the installation of SIRIUS should not require DLAF support as well.

Currently in the new commit, I keep `~dlaf` and only replace it with `+dlaf` when `-ef dlaf` is explicitly passed.

mkrack (approved):

Growl1234:
Installing DLA-Future via `make_cp2k.sh` with `-cv pdbg -np 2 -j 24` on my physical environment (Rocky Linux 10.2, GCC 14.3.1) causes OOM error (32GiB RAM + 16GiB swap), which doesn't happen with `-cv psmp`...

mkrack:
> Installing DLA-Future via `make_cp2k.sh` with `-cv pdbg -np 2 -j 24` on my physical environment (Rocky Linux 10.2, GCC 14.3.1) causes OOM error (32GiB RAM + 16GiB swap), which doesn't happen with `-cv psmp`...

Did you try already less than `-j 24`, e.g `-j 12`?

Growl1234:
And when running DLAF tests with psmp and >1 MPI ranks, for example:
```bash
mpirun -np 4 cp2k.popt H2O-6.inp
```
It actually uses all 32 logic threads and then gives a segfault after 1-2 SCF steps.

Setting `PIKA_THREADS` avoids the oversubscription that otherwise uses all 32 logical CPUs, but it does not prevent the segmentation fault. Moreover, the number of CPU threads actually used appears to be determined solely by `PIKA_THREADS`; it does not seem to scale with either the number of MPI ranks or the OpenMP settings...

Growl1234:
> Did you try already less than `-j 24`, e.g `-j 12`?

With `-np 4 -j 24` or `-np 2 -j 12`, things get better and no OOM encountered.

mkrack:
> > Did you try already less than `-j 24`, e.g `-j 12`?
> 
> With `-np 4 -j 24` or `-np 2 -j 12`, things get better and no OOM encountered.

As a rule of thumb, I estimate 4GiB of RAM per process, i.e. dividing the `j` value by `np` value multiplied by 4 should not exceed your total RAM.

PR #5501, opened 2026-07-01
Add local spack recipe for libxstream

mkrack:


tools/spack/spack_repo/cp2k_dev/packages/libxstream/package.py
```diff
@@ -0,0 +1,45 @@
...
+
+    depends_on("opencl")
+    depends_on("opencl-headers")
+    depends_on("opencl-icd-loader")
+
+    def cmake_args(self):
+        spec = self.spec
+        args = [
+            self.define_from_variant("CMAKE_POSITION_INDEPENDENT_CODE", "pic"),
+            self.define_from_variant("LIBXSTREAM_SHARED", "shared"),
```
mtaillefumier: `libxstream` uses `BUILD_SHARED_LIBS` to build shared libraries.

mtaillefumier (approved):

tools/spack/spack_repo/cp2k_dev/packages/libxstream/package.py
mkrack: Thanks for the hint.

mtaillefumier: It is not always clear what variable should be used. Some project uses their own version.

PR #5506, opened 2026-07-01
make_cp2k.sh: Avoid expanding unset SPACK_ENV

Growl1234:
Follow up to #5497: Avoid expanding `SPACK_ENV` after it has been [explicitly unset](https://github.com/cp2k/cp2k/blob/2c8265a529330abef610916e7989fde72eb0bdf5/make_cp2k.sh#L949-L950).

The existence of `cp2k_env` is already determined by `spack env list`; `SPACK_ENV` only describes the environment currently activated in the shell and is not needed here. This also prevents an `unbound variable` error with `set -u`.


mkrack (commented):
Encountered the problem with `unset SPACK_ENV` already in my tests. Thanks for fixing it.

mkrack (approved):

PR #5508, opened 2026-07-02
Docker: Test "keepalive" case specifically

Growl1234:
This PR runs the Spack-based CPU regtest configurations with `--flagslow` and remove `--keepalive` from their default test options.

However, from https://github.com/cp2k/cp2k/pull/5365#issuecomment-4732850294 and #5406 it seems some cases are detected only in keepalive mode. Therefore, add two tests dedicatedly to retain explicit coverage for that mode.

### DIFF
--- a/tools/dashboard/dashboard.conf
+++ b/tools/dashboard/dashboard.conf
@@ -298,6 +298,17 @@ timeout:     170
 host:        GCP
 report_url:  https://storage.googleapis.com/cp2k-ci/dashboard_psmp-4ranks_report.txt
 
+[current-psmp-keepalive]
+name:        Current Toolchain (psmp, keepalive)
+host:        GCP
+report_url:  https://storage.googleapis.com/cp2k-ci/dashboard_psmp-keepalive_report.txt
+
+[spack-psmp-keepalive]
+name:        Spack (psmp, keepalive)
+timeout:     170
+host:        GCP
+report_url:  https://storage.googleapis.com/cp2k-ci/dashboard_spack-psmp-keepalive_report.txt
+
 [spack-psmp-4x2]
 name:        Spack (psmp, 4x2)
 timeout:     170
--- a/tools/docker/Dockerfile.test_psmp_keepalive
+++ b/tools/docker/Dockerfile.test_psmp_keepalive
@@ -1,0 +1,86 @@
+#
+# This file was created by generate_dockerfiles.py.
+# Usage: podman build --shm-size=1g -f ./Dockerfile.test_psmp_keepalive ../../
+#
+
+FROM ubuntu:26.04
+
+# Install requirements for the toolchain.
+WORKDIR /opt/cp2k-toolchain
+COPY ./tools/toolchain/install_requirements*.sh ./
+RUN ./install_requirements.sh ubuntu:26.04
+
+# Install the toolchain.
+RUN mkdir scripts
+COPY ./tools/toolchain/scripts/VERSION \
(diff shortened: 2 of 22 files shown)


Growl1234:
There are 5 slow tests in Spack pdbg all regarding SIRIUS...

Growl1234:
> There are 5 slow tests in Spack pdbg all regarding SIRIUS...

It can be reproduced locally (`OMP_NUM_THREADS=1`):

- mpirun -np 2 cp2k.psmp Fe.inp
```
 -                                T I M I N G                                  -
 CP2K                                 1  1.0    0.003    0.003    3.692    3.692
 pwdft_calc_energy_force              1  2.0    2.668    2.668    2.668    2.668
 pwdft_init                           1  2.0    1.020    1.020    1.021    1.021

real  0m3.906s
user  0m7.439s
sys   0m0.251s
```
- mpirun -np 2 cp2k.pdbg Fe.inp
```
 -                                T I M I N G                                  -
 CP2K                                 1  1.0    0.006    0.006   20.600   20.600
 pwdft_calc_energy_force              1  2.0   11.725   11.725   11.725   11.725
 pwdft_init                           1  2.0    8.867    8.867    8.868    8.868

real  0m20.864s
user  0m41.046s
sys   0m0.584s
```

Growl1234:
I was missing something before...
SIRIUS built via toolchain is actually `Release` type, and the `SIRIUS_OPT` or `SIRIUS_DBG` seems not make sense.

tools/spack/cp2k_deps_p.yaml
```diff
@@ -165,6 +165,7 @@ spack:
 
     sirius:
       require:
+        - "build_type=Release"
```
mkrack: The purpose of the `spack pdbg` testers is to build a software stack for CP2K with debug information available for all dependencies. I am a bit reluctant to start bypassing this selectively, because this is not a problem of our testers or how we are testing things. Therefore any issue in that respect should be communicated to the SIRIUS developers, because the problems with a debug compilation should be addressed on the SIRIUS side and it is up to the SIRIUS developers how a debug compilation of SIRIUS should be performed.

oschuett:
It's true that `--keepalive` reveals problems that otherwise don't show up, because it re-initializes CP2K. However, in the past there was not much appetite for fixing these issues, so instead we started [skipping them](https://github.com/cp2k/cp2k/blob/5976d25a45174d28df9c77845b753e0a0585d6d8/tests/do_regtest.py#L45).

The original motivation for the `--keepalive` flag was to speedup [GPU tests](https://github.com/cp2k/cp2k/blob/5976d25a45174d28df9c77845b753e0a0585d6d8/tools/docker/scripts/test_regtest.sh#L49), by allowing DBCSR to resuse its JITed kernels. So we already have some coverage for it.

As for the `--flagslow`, it is just very flaky. Hence, we are only using it for [sdbg and pdbg](https://github.com/cp2k/cp2k/blob/5976d25a45174d28df9c77845b753e0a0585d6d8/tools/docker/scripts/test_regtest.sh#L54), where it already produces too much noise for my taste.

So, personally I don't think we should add these two new tests.

tools/spack/cp2k_deps_p.yaml
Growl1234: SIRIUS developer replied in https://github.com/electronic-structure/SIRIUS/issues/1161#issuecomment-4873979453:

>  This is normal. Debug mode triggers strict boundary checks of the access to mdarray elements. Production runs must always be compiled with Release mode.

PR #5503, opened 2026-07-01
Migrate development-related wiki pages

fstein93:
This PR migrates a variety of development-related wiki pages to our docs. I updated the information of the old wiki.

PR #5510, opened 2026-07-02
Libxc: Preserve aliases when registering XC functional sections

Growl1234:
Libxc can provide multiple names for the same functional ID. CP2K currently iterates over the IDs and obtains a name from each ID again. With the Libxc development branch, this will turn two aliases into the same canonical name, so adding the `XC_FUNCTIONAL` sections fails with a duplicate-section error. Therefore, use Libxc's list of functional names when creating the sections, and only resolve the functional ID afterwards.

The change was tested with both 7.0.0 and the current development branch. Since 7.0.0 already includes the required API, this is not considered to be a "breaking change".

### DIFF
--- a/src/xc/xc_libxc.F
+++ b/src/xc/xc_libxc.F
@@ -58,13 +58,11 @@ MODULE xc_libxc
                             xc_f03_func_init, &
                             xc_f03_func_end, &
                             xc_f03_func_info_t, &
-                            xc_f03_functional_get_name, &
                             xc_f03_func_get_info, &
                             xc_f03_func_info_get_family, &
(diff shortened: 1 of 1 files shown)


fstein93:
With your change, the manual cannot be built anymore. At least the functional `HYB_GGA_XC_LC_WPBESOL_WHS` has an invalid character afterwards. Check the artifacts of the Manual generation test of #5503 .

Growl1234:
Thanks I'll have a look.

Growl1234:
Looks like this exposes an issue in Libxc's Fortran name wrapper which exists in both 7.0.0 release and development branch. This PR might need to be reverted.

PR #5511, opened 2026-07-02
Set OpenMPI default binding policy to none via environment variable

Growl1234:
Open MPI's deefault process binding policy can break hybrid MPI+OpenMP parallelism by restricting each MPI rank, together with all of its OpenMP threads, to too few CPU cores. Open MPI does not infer `OMP_NUM_THREADS`, so a binding that is appropriate for MPI-only runs may force all OpenMP threads of a rank to share a single core. This is not user-friendly since CP2K is OpenMP-parallelized. Therefore, default to `--bind-to none` via setting environment to make hybrid parallelism (MPI+OpenMP) happy, which is also [suggested in OpenMPI documentation](https://docs.open-mpi.org/en/main/man-openmpi/man1/mpirun.1.html#quick-summary).

This change does not affect any OpenMPI Docker testers since they already have variables regarding root in both [toolchain tests](https://github.com/cp2k/cp2k/blob/5976d25a45174d28df9c77845b753e0a0585d6d8/tools/docker/scripts/test_regtest.sh#L41-L42) and [Spack tests](https://github.com/cp2k/cp2k/blob/5976d25a45174d28df9c77845b753e0a0585d6d8/make_cp2k.sh#L1527-L1528), as well as `--bind-to none` in [do_regtest.py](https://github.com/cp2k/cp2k/blob/5976d25a45174d28df9c77845b753e0a0585d6d8/tests/do_regtest.py#L70).

### DIFF
--- a/make_cp2k.sh
+++ b/make_cp2k.sh
@@ -1523,10 +1523,11 @@ done
 ln -sf cp2k."${VERSION}" cp2k_shell
 cd "${CP2K_ROOT}" || ${EXIT_CMD} 1
 
-# Allow to run as root with OpenMPI
+# Configure Open MPI environment for local CP2K runs
 if [[ "${MPI_MODE}" == "openmpi" ]]; then
-  OMPI_VARS="export OMPI_ALLOW_RUN_AS_ROOT=1 OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1 OMPI_MCA_plm_rsh_agent=/bin/false"
+  OMPI_VARS="export OMPI_MCA_plm_rsh_agent=/bin/false PRTE_MCA_hwloc_default_binding_policy=none"
   if [[ "${IN_CONTAINER}" == "yes" ]]; then
+    OMPI_VARS="${OMPI_VARS} OMPI_ALLOW_RUN_AS_ROOT=1 OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1"
     OMPI_VARS="${OMPI_VARS} OMPI_MCA_mpi_yield_when_idle=1 OMPI_MCA_btl=self,sm OMPI_MCA_pml=ob1"
   fi
 else
--- a/tools/toolchain/scripts/stage1/install_openmpi.sh
+++ b/tools/toolchain/scripts/stage1/install_openmpi.sh
@@ -105,6 +105,7 @@ export MPICXX="${MPICXX}"
 export MPIFC="${MPIFC}"
 export MPIFORT="${MPIFORT}"
 export MPIF77="${MPIF77}"
+export PRTE_MCA_hwloc_default_binding_policy=none
 EOF
   if [ "${with_openmpi}" != "__SYSTEM__" ]; then
     cat << EOF >> "${BUILDDIR}/setup_openmpi"


mtaillefumier (approved):

make_cp2k.sh
```diff
@@ -148,7 +148,7 @@ HELP="no"
 INSTALL_MESSAGE="NEVER"
 MPI_MODE="mpich"
 if command -v nproc &> /dev/null; then
-  MAX_PROCS=$(nproc)
+  MAX_PROCS=$(nproc --all)
```
mkrack: Why adding the `--alll` flag here which does not respect cgroups when running within containers. The `--`--all` flag was not added here on purpose.

Growl1234: I want to get rid of the limit brought by `OMP_NUM_THREADS`. Maybe there's better way?

mkrack: What do you mean with limit brought by `OMP_NUM_THREADS`? `OMP_NUM_THREADS` can be (is) set when a CP2K job is launched and thus the limit is explicitly chosen on purpose usually. Likewise, one can always unset `OMP_NUM_THREADS` before launching CP2K and OpenMP (given its greedy behaviour by default) will use all available CPU cores.

Growl1234: I intentionally initialized `OMP_NUM_THREADS` to 1 system-wide on my machine (writing it to `~/.bashrc`) in case some testing programs unconditionally uses MPI+OpenMP with all available logical threads, and explicitly set to another value when needed. However, when `OMP_NUM_THREADS` is set, `nproc` will unconditionally be equal to that value.

mkrack: So, that change is needed because of a specific setting in your shell environment? I think `OMP_NUM_THREADS` is not set by default. Is there any container base image which does it?

oschuett:
The `OMP_NUM_THREADS` variable is [always sets](https://github.com/cp2k/cp2k/blob/5976d25a45174d28df9c77845b753e0a0585d6d8/tests/do_regtest.py#L349) when running regtests.

And `OMPI_ALLOW_RUN_AS_ROOT` is clearly a security guardrail that we should not disable for our users.

mkrack:
> And `OMPI_ALLOW_RUN_AS_ROOT` is clearly a security guardrail that we should not disable for our users.

Originally, the definition of these variables  
```
OMPI_VARS="export OMPI_ALLOW_RUN_AS_ROOT=1 OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1 OMPI_MCA_plm_rsh_agent=/bin/false"
```
was restricted to builds within a container which should be fine in my opinion. That was [changed](https://github.com/cp2k/cp2k/commit/ed9ef77eaea4e3c5220e01e6aa886eadfc511544) in #4877 which passed my attention, sorry.

Growl1234:
Although I personally would like to doubt the actual point of rejecting root by OpenMPI as many other implementations e.g. MPICH and IntelMPI do not block it, I'm fine with removing `OMPI_ALLOW_RUN_AS_ROOT`.

mkrack:
> Alhough I personally would like to doubt the actual point of rejecting root by OpenMPI as many other implementations e.g. MPICH and IntelMPI do not block it, I'm fine with removing `OMPI_ALLOW_RUN_AS_ROOT`.

I suggest to recover the original if clause before #4877
```
if [[ "${MPI_MODE}" == "openmpi" ]] && [[ "${IN_CONTAINER}" == "yes" ]]; then
```
It should be safe (and needed) to set `OMPI_ALLOW_RUN_AS_ROOT` running rootless podman containers.

Growl1234:
But that clause also had
```bash
else
  OMPI_VARS="export OMPI_MCA_plm_rsh_agent=/bin/false"
fi
```
which made `OMPI_MCA_plm_rsh_agent` setting not valid for OpenMPI and in contrast expose to other build type (MPICH)...

mkrack (approved):

PR #5513, opened 2026-07-02
Fix perturbative SOC sign bug and molecular SOC path

samalbibek:
Fixing a sign bug in SOC, and now the accuracy of SO splittings in the band structure is 5-20 meV compared to experimental splittings and other benchmark theoretical methods, for alkali halides, elemental crystals, and compound semiconductors.


src/input_cp2k_properties_dft.F
```diff
@@ -2695,13 +2695,32 @@ SUBROUTINE create_soc_section(section)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
-      CALL keyword_create(keyword, __LOCATION__, name="ENERGY_WINDOW", &
-                          description="Apply SOC only for states with eigenvalues in the "// &
-                          "interval $[\varepsilon_\mathrm{VBM}-E_\mathrm{window}/2, "// &
-                          "\varepsilon_\mathrm{CBM}+E_\mathrm{window}/2]$. Might be necessary "// &
+      CALL keyword_create(keyword, __LOCATION__, name="SOC_WINDOW_OCC", &
+                          description="Apply SOC only for states with eigenvalues below VBM "// &
+                          "in the interval $[\varepsilon_\mathrm{VBM}-E_\mathrm{window\_occ}]$, "// &
```
JWilhelm: Please fix the window interval, a proper interval has an upper and lower limit.

src/input_cp2k_properties_dft.F
```diff
@@ -2695,13 +2695,32 @@ SUBROUTINE create_soc_section(section)
...
-                          default_r_val=cp_unit_to_cp2k(value=40.0_dp, unit_str="eV"), &
+                          usage="SOC_WINDOW_OCC 5.0", &
+                          default_r_val=-1.0_dp, &
+                          unit_str="eV")
+      CALL section_add_keyword(section, keyword)
+      CALL keyword_release(keyword)
+
+      CALL keyword_create(keyword, __LOCATION__, name="SOC_WINDOW_VIRT", &
+                          description="Apply SOC only for states with eigenvalues above CBM "// &
+                          "in the interval $[\varepsilon_\mathrm{CBM}+E_\mathrm{window\_virt}]$, "// &
```
JWilhelm: Please fix the window interval, a proper interval has an upper and lower limit.

JWilhelm (changes requested):
Thanks for adding this fix. Please add the following regtests to make sure the SOC code is properly covered by tests:

1) regtest with an occupied window
2) regtest with a virtual window
3) regtest with occupied window and virtual window and fermi smearing

These regtests should be designed in a way that leaving out the window/smearing keyword produces very different results such that an error in the regtests are triggered. (For example by using a very small occupied / virtual window.)

JWilhelm (approved):
Thanks for fixing this, Bibek. Looks good to me now.

PR #5514, opened 2026-07-02
Revert "Libxc: Preserve aliases when registering XC functional sections"

Growl1234:
Reverts cp2k/cp2k#5510

PR #5515, opened 2026-07-02
BSE: update references

MGraml:


MGraml:
I think the sphinx build fails, because libxs's `xc_f03_functional_get_name` doesnt trim the NUL, which then fails in sphinx' parser.
Probably this could be sanitized locally in 
https://github.com/cp2k/cp2k/blob/0a0a847578596eb1f9b5550b7bd075e48d0f1291/src/xc/xc_libxc.F#L246
by using `strip_control_codes` from https://github.com/cp2k/cp2k/blob/0a0a847578596eb1f9b5550b7bd075e48d0f1291/src/common/string_utilities.F#L3440

Should I include the fix directly in this PR?

oschuett:
> Should I include the fix directly in this PR?

Thanks, but this was already fixed via #5514.

Issue #5500, opened 2026-07-01
RFC: Versioning scheme for data files


oschuett:
There were a number of comments on #5376 pointing out that we should not modify existing data sets.

So, let's discuss how we could implement a versioning scheme for basis sets, pseudos, etc.

mkrack:
For the time being, I think the proposal to append `_v0` to the old file and `_v1` to the new file version while creating a symbolic link to the newer file version is just fine. Introducing a hash for each potential and basis set requires some coding and testing effort especially given the way how the datasets are currently parsed by CP2K.

oschuett:
See also #5520, where most of the discussion ended up taking place.

PR #5520, opened 2026-07-03
Append version tag to UZH database files

mkrack:
Apply suggestion from #5376 and RFC #5500


juerghutter:
I cannot make up my mind for this case. Originally, these were supposed to be dynamic libraries, that get
updated when weaknesses/errors are found. The old files (v0 now) has already been updated several times and
this is documented in the file headers. The current update is of medium size (ca.5-7 atoms) and there should be other updates coming. As I mentioned before, I can see the problem for users not checking the file information.
However, this tagging of file versions as v0 and v1 is very arbitrary.
The question is also what to do with the next error fix/update. Do we tag the files only for releases or for every change?

mkrack:
> I cannot make up my mind for this case. Originally, these were supposed to be dynamic libraries, that get updated when weaknesses/errors are found. The old files (v0 now) has already been updated several times and this is documented in the file headers. The current update is of medium size (ca.5-7 atoms) and there should be other updates coming. As I mentioned before, I can see the problem for users not checking the file information. However, this tagging of file versions as v0 and v1 is very arbitrary. The question is also what to do with the next error fix/update. Do we tag the files only for releases or for every change?

I fully agree and I had similar considerations, but the comment in #5376 is also fully valid from a user point of view. Therefore, I think we have to introduce some kind of (well quick-and-dirty) versioning for the database files, because that is better than to do nothing, because, otherwise, it might hit us after the release.
Now, the `_v0` refers to `2026.1` and `_v1` to `2026.2`. Though we are not forced to tightly bind each file version to a release, it makes sense to freeze the latest file version with a new release (only in case it was changed of course). So, you can create after the release `2026.2` a new file version `_v3` which is continuously updated until the next release.
In the meantime, we can work on a better solution.
Any other/better proposal how to address this issue is appreciated.

oschuett:
I agree that we only need to assign version tags on cp2k releases. The individual changes that make up a new version can still be documented in the header of each data file.

Looking at the history of [BASIS_MOLOPT_UZH](https://github.com/cp2k/cp2k/commits/master/data/BASIS_MOLOPT_UZH), we would now be on version `_v6`. So, perhaps it's indeed easier to just reuse the cp2k versions, e.g. `_2026.2`.

For what it's worth, [CODATA](https://www.nist.gov/programs-projects/codata-values-fundamental-physical-constants) also uses calendar versioning despite not releasing every year.

mkrack:
> However, this tagging of file versions as v0 and v1 is very arbitrary.
> The question is also what to do with the next error fix/update. Do we tag the files only for releases or for every change?

@juerghutter Now, the data files are tagged with the cp2k version, is that acceptable?

juerghutter:
Okay with me. This will probably cause the least amount of collateral damage.

juerghutter (approved):
Okay

tgmaxson:
I just wanted to follow up with this and say this feels like an acceptable but maybe not direct solution.  Users using MLIPs should be careful to specify things properly, but this at least gives a way to explicitly name the version.  However, if they pick the default we are back at square one.

Is there a reason CP2K doesn't simply move the "data" folder to an outside repo that can be versioned separately?  That would avoid this issue entirely as users can just always redownload the same "data" folder, unless there is an issue I do not see.  That would match the vasp/gpaw type versioning for example and as long as a given version isnt silently updated, there is no issue.

Sorry to be a pain here, I just encounter this type of issue too frequently.

oschuett:
> However, if they pick the default we are back at square one.

Generally, when people update to a newer of CP2K version, they should be ready for results to change (for the better). 
This can be due to bug fixes in the code or bug fixes in the parameters. If they then decided that they want to selectively stick to the old parameters, they now have the option to do so.

> Is there a reason CP2K doesn't simply move the "data" folder to an outside repo that can be versioned separately? 

While we could decouple the code from the parameters, it would make life harder for all developers and also most users. So, I think we actually landed on a pretty good tradeoff.

tgmaxson:
> > Is there a reason CP2K doesn't simply move the "data" folder to an outside repo that can be versioned separately?
> 
> While we could decouple the code from the parameters, it would make life harder for all developers and also most users. So, I think we actually landed on a pretty good tradeoff.

I am not sure it makes anything significantly harder to have users download a single extra folder though via a specific git tag or a requested specific spack version of the basis/potentials.  I feel like making life slightly easier for users / devs in return for potentially breaking the science of the research on accident is a bad idea.

Keep in mind, the current approach also makes it more difficult to report what is being used long term.  Is every data file going to be versioned going forward?  How should users report their choice of potential/basis.

PR #5527, opened 2026-07-04
[cmake] Cleanup accross the entire build system

mtaillefumier:
- Added cmake linter in the precommit requirements. **Notice**: make_pretty.sh may issue an error because the `cmake linter` is added in this PR but the server does not yet know about it.
- Cleaned various typo that trigger unconditional branches in cmake modules
- Removed code duplication of dftd4 api code
- Cleaned various target definitions
- Fix `WITH_CUDA_PROFILING/CP2K_WITH_GPU_PROFILING`
- Removed `NO_DEFAULT_PATH` as it turns off all conventional `cmake` path detection mechanisms including `CMAKE_PREFIX_PATH`.


oschuett:
I think, you forgot to include `cmake-lint.yaml`. Note also that `pyyaml` has to be [explicitly](https://cmake-format.readthedocs.io/en/latest/configuration.html) added to the [requirements](https://github.com/cp2k/cp2k/blob/master/tools/precommit/requirements.txt).

oschuett:
>  Notice: make_pretty.sh may issue an error because the cmake linter is added in this PR but the server does not yet know about it.

This is actually not a problem because the CI spins up its own [local server](https://github.com/cp2k/cp2k/blob/c24029e76188fffb6310aa742bdc4a476f65122a/tools/docker/scripts/test_precommit.sh#L13) instead of talking to precommit.cp2k.org.

CMakeLists.txt
Growl1234, line 877: Btw I am thinking if we could mark 3.x API of dftd4 library as deprecated here, so that we can remove it in the next release, which can also make D4 code a bit clearer. @jpoto what do you think?

mkrack (approved):

PR #5531, opened 2026-07-05
Docs: Clarify Wannier90 SCF MO reuse requirements

Growl1234:
Clarify that `REUSE_SCF_MOS` currently requires the SCF k-point data to remain within one k-point parallel group.

Update the Wannier90 reuse-validation inputs to use `PARALLEL_GROUP_SIZE 0`, so they exercise the intended reuse path instead of falling back to full-mesh diagonalization. This resolves the `WRONG RESULT`s in [4-MPI-ranks test](https://dashboard.cp2k.org/archive/psmp-4ranks/commit_e6548e293017dc583cc97de7f96a5e9a5868611d.txt).


Growl1234:
Btw, the four-rank failure of `H2_ONEDFT_PBE_CDFT_CI_NGROUPS.inp` appears to be a GauXC issue rather than a CP2K one. In the OneDFT gather path, GauXC uses `rt.comm_rank()` / `rt.comm_size()` from the CDFT subgroup, but still performs collectives on `MPI_COMM_WORLD`:

```cpp
MPI_Gather(..., sendcounts.data(), ..., MPI_COMM_WORLD);
```

For `NGROUPS 2` on four ranks, the count/displacement arrays are sized for two ranks while the collectives involve all four. The surrounding code already uses `rt.comm()` for another gather, so these OneDFT collectives likely need to use the runtime communicator consistently.

PR #5540, opened 2026-07-06
Make the grid gpu code less opaque

mtaillefumier:
- Name the different pointers appropriately instead of `ptr_dev[i]`.


oschuett:
For `diag_cu144_broy.inp` the [performance](https://storage.googleapis.com/cp2k-ci/run-cp2k-perf-cuda-volta-d3745562_report.txt) of `grid_integrate_task_list` increased from  34 to 42 seconds.

mtaillefumier:
I removed the loop flattening. Renaming objects alone can not produce 8 s performance degradation.

oschuett (approved):

PR #5541, opened 2026-07-06
Docs: Clean old-Makefile mentions

### LINKED ISSUE #5539: Docs: Clean up remaining mentions of arch files
An urgent task for the upcoming release that I could not resolve myself in #5522: despite arch files having been obsoleted by CMake in the last release, there are still legacy mentions in `docs/development/regtesting.md` and `docs/technologies/accelerators` (and maybe other places) which should definitely be cleaned up.

_Originally posted by @he-zilong in https://github.com/cp2k/cp2k/discussions/5339#discussioncomment-17542522_

Growl1234:
- [x] **[Accelerators](https://github.com/cp2k/cp2k/tree/master/docs/technologies/accelerators)**: Revise `opencl.md` so it is documented with CMake; remove  `fpga.md` as it is not supported with CMake configuration yet.
- [x] **[Regtesting](https://github.com/cp2k/cp2k/blob/master/docs/development/regtesting.md)**: Rewrite with current workflow.
- [x] **[Coding conventions](https://github.com/cp2k/cp2k/blob/master/docs/development/coding-conventions.md)**: `make doxify` in "Doxygen documentation" seems to be an old-Makefile workflow.
- [x] **[Python bindings](https://github.com/cp2k/cp2k/tree/master/src/start/python)**: Remove them as they are considered outdated and left unused (https://github.com/cp2k/cp2k/issues/5539#issuecomment-4893753157).
- [x] **Source code**: Target hint still uses "ARCH files".

Resolves #5539.


Growl1234:
Should we add things like "This commonly occurs when CP2K is built on a login node but run on compute nodes" in the CPU target hint, as I'm not sure if a generic build can (almost) always trigger this?

he-zilong:
> Should we add things like "This commonly occurs when CP2K is built on a login node but run on compute nodes" in the CPU target hint, as I'm not sure if a generic build can (almost) always trigger this?

I support this idea; "the compiler target... used to build this binary" is not always "generic". [The corresponding faq](https://www.cp2k.org/faq:hint_insufficiently_exploiting_cpu_extensions) may need to be migrated to the documentation as well...

src/openpmd_api.F
```diff
@@ -1216,6 +1216,6 @@ END SUBROUTINE openpmd_c_mesh_set_unit_dimension
          END SUBROUTINE openpmd_mesh_set_unit_dimension
 
 #endif
-         ! In two lines of this file, `make pretty` applies a wrong indentation and keeps it for the rest of this file.
+         ! In two lines of this file, "make pretty" applies a wrong indentation and keeps it for the rest of this file.
```
fstein93: This should be `make_pretty.sh`

fstein93 (commented):
Thank you for the revision

Growl1234:
> [The corresponding faq](https://www.cp2k.org/faq:hint_insufficiently_exploiting_cpu_extensions) may need to be migrated to the documentation as well...

Is the checklist there still relevant now, as CMake already uses `-march=native;-mtune=native` for Release and Debug builds, as well as `-mtune=generic` for Generic builds?

src/openpmd_api.F
Growl1234: I rechecked this and found that `make_pretty.sh` passes without these two comment lines, possibly because later changes altered the formatting conditions here. So, just remove them?

he-zilong:
> > [The corresponding faq](https://www.cp2k.org/faq:hint_insufficiently_exploiting_cpu_extensions) may need to be migrated to the documentation as well...
> 
> Is the checklist there still relevant now, as CMake already uses `-march=native;-mtune=native` for Release and Debug builds, as well as `-mtune=generic` for Generic builds?

On second thought, a full-fledged explanation about what the default values are and how they can be adjusted is better as a section in [Build from Source](https://manual.cp2k.org/trunk/getting-started/build-from-source.html). (I presume it is about checking `gfortran --target-help` for `Known valid arguments for -march= option` and `Known valid arguments for -mtune= option`, then applying to toolchain `--target-cpu` option and the following line of `cmake/CompilerConfiguration.cmake`?)
https://github.com/cp2k/cp2k/blob/fe787632bd21d8bb579232615b8aeebc7927b9bc/cmake/CompilerConfiguration.cmake#L55

After this addition, it would be easier to explain the messages about "The compiler target flags..." (and perhaps "Setting real_kernel for ELPA failed...") by linking to the section. Then the old FAQ can be considered obsoleted.

Besides this change, I think I can give my approval here as it is not critical anyways...

he-zilong (approved):

oschuett (approved):

PR #5545, opened 2026-07-07
Sync CI dockerfiles (see #5462)

mkrack:


mkrack:
CSCS HPC resources are still down. Test jobs reached maximum waiting time in queue.

PR #5546, opened 2026-07-07
make_cp2k.sh: Use Spack v1.2.0

Growl1234:
When `--use_externals` is enabled, Spack 1.1.1 can register the same Python installation more than once during external discovery. In particular, the Python installation used by the virtual environment and the system Python are both detected as equivalent external candidates. This leaves `meson` with multiple matching `python` dependencies and causes concretization to fail.

Spack 1.2.0 includes upstream PR https://github.com/spack/spack/pull/52100. External detection now keeps the first occurrence of a spec, warns about later identical detections at other prefixes, and does not add the duplicates to `packages.yaml`. This update therefore avoids the duplicate-Python external ambiguity during `make_cp2k.sh --use_externals` runs, without changing the package constraints or dependency recipes.

However, keep using old installer in `spack install` stage, because new installer of Spack v1.2 can't work correctly with Python 3.14 (see https://github.com/spack/spack/issues/52573).

### DIFF
--- a/make_cp2k.sh
+++ b/make_cp2k.sh
@@ -1159,7 +1159,7 @@ if [[ ! -f "${SPACK_BUILD_PATH}/BUILD_DEPENDENCIES_COMPLETED" ]]; then
   spack compiler list
 
   # Retrieve the newest compiler version found by spack
-  GCC_VERSION_NEWEST="$(spack compilers | awk '/gcc/ {print $2}' | sort -V | tail -n 1)"
+  GCC_VERSION_NEWEST="$(spack compilers | sed -nE 's/.*gcc@([0-9]+(\.[0-9]+)*).*/\1/p' | sort -V | tail -n 1)"
   echo "The newest GCC compiler version found by spack is ${GCC_VERSION_NEWEST}"
   GCC_VERSION_NEWEST="$(echo "${GCC_VERSION_NEWEST}" | sed -E -e 's/.*@([0-9]+).*/\1/' | cut -d. -f1)"
 
--- a/make_cp2k.sh
+++ b/make_cp2k.sh
@@ -914,7 +914,7 @@ echo ""
 ### Build CP2K dependencies with Spack if needed or requested ###
 
 # Spack version
-export SPACK_VERSION="${SPACK_VERSION:-1.2.0}"
+export SPACK_VERSION="${SPACK_VERSION:-1.2.1}"
 export SPACK_BUILD_PATH="${BUILD_PATH}/spack"
 export SPACK_ROOT="${SPACK_BUILD_PATH}/spack"
 
--- a/make_cp2k.sh
+++ b/make_cp2k.sh
@@ -914,7 +914,7 @@ echo ""
 ### Build CP2K dependencies with Spack if needed or requested ###
 
 # Spack version
-export SPACK_VERSION="${SPACK_VERSION:-1.1.1}"
+export SPACK_VERSION="${SPACK_VERSION:-1.2.0}"
 export SPACK_BUILD_PATH="${BUILD_PATH}/spack"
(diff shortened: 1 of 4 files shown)


mtaillefumier (approved):

mkrack (commented):
A build without the `-ue` flag fails.

Growl1234:
> A build without the `-ue` flag fails.

Flag `--no-user-config` and `--no-system-config` are no longer valid options in Spack 1.2.0. I dropped them as there has been proper configuration globally:
https://github.com/cp2k/cp2k/blob/1ae7fa728c0def7f3c16edff3855fa3ebfcd4b93/make_cp2k.sh#L921-L923

mkrack:
> However, keep using old installer in `spack install` stage, because new installer of Spack v1.2 can't work correctly with Python 3.14 (see [spack/spack#52573](https://github.com/spack/spack/issues/52573)).

Maybe, it is better to postpone the update to spack v1.2.0 after the release when the dust has settled and the new spack installer will work?

Growl1234:
My main concern for Spack v1.1.1 is the failure regarding duplicate Python detection when using local cache, which always happens when `meson` is detected in system. So, though this can be avoided through not using `-ue`, I still think the version update makes sense.

With current workaround, `make_cp2k.sh` leverages the fixes in Spack v1.2.0 to prevent some avoidable issues when parsing package configuration, and uses old-installer (default in version <= v1.1.1) for package installing.

mkrack:
Using a mixture of Spack versions 1.1.1 and 1.2.0 needs to be carefully tested, as this could introduce new, yet unknown problems, while the problems of version 1.1.1 are already better known.

Growl1234:
I think my previous wording might be misleading; this is not a mixture of Spack 1.1.1 and 1.2.0. The bootstrap, concretization, external detection, package repository, and installation environment are all from Spack 1.2.0. The only exception is that Spack 1.2.0 is explicitly configured to use its legacy installer implementation via `config:installer:old`; no Spack 1.1.1 code or state is used. Using old installer is a reasonable workaround for Python 3.14/forkserver hang in the 1.2.0 release, while retaining the upstream 1.2.0 fix for the duplicate Python external detection issue.

That said, I agree this need careful check. I didn't encounter problems with current configuration on my machine (both physical environment on Rocky 10.2 and Docker container as Ubuntu 26.04), but maybe there are some other cases sensitive to this?

mkrack (approved):

Growl1234:
Spack just released their [v1.2.1](https://github.com/spack/spack/releases/tag/v1.2.1) 10 hours ago...

Edit: Well the build fails because libint relocated with new installer cannot find its Boost and Eigen3 dependency... So I'll still need to roll back to legacy installer.

Growl1234:
> Edit: Well the build fails because libint relocated with new installer cannot find its Boost and Eigen3 dependency... So I'll still need to roll back to legacy installer.

I found the reason. `libint` declares `eigen` as a build-only dependency, so Eigen is not linked into the environment view generated by new Spack installer. However, the installed `libint2-config.cmake` still tries to locate Eigen when consumed by downstream CMake projects. Since the recorded build-time Eigen config path is not usable, the fallback lookup fails because Eigen is absent from the view.

mkrack:
> I found the reason. `libint` declares `eigen` as a build-only dependency, so Eigen is not linked into the environment view generated by new Spack installer. However, the installed `libint2-config.cmake` still tries to locate Eigen when consumed by downstream CMake projects. Since the recorded build-time Eigen config path is not usable, the fallback lookup fails because Eigen is absent from the view.

We use a [local spack recipe for libint](https://github.com/cp2k/cp2k/blob/master/tools/spack/spack_repo/cp2k_dev/packages/libint/package.py#L51-L52). You can try to remove the `type="build"` options from the dependencies `boost` and `eigen`.

PR #5549, opened 2026-07-07
 Fix CMake toolchain support for MKL and LIBXSMM with Intel builds

RTChou67:
## Summary

  This PR fixes two CMake/toolchain integration issues seen with Intel oneAPI builds:

  - passes the resolved MKL configuration through the generated CMake flags
  - fixes the exported LIBXSMM CMake configuration so CP2K sees a configured LIBXSMM build instead of reporting `LIBXSMM_VERSION: unconfigured`

## Details
The toolchain now propagates MKL-related settings into CMake, including the MKL root and CP2K BLAS/ScaLAPACK vendor options. The LIBXSMM install step also adjusts the generated CMake package metadata so downstream CP2K configuration receives the configured LIBXSMM compile definition. This prevents the installed CP2K binary from being linked against a LIBXSMM configuration that reports itself as unconfigured at runtime.

### DIFF
--- a/tools/toolchain/build_cp2k.sh
+++ b/tools/toolchain/build_cp2k.sh
@@ -197,6 +197,9 @@ fi
 if [ ${DEBUG_BUILD} == "__TRUE__" ]; then
   CMAKE_OPTIONS+=" -DCMAKE_BUILD_TYPE=Debug"
 fi
+if [ "${math_mode}" = "mkl" ]; then
+  CMAKE_OPTIONS+=" -DCP2K_BLAS_VENDOR=MKL -DCP2K_BLAS_THREADING=sequential"
+fi
 if [ -n "$(grep -- "--install-all" "${TOOLCHAIN_ROOTDIR}/toolchain_settings")" ]; then
   CMAKE_OPTIONS+=" -DCP2K_USE_EVERYTHING=ON -DCP2K_USE_DLAF=OFF -DCP2K_USE_PEXSI=OFF -DCP2K_USE_OPENPMD=OFF"
   # If MPI is disabled, set "CP2K_USE_MPI" to "OFF"
--- a/tools/toolchain/scripts/stage4/install_libxsmm.sh
+++ b/tools/toolchain/scripts/stage4/install_libxsmm.sh
@@ -47,6 +47,12 @@ EOF
         -DLIBXSMM_FORTRAN=ON \
         .. > configure.log 2>&1 || tail_excerpt configure.log
       make install -j $(get_nprocs) > make.log 2>&1 || tail_excerpt make.log
+      sed -i \
+        's/INTERFACE_COMPILE_DEFINITIONS "LIBXSMM_DEFAULT_CONFIG"/INTERFACE_COMPILE_DEFINITIONS "LIBXSMM_CONFIGURED"/' \
+        "${pkg_install_dir}/lib/cmake/libxsmm/libxsmmTargets.cmake"
+      if grep -R -q "LIBXSMM_DEFAULT_CONFIG" "${pkg_install_dir}/lib/cmake/libxsmm"; then
+        report_error ${LINENO} "Installed libxsmm target still exports LIBXSMM_DEFAULT_CONFIG."
+      fi
       cd ..
       write_checksums "${install_lock_file}" "${SCRIPT_DIR}/stage4/$(basename ${SCRIPT_NAME})"
     fi


tools/toolchain/build_cp2k.sh
```diff
@@ -185,12 +185,22 @@ source "${TOOLCHAIN_INSTALL_DIR}/toolchain.conf"
...
 if [[ ${CMAKE_INSTALL_PREFIX} == ${CP2K_ROOT}/* ]]; then
   CMAKE_OPTIONS+=" -DCP2K_DATA_DIR=${CP2K_ROOT}/data"
 fi
 if [ ${DEBUG_BUILD} == "__TRUE__" ]; then
   CMAKE_OPTIONS+=" -DCMAKE_BUILD_TYPE=Debug"
 fi
+if [ "${math_mode}" = "mkl" ]; then
+  if [ -z "${MKLROOT}" ]; then
+    report_error ${LINENO} "math_mode is mkl, but MKLROOT is not set by ${TOOLCHAIN_INSTALL_DIR}/setup."
+  fi
```
Growl1234, line 198: If math mode is MKL, `MKLROOT` should already be set in the system environment, as toolchain does detect them for parsing math environment. This guard might not necessary here in my view.

tools/toolchain/build_cp2k.sh
```diff
@@ -185,12 +185,22 @@ source "${TOOLCHAIN_INSTALL_DIR}/toolchain.conf"
...
+  CMAKE_OPTIONS+=" -DMKL_ROOT=${MKLROOT} -DCP2K_BLAS_VENDOR=MKL -DCP2K_BLAS_THREADING=sequential"
+  if [ "${mpi_mode}" != "no" ] && [ "${MKL_SCALAPACK}" = "yes" ]; then
+    CMAKE_OPTIONS+=" -DCP2K_SCALAPACK_VENDOR=MKL"
```
Growl1234: If this needs to be manually set, I will think that it's more likely an issue on the CMake side.

tools/toolchain/build_cp2k.sh
```diff
@@ -185,12 +185,22 @@ source "${TOOLCHAIN_INSTALL_DIR}/toolchain.conf"
 
 # Generate cmake options for compiling cp2k
 CMAKE_OPTIONS="-DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX} -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}"
+CMAKE_UNSET_OPTIONS=(-U "CP2K_USE_*")
```
Growl1234, line 188: What's the point of this?

Growl1234 (commented):
Note that CP2K cannot work with Intel oneAPI compilers, so the segfault will be rather normal when you compile CP2K with ifx (especially versions >=2025.3). Please use GCC with CP2K _always_.

The output regarding LIBXSMM actually makes no sense; the error is usually not related to it. You should run `gdb` to see what really happens. (That said, it's worth mentioning the output in the LIBXSMM repo.)

tools/toolchain/build_cp2k.sh
RTChou67: Yes, `MKLROOT` is set. However, without modifying the build script, it would run into errors. Anyway, as you said, CP2K cannot work with oneAPI compilers, so maybe these are indeed not necessary. 

> Shared libs: ON
>   -- Found MPI: TRUE (found version "4.1") found components: C CXX Fortran
>   -- Found MPI: TRUE (found version "4.1") found components: CXX C Fortran
>   -- Could NOT find MKL (missing: CP2K_MKL_INCLUDE_DIRS)
>   -- Checking for module 'scalapack'
>   --   Package 'scalapack' not found
>   CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:290 (message):
>     Could NOT find SCALAPACK (missing: CP2K_SCALAPACK_LINK_LIBRARIES)
>   Call Stack (most recent call first):
>     /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:654 (_FPHSA_FAILURE_MESSAGE)
>     cmake/modules/FindSCALAPACK.cmake:118 (find_package_handle_standard_args)
>     CMakeLists.txt:450 (find_package)

tools/toolchain/build_cp2k.sh
RTChou67: This is meant to avoid stale `CP2K_USE_*` cache entries when reconfiguring an existing build directory without `--clean`. For example, in my case, a previous CP2K build with GNU compilers may survive even when I swapped to oneAPI compilers. Other examples might also occur.

tools/toolchain/build_cp2k.sh
RTChou67: Well, I would agree. `MKL_ROOT` is redundant if `MKLROOT` is set. For `CP2K_BLAS_VENDOR=MKL`, my intent was not to work around CMake detection, but to propagate the toolchain's explicit `math_mode=mkl` choice into the CP2K CMake configure step. I agree that if ScaLAPACK auto-detection fails after BLAS has selected MKL, that should be fixed in CMake rather than forced here.

tools/toolchain/build_cp2k.sh
Growl1234: > -- Could NOT find MKL (missing: CP2K_MKL_INCLUDE_DIRS)

I think this is the reason; CMake fails to find MKL so it goes to general ScaLAPACK discovery? I need to investigate deeper into this...

tools/toolchain/build_cp2k.sh
Growl1234: I understand the intention. However, if the generated `CP2K_USE_*` settings actually change, the subsequent configure-and-build will in practice recompile nearly all of CP2K anyway; at most the FYPP generation may be reused. Therefore this does not provide a meaningful incremental-build benefit over `--clean`.

For compiler changes in particular, this does not help, since it does not reset CMake's cached `CMAKE_<LANG>_COMPILER` entries. Such a change still requires a clean build directory.

Therefore, I don't think we should unconditionally delete all user-visible `CP2K_USE_*` cache entries on every normal reconfiguration. If the rebuilding with different toolchain configuration is really annoying, we can switch to another design that deletes build directory anyways unless `--rebuild-only` is required.

RTChou67:
Agreed. The LIBXSMM version output was a separate toolchain/package issue: the generated CMake target exported `LIBXSMM_DEFAULT_CONFIG`, which made `cp2k.psmp --version` report `LIBXSMM_VERSION: unconfigured`. After fixing that target metadata, the version is reported correctly. However, runtime failures still occur, which i checked with gdb, finding that these do not point to LIBXSMM. For example:
  - `QS/regtest-sasccs/H2_sasccs.inp` crashes in `pw_poisson_solve`, called from `qs_sccs::sccs` at `src/qs_sccs.F:692`.
  - `xTB/regtest-gfn0/ch2o_3.inp` crashes in `pw_poisson_solve`, called from `spme::spme_potential` at `src/spme.F:515`.
  In both cases, the caller does not pass the optional `ehartree` argument, but the generated/specialized Poisson routine enters the `PRESENT(ehartree)` branch and writes through an inaccessible/null optional argument. So this looks like a separate CP2K-side issue. I have not fixed these issues.

Growl1234:
What version of oneAPI are you using? Intel oneAPI version 2025.3 or later have regression bugs mainly relates to ifx and Intel MPI. Moreover ifx cannot work correctly with libint (at least CP2K's HFX and libint wrappers)

tools/toolchain/build_cp2k.sh
RTChou67: I think you are right. I'll drop the unconditional `-U "CP2K_USE_*"` from this PR.

Growl1234:
`PRESENT(ehartree)` is indeed a point, but  you will soon find there will be more segfaults in more places, most of which has been confirmed to be due to regression from ifx or Intel MPI side. Since the code is mainly tested on GCC (we have only four CI tests for Intel compilers with limitations), I think we don't have too much motivation to work on this...

That said, it would be great if you are willing to help take a deeper look into this.

RTChou67:
I am using oneAPI 2025.3.2:
  - ifx 2025.3.2 20260112
  - icx 2025.3.2 20260112
  - Intel MPI 2021.17 Build 20251215
  - MKL 2025.3
This matches the version range you mentioned, so the SCCS/SPME crash may indeed be related to the known ifx / Intel MPI regression rather than LIBXSMM. HFX errors with oneAPI does exist, as you've said. Maybe I'll dig deeper into the oneAPI stuff before submitting PR. It's my first time contributing to a project, so maybe I'm too much in a hurry. 

> >>> /home/rtchou/Study/CP2K-CD/cp2k/regtesting/TEST-2026-07-08_01-06-46/QS/regtest-hfx
>     H2O-hfx-1.inp                                                                         -75.88215405           OK (   3.43 sec)
>     H2O-hfx-2.inp                                                                         -75.88215405           OK (   3.54 sec)
>     H2O-hfx-3.inp                                                                          -75.9481666           OK (   3.05 sec)
>     CH-hfx-md.inp                                                                         -38.26470346           OK (   5.01 sec)
>     CH-hfx-md-2.inp                                                                       -38.26470346           OK (   5.23 sec)
>     H2O_pw.inp                                                                            -3.790936332           OK (   4.54 sec)
>     H2O_pw_a.inp                                                                          -3.859607725           OK (   4.91 sec)
>     H2O_pw_b.inp                                                                          -1.922514227           OK (   4.41 sec)
>     4H2O-disk.inp                                                                         -303.0075049 WRONG RESULT (   6.40 sec)
>     4H2O-mix-disk-ram.inp                                                                 -303.0075049 WRONG RESULT (  16.06 sec)
>     4H2O-mix-disk-ram-on-the-fly.inp                                                      -303.0075049 WRONG RESULT (   6.14 sec)
>     H2-hfx-rtp.inp                                                                       -0.7669353208           OK (   4.67 sec)
>     H2-hfx-emd.inp                                                                       -0.7669356591           OK (   5.11 sec)
>     H2O-hfx-emd.inp                                                                       -16.87592804           OK (  12.04 sec)
>     H2O-hfx-emd-restart.inp                                                               -16.87592804           OK (   6.86 sec)
>     H2O-hfx-emd-ngs.inp                                                                   -15.85810933           OK (   7.58 sec)
>     H2O-hfx-atprop.inp                                                                    -16.95511513           OK (   3.76 sec)
>     H2O-hfx-ls-emd.inp                                                                    -16.83980279           OK (  10.08 sec)
>     H2O-hfx-ls-rtp.inp                                                                    -16.83980778           OK (   8.68 sec)
>     H2O-hfx-ls-rtp-bch.inp                                                                -16.83980778           OK (   7.13 sec)
>     H2O-hfx-ls-emd-bch.inp                                                                 -16.8398024           OK (   8.59 sec)
>     H2O-hfx-ls-emd-ngs.inp                                                                -16.78607911           OK (   9.78 sec)
>     H2-full.inp                                                                           -8.522369091           OK (   5.45 sec)
>     H2-ADMM-full.inp                                                                      -8.855712553           OK (   7.48 sec)
> <<< /home/rtchou/Study/CP2K-CD/cp2k/regtesting/TEST-2026-07-08_01-06-46/QS/regtest-hfx (1 of 1) done in 159.97 sec

tools/toolchain/build_cp2k.sh
he-zilong: > This is meant to avoid stale `CP2K_USE_*` cache entries when reconfiguring an existing build directory without `--clean`. For example, in my case, a previous CP2K build with GNU compilers may survive even when I swapped to oneAPI compilers. Other examples might also occur.

Hold on, it is possible to switch compilers only for CP2K but not the dependencies? I always thought that toolchain and CP2K must be consistent in terms of compiler (i.e. GNU toolchain for GNU CP2K, or Intel toolchain for Intel CP2K, but not GNU toolchain for Intel CP2K)...

Growl1234: > it is possible to switch compilers only for CP2K but not the dependencies?

This most likely results in errors because `.mod` files produced by different compilers are usually not ABI compatible.

oschuett:
This PR hasn't moved in a month. @RTChou67, are you still working on it?

DCM-Uni-Paderborn:
Updated this PR to the current `master` and addressed the review feedback:

- dropped the unconditional `-U "CP2K_USE_*"`; normal configuration now recreates the build directory anyway (since #5560)
- removed the redundant `MKLROOT` guard while keeping the explicit propagation of `math_mode=mkl` to `CP2K_BLAS_VENDOR=MKL` with sequential threading
- retained the LIBXSMM export fix and verified it against the current toolchain version LIBXSMM 2.1.0: the installed CMake target exports `LIBXSMM_DEFAULT_CONFIG` before the adjustment, while a CMake consumer reports version `2.1.0-2` afterward

Validation: targeted CP2K precommit checks for both modified shell scripts, `bash -n`, and a standalone CMake consumer using the adjusted LIBXSMM 2.1.0 package.

RTChou67:
> This PR hasn't moved in a month. @RTChou67, are you still working on it?

sorry that I replied late. there are my other works going on, so currently I'm not working on it.

Growl1234:
- The libxsmm configuration export is not the real issue causing segfault, so it will not resolve anything from CP2K side; would suggest report this issue upstream first. From my testing, changing `LIBXSMM_DEFAULT_CONFIG` to `LIBXSMM_CONFIGURED` in CMake will cause compilation error of libxsmm.
- I don't see the point of enforcing sequential threading as it does not seem to cause issues.

Growl1234:
@RTChou67 I investigated the output itself a bit deeper, and you can see https://github.com/libxsmm/libxsmm/issues/1088 and https://github.com/hfp/libxs/issues/30 for information. Note it is not referring to segfault itself; it's just libxsmm/libxs' behavior has hidden the real segfault info.

RTChou67:
@Growl1234 Many thanks for your investigation.

PR #5554, opened 2026-07-08
elpa 2026.02.001 -> 2026.02.002 (spack)

mkrack:


RMeli (approved):
LGTM.

It would be nice to have these updates upstream. Now that the packages and Spack itself are in different repositories it shouldn't be too bad to bump the `stack-packages` hash (we pin most versions anyways).

mkrack:
> LGTM.

Thanks for the quick check.

> It would be nice to have these updates upstream. 

Sure, I will do that. Unfortunately, the review process on the upstream spack-packages repository is quite slow and there is no formal mechanism (admin groups) like on conda-forge which one can address/remind as a contributor. I have several open PRs for more than a week now. I guess the admin team is overwhelmed by the number of PRs. 

> Now that the packages and Spack itself are in different repositories it shouldn't be too bad to bump the `stack-packages` hash (we pin most versions anyways).

The [current stack packages hash](https://github.com/cp2k/cp2k/blob/master/tools/spack/cp2k_deps_p.yaml#L185-L187) (if understand you correctly and that is what you mean) is from yesterday?!

RMeli:
> Unfortunately, the review process on the upstream spack-packages repository is quite slow and there is no formal mechanism (admin groups) like on conda-forge which one can address/remind as a contributor.

Fair point. I have merge rights but missed your PRs. Feel free to add me as reviewer of the upstream Spack PR when you need a review/merge.

> The [current stack packages hash](https://github.com/cp2k/cp2k/blob/master/tools/spack/cp2k_deps_p.yaml#L185-L187) (if understand you correctly and that is what you mean) is from yesterday?!

I did not check the current hash, sorry for the confusion. It was just to say that there should now be little drawbacks in keeping that hash very up-to-date and so if packages are merged quickly upstream (big IF!) the contributions could go there directly (and in CP2K one would only have to update the hash and pinned version).

PR #5555, opened 2026-07-08
Test native SKALA chunk routing across SCF iterations

jpoto:
## Summary

- Rebase the branch onto the current `master`.
- Exercise native SKALA atom-chunk routing across two SCF iterations in the existing H2 GPW stress test.
- Update the energy reference for the second SCF step.

## Routing semantics

No additional chunk-to-receive map is needed. `chunk_return_positions` packs canonical chunk rows into the original receive order before the reverse all-to-all. The returned gradient buffer is therefore in the source rank's send-buffer order and must be scattered through `route_send_local_rows(point_pos)`.

The previous implementation iterated over owner-side `chunk_row` while indexing source-side `route_send_local_rows`. These are different rank-local index spaces; with an imbalanced two-rank decomposition this caused the reported out-of-bounds access.

## Validation

Built from the current `master` with Debug checks, MPI, GauXC, and LibTorch, then tested on CPU:

| Test | MPI ranks | `E_total` [Ha] | Result |
| --- | ---: | ---: | --- |
| GPW atom chunks, two SCF steps | 1 | -1.04111543882624 | pass |
| GPW atom chunks, two SCF steps | 2 | -1.04111543840281 | pass |
| GAPW+GTH atom chunks | 2 | -0.97019845240713 | pass |

The one- and two-rank GPW energies differ by `4.2e-10` Ha. Forces and analytical stress are consistent across both decompositions, and no bounds error occurs. The focused precommit checks pass.


Growl1234:
Is task with more MPI ranks tested?

oschuett:
Looking at the [latest test results](https://storage.googleapis.com/cp2k-ci/run-cp2k-pdbg-9417834b_report.txt) it seems there is still a bug. Additionally, there is now a merge conflict.

DCM-Uni-Paderborn:
The branch is now rebased onto current master and the conflict is resolved. The pdbg failure came from mixing owner-side `chunk_row` indices with source-side `route_send_local_rows`; the attempted reverse map therefore created the out-of-bounds access. Current master already preserves the correct source send-buffer order, so I removed that source change and retained a two-step SCF regression that exercises route reuse. A fresh Debug/MPI build passes the GPW case with 1 and 2 ranks and the GAPW+GTH chunk case with 2 ranks.

PR #5559, opened 2026-07-09
Fix unix socket prefix - Make UNIX socket prefix configurable via DRIVER%PREFIX

### LINKED ISSUE #3884: Unix socket interface with i-PI not working in versions >= 2024.1
I tried to run simulations with i-PI and using CP2K as the force client and ran into problems with the newest versions. For 2024.1, I tried both a local cluster parallel build module and the available precompiled binary, for 2025.1 only the precompiled binary. All lead to the same error
`Error opening UNIX socket: path unavailable, or already existing: No such file or directory`; running the same input files with CP2K version 9.1 works smoothly.

I put together a MWE of a single water molecule simulation at a DFTB level that reproduces the error. The CP2K input is as follows:
```
&GLOBAL
  PROJECT water
  PREFERRED_FFT_LIBRARY FFTW
  FFTW_PLAN_TYPE MEASURE
  RUN_TYPE DRIVER
  PRINT_LEVEL LOW
&END GLOBAL

&FORCE_EVAL
  &DFT
    CHARGE 0
    &QS
      METHOD DFTB
      &DFTB
        DIAGONAL_DFTB3
        HB_SR_GAMMA
        &PARAMETER
          PARAM_FILE_PATH 3ob-3-1
          PARAM_FILE_NAME scc_parameter
          HB_SR_PARAM 4.0
        &END PARAMETER
      &END DFTB
      EXTRAPOLATION_ORDER 3
    &END QS
    &SCF
      MAX_SCF 200
      EPS_SCF 1.0E-7
      SCF_GUESS ATOMIC
      &MIXING
      &END
      &PRINT
        &RESTART OFF
        &END RESTART
      &END PRINT
    &END SCF
    &POISSON
      PERIODIC NONE
      POISSON_SOLVER WAVELET
    &END POISSON
  &END DFT
  &SUBSYS
    &TOPOLOGY
      COORD_FILE_NAME pos.xyz
      COORD_FILE_FORMAT XYZ
    &END TOPOLOGY
    &CELL
      ABC 50.0 50.0 50.0
      PERIODIC NONE
    &END CELL
    &KIND H
      DFTB3_PARAM -0.1857
    &END KIND
    &KIND O
      DFTB3_PARAM -0.1575
    &END KIND
  &END SUBSYS
&END FORCE_EVAL

&MOTION
  &DRIVER
    UNIX TRUE
    HOST myunixsocket
    PORT 31415
  &END DRIVER
&END MOTION
``` 

and the corresponding i-PI input is
```
<simulation verbosity='high'>

  <output prefix='water-test'>
    <properties stride='1' filename='out'>
      [step, time{picosecond}, conserved, temperature{kelvin}, potential]
    </properties>
    <properties stride='100' filename='ke'>
      [step, kinetic_td, kinetic_cv, kinetic_td(H), kinetic_cv(H), kinetic_td(O), kinetic_cv(O)]
    </properties>
    <checkpoint overwrite='false' stride='10000' filename='checkpoint'/>
    <trajectory cell_units='angstrom' stride='10' filename='positions'> positions{angstrom} </trajectory>
    <trajectory cell_units='angstrom' stride='100' filename='velocities'> velocities </trajectory>
    <trajectory cell_units='angstrom' stride='10' filename='positions-centroid'> x_centroid{angstrom} </trajectory>
    <trajectory cell_units='angstrom' stride='100' filename='velocities-centroid'> v_centroid </trajectory>
    <trajectory cell_units='angstrom' stride='100' filename='ke-atoms'> kinetic_cv </trajectory>
    <trajectory cell_units='angstrom' stride='100' filename='ke-atoms-od'> kinetic_od </trajectory>
    <trajectory cell_units='angstrom' stride='100' filename='forces'> forces </trajectory>
  </output>

  <total_steps>20000</total_steps>

  <ffsocket mode='unix' name='dftb'>
    <address>myunixsocket</address>
    <timeout>1500</timeout>
    <port>31415</port>
  </ffsocket>

  <system>
    <initialize nbeads='1'>
      <file mode="xyz" units="angstrom">pos.xyz</file>
      <cell units="angstrom">[50., 0.0, 0.0,
              0.0, 50., 0.0,
              0.0, 0.0, 50.]</cell>
      <velocities mode='thermal' units='kelvin'>300</velocities>
    </initialize>
    <forces>
      <force forcefield='dftb'  nbeads='1'></force>
    </forces>
    <motion mode='dynamics'>
      <dynamics mode='nvt'>
        <thermostat mode='pile_l'>
          <tau units="femtosecond">100</tau>
        </thermostat>
        <timestep units="femtosecond">0.5</timestep>
      </dynamics>
    </motion>
    <ensemble>
      <temperature units='kelvin'>300.0</temperature>
    </ensemble>
  </system>
</simulation>
```

jhaens:
## Summary

Fixes [#3884](https://github.com/cp2k/cp2k/issues/3884).

The UNIX-domain socket path used in i-PI driver mode was built in
`sockets.c` by hardcoding a `/tmp/qiskit_<host>` prefix ([line 108](https://github.com/cp2k/cp2k/blob/1b0cc9d292fdb61b71134cce16fe056c60892ae1/src/sockets.c#L108)). This breaks
interoperability with i-PI itself, which expects `/tmp/ipi_<host>`,
and with any other driver that uses a different naming convention -
CP2K and the peer process would both start and appear connected, but
would get stuck once the actual calculation started because they were
listening/connecting on different socket paths.

This PR:
- Removes the hardcoded prefix from `open_connect_socket()` in
  `src/sockets.c`. It now takes `host` as the full socket path,
  consistent with `open_bind_socket()`, which already worked this way.
- Adds a new `PREFIX` keyword to the `&MOTION/&DRIVER` input section
  (`src/start/input_cp2k_motion.F`), defaulting to `"ipi"` so existing
  input files keep working unchanged.
- Builds the full socket path (`/tmp/<PREFIX>_<HOST>`) in
  `src/ipi_driver.F` before handing it to the C socket layer. `PREFIX`
  is only meaningful when `UNIX` is true; it's ignored for INET
  sockets.
- Allows dynamic adjustment of the socket path used to connect to other drivers, enabling CP2K to interoperate with drivers that use different socket path prefixes.

Example input with the new `PREFIX` keyword:
```
&MOTION
  &DRIVER
    HOST localhost
    PORT 31423
    UNIX TRUE
    PREFIX ipi
  &END DRIVER
&END MOTION
```

## Test plan

- [x] Verified `src/sockets.c` compiles standalone.
- [x] Manually exercised the UNIX-socket driver path against an i-PI
      server with `PREFIX ipi` (default) and confirmed the socket path
      matches i-PI's own `/tmp/ipi_<host>` convention.
- [ ] No automated regtest was added: `tests/i-PI/ipi_client.inp` and
      `ipi_server.inp` are not wired into `TEST_FILES` since exercising
      this path requires an external driver process running
      alongside CP2K.

### DIFF
--- a/src/ipi_driver.F
+++ b/src/ipi_driver.F
@@ -87,7 +87,7 @@ CONTAINS
 #else
       INTEGER, PARAMETER                       :: MSGLEN = 12
 
-      CHARACTER(len=default_path_length)       :: c_hostname, drv_hostname
+      CHARACTER(len=default_path_length)       :: c_hostname, drv_hostname, drv_prefix
       CHARACTER(LEN=default_string_length)     :: header
       INTEGER                                  :: drv_port, handle, i_drv_unix, &
                                                   idir, ii, inet, ip, iwait, &
@@ -132,6 +132,7 @@ CONTAINS
       CALL section_vals_val_get(drv_section, "HOST", c_val=drv_hostname)
       CALL section_vals_val_get(drv_section, "PORT", i_val=drv_port)
       CALL section_vals_val_get(drv_section, "UNIX", l_val=drv_unix)
+      CALL section_vals_val_get(drv_section, "PREFIX", c_val=drv_prefix)
       CALL section_vals_val_get(drv_section, "SLEEP_TIME", r_val=sleeptime)
       CPASSERT(sleeptime >= 0)
 
@@ -145,7 +146,14 @@ CONTAINS
          WRITE (output_unit, *) "@ INPUT DATA: ", TRIM(drv_hostname), drv_port, drv_unix
       END IF
 
-      c_hostname = TRIM(drv_hostname)//C_NULL_CHAR
+      IF (drv_unix) THEN
+         ! for UNIX sockets, HOST names the socket file which lives at
+         ! /tmp/<PREFIX>_<HOST> (PREFIX must match the peer's convention,
+         ! e.g. "ipi" for i-PI)
+         c_hostname = "/tmp/"//TRIM(drv_prefix)//"_"//TRIM(drv_hostname)//C_NULL_CHAR
+      ELSE
+         c_hostname = TRIM(drv_hostname)//C_NULL_CHAR
+      END IF
       IF (ionode) CALL open_connect_socket(socket, i_drv_unix, drv_port, c_hostname)
 
       NULLIFY (wait_msg)
--- a/src/ipi_server.F
+++ b/src/ipi_server.F
@@ -87,7 +87,7 @@ CONTAINS
       CALL timeset(routineN, handle)
       CPABORT("CP2K was compiled with the __NO_SOCKETS option!")
 #else
-      CHARACTER(len=default_path_length)       :: c_hostname, drv_hostname
+      CHARACTER(len=default_path_length)       :: c_hostname, drv_hostname, drv_prefix
       INTEGER                                  :: drv_port, handle, i_drv_unix, &
                                                   output_unit, socket, comm_socket
       CHARACTER(len=msglength)                 :: msgbuffer
@@ -102,6 +102,7 @@ CONTAINS
       CALL section_vals_val_get(driver_section, "HOST", c_val=drv_hostname)
       CALL section_vals_val_get(driver_section, "PORT", i_val=drv_port)
       CALL section_vals_val_get(driver_section, "UNIX", l_val=drv_unix)
+      CALL section_vals_val_get(driver_section, "PREFIX", c_val=drv_prefix)
       IF (output_unit > 0) THEN
          WRITE (output_unit, *) "@ i-PI SERVER BEING STARTED"
          WRITE (output_unit, *) "@ HOSTNAME: ", TRIM(drv_hostname)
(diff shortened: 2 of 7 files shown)


oschuett:
Please also adopt the [test_i-pi.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/test_i-pi.sh).

And while you're at it, maybe also add a timeout because right now it will just hang forever.

jhaens:
Thanks for the feedback! I've pushed a follow-up commit that:

1. Adopted `test_i-pi.sh`: updated `tests/i-PI/ipi_client.inp` and `ipi_server.inp` to use the new `/tmp/<PREFIX>_<HOST>` convention (via `PREFIX ipi`) instead of the old hardcoded `/tmp/qiskit_myHost` workaround.
2. Added timeouts: wrapped every CP2K/i-PI process in `test_i-pi.sh` that waits on a socket connection with `timeout ${TIMEOUT_SEC}`, since a stuck connection consumes no CPU time and wasn't caught by the existing `ulimit -t `, it would otherwise hang forever.
3. Found and fixed a gap: while testing the client/server round trip locally, I found that `ipi_server.F` (server mode) was never updated with the same `PREFIX`/`/tmp/` path-building logic as `ipi_driver.F` (client mode). Without this, client and server no longer agreed on the socket path after the original fix. I extended the same logic to `ipi_server.F` and verified a full client/server geometry-optimization run completes successfully end-to-end.

oschuett (approved):

PR #5560, opened 2026-07-09
toolchain/build_cp2k.sh: Always re-create build dir if not --rebuild-only

Growl1234:
See https://github.com/cp2k/cp2k/pull/5549#discussion_r3538158106.


mkrack (approved):
LGTM

PR #5564, opened 2026-07-09
(P)DOS: Restore default behavior of output data

### LINKED ISSUE #5550: PDOS
Dear all

I do not think that the new version of the PDOS with already post-processed data is a good idea. 
I see that in DFT%PRINT%DOS%PDOS the possibility to get raw pods still exists, but I am strongly convinced that this should be the default and not an option that generates more files.  By  the way, I really dislike the additional extension _raw.
 Moreover,  pre-digested data ready to plot  disguise the real meaning of the  density of states, as it should be. 

Finally, the DFT%XAS%PRINT%PDOS seems not to even gave the option of getting  the raw data, as it is the case of DFT%XAS_TDP%PRINT%PDOS

My wish would be to revert the PDOS output as it was before, if this is really against the opinion of the majority, at least make the post processed PDOS a secondary option, maintaining the raw PDOS as the default.

Best
Marcella

Growl1234:
Closes #5550.

The input structure is now:
```text
&DOS                # Generate raw *.dos
  DELTA_E           # Works for both broadened output and histogram-mode total-DOS output.
  &PDOS             # Generate raw *.pdos and enable generating PDOS curves through &CURVE section
    COMPONENTS F
  &END PDOS
  &CURVE            # Generate broadened (P)DOS curve, can be named as *_curve.dos/.pdos
    ENERGY_UNIT
    ENERGY_ZERO
    &BROADEN
      ...
    &END BROADEN
  &END CURVE
&END DOS
```

Along with some additional change:
- Make the histogram-mode (not broadened) total-DOS output on a compact type by omitting bins where both density and occupation are zero.
- Remove duplicate Energy column (before beta-density) for histogram-mode and broadened total-DOS output.
- Make `ENERGY_ZERO` consistent for both alpha and beta spins when generating broadened output.

The k-point PDOS remains available only via broadened output. Hope we can resolve this in the next development cycle.

### DIFF
--- a/docs/methods/electronic_structure/dos.md
+++ b/docs/methods/electronic_structure/dos.md
@@ -28,10 +28,10 @@ used to assign the chemical character of the peaks or band edges.
 
 ```{important}
 Most of the DOS/PDOS functionality described below, including the unified
-`&DFT%PRINT%DOS` interface and the broadened, directly plottable `.dos` and `.pdos` files, is only
-available in CP2K 2026.2 and later versions. In earlier CP2K versions, `.dos` and `.pdos` referred
-to state-resolved raw outputs rather than broadened DOS/PDOS curves. The raw output of PDOS is
-preserved and now written as `.pdos_raw` through `&DFT%PRINT%DOS%PDOS%RAW` keyword.
+`&DFT%PRINT%DOS` interface and the `&CURVE` subsection for broadened, directly plottable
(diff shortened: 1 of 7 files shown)


oschuett:
@marci73, could you please take a look at this PR. It's one of the last few things holding up the release.

marci73 (approved):

Issue #5565, opened 2026-07-09
RFC: Remove remaining GOTO statements and make GOTO a precommit error


Growl1234:
CP2K already discourages the use of `GOTO`, and the convention checker already reports it as [c201](https://www.cp2k.org/conv#c201). However, the remaining legacy occurrences are currently covered by suppressions, so the rule is not fully enforced in practice. Since `GOTO` is really unclear and dangerous, I would like to propose strengthening this existing convention in two steps:

1. Remove the remaining `GOTO` statements from the current Fortran codebase, together with the corresponding `c201` suppressions.
2. Once the tree is clean, move or duplicate the `GOTO` check into the precommit checks and make it an unconditional hard error for the repository, without a suppression mechanism.

The cleanup itself could probably be done incrementally, especially for legacy numerical routines or control-flow-heavy code. However, after the remaining cases are removed, the precommit check should be simple and strict: any `GOTO` statement in the source tree should fail the check. The desired final state should be that no `GOTO` statement is contained and allowed by CP2K at all.

Growl1234:
The remaining number of `GOTO`/`GO TO` I can find:

- [x] `beta_gamma_psi.F`: ~34
- [x] `kpsym.F`: ~33
- [ ] `common/powell.F`: ~19
- [x] `motion/bfgs_optimizer.F`: ~5
- [x] `graphcon.F`: ~2

fstein93:
This is a tricky one. The `GOTO`s come from old legacy code copied from elsewhere. The ones in `common/powell`  are the famous implementations of the Powell solvers which were re-implemented more recently (see the discussion in #3202). These routines are still needed in some cases. I have just checked `beta_gamma_psi` which is trivial as the `GOTO`s are only found in a single routine which isn't used anywhere. Some of the `GOTO`s in `kpsym` were added more recently. Probably, removing them is a hard task while ensuring its functionality. The other two take probably a bit of time but beware of testing to ensure their functionality.

Growl1234:
> I have just checked `beta_gamma_psi` which is trivial as the `GOTO`s are only found in a single routine which isn't used anywhere.

Would that mean `SUBROUTINE bratio` is safe to remove? I just took a look at this file and it seems only `SUBROUTINE psi` is used in one other place...

fstein93:
[...]

PR #5568, opened 2026-07-10
Remove unused beta routines and move digamma to mathlib

Growl1234:
Towards https://github.com/cp2k/cp2k/issues/5565#issuecomment-4926893022.


fstein93 (changes requested):
I have two comments:
1. Let's rename `psi` to `digamma`, the name under which it is commonly known in maths and which is more precise considering that `psi` is also just a Greek letter which is used everywhere with slightly different meanings in Quantum chemistry.
2. I would rename the file and module as a whole considering it doesn't contain the beta and gamma functions anymore.

Growl1234:
> I would rename the file and module as a whole considering it doesn't contain the beta and gamma functions anymore.

What about `src/digamma_function.F` with `MODULE digamma_function` and `PURE FUNCTION digamma`?

he-zilong:
> > I would rename the file and module as a whole considering it doesn't contain the beta and gamma functions anymore.
> 
> What about `src/digamma_function.F` with `MODULE digamma_function` and `PURE FUNCTION digamma`?

I think it should become part of the `src/common/mathlib.F`.

Growl1234:
I think, it's also worth effort to adjust the comments in that code to improve readability.

fstein93 (approved):

oschuett (approved):

PR #5569, opened 2026-07-10
Improve input keyword metadata rendering

Growl1234:
This PR improves the rendering of keyword entries in the generated Input Reference. The data type and default value are moved out of the `py:data` signature and displayed as regular metadata fields alongside `Description`, `Usage`, `Lone keyword`, and other keyword properties.

The existing `py:data` directives, fully qualified object names, and legacy anchors are preserved, so cross-references and existing deep links should remain unchanged. This PR only changes the presentation of the generated documentation.

### DIFF
--- a/docs/generate_input_reference.py
+++ b/docs/generate_input_reference.py
@@ -340,20 +340,28 @@ def render_keyword(
         output += [f":module: {section_xref}"]
     else:
         output += [":noindex:"]
-    output += [f":type: '{data_type}{n_var_brackets}'"]
-    if default_value or default_unit:
-        default_unit_bracketed = f"[{default_unit}]" if default_unit else ""
-        output += [f":value: '{default_value} {default_unit_bracketed}'"]
     output += [""]
+
+    # Render keyword properties as compact, unbulleted lines instead of placing the
+    # type and default value in the object signature.
+    metadata = [f"**Type:** {escape_markdown(data_type + n_var_brackets)}"]
+    if default_value or default_unit:
+        default = default_value
+        if default_unit:
+            default += f" [{default_unit}]"
+        metadata += [f"**Default:** {escape_markdown(default.strip())}"]
     if repeats:
-        output += ["**Keyword can be repeated.**", ""]
+        metadata += ["**Repeatable:** yes"]
     if len(keyword_names) > 1:
-        aliases = " ,".join(keyword_names[1:])
-        output += [f"**Aliases:** {escape_markdown(aliases)}", ""]
+        aliases = ", ".join(keyword_names[1:])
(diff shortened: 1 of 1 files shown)


Growl1234:
One another possible idea would be to allow an input section or keyword to register a documentation note directly in the input source, which the generator could render as a MyST note admonition (` ```{note}`). This could be useful for important caveats or usage guidance that does not fit naturally into the ordinary description.

oschuett:
> One another possible idea would be to allow an input section or keyword to register a documentation note directly in the input source, which the generator could render as a MyST note admonition ( ```{note}). 

Descriptions are passed through to Sphinx with only [very light escaping](https://github.com/Growl1234/cp2k/blob/418efe7be4b1c6bb9af34b62bd887a6a81b5165c/docs/generate_input_reference.py#L434). So you should already be able to put admonitions into a description.

Regarding the [new rendering](https://ci.cp2k.org/artifacts/run-cp2k-manual-18fb16f7/manual/CP2K_INPUT/FORCE_EVAL/DFT.html#CP2K_INPUT.FORCE_EVAL.DFT.AUTO_BASIS) of keywords, it feels less glanceable to me. But then again, my programmer's eyes are really used to the notation `name: type = default`. I'd be curious what others think?

Growl1234:
> But then again, my programmer's eyes are really used to the notation `name: type = default`. I'd be curious what others think?

I think, the problem is that we cannot assume that most users, especially beginners, will immediately understand that, in the notation `name: type = default`, the value after the colon denotes the type and the value after `=` denotes the default. Explicit labels may be slightly less compact, but they remove this ambiguity.

oschuett:
How about we do both, i.e. we keep the current header line and just add the explicit labels below (redundantly)?

Growl1234:
I am not very convinced by keeping both, because it would duplicate exactly the same information and make each keyword entry visually heavier. We would have something like

> **UKS:** ***logical = F***

followed immediately by

> **Type:** logical
> **Default:** F

which feels rather redundant to me.

I would prefer to have one unambiguous representation rather than two equivalent ones. The explicit labels are slightly less compact, but I think they are clearer for users who are not already familiar with the `name: type = default` convention.

That said, I am happy to see what others prefer.

oschuett:
> I would prefer to have one unambiguous representation rather than two equivalent ones.

I'm happy to give this a try, but let's do it after the release so that we can course-correct if necessary.

oschuett:
Shall we merge this now and wait for feedback?

Growl1234:
Can be merged. I will also think of whether better ways can be found.

oschuett (approved):

PR #5571, opened 2026-07-10
CMake: Remove cmake_minimum_required from Config

Growl1234:
Setting `cmake_minimum_required()` in `cp2kConfig.cmake.in` will override the policy settings inherited from downstream projects, such as `CMP0144`.

### DIFF
--- a/cmake/cp2kConfig.cmake.in
+++ b/cmake/cp2kConfig.cmake.in
@@ -5,6 +5,10 @@
 #!   SPDX-License-Identifier: GPL-2.0-or-later                                                     !
 #!-------------------------------------------------------------------------------------------------!
 
+if(CMAKE_VERSION VERSION_LESS "3.24")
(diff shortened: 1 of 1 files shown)


mtaillefumier (approved):
We need to guard the find_dependency(HIP / ROCM) to at least the version indicated in the CMakeLists.txt because hip/rocm detection only start to work without too much trouble starting with cmake 3.23.

So a if statement like this should be added at the beginning of the config file 

`if(CMAKE_VERSION VERSION_LESS "3.24") message(FATAL_ERROR ...) endif()`

please check the version number.

Growl1234:
Thanks for the hint. I make the minimun-required version number consistent with `CMakeLists.txt`:
https://github.com/cp2k/cp2k/blob/a2edeeb2c6cd0ed68d5f3431fb60221baef61270/CMakeLists.txt#L8

PR #5574, opened 2026-07-10
Update local spack recipe for gauxc

mkrack:


mkrack:
@RMeli Please, could you review this request?

tools/spack/spack_repo/cp2k_dev/packages/gauxc/package.py
```diff
@@ -51,23 +51,24 @@ class Gauxc(CMakePackage, CudaPackage):
     variant("cutlass", default=False, description="Build with CUTLASS linear algebra support")
     variant("pic", default=True, description="Build position independent code")
 
-    conflicts("+magma", when="~cuda ~hip", msg="MAGMA requires CUDA or HIP")
+    # Skala resource file
+    SKALA_RESOURCES = {"1.1": "0c8432ac3f03c8f1276372df9aca5b7ee7f8939d47a8789eb158976e89aa0606"}
+    variant("skala_version", default="none", values=("none", *SKALA_RESOURCES.keys()), multi=False)
+    for skala_version, sha256 in SKALA_RESOURCES.items():
```
RMeli, line 57: This loops over all items, so makes the variant essentially useless. If you want to be able to specify multiple `skala_version`, then the variant should be `multi=True` and there is some helper functions to use to enable either `none` or a list of values (I will need to double check again the Spack documentation for this...). Then the loop should be over `self.spec.variants["skala_version"].value` (which for `multi=True` is a tuple) if `not self.spec.satisfies("skala_version=none")`.

RMeli (commented):
@mkrack sorry for the late review, but I was off on Friday.

PR #5575, opened 2026-07-10
Drop local spack recipe for tblite

mkrack:


Growl1234:
At least these lines are missing in `cp2k_deps_s-static.yaml`:
```yaml
    tblite:
      require:
        - "~shared"
```

However, IIRC Spack cannot make sure the `~shared` variant is passed to the dependencies of tblite. Therefore, keeping a minor recipe (including only CMake configuration as CP2K can only detect tblite through `tblite-config.cmake` anyways, so only around 40 lines can be kept) might still be preferred?

mkrack:
> At least these lines are missing in `cp2k_deps_s-static.yaml`:
> 
> ```yaml
>     tblite:
>       require:
>         - "~shared"
> ```
> 
> However, IIRC Spack cannot make sure the `~shared` variant is passed to the dependencies of tblite. Therefore, keeping a minor recipe (including only CMake configuration as CP2K can only detect tblite through `tblite-config.cmake` anyways, so only around 40 lines can be kept) might still be preferred?

Thanks for the reminder. I am running local tests currently to resolve the issue. If these do not work, we can keep the local `tblite` recipe.

mkrack:
It seems that dropping the local `tblite` recipe still requires a patching of upstream packages -> closing this request.

PR #5578, opened 2026-07-10
Disable DLA-Future only with MPICH

mkrack:


Growl1234:
My local tests show even enabling DLAF only with CP2K but not with SIRIUS brings significant time and memory consuming for debug builds...

mkrack:
@RMeli DLA-Future is now disabled again with OpenMPI to make this pull request mergable, but it is now easy to activate it again for MPICH and/or OpenMPI in `make_cp2k.sh`.

RMeli:
Ok, thanks @mkrack

PR #5579, opened 2026-07-10
Docs: write build timstamp to page footer, link past releases in CP2K_INPUT, add acronyms, add notes about NaN/Inf, edit convergence.md

he-zilong:


docs/generate_input_reference.py
```diff
@@ -107,7 +107,11 @@ def build_input_reference(root: ET.Element, output_dir: Path) -> None:
 
     assert compile_revision.startswith("git:")
     github_url = f"https://github.com/cp2k/cp2k/tree/{compile_revision[4:]}"
-    output += [f"Based on {cp2k_version} ([{compile_revision}]({github_url}))", ""]
+    output += [
+        f"Built at {tstr} for {cp2k_version} ([{compile_revision}]({github_url}))",
```
oschuett: This renders to:

> Built at 2026-07-10T17:26:24+00:00 for CP2K version 2026.1 (Development Version) ([git:f7ff04c](https://github.com/cp2k/cp2k/tree/f7ff04c))

Do we really need the time down to the second? How just showing the date:
```suggestion
    from datetime import date
    today = str(date.today())
    output += [
        f"Built on {today} for {cp2k_version} ([{compile_revision}]({github_url}))",
```

he-zilong: Is the Input Reference at https://manual.cp2k.org/trunk/index.html built once evey 24 hours like the frequency "daily" specified in `cp2k-ci.conf` for the `manual generation` test? (It appears that the dashboard automatic run is not strictly once every 24 hours; maybe it is based on the number of recent commits?) I use the full HH:MM:SS format in case some user want to build the Input Reference (and other documentations) locally for the development version (which may be multiple times a day).

Growl1234: > Is the Input Reference at https://manual.cp2k.org/trunk/index.html built once evey 24 hours like the frequency "daily" specified in `cp2k-ci.conf` for the `manual generation` test?

It's daily:
https://github.com/cp2k/cp2k/blob/440cf4f012e149c632c3a8b884bbbd90dd385da1/tools/docker/cp2k-ci.conf#L197-L205

docs/methods/dft/convergence.md
```diff
@@ -13,6 +13,19 @@
...
+ *   ___                                                                       *
+ *  /   \                                                                      *
+ * [ABORT]                                                                     *
+ *  \___/                KS energy is an abnormal value (NaN/Inf).             *
+ *    |                                                                        *
+ *  O/|                                                                        *
+ * /| |                                                                        *
+ * / \                                                    qs_ks_methods.F:1166 *
+ *******************************************************************************
+```
```
Growl1234, line 34: IIUC this is not an error about SCF convergence?

he-zilong: But it is also hampering convergence and the resolutions as far as I know are more or less the same.

he-zilong: I have noted that this is another symptom for SCF going wild before MAX_SCF.

he-zilong:
It turns out that there are Sphinx configurations for inserting last updated time into the footer of every page, which is on the same line as the copyright and is far more visible than a single occurrence on the CP2K_INPUT.
```
© Copyright 2000-2026, CP2K Developers. Last updated on Mon, 13 Jul 2026 04:16:48 +0000.
```
I think this PR is good enough for now.

docs/generate_input_reference.py
```diff
@@ -107,7 +111,11 @@ def build_input_reference(root: ET.Element, output_dir: Path) -> None:
 
     assert compile_revision.startswith("git:")
     github_url = f"https://github.com/cp2k/cp2k/tree/{compile_revision[4:]}"
-    output += [f"Based on {cp2k_version} ([{compile_revision}]({github_url}))", ""]
+    output += [
+        f"Built {prep} {tstr} for {cp2k_version} ([{compile_revision}]({github_url})).",
+        "",
+        "For past stable releases, see [](./versions).",
+    ]
```
oschuett, line 112: Since we're now enabling `html_last_updated_fmt`, could you please restore this file - except for the hint about previous releases, which is a good idea.
```suggestion
    output += [f"Based on {cp2k_version} ([{compile_revision}]({github_url}))"]
    output += ["For previous releases, see [](./versions).", ""]
```

PR #5581, opened 2026-07-12
Fix Bloch phases for fully symmetry-reduced k-point grids

DCM-Uni-Paderborn:
## Summary

- Track crystallographic atom-cell shifts separately from CP2K internal periodic-image gauge shifts.
- Select the Bloch-phase direction for each symmetry operation by testing overlap-matrix covariance.
- Rebuild symmetry for moving and finite-difference geometries, and use full current-geometry symmetry for DEBUG finite differences by default.
- Fall back to inversion/time-reversal reduction when approximate K290 operations form overlapping, non-closed mesh orbits or when the cell representation does not support full reduction.
- Add DFT and xTB/TBLITE regressions for boundary-crossing atoms, even Gamma-centered meshes, cell optimization, and nonsymmorphic SPGLIB and K290 operations.

## Motivation

Full space-group k-point reduction could use a Bloch-phase convention inconsistent with CP2K periodic neighbor-list images. This produced incorrect symmetry-transformed matrices and could make energies, forces, and stress depend on atom wrapping or on the selected irreducible mesh. The issue is most visible for nonsymmorphic operations and atoms near periodic cell boundaries.

The overlap matrix provides an invariant reference for choosing the phase convention. Calibrating against overlap covariance avoids grid-parity heuristics and works for both DFT and tight-binding paths.

## Validation

- CMake build with GCC 16.1.0, SPGLIB, and TBLITE: `cp2k.ssmp` built successfully.
- Eight new regression inputs pass their registered CP2K matchers.
- Monoclinic DFT DEBUG finite differences: force sum error `0.0`, stress sum error `4.81e-10`; the test retains the `2/m` to `-1` symmetry transition and the two-to-four irreducible k-point transition.
- The monoclinic DEBUG case takes `7.28 s` locally instead of `61.80 s` while keeping the full finite-difference stress tensor.
- SPGLIB and K290 phase tests reproduce the expected energies and irreducible k-point counts.
- The k-point cell-optimization regression reproduces its energy and six-point reduced mesh.
- The pre-existing Si GFN2 3x3x3 finite-difference stress regression retains all three diagonal components and six irreducible Bloch k-points; `EPS_SCF 5e-7` reduces its local runtime from `8.11 s` to `6.21 s`, while the registered stress sum remains `9.84e-5` (tolerance `2e-4`).


Growl1234:
Would it resolve the density issue mentioned in #5280?

Growl1234:
And looks like this somewhat breaks SCF reuse of MO in Wannier90 path?

DCM-Uni-Paderborn:
> [Would it resolve the density issue mentioned in #5280?](https://github.com/cp2k/cp2k/pull/5581#issuecomment-4951468108)

@Growl1234 Yes, this PR addresses the separate incorrect-density issue described in #5280: full atomic k-point symmetry could use Bloch phases inconsistent with CP2K's current PBC image convention, producing an incorrect density even with `USE_GUESS`. The fix separates crystallographic atom-cell shifts from CP2K's internal neighbor-list gauge and selects the phase convention from overlap-matrix covariance. It does not implement the `USE_PREV_WF`/WFN-history reuse proposed in #5280; that remains a separate topic. Boundary-crossing and shifted-cell regressions cover the density/gauge problem here.

DCM-Uni-Paderborn:
> [And looks like this somewhat breaks SCF reuse of MO in Wannier90 path?](https://github.com/cp2k/cp2k/pull/5581#issuecomment-4951831670)

@Growl1234 Yes, good catch. The first PR head broke SCF-MO reuse because the Wannier90 transform applied an unconditional Bloch-phase sign. Commit `dd3b10f` fixes this by using the gauge-adjusted atom-cell shifts and the phase direction selected from overlap covariance; the original heuristic remains only as a fallback before calibration. The Harris k-point path was fixed at the same time to provide its overlap matrix to the calibration. Focused regressions pass, and the preceding full `pdbg` CI run passed all 5696 tests; the current head only adjusts a regression-test SCF tolerance to address the independent `sdbg` runtime limit.

Growl1234:
> Yes, this PR addresses the separate incorrect-density issue described in #5280...

Thank you! I validated it on my local machine and it worked!

> It does not implement the `USE_PREV_WF`/WFN-history reuse proposed in #5280; that remains a separate topic.

I will take care of this, but the patch here will definitely make things much better for my work.

Growl1234:
Is this PR ready for merge? I hope this fix can be included in upcoming release so that users would not complain (or complain less) about a k-point summetry implementation that's not yet very robust.

CC: @oschuett

DCM-Uni-Paderborn:
> Is this PR ready for merge? I hope this fix can be included in upcoming release so that users would not complain (or complain less) about a k-point summetry implementation that's not yet very robust.
> 
> CC: @oschuett

IMO yes, but it's hard to foresee the impact on the dashboard. I leave it to @oschuett to make the final call on #5581 & #5582, respectively.

oschuett:
Since k-points have been a major topic of the last months, let's make sure we get it right.

Btw, please also check that your contributions are well represented in the [changelog](https://manual.cp2k.org/trunk/changelog.html).

oschuett:
The dashboard found a few failures of `regtest-kp-1-spg/c_geo_opt_kpsym_spglib.inp`:
```
 *  \___/    No Bloch-phase direction preserves overlap covariance; residual=  *
 *    |               2.8207E-01, irreducible k-point=1, operation=2           *
 * / \                                                   kpoint_methods.F:2206 *
```

@DCM-Uni-Paderborn, those are benign, right?

Growl1234:
I feel this is most likely a numerical stability issue? I cannot reproduce it locally. Or perhaps we can increase the `CUTOFF` and `REL_CUTOFF`.

oschuett:
Yes, I think that tests was just pushed a bit too far to make it run quicker. It's also reassuring that it aborts instead of giving a wrong result.

oschuett:
We still see occasional failures of `regtest-kp-1-spg/c_geo_opt_kpsym_spglib.inp`. Today e.g. in [psmp](https://dashboard.cp2k.org/archive/current-psmp/commit_a0df33dd6ace3f0789f82cf49f0d553fde41bf3c.txt).

@DCM-Uni-Paderborn, could you please take a look.

Growl1234:
I am able to reproduce it now. It is more likely to happen with hybrid parallelism, but still occasionally (1 in 5).

Growl1234:
Found it. The failure is most likely caused by an unstable periodic-image gauge for atoms on the fractional-cell boundary: tiny floating-point differences make `pbc()` choose either +0.5 or -0.5, resulting in inconsistent real-space image assignments and Bloch phases after the geometry-optimization update.

(This is somewhat similar to what I wanted to fix in #5219 for k-point WFN extrapolation.)

oschuett:
> Found it. The failure is most likely caused by an unstable periodic-image gauge for atoms on the fractional-cell boundary...

Great work! So this is a bug after all then. So should we close #5625 and instead open an issue to track this problem?

DCM-Uni-Paderborn:
The intermittent half-cell boundary failure reported in this PR is now addressed by #5646, merged as [`7b6f4dd0`](https://github.com/cp2k/cp2k/commit/7b6f4dd061754e37c41e190ef48877c08126206a). The fix stabilizes periodic-image selection near fractional half-cell boundaries, keeps the PBC-derived Bloch gauge consistent in the affected paths, and adds deterministic unit coverage for the previously intermittent condition. Tracking issue #5626 has therefore been closed.

PR #5582, opened 2026-07-12
Fix TBLITE forces and stress for periodic k-point calculations

DCM-Uni-Paderborn:
## Summary

- Contract the native TBLITE Hamiltonian, overlap, coordination-number, SCC, and multipole derivatives with the CP2K density matrices used for forces.
- Accumulate explicit real-space strain-derivative matrices for periodic k-point calculations and add them consistently to the CP2K virial.
- Handle spin-resolved SCC terms, Gamma sampling, and Nyquist self-image contributions without the previous mesh-size scaling heuristics.
- Route overlap derivatives through CP2K native overlap-force machinery and keep atomic and cell derivatives on the same density reference.
- Add finite-difference force and stress regressions for GFN1-xTB and GFN2-xTB at Gamma and with native Bloch k-points.

## Motivation

The CP2K-native TBLITE path produced forces and stress that were inconsistent with finite differences for periodic systems, especially with k-point sampling. The previous implementation estimated several periodic image and multipole contributions from mesh parity and image counts, while some Hamiltonian, overlap, SCC, and coordination-number response terms were not contracted on the same footing.

This change evaluates those derivative contributions directly from the real-space matrices and the converged CP2K density, removing the empirical correction factors.

## Dependency

#5581 has been merged. This branch is rebased onto current `master` and no longer carries the #5581 commits; the PR diff now contains only the TBLITE derivative implementation and its focused regressions.

## Validation

- CMake build of the rebased head with GCC 16.1.0 and TBLITE 0.6: `cp2k.ssmp` built successfully.
- The complete `regtest-tblite-ipea1`, `regtest-tblite-gfn1-grad`, and `regtest-tblite-gfn2` selections pass all 113 registered matchers.
- GFN1 H2O k-point DEBUG: stress sum `1.56e-9`, periodic stress sum `3.95e-10`, force sum `0.0`.
- GFN1 h-BN k-point DEBUG: stress sum `5.67e-7`, force sum `7e-8`.
- GFN2 O2 unrestricted k-point DEBUG: stress sum `7.14e-7`, force sum `7e-8`.
- GFN2 Si k222 and k333 stress sums: `4.13e-6` and `1.65e-5`; the compact N2 CP2K-mixer full-tensor k222 check gives `6.29e-9`.
- The complete two-rank GFN2 selection passes `--flagslow` with no slow tests; the full-tensor CP2K-mixer case takes `3.87 s` locally instead of `21.45 s` for the superseded Si case.
- Two-rank MPI runs of the three complete TBLITE selections pass all 113 registered matchers (11 IPEA1, 33 GFN1, and 69 GFN2).
- The IPEA1 k-point finite-difference stress remains rank independent: sums of `6.89e-9`, `7.43e-9`, and `7.43e-9` with 1, 2, and 4 MPI ranks.
- CP2K precommit checks: all 8924 files passed.
- The inherited Si GFN2 k333 DEBUG case retains all three diagonal finite-difference stress components and six irreducible Bloch k-points; `EPS_SCF 3e-7` takes `6.42 s` locally and gives a stress sum of `1.65e-5` (tolerance `2e-5`), balancing the strict derivative matcher with the `sdbg` runtime limit.


oschuett:
Since this PR also mostly fixes problems with existing features, I'd include it in the release.

However, now there are some conflicts and @jpoto has not yet approved it.

Maybe for the future we should add @DCM-Uni-Paderborn as a second code owner for _dispersion + tblite_ ?

oschuett:
@DCM-Uni-Paderborn could you please resolve the merge conflicts.

src/kpoint_methods.F
```diff
@@ -184,11 +186,16 @@ SUBROUTINE kpoint_initialize(kpoint, particle_set, cell)
          ! kind type list
          ALLOCATE (kpoint%atype(natom))
          kpoint%atype = atype
-         ! Match the periodic gauge used by the k-space matrices.
+         ! Match the atom images used by CP2K's periodic neighbor lists.
```
jpoto, line 189: I don't know if the neighbor list iterators would be more consistent.

src/tblite_interface.F
```diff
@@ -1094,6 +1134,13 @@ SUBROUTINE build_tblite_matrices(qs_env, calculate_forces)
          jatom = sab_orb(1)%nlist_task(slot)%jatom
          cell(:) = sab_orb(1)%nlist_task(slot)%cell(:)
          rij(1:3) = sab_orb(1)%nlist_task(slot)%r(1:3)
+         raw_iatom = iatom
+         raw_jatom = jatom
+         raw_rij = rij
+         native_cn_icol = 0.0_dp
```
jpoto, line 1140: How are these cn related to the ones determined by tblite?

jpoto (approved):
looks good to me.

PR #5586, opened 2026-07-13
GLBOPT: Reduce output of tests

oschuett:


he-zilong:
I had no idea that individual optimization tasks could accumulate too many `FINAL_STRUCTURE` files for a swarm task. That said, could the regtests use `LOW` or even `SILENT` for `PRINT_LEVEL` given that the default is `MEDIUM`?

oschuett:
Yes, probably that would have worked too.

PR #5588, opened 2026-07-14
Address issues #1498 and #4023

### LINKED ISSUE #1498: Restart from calculation with &EXT_RESTART and &PERIODIC_EFIELD fails with CPASSERT
When trying to use the restart file written by CP2K for a simulation with the following `EXT_RESTART` part AND a `&PERIODIC_EFIELD`:

```
&EXT_RESTART
        RESTART_FILE_NAME   ${RESFILE}-1.restart
        RESTART_COUNTERS T
        RESTART_POS T
        RESTART_BAND T
&END
```

CP2K immediately crashes with:

```
 *  \___/                             CPASSERT failed                          *
 * / \                                                  cp_control_utils.F:424 *
```

The culprit is the internal generation of the `REAL_TIME_PROPAGATION` section in case `&EXT_RESTART` is used and `RESTART_RTP` is not explicitly set the `.FALSE.`
This then triggers the assertion which checks that RTP and the periodic efield are not enabled at the same time.

The user can salvage the restart file by explicitly setting `RESTART_RTP .FALSE.`.

### LINKED ISSUE #4023: Combining AUTO_BASIS with GHOST atoms doesn't work
I get an error when combining automatic generation of the RI basis is combined with ghost atoms.
Input:
[dft_rpa_ghost.inp.txt](https://github.com/user-attachments/files/18907665/dft_rpa_ghost.inp.txt)
Output:
[dft_rpa_ghost.inp.out.txt](https://github.com/user-attachments/files/18907663/dft_rpa_ghost.inp.out.txt)
Error:
[output.txt](https://github.com/user-attachments/files/18907664/output.txt)

The problem is that econf:
(https://github.com/cp2k/cp2k/blob/master/src/auto_basis.F#L103)
is not initialized without the potential. 

A workaround is to generate the RI basis by a separate calculation.

he-zilong:
This PR:

- closes #1498 by turning the CPASSERT into a CPABORT with a clear note;
- closes #4023 with error messages as suggested by Prof. Hutter.

Also, the possible arguments for `AUTO_BASIS` are listed simply in the description as a workaround; converting it to a "Valid values" list with full-fledged explanations would take more efforts.

### DIFF
--- a/src/auto_basis.F
+++ b/src/auto_basis.F
@@ -61,7 +61,7 @@ CONTAINS
       INTEGER, INTENT(IN), OPTIONAL                      :: basis_sort
 
       CHARACTER(LEN=2)                                   :: element_symbol
-      CHARACTER(LEN=default_string_length)               :: bsname
+      CHARACTER(LEN=default_string_length)               :: bsname, kname
       INTEGER                                            :: i, j, jj, l, laux, linc, lmax, lval, lx, &
                                                             nsets, nx, z
       INTEGER, DIMENSION(0:18)                           :: nval
@@ -84,7 +84,7 @@ CONTAINS
                   2.0_dp, 2.0_dp, 2.0_dp, 2.0_dp, 2.0_dp, 2.0_dp, 2.0_dp, 2.0_dp]
       !
       CPASSERT(.NOT. ASSOCIATED(ri_aux_basis_set))
-      NULLIFY (orb_basis_set)
+      NULLIFY (orb_basis_set, econf)
       IF (.NOT. PRESENT(basis_type)) THEN
          CALL get_qs_kind(qs_kind, basis_set=orb_basis_set, basis_type="ORB")
       ELSE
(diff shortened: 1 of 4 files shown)


fstein93:
I have just checked the code because I was never aware that there was the size option `X`. Unfortunately, it is never used. I guess it was implemented as a place holder but it always results in `CPABORT`s at the actual creation step. So, I would rather remove this option.

he-zilong:
> I have just checked the code because I was never aware that there was the size option `X`. Unfortunately, it is never used. I guess it was implemented as a place holder but it always results in `CPABORT`s at the actual creation step. So, I would rather remove this option.

Good catch. Using `X` to specify the size passes `basis_cntrl = -1` to `create_ri_aux_basis_set()` or `create_lri_aux_basis_set()` which is surely invalid. This problem is also affecting the `AUTO_BASIS` keyword under the `FORCE_EVAL/PROPERTIES/TDDFPT` and `FORCE_EVAL/PROPERTIES/RIXS/TDDFPT` sections.

Edit: it seems that the purpose of `X` as the size is to accompany `X` as the type for the default behavior of not automatically generating auxiliary basis sets. So a setup like `AUTO_BASIS RI_AUX X` should not be valid, but the representation of the default behavior needs to be redesigned.

he-zilong:
Well, I have revised the logic so that the second argument `X` does not get mapped to a size value of -1. Now it checks if the first argument is also `X`, and if not (e.g. `AUTO_BASIS RI_AUX X`) the program will abort with an explanation. `AUTO_BASIS X X` is reserved as a special syntax for not using any automatically generated basis sets by default.

Unfortunately this is a bit late [for the upcoming release](https://github.com/cp2k/cp2k/issues/5591#issuecomment-4968878286)...

PR #5591, opened 2026-07-14
tblite v0.7.0 and trexio v2.6.1

Growl1234:
- Update tblite to v0.7.0, along with API change and version requirement.
- Update trexio to v2.6.1 (Spack; toolchain already latest).
- Adjust reference values accordingly considering the improvement for periodic implementation in tblite-0.7.0 (https://github.com/cp2k/cp2k/pull/5591#issuecomment-5030432381, https://github.com/cp2k/cp2k/pull/5591#issuecomment-5030550797).
- Drop redundant tblite patches in toolchain (https://github.com/cp2k/cp2k/pull/5591#issuecomment-5031026319).
- Try preventing FPE errors.


tools/spack/spack_repo/cp2k_dev/packages/tblite/package.py
RMeli, line 1: LGTM, but would be nice to have this added upstream if possible (support for `meson` seem simple enough to add in this case).

RMeli: Feel free to add me as a reviewer upstream, I can ensure a timely merge.

Growl1234: If you means the version update, it is here: https://github.com/spack/spack-packages/pull/5577

oschuett:
Just as a heads up: Since version upgrades tend to require follow up work, I'd like to merge this after the release.

 I just merged #5582, which I intend to be the final PR of this release. For the rest of the day I'll refresh the dashboard. If all goes well then I'll tag the release tonight.

tools/spack/spack_repo/cp2k_dev/packages/tblite/package.py
RMeli: Thanks for pointing it out. Why not adding the new variants upstream too?

Feel free to add me as a reviewer the next time, so I'm aware of the Spack PRs. Since I have merge rights I can merge it quickly (after package maintainers also had a look).

Growl1234: > Why not adding the new variants upstream too?

Actually the new variants have been added in https://github.com/spack/spack-packages/pull/5349 and I copied them here.

> Feel free to add me as a reviewer the next time, so I'm aware of the Spack PRs. Since I have merge rights I can merge it quickly (after package maintainers also had a look).

Sorry but unfortunately it seems I don't have the triage access in `spack-packages` repo to request a review.

RMeli: Ah, I see. I forgot about that, and they have been added before the release. I just merged the version bump, so in principle you should be able to bump the hash to `331fc1e9cf6aee8f078ff8b7dd600a2e770fbc4c` and remove the custom package.

> Sorry but unfortunately it seems I don't have the triage access in spack-packages repo to request a review.

I see. You can just tag me normally in a comment.

Growl1234: Thank you. Currently the local recipe is kept on purpose to make `spack-ssmp-static` work, as the reason described in https://github.com/cp2k/cp2k/pull/5575#issuecomment-4935625417:

> IIRC Spack cannot make sure the `~shared` variant is passed to the dependencies of tblite.

With the minor recipe here, we can make sure tblite uses its `subproject` to handle dependencies and make the intended `BUILD_SHARED_LIBS=OFF` consistent through them.

he-zilong:
> Just as a heads up: Since version upgrades tend to require follow up work, I'd like to merge this after the release.
> 
> I just merged #5582, which I intend to be the final PR of this release. For the rest of the day I'll refresh the dashboard. If all goes well then I'll tag the release tonight.

Hey, we are not even finished with the [changelog](https://github.com/cp2k/cp2k/blob/3c427e6f4deaa53fba46edebc92f8b003492eb80/docs/changelog.md) yet... Could #5585 and #5588 be at least included since one is a small CMake adjustment and another is a bug fix with clarified documentation?

Growl1234:
There is a CMake error, which I can reproduce with Spack but cannot with toolchain:
```
CMake Error at CMakeLists.txt:887 (find_package):
  Found package configuration file:

    /opt/cp2k/spack/spack/opt/spack/view/lib/cmake/dftd4/dftd4-config.cmake

  but it set dftd4_FOUND to FALSE so package "dftd4" is considered to be NOT
  FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing:
  mctc-lib::mctc-lib
```
@awvwgk maybe could you have a look?

oschuett:
> Could https://github.com/cp2k/cp2k/pull/5585 and https://github.com/cp2k/cp2k/pull/5588 be at least included ...

Neither of these PRs is a major bug fix and I already launched the tests. So, I'd really like to finish this off today.

On the bright side, the next release is just 6 months away...  ;-)

tools/toolchain/scripts/stage8/install_tblite.sh
```diff
@@ -32,7 +32,9 @@ case "$with_tblite" in
     if verify_checksums "${install_lock_file}"; then
       echo "tblite-${tblite_ver} is already installed, skipping it."
     else
-      retrieve_package "${tblite_sha256}" "tblite-${tblite_ver}.tar.xz"
+      # retrieve_package "${tblite_sha256}" "tblite-${tblite_ver}.tar.xz"
```
jpoto: @oschuett can you give me an account for cp2k.org or but the new file there?

CMakeLists.txt
```diff
@@ -909,7 +909,7 @@ if(CP2K_USE_ACE)
 endif()
 
 if(CP2K_USE_TBLITE)
-  find_package(tblite REQUIRED)
+  find_package(tblite 0.7.0 REQUIRED)
```
jpoto: I don't know if we want a strict version dependence.

Growl1234: The reason why >=0.7.0 is required is due to the [API change](https://github.com/cp2k/cp2k/pull/5591/changes#diff-bcb4c62dc5f01e7290dab4db5c5a625ec723dfe2ca6471bae33faedd466be3fcL479-R479) which is related to [here](https://github.com/tblite/tblite/pull/338/changes#diff-95ba87f4134e760d4ac01c81d116de54485cff57b19818286176a714f11e384bL265-R272) in https://github.com/tblite/tblite/pull/338. I don't quite want to make the "branch" again like dftd4 v3.x, v4.0.0~v4.1.1 and v4.2.0. That said, we can leave it for discussion.

he-zilong:
Okay then.

@Growl1234 Could you please take the opportunity to make the first discussion thread for the upcoming release on our forums _in time_, i.e. ahead of others? I really really dislike seeing machine translations or AI interpretations for the change log...

Growl1234:
> Could you please take the opportunity to make the first discussion thread for the upcoming release on our forums _in time_, i.e. ahead of others? I really really dislike seeing machine translations or AI interpretations for the change log...

I can take it and I already had a draft `.docx` in Chinese with some aspects I'm interested in, but I'm afraid I cannot make sure all changes can be properly included...

tools/toolchain/scripts/stage8/install_tblite.sh
oschuett: https://www.cp2k.org/static/downloads/tblite-0.7.0.tar.xz

I'm also happy to give you an account. In the long run we should build a better solution for this though.

Growl1234:
> For the rest of the day I'll refresh the dashboard. If all goes well then I'll tag the release tonight.

Most of Spack testers are failing due to not able to fetch `skala.fun`...

Growl1234:
@he-zilong I've posted one: http://bbs.keinsci.com/thread-59952-1-1.html. Feel free to point out problem, as my post might be relatively incomplete.

Growl1234:
The CMake failure turns out to be a problem on the dftd4 side (`dftd4-config.cmake` includes its target file before finding dependency), which has been fixed upstream after 4.2.0 release. I added a workaround to find `mctc-lib` before `dftd4`; once new dftd4 and tblite releases are out it can be dropped.

Growl1234:
The regtests with newer tblite show many wrong results.

@DCM-Uni-Paderborn, I believe you are willing to have a look.

Growl1234:
Looks like the "wrong result"s are related to https://github.com/tblite/tblite/pull/350 (and possibly https://github.com/tblite/tblite/pull/343 and https://github.com/tblite/tblite/pull/356), which means the old reference values might be no longer valid.

Growl1234:
I checked the consistency with tblite CLI with [this input](https://github.com/user-attachments/files/30213527/bn_gfn1_kp_spglib_phase.inp.txt) via:

```
&REFERENCE_CLI
  STOP_ON_ERROR T
  CHECK_VIRIAL T
  ERROR_LIMIT 5.0E-8
&END REFERENCE_CLI
```

The results are as below:

- Version 0.6.0:
```
 Energy CP2K/CLI/absdiff:   -1.863580000164E+01   -1.863580003102E+01    2.938107357409E-08
 Gradient diff sum/max:    5.146529057679E-15    5.612600529949E-16
 Virial diff sum/max:    2.670909369674E-01    8.903031232248E-02
```

- Version 0.7.0:
```
 Energy CP2K/CLI/absdiff:   -1.857844779689E+01   -1.857844782367E+01    2.677639798776E-08
 Gradient diff sum/max:    6.238106039281E-15    8.567792675745E-16
 Virial diff sum/max:    8.186482864389E-07    2.728827626086E-07
```

So it looks like tblite 0.7.0 actually improves the periodic implementation, and the reference values should probably be updated accordingly? If so, this would also support the intent to require tblite 0.7.0 or newer in CMake.

Growl1234:
I also removed the two tblite patch from toolchain, as:

- No evidence shows these patches have mathmatical effect on the tblite code path. (OpenMP reduction might be of help but not significantly)
- Spack recipe doesn't contain them and Spack tests work well at the same time (which also leads to part of reason 1); needs to keep dependency behavior consistent between toolchain and Spack.
- Calculation settings should *not* be controlled through environment variables.

Growl1234:
I cannot reproduce the FPE in debug type of OpenBLAS-built CP2K locally. The commit regarding Arnoldi diagonalization is what I found when running pdbg with MKL, but appearently it's not the same path as what CI detected.

It's also strange that the FPE happens with toolchain but not Spack.

[Edit: From backtrace I guess it's related to TC2 purification; let's try switching off non-monotonic TC2 to see how it goes.]

jpoto (approved):
Nice update to tblite version with the patches included. 

@mkrack is there a way to ensure that cp2k uses tblite 0.7.0 or higher (including main) in spack

Growl1234:
> is there a way to ensure that cp2k uses tblite 0.7.0 or higher (including main) in spack

I think, just update Spack-package recipe with

```python
with when("+tblite"):
    depends_on("tblite build_system=cmake")
    depends_on("tblite@0.7:", when="@2027.1:")
```

IIRC `@2027.1:` here should include the development branch.

tools/toolchain/scripts/stage8/install_tblite.sh
Growl1234: I think, perhaps we can assign the package uploading to CI? For example, if one submitted a PR with updates of toolchain package and test `Regtest pdbg` passed, then CI could parse the URL and upload the tarball of correct name and SHA. Or, just design a `/cp2kci upload-package` that's valid for members (with triage access).

mkrack (approved):

Growl1234:
Appearently this has made [ifort tester](https://dashboard.cp2k.org/archive/intel-ifort-ssmp/index.html) work... ? 🤔

PR #5592, opened 2026-07-15
Colvar: add VoronoiIPZ to identify the absolute location of H3O+ ions along the z-axis

cpark1602:
Add a new collective variable (CV) named VoronoiIPZ to CP2K's colvar infrastructure. This collective variable is designed to identify the absolute location of ions in a slab system relative to a reference layer.

Specifically, it calculates the absolute location of H3O+ ions along the z-axis relative to the designated reference layer in slab geometries. This feature is useful for examining interface phenomena and ion-slab interactions.

The algorithm is ported from the original C++ implementation in the PLUMED software package.

Reference:
- https://doi.org/10.1021/acs.langmuir.4c05004

### DIFF
--- a/src/colvar_methods.F
+++ b/src/colvar_methods.F
@@ -28,7 +28,7 @@ MODULE colvar_methods
         plane_def_vec, plane_distance_colvar_id, plane_plane_angle_colvar_id, &
         population_colvar_id, qparm_colvar_id, reaction_path_colvar_id, ring_puckering_colvar_id, &
         rmsd_colvar_id, rotation_colvar_id, torsion_colvar_id, u_colvar_id, xyz_diag_colvar_id, &
-        xyz_outerdiag_colvar_id
+        xyz_outerdiag_colvar_id, voronoiipz_colvar_id
    USE constraint_fxd,                  ONLY: check_fixed_atom_cns_colv
    USE cp_log_handling,                 ONLY: cp_get_default_logger,&
                                               cp_logger_get_default_io_unit,&
@@ -101,6 +101,8 @@ MODULE colvar_methods
    IMPLICIT NONE
    PRIVATE
 
+   INTEGER :: i, j, n
+
    CHARACTER(len=*), PARAMETER, PRIVATE :: moduleN = 'colvar_methods'
    REAL(KIND=dp), PRIVATE, PARAMETER    :: tolerance_acos = 1.0E-5_dp
 
@@ -143,7 +145,8 @@ CONTAINS
       INTEGER, DIMENSION(:), POINTER                     :: iatms
       INTEGER, DIMENSION(:, :), POINTER                  :: p_bounds
       LOGICAL                                            :: check, use_mixed_energy
-      LOGICAL, DIMENSION(26)                             :: my_subsection
+      LOGICAL, DIMENSION(27)                             :: my_subsection
+
       REAL(dp), DIMENSION(:), POINTER                    :: s1, wei, weights
       REAL(dp), DIMENSION(:, :), POINTER                 :: p_range, s1v
       REAL(KIND=dp), DIMENSION(1)                        :: my_val
@@ -153,7 +156,7 @@ CONTAINS
       TYPE(enumeration_type), POINTER                    :: enum
       TYPE(keyword_type), POINTER                        :: keyword
       TYPE(section_type), POINTER                        :: section
-      TYPE(section_vals_type), POINTER :: acid_hyd_dist_section, acid_hyd_shell_section, &
+      TYPE(section_vals_type), POINTER :: voronoiipz_section, acid_hyd_dist_section, acid_hyd_shell_section, &
          angle_section, colvar_subsection, combine_section, coordination_section, dfunct_section, &
          distance_from_path_section, distance_section, frame_section, gyration_section, &
          HBP_section, hydronium_dist_section, hydronium_shell_section, mindist_section, &
@@ -181,6 +184,8 @@ CONTAINS
       hydronium_shell_section => section_vals_get_subs_vals(colvar_section, "HYDRONIUM_SHELL", i_rep_section=icol)
       hydronium_dist_section => section_vals_get_subs_vals(colvar_section, "HYDRONIUM_DISTANCE", i_rep_section=icol)
       acid_hyd_dist_section => section_vals_get_subs_vals(colvar_section, "ACID_HYDRONIUM_DISTANCE", i_rep_section=icol)
+      voronoiipz_section => section_vals_get_subs_vals(colvar_section, "VORONOIIPZ", i_rep_section=icol)
+
       acid_hyd_shell_section &
          => section_vals_get_subs_vals(colvar_section, "ACID_HYDRONIUM_SHELL", i_rep_section=icol)
       reaction_path_section => section_vals_get_subs_vals(colvar_section, "REACTION_PATH", i_rep_section=icol, &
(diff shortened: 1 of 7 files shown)


src/subsys/colvar_types.F
```diff
@@ -1333,10 +1372,25 @@ RECURSIVE SUBROUTINE colvar_release(colvar)
...
+         DEALLOCATE (colvar%voronoiipz_params)
+
+       ! ----
+
+
       CASE (no_colvar_id)
          ! Do nothing
       CASE default
-         CPABORT("Unknown colvar type")
+         CPABORT("")
```
fstein93, line 1393: Why have you removed the more specific message?

src/colvar_methods.F
```diff
@@ -1065,6 +1071,13 @@ RECURSIVE SUBROUTINE colvar_read(colvar, icol, colvar_section, para_env, cell)
                                      colvar%hydronium_dist_param%n_hydrogens, &
                                      colvar%hydronium_dist_param%i_oxygens, &
                                      colvar%hydronium_dist_param%i_hydrogens)
+      ELSE IF (my_subsection(27)) THEN
+         CALL colvar_create(colvar, voronoiipz_colvar_id)
+         IF (ASSOCIATED(voronoiipz_section)) THEN
+            CALL read_voronoiipz_colvars(voronoiipz_section, colvar)
+         ELSE
+            WRITE(*,*) "ERROR: voronoiipz_section is NULL at index:", icol
```
fstein93, line 1079: Maybe `CPWARN` is an option here?

src/colvar_methods.F
```diff
@@ -1808,8 +1825,14 @@ SUBROUTINE colvar_eval_glob_f(icolvar, force_env)
          CALL ring_puckering_colvar(colvar, cell, subsys=subsys)
       CASE (mindist_colvar_id)
          CALL mindist_colvar(colvar, cell, subsys=subsys)
+      !*CASE (voronoiipz_colvar_id)
+      !*   CALL colvar_eval_voronoiipz(colvar, cell, particles_set=subsys%particles%els)
```
fstein93, line 1829: Please remove it if unused

src/colvar_methods.F
```diff
@@ -6110,4 +6134,245 @@ END FUNCTION distance
 
    END SUBROUTINE HBP_colvar
 
+!     +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+!     COLVAR_VORONOIIPZ EVALUATION
+!     +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+SUBROUTINE colvar_eval_voronoiipz(colvar, cell, subsys, particles)
```
Growl1234, line 6140: Please make the formating of the new subroutines (including header notes) consistent with others. Ideally this can be done with `make_pretty.sh`, which also fixes most of the precommit errors (except the blank `CPABORT` which I really can't understand why you did it).

he-zilong (commented):
About the CPABORT messages turned empty: I suppose you started working on a local clone of the repository before my #5491 ? Please consider rebasing (not merging as we want to keep the git history linear) the intended changes to on top of the latest revision before submitting a PR. Read about some explanations [here](https://github.com/cp2k/cp2k/wiki/CP2K-CI).

Please make the title more indicative: something along the lines of `Colvar: add VoronoiIPZ to identify the absolute location of H3O+ ions along the z-axis` would be better.

Do not use `WRITE (*, *)` constructs as they are violating our coding convention for the output unit. If you do not need input control on the printed contents, the minimal acceptable practice would import the `cp_log_handling` module...
```
[...]
   USE cp_log_handling,                 ONLY: cp_get_default_logger,&
                                              cp_logger_get_default_io_unit
[...]
```
... and take the output unit as follows in the subroutine.
```
[...]
      INTEGER                                            :: output_unit
      TYPE(cp_logger_type), POINTER                      :: logger
[...]
      logger => cp_get_default_logger()
      output_unit = cp_logger_get_default_io_unit(logger)
      IF (output_unit > 0) WRITE (UNIT=output_unit, ...) ...
```

Literature citations should be added to [`src/common/bibliography.F`](https://github.com/cp2k/cp2k/blob/master/src/common/bibliography.F) and invoked with the `citations=[]` parameter of `CALL section_create()`.

Finally, a question regarding the colvar itself: does it rely on the fact that the H3O+ ion of interest have fixed atomic indices and/or O-H distances? In DFT calculations, the proton hopping phenomena (with grotthuss mechanism or similar) may make the identification of the H3O+ ion difficult.

oschuett:
Hi Chanbum, thanks a lot for contributing!

I realize you already got a lot of feedback, but it would be great if you could also add a regtest for the new feature.

src/subsys/colvar_types.F
cpark1602: Thanks for catching this! That was an accidental change on my part during cleanup. I will restore the "Unknown colvar type" message and push an updated commit shortly.

src/colvar_methods.F
cpark1602: Removed the commented-out code block.

src/colvar_methods.F
cpark1602: Thanks for the suggestion! Replaced the raw WRITE(*,*) statement with CALL CPWARN(...).

src/colvar_methods.F
DCM-Uni-Paderborn: The new routines and headers have been reformatted consistently, all blank `CPABORT` messages have been restored to specific diagnostics, and the CP2K precommit checks pass for all seven changed files.

src/subsys/colvar_types.F
DCM-Uni-Paderborn: Restored the specific `CPABORT` messages, including `Unknown colvar type`, throughout the affected dispatch and lifecycle paths.

src/colvar_methods.F
DCM-Uni-Paderborn: The raw `WRITE (*, *)` path has been removed entirely. The selected subsection is now asserted to be associated before it is read, so there is no warning-only condition left here.

src/colvar_methods.F
DCM-Uni-Paderborn: Removed the unused commented-out dispatch code.

DCM-Uni-Paderborn:
Addressed the review feedback in `bab55de8e3` after rebasing the PR onto the current `master`.

- Restored the specific `CPABORT` diagnostics, removed the raw `WRITE (*, *)` and dead commented code, and applied the CP2K formatting rules.
- Added Zhang and Xu (2025), DOI [10.1021/acs.langmuir.4c05004](https://doi.org/10.1021/acs.langmuir.4c05004), to `bibliography.F` and attached the citation to the input section.
- Added a FIST regression test with non-contiguous atom indices and finite-difference force checks. The debug test passes with both 1 MPI rank / 1 OpenMP thread and 2 MPI ranks / 2 OpenMP threads; all nine checked force components agree with the numerical derivatives.
- Corrected the local/global atom indexing, the Voronoi-weight chain rule, numerical stabilization, clone/release handling, and the former OpenMP races.

Regarding proton hopping: `GROUPA` and `GROUPB` define fixed candidate atom sets, not fixed O-H bonds or a fixed hydronium oxygen. Every hydrogen is assigned continuously to the candidate oxygen sites through distance-dependent Voronoi weights. The positive excess coordination can therefore move continuously between oxygen atoms during Grotthuss hopping without requiring fixed O-H distances or a fixed H3O+ atom index.

DCM-Uni-Paderborn:
From my side, the outstanding review points have been addressed and the regression and finite-difference force checks pass. I think this PR can be merged in its current form, provided @cpark1602 agrees with the changes made to the branch.

oschuett:
@cpark1602, do you have any objections?

cpark1602:
> @cpark1602, do you have any objections?

Dear all,

Thank you very much for your valuable comments and efforts. I will be back soon.

Best,
Chanbum

cpark1602:
> @cpark1602, do you have any objections?

@oschuett  Please merge this commits. Thank you all for the efforts!

PR #5597, opened 2026-07-16
Bump fortitude to version 0.9.0 and consider some of its new rules

fstein93:
For now, I am only into upgrading Fortitude. There will be some discussion o a few cases needed later.


dryndyk (commented):
I see mainly that 
```
IF (...)  &
   code
```
is replaced by
```
IF (...) THEN
   code
END IF
```

Can't say I like it, but OK.

src/hdf5_wrapper.F
```diff
@@ -35,6 +35,8 @@ MODULE hdf5_wrapper
 
    IMPLICIT NONE
 
+   PRIVATE
```
Growl1234, line 38: Looks like this breaks compilation of other code using this module.

fstein93:
> I see mainly that
> 
> ```
> IF (...)  &
>    code
> ```
> 
> is replaced by
> 
> ```
> IF (...) THEN
>    code
> END IF
> ```
> 
> Can't say I like it, but OK.

I understand your concern. I guess Fortitude's recommendation is to use the inline-version of `IF` only if its body stays within this line.

jpoto (approved):
looks fine to me

fstein93:
I am merging it now to prevent merge conflicts due to the huge number of files affected by it.

PR #5598, opened 2026-07-16
Move SCF MO k-point symmetry transform helper out of qs_wannier90

Growl1234:
Migrates the symmetry-related utility functions from `qs_wannier90.F` into a new module `kpoint_mo_symmetry_methods.F` to make them easier for reuse elsewhere, which would also be a prerequisite for the work on enabling WFN history for symmetry-reduced k-point grids.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -329,6 +329,7 @@ list(
   kpoint_k_r_trafo_simple.F
   kpoint_methods.F
   kpoint_mo_dump.F
+  kpoint_mo_symmetry_methods.F
   kpoint_transitional.F
   kpoint_types.F
   kpsym.F
--- a/src/kpoint_mo_symmetry_methods.F
+++ b/src/kpoint_mo_symmetry_methods.F
@@ -1,0 +1,238 @@
+!--------------------------------------------------------------------------------------------------!
+!   CP2K: A general program to perform molecular dynamics simulations                              !
+!   Copyright 2000-2026 CP2K developers group <https://cp2k.org>                                   !
+!                                                                                                  !
+!   SPDX-License-Identifier: GPL-2.0-or-later                                                      !
+!--------------------------------------------------------------------------------------------------!
+
+! **************************************************************************************************
+!> \brief Utilities for transforming k-point MO coefficients under symmetry operations
+! **************************************************************************************************
+MODULE kpoint_mo_symmetry_methods
+   USE cp_fm_types,                     ONLY: cp_fm_copy_general,&
+                                              cp_fm_get_info,&
+                                              cp_fm_get_submatrix,&
+                                              cp_fm_set_submatrix,&
+                                              cp_fm_type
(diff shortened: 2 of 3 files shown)


DCM-Uni-Paderborn (approved):
I found no blocking issues and approve this refactoring. The moved routines are semantically unchanged, the new module boundary is clean, and the formatting, local MPI build, and my own K290/SPGLIB Wannier90 tests passed. 

Wannier90-SCF-MO-reconstruction with K290 and SPGLIB: 
identical total energy -4.309643106506754 Ha
subspace deviation 3.331E-15

One non-blocking point for the follow-up WFN-history work: the inherited implementation currently materializes complete (nao x nmo) real and imaginary matrices on every MPI rank. That is unchanged by this PR and is not a blocker here, but it should be revisited before the transformation is called frequently for WFN-history updates in large-scale calculations.

DCM-Uni-Paderborn (approved):

Growl1234:
The Skala upstream [*replaced* their original files](https://huggingface.co/microsoft/skala-1.1/commit/5925d43f364e7a7d6cc35e6ce68d070b4ad640a9), which causes the sha256 mismatch and breaks the CI workflow (both toolchain and Spack will be affected).

@oschuett @awvwgk This needs to get attention as it affects the released v2026.2 tarball. It also turns out we have forgotten to upload the files to be downloaded in `toolchain/scripts/stage6/install_gauxc.sh` onto the CP2K download server.

awvwgk:
FYI @tvogels

oschuett:
> This needs to get attention as it affects the released v2026.2 tarball.

I intentionally did not move the files in `install_gauxc.sh` to cp2k.org because they went to [great lengths](https://github.com/cp2k/cp2k/blob/39fe4c47c6ffbb58ee35c86b875c1acad04cd37b/tools/toolchain/scripts/stage6/install_gauxc.sh#L32-L57) to download them from elsewhere. Of course, it's really unfortunate that the files got removed the day after our release. However, I don't think it makes sense for us to cut another release until the whole GauXC installation has been properly cleaned up.

Growl1234:
I didn't think of cutting one more release. What I want to say is, simply replacing an existing file is not a good habit as it most likely breaks some downstream workflows, which happens here (but may not only with CP2K). Therefore, my wish for upstream Skala developers would be to restore the old `skala-1.1.fun` and upload new file that names like `skala-1.1-1.fun` or `skala-1.1-retraced.fun` instead. *[Update: Thank you tvogels for reverting the replacement!]*

tvogels:
Sorry, we didn't realize replacing the checkpoint would break checksum checks for downstream users. The new version of the checkpoint produces identical numbers (up to numerical noise) but is significantly faster. 

We won't replace the files again -- we'll decide how to mark versions of the same functional. You can expect that one soon.

By the way, there was a hugging face outage today as well. Maybe I would recommend downloading the checkpoint and shipping it with the software.

PR #5602, opened 2026-07-17
CMake: Make native optimization an option

Growl1234:
Downstream Fedora RPM-build project has been enabling ppc64le build, which currently doesn't work because ppc64le does not know `-mtune=` flag. See https://src.fedoraproject.org/rpms/cp2k/pull-request/12.

This PR uses `-mcpu=native` as native flag for ppc64le, and drops `-mtune=generic` for ppc64le to use GCC default.

### DIFF
--- a/cmake/CompilerConfiguration.cmake
+++ b/cmake/CompilerConfiguration.cmake
@@ -53,12 +53,16 @@ add_compile_options(
 # -- Apple Silicon + GCC: -march=native expands internally to -march=apple-m1
 # (invalid). Use -mcpu=native instead.
 set(_CP2K_GNU_NATIVE_TUNE "-march=native;-mtune=native")
+set(_CP2K_GNU_GENERIC_TUNE "-mtune=generic")
 if(APPLE
    AND CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64"
    AND (CMAKE_Fortran_COMPILER_ID STREQUAL "GNU"
         OR CMAKE_C_COMPILER_ID STREQUAL "GNU"
         OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU"))
   set(_CP2K_GNU_NATIVE_TUNE "-mcpu=native")
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64|ppc64le")
+  set(_CP2K_GNU_NATIVE_TUNE "-mcpu=native")
+  set(_CP2K_GNU_GENERIC_TUNE "")
 endif()
 if(APPLE)
   add_definitions(-D__MACOSX)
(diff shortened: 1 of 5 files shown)


cmake/CompilerConfiguration.cmake
```diff
@@ -53,12 +53,16 @@ add_compile_options(
 # -- Apple Silicon + GCC: -march=native expands internally to -march=apple-m1
 # (invalid). Use -mcpu=native instead.
 set(_CP2K_GNU_NATIVE_TUNE "-march=native;-mtune=native")
+set(_CP2K_GNU_GENERIC_TUNE "-mtune=generic")
```
LecrisUT: Instead of these, please make these public cache variables with

```
set(CP2K_NATIVE_FLAGS "..." CACHE STRING "help message"
```
In order to include the if statements, make it a [gen-ex format](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html).

Or you can make it an option and use the normal `if` gating to avoid a super ugly default value.

Frankly though I would prefer for this to be explicitly opt in via cmake presets

Growl1234: I am personally a bit hesitant with it, as the flags are mainly limited to GNU compilers and supported architectures after all. I would prefer keep it as an internal logic and let CMake decide the flags to be used according to architecture detected. Moreover I'm not sure if core developers will be interested in testing more architectures.

oschuett:
I believe we discussed ppc64 before. The _Power_ architecture is not relevant to our users. Hence, we shouldn't spend any time on it. I'd suggest to simply not package cp2k for it.

LecrisUT:
More context from downstream PR, we have to disable the `march` and `mtune` regardless.

Growl1234:
Uh I realized the `-mtune` issue also happens with RISC-V.

Growl1234:
I can think of a more general way, to check if the flags are supported by target architectures and then apply them. This might make us get rid of the confusion.
```cmake
# Detect supported native/generic flags
set(_CP2K_GNU_NATIVE_TUNE "")
set(_CP2K_GNU_GENERIC_TUNE "")

check_c_compiler_flag("-mtune=generic" HAS_MTUNE_GENERIC)
if(HAS_MTUNE_GENERIC)
  set(_CP2K_GNU_GENERIC_TUNE "-mtune=generic")
endif()

set(_NATIVE_CANDIDATES "-march=native -mtune=native" "-mcpu=native")
foreach(_flags IN LISTS _NATIVE_CANDIDATES)
  string(REGEX REPLACE "[ -=]" "_" _cache_var "HAS_${_flags}")
  string(TOUPPER "${_cache_var}" _cache_var)
  check_c_compiler_flag("${_flags}" ${_cache_var})
  if(${_cache_var})
    string(REPLACE " " ";" _CP2K_GNU_NATIVE_TUNE "${_flags}")
    break()
  endif()
endforeach()
```

LecrisUT:
> More context from downstream PR, we have to disable the `march` and `mtune` regardless.

Again, the preferred solution is to have a global settings that downstream can turn off. We do not dictate what hardware we should be building on with that. For specific avx support in downstream https://fedoraproject.org/wiki/Changes/Optimized_Binaries_for_the_AMD64_Architecture_v2 approach should be used, but there is still no reference application we can check yet.

mtaillefumier (commented):
- While the PR is clean, I would argue that architecture specific options should not belong in 'CMakeLists.txt' and should stay within user control.  
- Power CPUs are arguably not the main target of cp2k however cp2k should properly run on these CPU in my view otherwise, as portability would only be limited to mainstream ARM and X86 families.
- These options are in general compiler dependent that's why only few projects directly handle these architecture specific flags manually.

I may approve the PR, however Do NOT make these options on by default and more importantly do not expose them to any spack package because it breaks the manual compilations settings mechanism that spack offer to users. 

Many of these compilations options should stay in user control that's why I already find the practice of modifying CMAKE_XXX_FLAGS internally very brittle.

mtaillefumier:
this issue https://github.com/spack/spack-packages/issues/5780 confirms that setting hardware specific options by default leads to some problem downstream. 

This PR solves this specific issue provided that there are not set on be default. We must also make sure that the build system does **NOT** overwrite users specified flags.

Growl1234:
I'll submit a fresh PR to remove `-march` and `-mtune` flags from internal CMake logic. With such patch the build type `Generic` would probably be no longer needed.

PR #5603, opened 2026-07-17
Remove gotos

### LINKED ISSUE #5565: RFC: Remove remaining GOTO statements and make GOTO a precommit error

fstein93:
This PR removes some of the `GOTO`s exluding the ones in `src/powell.F` (too entangled code, see #5565 ) and `srx/xc/xc_gauxc.F` (Taken care of in #5600 ). I also activated one more rule of Fortitude. I am currently waiting for Fortitude to implement a fix such that we can remove some of the new exceptions.
Closes #5565

### DIFF
--- a/src/graphcon.F
+++ b/src/graphcon.F
@@ -528,6 +528,7 @@ CONTAINS
       LOGICAL                                            :: first
 
       INTEGER                                            :: k, m, t
+      LOGICAL                                            :: q_is_identity
 
       IF (n == 1) RETURN
       IF (first) THEN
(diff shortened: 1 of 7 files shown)


fstein93:
FYI: I may not reply for the next two weeks because of holidays.

mkrack (approved):

PR #5615, opened 2026-07-21
Fix duplicated regtest duration accounting for multiple matchers

Growl1234:
A single regtest input may define multiple matchers, but each matcher result currently carries the full execution time of the input. As a result, directory totals will count the same run multiple times.

This change marks only the first matcher result as contributing to the execution time. Subsequent matcher results still report their individual status, but no longer duplicate the same duration.

### DIFF
--- a/tests/do_regtest.py
+++ b/tests/do_regtest.py
@@ -209,7 +209,7 @@ async def main() -> None:
     print("\n".join(r.error for r in all_results if r.error))
 
     print("\n------------------------------- Timings --------------------------------")
-    timings = sorted(r.duration for r in all_results)
+    timings = sorted(r.duration for r in all_results if r.duration)
     print('Plot: name="timings", title="Timing Distribution", ylabel="time [s]"')
     for p in (100, 99, 98, 95, 90, 80):
         y = percentile(timings, p / 100.0)
@@ -219,7 +219,9 @@ async def main() -> None:
     if cfg.flag_slow:
         print("\n" + "-" * 15 + "--------------- Slow Tests ---------------" + "-" * 15)
         threshold = 2 * percentile(timings, 0.95)
-        outliers = [r for r in all_results if r.duration > 0.95 * threshold]
+        outliers = [
+            r for r in all_results if r.duration and r.duration > 0.95 * threshold
+        ]
         maybe_slow = [r for r in outliers if r.fullname not in cfg.slow_suppressions]
         num_suppressed = len(outliers) - len(maybe_slow)
         rerun_tasks: List[Task[BatchResult]] = []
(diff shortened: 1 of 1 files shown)


tests/do_regtest.py
```diff
@@ -209,7 +209,8 @@ async def main() -> None:
     print("\n".join(r.error for r in all_results if r.error))
 
     print("\n------------------------------- Timings --------------------------------")
-    timings = sorted(r.duration for r in all_results)
+    timing_results = [r for r in all_results if r.counts_duration]
```
oschuett: Couldn't we just dedup instead of introducing the `counts_duration` ?
```suggestion
    timing_results = {r.fullname : r for r in all_results}.values()
```

Growl1234: Deduplication fixes the current aggregations, but not the output: every matcher would still report the full execution time, which is misleading and remains easy for downstream parsers to double-count. `counts_duration` explicitly distinguishes the single execution result from its additional matcher results and lets us suppress those duplicated timings.

oschuett: Ok, then how about we make `duration: Optional[float]` in `TestResult` and only set it on the first one.

This would be very similar to your proposal, but you could write `... if r.duration`.

Growl1234: OK I've changed it.

oschuett (approved):

tests/do_regtest.py
```diff
@@ -228,8 +230,14 @@ async def main() -> None:
...
-            rerun_times.update({r.fullname: r.duration for r in t.results})
-        stats = {r.fullname: [r.duration, rerun_times[r.fullname]] for r in maybe_slow}
+            rerun_times.update(
+                {r.fullname: r.duration for r in t.results if r.duration}
+            )
+        stats = {
+            r.fullname: [r.duration, rerun_times[r.fullname]]
+            for r in maybe_slow
+            if r.duration
+        }
```
oschuett: Thanks a lot for making the change!

One last nit pick (sorry). I think, we don't need the `if r.duration` here because `maybe_slow` was created from `outliers`, which in turn got already filtered.

It would be really nice it this could remain a one-liner.

```suggestion
        stats = {r.fullname: [r.duration, rerun_times[r.fullname]] for r in maybe_slow}
```

Growl1234: Looks like applying this change causes [type-checking errors](https://storage.googleapis.com/cp2k-ci/run-cp2k-misc-3cd3c26d_report.txt). `mypy` does not preserve the narrowing of `duration` through `outliers` and `maybe_slow`.

oschuett: Too bad. Thanks a lot for trying!

PR #5618, opened 2026-07-22
build: drop UCC requirement for cuSOLVERMp 0.7+

Stardust0831:
## Description

cuSOLVERMp 0.7 replaced the legacy CAL communication backend with NCCL, as already reflected by the version-dependent backend selection introduced in #4766. However, CP2K still found and linked UCC unconditionally in both the
CMake and toolchain paths. This prevented a current cuSOLVERMp installation from being used unless the unrelated legacy UCC/UCX stack was also installed.

This change completes the version-dependent dependency selection:

- cuSOLVERMp 0.7 and newer find and link NCCL only;
- older cuSOLVERMp releases continue to find and link CAL plus UCC/UCX;
- the CMake summary, toolchain setup, and documentation report the matching dependencies.

The backend cache value is forced to the detected result so that reconfiguring the same build directory with a different cuSOLVERMp version cannot retain a stale communication backend.

## Verification

A stacked integration branch, including the follow-up Spack support planned as a separate PR, was built and tested on two NVIDIA V100 GPUs:

https://github.com/Stardust0831/cp2k/actions/runs/29884688275

The successful run provides the following artifacts and checks:

- Spack installed `cusolvermp@0.7.2.888` and NCCL, with no `ucc` package in the environment.
- `cp2k.psmp --version` reports `cusolvermp cusolvermp_nccl`.
- `ldd libcusolverMp.so.0` contains NCCL and CUDA libraries, and no UCC or UCX  libraries.
- Two MPI ranks were bound to separate V100 devices: rank 0 to GPU 0 and rank 1 to GPU 1.
- The real regression case executed `cp_fm_general_cusolver` and produced  `-31.187602969867200 Ha`, differing from the reference by `1.421e-14 Ha`.
- The complex regression case executed `cp_cfm_general_cusolver` and produced  `-31.461089087225645 Ha`, differing from the reference by `7.105e-15 Ha`.

UCX is present in the complete CP2K process through MPICH's selected network transport. It is not a dependency of `libcusolverMp.so.0`, and UCC is absent.

The version split was also checked independently of the GPU run:

- minimal CMake configurations verified the exact interface link libraries for both a synthetic 0.7 header (`cuSOLVERMp;NCCL`) and a synthetic 0.6 header  (`cuSOLVERMp;CAL;UCC`);
- the actual toolchain installer was run through legacy-to-0.7, 0.7-to-legacy, and legacy-to-0.7 transitions. Each generated setup and persisted  `toolchain.env` contained only the roots for the selected backend.

The four changed files also pass the CP2K precommit service:

```text
Summary: Found 4, skipped 0, checked 4, and failed 0 files.
Status: OK
```

### DIFF
--- a/cmake/modules/FindCuSolverMP.cmake
+++ b/cmake/modules/FindCuSolverMP.cmake
@@ -12,8 +12,6 @@
 include(FindPackageHandleStandardArgs)
 include(cp2k_utils)
 
-find_package(ucc REQUIRED)
-
 # First, find CuSolverMP library and headers
 cp2k_set_default_paths(CUSOLVER_MP "CUSOLVER_MP")
 cp2k_find_libraries(CUSOLVER_MP "cusolverMp")
@@ -41,12 +39,13 @@ if(CP2K_CUSOLVER_MP_INCLUDE_DIRS)
     find_package(Nccl REQUIRED)
     set(CP2K_CUSOLVERMP_USE_NCCL
         ON
-        CACHE BOOL "CuSolverMP uses NCCL for communication")
+        CACHE BOOL "CuSolverMP uses NCCL for communication" FORCE)
   else()
     find_package(Cal REQUIRED)
+    find_package(ucc REQUIRED)
     set(CP2K_CUSOLVERMP_USE_NCCL
         OFF
-        CACHE BOOL "CuSolverMP uses Cal for communication")
+        CACHE BOOL "CuSolverMP uses Cal for communication" FORCE)
   endif()
 endif()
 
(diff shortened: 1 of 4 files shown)


cmake/modules/FindCuSolverMP.cmake
```diff
@@ -41,12 +39,13 @@ if(CP2K_CUSOLVER_MP_INCLUDE_DIRS)
     find_package(Nccl REQUIRED)
     set(CP2K_CUSOLVERMP_USE_NCCL
         ON
-        CACHE BOOL "CuSolverMP uses NCCL for communication")
+        CACHE BOOL "CuSolverMP uses NCCL for communication" FORCE)
```
Growl1234, line 42: Why add `FORCE` here? This is a *cache* variable, able to be set by users.

cmake/modules/FindCuSolverMP.cmake
```diff
@@ -41,12 +39,13 @@ if(CP2K_CUSOLVER_MP_INCLUDE_DIRS)
...
   else()
     find_package(Cal REQUIRED)
+    find_package(ucc REQUIRED)
     set(CP2K_CUSOLVERMP_USE_NCCL
         OFF
-        CACHE BOOL "CuSolverMP uses Cal for communication")
+        CACHE BOOL "CuSolverMP uses Cal for communication" FORCE)
```
Growl1234, line 48: Same.

cmake/modules/FindCuSolverMP.cmake
Stardust0831: I added FORCE to avoid retaining a stale backend selection when reconfiguring the same build directory with a different cuSOLVERMp version. Without it, a cached OFF from 0.6 can survive after detecting 0.7, and vice versa, leaving the selected link interface inconsistent with the detected version.
Since the backend is derived from the cuSOLVERMp header rather than being an independent choice, would you prefer this to be a normal non-cache variable instead?

Growl1234:
I'd like to request reviews from some more people who have major contribution on cuSolverMp support for CP2K (I see from Git history).

jpoto:
Should there be backward compatibility?

Stardust0831:
> Should there be backward compatibility?

Thanks.
If you mean compatibility with cuSOLVERMp releases older than 0.7, yes. The PR keeps the legacy path: CMake selects CAL + UCC for versions below 0.7, and the toolchain continues to require CAL, UCC, and UCX there. Only cuSOLVERMp 0.7 and newer use NCCL without UCC.The synthetic CMake checks covered both 0.6 and 0.7, and the toolchain was also tested across transitions in both directions. Did you have another compatibility aspect in mind?

jpoto (approved):
I have not tested it, but it looks ok.

oschuett (approved):

PR #5624, opened 2026-07-24
build: add Spack support for cuSOLVERMp

Stardust0831:
This follows two merged prerequisite changes:

- cp2k/cp2k#5618 added version-aware CMake and toolchain support for the NCCL backend used by cuSOLVERMp 0.7+;
- spack/spack-packages#5647 marked standalone cuSOLVERMp source and binary artifacts as non-redistributable, preventing them from being published to public Spack mirrors and buildcaches by default.

## Description

Add cuSOLVERMp 0.7.2 as an opt-in dependency of the CP2K Spack/CMake build:

- add the `cusolvermp@0.7.2.888` root spec already provided by CP2K's pinned Spack package repository;
- make `--enable_feature cusolver_mp` enable both the Spack root and `CP2K_USE_CUSOLVER_MP`;
- reject cuSOLVERMp requests without CUDA or for a serial CP2K binary before starting a dependency build;
- stop CUDA builds from implicitly enabling `libxstream`, which is the OpenCL backend and remains enabled explicitly through `-opencl`.

No UCC root is added. With the backend split merged in #5618, cuSOLVERMp 0.7 uses its NCCL communication backend directly.

This Spack integration intentionally targets cuSOLVERMp 0.7.2. The current [builtin Spack recipe](https://github.com/spack/spack-packages/blob/57ae97f3307b6c0f15a18fa87e52a76bd331ba09/repos/spack_repo/builtin/packages/cusolvermp/package.py) provides only 0.7.0.833 and 0.7.2.888 and depends on NCCL; it does not provide a pre-0.7 release or the CAL and UCC/UCX dependency path required by those releases. Existing CMake and toolchain support for externally provided pre-0.7 installations remains unchanged.

## Verification

The PR source commit (`371eb754f1`) was built and tested on two NVIDIA V100 GPUs through the temporary validation commit `2d575e0da7`:

https://github.com/Stardust0831/cp2k/actions/runs/30113610004

The job completed successfully in 44m21s on a Slurm allocation with 16 CPU slots and two Tesla V100-SXM2-16GB devices.

The successful run verified:

- Spack concretized and installed `cusolvermp@0.7.2.888`, `nccl@2.29.7-1`, CUDA-enabled MPICH, DBCSR, OpenBLAS, ScaLAPACK, and FFTW;
- the concretized environment contains no `ucc` package and does not enable `libxstream`;
- DBCSR concretized as `+cuda ~opencl ~rocm` for CUDA architecture 70;
- CP2K configured with GCC 13.3.0 and CUDA 12.4, and `cp2k.psmp --version` reports `cusolvermp cusolvermp_nccl`;
- two MPI ranks were bound to separate V100 devices;
- real and complex regression cases executed `cp_fm_general_cusolver` and `cp_cfm_general_cusolver`, with energy deviations of `3.553e-15 Ha` and `0.000e+00 Ha`, respectively;
- `libcusolverMp.so.0` links NCCL and CUDA, but not UCC or UCX, while `cp2k.psmp` links neither UCC, OpenCL, nor libxstream.

UCX is present because MPICH selected it as its network transport. It is not a cuSOLVERMp dependency, and UCC is absent.

The temporary validation commit also carried an isolated workaround for preserving the explicitly selected `--gcc_version 13` on this cluster, whose system compiler is too old. That unrelated workaround is intentionally not part of this PR.

Additional focused checks:

- generated CUDA-only dependency selection: one enabled `cusolvermp` root, zero enabled `libxstream` roots, and zero enabled `ucc` roots;
- repeated feature arguments follow the final enable/disable state;
- missing-CUDA, serial-build, and `sdbg` with an overridden MPI mode fail with the intended diagnostics;
- CP2K precommit checked both changed files with zero failures.


make_cp2k.sh
```diff
@@ -387,7 +388,11 @@ while [[ $# -gt 0 ]]; do
             SED_PATTERN_LIST+=" -e '/\s*-\s+\"spla@/ ${SUBST}"
             SED_PATTERN_LIST+=" -e '/\s*-\s+\"sirius@/ ${SUBST}"
             ;;
-          cray_pm_accel_energy | cusolver_mp | spla_gemm_offloading | unified_memory)
+          cusolver_mp)
+            USE_CUSOLVER_MP="${ON_OFF}"
```
Growl1234:
```suggestion
            USE_CUSOLVER_MP="${ON_OFF}"
            CMAKE_FEATURE_FLAGS_GPU+=" -DCP2K_USE_${2^^}=${ON_OFF}"
```

make_cp2k.sh
```diff
@@ -387,7 +388,11 @@ while [[ $# -gt 0 ]]; do
...
+            SED_PATTERN_LIST+=" -e '/\s*-\s+\"cusolvermp@/ ${SUBST}"
+            ;;
```
Growl1234, line 395: How about handling `cusolver_mp` in the common `ace | cosma | deepmd | ...` branch and appending
```bash
CMAKE_FEATURE_FLAGS_GPU+=" -DCP2K_USE_${2^^}=${ON_OFF}"
```
there? This would avoid the cuSolverMp-specific `if` in the CUDA configuration block.

make_cp2k.sh
```diff
@@ -898,6 +915,9 @@ if ((CUDA_SM_CODE > 0)); then
   CMAKE_CUDA_FLAGS="-DCP2K_USE_ACCEL=CUDA"
   CMAKE_CUDA_FLAGS+=" -DCP2K_WITH_GPU=${GPU_MODEL}"
   CMAKE_CUDA_FLAGS+=" -DCMAKE_CUDA_ARCHITECTURES=${CUDA_SM_CODE}"
+  if [[ -n "${USE_CUSOLVER_MP}" ]]; then
+    CMAKE_FEATURE_FLAGS_GPU+=" -DCP2K_USE_CUSOLVER_MP=${USE_CUSOLVER_MP}"
+  fi
```
Growl1234: With above suggetion, this `if` clause will be no longer needed.
```suggestion
```

make_cp2k.sh
Growl1234: Can also add `pdbg` check here.

Growl1234 (commented):
Some minor comments :)

Better double-check if `-ef/-df all` has effect on this option (in principle it should do), e.g.
```bash
-ef cusolver_mp -df all
```

make_cp2k.sh
Stardust0831: Implemented in 00aacefd4d. `cusolver_mp` is now handled in the common dependency-selection branch and appends its CMake option directly to `CMAKE_FEATURE_FLAGS_GPU`. I also added it to the `all` handling so argument order is preserved: `-ef cusolver_mp -df all` ends disabled, while `-df all -ef cusolver_mp` ends enabled, for both the Spack root and the final GPU CMake option.

make_cp2k.sh
Stardust0831: Applied the suggestion in 8201e68624 and completed the corresponding flag handling in 00aacefd4d. The CUDA configuration no longer contains the cuSOLVERMp-specific `if` clause.

make_cp2k.sh
Stardust0831: Added the `pdbg` check in 00aacefd4d. Both MPI-parallel variants (`pdbg` and `psmp`) are now rejected when `MPI_MODE=no`; the focused diagnostic check passes.

Growl1234 (approved):

mkrack (changes requested):
`-df all / -ef all` corresponds to `CP2K_USE_EVERYTHING=OFF|ON` in the `CMakeLists.txt` file. The option `CP2K_USE_CUSOLVER_MP` is not triggered by `CP2K_USE_EVERYTHING` and thus it should not be defined/set by `-df all` or `-ef all`.

Stardust0831:
> `-df all / -ef all` corresponds to `CP2K_USE_EVERYTHING=OFF|ON` in the `CMakeLists.txt` file. The option `CP2K_USE_CUSOLVER_MP` is not triggered by `CP2K_USE_EVERYTHING` and thus it should not be defined/set by `-df all` or `-ef all`.

@mkrack Thanks.
Fixed in d1f87a6f97. `-ef/-df all` now changes only `CP2K_USE_EVERYTHING` and does not alter the cuSOLVERMp Spack root, `CP2K_USE_CUSOLVER_MP`, or its GPU CMake flag. Explicit `cusolver_mp` selection remains independent before or after `all`, matching the CMake option definition.

make_cp2k.sh
```diff
@@ -289,10 +290,14 @@ while [[ $# -gt 0 ]]; do
...
             ;;
-          ace | cosma | deepmd | dftd4 | dlaf | elpa | fftw3 | gauxc | greenx | hdf5 | libfci | \
-            libgint | libint2 | libsmeagol | libtorch | libxc | libxs | mimic | openpmd | pexsi | \
-            plumed | spglib | tblite | trexio | vori)
-            CMAKE_FEATURE_FLAGS+=" -DCP2K_USE_${2^^}=${ON_OFF}"
+          ace | cosma | cusolver_mp | deepmd | dftd4 | dlaf | elpa | fftw3 | gauxc | greenx | \
+            hdf5 | libfci | libgint | libint2 | libsmeagol | libtorch | libxc | libxs | mimic | \
+            openpmd | pexsi | plumed | spglib | tblite | trexio | vori)
+            if [[ "${2,,}" == "cusolver_mp" ]]; then
+              CMAKE_FEATURE_FLAGS_GPU+=" -DCP2K_USE_${2^^}=${ON_OFF}"
```
mkrack: I am wondering if it would be possible to move this update of `CMAKE_FEATURE_FLAGS_GPU` down to the `cusolver_mp` case and drop the if clause here?

Stardust0831: Implemented in 80bea04405. `cusolver_mp` now has its own top-level case, where `USE_CUSOLVER_MP`, `CMAKE_FEATURE_FLAGS_GPU`, and the Spack root selection are updated together. The common feature branch is back to an unconditional `CMAKE_FEATURE_FLAGS` update with no cuSOLVERMp-specific `if`.

mkrack (approved):

PR #5627, opened 2026-07-27
Initial commit of linearized real time propagation of the Bethe Salpeter equation (linRTBSE) and open-shell for the existing BSE module

MGraml:
Here, I propose the implementation of the linearized real-time Bethe–Salpeter equation (linRTBSE) (closed-shell and open-shell) as well as an open-shell extension to the linear-response (Casida-type) BSE, and the cross-wiring of the real-space RI (RI-RS) module to compute the linRTBSE kernel.

linRTBSE propagates the first-order density-matrix response Δρ inside an active MO window, both in the Tamm-Dancoff approximation or with the full (ABBA) coupling, within linear response using a Runge Kutta solver (rk4), which leads to much smaller cost than full RTBSE propagation.
Besides the time propagation, it is also possible to diagonalize the Liouvillian for eigenvalues in a Casida-like fashion for debugging purposes, which allows direct comparison with the existing linear-response BSE.

Besides, the LRBSE can now access open-shell systems and the RI-RS module is extended to take custom grid files (Credit to @RemiPasquier ).

The new input keywords, which activate the linearized propagation, are in `DFT%REAL_TIME_PROPAGATION%RTBSE`:

- `LINEARIZED_BSE_PROPAGATION` (alias `LRRTBSE`) - selects linRTBSE.
- `KERNEL_RI {DEFAULT,RS,AO}` - RI representation used for the Hartree and screened-exchange kernels.

The custom grids are specified in `PROPERTIES%BANDSTRUCTURE%GW%RI_RS`:

- `GRID_SELECT 3` with `GRID_FILE_SUFFIX`, in  - per-element
  user-provided real-space grids, read as `ri_rs_grid/<Element><suffix>`.

## Testing

New `tests/QS/regtest-rtbse-linearized` covering TDA and ABBA, closed and open shell, AO and RI-RS kernels, and restarts. Open-shell linRTBSE is covered by H₂ forced-LSD tests only; open-shell LRBSE additionally covers O₂. Open-shell excitation energies were checked against MolGW; open-shell oscillator strengths are not cross-validated yet.

## Status and known limitations

linRTBSE and open-shell BSE are work in progress - both warn at startup and results need careful checking before used for production.

- `TDA_SHIFT_TO_FIRST_PEAK` is not implemented for open shell and aborts.
- Only the δ-kick is implemented; continuous or pulsed E(t) coupling aborts.
- Open-shell oscillator strengths are not validated.

The restart path is planned to be reworked: this PR keeps the new MO-active restart confined to linRTBSE and leaves the full RT-BSE restart at its upstream format, so that the two can be unified later.

Best,
Max


Growl1234:
I'd suggest to split `QS/regtest-rtbse-linearized` into several small directories, so that the parallel mechanism can be leveraged, rather than running a single directory for 40min (including rerunning).

MGraml:
Thank you @Growl1234 for the suggestion, I think I have an idea how to split them in a logically sound way. I will push the change in a second

tests/SLOW_TESTS_SUPPRESSIONS
```diff
@@ -99,4 +99,17 @@ QS/regtest-gauxc-gpw-mech/H2P_NATIVE_SKALA_GPW_UKS_PBC_FORCE_DEBUG.inp
...
+# Linearized RT-BSE and open-shell BSE: each of these runs an SCF, a G0W0 step,
+# the BSE kernel setup and a real-time propagation in a single job, so they cannot
+# reach the generous time limit. The suite was already reduced for this: cutoffs and
+# SCF thresholds were relaxed, the propagation shortened, and redundant
+# coupling/kernel/spin combinations dropped. The remainder is irreducible.
+QS/regtest-rtbse-linearized/h2_linrtbse_tda.inp
+QS/regtest-rtbse-linearized-open-shell/h2_linrtbse_tda_full_rirs_open_shell.inp
+QS/regtest-rtbse-linearized-open-shell/h2_linrtbse_abba_rirs_open_shell.inp
+QS/regtest-rtbse-linearized-open-shell/h2_linrtbse_tda_open_shell_restart.inp
+QS/regtest-rtbse-linearized-restart/h2_linrtbse_tda_static_pol_shift_restart.inp
```
oschuett: I believe these suppressions are not needed anymore.

JWilhelm (approved):

tests/SLOW_TESTS_SUPPRESSIONS
MGraml: Yes, thanks for spotting that! I was too cautious there because of the timings on my local machine.

oschuett (approved):

Issue #5548, opened 2026-07-07
Expose architecture-specific compile options (`-march`, `-mtune`, etc) and track CPU instruction set support w/ regtests


he-zilong:
> > [The corresponding faq](https://www.cp2k.org/faq:hint_insufficiently_exploiting_cpu_extensions) may need to be migrated to the documentation as well...
> 
> Is the checklist there still relevant now, as CMake already uses `-march=native;-mtune=native` for Release and Debug builds, as well as `-mtune=generic` for Generic builds?

On second thought, a full-fledged explanation about what the default values are and how they can be adjusted is better as a section in [Build from Source](https://manual.cp2k.org/trunk/getting-started/build-from-source.html). (I presume it is about checking `gfortran --target-help` for `Known valid arguments for -march= option` and `Known valid arguments for -mtune= option`, then applying to toolchain `--target-cpu` option and the following line of `cmake/CompilerConfiguration.cmake`?)
https://github.com/cp2k/cp2k/blob/fe787632bd21d8bb579232615b8aeebc7927b9bc/cmake/CompilerConfiguration.cmake#L55

After this addition, it would be easier to explain the messages about "The compiler target flags..." (and perhaps "Setting real_kernel for ELPA failed...") by linking to the section. Then the old FAQ can be considered obsoleted.

_Originally posted by @he-zilong in https://github.com/cp2k/cp2k/issues/5541#issuecomment-4895332935_

----

> We should also consider to include the information about the compiler (version) support for a release version. Maybe, it is better to collect that information per release than continuously updating the [compiler support wiki page](https://www.cp2k.org/dev:compiler_support) which is often forgotten.
> 
> In addition to the support for compilers, the support for CPU architectures and instruction sets is also definitely worth clarifying. One of the frequent user questions is about a hint like this: (the outdated reference to ARCH file needs fixing as well...)
> ```
> *** HINT in environment.F:920 :: The compiler target flags (generic) used ***
> *** to build this binary cannot exploit all extensions of this CPU model  ***
> *** (x86_avx512). Consider compiler target flags as part of FCFLAGS and   ***
> *** CFLAGS (ARCH file).                                                   ***
> ```
> The AVX-512 instruction set is available on some new high-performance CPU (such as the 4th Gen AMD EPYC processors) that is likely only on the computing node and not on the login node of supercomputer clusters. I think the whole toolchain (especially math libraries like OpenBLAS) as well as CP2K itself should be checked for compatibility. 

[...]

PR #5630, opened 2026-07-29
CMake: Use CMake presets to handle native optimization flags

Growl1234:
Follows the discussion in #5602.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,25 @@
+{
+  "version": 5,
+  "configurePresets": [
+    {
+      "name": "native-gnu-x86_64",
+      "displayName": "GNU native build for x86_64",
+      "description": "Use -march=native with GNU compilers on x86_64 architecture.",
+      "cacheVariables": {
+        "CMAKE_C_FLAGS": "$penv{CFLAGS} -march=native",
+        "CMAKE_CXX_FLAGS": "$penv{CXXFLAGS} -march=native",
+        "CMAKE_Fortran_FLAGS": "$penv{FFLAGS} -march=native"
+      }
+    },
+    {
+      "name": "native-gnu-arm64",
+      "displayName": "GNU native build for Arm64",
+      "description": "Use -mcpu=native with GNU compilers on Arm64/AArch64, including GNU/Linux and Apple Silicon.",
+      "cacheVariables": {
+        "CMAKE_C_FLAGS": "$penv{CFLAGS} -mcpu=native",
+        "CMAKE_CXX_FLAGS": "$penv{CXXFLAGS} -mcpu=native",
+        "CMAKE_Fortran_FLAGS": "$penv{FFLAGS} -mcpu=native"
+      }
+    }
+  ]
+}
--- a/tools/docker/scripts/cmake_cp2k.sh
+++ b/tools/docker/scripts/cmake_cp2k.sh
@@ -178,7 +178,6 @@ elif [[ "${PROFILE}" == "toolchain_hip_"* ]] && [[ "${VERSION}" == "psmp" ]]; th
 elif [[ "${PROFILE}" == "toolchain_generic" ]] && [[ "${VERSION}" == "psmp" ]]; then
   cmake \
     -GNinja \
-    -DCMAKE_BUILD_TYPE="Generic" \
     -DCMAKE_INSTALL_PREFIX="${INSTALL_PREFIX}" \
     -DCP2K_USE_EVERYTHING=ON \
     -DCP2K_USE_DLAF=OFF \
--- a/tools/docker/scripts/cmake_cp2k.sh
+++ b/tools/docker/scripts/cmake_cp2k.sh
@@ -64,7 +64,7 @@ elif [[ "${PROFILE}" == "toolchain" ]] && [[ "${VERSION}" == "sdbg" ]]; then
   CMAKE_EXIT_CODE=$?
 
 elif [[ "${PROFILE}" == "toolchain" ]] && [[ "${VERSION}" == "psmp" ]]; then
-  cmake \
+  cmake --preset "native-gnu-x86_64" \
     -GNinja \
     -DCMAKE_INSTALL_PREFIX="${INSTALL_PREFIX}" \
     -DCP2K_USE_EVERYTHING=ON \
@@ -76,7 +76,7 @@ elif [[ "${PROFILE}" == "toolchain" ]] && [[ "${VERSION}" == "psmp" ]]; then
   CMAKE_EXIT_CODE=$?
 
 elif [[ "${PROFILE}" == "toolchain" ]] && [[ "${VERSION}" == "ssmp" ]]; then
-  cmake \
+  cmake --preset "native-gnu-x86_64" \
     -GNinja \
     -DCMAKE_INSTALL_PREFIX="${INSTALL_PREFIX}" \
     -DCP2K_USE_EVERYTHING=ON \
@@ -119,7 +119,7 @@ elif [[ "${PROFILE}" == "toolchain_intel" ]] && [[ "${VERSION}" == "ssmp" ]]; th
   CMAKE_EXIT_CODE=$?
 
 elif [[ "${PROFILE}" == "toolchain_arm64" ]] && [[ "${VERSION}" == "psmp" ]]; then
-  cmake \
+  cmake --preset "native-gnu-arm64" \
     -GNinja \
     -DCMAKE_INSTALL_PREFIX="${INSTALL_PREFIX}" \
     -DCP2K_USE_EVERYTHING=ON \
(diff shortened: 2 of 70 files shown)


mtaillefumier (approved):

oschuett:
It seems this makes all our builds _generic_ now. So, what's the proper way to enable `-march=native;-mtune=native`?

Growl1234:
Pass it through `CMAKE_<LANG>_FLAGS`?

mtaillefumier:
> Pass it through `CMAKE_<LANG>_FLAGS`?

Yes that the cleanest way to do this that won't break anything downstream. If this option is wanted then a mechanism must be provided and this mechanism **should not be ON by default**. they are several way to implement this, using build types can be a way to implement this as well (RELEASE_NATIVE for instance). At least it makes it clear that the user wants there options to be used.

Growl1234:
If there's a easier way to enable native optimization, it's worth consideration to apply it instead. That's why I once intended the changes in #5602; it's controllable through an option anyways, and it chooses the right native flags through detection so architecture-relevant issue like what happened on Power and RISC-V won't happen again. The only thing that make us hesitant is that it applies the flags *internally*.

That said, I think what `RELEASE_NATIVE` type does is similar to what `CP2K_ENABLE_NATIVE_OPTIMIZATION` does; such type will most likely still apply the compiler flag internally with controllable mechanism.

Regardless of above discussion, I want to say for now that we will start to need a clear documentation regarding target-specific flags (especially native opt), no matter if it's applied in the internal CMake mechanism or just left through `CMAKE_<LANG>_FLAGS`. It becomes important once we have decided not to enable native optimization globally (except Generic build type), because there must be many people who want to leverage the architecture-flags to gain performance and are used to the current layout.

@he-zilong @LecrisUT I also want to hear your opinion on this.

LecrisUT:
My grand recommendation is to move _all flags_ into cmake presets, except for truly _required_ flags that would otherwise break compilation (but even those would be fine not to bother with). As long as the consumer has control of how they can opt-in into flags, it should be fine, but that is still generally missing. I understand the want to give the users a most optimal performance by default, but that's also one of the reasons why the presets were introduced.

Growl1234:
I really like the idea of CMake presets, at least for the purpose of native optimization. They do not change the internal logic, while making such requirements more explicit and discoverable, and they are also easy to extend if more cases are really needed.

oschuett:
We considered _CMake presets_ in the past, but decided back then that it would just add one more layer of indirection.

However, in the meantime we have accumulated several places where CMake flags are being maintained:

- [make_cp2k.sh](https://github.com/cp2k/cp2k/blob/master/make_cp2k.sh)
- [build_cp2k.sh](https://github.com/cp2k/cp2k/blob/master/tools/toolchain/build_cp2k.sh)
- [cmake_cp2k.sh](https://github.com/cp2k/cp2k/blob/master/tools/docker/scripts/cmake_cp2k.sh)
- [Spack package](https://github.com/spack/spack-packages/blob/fbbb4f6243447fd907936eeec549a5597eb36aff/repos/spack_repo/builtin/packages/cp2k/package.py#L1226)

So, perhaps _CMake presets_ would be a way to dedup some of this?

he-zilong:
Yeah, the first part of #5548 may be taken care of by CMake presets provided sufficient options.

This may or may not be on topic, but do we need to consider at the same time the CMake flags for dependencies like OpenBLAS/fftw/ELPA? There are user demands on explicit AVX-512 support in these libraries, though I do not know how much improvement on performance it will allow for.

mtaillefumier:
CMake preset can solve many of these issues. I would not touch the spack package file at all as spack does not set the compilation options in the package file at all. So if the idea is to encapsulate -DCP2K_XX_XX options in presets and remove them from the package.py recipe then it is a no-go. presets are a good solution for people using cmake directly not for packaging software.

Growl1234:
If I was to leverage CMake presets, I would not include any user-facing `CP2K_XXX` feature options in them; those should remain under the control of users and downstream packaging. What presets could meaningfully express are build profiles that require specific compiler flags, such as a native-optimization profile applying the appropriate architecture-specific flags. I fully agree that Spack should continue to manage its own variants and compiler flags and should not depend on these presets.

Growl1234:
With these considerations in mind, I think we should also be careful not to overengineer CMake presets in CP2K. They should provide optional build-strategy profiles rather than predefined feature sets.

Growl1234:
I submitted an initial version of `CMakePresets.json` to be the place to manage native optimization building.

LecrisUT:
> If I was to leverage CMake presets, I would not include any user-facing `CP2K_XXX` feature options in them; those should remain under the control of users and downstream packaging.

I would recommend to add any relevant option setting. The user can combine presets with `-D` options (and all other flags) to override what they desire. The presets are meant for the developers to advertise various patterns both for the developers and the users. Advanced users like packagers may or may not use the presets depending on the control they need, but I would still recommend to make them as complete as possible (with various layers that can be merged).

But let's not loose track of the main objective here "Drop architecture optimization flags" or the broader "Move hard-coded compilation flags". The need for this is in agreement with all downstreams. How to keep the compatibility for the general user though, that is the flexible part which can be addressed in multiple ways, with the recommendation from CMake being the presets.

mtaillefumier:
> > If I was to leverage CMake presets, I would not include any user-facing `CP2K_XXX` feature options in them; those should remain under the control of users and downstream packaging.
> 
> I would recommend to add any relevant option setting. The user can combine presets with `-D` options (and all other flags) to override what they desire. The presets are meant for the developers to advertise various patterns both for the developers and the users. Advanced users like packagers may or may not use the presets depending on the control they need, but I would still recommend to make them as complete as possible (with various layers that can be merged).
> 
> But let's not loose track of the main objective here "Drop architecture optimization flags" or the broader "Move hard-coded compilation flags". The need for this is in agreement with all downstreams. How to keep the compatibility for the general user though, that is the flexible part which can be addressed in multiple ways, with the recommendation from CMake being the presets.

that's a sound solution. 

The issue comes from options that are directly tied to the hardware because they can cause a lot of problem downstream '-mtune' is one of them. Warnings are harmless in that regard. Other optimization options are also not problematic. So one option is to move these in presets configurations so that the scripts don't have to replicate everything all the time.

mtaillefumier (approved):

mtaillefumier:
> With these considerations in mind, I think we should also be careful not to overengineer CMake presets in CP2K. They should provide optional build-strategy profiles rather than predefined feature sets.

The current preset setup seems to cover what is desired.

tools/docker/scripts/cmake_cp2k.sh
```diff
@@ -178,7 +178,6 @@ elif [[ "${PROFILE}" == "toolchain_hip_"* ]] && [[ "${VERSION}" == "psmp" ]]; th
 elif [[ "${PROFILE}" == "toolchain_generic" ]] && [[ "${VERSION}" == "psmp" ]]; then
   cmake \
     -GNinja \
-    -DCMAKE_BUILD_TYPE="Generic" \
```
oschuett, line 181: Since generic will then be the default, could you please add `--preset=native-gnu-...` to all the others.

CMakePresets.json
```diff
@@ -0,0 +1,25 @@
...
+  "version": 5,
+  "configurePresets": [
+    {
+      "name": "native-gnu-x86_64",
+      "displayName": "GNU native build for x86_64",
+      "description": "Use -march=native with GNU compilers on x86_64 architecture.",
+      "cacheVariables": {
+        "CMAKE_C_FLAGS": "$penv{CFLAGS} -march=native",
+        "CMAKE_CXX_FLAGS": "$penv{CXXFLAGS} -march=native",
+        "CMAKE_Fortran_FLAGS": "$penv{FFLAGS} -march=native"
```
oschuett, line 11: Should we also add `-mtune=native`, like before?

Growl1234: Usually `-march=native` is enough for x86_64, because compiler will handle it and make both `-march` and `-mtune` the proper option, e.g. on my machine:
```
$ gcc -march=native -### -c -x c /dev/null 2>&1 |
    grep -oE -- '-m(arch|tune)=[^ "]+'
-march=alderlake
-mtune=alderlake
```
For comparison:
```
$ gcc -### -c -x c /dev/null 2>&1 |
    grep -oE -- '-m(arch|tune)=[^ "]+'
-march=x86-64-v3   # decided by how GCC is packaged by system; most use x86-64
-mtune=generic
```
```
$ gcc -march=native -mtune=native -### -c -x c /dev/null 2>&1 |
    grep -oE -- '-m(arch|tune)=[^ "]+'
-march=alderlake
-mtune=alderlake
```

he-zilong:
I cannot see how `CMakePresets.json` is used yet. And why json instead of, say, yaml or toml?

Growl1234:
> I cannot see how `CMakePresets.json` is used yet.

`CMakePresets.json` is consumed directly by CMake, for example:

```bash
cmake --preset "native-gnu-x86_64" [other options]
```

You can also see available presets through

```bash
cmake --list-presets
```

The JSON format is not a choice made by CP2K; it is the format specified and natively supported by CMake for presets. CMake recognizes `CMakePresets.json` and `CMakeUserPresets.json` directly and parses them according to its preset schema.[^1] Using YAML or TOML would require a CP2K-specific parsing or conversion layer and would no longer be a native CMake preset.

[^1]: https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html?utm_source=chatgpt.com "cmake-presets(7) — CMake 4.4.0 Documentation"

tools/docker/scripts/cmake_cp2k.sh
Growl1234: Done. That said, I didn't apply the native preset for debug building, Ubuntu, and Spack ssmp-static tests.

oschuett (approved):

Issue #2391, opened 2022-11-10
Feature request:  DFT+U scheme (Liechtenstein) with explicit U and J 


robinzyb:
Dear CP2K developers,

is that possible to add the rotationally invariant DFT+U scheme with explicit U and J, as introduced by Liechtenstein?
[Liechtenstein scheme](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.52.R5467)
To my knowledge, semiconducting MnO2 will open its band gap only if the explicit exchange interaction (J) is considered.
Hybrid functional is also nice, but it will be rather expensive for large systems.

See comparison between (a) and (c) in this paper (doi:  10.1103/PhysRevB.86.205126)
<img width="711" alt="Screen Shot 2022-11-10 at 20 50 12" src="https://user-images.githubusercontent.com/38876805/201095955-19cb7237-6941-4aa7-be98-954238f734b1.png">

mkrack:
The DFT+U approach of Liechtenstein is not yet implemented in CP2K/QS (DFT module Quickstep). This is still on my to-do list. A separate U and J treatment is only available via the PW_DFT interface to SIRIUS, currently.

mattatlincoln:
Dear Matthias,
I am aware of a group that has implemented this and am encouraging them to commit into the code base. Perhaps can merge some of the implementations together (they also implemented Lowdin forces...).
Matt



Dr. Matt Watkins | Professor of Chemical Physics
School of Mathematics and Physics
University of Lincoln. Brayford Pool, Lincoln, Lincolnshire. LN6 7TS
tel: +44 (0)1522 835867
mattatlincoln.github.io<https://mattatlincoln.github.io/>



                       
                        
The University of Lincoln, UK, is rated among the UK's top 20 universities for student satisfaction in the Guardian University Guide 2022 and the Complete University Guide 2022 and in the top 30 UK universities overall in the WhatUni Student Choice Awards 2020. We hold a top five-star rating in the QS Stars ratings system of global universities and are placed among the world's top 150 young universities in The Times Higher Education Young University Rankings 2021.
                         
The information in this e-mail and any attachments may be confidential. If you have received this email in error please notify the sender immediately and remove it from your system. Do not disclose the contents to another person or take copies.
​
Email is not secure and may contain viruses. The University of Lincoln makes every effort to ensure email is sent without viruses, but cannot guarantee this and recommends recipients take appropriate precautions.
​
[...]

PR #5631, opened 2026-07-29
Implement tensorial DFT+U+J and self consistent MTLR

ZiweiChai:
Summary
- Implements tensorial DFT+U+J with analytical forces.
- Extends the Lowdin formulation to DFT+U+J.
- Implements self-consistent minimum-tracking linear-response calculations of U and J.
- Adds the required input keywords, initialization, and SCF integration.
- Adds regression tests for tensorial and Lowdin MTLR, energies, and forces.

Testing
- make_pretty.sh: passed for all modified files
- Clean compilation: passed
- QS/regtest-mtlr: passed
- QS/regtest-plus_u: passed
- QS regression tests: passed


src/qs_kind_types.F
```diff
@@ -2182,6 +2305,59 @@ SUBROUTINE read_qs_kind(qs_kind, kind_section, para_env, force_env_section, &
...
+               END IF
+               IF (MAXVAL(tmp_perturbation_strength) - MINVAL(tmp_perturbation_strength) <= &
+                   100.0_dp*EPSILON(1.0_dp)) THEN
+                  CPABORT("MTLR perturbation strengths must contain distinct values.")
+               END IF
+               IF (ALLOCATED(qs_kind%dft_plus_u%perturbation_strength)) THEN
+                  DEALLOCATE (qs_kind%dft_plus_u%perturbation_strength)
+               END IF
+               ALLOCATE (qs_kind%dft_plus_u%perturbation_strength(SIZE(tmp_perturbation_strength)))
+               qs_kind%dft_plus_u%perturbation_strength = tmp_perturbation_strength
```
oschuett: To prevent automatic resizing:
```suggestion
               qs_kind%dft_plus_u%perturbation_strength(:) = tmp_perturbation_strength(:)
```

src/input_cp2k_dft.F
```diff
@@ -202,20 +202,43 @@ SUBROUTINE create_dft_section(section)
...
       CALL keyword_release(keyword)
 
+      CALL keyword_create(keyword, __LOCATION__, &
+                          name="EPS_U_J_LOOP", &
+                          description="Convergence threshold for the self-consistent update of U and J parameters "// &
+                          "in the minimum tracking linear response calculation.", &
+                          repeats=.FALSE., &
+                          n_var=1, &
+                          unit_str="hartree", &
+                          default_r_val=1.0E-2_dp, &
```
Growl1234, line 228: Is `0.01 Hartree` really what you want to by default?

```suggestion
                          default_r_val=cp_unit_to_cp2k(value=0.01_dp, unit_str="eV"), &
```

src/mtlr_u_j_methods.F
```diff
@@ -0,0 +1,566 @@
...
+      u_old(:) = 0.0_dp
+      j_old(:) = 0.0_dp
+      converged = .FALSE.
+      eps_u_j_loop = dft_control%eps_u_j_loop
+      max_mtlr_iter = dft_control%max_mtlr_iter
+
+      ! Ensure that the DFT+U+J machinery remains active during the
+      ! initial MTLR calculation, even when a parameter starts from zero.
+      DO ikind = 1, nkind
+         IF (.NOT. qs_kind_set(ikind)%dft_plus_u%do_mtlr) CYCLE
```
Growl1234: Could you guard this access with `ASSOCIATED(qs_kind_set(ikind)%dft_plus_u)`?

```suggestion
         IF (ASSOCIATED(qs_kind_set(ikind)%dft_plus_u)
            IF (.NOT. qs_kind_set(ikind)%dft_plus_u%do_mtlr) CYCLE
         END IF
```

src/mtlr_u_j_methods.F
```diff
@@ -0,0 +1,566 @@
...
+                     perturbation_strength(p_iter)*evolt, " eV"
+                  WRITE (UNIT=output_unit, FMT="(T2,78('='))")
+               END IF
+
+               dft_control%perturbation_strength = perturbation_strength(p_iter)
+               dft_control%mtlr_dft_with_perturbation = .TRUE.
+
+               IF (do_reference_scf) THEN
+                  CALL force_env_calc_energy_force(force_env, calc_force=.FALSE.)
+               END IF
```
Growl1234, line 276: Sorry if I misunderstood something, but I don't quite understand why `force_env_calc_energy_force` here is guarded by `do_reference_scf`. Should the perturbation SCF be executed unconditionally here? With `SCF_GUESS ATOMIC`, `do_reference_scf` is false, so no perturbed SCF is performed and the subsequent fit uses stale or uninitialized response data.

```suggestion
               CALL force_env_calc_energy_force(force_env, calc_force=.FALSE.)
```

src/input_cp2k_dft.F
ZiweiChai: Yes.

src/mtlr_u_j_methods.F
ZiweiChai: If SCF_GUESS is RESTART, then the reference SCF will be first performed and then the perturbed SCF runs will be restarted from the converged .wfn. So, the calculations will be more fast. But if one would like to have LR U and J calculated from the same ATOMIC guess, then the reference SCF is not needed. The calculations will be slow, but they will all start from the same ATOMIC references.

src/mtlr_u_j_methods.F
```diff
@@ -0,0 +1,568 @@
...
+      u_old(:) = 0.0_dp
+      j_old(:) = 0.0_dp
+      converged = .FALSE.
+      eps_u_j_loop = dft_control%eps_u_j_loop
+      max_mtlr_iter = dft_control%max_mtlr_iter
+
+      ! Ensure that the DFT+U+J machinery remains active during the
+      ! initial MTLR calculation, even when a parameter starts from zero.
+      DO ikind = 1, nkind
+         IF (ASSOCIATED(qs_kind_set(ikind)%dft_plus_u)
```
Growl1234: Sorry, my fault.
```suggestion
         IF (ASSOCIATED(qs_kind_set(ikind)%dft_plus_u) THEN
```

Growl1234:
Just a side note: I think it's time for us to prepare a document about DFT+U 🤔

CC: @he-zilong

he-zilong:
> Just a side note: I think it's time for us to prepare a document about DFT+U 🤔

Certainly. The biggest problems about DFT+U in CP2K are the lack of existing reusable parameters from literatures (note that the parameters for other softwares, in particular plane-wave implementations, do NOT count), and the lack of estabilished procedures to fit custom DFT+U parameters for CP2K. Thus, I expect that there are more to prepare for the documentation than simple notes about setting up the keywords...

Also, is this PR addressing the #2391 feature request?

src/mtlr_u_j_methods.F
Growl1234:
```suggestion
         IF (ASSOCIATED(qs_kind_set(ikind)%dft_plus_u)) THEN
```

Growl1234:
> Certainly. The biggest problems about DFT+U in CP2K are the lack of existing reusable parameters from literatures (note that the parameters for other softwares, in particular plane-wave implementations, do NOT count), and the lack of estabilished procedures to fit custom DFT+U parameters for CP2K. Thus, I expect that there are more to prepare for the documentation than simple notes about setting up the keywords...

I think the current PR is really a good start, as one will be able to calculate the U and J paramater through MTLR.

> Also, is this PR addressing the #2391 feature request?

I think, yes, to some extent, though limited to the Löwdin and tensorial DFT+U implementations. In my view, that would already be sufficient, since both the Löwdin and tensorial approaches are more robust than the Mulliken approach because of the inherent limitations of Mulliken population analysis.

What‘s still missing is k-point support. As I understand it, adding k-point support to the tensorial implementation should be easier than doing so for the Löwdin implementation, and fortunately there is already a [paper](https://doi.org/10.1021/acs.jctc.4c01515) touching it.

oschuett (approved):

Growl1234:
From my local tests the code is still easy to cause some runtime errors. I will take a look and see if we can make it robust.

ZiweiChai:
> From my local tests the code is still easy to cause some runtime errors. I will take a look and see if we can make it robust.

A line of code which I added in qs_kind_types.F caused the runtime error. I will submit a correction later together with another modification of new_pulay_mixing for correcting R_cond problem when performing geo_opt.

Growl1234:
> A line of code which I added in qs_kind_types.F caused the runtime error. I will submit a correction later together with another modification of new_pulay_mixing for correcting R_cond problem when performing geo_opt.

I also found some other points in MTLR code. I've tried to prepare and am currently testing a fix regarding `mtlr_u_j_methods.F`, and will ping you for requesting review once a PR is ready.

src/mtlr_u_j_methods.F
Growl1234: My tests have shown dropping `IF` clause here is necessary, otherwise it turns out an error when I explicitly set `SCF_GUESS ATOMIC`:
```
 *  \___/                      MTLR regression is singular.                    *
 * / \                                                  mtlr_u_j_methods.F:321 *
```

ZiweiChai:
> > A line of code which I added in qs_kind_types.F caused the runtime error. I will submit a correction later together with another modification of new_pulay_mixing for correcting R_cond problem when performing geo_opt.
> 
> I also found some other points in MTLR code. I've tried to prepare and am currently testing a fix regarding `mtlr_u_j_methods.F`, and will ping you for requesting review once a PR is ready.

Yes, this is a bug. I will fix it.

Growl1234:
You could have a look at #5638, where I tried to fixed the runtime errors and added some clearer `CPABORT`, and suggest your changes there as my consideration is mainly about coding itself, or if you already have a reasonable solution I can close it [not duplicating effort :) ]

src/mtlr_u_j_methods.F
Growl1234: I am reconsidering the current condition: when `SCF_GUESS` is set to `RESTART`, even if the initial calculation falls back to starting from scratch because no restart file exists, all subsequent calculations start from the previously converged wavefunction — effectively the saved `.wfn` file. In contrast, with `SCF_GUESS ATOMIC`, every calculation starts from an atomic-density guess. I have doubt if this discrepancy is desirable.

Would it be possible to apply an internal reuse of the reference wavefunction (from the very first, not-perturbed SCF) within the MTLR workflow to reduce the computational cost?

PR #5031, opened 2026-03-30
Energy Correction Harris functional with k-points

juerghutter:
Implements k-point eneabled Harris functional calculations (energy, forces, stress tensor) based on
a Gamma-point only KS density.
Enables output of a Kpoint restartable density matrix file from the Harris functional. (Gamma point - kpoints initial guess)

PR #5636, opened 2026-07-30
Adding restart documentation

jpoto:
I got some questions regarding restart and think adding some documentations will be helpful.


he-zilong:
Why is the crucial distinction between `.wfn` gamma-point wavefunction restart file and `.kp` k-point wavefunction restart file not mentioned? This page should not only discuss that, but also demonstrate how the energy correction Harris functional introduced in #5031 can be used to produce a wavefunction restart file. When I wrote the `How to make a SCF run converge` page I mentioned this briefly and expected elaboration elsewhere.
https://github.com/cp2k/cp2k/blob/678364bc47497351bf4295df58ee914d8434cc09/docs/methods/dft/convergence.md?plain=1#L113-L121

The input files provided in the documentation are laden with gratuitous preprocessor directives `@SET` and `@IF`; I think they need to be removed, and in the spirit of #4972 the usage of preprocessor directives need a separate page for explanation.

I am not a fan of the current page organization where both the wavefunction restart and motion restart are discussed together; the documentation under `methods/dft` is more suitable for the former as the latter is not restricted to QuickStep calculations.

jpoto:
@juerghutter @DCM-Uni-Paderborn 
Is there something important missing?

jpoto:
> Why is the crucial distinction between `.wfn` gamma-point wavefunction restart file and `.kp` k-point wavefunction restart file not mentioned? This page should not only discuss that, but also demonstrate how the energy correction Harris functional introduced in #5031 can be used to produce a wavefunction restart file. When I wrote the `How to make a SCF run converge` page I mentioned this briefly and expected elaboration elsewhere.
> 
> https://github.com/cp2k/cp2k/blob/678364bc47497351bf4295df58ee914d8434cc09/docs/methods/dft/convergence.md?plain=1#L113-L121
> 
> The input files provided in the documentation are laden with gratuitous preprocessor directives `@SET` and `@IF`; I think they need to be removed, and in the spirit of #4972 the usage of preprocessor directives need a separate page for explanation.
> 
> I am not a fan of the current page organization where both the wavefunction restart and motion restart are discussed together; the documentation under `methods/dft` is more suitable for the former as the latter is not restricted to QuickStep calculations.

I think it is better to keep it in one file to avoid duplication/confusion, but I moved it.

PR #5638, opened 2026-07-31
MTLR: Fix runtime errors

Growl1234:
Fixes several runtime errors in the MTLR implementation and adds validation for invalid or incomplete configurations. Most notably:

- Avoid a double-free of `PERTURBATION_STRENGTH`. The pointer returned by `section_vals_val_get` refers to storage owned by the input tree and must not be deallocated by `qs_kind_types`.
- Restrict all MTLR operations to KINDs with an enabled `&MTLR` section. This prevents dereferencing an unassociated `dft_plus_u` pointer and removes the need to add dummy `DFT_PLUS_U` sections to unrelated KINDs.
- Always run the perturbed SCF calculations. Previously, `SCF_GUESS ATOMIC` disabled both the reference SCF and the perturbed SCFs, producing identical response data and eventually a singular regression. A failed `RESTART` appeared to work only because it retained the restart control path after falling back to an atomic guess.
- Read `MAX_MTLR_LOOP` as an integer when writing the DFT summary.
- Validate the number of perturbation strengths and the initialization of the projected Hxc matrices.
- Validate tensorial atomic-projector indices before accessing the orbital array.
- Apply convergence checks only to MTLR-enabled KINDs.

These changes make MTLR fail early with a clear diagnostic for invalid input and prevent the observed null-pointer access, double-free, out-of-bounds access, and formatted-I/O errors.

### DIFF
--- a/src/atom_kind_orbitals.F
+++ b/src/atom_kind_orbitals.F
@@ -108,7 +108,8 @@ CONTAINS
          OPTIONAL, POINTER                               :: ao_coef
 
       INTEGER                                            :: i, ii, j, k, k1, k2, l, ll, m, mb, mo, &
-                                                            nr, nset, nsgf, valence_n, z
+                                                            nr, nset, nsgf, projector_index, &
+                                                            valence_n, z
       INTEGER, DIMENSION(0:lmat)                         :: nbb
       INTEGER, DIMENSION(0:lmat, 10)                     :: ncalc, ncore, nelem
       INTEGER, DIMENSION(0:lmat, 100)                    :: set_index, shell_index
@@ -523,10 +524,15 @@ CONTAINS
          CPASSERT(PRESENT(which_n))
          CPASSERT(which_l >= 0 .AND. which_l <= lmat)
          valence_n = COUNT(ncore(which_l, :) > 0) + which_l + 1
-         CPASSERT(which_n >= valence_n)
+         projector_index = which_n - valence_n + 1
+         IF (projector_index < 1 .OR. projector_index > SIZE(atom%orbitals%wfn, 2)) THEN
+            CALL cp_abort(__LOCATION__, &
+                          "The requested tensorial atomic projector N is outside "// &
+                          "the available orbital range for the selected L.")
+         END IF
          IF (ASSOCIATED(ao_coef)) DEALLOCATE (ao_coef)
          ALLOCATE (ao_coef(SIZE(atom%orbitals%wfn(:, 1, which_l))))
-         ao_coef(:) = atom%orbitals%wfn(:, which_n - valence_n + 1, which_l)
+         ao_coef(:) = atom%orbitals%wfn(:, projector_index, which_l)
          IF (MAXVAL(ABS(ao_coef)) <= 100.0_dp*EPSILON(1.0_dp)) THEN
             CALL cp_abort(__LOCATION__, &
                           "The requested atomic projector is zero. Check the "// &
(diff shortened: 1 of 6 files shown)


Growl1234:
There's also a bug occurring when entering Line search in OT CG.

ZiweiChai:
Thank you for fixing the runtime issues. I agree with several of the changes/corrections in this PR. However, the new in-memory MO reuse implementation changes the intended behavior of the MTLR SCF initialization. The two modes were meant to behave differently:
With SCF_GUESS RESTART, a reference SCF is performed and each perturbed SCF starts from the reference WFN.
With SCF_GUESS ATOMIC, each perturbed SCF should start from an atomic guess, without imposing a separate reference electronic structure. The original condition around the perturbed SCF call was indeed wrong, because it prevented the perturbed calculations from running in the ATOMIC case. That should be fixed. However, always performing a reference SCF and restoring reference_mos before each perturbation removes the original ATOMIC option. In the current version, SCF_GUESS ATOMIC only affects the first reference calculation, while all perturbed calculations still start from the stored reference MOs. These two choices are not necessarily equivalent. Starting from an atomic guess and starting from a converged reference electronic structure can lead to different SCF solutions, especially for systems with localized states, metastable electronic configurations, etc. I would therefore like users to keep the choice between the two initialization schemes. I have not tested the new MO-copying implementation. The existing WFN-restart route, on the other hand, has already been tested on several systems on my side and has worked reliably. The MO-based approach may also work if the complete SCF state is restored consistently and no additional problems are introduced, but this still needs to be demonstrated.
For this PR, I suggest keeping the independent runtime and safety fixes, while preserving the original ATOMIC and RESTART behaviors. The new in-memory MO restoration could be discussed and tested separately, with regression tests showing that it reproduces the established WFN-restart results. If the two implementations show any difference in practice, I would prefer to retain the established WFN-restart route. In all systems I have tested, the perturbed SCF calculations converge very quickly from the reference WFN, and the additional restart overhead is negligible compared with the overall MTLR calculation.

Growl1234:
Thanks for your explanation! I will move the commit out to a seperate branch. This thread is very worth discussing, as it will gain practice performance anyways; once it has been clear that such an implementation doesn't introduce regression, we can go back to it.

ZiweiChai (approved):
Thanks for separating the WFN-reuse implementation from this PR. I have reviewed the remaining runtime and safety fixes, and they look good to me.

Growl1234:
Looks like MTLR tests still occassionally fail due to SCF convergence issue.

ZiweiChai:
> Looks like MTLR tests still occassionally fail due to SCF convergence issue.

Which systems and computational setups showed these SCF convergence failures?

Growl1234:
> > Looks like MTLR tests still occassionally fail due to SCF convergence issue.
> 
> Which systems and computational setups showed these SCF convergence failures?

From dashboard, such as:
https://storage.googleapis.com/cp2k-ci/dashboard_spack-pdbg_report.txt
The two regtests occasionally encounter such failures, but I expect it not easy to investigate (may or may not be related to MTLR itself)

Growl1234:
Uh I just saw one commit from #5652 is trying to stabilize them

DCM-Uni-Paderborn:
Follow-up: the stabilization for the two MTLR regression inputs has now been merged via #5652. It limits the OT-DIIS history to `N_HISTORY_VEC 2` and raises `MAX_SCF` to 100. The optional `Spack pdbg` check passed on the final PR head; both inputs had also passed five consecutive local runs with two MPI ranks before merge.

PR #5640, opened 2026-07-31
MTLR: Enable WFN reuse

Growl1234:
Reuses a fixed reference wavefunction for the perturbed SCF calculations in the MTLR workflow.

Previously, the initialization depended on `SCF_GUESS`: the `RESTART` path repeatedly loaded the reference wavefunction from disk, whereas the `ATOMIC` path rebuilt an atomic-density guess for every perturbation. Besides unnecessary computational and I/O costs, this made the two paths behave inconsistently.

The updated workflow:

- performs an unperturbed reference SCF before each (U/J) iteration;
- stores the converged reference MOs in memory;
- restores the same reference MOs before every perturbation;
- uses `USE_PREV_WF` to construct the initial density from the restored MOs;
- synchronizes the restored dense MO coefficients with `mo_coeff_b` for OT calculations.

The solution from one perturbation is deliberately not reused by the next one, as doing so could introduce a dependence on the ordering of `PERTURBATION_STRENGTH`.

With this change, `SCF_GUESS` only determines how the first reference SCF is initialized. An explicit `WFN_RESTART_FILE_NAME` is therefore allowed again and is only used for that initial calculation.

The MTLR regression tests now disable wavefunction restart output and cover both `ATOMIC` and failed-`RESTART` fallback initialization, ensuring that subsequent calculations rely on the in-memory reference rather than a `.wfn` file.


Growl1234:
This is rebased on #5638 (that's why the commits from there are included) and is left as a draft for discussion (see https://github.com/cp2k/cp2k/pull/5638#issuecomment-5141385810). More practicial tests should be needed to make sure the design introduces no regression.

CC: @ZiweiChai

DCM-Uni-Paderborn:
Dependency note after #5652: `master` now stabilizes both MTLR regression inputs with `MAX_SCF 100` and `N_HISTORY_VEC 2`. This draft currently conflicts in those two inputs; when rebasing, please preserve these settings alongside the `USE_PREV_WF` and restart-output changes. The stabilization passed the optional `Spack pdbg` check and repeated local two-rank runs.

DCM-Uni-Paderborn (commented):
Thanks for putting this together. Reusing an in-memory reference wavefunction is an attractive direction, and the implementation looks carefully structured, particularly the restoration before each perturbation and the handling of mo_coeff_b for OT.
I may be missing some context regarding the intended MTLR semantics, so please treat the following as a concern for clarification rather than as a claim that the implementation is incorrect.
My main hesitation is that this PR appears to change the observable meaning of SCF_GUESS ATOMIC. With the current implementation, ATOMIC initializes the first reference SCF, but all perturbed calculations subsequently start from the converged reference MOs. This may be a perfectly reasonable definition for MTLR, but it is not necessarily equivalent to starting every perturbation from an atomic guess. In systems with localized or metastable electronic states, the two approaches could converge to different SCF solutions.
The existing H₂⁺ regression tests are encouraging, but they probably cannot exercise the cases where this distinction matters. Before merging, I would feel more comfortable with:
confirmation from the MTLR domain maintainers that the unified reference-WFN behavior is intended;
a comparison between an actual file-based WFN restart and the in-memory restoration;
a test with reversed PERTURBATION_STRENGTH ordering;
if feasible, one more representative system with localized or competing SCF solutions.
It may also be worth documenting or measuring the memory trade-off, since the additional MO snapshot can include both the full-matrix coefficients and mo_coeff_b for OT.
I am not opposed to the approach, and the current results look promising. My preference would simply be to keep the PR as a draft until the intended semantics and the equivalence of the restart paths have been confirmed. If the consensus is that every perturbation should use the same reference WFN and the additional validation supports this, I would be happy with merging it.

Growl1234:
@ZiweiChai I believe you are most familiar with MTLR code since you are the author, and would appreciate your support to add more regtests against different types of system under `QS/regtest-mtlr` in a seperate PR (I will rebase this one if things are ready).

ZiweiChai:
> @ZiweiChai I believe you are most familiar with MTLR code since you are the author, and would appreciate your support to add more regtests against different types of system under `QS/regtest-mtlr` in a seperate PR (I will rebase this one if things are ready).

Okay, I’ll take a look and work on this in the next few days. I’m currently snowed under with other work.

Growl1234:
@ZiweiChai Please have a final check. Then we can merge it.

ZiweiChai (approved):

PR #5641, opened 2026-07-31
Drop legacy dftd4-API compatibility

Growl1234:
Together with #5591, I start to think of also dropping the API compatibility of the old dftd4 version. Supporting several generations of the dftd4 API has introduced a substantial amount of conditional compilation into the D4 implementation. The affected branches differ not only in procedure signatures, but also in array ranks, charge interfaces, error handling, and model metadata. Keeping these paths in parallel makes the implementation harder to review and increases the chance that fixes or new functionality are applied to only one API variant.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -861,28 +861,12 @@ if(CP2K_USE_SPLA)
   endif()
 endif()
 
-# Sets CP2K_USE_DFTD4_V3 / CP2K_USE_DFTD4_V4_2 from dftd4_VERSION, which is
-# populated by dftd4-config-version.cmake either directly via
-# find_package(dftd4) or transitively via find_package(tblite).
-macro(cp2k_detect_dftd4_api)
-  if(dftd4_VERSION VERSION_LESS "4.0.0")
-    message(STATUS "DFTD4: found version ${dftd4_VERSION}, using v3.x API")
-    set(CP2K_USE_DFTD4_V3 ON)
-  else()
-    set(CP2K_USE_DFTD4_V3 OFF)
-  endif()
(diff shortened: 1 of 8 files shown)


mkrack (approved):
LGTM

he-zilong:
Could you please explicitly state in the relevant sections of the [Libraries](https://manual.cp2k.org/trunk/technologies/libraries.html) page that 2026.2 is the last version to support legacy APIs and a minimum version requirement for the DFT-D4 dependency will become effective?

Growl1234:
> Could you please explicitly state in the relevant sections of the [Libraries](https://manual.cp2k.org/trunk/technologies/libraries.html) page that 2026.2 is the last version to support legacy APIs and a minimum version requirement for the DFT-D4 dependency will become effective?

Surely I can, though hard to say how we deal with dependency version requirement in documentation. For example, we requires libxc >=7, while Ubuntu system still packages v5.2.3. Tblite has not converged to a fully stable ABI, hence 0.x version and why we have to start requiring tblite version in #5591 due to a small ABI change; it's not clear to me whether tblite v0.4.0~0.5.0 work with v2026.2. I didn't take too much care of it as I assumed users install CP2K with its dependencies through dedicated dependency-handling workflow, such as toolchain, Spack, or Conda-forge; but I agree it should be taken care if needed.

mtaillefumier (approved):

oschuett (approved):

mkrack:
It seems that the failing tests with `Spack pdbg` are not related to this PR.
@Growl1234 Do you think that we could merge this PR therefore anyway?

Growl1234:
I see no problem.

PR #5642, opened 2026-07-31
Floquet parallelized

ShridharShanbhag:
This PR makes the floquet code now parallelized across MPI groups split from the global communicator para_env.
We also change add a new method for finding the band structure of the Floquet engineered bands for the m=0 bands along with their weights. At zero drive, this gives back the band structure of the solid with all weights = 1.
This new result is printed in a new file. The band structure, density of states and Quasi-energies are now with respect to the VBM.
The regtests have been updated likewise to test the new and existing features.

Co-Authored-By: Claude


ShridharShanbhag:
@oschuett could you verify if there is a problem with the do_regtest.py file? The regtest failure report does not indicate a problem with this pull request but rather with the python file for regtesting

tests/QS/regtest-floquet/TEST_FILES.toml
```diff
@@ -6,6 +6,8 @@
 #
 # compute ground state for graphene with PBE and SZV
 # compute the floquet DOS and QUASI_ENERGIES at K-point
-"Floquet_KP.inp"  = [{matcher="Quasienergy", file="QUASI_ENERGIES.bs", tol=1.0E-06, ref=0.00534452},
-                     {matcher="Floquet_DOS", file="FLOQUET_DOS.out",   tol=1e-8,  ref=24.91}]
+"Floquet_KP.inp"  = [{matcher="Quasienergy", file="QUASI_ENERGIES.bs", tol=1.0E-06, ref=0.12018321},
+                     {matcher="Floquet_BS", file="FLOQUET_BANDSTRUCTURE.bs",   tol=1e-8,  ref=0.12018321},
+                     {matcher="Floquet_DOS", file="FLOQUET_DOS.out",   tol=1e-8,  ref=29.74}]
+                     {matcher="Floquet_OCC", file="FLOQUET_DOS.out",   tol=1e-8,  ref=0.2839}]
```
oschuett:
```suggestion
                     {matcher="Floquet_DOS", file="FLOQUET_DOS.out",   tol=1e-8,  ref=29.74},
                     {matcher="Floquet_OCC", file="FLOQUET_DOS.out",   tol=1e-8,  ref=0.2839}]
```

src/floquet_utils.F
```diff
@@ -321,337 +335,773 @@ SUBROUTINE build_off_diagonal_matrix(qs_env, bs_env, xkp, ispin, off_diag_m)
...
-         END IF
-      END IF
-      DEALLOCATE (vector_block_0, vector_block_1)
       CALL timestop(handle)
+   END SUBROUTINE compute_e_k_de_dk_dipole
 
-   END SUBROUTINE check_floquet_conversion
+! **************************************************************************************************
+!> \brief Finds the number of MPI ranks that share a physical node, determined by splitting the
+!>        given communicator into node-local communicators keyed on the hostname.
```
oschuett: A better way to do this would be `MPI_Comm_split_type(...,MPI_COMM_TYPE_SHARED,..)`.

ShridharShanbhag: I agree, however I see that the MPI_COMM_TYPE_SHARED does not exist in the wrapper we use mpiwrap/message_passing.F (not sure why). Do you think it would be a good Idea to include this?

oschuett: Oh, I didn't realize we don't have it yet.

I think, separating intra- from inter-node MPI-communication will become more common as we're moving to fatter nodes. So, I'd be very much in favor of adding `MPI_Comm_split_type` to our `mpiwrap`.

What do you think @fstein93?

fstein93: We should definitely add this aspect to our wrapper. This should NOT be hidden somewhere inside of `floquet_utils.F`.

fstein93: In the MPI-wrapper, one could even use the MPI primitives directly.

ShridharShanbhag: Okay, I will make these changes in this PR.

fstein93:
I have checked the MPI standard on the allowed split types. I would wrap the split types in a separate type with two components: the split type and the required info. The reason is that the split type may be also `MPI_COMM_TYPE_HW_GUIDED` or `MPI_COMM_TYPE_HW_GUIDED` (MPI 4.0) with the info handle being an info on the hardware level (L1, L2, L3, Socket etc.).

ShridharShanbhag:
@fstein93 I included a new mp_split_type that bundles the split_type (INTEGER) and info (mp_info_type). I could not find MPI_COMM_TYPE_HW_GUIDED being used anywhere in cp2k but when it is implemented, mp_split_type could then be used for that just as I currently use it for the SHARED type.

fstein93 (commented):
Can you add `INTENT`s to at least all non-pointer arguments? This serves the documentation of the procedure interfaces.

ShridharShanbhag:
The subroutine mp_comm_split_type already has `INTENT` on all the arguments. I have updated the floquet files to have INTENT for all the non-pointer arguments on all the subroutines.

oschuett (approved):

PR #5643, opened 2026-07-31
Use CMake preset for all docker builds and set default in make_cp2k.sh

oschuett:
Follow up to #5630

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -20,6 +20,16 @@
         "CMAKE_CXX_FLAGS": "$penv{CXXFLAGS} -mcpu=native",
         "CMAKE_Fortran_FLAGS": "$penv{FFLAGS} -mcpu=native"
       }
+    },
+    {
+      "name": "native-intel",
+      "displayName": "Intel native build",
+      "description": "Use -xHOST with Intel compilers.",
+      "cacheVariables": {
+        "CMAKE_C_FLAGS": "$penv{CFLAGS} -xHOST",
+        "CMAKE_CXX_FLAGS": "$penv{CXXFLAGS} -xHOST",
+        "CMAKE_Fortran_FLAGS": "$penv{FFLAGS} -xHOST"
+      }
     }
   ]
 }
--- a/make_cp2k.sh
+++ b/make_cp2k.sh
@@ -139,8 +139,8 @@ CMAKE_FEATURE_FLAGS="-DCP2K_BLAS_VENDOR=OpenBLAS" # LAPACK/BLAS from OpenBLAS by
 CMAKE_FEATURE_FLAGS+=" -DCP2K_USE_FFTW3=ON"       # FFTW3 is always activated unless explicitly disabled
 CMAKE_FEATURE_FLAG_MPI="-DCP2K_USE_MPI=ON"        # MPI is switched on by default
 CMAKE_FEATURE_FLAGS_GPU="-DCP2K_USE_SPLA_GEMM_OFFLOADING=ON"
-CMAKE_PRESET=""
-CMAKE_PRESET_ARGS=()
+CMAKE_PRESET="native-gnu-x86_64"
+CMAKE_PRESET_ARGS=(--preset "${CMAKE_PRESET}")
 CRAY="no"
 CUDA_SM_CODE=0
 GCC_VERSION="auto"
@@ -555,8 +555,13 @@ while [[ $# -gt 0 ]]; do
       ;;
     -ps | --preset)
       if (($# > 1)); then
-        CMAKE_PRESET="${2}"
-        CMAKE_PRESET_ARGS=(--preset "${CMAKE_PRESET}")
+        if [[ "${2}" == "none" ]]; then
+          CMAKE_PRESET=""
+          CMAKE_PRESET_ARGS=()
+        else
+          CMAKE_PRESET="${2}"
+          CMAKE_PRESET_ARGS=(--preset "${CMAKE_PRESET}")
+        fi
       else
         echo "ERROR: No CMake preset found for flag \"${1}\""
         ${EXIT_CMD} 1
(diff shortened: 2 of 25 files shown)


mkrack (commented):
I am wondering if all these additional `--preset` flags for `make_cp2k.sh` builds for `psmp|ssmp` binaries are really needed. `make_cp2k.sh` should define automatically a `native` preset for `Release` and a `debug` preset for `RelWithDebInfo` builds.

CMakePresets.json
```diff
@@ -2,24 +2,34 @@
   "version": 5,
   "configurePresets": [
     {
-      "name": "native-gnu-x86_64",
-      "displayName": "GNU native build for x86_64",
-      "description": "Use -march=native with GNU compilers on x86_64 architecture.",
+      "name": "native-gnu-linux",
```
Growl1234: Please note that only x86_64 and aarch64 architectures are happy with `-march=native`, others will need `-mcpu=native`; and aarch64 actually just parse `-march=native` as `-mcpu=native`. Although I understand that CP2K has no interest in other alternative architectures, I think we'd better find a way to avoid confusion.

Growl1234:
> I am wondering if all these additional `--preset` flags for `make_cp2k.sh` builds for `psmp|ssmp` binaries are really needed. `make_cp2k.sh` should define automatically a `native` preset for `Release` and a `debug` preset for `RelWithDebInfo` builds.

We hope to leave the choice to users, which is exactly what @mtaillefumier is arguing and why we instead pull in a CMake presets file 🤔

mkrack:
> We hope to leave the choice to users, which is exactly what @mtaillefumier is arguing and why we instead pull in a CMake presets file 🤔

I am not questioning the introduction of a preset file but the explicit addition of `--preset` flags with `make_cp2k.sh` in many places. The selection of the cp2k binary version should already define the corresponding preset automatically, i.e. a default preset for `ssmp|psmp` and another for `sdbg|pdbg` corresponding to `Release` and `RelWithDebugInfo` builds. It is then still up to the user to overwrite these default presets via the `--preset` flag. The addition of presets should make things more flexible but not less convenient.

CMakePresets.json
oschuett: Admittedly it's a mess. How did you figure out what these flags do on the various platforms? I couldn't find any useful docus. My main motivation for changing the name was to stay true to our [original behaviour](https://github.com/cp2k/cp2k/blob/b95687185d47417b39e6150adcf90807c7d45812/cmake/CompilerConfiguration.cmake#L56).

Growl1234: `gfortran --help=target` is a place to see all target-specific flags support by gfortran compiled on the corresponding architecture. Also we can see details from [GCC docs](https://gcc.gnu.org/onlinedocs/gcc/Target-Specific-Options.html), e.g. the options suitable for [x86_64](https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html) and [aarch64](https://gcc.gnu.org/onlinedocs/gcc/AArch64-Options.html).

oschuett:
> make_cp2k.sh should define automatically a native preset...

I've now set `native-gnu-linux` as default preset in `make_cp2k.sh`.

Growl1234:
> > make_cp2k.sh should define automatically a native preset...
> 
> I've now set `native-gnu-linux` as default preset in `make_cp2k.sh`.

It would be better to allow passing `--preset none` for ones who don't want to apply native flags. 🤔

CMakePresets.json
oschuett: Thanks! So it seems to me `-march=native` is the preferred flag to use in general, but it's broken on macOS.

Growl1234: > Thanks! So it seems to me `-march=native` is the preferred flag to use in general, but it's broken on macOS.

It also breaks some other architectures, for example:
1. RISC-V does not support either `-march=native` or `-mcpu=native`; it requires specified arhitecture option.
2. PowerPC does not have `-march` flag.

oschuett: Ok, RISC-V will certainly become relevant in the future. I'm starting to see your point. I'll revert the names.

mkrack:
> > > make_cp2k.sh should define automatically a native preset...
> > 
> > 
> > I've now set `native-gnu-linux` as default preset in `make_cp2k.sh`.
> 
> It would be better to allow passing `--preset none` for ones who don't want to apply native flags. 🤔

I am still missing the link between `preset` and `CP2K_BUILD_TYPE` in `make_cp2k.sh` concerning defaults. I anticipated that the introduction of presets would facilitate the compiler flag selection for different platforms for the given `CMAKE_BUILD_TYPE` and more importantly make that selection more robust as the target platform is known. So far, this benefit is not obvious to me. I am glad to learn more about it.

Growl1234:
Currently the CMake preset in CP2K does not actually link with build type, one of the reason is previously `${_CP2K_GNU_NATIVE_TUNE}` was applied to every build type except `Generic`.

oschuett:
> So far, this benefit is not obvious to me.

Are you saying that, introducing the presets might have been a mistake?

Otherwise, I'd prefer to merge this now to fix the [sdbg regtests](https://dashboard.cp2k.org/archive/current-sdbg/index.html). We can still add more finesse to `make_cp2k.sh` later.

mkrack:
> > So far, this benefit is not obvious to me.
> 
> Are you saying that, introducing the presets might have been a mistake?
> 
> Otherwise, I'd prefer to merge this now to fix the [sdbg regtests](https://dashboard.cp2k.org/archive/current-sdbg/index.html). We can still add more finesse to `make_cp2k.sh` later.

That's fine with me. I am still in the process to learn how to benefit from presets.

mkrack (approved):

mtaillefumier:
I am bit confused. 

- What I am saying is: do not set the `-mtune -mcpu` compilation flags by **default**. Let the user choose what they want to do with these two options. The default Release, RelWithDebugInfo (all the default cmake build) should be clear from these two options.  
- It is possible to add builds in the presets file with customized compilations flags: the doc is here https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#build-preset. For instance, it would be possible to define psmp or pdbg builds and set the compilation flags as wished.
- What `make_cp2k.sh` does is a different story and to be fair secondary for `spack` or any other package manager. Now I understand that some users prefer to use the script instead.

PR #5647, opened 2026-08-02
Fix shared SIRIUS and tblite link interfaces

DCM-Uni-Paderborn:
## Summary

- restore the shared SIRIUS C++ link interface through a CP2K wrapper target using `LINK_ONLY`
- build shared tblite against the external Spack DFTD3, DFTD4, MCTC, mstore, and toml-f dependency graph
- preserve bundled tblite subprojects for `ssmp-static` through the explicit `tblite~shared` constraint
- export documented C API entry points that GCC 16 otherwise keeps local in DFTD4, simple-dftd3, and ELPA

## Background

SIRIUS 7.11.1 links its Fortran library to `sirius_cxx` privately. Its package configuration restores that dependency for static builds, but not for shared builds. Consequently, CP2K linked only `libsirius.so` in the Rawhide Spack build, while the final executable also needs the link interface of `libsirius_cxx.so`.

The first revision restored that target and exposed a second issue: the local tblite recipe built bundled static DFTD3 and DFTD4 libraries, while shared SIRIUS used the external Spack libraries. Mixing both implementations in the final link produced hidden-symbol failures.

For shared builds, tblite now uses the same external dependency graph as SIRIUS. Static tblite keeps its bundled subprojects, which are required for the `ssmp-static` dependency graph. No tests or dependencies are disabled.

`LINK_ONLY` prevents the C++ target include and compile requirements from propagating into CP2K Fortran compilation. The SIRIUS Fortran package should ultimately export a complete shared link interface upstream. The small DFTD3, DFTD4, and ELPA visibility patches should likewise be fixed upstream; they expose only documented C API routines needed by the shared stack.

## Verification

- `make_pretty.sh` completed successfully for all changed files
- `git diff --check` completed successfully
- fresh Rawhide/GCC 16 Spack concretization produced one shared DFTD3, DFTD4, and MCTC graph for SIRIUS and tblite
- patched DFTD3, DFTD4, ELPA, tblite, and SIRIUS built successfully
- dynamic symbol checks confirmed all DFTD3, DFTD4, and ELPA entry points referenced by `libsirius_cxx.so`
- full CP2K CMake build linked `cp2k.pdbg` and all unit-test executables successfully
- separate `tblite~shared` concretization contains no external DFTD3, DFTD4, MCTC, or toml-f dependencies


Growl1234:
IIRC SIRIUS did so because there was an issue with downstream MKL discovery (https://github.com/electronic-structure/-SIRIUS/pull/971). I have concern whether restore the behavior from the CP2K side will introduce more issues than benefits.

Moreover we should take a look why toolchain building with SIRIUS (Rawhide pdbg) succeeds, ~which may be of help~ *[Edit: looks like it's because SIRIUS is built as static library in toolchain...]*.

Growl1234:
The new failure exposes another problem: the `s-dftd3` and `dftd4` that SIRIUS is using (pulled in from Spack) is not the same as what CP2K is using (via local recipe of tblite whose tarball already has bundled `s-dftd3` and `dftd4`). Unfortunately that local recipe is currently required if we want to include it into `ssmp-static` Spack building.

mtaillefumier (approved):
it should probably be fixed upstream as well

Growl1234:
Would suggest to submit public C API patches upstream.

DCM-Uni-Paderborn:
The Rawhide Spack build and the shared SIRIUS/tblite link stack completed successfully. The check failed only in the two recently added MTLR regression inputs because OT-DIIS became unstable under GCC 16 + MPICH.

Commit ce8bf381a5 limits the DIIS history and raises the SCF iteration reserve for those small tests. Both inputs pass locally with two MPI ranks, retain U/J values within the existing 0.01 eV tolerance, and pass CP2K precommit.

mkrack (commented):
The `ssmp-static` tester (and possibly others) should be run without cache to perform a test of the modified spack recipes.

Issue #4038, opened 2025-02-28
Analytic Poisson solver for 2d systems produces inconsistent results


AndrasPetho:
During my work on two dimensional (pbc XY) systems I have noticed that the Poisson solver gives inconsistent results for XY periodicity with the analytic keyword. This issue seems to effect versions 2023.2 onwards with 2023.1 still giving similar results, compared to the implicit solver. Setting the poisson solver to any other periodicity seems to return similar energies to the implicit solver, however generates a warning about the different periodic boundary between the cell and the poisson solver. 

I have noticed the issue on more complex systems but have reproduced it on a He+ in a large box with the following input files: (only difference is the solver)
[He_analytic_input.txt](https://github.com/user-attachments/files/19032645/He_analytic_input.txt)
[He_implicit_input.txt](https://github.com/user-attachments/files/19032644/He_implicit_input.txt)

It seems that the difference between the 2d analytic solver and implicit increases with system size with a small molecular junction consisting of 100 atoms having 8 a.u. difference in energy, this is not present in the 2023.1 version

Also attaching my outputs for the versions 8.2, 9.1, 2022.2 , 2023.1, 2023.2, 2024.1, 2024.2, 2024.3 for 2D and 2023.1 and 2023.2 for 3D, 1D and 0D periodicity (those would use the relevant periodic keyword for the poisson solver)   

v8.2

[HE_analytic_v82.txt](https://github.com/user-attachments/files/19032685/HE_analytic_v82.txt)
[HE_implicit_v82.txt](https://github.com/user-attachments/files/19032686/HE_implicit_v82.txt)

v9.1

[HE_analytic_v91.txt](https://github.com/user-attachments/files/19032701/HE_analytic_v91.txt)
[HE_implicit_v91.txt](https://github.com/user-attachments/files/19032700/HE_implicit_v91.txt)

v2022.2

[HE_implicit_v222.txt](https://github.com/user-attachments/files/19032703/HE_implicit_v222.txt)
[HE_analytic_v222.txt](https://github.com/user-attachments/files/19032702/HE_analytic_v222.txt)

v2023.1

[HE_implicit_v231.txt](https://github.com/user-attachments/files/19032707/HE_implicit_v231.txt)
[HE_analytic_v231.txt](https://github.com/user-attachments/files/19032708/HE_analytic_v231.txt)

v2023.2

[HE_implicit_v232.txt](https://github.com/user-attachments/files/19032710/HE_implicit_v232.txt)
[...]

PR #5651, opened 2026-08-03
Document Poisson solvers and charged-system convergence

### LINKED ISSUE #5632: Add docs for Poisson solvers and regtests for energy convergence of charged systems w.r.t. vacuum size
A crucial topic that is not yet covered by documentation is the Poisson solvers for different periodicities, how exactly a system with nonzero net charge is handled (#4442), and how the energy of a system converges with respect to the size of vacuum (if any) for neutral and charged cases.

This was originally sent to #4972 but did not receive response there. Now I think that its scope can be expanded beyond documentation as a set of tracking regtests could possibly resolve issues like #4038.

DCM-Uni-Paderborn:
## Summary

- add a dedicated DFT electrostatics page covering periodicity settings and supported Poisson solvers
- clarify that the wavelet solver requires FFTSG-compatible grid dimensions, which the default `EXTENDED_FFT_LENGTHS .FALSE.` guarantees independently of the selected FFT library
- document the homogeneous-background convention for charged 3D cells and the absence of a vacuum-independent limit for uncompensated charged 1D/2D systems
- add four focused `ANALYTIC` 2D regression tests for neutral and charged He at 10 and 20 Å cell lengths

## Motivation

Electrostatic boundary conditions are part of the physical model, but the manual did not explain how `CELL%PERIODIC`, `POISSON%PERIODIC`, and `IMPLICIT%BOUNDARY_CONDITIONS` interact. This also made charged-system energy comparisons across cell sizes or solvers easy to misinterpret.

The new tests track the neutral plateau and the expected cell-size dependence of an uncompensated charged 2D system independently. They avoid using the `IMPLICIT` input from #4038 as a 2D reference because its default `IMPLICIT%BOUNDARY_CONDITIONS PERIODIC` selects a 3D-periodic problem.

Closes #5632.

## Validation

- CP2K precommit checks for all nine changed files
- official regression-test driver with two MPI ranks: 4/4 tests passed
- one- and two-rank direct runs reproduced the reference energies
- generated the full input reference and rendered the new Sphinx page; all new input and bibliography links resolve


docs/methods/dft/electrostatics/index.md
```diff
@@ -0,0 +1,133 @@
...
+| `PERIODIC`  | 3D                            | Standard fully periodic reciprocal-space solution.                                                                                                                          |
+| `ANALYTIC`  | 0D, 1D, 2D                    | Analytic reciprocal-space Green's functions; convergence can be slow.                                                                                                       |
+| `MT`        | 0D, 2D                        | Martyna--Tuckerman image decoupling. The cell must be at least twice the extent of the complete electronic density. [](#Martyna1999)                                        |
+| `MULTIPOLE` | 0D                            | Fits the total charge with atom-centred Gaussians to decouple periodic images. [](#Bl%C3%B6chl1995)                                                                         |
+| `WAVELET`   | 0D, 2D, 3D                    | Requires `PREFERRED_FFT_LIBRARY FFTSG`; the density must decay to zero at nonperiodic cell faces. [](#Genovese2006) [](#Genovese2007)                                       |
```
fstein93: The wavelet solver requires the grid size to be compatible with `FFTSG`, not the library to be `FFTSG`. Thus, the keyword `EXTENDED_FFT_LENGTHS .FALSE.` (default)

fstein93 (changes requested):

fstein93 (approved):

PR #5652, opened 2026-08-03
Update GauXC to fix subgroup MPI collectives

### LINKED ISSUE #5537: GauXC: Fix H2_ONEDFT_PBE_CDFT_CI_NGROUPS.inp with 4 MPI ranks
Btw, the four-rank failure of `H2_ONEDFT_PBE_CDFT_CI_NGROUPS.inp` appears to be a GauXC issue rather than a CP2K one. In the OneDFT gather path, GauXC uses `rt.comm_rank()` / `rt.comm_size()` from the CDFT subgroup, but still performs collectives on `MPI_COMM_WORLD`:

```cpp
MPI_Gather(..., sendcounts.data(), ..., MPI_COMM_WORLD);
```

For `NGROUPS 2` on four ranks, the count/displacement arrays are sized for two ranks while the collectives involve all four. The surrounding code already uses `rt.comm()` for another gather, so these OneDFT collectives likely need to use the runtime communicator consistently.

_Originally posted by @Growl1234 in https://github.com/cp2k/cp2k/issues/5531#issuecomment-4885711746_

DCM-Uni-Paderborn:
## Summary

- Update the GauXC revision used by the CP2K toolchain to `ea0f1fc7`.
- Add the corresponding Spack version `dev20260717`.
- Use the same GauXC revision in both the parallel and serial Spack environments.
- Update the toolchain archive checksum.

This keeps the GauXC revisions used by the toolchain and the local Spack recipe consistent.

## Motivation

GauXC used `MPI_COMM_WORLD` for OneDFT gather/scatter collectives while CP2K supplied a subgroup communicator. With `NGROUPS 2` and more than two MPI ranks, communicator sizes and the associated count/displacement arrays could therefore disagree.

The issue was fixed upstream in wavefunction91/GauXC#217. Both GauXC revisions currently used by CP2K predate that fix, so standard toolchain and Spack installations do not yet include it.

## Validation

- Verified the SHA256 checksum of the updated GauXC archive.
- Verified that the existing CP2K-specific GauXC patches apply cleanly to the new revision.
- Checked the toolchain script, Spack package syntax, and both YAML environments.
- CP2K precommit checks: 4/4 passed.
- The upstream GauXC change includes a dedicated four-rank subgroup-communicator regression test.

Closes #5537


mkrack (approved):
LGTM

DCM-Uni-Paderborn:
The optional `Spack pdbg` check consistently exposed an SCF instability in the recently added MTLR regression tests under GCC 15 + MPICH; the GauXC build and GauXC regression tests themselves passed.

Commit af34ea3565 limits the OT-DIIS history for these small MTLR tests and raises the SCF iteration reserve. Both MTLR inputs then passed five consecutive local runs with two MPI ranks, and the resulting U/J values remain well within the existing 0.01 eV regression tolerance.

PR #5663, opened 2026-08-04
Extension of native xTB to flexible number of primitive Gaussians (Ma…

annahehn:
…risa Schneider)


src/xtb_coulomb.F
```diff
@@ -226,9 +231,14 @@ SUBROUTINE build_xtb_coulomb(qs_env, ks_matrix, rho, charges, mcharge, energy, &
          rcut = rcuta + rcutb
          dr = SQRT(SUM(rij(:)**2))
          CALL gamma_rab_sr(gammab, dr, ni, kappaa, etaa, nj, kappab, etab, kg, rcut)
-         gchrg(iatom, 1:ni, 1) = gchrg(iatom, 1:ni, 1) + MATMUL(gammab, charges(jatom, 1:nj))
+         IF (iatom == jatom) THEN
+            gchrg(iatom, 1:ni, 1) = gchrg(iatom, 1:ni, 1) + MATMUL(gammab, charges(jatom, 1:nj))
+         END IF
          IF (iatom /= jatom) THEN
```
jpoto, line 237: You could use an else here. 
Why did you remove the contributions iatom<jatom?

jpoto (commented):
In order to check internal consistency a debug gradient test would be good. 
There is still a problem with the tests.

DCM-Uni-Paderborn:
Addressed the review feedback in 016d6e3163:

- Restored the symmetric sab_xtbe neighbor list and the original accumulation of charge, force, and virial contributions for both atoms of a pair. The filtered/non-symmetric version in the original PR gave a DEBUG force-difference sum of 1.776e-5; after restoring consistent pair handling it is 5e-8 with a 1e-6 test tolerance.
- Reworked the flexible STO-to-GTO path to allocate from the maximum requested per-shell primitive count, validate counts in the supported range 1...6, pass each shell count through the existing get_sto_ng argument, and correctly combine variable-length primitive blocks during orthogonalization. The stored Gaussian-count array now matches the five available shell slots.
- Removed the unrelated tblite changes and the redundant get_sto_ng interface extension so that the PR remains scoped to STO_FLEX.
- Replaced the original CdS setup using MAX_SCF 1 and a missing restart with a genuinely converged SCF setup. The old reference energies represented the first unconverged SCF step, so they were replaced by converged values. The flexible energy test and the DEBUG gradient test are separate to avoid matching the energy of the final displaced geometry. The flexible energy tolerance is 1e-10 because repeated parallel runs differed by about 2.7e-11 Hartree; the non-flexible test remains at 1e-12.
- A full release build succeeds, and xTB/regtest-1 passes 20/20 tests with 2 MPI ranks and 2 OpenMP threads.

The numerical values in number_of_primitives were not changed by this follow-up.

oschuett:
@annahehn do you agree with the changes?

oschuett:
I think, the changes pushed by @DCM-Uni-Paderborn are pretty much in line with the original PR.

So, I'd suggest if @annahehn doesn't reply today, then we'll just merge this as is.

Issue #5298, opened 2026-05-26
RFC: Should USE_PREV_RHO_R be deprecated?


Growl1234:
`USE_PREV_RHO_R` seems misleading and possibly broken.

The name suggests that it reuses the previous real-space density. However, in `qs_wf_history_methods.F`, it appears to store the AO density matrix instead:
https://github.com/cp2k/cp2k/blob/f506c3be655993367ada9324c884b599c83f3cf6/src/qs_wf_history_methods.F#L536-L538
It does not seem to set `store_rho_r = .TRUE.`.

During extrapolation, it copies the previous AO density matrix but does not rebuild the real-space density as `USE_PREV_P` and other methods does:
https://github.com/cp2k/cp2k/blob/f506c3be655993367ada9324c884b599c83f3cf6/src/qs_wf_history_methods.F#L873-L874
Instead, the current branch only marks `rho_g` as valid:
https://github.com/cp2k/cp2k/blob/f506c3be655993367ada9324c884b599c83f3cf6/src/qs_wf_history_methods.F#L822-L823
This may leave inconsistent density representations and can lead to very poor initial guesses or even `KS energy abnormal value`.

Conceptually, a naive reuse of the previous real-space density would be questionable for changing geometries, since the density field itself does not automatically follow the moving atoms. Reusing the AO density matrix, as in `USE_PREV_P`, is more natural in CP2K because the atom-centered basis moves with the atoms. WFN-based extrapolation paths are usually much preferable, as they preserve the orbital subspace information rather than only a fixed grid density.

Given this, it may be better to deprecate the `USE_PREV_RHO_R` method in favor of the existing density-matrix-based and WFN-based extrapolation paths, unless there is a clear reason to keep and redesign it as a proper real-space-density predictor.

DCM-Uni-Paderborn:
This issue looks valid to me. The current implementation does not appear to
match the documented meaning of `USE_PREV_RHO_R`: it is documented as reusing
the previous real-space density, but the history setup stores the AO density
matrix instead. The extrapolation path then copies that AO density matrix but
does not rebuild the dependent density representations in the same way as
`USE_PREV_P`.

Yet, I think the immediate bug can probably be fixed in a small PR by making
`USE_PREV_RHO_R` behave consistently with what it currently stores, namely the
AO density matrix. In practice this means treating it like `USE_PREV_P` and
rebuilding the dependent density representations after copying the AO density
matrix.

[...]

PR #5666, opened 2026-08-04
wfn-extrap: Remove deprecated USE_PREV_RHO_R

Growl1234:
Based on #5298 and follow up to #5300.

This PR removes the `USE_PREV_RHO_R` extrapolation option, which was deprecated and treated as an alias of `USE_PREV_P` instead. Including the removal of the relevant tests.


DCM-Uni-Paderborn (approved):
LGTM

PR #5673, opened 2026-08-05
Split slow regtest directories and reorder TEST_DIRS

Growl1234:
This PR splits several long-running regtest directories into smaller, better-balanced batches:

- `QS/regtest-acc-1`
- `QS/regtest-acc-5`
- `QS/regtest-tddfpt-force-gapw`
- `xTB/regtest-tblite-gfn2`
- `DFTB/regtest-scc`

Each original directory is split into five subdirectories based on the test timings observed in the Spack and toolchain reports. This allows the batches to run concurrently and reduces their directory-level critical path from roughly 3–5 minutes to about one minute when enough workers are available.nRestart-dependent tests remain in the same directory and retain their original order.

No tests or reference values are changed.


oschuett:
Thanks a lot for taking care of our regtests!  What also helps is to move slow tests dirs up - aka the _big rocks strategy_.

oschuett:
I just recalled that, I actually wrote a script for that: [regtesting/optimize_test_dirs.py](https://github.com/cp2k/cp2k/blob/master/tools/regtesting/optimize_test_dirs.py).

Growl1234:
That's good news, so we could simply check whether that script work as it has stalled for years.

Growl1234:
Could it be of help when re-running slow test dirs?

Growl1234:
I had a look, only one line need modification:
```patch
diff --git a/tools/regtesting/optimize_test_dirs.py b/tools/regtesting/optimize_test_dirs.py
index cd594b4565..efcfe64410 100755
--- a/tools/regtesting/optimize_test_dirs.py
+++ b/tools/regtesting/optimize_test_dirs.py
@@ -14,7 +14,7 @@ def main() -> None:
     args = parser.parse_args()
 
     report = Path(args.regtest_report).read_text(encoding="utf8")
-    match = re.search(">>> (.*?)/UNIT\n", report)
+    match = re.search(r"^>>> (.*?)/UNIT/", report, re.MULTILINE)
     assert match
     report_basedir = match.group(1)
 
```

oschuett:
> Could it be of help when re-running slow test dirs?

In principle yes, but typically there are fewer reruns than workers and therefore the order doesn't matter.

On the topic, ideally there would be no reruns at all because all tests stay far away from the slowness threshold.

Growl1234:
Might miss something like tests that need more MPI ranks or need dependencies that toolchain doesn't cover e.g. DLAF, but I think generally it would be OK 🤔

PR #5674, opened 2026-08-05
Test CUDA 12.9.2 -> 13.1.2

mkrack:


mtaillefumier (approved):

mkrack:
SIRIUS [v7.11.1](https://github.com/electronic-structure/SIRIUS/blob/release/v7.11.1/cmake/sirius_cxxConfig.cmake.in#L101) (likewise in the branch [develop](https://github.com/electronic-structure/SIRIUS/blob/develop/cmake/sirius_cxxConfig.cmake.in#L108)) performs a `find_dependency(CUDA REQUIRED)` which causes a CMake warning treated as an error because of the CMake flag `-Werror=dev`. Currently, I worked around it by disabling the CMake developer warnings (`-Werror=dev` -> `-Wno-error=dev`).

@mtaillefumier Do you have a better solution for that problem?

PR #5681, opened 2026-08-05
Keep native SKALA CUDA inputs in FP32

DCM-Uni-Paderborn:
## Summary

- keep dynamic native SKALA density, density-gradient, and kinetic-energy tensors in FP32 on CUDA
- preserve the existing FP64 CPU path
- reuse cached leaf tensors only when their shape, device, and requested dtype match
- convert autograd gradients back to FP64 CPU tensors before they are consumed by CP2K

## Motivation

The CUDA path previously copied CP2K's FP64 feature arrays into FP64 CUDA leaf tensors even when the exported SKALA model evaluated its learned operations in FP32. This increased dynamic-tensor storage and memory traffic without adding useful model precision.

The change performs the FP64-to-FP32 conversion while resetting the cached CUDA leaf tensors. Static layout tensors, direct exchange-correlation-density output, CP2K potentials, and CPU execution retain their existing precision. Gradient readback is explicitly converted to FP64 so the Fortran interface and downstream accumulation remain unchanged.

This is a focused GPU performance and memory improvement toward #5439.

## Performance

With the CUDA-exported SKALA model on Spark (GB10), one GPU, and H2O32:

- total CP2K time: 1.7% lower
- native SKALA forward time: 3.1% lower
- sampled GPU time: 4.5% lower
- host memory: unchanged within measurement noise

The total-energy difference relative to the FP64-input CUDA path was `4.18e-8 Ha`. For H2O6, combining the current model export with this input change reduced total time by 5.6% and sampled GPU time by 24.4% relative to the previously published Rev1 path; the incremental FP32-input effect itself was approximately neutral in total time while reducing sampled GPU use by 6.7%.

## Validation

- CP2K precommit checks for all four changed files
- clean CMake build with MPI, CUDA LibTorch, and GauXC disabled on Spark, rebased onto current `master`
- single-GPU H2O6 and H2O32 energy/performance runs
- native SKALA H2 energy, forces, and analytical stress smoke tests
- published Rev1 CUDA model compatibility smoke test

No paper, regression-result, or benchmark-script changes are included.

### DIFF
--- a/src/skala_gpw_features.F
+++ b/src/skala_gpw_features.F
@@ -189,11 +189,12 @@ CONTAINS
 !> \param use_atom_chunks ...
 !> \param route_atom_chunks ...
 !> \param atom_partition ...
+!> \param use_float32_dynamic Store dynamic CUDA inputs in single precision.
 ! **************************************************************************************************
    SUBROUTINE skala_gpw_feature_build(features, rho_set, rho_r, particle_set, cell, &
                                       requires_grad, weights, requires_coordinate_grad, &
                                       requires_stress_grad, use_atom_chunks, route_atom_chunks, &
-                                      atom_partition)
+                                      atom_partition, use_float32_dynamic)
       TYPE(skala_gpw_feature_type), INTENT(INOUT)        :: features
       TYPE(xc_rho_set_type), INTENT(IN)                  :: rho_set
       TYPE(pw_r3d_rs_type), DIMENSION(:), INTENT(IN)     :: rho_r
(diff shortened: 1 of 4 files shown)


DCM-Uni-Paderborn:
Withdrawing this PR after molecular CUDA validation uncovered a correctness regression that is not covered by the current CI: the first Kohn-Sham matrix becomes NaN and CP2K can then report false convergence after one SCF step. Density-only, gradient-only, kinetic-only, and gradient-plus-kinetic FP32 variants all reproduce the failure on 2 x A40. The previously reported small timing gains therefore are not valid full-SCF results and do not justify the correctness risk. PR #5682 is independent of this change and has been merged.

PR #5692, opened 2026-08-07
pass on gpu path for spla

jpoto:
Right now the GPU version is ignored during installation.


mkrack (commented):
The `CUDA Pascal` test is failing.

Growl1234:
This might be an issue need to be fixed by SIRIUS. The CUDA discovery is placed behind SPLA.

Growl1234 (changes requested):
Generally, I don't think the issue worth a workaround that's heavy for toolchain scripts. Did you try modifying SIRIUS' CMake, which can be added to toolchain as a patch? Or even after modifying SIRIUS the GPU build will still fail for CP2K?

Growl1234 (approved):

PR #5693, opened 2026-08-07
Remove unused variables

mkrack:


Growl1234:
It's so strange that `spack sdbg` doesn't find include directory of BLAS...
```
  - BLAS
    - Vendor: OpenBLAS
    - Include directories: CP2K_BLAS_INCLUDE_DIRS-NOTFOUND
    - Libraries: /opt/cp2k/spack/spack/opt/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_pl/linux-zen2/openblas-0.3.33-fwwckabxsmmkc42oqgomdemxtashkkce/lib/libopenblas.so
```

And since #5675 the comment is no longer correct: https://github.com/cp2k/cp2k/blob/0b7123d0137e111eefc36787a5904a79b7b8be45/cmake/modules/FindBlas.cmake#L130-L132

mkrack:
@Growl1234 Thanks for addressing the problem with `spack sdbg` via #5694.

PR #5607, opened 2026-07-18
Libxc 7.0.0 -> 7.1.1

Growl1234:

PR #5694, opened 2026-08-07
CMake/Spack: Resolve OpenBLAS discovery

Growl1234:
Since #5675 introduced [the inclusion of `cblas.h`](https://github.com/cp2k/cp2k/blob/852b9a412ddc30c10b4546cf9c8743185fdcc096/src/torch_c_api.cpp#L21-L23), we must treat `CP2K_BLAS_INCLUDE_DIRS` as a required variable.

The `FindOpenBLAS.cmake` seems to lose the include directory discovery when OpenBLAS is built with CMake, as OpenBLAS CMake installs include headers under `include/openblas` rather `include` (not consistent with Makefile build...). The unconditional `cp2k_include_dirs(OPENBLAS "cblas.h")` then overrides the correct value the `pkg-config` of OpenBLAS already provides, as it search for `<prefix>/include/cblas.h`, `<prefix>/include/OPENBLAS/cblas.h`, and `<prefix>/OPENBLAS/cblas.h`, but not `<prefix>/include/openblas/cblas.h`.

Since CMake is explicitly marked as "experimental" by upstream, I also make `build_system=makefile` the requirement for OpenBLAS in Spack `yaml` file.

### DIFF
--- a/cmake/modules/FindOpenBLAS.cmake
+++ b/cmake/modules/FindOpenBLAS.cmake
@@ -32,7 +32,9 @@ if(NOT CP2K_OPENBLAS_FOUND)
   cp2k_find_libraries(OPENBLAS_THREADS64 "openblas64_threads;openblas64_omp")
 endif()
 
-cp2k_include_dirs(OPENBLAS "cblas.h")
+if(NOT CP2K_OPENBLAS_INCLUDE_DIRS)
+  cp2k_include_dirs(openblas "cblas.h")
+endif()
 
 # check if found
 if(CP2K_OPENBLAS_INCLUDE_DIRS)
--- a/cmake/modules/FindBlas.cmake
+++ b/cmake/modules/FindBlas.cmake
@@ -127,13 +127,10 @@ endif()
 # cleanup list (regularly contains empty items)
 list(FILTER CP2K_BLAS_LINK_LIBRARIES EXCLUDE REGEX "^$")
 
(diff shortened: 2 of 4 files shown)


mtaillefumier (approved):

mkrack (approved):

cmake/modules/FindOpenBLAS.cmake
```diff
@@ -32,7 +32,9 @@ if(NOT CP2K_OPENBLAS_FOUND)
   cp2k_find_libraries(OPENBLAS_THREADS64 "openblas64_threads;openblas64_omp")
 endif()
 
-cp2k_include_dirs(OPENBLAS "cblas.h")
+if(NOT CP2K_OPENBLAS_INCLUDE_DIRS)
+  cp2k_include_dirs(openblas "cblas.h")
```
Growl1234, line 36: Oh no I made a mistake here. It should still be upper-case, or it returns `CP2K_openblas_INCLUDE_DIRS`
```suggestion
  cp2k_include_dirs(OPENBLAS "cblas.h")
```

mkrack: Does this cause that the `spack ssmp-static` tester is now failing? (see #5696)

Growl1234: I suspect so.

Growl1234: Hmm from #5607 it looks like there are other reasons...

PR #5697, opened 2026-08-07
CMake/Spack: Enhance OpenBLAS discovery

Growl1234:
Using `cp2k_include_dirs(openblas "cblas.h")` is incorrect because it returns `CP2K_openblas_INCLUDE_DIRS` that cannot be used (CMake wants `CP2K_OPENBLAS_INCLUDE_DIRS`.

As simply reverting that change will again break the discovery of `include/openblas/cblas.h`, I modified the `cp2k_includr_dirs` function to make such discovery possible.

Follow up to #5694.

### DIFF
--- a/cmake/modules/FindOpenBLAS.cmake
+++ b/cmake/modules/FindOpenBLAS.cmake
@@ -33,7 +33,7 @@ if(NOT CP2K_OPENBLAS_FOUND)
 endif()
 
 if(NOT CP2K_OPENBLAS_INCLUDE_DIRS)
-  cp2k_include_dirs(openblas "cblas.h")
+  cp2k_include_dirs(OPENBLAS "cblas.h")
 endif()
 
 # check if found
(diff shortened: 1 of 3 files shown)


Growl1234:
@mkrack I guess the reason might be:
https://github.com/cp2k/cp2k/blob/966235b268060496e96ace6d137cbbbeb92d7812/make_cp2k.sh#L1449

with this CMake of CP2K skips OpenBLAS discovery via `set(CP2K_BLAS_FOUND TRUE)` so `CP2K_BLAS_INCLUDE_DIRS` is empty.

mkrack:
> @mkrack I guess the reason might be:
> 
> https://github.com/cp2k/cp2k/blob/966235b268060496e96ace6d137cbbbeb92d7812/make_cp2k.sh#L1449
> 
> with this CMake of CP2K skips OpenBLAS discovery via `set(CP2K_BLAS_FOUND TRUE)` so `CP2K_BLAS_INCLUDE_DIRS` is empty.

@Growl1234 Good catch, thanks!

mkrack (approved):

PR #5702, opened 2026-08-10
Revise spack build of libgint

mkrack:
- allow for passing cuda_arch to the libgint spack recipe
- disable libgint if CUDA is not available

### DIFF
--- a/tools/spack/spack_repo/cp2k_dev/packages/libgint/package.py
+++ b/tools/spack/spack_repo/cp2k_dev/packages/libgint/package.py
@@ -3,11 +3,12 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack_repo.builtin.build_systems.makefile import MakefilePackage
+from spack_repo.builtin.build_systems.cuda import CudaPackage
 
 from spack.package import *
 
 
-class Libgint(MakefilePackage):
+class Libgint(MakefilePackage, CudaPackage):
     """libGint: GPU library for four-centre two-electron integrals and
     Hartree–Fock exchange calculations."""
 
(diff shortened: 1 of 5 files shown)


tools/spack/spack_repo/cp2k_dev/packages/libgint/package.py
```diff
@@ -19,18 +20,34 @@ class Libgint(MakefilePackage):
...
         sha256="cc0dfeb6022ebfe0c3028a131045ff49b4c1005ad9cb77c5736fb3dac045b192",
     )
 
+    variant(
+        "cuda_arch",
+        values=CudaPackage.cuda_arch_values,
+        default="70",
+        multi=False,
+        description="CUDA architecture",
+    )
```
RMeli, line 36: You get this variant already by inheriting from `CudaPackage`. The redefinition is to set a default value?

tools/spack/spack_repo/cp2k_dev/packages/libgint/package.py
```diff
@@ -19,18 +20,34 @@ class Libgint(MakefilePackage):
...
+
     depends_on("cuda")
```
RMeli: You can remove this now that it is derived fro `CudaPackage`. The `CudaPackage` already has `depends_on("cuda", when="+cuda")`.  

> disable libgint if CUDA is not available

We might want to add
```
conflicts("~cuda")
```

tools/spack/spack_repo/cp2k_dev/packages/libgint/package.py
mkrack: Yes, to set the default `70` from the Makefile.

tools/spack/spack_repo/cp2k_dev/packages/libgint/package.py
mkrack: Thanks, makes sense.

RMeli (approved):

PR #5703, opened 2026-08-10
Jvp ftorch gauxc

jpoto:
@franzpoeschel @DCM-Uni-Paderborn 

I have separated the ftorch and gauxc installation/testing. 
It was necessary to update some reference values.


Growl1234:
I’m a bit confused. So what are the respective roles of GauXC and SKALA now? Would this make the corresponding workflow (including installation) cleaner?

jpoto:
So far everything was installed by the install_gauxc,sh script. 

In this pull request it is split up:
- install_skala.sh installs the the information (weights, ...) for the ML density functional. 
- install_skala_ftorch.sh installs the ftorch, which is necessary to evaluate it on CP2K grids.
- install_gauxc.sh additionally installs gauxc, which construct its own grid and evaluates the expressions based on basis information and density matrix from CP2K. Since the density functional was trained on this grid it is probably more accurate. 

Additionally the tests are split up in the ones that require gauxc and the ones that can also be run with ftorch.

make_cp2k.sh
```diff
@@ -281,8 +281,8 @@ while [[ $# -gt 0 ]]; do
             # Enable or disable all features
             CMAKE_FEATURE_FLAG_ALL="-DCP2K_USE_EVERYTHING=${ON_OFF}"
             for package in adios2 cosma deepmdkit dla-future dla-future-fortran elpa \
-              gauxc greenx hdf5 libfabric libfci libint libvdwxc libsmeagol libvori \
-              libxc libxs libxsmm mimic-mcl openpmd-api pace pexsi plumed py-torch sirius \
+              gauxc greenx hdf5 libfabric libfci libint libvdwxc libsmeagol libtorch libvori \
+              libxc libxs libxsmm mimic-mcl openpmd-api pace pexsi plumed py-torch scala sirius \
```
awvwgk:
```suggestion
              libxc libxs libxsmm mimic-mcl openpmd-api pace pexsi plumed py-torch skala sirius \
```

make_cp2k.sh
```diff
@@ -403,6 +403,12 @@ while [[ $# -gt 0 ]]; do
           none)
             # do nothing
             ;;
+          scala)
+            CMAKE_FEATURE_FLAGS+=" -DCP2K_USE_SCALA=${ON_OFF}"
```
awvwgk:
```suggestion
          skala)
            CMAKE_FEATURE_FLAGS+=" -DCP2K_USE_SKALA=${ON_OFF}"
```

tools/toolchain/scripts/stage6/install_skala_ftorch.sh
```diff
@@ -0,0 +1,425 @@
...
+      echo "Download attempt ${__attempt} for ${__filename} failed."
+      sleep $((__attempt * 5))
+    done
+    return 1
+  else
+    echo "${__outfile} is found and checksum is right"
+  fi
+}
+
+cat > skala_ftorch.cpp << 'SKALA_CXX_EOF'
```
awvwgk: Why would you keep the C++ and Fortran source files as HEREDOCs in the install script? This would make it impossible to use via other installers like spack if I understand correctly. These sources are available under MIT license (they look like the ones from our skala repo) and could be committed directly to the CP2K source code. If you prefer to download them from our skala repo, I am happy to create a release artifact for them as well.

awvwgk (commented):
Thanks for trying to disentangle those two dependencies. I see there are currently two somewhat independent routes for Skala in CP2K:

1. Via GauXC with Skala support
   - requires GauXC
   - requires libtorch
   - requires Skala checkpoints from HuggingFace
2. Via FTorch
   - requires example files from Skala repo
   - requires FTorch
   - requires libtorch
   - requires Skala checkpoints from HuggingFace

For the example files from the Skala repo, we are happy to provide a release archive or you can just commit them to the CP2K repository directly. What is your preference here?

While in principle these dependencies are all independent, the user experience of enabling Skala in CP2K would be suboptimal if we require `--with-{gauxc,skala-ftorch,ftorch,libtorch,skala-checkpoint}=true` to enable the full feature set, maybe there is a better way to handle those?

Also, how do you want to handle these two paths in spack? As far as I can see, we cover only the GauXC part there right now.

ping @jenswehner, @tvogels

Growl1234:
I think, it would be very nice if skala installation could be wrapped in one clean script. I also agree with @awvwgk, it's not suitable to directly put code in the installation script; either use another folder to explicitly contain code files or download a tarball including them.

tools/toolchain/scripts/stage6/install_skala_ftorch.sh
```diff
@@ -0,0 +1,136 @@
...
+
+      mkdir -p build_ftorch
+      cd build_ftorch
+
+      cmake "${BUILDDIR}/FTorch-${ftorch_ver}" \
+        -DCMAKE_INSTALL_PREFIX="${pkg_install_dir}" \
+        -DCMAKE_INSTALL_LIBDIR=lib \
+        -DCMAKE_C_COMPILER="${CC}" \
+        -DCMAKE_CXX_COMPILER="${CXX}" \
+        -DCMAKE_Fortran_COMPILER="${FC}" \
```
Growl1234, line 79: IIRC toolchain already defines the env variables `CC` `FC` etc internally, so these three lines might be redundant.

tools/toolchain/scripts/stage5/install_stage5.sh
```diff
@@ -5,4 +5,9 @@
 
 ./scripts/stage5/install_elpa.sh
 
+# Install skala model if needed by GauXC or skala_ftorch
+if [ "${with_gauxc}" != "__DONTUSE__" ] || [ "${with_skala_ftorch}" != "__DONTUSE__" ]; then
+  ./scripts/stage5/install_skala.sh
+fi
```
Growl1234, line 11: Please leave the judgement at top-level toolchain script. There's no need to add a second guard here, otherwise it brings another avoidable maintenance burden.

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -1117,6 +1121,24 @@ if [ "${ENABLE_GAUXC_CUTLASS}" = "__TRUE__" ]; then
...
+  echo "Info: Installing shared skala model for GauXC and/or skala_ftorch"
+  with_skala="__INSTALL__"
+fi
+
+# Set SKALA flag if either GAUXC or skala_ftorch is enabled
+if [ "${with_gauxc}" != "__DONTUSE__" ] || [ "${with_skala_ftorch}" != "__DONTUSE__" ]; then
+  export SKALA="__TRUE__"
+else
+  export SKALA="__FALSE__"
+fi
```
Growl1234, line 1074: Could this if clause be merged with above one?

```bash
if [ "${with_gauxc}" != "__DONTUSE__" ] || [ "${with_skala_ftorch}" != "__DONTUSE__" ]; then
  echo "Info: Installing shared skala model for GauXC and/or skala_ftorch"
  with_skala="__INSTALL__"
  export SKALA="__TRUE__"
else
  with_skala="__DONTUSE__"
  export SKALA="__FALSE__"
fi
```

tools/toolchain/scripts/stage5/install_skala.sh
Growl1234, line 1: Along with above two comments, here we should have `__INSTALL__ ` and `__DONTUSE__` two branches, the latter doing nothing but skipping.

make_cp2k.sh
```diff
@@ -303,8 +303,8 @@ while [[ $# -gt 0 ]]; do
             # Enable or disable all features
             CMAKE_FEATURE_FLAG_ALL="-DCP2K_USE_EVERYTHING=${ON_OFF}"
             for package in adios2 cosma deepmdkit dla-future dla-future-fortran elpa \
-              gauxc greenx hdf5 libfabric libfci libint libvdwxc libsmeagol libvori \
-              libxc libxs libxsmm mimic-mcl openpmd-api pace pexsi plumed py-torch sirius \
+              gauxc greenx hdf5 libfabric libfci libint libvdwxc libsmeagol libtorch libvori \
```
mkrack, line 306: Why adding `libtorch` here? There is no package `libtorch` in the spec list of the `cp2k_deps_p.yaml` file, only `py-torch`. Unfortunately, spec names and `CP2K_USE_...` flags do not match always.

make_cp2k.sh
```diff
@@ -425,6 +425,9 @@ while [[ $# -gt 0 ]]; do
           none)
             # do nothing
             ;;
+          skala_ftorch)
```
mkrack: If feature name and `CP2K_USE_...` flag are consistent and the feature triggers no side-effects or implies other dependencies, it is sufficient just to add the `skala_ftorch` feature to the list above `ace | cosma | ...`.

Growl1234 (approved):

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -1064,11 +1064,12 @@ fi
 if [ "${with_gauxc}" != "__DONTUSE__" ] || [ "${with_skala_ftorch}" != "__DONTUSE__" ]; then
   echo "Info: Installing shared skala model for GauXC or Skala_Ftorch"
   with_skala="__INSTALL__"
-  export with_skala
   export SKALA="__TRUE__"
 else
+  with_skala="__DONTUSE__"
   export SKALA="__FALSE__"
 fi
+export with_skala
```
Growl1234: I wonder why this is needed 🤔

Growl1234, line 1072: I wonder why this is needed 🤔

jpoto:
If no one sees a problem it could be merged. The cuda pascal test failures are in unrelated tests. 
For the toolchain there are now two options (skala-ftorch and gauxc), whereas spack only has the gauxc option. 

Based on this branch I tried adding GPU functionality to skala-ftorch:
https://github.com/jpoto/cp2k/commit/3073bff969bcdac2f4c3082f6b3980fd63a7f814
It works, but still needs cleaning up and I am not certain which libtorch cuda to select.

jpoto:
Can it be merged?

jpoto:
@DCM-Uni-Paderborn Any comments?
I think it can now be merged.

mkrack (commented):
What's about the `libtorch` in line 318 of `make_cp2k.sh`?

jpoto:
> What's about the `libtorch` in line 318 of `make_cp2k.sh`?

It is from an earlier attempt. I will remove it.

PR #5706, opened 2026-08-11
make_cp2k.sh: add option for performance test run

mkrack:
Add spack-based performance tester


oschuett:
Should we also add this new tester to the dashboard?

mkrack:
> Should we also add this new tester to the dashboard?

I would appreciate an inclusion to check the performance of spack-based CP2K binary builds continuously.

PR #5708, opened 2026-08-11
Guard LibXS JIT use against missing handles

DCM-Uni-Paderborn:
## Summary

- patch bundled LibXS 1.0.0 to require a valid JIT handle before retrieving an MKL DGEMM or SGEMM kernel
- execute a JIT kernel only when both its function pointer and handle are present
- use the existing XGEMM/BLAS fallback for incomplete JIT configurations
- include the patch in the toolchain checksum so cached installations are invalidated correctly

## Motivation

The Intel oneAPI (ssmp, ifort) dashboard originally aborted in `G0W0_H2O_PBE0_30_pts.inp` at LibXS' `config->jitter` assertion. A JIT create callback can return without populating the handle, so dispatch must not request or store a kernel in that case.

After that dispatch guard, the ifort-psmp check exposed a second path in `diamond_gpw_stress.inp`: while the shared GEMM configuration is updated during JIT warm-up, an OpenMP consumer can observe a JIT function pointer without its handle. Treating the function pointer and handle as a pair at the call site selects LibXS' existing fallback for this incomplete state.

The complete fix and regression tests are submitted upstream as hfp/libxs#27 (tracking issue: hfp/libxs#26). This PR remains version-bound to bundled LibXS 1.0.0 and does not change CP2K sources, regtest references, or tolerances.

## Testing

- upstream LibXS mock tests cover missing handles during DGEMM/SGEMM dispatch and call, including fallback execution
- focused LibXS CMake test passes with Linux/GCC/OpenMP
- Intel 2024.2 ifort ssmp: `QS/regtest-gw-cubic`, 13/13 correct; the original failing test returns `15.6066`
- Intel 2024.2 ifort psmp, 2 MPI x 2 OpenMP, portable CPU target: `QS/regtest-kp-hfx-ri-2`, 6/6 correct
- the psmp CI reproducer `diamond_gpw_stress.inp` returns `-10.23639573` without the LibXS assertion
- patch applies cleanly to the bundled LibXS 1.0.0 tarball
- `make_pretty.sh` and `git diff --check`: pass

### DIFF
--- a/tools/toolchain/scripts/stage4/install_libxs.sh
+++ b/tools/toolchain/scripts/stage4/install_libxs.sh
@@ -33,6 +33,8 @@ case "$with_libxs" in
 
       echo "Installing from scratch into ${pkg_install_dir}"
       cd libxs-${libxs_ver}
+      patch -l -p1 < "${SCRIPT_DIR}/stage4/libxs-${libxs_ver}-jit-handle.patch" \
+        > libxs_jit_handle.patch.log 2>&1 || tail_excerpt libxs_jit_handle.patch.log
       mkdir build && cd build
       cmake \
         -DCMAKE_INSTALL_PREFIX="${pkg_install_dir}" \
@@ -42,7 +44,8 @@ case "$with_libxs" in
         .. > configure.log 2>&1 || tail_excerpt configure.log
       make install -j $(get_nprocs) > make.log 2>&1 || tail_excerpt make.log
       cd ..
-      write_checksums "${install_lock_file}" "${SCRIPT_DIR}/stage4/$(basename ${SCRIPT_NAME})"
+      write_checksums "${install_lock_file}" "${SCRIPT_DIR}/stage4/$(basename ${SCRIPT_NAME})" \
+        "${SCRIPT_DIR}/stage4/libxs-${libxs_ver}-jit-handle.patch"
     fi
     ;;
   __SYSTEM__)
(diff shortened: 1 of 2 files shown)


Growl1234:
Please feel free to submit a PR upstream.

DCM-Uni-Paderborn:
Thanks. I have submitted the fix upstream as hfp/libxs#27, with a mock-backend regression test covering both DGEMM and SGEMM. The CP2K patch remains version-bound to bundled LibXS 1.0.0.

DCM-Uni-Paderborn:
The failing ifort-psmp check exposed a second incomplete-JIT path: `libxs_gemm_call` could observe a JIT function pointer while `config->jitter` was null during shared-config warm-up. The bundled 1.0.0 patch now requires the function pointer and handle as a pair at the call site and otherwise uses the existing XGEMM/BLAS fallback. The same change and DGEMM/SGEMM regression coverage were added to hfp/libxs#27. A local Intel 2024.2 ifort-psmp run with 2 MPI ranks x 2 OpenMP threads now passes all 6 tests in `QS/regtest-kp-hfx-ri-2`, including the previously aborting `diamond_gpw_stress.inp`.

DCM-Uni-Paderborn:
The current ifort-psmp rerun stopped updating at 295/464 directories without reporting a wrong or failed test. I reproduced the same patched Intel ifort-psmp image on x86_64 with 2 MPI ranks x 2 OpenMP threads: the five directories pending at the stall completed 29/29 tests correctly in 33 seconds, and a smoke run reached all 464 directories. The smoke run's only mismatch was the existing `si128-elpa.inp` ELPA-QR value on the AMD host (`2.89e-7` from the Intel-host reference), unrelated to LibXS. This points to a stalled CI worker rather than another PR regression, so I am retrying only the ifort-psmp check once.

mkrack (approved):

PR #5711, opened 2026-08-11
Remove the GOTOs in powell.F

fstein93:


fstein93:
@oschuett I am not sure why the prettifier tries to extend the doxygen string of one of the routines over and over again.

src/common/powell.F
```diff
@@ -336,577 +334,643 @@ SUBROUTINE newuob(n, npt, x, rhobeg, rhoend, maxfun, xbase, &
...
-         DO j = 1, n
-            distsq = distsq + (xpt(k, j) - xopt(j))**2
-         END DO
-         IF (distsq > rhosq) temp = temp*(distsq/rhosq)**3
-         IF (temp > detrat .AND. k /= ktemp) THEN
-            detrat = temp
-            knew = k
+               CALL finalize_optimization_after_failure
+            END IF
+            do_return = .TRUE.
```
Growl1234, line 735: `do_return` is undefined whenever `update_rho` takes the normal `rho > rhoend` branch, but both callers test it immediately afterwards. If it happens to be true, `newuob` returns without generating a new trial point or saving the updated state, while `opt%state` remains 2. This can repeatedly evaluate the same point and explains the widespread pdbg timeouts.

I think commit `f3b93ec` already had the robust form: pass `do_return` as an `INTENT(OUT)` argument and initialize it to `.FALSE. `on every call. Could we restore that implementation?

Growl1234 (changes requested):

Growl1234:
> @oschuett I am not sure why the prettifier tries to extend the doxygen string of one of the routines over and over again.

Most likely because the `fixcomments.pl` misreads `_function` in subroutine name as `FUNCTION` keyword.

src/common/powell.F
fstein93: I will just initialize `do_return` at the very beginning. As soon as it is set to `.TRUE.`, we are leaving anyways.

fstein93:
> > @oschuett I am not sure why the prettifier tries to extend the doxygen string of one of the routines over and over again.
> 
> Most likely because the `fixcomments.pl` misreads `_function` in subroutine name as `FUNCTION` keyword.

This sounds like a bug to me.

Growl1234 (approved):

mkrack (approved):

PR #5718, opened 2026-08-11
Write effective charges to additional GPW cube files

### LINKED ISSUE #4518: Gaussian cube file output is not written with correct effective nuclear charge
Dear CP2K developers,

Apologies if this is cluttering the backlog/to-do list, but yet another problem has been observed about undesirably missing information in the output files after a Quickstep calculation, this time concerning Gaussian cube files. As specified in official documentation of [Gaussian cubegen](https://gaussian.com/cubegen/) and explained in third-party manuals from e.g. [h5cube project](https://h5cube-spec.readthedocs.io/en/latest/cubeformat.html) and [VMD cube plugin](https://www.ks.uiuc.edu/Research/vmd/plugins/molfile/cubeplugin.html), the records of atoms should have (effective) nuclear charge after atomic number and before three coordinates. Like the proposed ECP addition to molden file in #4353 this is needed for wavefunction post-analyses.

However, the subroutine `pw_to_cube` for printing Gaussian cube files just writes `0._dp` which cannot be correct for either all-electron or pseudopotential/ECP basis sets (note that the example from [Bourke's notes](https://paulbourke.net/dataformats/cube/) is also incorrect in this regard):
https://github.com/cp2k/cp2k/blob/fc6cf0ff43c4c312e8f79ec4a7b3601c9eabcced/src/pw/realspace_grid_cube.F#L127-L131
In the wrapper `cp_pw_to_cube` only atomic numbers and positions are fetched:
https://github.com/cp2k/cp2k/blob/fc6cf0ff43c4c312e8f79ec4a7b3601c9eabcced/src/cp_realspace_grid_cube.F#L59-L71
And indeed `get_atomic_kind` in `src/subsys/atomic_kind_types.F` does not deal with effective nuclear charge.

The effective nuclear charge, which the cube file should contain, is present in the printout as `Z(eff)` that is requested by keyword `&FORCE_EVAL/&SUBSYS/&PRINT/&ATOMIC_COORDINATES` and handled by `write_qs_particle_coordinates`:
https://github.com/cp2k/cp2k/blob/fc6cf0ff43c4c312e8f79ec4a7b3601c9eabcced/src/particle_methods.F#L573-L585
So a possible solution would be calling `get_qs_kind` from `src/qs_kind_types.F` in the cube-making procedure.

Hopefully this little analysis could help, and thank you for the efforts in improvements.

DCM-Uni-Paderborn:
## Summary

- factor the existing GPW effective-core-charge lookup into a shared helper
- write effective charges to occupied and unoccupied MO, ELF, local-energy, local-stress, dielectric, and Dirichlet cube files
- extend the existing local-stress and implicit-Poisson regressions to validate the affected cube headers

## Root cause

#4584 taught the Cube writer to accept effective charges and enabled them for most GPW outputs. These post-SCF call sites still supplied only the particles, so the Cube writer fell back to zero for the charge field.

The shared helper reuses the same `get_qs_kind(..., zeff=...)` logic already used by the other GPW Cube outputs, including the existing potential-specific handling. Passing the charges at the GPW call sites also avoids adding a generic warning for the optional low-level argument, which would affect method-specific callers that do not expose the same charge information.

## Verification

- reproduced the five original missing charges with the existing Si8 local-stress setup: each header contained `0.000000` on current `master`
- confirmed those five headers contain the GTH-PBE-q4 effective charge `4.000000` after the fix
- added coverage for the three implicit-Poisson cube variants and confirmed the O header contains the GTH-PBE-q6 effective charge `6.000000`
- reproduced the implicit-Poisson reference energy `-17.219865889534653` Ha
- built `cp2k.psmp` with GCC 16 and OpenMPI on macOS
- passed the full two-rank `QS/regtest-gpw-8` suite (`26/26`)
- passed all CP2K pre-push checks on the changed files

Closes #4518

### DIFF
--- a/src/pw/realspace_grid_cube.F
+++ b/src/pw/realspace_grid_cube.F
@@ -121,8 +121,10 @@ CONTAINS
       IF (PRESENT(mpi_io)) parallel_write = mpi_io
       IF (PRESENT(max_file_size_mb)) my_max_file_size_mb = max_file_size_mb
       CPASSERT(my_max_file_size_mb >= 0)
-      CPWARN_IF(.NOT. parallel_write .AND. PRESENT(particles_z) .AND. .NOT. PRESENT(particles_zeff), &
-                missing_zeff_warning)
+      IF (.NOT. parallel_write .AND. PRESENT(particles_z) .AND. &
+          .NOT. PRESENT(particles_zeff)) THEN
+         CPWARN(missing_zeff_warning)
+      END IF
 
       my_stride = 1
       IF (PRESENT(stride)) THEN
@@ -706,7 +708,9 @@ CONTAINS
       !get rs grids and parallel envnment
       gid = grid%pw_grid%para%group
       my_rank = grid%pw_grid%para%group%mepos
-      CPWARN_IF(PRESENT(particles_z) .AND. .NOT. PRESENT(particles_zeff), missing_zeff_warning)
+      IF (PRESENT(particles_z) .AND. .NOT. PRESENT(particles_zeff)) THEN
+         CPWARN(missing_zeff_warning)
+      END IF
 
       ! Shortcut
       lbounds = grid%pw_grid%bounds(1, :)
--- a/src/pw/realspace_grid_cube.F
+++ b/src/pw/realspace_grid_cube.F
@@ -34,6 +34,9 @@ MODULE realspace_grid_cube
    INTEGER, PARAMETER, PRIVATE          :: cube_line_len = cube_entry_len*cube_num_entries_line
    CHARACTER(len=*), PARAMETER          :: cube_value_format = '(1X,ES12.4E3)', &
                                            cube_values_format = '(6(1X,ES12.4E3))'
+   CHARACTER(len=*), PARAMETER, PRIVATE :: missing_zeff_warning = &
+                                           "Effective nuclear charges were not supplied; "// &
+                                           "zeros will be written to the Cube atom records."
    LOGICAL, PARAMETER, PRIVATE          :: debug_this_module = .FALSE.
 
 CONTAINS
@@ -118,6 +121,8 @@ CONTAINS
       IF (PRESENT(mpi_io)) parallel_write = mpi_io
       IF (PRESENT(max_file_size_mb)) my_max_file_size_mb = max_file_size_mb
       CPASSERT(my_max_file_size_mb >= 0)
+      CPWARN_IF(.NOT. parallel_write .AND. PRESENT(particles_z) .AND. .NOT. PRESENT(particles_zeff), &
+                missing_zeff_warning)
 
       my_stride = 1
       IF (PRESENT(stride)) THEN
@@ -701,6 +706,7 @@ CONTAINS
       !get rs grids and parallel envnment
       gid = grid%pw_grid%para%group
       my_rank = grid%pw_grid%para%group%mepos
+      CPWARN_IF(PRESENT(particles_z) .AND. .NOT. PRESENT(particles_zeff), missing_zeff_warning)
 
       ! Shortcut
       lbounds = grid%pw_grid%bounds(1, :)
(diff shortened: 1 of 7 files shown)


he-zilong:
I like this follow-up PR; thank you for the efforts!

Meanwhile, could we confirm that cube files from a TB (`src/qs_scf_post_tb.F`) or SE (`src/qs_scf_post_se.F`) calculation have the correct method-specific effective nuclear charges? If no, I would like to see some more subsequent works.

DCM-Uni-Paderborn:
Thanks for raising this. The supported xTB cube writers already obtain the method-specific effective nuclear charges via `get_xtb_atom_param(..., zeff=...)` and pass them to all relevant cube-writing calls. Cube output is currently disabled for DFTB, while the corresponding semi-empirical cube outputs, including MO cubes, are not implemented. Therefore, I do not see an additional functional fix needed in this GPW-focused PR.

DCM-Uni-Paderborn:
To clarify what follow-up support would require:

For semi-empirical methods, CP2K already constructs an ORB GTO approximation from the method-specific STO parameters and has the AO density matrix and MO coefficients available. A follow-up would need to initialize the plane-wave environment and task list for post-processing, refactor or expose the existing TB cube-writing routines for reuse, obtain the per-atom effective nuclear charges via `get_se_param(..., zeff=...)`, and add regression coverage for the supported cube types. This can be done in a potential follow-up PR ...

DFTB, however, is more involved. Its current setup does not provide an ORB basis suitable for evaluating densities or molecular orbitals on a real-space grid. The Slater–Koster parameterization provides tabulated matrix elements rather than the required real-space orbital representation, so merely enabling the existing cube path would not be sufficient. DFTB cube support would first require an agreed source and definition for the radial orbitals, the corresponding grid-evaluation machinery, and validation of the resulting density and MO cube semantics.

he-zilong:
> Thanks for raising this. The supported xTB cube writers already obtain the method-specific effective nuclear charges via `get_xtb_atom_param(..., zeff=...)` and pass them to all relevant cube-writing calls. Cube output is currently disabled for DFTB, while the corresponding semi-empirical cube outputs, including MO cubes, are not implemented. Therefore, I do not see an additional functional fix needed in this GPW-focused PR.

Understood, thank you! By "some more subsequent works" I meant what ensure every existing cube printouts (I was thinking more of GPW and TB) has the effective nuclear charge(s), and not really that new cube outputs should be invented for DFTB or SE methods if theoretical formulation not alreay available.

(Edit: the rest of this comment converted to a review)

he-zilong (commented):
At the moment, the `particles_zeff` argument is `OPTIONAL` for both `SUBROUTINE pw_to_cube` and `SUBROUTINE pw_to_cube_parallel` in `src/pw/realspace_grid_cube.F`, and a value of 0 is _silently_ used as the effective charge(s) when it is not present.

https://github.com/cp2k/cp2k/blob/406ea2db8b3eba6c4d4d2ff961bb5b7282ad99cf/src/pw/realspace_grid_cube.F#L179-L189
https://github.com/cp2k/cp2k/blob/406ea2db8b3eba6c4d4d2ff961bb5b7282ad99cf/src/pw/realspace_grid_cube.F#L782-L796

I think the check for the presence of `particles_zeff` needs to be much earlier in the two subroutines, and if not present an explicit warning should be made. It is fine to preserve the fallback to 0, but then developers creating new cube outputs need to be notified in the tests when they forget to pass the effective nuclear charge.

Apart from this little change, this PR looks good to me for now.

DCM-Uni-Paderborn:
Thanks, this has now been addressed in bbf632ba32 (with the line-length follow-up in 4f027b4061). Both Cube-writing paths warn explicitly when atom records are requested without `particles_zeff`, while retaining the existing zero-valued fallback. The serial wrapper suppresses its warning when it delegates to the parallel writer, so each affected output emits the warning only once.

he-zilong (approved):

he-zilong:
Is this PR "blocked" from merging only due to a lack of code owner review(s)?

fstein93 (approved):

marci73 (approved):

PR #5724, opened 2026-08-12
Use the FFTW3 Fortran execution interface in fftw31dm

DCM-Uni-Paderborn:
## Summary

- replace the legacy `dfftw_execute_dft` calls in `fftw31dm` with FFTW3's explicit Fortran 2003 interface
- pass the first array element at each FFT offset instead of one-element array sections
- declare the input array `INTENT(INOUT)`, as required by `fftw_execute_dft`

## Motivation

The CSCS Daint H100 psmp regtests consistently segfault in the OpenMP part of `fftw31dm`. The legacy call has no explicit Fortran interface and receives one-element array sections. Using the `fftw3.f03` interface gives the compiler the correct `C_PTR` and assumed-size array contract and avoids array descriptor or temporary handling at the call site.

This keeps the existing FFT plans, work sharing, results, and test tolerances unchanged.

## Testing

- `./make_pretty.sh src/pw/fft/fftw3_lib.F`
- Spark, 2 MPI ranks x 2 OpenMP threads: all five `tests/LIBTEST/misc/test_pw*.inp` inputs passed
- Spark, 2 MPI ranks x 2 OpenMP threads: `w512_pint_qtb_fp0.inp` and `w512_pint_qtb_fp1.inp` passed
- CSCS Daint H100: all `QS/regtest-rtbse-linearized` inputs and matchers passed in the combined stack run

## Follow-up

The four remaining legacy `dfftw_plan_many_dft` planning calls in this file should be converted separately with dedicated coverage, as suggested in review.

PR #5725, opened 2026-08-12
Run GPU-memory-intensive P100 tests exclusively

DCM-Uni-Paderborn:
## Summary

- mark the LibGint, native SKALA, and PAO regtest directories as GPU-memory-intensive batches
- add an optional regtest-runner mode that executes marked batches without any other CP2K process running
- prevent an exclusive batch from starving while newer ordinary batches acquire released worker slots
- run the single-GPU P100 `psmp` and `ssmp` jobs with six OpenMP threads and exclusive GPU-memory batches
- reduce the grid cutoff of six native-SKALA k-point stress smoke tests from 80 to 70 Ry so their differentiable force/stress graph fits in 7.42 GiB

## Motivation

The GCP runner provides 24 CPU cores but only 7.42 GiB of usable device memory. With the former defaults, `psmp` launched six test-directory workers with two MPI ranks each, so up to twelve CP2K CUDA processes shared that GPU.

Serializing only the marked batches was insufficient because ordinary CUDA-enabled batches could still overlap with them. The runner therefore lets an exclusive batch reserve all worker slots. A shared admission lock prevents newer ordinary batches from taking released slots while an exclusive batch is draining the active workers.

The resulting live `ssmp` run showed that concurrency was no longer the cause of the remaining SKALA failures: each of the six k-point force/stress tests alone occupied about 7.31-7.32 GiB and then failed another 308 MiB allocation. These tests require the full differentiable graph, so atom chunking is intentionally disabled for force/virial calculations. Reducing only their smoke-test cutoff retains CUDA, k points, symmetry/inversion, forces, analytical stress, tolerances, and matcher coverage while reducing the graph size.

The scheduling marker has no effect unless a runner explicitly supplies `--exclusive-gpu-memory-batches`; all other regtest configurations retain their existing scheduling.

## Testing

- `python3 -m py_compile tests/do_regtest.py`
- focused asyncio scheduling test: an exclusive batch starts alone and before a later ordinary batch after two active workers finish
- `./make_pretty.sh`
- pre-push checks for all changed files
- first live P100 `ssmp` diagnostic: 5709/5751 correct, zero wrong results; 34 FFTW/RTBSE/PIMD failures covered by #5724, two PAO device failures covered by #5726, and six isolated SKALA out-of-memory failures addressed here

The next P100 run will supply the exact reference values for the six cutoff-adjusted tests. Those values will be updated from the real target run rather than inferred on a different device.

### DIFF
--- a/tests/QS/regtest-gauxc-cuda/H2_NATIVE_SKALA_GAPW_AE_KP_INV_STRESS.inp
+++ b/tests/QS/regtest-gauxc-cuda/H2_NATIVE_SKALA_GAPW_AE_KP_INV_STRESS.inp
@@ -20,7 +20,7 @@
       SYMMETRY ON
     &END KPOINTS
     &MGRID
-      CUTOFF 80
+      CUTOFF 70
       REL_CUTOFF 20
     &END MGRID
     &POISSON
--- a/tests/QS/regtest-gauxc-cuda/H2_NATIVE_SKALA_GAPW_AE_KP_INV_STRESS.inp
+++ b/tests/QS/regtest-gauxc-cuda/H2_NATIVE_SKALA_GAPW_AE_KP_INV_STRESS.inp
@@ -20,7 +20,7 @@
       SYMMETRY ON
     &END KPOINTS
     &MGRID
-      CUTOFF 70
+      CUTOFF 80
       REL_CUTOFF 20
     &END MGRID
     &POISSON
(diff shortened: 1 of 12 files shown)


oschuett:
It took me a while to realize that this did not get merged into our `master` branch. Since we already have a lot of traffic on this repo, should we maybe move feature branches elsewhere?

DCM-Uni-Paderborn:
Thanks, and sorry for the confusion. #5725 was a stacked PR targeting the feature branch used by #5726 rather than `master`. I agree that GitHub's “merged” status is misleading in this setup. For future stacked work, I will keep intermediate feature branches in our fork and only open the final PR against `cp2k/master`.

Following the review of #5726, I am also removing the generic `do_regtest.py` scheduling infrastructure introduced by #5725. Its intended P4/P100 memory isolation will instead be implemented locally in the two Docker testers using the existing regtest options. Thus, #5725 will not be merged into `master` verbatim, but its required operational behavior will be preserved.

oschuett:
>  For future stacked work, I will keep intermediate feature branches in our fork and only open the final PR against cp2k/master.

Alright, I've added a Github rule set to prevent the creation of new branches. The [existing branches](https://github.com/cp2k/cp2k/branches/all) are still there.

PR #5726, opened 2026-08-12
Make PAO TorchScript models CUDA-device safe

DCM-Uni-Paderborn:
## Summary

- register PAO model constants as buffers and create temporary tensors on the input device with the input dtype
- migrate tensor attributes and device-less tensor factories when loading existing serialized PAO v2 models
- initialize PyTorch CUDA linear algebra once before PAO enters threaded prediction

## Rationale

Newly scripted PAO models should encode device handling directly in the model source. `PaoReadout.D_yzx_to_xyz` and the precomputed Wigner blocks are therefore registered buffers, and tensor factories in `SymmetricMatrix.forward()` inherit the input device and dtype.

The H/O models currently shipped with CP2K, as well as user-created PAO v2 model files, predate that source fix. Re-scripting them is not generally possible without their original model/training setup. The narrow loader migration keeps these existing serialized models usable by moving ordinary tensor attributes and replacing `device=None` tensor factories with the selected execution device. CUDA linear algebra is initialized once before OpenMP threads can enter its lazy initialization concurrently.

Although #5725 remains visible in the feature-branch history because it was merged into this branch, its broad P4 scheduler and regtest changes are superseded by the final focusing commit. They are not present in the net diff of this PR. The final diff is limited to `src/pao_model.F`, `src/torch_c_api.cpp`, and `tools/pao-ml/pao/model.py`.

## Testing

- complete `Dockerfile.test_misc` container test: PAO training and retraining, validation of the newly generated model, validation of the shipped H/O models, and `mypy --strict ./tools/pao-ml/`
- CUDA PAO energy/force tests and the host-load/device-move path on Terok and Spark
- `./make_pretty.sh src/pao_model.F src/torch_c_api.cpp tools/pao-ml/pao/model.py`
- CP2K pre-push hook and `git diff --check`


mkrack:
@DCM-Uni-Paderborn Could you trigger the checks, e.g. by rebase or an empty commit?

oschuett:
Thanks a lot for getting PAO-ML ready for GPUs. Admittedly, we're still focused on training the model in the first place.

Since we own the [model](https://github.com/cp2k/cp2k/blob/master/tools/pao-ml/pao/model.py), we can also fix any problems directly at the source.

Regarding the changes to `do_regtest.py`, I'm hesitant. They add a lot of complexity and moving parts. I think, we can achieve essentially the same with the existing `--maxtasks` and `--restrictdir` flags.

And in a few weeks, when I'm done with the new CI, we should be able to finally ditch the P4 and test on larger GPUs.

DCM-Uni-Paderborn:
Thanks, I agree. I have removed the `do_regtest.py`, `TEST_DIRS`, P100-runner, and unrelated regtest/Spack changes from the net diff. New PAO models now handle device placement directly in `tools/pao-ml/pao/model.py` through registered buffers and input-derived tensor factories. I retained only the narrow loader migration needed for existing serialized PAO v2 models, plus the one-time CUDA linear-algebra initialization before threaded prediction. #5725 remains traceable in the feature-branch history, but its broad P4-specific changes are superseded in this PR.

oschuett (approved):

PR #5739, opened 2026-08-13
FIST: add symmetrix as a torch-free BACKEND for the &MACE potential

DhruvSkyy:
## What this adds

CP2K can already run MACE machine-learning potentials, but only through TorchScript, which means every build and every run has to carry the full LibTorch (the PyTorch C++ library). This pull request adds a second way to run the same MACE models, called [symmetrix](https://github.com/wcwitt/symmetrix). symmetrix is a small C++ library that evaluates a MACE model directly, without PyTorch, so it links a lightweight static library instead of the whole LibTorch stack and builds with the same compiler as the rest of CP2K.

Nothing changes for existing users. The `&MACE` section (added in cp2k#5580) gains one keyword, `BACKEND`, which chooses the engine. `BACKEND TORCH` is the default and behaves exactly as before, while `BACKEND SYMMETRIX` runs the new torch-free path. The rest of the input is identical, so switching between them is a one-line change.

## How it works

When CP2K evaluates the potential in parallel, symmetrix splits the atoms across the MPI ranks: each rank takes a contiguous block of the simulation box and computes the energy and forces for just those atoms, keeping the surrounding atoms as fixed context so the result stays exact. Each rank produces its own partial energy, forces and stress, and CP2K's normal cross-rank sum adds them into the totals, exactly as it does for the built-in potentials. The neighbour-list bookkeeping that both the TorchScript path and symmetrix need is the same, so it is pulled out into one shared helper module (`manybody_mlp_neighbor.F`) that both now use.

## Why it is worth having

The point of symmetrix is speed and the size of system you can run. The tables below compare it with the TorchScript backend (`BACKEND TORCH`, from cp2k#5580) on one A100 GPU. Times are milliseconds per MD step, and the forces agree to within about 1e-6, so this is the same physics run two ways. The speed-up column is torch divided by symmetrix.

By model size (768 atoms):

| Model | torch | symmetrix | speed-up |
|---|---|---|---|
| MACE-OMAT-0 small | 79.5  | 44.7 | 1.8x |
| MACE-OFF23 small  | 41.4  | 18.2 | 2.3x |
| MACE-OFF23 medium | 131.8 | 36.5 | 3.6x |
| MACE-OFF23 large  | 460.3 | 98.0 | 4.7x |

By system size (MACE-OMAT-0 small):

| Atoms  | torch | symmetrix | speed-up |
|---|---|---|---|
| 192    | 26.2  | 15.9  | 1.6x |
| 768    | 79.5  | 44.7  | 1.8x |
| 3,072  | 318.4 | 168.1 | 1.9x |
| 6,144  | 682.7 | 324.2 | 2.1x |
| 12,288 | out of memory | 622.0 | — |

The pattern is simple: the bigger the MACE model, the more work PyTorch does each step, so the more symmetrix wins, from about 1.6x on the smallest case to 4.7x on the largest model. The last row is the important one. At 12,288 atoms only symmetrix still runs; the TorchScript path runs out of memory on the 80 GB card, because PyTorch has to hold a large computational graph while symmetrix does not. The one price symmetrix pays is a slightly slower start, about 0.6 s to read its 50 MB model file against 0.01 s for the 7.4 MB TorchScript file, which is a one-time cost and negligible for a real run.

## Building and testing

symmetrix is off by default and opt-in with `-DCP2K_USE_SYMMETRIX=ON`. It can be installed either through the CP2K toolchain (`--with-symmetrix`) or through spack (`cp2k_dev/symmetrix`). The PR adds the CMake `FindSymmetrix` module, documentation, and six regression tests built on a small committed model, covering the energy, the stress, an analytic-versus-numerical force check, a triclinic cell, a mixed MACE-plus-classical system, and a short molecular-dynamics run. They pass identically at 1 and 3 MPI ranks:

```
Number of FAILED  tests 0
Number of WRONG   tests 0
Number of CORRECT tests 12
Status: OK
```

The two-water reference energy is `-5.585977103627951` Ha.

cc @cschran


CMakeLists.txt
```diff
@@ -129,6 +129,23 @@ option(CP2K_USE_LIBFCI "Enable LibFCI active-space solver support"
        ${CP2K_USE_EVERYTHING})
 option(CP2K_USE_LIBINT2 "Enable Libint2 support" ${CP2K_USE_EVERYTHING})
 option(CP2K_USE_LIBTORCH "Enable LibTorch support" ${CP2K_USE_EVERYTHING})
+option(CP2K_USE_SYMMETRIX "Enable symmetrix MACE interatomic potentials" OFF)
```
oschuett, line 132:
```suggestion
option(CP2K_USE_SYMMETRIX "Enable symmetrix MACE interatomic potentials" ${CP2K_USE_EVERYTHING})
```

tools/toolchain/scripts/stage6/install_symmetrix.sh
```diff
@@ -0,0 +1,97 @@
...
+    pkg_install_dir="${INSTALLDIR}/symmetrix-${symmetrix_commit}"
+    install_lock_file="${pkg_install_dir}/install_successful"
+    symmetrix_root="${pkg_install_dir}/symmetrix"
+    if verify_checksums "${install_lock_file}"; then
+      echo "symmetrix-${symmetrix_commit} is already installed, skipping it."
+    else
+      echo "Cloning ${symmetrix_repo} @ ${symmetrix_commit} into ${symmetrix_root}"
+      [ -d "${pkg_install_dir}" ] && rm -rf "${pkg_install_dir}"
+      mkdir -p "${pkg_install_dir}"
+      git clone --recursive "${symmetrix_repo}" "${symmetrix_root}" \
```
oschuett, line 43: Unfortunately, installing directly from Github has proven to be quite unreliable. So, maybe you could bundle this as a tar-ball that we can host on cp2k.org?

oschuett (commented):
Overall looks good.
Is there anything we can help you with to move this out of the draft status?

DhruvSkyy:
> Overall looks good. Is there anything we can help you with to move this out of the draft status?

I'm just testing a few things to see if it can be more efficient but it should be ready soon.

Issue #5741, opened 2026-08-14
Consider using Libint backend for existing Gaussian integral implementations


RTChou67:
CP2K currently contains several recurrence implementations for Gaussian integrals, including: 
  - coulomb2 in src/aobasis/ai_coulomb.F;
  - coulomb2_new in src/aobasis/ai_coulomb.F;
  - coulomb3 in src/aobasis/ai_coulomb.F;
  - and many other APIs. 
 
At the same time, CP2K already provides libint wrappers, which raised a question: should some of the existing APIs migrated to an external integral backend, such as libint or libcint mentioned in [#1232](https://github.com/cp2k/cp2k/issues/1232)?

Growl1234:
I think, instead of tackling what could be a fairly involved change right now, we should look ahead to LibintX.

RTChou67:
WOWWWW. That's indeed fast. I fell in love with it

fstein93:
Because of functionality, the other implementations should not be removed. Otherwise, more features will require an external library. The only thing I could imagine is to use the implementations mentioned by you as an alternative backend and the libraries as faster variants.

Growl1234:
That makes sense to me. In that case, perhaps the useful direction here would be to expose the internal implementations and external libraries through a common backend interface, rather than replacing the former. There are also something this can bring:
- a chance to refactor libint wrappers, the problem was pointed out by #1232;
- a cleaner path for LibintX in the future.

PR #5748, opened 2026-08-14
Add native/Libint backend selection for several electronic integral APIs

RTChou67:
This change adds independent backend selection for three existing Gaussian Coulomb integral APIs: `coulomb2`, `coulomb2_new`, and `coulomb3`. The existing native implementations remain available. The Libint paths use the existing `eri_2center` and `eri_3center` wrappers.

Backend selection is controlled under `DFT%QS%BACKEND`:

```text
&BACKEND
  COULOMB2 NATIVE|LIBINT
  COULOMB2_NEW NATIVE|LIBINT
  COULOMB3 NATIVE|LIBINT
&END BACKEND
```
The benchmark uses RI-SOS-MP2 energy calculations for `(H2O)4` and `(H2O)8` clusters, tested 4 times. Each case uses eight MPI ranks on a i9-13900HX laptop. The values below are median wall times in seconds:

```text
CASE   BACKEND  MPI_RANKS  MEDIAN_SEC
H2O4   NATIVE   8          2.435
H2O4   LIBINT   8          2.020
H2O8   NATIVE   8          9.905
H2O8   LIBINT   8          6.820
```

## Future Work

- Add backend selection for other electronic integral APIs.
- Add regtests covering the libint backend paths.
- Review native APIs for duplicated recurrence code and combine duplicate recurrence codes.


RTChou67:
This PR is a direct follow-up to #5741. This also provides a first step toward the broader refactoring discussed in #5741, namely making better use of the existing Libint wrappers and reducing duplicated Gaussian integral recurrence implementations. Further backend coverage and consolidation of duplicated recurrence code can be addressed in follow-up work.
I would appreciate any feedback or suggestions on the current approach, particularly on regtest design, or things related to libint, which i'm not very familiar with.

fstein93:
I'm not very familiar with the Coulomb2 vs Coulomb2_new backends. Is it necessary to have them separate? This may be confusing for users. Can we consider both to be backends for the calculation of two-center-electron repulsion integrals?

RTChou67:
I also don't know why there are two separate APIs instead of a single upgraded version, since I didn't write them. These APIs do more than just calculate electronic integrals, and I only added an alternative implementation for the recurrence part. Their recurrence codes seem to be the same, while the other parts are different. I think merging them could be a very good idea after a thorough review. They are not the only pair sharing the same recurrence core. In my other PRs I merged some, but there are too many.

Growl1234:
Hmm, you simply removed `coulomb2_new` API? 🤔

RTChou67:
> Hmm, you simply removed `coulomb2_new` API? 🤔

Not really. They share the same two-center Coulomb recurrence code, so can be merged. coulomb2_new does not accept screening, so I add a little switch to the coulomb2 API. Besides, I removed the unused two-center maxder/vac_plus arguments, as it is never CALLed.

RTChou67:
> Hmm, you simply removed `coulomb2_new` API? 🤔

Not really. They share the same two-center Coulomb recurrence code, so they can be merged. coulomb2_new does not accept screening, so I added a little switch to the coulomb2 API. Besides, I removed the unused two-center maxder/vac_plus arguments, as it is never CALLed, i think that's ok.

Growl1234:
Okay. Looking forward to the follow-up work!

RTChou67:
The duplicate handwritten recurrence differs from `os_rr_ovlp` in `src/ai_os_rr.F`, which may lead to numerical discrepancies. Might have to change the regtest results.

RTChou67:
Interesting. I reran the test locally using GNU 13.3.0 on Ubuntu 24.04.4. Both the psmp and pdbg builds produced the expected result. This suggests that the reported `WRONG RESULT` may be environment-dependent. Could someone let me know more about the CI test environment?
[Au13ico_mtd.inp.out.psmp.txt](https://github.com/user-attachments/files/31119509/Au13ico_mtd.inp.out.psmp.txt)
[Au13ico_mtd.inp.out.pdbg.txt](https://github.com/user-attachments/files/31119514/Au13ico_mtd.inp.out.pdbg.txt)

fstein93:
Some information (CPU type, instruction set) are provided in the CP2K output file.

Growl1234:
You can simply see the Dockerfiles for environment.

It runs on Ubuntu 26.04 container (gcc-15.2.0), and all dependencies are installed via toolchain.

RTChou67:
Thanks for the replies.

RTChou67:
[Au13ico_mtd.inp.out.GCC15.txt](https://github.com/user-attachments/files/31134185/Au13ico_mtd.inp.out.GCC15.txt)
Strange. I compiled on Ubuntu 26.04 with GNU 15.2.0; other dependencies compiled. Still, it is completely correct. I don't know why the CI test fail.

mkrack:
> [Au13ico_mtd.inp.out.GCC15.txt](https://github.com/user-attachments/files/31134185/Au13ico_mtd.inp.out.GCC15.txt) Strange. I compiled on Ubuntu 26.04 with GNU 15.2.0; other dependencies compiled. Still, it is completely correct. I don't know why the CI test fail.

For me that looks like normal, acceptable numerical noise. I suggest to increase the tolerance for that test to `5.0E-9`.

RTChou67:
> > [Au13ico_mtd.inp.out.GCC15.txt](https://github.com/user-attachments/files/31134185/Au13ico_mtd.inp.out.GCC15.txt) Strange. I compiled on Ubuntu 26.04 with GNU 15.2.0; other dependencies compiled. Still, it is completely correct. I don't know why the CI test fail.
> 
> For me that looks like normal, acceptable numerical noise. I suggest to increase the tolerance for that test to `5.0E-9`.

Thanks for the approval! I'll loosen the tolerance and update the regtest accordingly.

fstein93 (approved):

JWilhelm (commented):
The implementation looks good to me overall. I only have a question about the naming and placement of the new input section.

Is there a particular reason for calling it `BACKEND`? This name seems rather generic, whereas the section currently selects the implementation of specific two- and three-center Coulomb integral routines. Since the intended scope appears to be a very specific Coulomb-integral backend, would a more specific name such as  `COULOMB_INTEGRALS` be clearer?

Relatedly, is there a reason to place this section under `DFT%QS` rather than under `GLOBAL`? `GLOBAL` already contains several program-wide library-selection options, such as `PREFERRED_DIAG_LIBRARY`, `PREFERRED_FFT_LIBRARY`, and `PREFERRED_CHOLESKY_LIBRARY`. Since this section likewise selects between native CP2K and external-library implementations, I wonder whether it would fit more naturally there. I may be missing a reason why this choice needs to be specific to the QS environment; if so, could you please clarify?

RTChou67:
Thank you for the detailed suggestions. Currently, coulomb2 and coulomb3 represent two distinct electronic integral APIs, so I would prefer to keep them as separate keywords. This may also be useful for debugging, especially as more electronic integral APIs gain alternative backends from external libraries.

I agree that GLOBAL is a more natural location for the backend selection. I will move it from DFT%QS to GLOBAL.

JWilhelm (approved):

fstein93:
If the backend is a global property, I would be in favor of setting it at an earlier stage (similar to other libraries such as with FFTs, (P)DGEMM or diagonalization) and wrap all respective operations such that there is no need to pass around the backend variable between all routines.

RTChou67:
Thanks, I agree with this approach. The COULOMB2 and COULOMB3 library choices are now read during global initialization from GLOBAL%PREFERRED_INTEGRAL_LIBRARY. The selected libraries are stored in the global environment and propagated to an active integral-library state used by the integral operations.

src/input_cp2k_global.F
```diff
@@ -275,6 +277,36 @@ SUBROUTINE create_global_section(section)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
+      NULLIFY (sub_section)
+      CALL section_create(sub_section, __LOCATION__, name="preferred_integral_library", &
+                          description="Selects the implementation of individual Gaussian integral operations.", &
+                          n_keywords=2, n_subsections=0, repeats=.FALSE.)
+
+      CALL keyword_create(keyword, __LOCATION__, name="COULOMB2", &
+                          description="Library for the coulomb2 API.", &
```
fstein93: `coulomb2` is internal and not a good wording for users. I would rather write something like `two-center Coulomb integrals` and `Coulomb_2c`

src/input_cp2k_global.F
```diff
@@ -275,6 +277,36 @@ SUBROUTINE create_global_section(section)
...
+                          usage="COULOMB2 NATIVE", &
+                          enum_c_vals=s2a("NATIVE", "LIBINT"), &
+                          enum_i_vals=[library_native, library_libint], &
+                          enum_desc=s2a("Use CP2K's native implementation.", &
+                                        "Use the Libint implementation."), &
+                          default_i_val=library_native)
+      CALL section_add_keyword(sub_section, keyword)
+      CALL keyword_release(keyword)
+
+      CALL keyword_create(keyword, __LOCATION__, name="COULOMB3", &
```
fstein93: Similar here

src/mp2_eri.F
```diff
@@ -160,6 +172,13 @@ SUBROUTINE mp2_eri_2c_integrate(param, potential_parameter, para_env, qs_env, ba
       my_eri_method = do_eri_mme
       IF (PRESENT(eri_method)) my_eri_method = eri_method
 
+      use_coulomb2_libint = my_eri_method == do_eri_os .AND. &
```
fstein93: Because we now have a unified backend for two- and three-center integrals, I consider the differences between the backends as internal and would move the initialization there (similar to the FFT-backend where the native backend does not require an initialization but FFTW).

fstein93: I would even move this initialization step upstream to `f77_interface` similarly to the other backends.

src/mp2_eri.F
```diff
@@ -182,6 +201,13 @@ SUBROUTINE mp2_eri_2c_integrate(param, potential_parameter, para_env, qs_env, ba
       CALL get_qs_env(qs_env, atomic_kind_set=atomic_kind_set, qs_kind_set=qs_kind_set, &
                       cell=cell, particle_set=particle_set, natom=natom, nkind=nkind)
 
+      IF (use_coulomb2_libint) THEN
+         CALL get_qs_kind_set(qs_kind_set=qs_kind_set, maxlgto=max_am_a, basis_type=basis_type_a)
+         CALL get_qs_kind_set(qs_kind_set=qs_kind_set, maxlgto=max_am_b, basis_type=basis_type_b)
+         CALL cp_libint_init_2eri(lib, MAX(max_am_a, max_am_b))
+         CALL cp_libint_set_contrdepth(lib, 1)
+      END IF
```
fstein93, line 195: This should go now to the interface

src/mp2_eri.F
```diff
@@ -306,6 +336,8 @@ SUBROUTINE mp2_eri_2c_integrate(param, potential_parameter, para_env, qs_env, ba
          END DO
       END DO
 
+      IF (use_coulomb2_libint) CALL cp_libint_cleanup_2eri(lib)
```
fstein93: This branch should go to the interface and additionally, if you move the initialization to `f77_interface`, then this one should also go there.

src/mp2_eri.F
```diff
@@ -534,16 +571,23 @@ SUBROUTINE integrate_set_2c(param, potential_parameter, la_min, la_max, lb_min,
...
+            ALLOCATE (f_work(0:la_max + lb_max + 2))
+            ALLOCATE (v_work(ncoa, ncob, la_max + lb_max + 1))
 
-         CALL coulomb2_new(la_max, npgfa, zeta, la_min, lb_max, npgfb, zetb, lb_min, &
-                           rb - ra, rab2, hab_uncontr, v_work, f_work)
+            CALL coulomb2(la_max, npgfa, zeta, rpgfa, la_min, lb_max, npgfb, zetb, rpgfb, lb_min, &
+                          rb - ra, rab2, hab_uncontr, v_work, f_work, screening=.FALSE.)
 
-         DEALLOCATE (v_work, f_work)
+            DEALLOCATE (v_work, f_work)
```
fstein93: Move this to the interface

src/mp2_eri.F
```diff
@@ -675,6 +721,13 @@ SUBROUTINE mp2_eri_3c_integrate(param, potential_parameter, para_env, qs_env, &
       my_eri_method = do_eri_mme
       IF (PRESENT(eri_method)) my_eri_method = eri_method
 
+      use_coulomb3_libint = my_eri_method == do_eri_os .AND. &
+                            active_integral_library%coulomb3_library == library_libint
+      IF (my_eri_method == do_eri_os .AND. active_integral_library%coulomb3_library /= library_native .AND. &
+          .NOT. use_coulomb3_libint) THEN
+         CPABORT("Requested three-center Coulomb library is not implemented")
+      END IF
```
fstein93: This goes to the interface.

src/mp2_eri.F
```diff
@@ -723,6 +776,14 @@ SUBROUTINE mp2_eri_3c_integrate(param, potential_parameter, para_env, qs_env, &
       CALL basis_set_list_setup(basis_set_list_a, basis_type_a, qs_kind_set)
       CALL basis_set_list_setup(basis_set_list_b, basis_type_b, qs_kind_set)
 
+      IF (use_coulomb3_libint) THEN
+         CALL get_qs_kind_set(qs_kind_set=qs_kind_set, maxlgto=max_am_a, basis_type=basis_type_a)
+         CALL get_qs_kind_set(qs_kind_set=qs_kind_set, maxlgto=max_am_b, basis_type=basis_type_b)
+         CALL get_qs_kind_set(qs_kind_set=qs_kind_set, maxlgto=max_am_c, basis_type=basis_type_c)
+         CALL cp_libint_init_3eri(lib, MAX(max_am_a, max_am_b, max_am_c))
+         CALL cp_libint_set_contrdepth(lib, 1)
+      END IF
```
fstein93, line 748: This goes to the interface.

src/mp2_eri.F
```diff
@@ -998,6 +1063,8 @@ SUBROUTINE mp2_eri_3c_integrate(param, potential_parameter, para_env, qs_env, &
       DEALLOCATE (basis_set_list_a, basis_set_list_b)
       CALL neighbor_list_iterator_release(nl_iterator)
 
+      IF (use_coulomb3_libint) CALL cp_libint_cleanup_3eri(lib)
+
```
fstein93, line 1030: This goes to the interface.

src/mp2_eri.F
```diff
@@ -1302,40 +1374,49 @@ SUBROUTINE integrate_set_3c(param, potential_parameter, &
...
+            nc_end = ncoset(lc_max)
+
+            CALL coulomb3(la_max, npgfa, zeta(:), rpgfa_work(:), la_min, &
+                          lb_max, npgfb, zetb(:), rpgfb_work(:), lb_min, &
+                          lc_max, zetc(kpgf), rpgfc_work(kpgf), lc_min, &
+                          gccc, rb - ra, rab2, rc - ra, rac2, rbc2, &
+                          vabc, habc_uncontr(:, :, nc_start:nc_end), v_work, f_work)
+
+            DEALLOCATE (v_work, f_work, rpgfa_work, rpgfb_work, rpgfc_work, gccc, vabc)
+         END IF
```
fstein93: This goes to the interface.

src/mp2_gpw.F
```diff
@@ -292,6 +296,14 @@ SUBROUTINE mp2_gpw_main(qs_env, mp2_env, Emp2, Emp2_Cou, Emp2_EX, Emp2_S, Emp2_T
...
       eri_method = mp2_env%eri_method
 
+      IF ((active_integral_library%coulomb2_library == library_libint .OR. &
+           active_integral_library%coulomb3_library == library_libint) .AND. &
+          mp2_env%eri_method == do_eri_os .AND. &
+          .NOT. libint_initialized) THEN
+         CALL cp_libint_static_init()
+         libint_initialized = .TRUE.
+      END IF
+
```
fstein93: This goes to the interface.

src/mp2_gpw.F
```diff
@@ -623,6 +634,8 @@ SUBROUTINE mp2_gpw_main(qs_env, mp2_env, Emp2, Emp2_Cou, Emp2_EX, Emp2_S, Emp2_T
 
       END IF
 
+      IF (libint_initialized) CALL cp_libint_static_cleanup()
+
```
fstein93: This goes to the interface.

src/qs_environment.F
```diff
@@ -1054,6 +1059,9 @@ SUBROUTINE qs_init_subsys(qs_env, para_env, subsys, cell, cell_ref, use_ref_cell
       TYPE(xtb_control_type), POINTER                    :: xtb_control
 
       CALL timeset(routineN, handle)
+      coulomb2_libint_initialized = &
+         active_integral_library%coulomb2_library == library_libint
+      IF (coulomb2_libint_initialized) CALL cp_libint_static_init()
```
fstein93: This goes to the interface.

src/qs_environment.F
```diff
@@ -1541,6 +1549,7 @@ SUBROUTINE qs_init_subsys(qs_env, para_env, subsys, cell, cell_ref, use_ref_cell
 
       !  Initialise the qs_kind_set
       CALL init_qs_kind_set(qs_kind_set)
+      IF (coulomb2_libint_initialized) CALL cp_libint_static_cleanup()
```
fstein93: This goes to the interface.

fstein93 (changes requested):
Sorry for spamming so many new comments but with the changes you introduced now, the branching should go to the interface. The caller should not need to worry about the backend in use.
Regarding moving the initalization/cleanup steps of the backends to the `f77_interface`, I do not know whether it makes sense to initialize the libraries in `f77_interface` whenever CP2K is called or only when needed by the caller. The many comments are just a guide for you to not omit any spot.

src/input_cp2k_global.F
RTChou67: Thank you for your detailed review and suggestions. Do you think the keywords themselves should also be changed for clarity, or should we keep `COULOMB2` and `COULOMB3` for simplicity?

Growl1234: I'd prefer `COULOMB_2C` and `COULOMB_3C`.

RTChou67: Sounds good to me. I'll rename that.

RTChou67:
I added coulomb_integral_interface.F to provide a common CP2K-level entry point for the existing two- and three-center Coulomb integral implementations. Consequently, callers no longer need to inspect the selected library or manage libint engines themselves. The process-level Libint initialization and cleanup belong in f77_interface, as you suggested. These calls only manage the global Libint runtime and can be performed once when CP2K starts and terminates.

Growl1234:
Should we also have a look at `src/aobasis/ai_coulmb.F`? There are also `USE libint_2c_3c` and `USE libint_wrapper` there.

fstein93:
> Should we also have a look at `src/aobasis/ai_coulmb.F`? There are also `USE libint_2c_3c` and `USE libint_wrapper` there.

One could also add a check to the convention checker similar to not calling cp_fm_gemm instead of parallel_gemm_api or similar

src/coulomb_integral_interface.F
```diff
@@ -0,0 +1,273 @@
...
+      TYPE(coulomb_integral_context_type), INTENT(INOUT) :: context
+      INTEGER, INTENT(IN), OPTIONAL                      :: max_am_2c, max_am_3c
+
+      IF (PRESENT(max_am_2c)) THEN
+         SELECT CASE (active_integral_library%coulomb2_library)
+         CASE (library_libint)
+            CALL cp_libint_init_2eri(context%lib_2c, max_am_2c)
+            CALL cp_libint_set_contrdepth(context%lib_2c, 1)
+         CASE (library_native)
+            CONTINUE
```
fstein93: I would use a comment `'! Nothing to do` instead. `CONTINUE` is mainly used as a reference for `GOTO`.

src/coulomb_integral_interface.F
```diff
@@ -0,0 +1,273 @@
...
+         context%initialized_2c = .TRUE.
+      END IF
+
+      IF (PRESENT(max_am_3c)) THEN
+         SELECT CASE (active_integral_library%coulomb3_library)
+         CASE (library_libint)
+            CALL cp_libint_init_3eri(context%lib_3c, max_am_3c)
+            CALL cp_libint_set_contrdepth(context%lib_3c, 1)
+         CASE (library_native)
+            CONTINUE
```
fstein93: Same

src/coulomb_integral_interface.F
```diff
@@ -0,0 +1,273 @@
...
+! **************************************************************************************************
+   SUBROUTINE coulomb_integral_cleanup(context)
+      TYPE(coulomb_integral_context_type), INTENT(INOUT) :: context
+
+      IF (context%initialized_2c) THEN
+         SELECT CASE (active_integral_library%coulomb2_library)
+         CASE (library_libint)
+            CALL cp_libint_cleanup_2eri(context%lib_2c)
+         CASE (library_native)
+            CONTINUE
```
fstein93: Same

src/coulomb_integral_interface.F
```diff
@@ -0,0 +1,273 @@
...
+         END SELECT
+         context%initialized_2c = .FALSE.
+      END IF
+
+      IF (context%initialized_3c) THEN
+         SELECT CASE (active_integral_library%coulomb3_library)
+         CASE (library_libint)
+            CALL cp_libint_cleanup_3eri(context%lib_3c)
+         CASE (library_native)
+            CONTINUE
```
fstein93: Same

src/f77_interface.F
```diff
@@ -315,6 +317,9 @@ SUBROUTINE init_cp2k(init_mpi, ierr, mpi_comm)
          CALL pw_gpu_init()
          CALL grid_library_init()
          CALL dbm_library_init()
+#if(__LIBINT)
```
fstein93: Please use `#if defined(__LIBINT)` instead.

src/f77_interface.F
```diff
@@ -350,6 +355,10 @@ SUBROUTINE finalize_cp2k(finalize_mpi, ierr)
          END DO
          DEALLOCATE (f_envs)
 
+#if(__LIBINT)
```
fstein93: Same here

src/environment.F
```diff
@@ -550,6 +551,11 @@ SUBROUTINE read_global_section(root_section, para_env, globenv)
                                 l_val=globenv%direct_generalized_diagonalization)
       CALL section_vals_val_get(global_section, "PREFERRED_CHOLESKY_LIBRARY", i_val=i_cholesky)
       CALL section_vals_val_get(global_section, "PREFERRED_DGEMM_LIBRARY", i_val=i_dgemm)
+      CALL section_vals_val_get(global_section, "PREFERRED_INTEGRAL_LIBRARY%COULOMB_2C", &
+                                i_val=globenv%integral_library%coulomb2_library)
```
fstein93: I would rather have it inside of `integral_library_types` and use a global variable there similar to the linear algebra libraries.

RTChou67:
> > Should we also have a look at `src/aobasis/ai_coulmb.F`? There are also `USE libint_2c_3c` and `USE libint_wrapper` there.
> 
> One could also add a check to the convention checker similar to not calling cp_fm_gemm instead of parallel_gemm_api or similar

that seem another ton of work, as cp_fm_gemm is still a public API.

fstein93 (changes requested):
I also do not quite understand why there are two separate modules `coulomb_integral_interface` and `integral_library_types`

fstein93:
> > > Should we also have a look at `src/aobasis/ai_coulmb.F`? There are also `USE libint_2c_3c` and `USE libint_wrapper` there.
> > 
> > 
> > One could also add a check to the convention checker similar to not calling cp_fm_gemm instead of parallel_gemm_api or similar
> 
> that seem another ton of work, as cp_fm_gemm is still a public API.

Could be for a follow-up PR. My idea is just to prevent developers from using the low-level routines instead of the new common API.

RTChou67:
> I also do not quite understand why there are two separate modules `coulomb_integral_interface` and `integral_library_types`

My thoughts are that `integral_library_types` is a lightweight configuration module that only defines the library identifiers, the global integral_library_type, and the active library selection. So maybe it's better not to depend on the Coulomb implementation or libint. `coulomb_integral_interface` does the dispatch and calls different backends. Merging the two modules might make it a bit too heavy. If in the future, other electronic integrals and other libraries are added, it might become even messier.

RTChou67:
> > > > Should we also have a look at `src/aobasis/ai_coulmb.F`? There are also `USE libint_2c_3c` and `USE libint_wrapper` there.
> > > 
> > > 
> > > One could also add a check to the convention checker similar to not calling cp_fm_gemm instead of parallel_gemm_api or similar
> > 
> > 
> > that seem another ton of work, as cp_fm_gemm is still a public API.
> 
> Could be for a follow-up PR. My idea is just to prevent developers from using the low-level routines instead of the new common API.

Yes, I agree. I don't know if i have time though.

Growl1234:
Sorry for my smushed brain. But I might be able to take care of the convention checker.

RTChou67:
> > One could also add a check to the convention checker similar to not calling cp_fm_gemm instead of parallel_gemm_api or similar
> 
> Isn't there already one? https://www.cp2k.org/conv#c101

I'm not familiar with the precommit tool. As CP2K can maintain automatically generated Doxygen, would it be possible to ban these low-level routines in the precommit script?

oschuett:
> I'm not familiar with the precommit tool. 

The checks for banning routines are implemented here:
https://github.com/cp2k/cp2k/blob/89b16026ca000a26015dd4c5a772b645aa970245/tools/conventions/analyze_gfortran_ast.py#L147-L161

src/coulomb_integral_interface.F
```diff
@@ -0,0 +1,273 @@
...
+             coulomb_integral_init, &
+             compute_coulomb_2c, &
+             compute_coulomb_3c
+
+   TYPE, PUBLIC :: coulomb_integral_context_type
+      TYPE(cp_libint_t) :: lib_2c
+      TYPE(cp_libint_t) :: lib_3c
+      LOGICAL           :: initialized_2c = .FALSE.
+      LOGICAL           :: initialized_3c = .FALSE.
+   END TYPE coulomb_integral_context_type
```
Growl1234, line 46: Could the components of this public type be made private? `lib_2c` and `lib_3c` look like backend-specific implementation details, and hiding them would prevent callers from depending on Libint state and make adding/replacing backends easier.

src/coulomb_integral_interface.F
```diff
@@ -0,0 +1,273 @@
...
+      TYPE(coulomb_integral_context_type), INTENT(INOUT) :: context
+      INTEGER, INTENT(IN)                                :: la_min, la_max, lb_min, lb_max, npgfa, &
+                                                            npgfb
+      REAL(dp), DIMENSION(npgfa), INTENT(IN)             :: zeta
+      REAL(dp), DIMENSION(npgfb), INTENT(IN)             :: zetb
+      REAL(dp), DIMENSION(npgfa), INTENT(IN)             :: rpgfa
+      REAL(dp), DIMENSION(npgfb), INTENT(IN)             :: rpgfb
+      REAL(dp), DIMENSION(3), INTENT(IN)                 :: ra, rb
+      REAL(dp), DIMENSION(:, :), INTENT(INOUT)           :: hab
+      TYPE(libint_potential_type), INTENT(IN)            :: potential_parameter
```
Growl1234: Could this be a backend-neutral potential/operator type? Otherwise the common interface still exposes a Libint-specific type.

src/coulomb_integral_interface.F
```diff
@@ -0,0 +1,273 @@
...
+                                                            lc_max, npgfa, npgfb, npgfc
+      REAL(dp), DIMENSION(npgfa), INTENT(IN)             :: zeta
+      REAL(dp), DIMENSION(npgfb), INTENT(IN)             :: zetb
+      REAL(dp), DIMENSION(npgfc), INTENT(IN)             :: zetc
+      REAL(dp), DIMENSION(npgfa), INTENT(IN)             :: rpgfa
+      REAL(dp), DIMENSION(npgfb), INTENT(IN)             :: rpgfb
+      REAL(dp), DIMENSION(npgfc), INTENT(IN)             :: rpgfc
+      REAL(dp), DIMENSION(3), INTENT(IN)                 :: ra, rb, rc
+      REAL(dp), DIMENSION(:, :, :), INTENT(INOUT)        :: habc
+      TYPE(libint_potential_type), INTENT(IN)            :: potential_parameter
```
Growl1234: Same.

src/coulomb_integral_interface.F
```diff
@@ -0,0 +1,273 @@
...
+         ALLOCATE (rpgfa_work(npgfa))
+         ALLOCATE (rpgfb_work(npgfb))
+         ALLOCATE (rpgfc_work(npgfc))
+         rpgfa_work(:) = 1.0E10_dp
+         rpgfb_work(:) = 1.0E10_dp
+         rpgfc_work(:) = 1.0E10_dp
+         ALLOCATE (gccc(ncoc))
+         gccc(:) = 0.0_dp
+         ALLOCATE (vabc(ncoa, ncob))
+         vabc(:, :) = 0.0_dp
```
Growl1234, line 253: Is the no-screening behavior equivalent for the Libint backend? The native path explicitly disables all three radius checks here, while `eri_3center` still applies the AB screening because `dr_ab` remains zero for the Coulomb operator.

src/coulomb_integral_interface.F
RTChou67: Good advise. I'll modify

src/coulomb_integral_interface.F
RTChou67: Yes, I agree. The public Coulomb interface should not expose libint_potential_type, since this would still make callers depend on Libint-specific data structures. Adding a backend-neutral type module and changing coulomb_integral_interface to use that type would be better. This would be a quite large modification which requiers 
changing a lot of code .

src/coulomb_integral_interface.F
RTChou67: humm. This seems to be introduced before my implementation. Adding a  dr_ab = 1000000.0_dp may address the problem, but might cause performance issues

Growl1234: Right, I agree this predates the PR. My concern is mainly backend equivalence now that the two implementations are exposed as alternatives. Before disabling the AB screening in Libint, perhaps we should first establish whether the native `no screening` behavior is actually required; if AB screening is valid here, aligning the native path instead might be preferable.

src/coulomb_integral_interface.F
Growl1234: It may not have to require changing all existing callers in this PR. For example, it might be possible to move the type itself to a backend-neutral module and temporarily re-export it from `libint_2c_3c` under the existing `libint_potential_type` name; then the new common interface could use the neutral name directly, while existing callers could be migrated separately.

src/coulomb_integral_interface.F
```diff
@@ -143,7 +145,8 @@ SUBROUTINE compute_coulomb_2c(context, la_min, la_max, lb_min, lb_max, npgfa, np
       REAL(dp), DIMENSION(npgfb), INTENT(IN)             :: rpgfb
       REAL(dp), DIMENSION(3), INTENT(IN)                 :: ra, rb
       REAL(dp), DIMENSION(:, :), INTENT(INOUT)           :: hab
-      TYPE(libint_potential_type), INTENT(IN)            :: potential_parameter
+
+      CLASS(coulomb_operator_type), INTENT(IN)           :: potential_parameter
```
Growl1234: Why change it to `CLASS`? I think `TYPE` is OK.

RTChou67: That's just my stupid idea that it may be useful in the future.It seems indeed useless.

Growl1234 (approved):

src/coulomb_integral_interface.F
Growl1234: Since this shouldn't be the blocker in my view, I will consider this as "resolved"; it would be best if this is mentioned in the related issue.

Could you also check other comments and click "Resolve" if you have addressed them?

fstein93 (approved):

src/coulomb_integral_interface.F
RTChou67: Thanks for reminding. I suppose all comments now are Resolved.

Issue #4682, opened 2026-01-09
QM/MM Protein-Ligand Simulations


vishwajitpatil208-alt:
Hello everyone.
I'm working on protein-ligand simulations using QM/MM approach. I tried creating environment as mentioned on one of the websites. However, I'm not able to create a fresh conda environment with all necessary dependencies.

I used following link:
[https://manual.gromacs.org/current/install-guide/index.html](https://github.com/Valdes-Tresanco-MS/gmx_MMPBSA/issues/url)

I would be happy and grateful if someone could provide me an environment.yml file to install and compile GROMACS 2025 with cp2k with necessary dependencies.

Looking forward for the response.

Thanks.

oschuett:
We regularly test CP2K+Gromacs using [this Dockerfile](https://github.com/cp2k/cp2k/blob/master/tools/docker/Dockerfile.test_gromacs). The installation of Gromacs starts [here](
https://github.com/cp2k/cp2k/blob/62f3f8ee301b1a04e07c397ced95301d97190bbe/tools/docker/scripts/test_gromacs.sh#L16).

vishwajitpatil208-alt:
Thanks for the above information @oschuett , 
Do i need to run the docker file first and run as a container and then do i need to execute the test_gromacs.sh inside the containers environment. 
Let's say do i need to approach this way by running the command " docker exec -it <gromacs_container> /bin/bash"

DCM-Uni-Paderborn:
The Dockerfile is meant primarily as a reproducible build-and-test recipe, not as a conda `environment.yml`.

The usual workflow is:

```bash
cd cp2k/tools/docker
podman build --shm-size=1g -f ./Dockerfile.test_gromacs ../../
```

During that `podman build`, the Dockerfile already copies and runs `tools/docker/scripts/test_gromacs.sh`:

```dockerfile
COPY ./tools/docker/scripts/test_gromacs.sh ./
RUN ./test_gromacs.sh 2>&1 | tee report.log
```

So you normally do **not** need to first start a container and then run `docker exec ... ./test_gromacs.sh`; the test is executed as part of the image build. If you want to inspect the environment interactively afterwards, you can run a shell from the built image, but that is optional/debugging only.

For a non-Docker/HPC installation, the important part of the recipe is the same sequence used in `test_gromacs.sh`:

1. Build and install CP2K with CMake.
2. Configure GROMACS with CP2K support, e.g.
   ```bash
   cmake .. \
[...]

PR #5754, opened 2026-08-15
Add internal cross-link

mkrack:


he-zilong:
Do these cross-links in the documentation support special characters like slashes in the target section header?

The issue #4682 could finally be closed after #5742 and #5745 and this one, I suppose.

oschuett:
Currently, [section anchors](https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#syntax-header-anchors) are only auto-generated for the first three heading levels:

https://github.com/cp2k/cp2k/blob/db4011826da0589fa52d24e7d0bd2a812955f836/docs/conf.py#L26

Alternatively, you can always add an [explicit cross-ref target](https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#syntax-targets):
```
(gromacs-cp2k)=
#### GROMACS/CP2K QM/MM
```

mkrack:
> Alternatively, you can always add an [explicit cross-ref target](https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#syntax-targets):
> 
> ```
> (gromacs-cp2k)=
> #### GROMACS/CP2K QM/MM
> ```
Thanks @oschuett. Unfortunately, that does not work because `make_pretty.sh` inserts automatically an empty line between target and heading which MyST does not allow.

docs/methods/qm_mm/gromacs.md
```diff
@@ -12,8 +12,9 @@ molecular complexes, and embedded active sites.
 
 ## Setup Outline
 
-1. Build GROMACS with CP2K QM/MM support and make sure it can find the CP2K executable and data
-   directory.
+1. Build GROMACS with CP2K QM/MM support (see
+   [here](../../getting-started/build-from-source.md#build-gromacs-cp2k)) and make sure it can find
```
oschuett: It seems explicit targets are global ids. So, try the following:

```suggestion
   [here](#build-gromacs-cp2k)) and make sure it can find
```

PR #5755, opened 2026-08-15
Fix second derivatives of overlap integrals

RTChou67:
The current implementation of the first term in ai_overlap:overlap_ab uses `apbm`, which represents (a+1|b-1). however, the recurrence requires `apbp` which is (a+1|b+1). This PR fixes this bug.


mkrack:
That looks indeed like a bug. The recurrence relation in the comment appears correct to me and it does not match the coded one. The impact on the test results, however, is minor. Do you have test case which shows a larger deviation?

RTChou67:
Thanks. I'll think of one

RTChou67:
> That looks indeed like a bug. The recurrence relation in the comment appears correct to me and it does not match the coded one. The impact on the test results, however, is minor. Do you have test case which shows a larger deviation?

I tested two larger asymmetric systems, (H2O)4 and (H2O)8, with EHRENFEST_DYN. Comparing the corrected and uncorrected implementations gives a much larger difference in the Ehrenfest force:

(H2O)4: up to approximately 1.93 × 10^-5 Ha/Bohr;
(H2O)8: up to approximately 3.01 × 10^-5 Ha/Bohr.

The final total-energy differences are smaller, approximately 1.19E-10 Ha for (H2O)4 and 6.17E-11 Ha for (H2O)8.

mkrack:
Thanks for testing. Would it be possible to add a test input checking the Ehrenfest forces?

RTChou67:
> Thanks for testing. Would it be possible to add a test input checking the Ehrenfest forces?
Added.

mkrack:
Run `make_pretty.sh`.

RTChou67:
> Run `make_pretty.sh`.

Thanks for pointing that out.

mkrack:
The new test `QS/regtest-ls-rtp/H2O4-emd-overlap-second-deriv.inp` is quite slow with 98s. Is it possible to accelerate that test, e.g. by running less NVE step?

mkrack:
The test is still too slow with >70s. You can also add the test to the list of [SLOW_TESTS_SUPPRESSIONS](https://github.com/cp2k/cp2k/blob/master/tests/SLOW_TESTS_SUPPRESSIONS) in case it is really not possible to accelerate it further.

RTChou67:
> The test is still too slow with >70s. You can also add the test to the list of [SLOW_TESTS_SUPPRESSIONS](https://github.com/cp2k/cp2k/blob/master/tests/SLOW_TESTS_SUPPRESSIONS) in case it is really not possible to accelerate it further.

Well it is possible, but the thing is that i have to test on the same environment as the CP2K-CI, so I have to push again and again. This time the system is much smaller, so i think it will be ok

RTChou67:
suppressed to 27s but still slow. I guess it's necessary to add it to [SLOW_TESTS_SUPPRESSIONS](https://github.com/cp2k/cp2k/blob/master/tests/SLOW_TESTS_SUPPRESSIONS)

tests/SLOW_TESTS_SUPPRESSIONS
```diff
@@ -38,6 +38,9 @@ QS/regtest-kp-2/cc2.inp
 QS/regtest-kp-3/cc1.inp
 QS/regtest-loc_powf/run.inp
 QS/regtest-ls-rtp/Ar_mixed_aa_planar-rtp-osc-field.inp
+
+
```
mkrack: Why the two blank lines?

RTChou67: sry. fixed

PR #5758, opened 2026-08-15
Remove `src/mod.header` and `src/README.md`

he-zilong:
- `src/mod.header` contains an outdated header template for modules that is not used elsewhere and last updated 2020; commit eee4402 in 2011 removed a `src/sub.header` but preserved `src/mod.header` for some reason.
- `src/README.md` contains essentially only an link pointing to www.cp2k.org/dev:codestructure whose content has been transferred to [the documentation the code structure](https://manual.cp2k.org/trunk/development/code-structure.html) and updated at #5737; even if not, it may be retired in favor of the entire `Development` section of the continuously maintained manual which should be informative and visible enough.


Growl1234:
I think, it would be fine to make `src/README.md` a symbol link of `docs/development/code-structure.md`

he-zilong:
> I think, it would be fine to make `src/README.md` a symbol link of `docs/development/code-structure.md`

Will do in my next PR on docs

PR #5763, opened 2026-08-16
Limit CSCS H100 regtest concurrency

DCM-Uni-Paderborn:
## Summary

- run eight concurrent two-rank, single-threaded H100 regtest workers instead of 56
- allow up to 400 seconds for slower ARM/GH200 SIRIUS cases
- extend the SLURM allocation safeguard to 90 minutes
- keep every test, matcher, dependency, and GPU backend enabled

## Root cause

The H100 dashboard consistently failed with about 49-51 timed-out tests. Its allocation launched 56 concurrent regression-test directories on four GPUs, placing 28 simultaneous MPI ranks on each GPU.

The first revision reduced this to 16 workers and cut the failures from 51 to 12. The second revision switched to eight single-threaded workers and reduced the result further to three timeouts in a 37-minute run. The earlier SIRIUS full-potential Cholesky failure did not recur. Because worker concurrency and OpenMP thread count changed together, this observation does not establish a SIRIUS/OpenMP defect.

A subsequent run with a 300-second process timeout passed 5949 of 5950 tests in 38 minutes. The only remaining failure was `SIRIUS/regtest-1/Au.inp`, which continued normal SCF iterations until it was terminated after 302.62 seconds.

PR #5721 lowered the current global regtest timeout from the former 400 seconds to 150 seconds. This H100 runner now explicitly uses the former 400-second threshold. With `--ompthreads 1` and 16 task slots, `do_regtest.py` creates eight two-rank workers, or two workers per GPU. The 90-minute SLURM limit remains a protective bound and leaves headroom for the slower test. The separate H100 benchmark jobs still exercise hybrid MPI/OpenMP execution.

## Validation

- `./make_pretty.sh ci/cscs_daint_spack_psmp_h100.yaml`
- YAML parsing and worker/rank/timeout check
- `git diff --check`
- CP2K pre-push hook

### DIFF
--- a/ci/cscs_daint_spack_psmp_h100.yaml
+++ b/ci/cscs_daint_spack_psmp_h100.yaml
@@ -95,7 +95,7 @@ regression test cp2k daint:
   timeout: 6h
   script:
     - git --no-pager log -1 --pretty="%nCommitSHA:%x20%H%nCommitTime:%x20%ci%nCommitAuthor:%x20%an%nCommitSubject:%x20%s%n"
-    - podman run --device nvidia.com/gpu=all --shm-size=1g -v ${PWD}:/mnt ${DOCKERHUB_IMAGE_NAME} /opt/cp2k/tests/do_regtest.py --cp2kdatadir /opt/cp2k/install/share/cp2k/data --maxtasks ${SLURM_NTASKS} --ompthreads 1 --mpiexec "mpiexec -bind-to none" --keepalive /opt/cp2k/install/bin ${VERSION} 2>&1 || true
+    - podman run --device nvidia.com/gpu=all --shm-size=1g -v ${PWD}:/mnt ${DOCKERHUB_IMAGE_NAME} /opt/cp2k/tests/do_regtest.py --cp2kdatadir /opt/cp2k/install/share/cp2k/data --maxtasks ${SLURM_NTASKS} --ompthreads 1 --timeout 300 --mpiexec "mpiexec -bind-to none" --keepalive /opt/cp2k/install/bin ${VERSION} 2>&1 || true
   variables:
     SLURM_DEBUG: 1
     SLURM_HINT: nomultithread
--- a/ci/cscs_daint_spack_psmp_h100.yaml
+++ b/ci/cscs_daint_spack_psmp_h100.yaml
@@ -95,13 +95,14 @@ regression test cp2k daint:
   timeout: 6h
   script:
     - git --no-pager log -1 --pretty="%nCommitSHA:%x20%H%nCommitTime:%x20%ci%nCommitAuthor:%x20%an%nCommitSubject:%x20%s%n"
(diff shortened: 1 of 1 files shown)


mkrack:
> suite takes at most 72 seconds in the successful Spack pdbg check. The H100 job therefore uses `--timeout 300`; even if all three remaining tests consume the full allowance, the run remains within its existing 60-minute SLURM limit.

The SLURM timelimit could be increased. It serves mainly as a protective mechanism against excessive resource waste.

ci/cscs_daint_spack_psmp_h100.yaml
```diff
@@ -95,12 +95,14 @@ regression test cp2k daint:
...
     - git --no-pager log -1 --pretty="%nCommitSHA:%x20%H%nCommitTime:%x20%ci%nCommitAuthor:%x20%an%nCommitSubject:%x20%s%n"
-    - podman run --device nvidia.com/gpu=all --shm-size=1g -v ${PWD}:/mnt ${DOCKERHUB_IMAGE_NAME} /opt/cp2k/tests/do_regtest.py --cp2kdatadir /opt/cp2k/install/share/cp2k/data --maxtasks ${SLURM_NTASKS} --mpiexec "mpiexec -bind-to none" --keepalive /opt/cp2k/install/bin ${VERSION} 2>&1 || true
+    - podman run --device nvidia.com/gpu=all --shm-size=1g -v ${PWD}:/mnt ${DOCKERHUB_IMAGE_NAME} /opt/cp2k/tests/do_regtest.py --cp2kdatadir /opt/cp2k/install/share/cp2k/data --maxtasks ${SLURM_NTASKS} --ompthreads 1 --timeout 300 --mpiexec "mpiexec -bind-to none" --keepalive /opt/cp2k/install/bin ${VERSION} 2>&1 || true
   variables:
     SLURM_DEBUG: 1
     SLURM_HINT: nomultithread
     SLURM_JOB_NUM_NODES: 1
-    SLURM_NTASKS: 224
+    # Run eight two-rank workers to limit device contention. SIRIUS full-
+    # potential tests are unstable with multiple OpenMP threads.
```
mkrack: Has that already been communicated upstream?

DCM-Uni-Paderborn: No, not yet. More importantly, these runs changed both worker concurrency and OpenMP thread count, so they do not isolate a SIRIUS/OpenMP issue. I have reworded the comment to describe `--ompthreads 1` as a resource-allocation choice. If the Cholesky failure recurs under low device contention, I will prepare a reproducer and report it upstream.

DCM-Uni-Paderborn:
Thanks, that makes sense. I have increased `SLURM_TIMELIMIT` from 60 to 90 minutes, so the 300-second per-test timeout has enough headroom while the allocation still retains a protective limit.

mkrack:
Great, only one test input failed. Do you think, it would pass with the default time limit of 400s?

DCM-Uni-Paderborn:
> Great, only one test input failed. Do you think, it would pass with the default time limit of 400s?

Dear Matthias, 

yes I think that is worth testing. Au.inp was still making normal SCF progress when it was terminated after 302.6 seconds, and it is now the only remaining failure. One detail is that #5721 reduced the current default timeout from 400 to 150 seconds, which is why this runner currently overrides it with 300 seconds. I will raise the H100-specific timeout to the former 400-second default and rerun the test.

Greetings, 
Thomas

mkrack (approved):

PR #5765, opened 2026-08-17
Declare OpenMP data sharing in native SKALA paths

DCM-Uni-Paderborn:
## Summary

- add `DEFAULT(NONE)` to the six OpenMP regions introduced for native SKALA atom-composite execution in #5759;
- declare their shared data explicitly while preserving the existing private variables and reductions;
- restore the Coding conventions dashboard check without changing the numerical algorithm.

## Root cause

The dashboard reports the failures at current head `98357aa`, but the two new convention issues originate from the preceding merge commit `c4ed851` (#5759). Its new OpenMP regions relied on implicit data sharing, whereas CP2K requires `DEFAULT(NONE)`.

The compilation failures shown at the same dashboard head have a separate root cause in `torch_api.F` and are handled by #5764. The CSCS H100 failure is likewise independent and is handled by #5763. This PR intentionally does not duplicate either fix.

## Validation

- `make_pretty.sh src/qs_vxc_atom.F`;
- `git diff --check`;
- GCC 16 OpenMP debug compilation of `qs_vxc_atom.F`;
- CP2K's gfortran AST convention analyzer: zero issues and zero warnings for `qs_vxc_atom.F`;
- complete local GCC 16 OpenMP debug build, with #5764 applied locally because current master does not compile without it;
- complete Spark AArch64 GCC 13 build with MPI, OpenMP, GauXC, LibTorch, LibXC, Libint and Spglib, again with #5764 applied locally;
- `H2_NATIVE_SKALA_GAPW_GTH_ATOM_COMPOSITE.inp` on Spark with one and four OpenMP threads. Energy, total force, stress trace and atom-composite electron count are identical at the printed precision and within the existing regression-test tolerances.

### DIFF
--- a/src/qs_vxc_atom.F
+++ b/src/qs_vxc_atom.F
@@ -966,12 +966,27 @@ CONTAINS
                         composite_image_periodicity, composite_descriptor_image_coords, &
                         composite_descriptor_target_image)
                   END IF
-!$OMP PARALLEL DO COLLAPSE(2) IF (image_partition_atom_composite) SCHEDULE(STATIC) &
+!$OMP PARALLEL DO COLLAPSE(2) IF (image_partition_atom_composite) SCHEDULE(STATIC) DEFAULT(NONE) &
 !$OMP PRIVATE(composite_row, composite_point, composite_smooth_density_value, &
 !$OMP         composite_smooth_gradient_value, composite_smooth_kin_value, &
 !$OMP         descriptor_window_weight, idir, ispin, &
 !$OMP         gth_potential, nlcc_density, nlcc_gradient, nlcc_hessian, nlcc_spin_factor, &
-!$OMP         interpolation_stencil, partition_scale, partition_weight, sgp_potential, source_atom)
+!$OMP         interpolation_stencil, partition_scale, partition_weight, sgp_potential, source_atom) &
+!$OMP SHARED(atom_composite_reference, cell, composite_atom_coords, composite_atom_kind, &
+!$OMP        composite_atom_start, composite_atomic_grid_weights, composite_base_grid_weights, &
+!$OMP        composite_density, composite_descriptor_image_coords, &
+!$OMP        composite_descriptor_target_image, composite_distances, composite_grad, &
+!$OMP        composite_grid_coords, composite_grid_weights, composite_kin, &
+!$OMP        composite_partition_atom_coords, composite_partition_image_coords, &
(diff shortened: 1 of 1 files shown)


Growl1234:
Looks like this causes Intel ssmp failures:
```
/opt/cp2k/build/src/qs_vxc_atom.F(1667): error #6752: Since the OpenMP* DEFAULT(NONE) clause applies, the PRIVATE, SHARED, REDUCTION, FIRSTPRIVATE, or LASTPRIVATE attribute must be explicitly specified for every variable.   [LOCAL_PARTITION_WEIGHT]
                                       local_partition_dstrain(3, 3), local_partition_weight
----------------------------------------------------------------------^
/opt/cp2k/build/src/qs_vxc_atom.F(1666): error #6752: Since the OpenMP* DEFAULT(NONE) clause applies, the PRIVATE, SHARED, REDUCTION, FIRSTPRIVATE, or LASTPRIVATE attribute must be explicitly specified for every variable.   [LOCAL_PARTITION_DATOM]
                                       local_descriptor_dstrain(3, 3), local_partition_datom(3, SIZE(particle_set)), &
-----------------------------------------------------------------------^
/opt/cp2k/build/src/qs_vxc_atom.F(1667): error #6752: Since the OpenMP* DEFAULT(NONE) clause applies, the PRIVATE, SHARED, REDUCTION, FIRSTPRIVATE, or LASTPRIVATE attribute must be explicitly specified for every variable.   [LOCAL_PARTITION_DSTRAIN]
                                       local_partition_dstrain(3, 3), local_partition_weight
---------------------------------------^
/opt/cp2k/build/src/qs_vxc_atom.F(1665): error #6752: Since the OpenMP* DEFAULT(NONE) clause applies, the PRIVATE, SHARED, REDUCTION, FIRSTPRIVATE, or LASTPRIVATE attribute must be explicitly specified for every variable.   [LOCAL_DESCRIPTOR_DATOM]
                           REAL(dp) :: local_descriptor_datom(3, 1), &
---------------------------------------^
/opt/cp2k/build/src/qs_vxc_atom.F(1666): error #6752: Since the OpenMP* DEFAULT(NONE) clause applies, the PRIVATE, SHARED, REDUCTION, FIRSTPRIVATE, or LASTPRIVATE attribute must be explicitly specified for every variable.   [LOCAL_DESCRIPTOR_DSTRAIN]
                                       local_descriptor_dstrain(3, 3), local_partition_datom(3, SIZE(particle_set)), &
---------------------------------------^
```

PR #5776, opened 2026-08-19
Fix DDAPC forces on empty MPI ranks

### LINKED ISSUE #5774: Bug with NaN forces when number of number of spherical basis functions is lower than the number of MPI processes
When number of spherical basis function is lower than the number of MPI processes `ch_pulay` and `other` component of forces become NaN.
Apparently problem is with work distribution and gather of forces over the MPI.

Attached input files to reproduce the bug.
System in the input has 121 spherical basis functions, so the bug appears on 122+ MPI processes (attached also sample output with 122 MPI processes that reproduces the problem)

Tested on the master branch, but may also affect release branches.

[pbe_cp2k.inp](https://drive.google.com/file/d/12jIHw_xuOhsrorJXroZxaPqyQ_q82DrO/view?usp=sharing)
[pbe_cp2k.pdb](https://drive.google.com/file/d/1eSZRANLDow5WqzXH24m_ofsGuoDSUa8H/view?usp=sharing)
[pbe_cp2k_nan_forces.out](https://drive.google.com/file/d/1tnMg5pgl-1xseP5jYIeUCNNdlY3rWdNS/view?usp=sharing)

DCM-Uni-Paderborn:
## Summary

- initialize DDAPC derivative-matrix rows on MPI ranks without local G-vectors below the cutoff
- pass valid zero-sized trigonometric work arrays through the empty-rank path
- add a four-rank QMMM regression test with fewer spherical basis functions than MPI ranks

## Root cause

When a plane-wave distribution left an MPI rank without G-vectors below `GCUT`, `build_der_A_matrix_rows` skipped its DGEMM path without initializing the requested rows of `dAm`. Those uninitialized values propagated into the DDAPC charge derivatives and contaminated the globally reduced `ch_pulay` forces with NaNs.

## Validation

- built `cp2k.psmp` from the current `master` source with MPI enabled
- CP2K precommit checks pass for all changed files
- four-rank `QMMM/QS/regtest-mpi` run: 1/1 correct

Fixes #5774


Growl1234 (approved):
From https://github.com/cp2k/cp2k/issues/5774#issuecomment-5342975254 the issue author has confirmed that this PR fixes that issue.

PR #5780, opened 2026-08-19
RIRS: dedicated &RI_RS input section, linear-dependency fix in the GW minimax grid, and evGW0

RitajTyagi:
Three changes to the real-space RI GW (RI-RS GW) code:

1. the RI-RS keywords move into a dedicated `&RI_RS` input section,
2. the placeholder eigenvalues of linearly dependent basis modes are kept out of the
   GW minimax grid (this changes G0W0 results on affected systems),
3. evGW0 is added for non-periodic systems.

### New `&RI_RS` input section

The RI-RS keywords move out of `&GW` into a dedicated `&RI_RS` subsection. This
changes existing RI-RS inputs: the old logical `RI_RS` becomes
`&RI_RS ... &END RI_RS`, and the keywords below now sit inside it.

`TIKHONOV`, `GRID_SELECT`, `GRID_FILE_SUFFIX`, `CUTOFF_RADIUS_RL_RI`,
`CUTOFF_RADIUS_RL_AO`, `CUTOFF_RADIUS_RL_W`, `CUTOFF_RADIUS_G_W`, `N_PANELS`,
`N_PROCS_PER_ATOM_Z_LP`, `KEEP_SPARSITY_RL`

Tests 01-06 are updated to the new form, as are the five RI-RS tests in
`regtest-rtbse-linearized-rirs` and `regtest-rtbse-linearized-open-shell`; their
references are unchanged. The section description was rewritten while moving it, which
also removes two `\{ ... \}` constructs that MathJax failed to render in the manual.

### Linear-dependency placeholders - this also changes G0W0

The canonical orthogonalization drops linearly dependent basis modes and parks their
eigenvalues at 10000 Ha. GW was treating those as real states when setting up the
minimax grid. On a system with 132 dependent modes this gave `E_max = 272383 eV` and
`E_range = 147471` instead of about 235, and the Pade continuation dropped to 7
parameters instead of 16, since `nparam_pade` is capped by the number of minimax
points below `FREQ_MAX_FIT`.

`cp_fm_geeig_canon` and `cp_cfm_geeig_canon` now optionally return how many modes they
kept, and the GW grid setup uses that count instead of the length of the eigenvalue
array. Details:

- The states are selected by count, not by testing the eigenvalue value. The
  placeholder value is defined once as `set_removed_eigval_to` in `cp_fm_diag` and
  used in both diag routines.
- For k-points the retained count can differ between k-points. The minimum over the
  k-point mesh is used, since the minimax energy range is a single global quantity.

Systems without linear dependencies are unaffected - the references for tests 01-06
are unchanged.

### evGW0 (experimental)

evGW0 for molecules: the quasiparticle energies are iterated in the Green's function
while W stays frozen at RPA@DFT, so each cycle only repeats the correlation
self-energy and the quasiparticle solve. The grid, `Phi_mu_l`, `Z_lP`, W and
`Sigma^x` are computed once and reused.

evGW0 is activated by an `&EVGW0` subsection inside `&RI_RS`. Without that section a
single-shot G0W0 calculation is performed, so nothing changes unless evGW0 is
requested explicitly.

| Keyword | Default | |
|---|---|---|
| `MAX_ITER` | 20 | maximum number of evGW0 cycles |
| `EPS_ITER` | 5e-5 | Ha |

```
&RI_RS
  &EVGW0
    MAX_ITER 20
    EPS_ITER 0.00005
  &END EVGW0
&END RI_RS
```

All states are updated in G. The exception is the linear-dependency placeholders
described above: `eigenval_G0W0` carries a finite number for those indices, so copying
it there would drop the removed modes into the physical spectrum and let them
contribute to G. Entries above the retained count are therefore left alone. The
chemical potential is recomputed from the quasiparticle energies in every cycle.

The cycle stops once the quasiparticle HOMO, the LUMO and the HOMO-LUMO gap all change
by less than `EPS_ITER` between two successive cycles. Those three values are printed after 
every cycle, and the full spectrum of every cycle is appended to `bandstructure_SCF_and_G0W0` 
under an `evGW0 cycle: N` header. Open shell is supported; both spin channels must satisfy the
convergence criterion.

Requesting evGW0 prints a warning that the feature is experimental. In G0W0 an
analytic-continuation error affects one state; in evGW0 it is fed back into G and
accumulates over cycles. Against FHI-aims evGW0 on small molecules the deviation
stayed below 100 meV; this has not been checked beyond small molecules.

### Tests

Two new: `07_RIRS_EVGW0_GPW_PBE_H2O` (closed shell) and `08_RIRS_EVGW0_PBE0_O2` (open
shell). 

All GW and RT-BSE regtest suites pass locally, 191/191: `regtest-gw`,
`regtest-gw-cubic`, `regtest-gw2x`, `regtest-gw-ic-model`, `regtest-gw-realspace`,
`regtest-scalable-gw`, `regtest-rtbse`, `regtest-rtbse-linearized`,
`regtest-rtbse-linearized-open-shell`, `regtest-rtbse-linearized-restart` and
`regtest-rtbse-linearized-rirs`.


fstein93 (approved):

src/gw_large_cell_gamma.F
```diff
@@ -2468,29 +2469,61 @@ SUBROUTINE compute_QP_energies(bs_env, qs_env, fm_Sigma_x_Gamma, fm_Sigma_c_Gamm
       ALLOCATE (Sigma_c_ikp_n_time(bs_env%n_ao, bs_env%num_time_freq_points, 2))
       ALLOCATE (Sigma_c_ikp_n_freq(bs_env%n_ao, bs_env%num_time_freq_points, 2))
 
+      ! Steps 1-5 below only depend on the (frozen) Kohn-Sham orbitals, on v^xc and on Σ^x, so
+      ! in an evGW0 cycle they give the same result in every iteration. Cache them on the first
+      ! call and reuse them afterwards. The cache holds one complex n_ao x n_ao matrix per
+      ! k-point and spin, which is why it is only built for the molecular single-k-point case.
```
JWilhelm: Does this caching safe a lot of computation time? If not, I would be in favor of avoiding this because it makes the code considerably more complicated at little computational benefit.

src/input_cp2k_properties_dft.F
```diff
@@ -2733,10 +2763,85 @@ SUBROUTINE create_gw_section(section)
       CALL section_add_keyword(section, keyword)
       CALL keyword_release(keyword)
 
-      CALL section_add_subsection(section, subsection)
-      CALL section_release(subsection)
+      CALL keyword_create(keyword, __LOCATION__, name="SELF_CONSISTENCY", &
```
JWilhelm: would it be also possible to have a section 

&EVGW0

that activates evGW0, and you could put all evGW0 parameters inside that section? Then you could remove the keyword SELF_CONSISTENCY

src/post_scf_bandstructure_types.F
```diff
@@ -87,6 +88,36 @@ MODULE post_scf_bandstructure_types
       REAL(KIND=dp)                                    :: cutoff_radius_g_w = -1.0_dp
       CHARACTER(LEN=default_string_length)             :: grid_file_suffix = ""
 
+      ! Eigenvalue self-consistency (evGW0): G is rebuilt from the quasiparticle energies of
+      ! the previous cycle while W stays frozen at the RPA@DFT level.
+      INTEGER                                          :: self_consistency = G0W0
+      INTEGER                                          :: sc_gw0_iter = 20
```
JWilhelm: please unify the notation, always evGW0, never sc_gw0

src/post_scf_bandstructure_types.F
```diff
@@ -87,6 +88,36 @@ MODULE post_scf_bandstructure_types
...
+      REAL(KIND=dp)                                    :: eps_iter = 5.0E-5_dp
+      LOGICAL                                          :: do_ev_sc = .FALSE.
```
JWilhelm: please unify notation, evGW0

src/post_scf_bandstructure_types.F
```diff
@@ -87,6 +88,36 @@ MODULE post_scf_bandstructure_types
...
+      ! Number of occupied and of empty states per spin channel, counted by Kohn-Sham index
+      ! outwards from the gap. These are the states whose quasiparticle energies decide
+      ! convergence and the only ones fed back into the Green's function
+      INTEGER                                          :: evgw0_nstate = 20
```
JWilhelm: do we need this parameter? isn't evGW0 done on all states?

src/post_scf_bandstructure_types.F
```diff
@@ -87,6 +88,36 @@ MODULE post_scf_bandstructure_types
...
+      ! .TRUE.  : recompute the chemical potential from the quasiparticle energies every cycle
+      !           (the convention of CP2K's RI-V evGW0)
+      ! .FALSE. : keep it at the SCF value for the whole cycle (the convention of FHI-AIMS)
+      LOGICAL                                          :: evgw0_update_fermi = .TRUE.
+      ! Outside the EVGW0_NSTATE window: .TRUE. displaces the state by the mean correction of
+      ! the states inside it of the same character, .FALSE. keeps its DFT eigenvalue.
+      LOGICAL                                          :: evgw0_shift_outer = .TRUE.
```
JWilhelm: also not needed if we correct all states from evGW0

src/post_scf_bandstructure_types.F
```diff
@@ -87,6 +88,36 @@ MODULE post_scf_bandstructure_types
...
+      REAL(KIND=dp)                                    :: eps_iter = 5.0E-5_dp
+      LOGICAL                                          :: do_ev_sc = .FALSE.
+      ! Number of occupied and of empty states per spin channel, counted by Kohn-Sham index
+      ! outwards from the gap. These are the states whose quasiparticle energies decide
+      ! convergence and the only ones fed back into the Green's function
+      INTEGER                                          :: evgw0_nstate = 20
+      ! .TRUE.  : recompute the chemical potential from the quasiparticle energies every cycle
+      !           (the convention of CP2K's RI-V evGW0)
+      ! .FALSE. : keep it at the SCF value for the whole cycle (the convention of FHI-AIMS)
+      LOGICAL                                          :: evgw0_update_fermi = .TRUE.
```
JWilhelm: do we need this logical?

src/post_scf_bandstructure_types.F
```diff
@@ -87,6 +88,36 @@ MODULE post_scf_bandstructure_types
...
+      ! Outside the EVGW0_NSTATE window: .TRUE. displaces the state by the mean correction of
+      ! the states inside it of the same character, .FALSE. keeps its DFT eigenvalue.
+      LOGICAL                                          :: evgw0_shift_outer = .TRUE.
+
+      ! Caches of the evGW0-invariant quantities. Only filled when do_ev_sc is true; every
```
JWilhelm: I agree we need to store W, but not the other quantities. Can you please shrink the comment to a single line?

src/post_scf_bandstructure_types.F
```diff
@@ -153,6 +184,12 @@ MODULE post_scf_bandstructure_types
       TYPE(ri_rs_env)                                 :: ri_rs
 
       ! general parameters on molecular orbitals and basis sets
+      ! Number of basis modes that survived the linear-dependency filter of the canonical
```
JWilhelm: please shrink comment to a single line

src/post_scf_bandstructure_types.F
```diff
@@ -87,6 +88,36 @@ MODULE post_scf_bandstructure_types
...
+      ! .TRUE.  : recompute the chemical potential from the quasiparticle energies every cycle
+      !           (the convention of CP2K's RI-V evGW0)
+      ! .FALSE. : keep it at the SCF value for the whole cycle (the convention of FHI-AIMS)
+      LOGICAL                                          :: evgw0_update_fermi = .TRUE.
+      ! Outside the EVGW0_NSTATE window: .TRUE. displaces the state by the mean correction of
+      ! the states inside it of the same character, .FALSE. keeps its DFT eigenvalue.
+      LOGICAL                                          :: evgw0_shift_outer = .TRUE.
+
+      ! Caches of the evGW0-invariant quantities. Only filled when do_ev_sc is true; every
+      ! consumer falls back to recomputing them when the corresponding array is unallocated,
```
JWilhelm: can you please check in general all Claude comments for strange words like "consumer"?

src/post_scf_bandstructure_utils.F
```diff
@@ -675,9 +677,14 @@ SUBROUTINE compute_cfm_mo_coeff_kp_and_eigenval_scf_kp(qs_env, bs_env)
 
             ! Diagonalize KS-matrix via Rothaan-Hall equation:
             ! H^KS(k) C(k) = S(k) C(k) ε(k)
+            ! nmo_retained counts the basis modes that survived the linear-dependency
```
JWilhelm: can you please remove this comment? now it makes the impression the nmo_retained is the most important thing in that routine, while it almost never is relevant

src/post_scf_bandstructure_utils.F
```diff
@@ -776,27 +783,32 @@ SUBROUTINE diagonalize_ks_matrix(bs_env)
...
+
       DO ispin = 1, bs_env%n_spin
 
          ! use work matrices because the matrices are overwritten in cp_fm_geeig_canon
          CALL cp_fm_to_fm(bs_env%fm_ks_Gamma(ispin), bs_env%fm_work_mo(1))
          CALL cp_fm_to_fm(bs_env%fm_s_Gamma, bs_env%fm_work_mo(2))
 
          ! diagonalize the Kohn-Sham matrix to obtain MO coefficients and SCF eigenvalues
-         ! (at the Gamma-point)
+         ! (at the Gamma-point). nmo_retained counts the basis modes that survived the
```
JWilhelm: can you please remove this comment? now it makes the impression the nmo_retained is the most important thing in that routine, while it almost never is relevant

src/post_scf_bandstructure_utils.F
```diff
@@ -776,27 +783,32 @@ SUBROUTINE diagonalize_ks_matrix(bs_env)
...
 
-      INTEGER                                            :: handle, ikp, ispin
+      INTEGER                                            :: handle, ikp, ispin, nmo_retained
       REAL(KIND=dp)                                      :: CBM, VBM
 
       CALL timeset(routineN, handle)
 
       ALLOCATE (bs_env%eigenval_scf_Gamma(bs_env%n_ao, bs_env%n_spin))
 
+      bs_env%n_mo_retained = bs_env%n_ao
```
JWilhelm: please set this in 

get_parameters_from_qs_env

src/post_scf_bandstructure_utils.F
```diff
@@ -643,6 +644,7 @@ SUBROUTINE compute_cfm_mo_coeff_kp_and_eigenval_scf_kp(qs_env, bs_env)
 
       ! nkp_bs_and_DOS contains desymmetrized k-point mesh from SCF and k-points from GW bandstructure
       nkp_bs_and_DOS = bs_env%nkp_bs_and_DOS
+      bs_env%n_mo_retained = bs_env%n_ao
```
JWilhelm: this seems to be done twice, should be done in get_parameters_from_qs_env

JWilhelm (changes requested):

src/gw_large_cell_gamma.F
RitajTyagi: the computational time is far less compared to other routines like Z_coeff and Sigma_c. I agree with your point and it has been removed now except for W matrices.

src/input_cp2k_properties_dft.F
RitajTyagi: Created a separate section for EVGW0

src/post_scf_bandstructure_types.F
RitajTyagi: Done !

src/post_scf_bandstructure_types.F
RitajTyagi: Done !

src/post_scf_bandstructure_types.F
RitajTyagi: All states are updated now so this parameter not required

src/post_scf_bandstructure_types.F
RitajTyagi: Removed

src/post_scf_bandstructure_types.F
RitajTyagi: Removed

src/post_scf_bandstructure_types.F
RitajTyagi: Done

src/post_scf_bandstructure_types.F
RitajTyagi: Done

src/post_scf_bandstructure_types.F
RitajTyagi: Checked

src/post_scf_bandstructure_utils.F
RitajTyagi: Done

src/post_scf_bandstructure_utils.F
RitajTyagi: Done

src/post_scf_bandstructure_utils.F
RitajTyagi: Done

src/post_scf_bandstructure_utils.F
RitajTyagi: Done only once

src/gw_non_periodic_ri_rs.F
```diff
@@ -206,28 +209,221 @@ SUBROUTINE gw_calc_non_periodic_ri_rs(qs_env, bs_env)
...
-      !      Σ^c_λσ(iτ)    =  Σ_ll' Φ_λ(r_l) Σ^c_ll'(iτ) Φ_σ(r_l')
-      ! ========================================================================
-      CALL compute_Sigma_c(bs_env, fm_W_time, bs_env%ri_rs%mat_phi_mu_l, &
-                           bs_env%ri_rs%mat_Z_lP, fm_Sigma_c_Gamma_time)
-
-      ! ========================================================================
-      ! 8. Quasiparticle energies (analytic continuation iτ -> iω -> real ϵ)
-      !      Σ^c_λσ(iτ) -> Σ^c_nn(ϵ)
-      !      ϵ_n^GW = ϵ_n^DFT + Σ^c_nn(ϵ_n^GW) + Σ^x_nn - v^xc_nn
+      ! 7./8. Correlation self-energy Σ^c and quasiparticle energies, iterated
```
JWilhelm: Why is it 7./8. ?

I would somehow favor to write the equation that were previously as comment there again here as comment. Then all GW RIRS equations are in a compact form in the routine gw_calc_non_periodic_ri_rs. What would you think?

src/gw_non_periodic_ri_rs.F
```diff
@@ -206,28 +209,221 @@ SUBROUTINE gw_calc_non_periodic_ri_rs(qs_env, bs_env)
...
+                  'evGW0 eigenvalue self-consistency reached in', i_iter, ' cycles.'
+               WRITE (u, '(A)') ' '
+            END IF
+            EXIT
+         END IF
+
+         IF (i_iter == n_iter) EXIT
+
+         ! ---- update the eigenvalues entering G for the next cycle ----
+         ! ikp = 1 is the Gamma point: this is the non-periodic (molecular) path, so
```
JWilhelm: is this comment about ikp = 1 = Gamma-point necessary here? This is assumed in the whole code, not just at this specific spot. Maybe removing this comment would be better...

src/gw_non_periodic_ri_rs.F
```diff
@@ -206,28 +209,221 @@ SUBROUTINE gw_calc_non_periodic_ri_rs(qs_env, bs_env)
...
+         ! eigenval_scf_Gamma and eigenval_G0W0(:, 1, :) describe the same k-point.
+         bs_env%eigenval_evGW0(:, :, :) = bs_env%eigenval_G0W0(:, :, :)
+         DO ispin = 1, bs_env%n_spin
+            ! Every real state is updated. Entries above n_mo_retained are the placeholders of
```
JWilhelm: please remove this comment or shrink to a single line, that's no important information here.

src/gw_non_periodic_ri_rs.F
```diff
@@ -206,28 +209,221 @@ SUBROUTINE gw_calc_non_periodic_ri_rs(qs_env, bs_env)
...
+
+      DO i_iter = 1, n_iter
+
+         bs_env%ri_rs%evgw0_i_iter = i_iter
+
+         CALL compute_Sigma_c(bs_env, fm_W_time, bs_env%ri_rs%mat_phi_mu_l, &
+                              bs_env%ri_rs%mat_Z_lP, fm_Sigma_c_Gamma_time)
+
+         ! W has been copied into the DBCSR cache; the dense copy is no longer needed
+         IF (bs_env%ri_rs%do_evgw0 .AND. i_iter == 1) CALL cp_fm_release(fm_W_time)
```
JWilhelm: the memory win by deallocating fm_W_time should be negligible, maybe for code readability only deallocate after i_iter loop. Then also the "ELSE" before "CALL cp_fm_release(fm_W_time)" below can be dropped.

src/gw_non_periodic_ri_rs.F
```diff
@@ -4363,16 +4565,25 @@ SUBROUTINE compute_Sigma_c(bs_env, fm_W_time, mat_phi_mu_l, mat_Z_lP, fm_Sigma_c
       DO i_t = 1, bs_env%num_time_freq_points
          tau = bs_env%imag_time_points(i_t)
 
-         CALL dbcsr_create(matrix_W_aux, "W_aux", dist_aux_aux, dbcsr_type_no_symmetry, blk_aux, blk_aux)
-         IF (bs_env%ri_rs%cutoff_radius_g_w > 0.0_dp .AND. &
-             ALLOCATED(bs_env%ri_rs%atom_centers)) THEN
-            CALL reserve_blocks_within_radius(matrix_W_aux, bs_env%ri_rs%atom_centers, &
-                                              bs_env%ri_rs%cutoff_radius_g_w)
-            CALL copy_fm_to_dbcsr(fm_W_time(i_t), matrix_W_aux, keep_sparsity=.TRUE.)
+         IF (do_cache_W) THEN
```
JWilhelm: now I am confused about the computational saving because of caching W: To me it seems we only win by caching W that we don't need to do this: 

Image

If I'm not wrong, this operation is super fast. I would understand if you like not to execute W_ll' = Z_lP W_PQ Z_l'Q, but this is probably a bad idea because W_ll'(it) for all t points is probably occupying a lot of memory.

Or am I overlooking something that one gains because of caching W?

src/fm/cp_fm_diag.F
```diff
@@ -83,6 +83,14 @@ MODULE cp_fm_diag
 
    REAL(KIND=dp), PARAMETER, PUBLIC :: eps_check_diag_default = 5.0E-14_dp
 
+   ! Eigenvalue assigned to the basis modes that the canonical orthogonalization removes
```
JWilhelm: can you please reduce the length of that comment to at most 2 lines?

JWilhelm (changes requested):

src/fm/cp_fm_diag.F
RitajTyagi: Done

src/gw_non_periodic_ri_rs.F
RitajTyagi: You're right. I checked and the cache only skips the dbcsr_create / reserve_blocks_within_radius / copy_fm_to_dbcsr / dbcsr_filter step, the panel contraction runs every cycle either way. On a medium-sized molecule contract_grid_panels_sigma_c is 873 s of the 890 s spent in compute_Sigma_c, out of a 1674 s run, so everything the cache could possibly save sits inside a ~17 s remainder
that also contains build_G_ao.

Also the cache holds W_PQ in the RI basis (n_RI x n_RI), not W_ll'(it) — setup_square_topology uses the column block sizes of Z_lP. So it's the same shape as fm_W_time, not a grid-sized object. But both are alive during the first cycle, so the peak still goes up rather than down.

Whether it would start to pay off for really large systems I don't know, I haven't tested that, so I don't want to claim it does. For now I've removed it. fm_W_time is released once after the loop and the ELSE is gone, as you suggested.

src/gw_non_periodic_ri_rs.F
RitajTyagi: Done

src/gw_non_periodic_ri_rs.F
RitajTyagi: Removed

src/gw_non_periodic_ri_rs.F
RitajTyagi: The 7/8 is there since these two are different steps wrap in one routine. But I see your point. I have modified the description.

src/gw_non_periodic_ri_rs.F
```diff
@@ -171,63 +174,249 @@ SUBROUTINE gw_calc_non_periodic_ri_rs(qs_env, bs_env)
...
+
+         ! eigenvalues currently in G; the first cycle starts from the DFT ones
+         IF (.NOT. ALLOCATED(bs_env%eigenval_evGW0)) THEN
+            ALLOCATE (bs_env%eigenval_evGW0(bs_env%n_ao, bs_env%nkp_bs_and_DOS, bs_env%n_spin))
+         END IF
+         bs_env%eigenval_evGW0(:, :, :) = bs_env%eigenval_scf(:, :, :)
+         band_prev(:, :) = 0.0_dp
+         ! the loop overwrites these; post-GW printing expects the DFT values back
+         ALLOCATE (eigenval_scf_Gamma_dft, SOURCE=bs_env%eigenval_scf_Gamma)
+         e_fermi_dft(:) = bs_env%e_fermi(:)
```
JWilhelm, line 348: Can you please move this entire block (line 273-283) into a routine init_evGW0 ? The main issue in SUBROUTINE compute_Sigma_c_and_QP_energies is mixing many hierarchies: high level routines compute_Sigma_c and compute_QP_energies and big blocks of initialization and printing

src/gw_non_periodic_ri_rs.F
```diff
@@ -171,63 +174,249 @@ SUBROUTINE gw_calc_non_periodic_ri_rs(qs_env, bs_env)
...
+
+         IF (converged) THEN
+            IF (u > 0) THEN
+               WRITE (u, '(A)') ' '
+               WRITE (u, '(T2,A,I4,A)') &
+                  'evGW0 eigenvalue self-consistency reached in', i_iter, ' cycles.'
+               WRITE (u, '(A)') ' '
+            END IF
+            EXIT
+         END IF
```
JWilhelm: please move this print statement into print_evGW0_band_edges

src/gw_non_periodic_ri_rs.F
```diff
@@ -171,63 +174,249 @@ SUBROUTINE gw_calc_non_periodic_ri_rs(qs_env, bs_env)
...
+
+         IF (i_iter == n_iter) EXIT
```
JWilhelm: ->  IF (i_iter == n_iter .OR. converged) EXIT

src/gw_non_periodic_ri_rs.F
```diff
@@ -171,63 +174,249 @@ SUBROUTINE gw_calc_non_periodic_ri_rs(qs_env, bs_env)
...
+         bs_env%eigenval_evGW0(:, :, :) = bs_env%eigenval_G0W0(:, :, :)
+         DO ispin = 1, bs_env%n_spin
+            ! Update all physical states; exclude linear-dependency placeholders.
+            DO i_mo = 1, bs_env%n_mo_retained
+               bs_env%eigenval_scf_Gamma(i_mo, ispin) = bs_env%eigenval_G0W0(i_mo, 1, ispin)
+            END DO
+            bs_env%e_fermi(ispin) = &
+               0.5_dp*(bs_env%eigenval_G0W0(bs_env%n_occ(ispin), 1, ispin) + &
+                       bs_env%eigenval_G0W0(bs_env%n_occ(ispin) + 1, 1, ispin))
+         END DO
```
JWilhelm: please move into a routine update_eigenvalues_G (or similar)

src/gw_non_periodic_ri_rs.F
```diff
@@ -171,63 +174,249 @@ SUBROUTINE gw_calc_non_periodic_ri_rs(qs_env, bs_env)
...
+
+      CALL cp_fm_release(fm_W_time)
+
+      IF (bs_env%ri_rs%do_evgw0) THEN
+         IF (.NOT. converged) THEN
+            CALL cp_warn(__LOCATION__, &
+                         "The evGW0 eigenvalue self-consistency cycle did not converge "// &
+                         "within MAX_ITER cycles. The reported quasiparticle energies are "// &
+                         "those of the last cycle.")
+         END IF
```
JWilhelm: please move into print_evGW0_band_edges, probably with IF-check 

IF(i_iter == n_iter .AND. .NOT. converged)

src/gw_non_periodic_ri_rs.F
```diff
@@ -171,63 +174,249 @@ SUBROUTINE gw_calc_non_periodic_ri_rs(qs_env, bs_env)
...
+         ! put the DFT reference back for the post-GW DOS/band-edge printing
+         bs_env%eigenval_scf_Gamma(:, :) = eigenval_scf_Gamma_dft(:, :)
+         bs_env%e_fermi(:) = e_fermi_dft(:)
+         DEALLOCATE (eigenval_scf_Gamma_dft)
+         ! compute_QP_energies releases Σ^x itself in the G0W0 case
+         CALL cp_fm_release(fm_Sigma_x_Gamma)
```
JWilhelm, line 416: please move into a routine, for example reset_and_clean_bs_env

JWilhelm (changes requested):

src/gw_non_periodic_ri_rs.F
```diff
@@ -277,60 +334,69 @@ SUBROUTINE compute_Sigma_c_and_QP_energies(bs_env, qs_env, fm_W_time, fm_Sigma_x
...
-               WRITE (u, '(T2,A,I4,A)') &
-                  'evGW0 eigenvalue self-consistency reached in', i_iter, ' cycles.'
-               WRITE (u, '(A)') ' '
-            END IF
-            EXIT
-         END IF
+      CALL timeset(routineN, handle)
 
-         IF (i_iter == n_iter) EXIT
+      bs_env%eigenval_evGW0(:, :, :) = bs_env%eigenval_G0W0(:, :, :)
```
JWilhelm: still one confusion: are bs_env%eigenval_G0W0(:, :, :) overwritten during the evGW0 cycle? If yes, is bs_env%eigenval_G0W0(:, :, :) then resettet later to contain G0W0 eigenvalues rather than evGW0 eigenvalues?

src/gw_non_periodic_ri_rs.F
```diff
@@ -171,63 +174,324 @@ SUBROUTINE gw_calc_non_periodic_ri_rs(qs_env, bs_env)
...
+
+      converged = .FALSE.
+
+      CALL init_evGW0(bs_env, n_iter, band_prev, eigenval_scf_Gamma_dft, e_fermi_dft)
+
+      DO i_iter = 1, n_iter
+
+         bs_env%ri_rs%evgw0_i_iter = i_iter
+
+         CALL compute_Sigma_c(bs_env, fm_W_time, bs_env%ri_rs%mat_phi_mu_l, &
```
JWilhelm, line 282: can you please put on top the formulae? 

      !       W_ll'(iτ)   =  Σ_PQ Z_lP W^MIC_PQ(iτ) Z_l'Q
      !        Σ^c_ll'(iτ) = -G^occ_ll'(i|τ|) ∘ W_ll'(iτ),  τ < 0
      !        Σ^c_ll'(iτ) =  G^vir_ll'(i|τ|) ∘ W_ll'(iτ),  τ > 0
      !        Σ^c_λσ(iτ)  =  Σ_ll' Φ_λ(r_l) Σ^c_ll'(iτ) Φ_σ(r_l')

src/gw_non_periodic_ri_rs.F
```diff
@@ -171,63 +174,324 @@ SUBROUTINE gw_calc_non_periodic_ri_rs(qs_env, bs_env)
...
+                              bs_env%ri_rs%mat_Z_lP, fm_Sigma_c_Gamma_time)
+
+         CALL compute_QP_energies(bs_env, qs_env, fm_Sigma_x_Gamma, fm_Sigma_c_Gamma_time)
```
JWilhelm, line 287: can you please put on top the formulae?

      !       Σ^c_λσ(iτ) -> Σ^c_nn(ϵ)
      !        ϵ_n^GW,(i) = ϵ_n^DFT + Σ^c_nn[G^(i-1),W](ϵ_n^GW,(i)) + Σ^x_nn - v^xc_nn

src/gw_non_periodic_ri_rs.F
RitajTyagi: Yes bs_env%eigenval_G0W0(:, :, :) overwritten during the evGW0 cycle. After the loop it holds the last cycle's QP energies, i.e. the evGW0 result. I have not reset back it to G0W0. Also at the ends the labels also changes from G0W0 to evGW0.

src/gw_non_periodic_ri_rs.F
```diff
@@ -171,63 +174,324 @@ SUBROUTINE gw_calc_non_periodic_ri_rs(qs_env, bs_env)
...
+
+         IF (.NOT. bs_env%ri_rs%do_evgw0) EXIT
+
+         CALL print_evGW0_band_edges(bs_env, band_prev, i_iter, n_iter, converged)
+
+         IF (i_iter == n_iter .OR. converged) EXIT
+
+         CALL update_eigenvalues_G(bs_env)
```
JWilhelm, line 298: can you please put formulas on top: 
      ! eigenvalues ϵ_n to be updated in G:
      !      G^occ_µλ(i|τ|)  = Σ_n^occ C_µn e^(-|(ϵ_n-ϵ_F)τ|) C_λn
      !      G^vir_µλ(i|τ|)  = Σ_n^vir C_µn e^(-|(ϵ_n-ϵ_F)τ|) C_λn

JWilhelm (changes requested):

src/gw_large_cell_gamma.F
```diff
@@ -2518,7 +2519,7 @@ SUBROUTINE compute_QP_energies(bs_env, qs_env, fm_Sigma_x_Gamma, fm_Sigma_c_Gamm
 
       ! Σ^x is frozen in an evGW0 cycle, so there the caller keeps it and releases it once the
       ! self-consistency loop is done
-      IF (.NOT. bs_env%ri_rs%do_evgw0) CALL cp_fm_release(fm_Sigma_x_Gamma)
+      IF (bs_env%gw_flavour /= evGW0) CALL cp_fm_release(fm_Sigma_x_Gamma)
```
JWilhelm: IF (bs_env%gw_flavour == G0W0) 

will be safer if you will have more GW flavors later

src/gw_non_periodic_ri_rs.F
```diff
@@ -277,7 +278,7 @@ SUBROUTINE compute_Sigma_c_and_QP_energies(bs_env, qs_env, fm_W_time, fm_Sigma_x
 
          CALL compute_QP_energies(bs_env, qs_env, fm_Sigma_x_Gamma, fm_Sigma_c_Gamma_time)
 
-         IF (.NOT. bs_env%ri_rs%do_evgw0) EXIT
+         IF (bs_env%gw_flavour /= evGW0) EXIT
```
JWilhelm: IF (bs_env%gw_flavour == G0W0) 

will be safer if you will have more GW flavors later

src/gw_non_periodic_ri_rs.F
```diff
@@ -396,12 +395,14 @@ SUBROUTINE reset_and_clean_bs_env(bs_env, eigenval_scf_Gamma_dft, e_fermi_dft, f
...
-      IF (bs_env%ri_rs%do_evgw0) THEN
+      IF (bs_env%gw_flavour == evGW0) THEN
+         ! Update array with final evGW0 result
+         bs_env%eigenval_evGW0(:, :, :) = bs_env%eigenval_GW(:, :, :)
          ! put the DFT reference back for the post-GW DOS/band-edge printing
          bs_env%eigenval_scf_Gamma(:, :) = eigenval_scf_Gamma_dft(:, :)
          bs_env%e_fermi(:) = e_fermi_dft(:)
          DEALLOCATE (eigenval_scf_Gamma_dft)
-         ! compute_QP_energies releases Σ^x itself in the G0W0 case
+         ! Releases Σ^x in the G0W0 case
```
JWilhelm: ! Releases Σ^x in the evGW0 case
         
         
or entirely remove the comment

JWilhelm (changes requested):

JWilhelm (approved):

PR #5782, opened 2026-08-19
Reduce device contention for P100 tests (see #5763)

mkrack:


DCM-Uni-Paderborn:
Thanks for looking into this. The live run suggests that `--maxtasks 8` is too restrictive for this runner: it reduces the job from six to two directory workers, and after about 2.5 hours it had completed only 148 of 489 directories.

It also does not resolve the main `psmp` failure. All nine pseudopotential SIRIUS tests still time out at 400 s and the SIRIUS batch alone takes 3652 s. The same CUDA-enabled SIRIUS 7.11.1 tests finish in 99 s in the toolchain Pascal job, while the CPU Spack job finishes them in 49 s. The relevant configuration difference appears to be the Spack `SIRIUS +memory_pool`/Umpire build; the working toolchain explicitly sets `SIRIUS_USE_MEMORY_POOL=OFF`.

I therefore suggest retaining the existing worker count, increasing the per-test timeout, and testing `SIRIUS ~memory_pool` for the Spack P100 build. This keeps CUDA-enabled SIRIUS and all regtests enabled while addressing the configuration difference instead of making the complete dashboard run several hours long.

mkrack:
Closing this PR in favour of #5792.

PR #5785, opened 2026-08-19
Packit: Use label requirement for copr_build job in PR

Growl1234:


Growl1234:
@LecrisUT Did I do it right? It seems adding label itself doesn't trigger Packit job; if I understand correctly, Packit detect the label only when a commit is pushed or PR is opened, so one might still need `/packit build` if PR is opened before labelled?

LecrisUT:
> Packit detect the label only when a commit is pushed or PR is opened, so one might still need `/packit build` if PR is opened before labelled?

Yes, that might be the case. Don't quite remember the process on that. I've opened an [issue](https://github.com/packit/packit-service/issues/3166) just in case, and I might take that one soon.

You should also add it to the testing-farm job. Don't think the separation of those jobs makes sense now either. You can use `_` key to write the yaml anchors and keep things a bit tidy.

Growl1234:
I'm less sure about putting these PR jobs behind the same label, as a Copr build already spends more than 50 minutes.

oschuett (approved):

Issue #4849, opened 2026-02-17
Failing CUDA P100 tests (ssmp and psmp)


mkrack:
The new spack based CUDA P100 tester [ssmp](https://storage.googleapis.com/cp2k-ci/run-cp2k-spack-ssmp-p100-d43c5fce_report.txt) and [psmp](https://storage.googleapis.com/cp2k-ci/run-cp2k-spack-psmp-p100-8c4fa722_report.txt) show some runtime failures (segfaults) mostly in `src/pw/fft/fftw3_lib.F:1165`

```Backtrace for this error:
#0  0x1238e94c8e59 in ???
#1  0x1238e94c7e75 in ???
#2  0x1238e92d832f in ???
#3  0x1238d7584f26 in ???
#4  0x1238ec7164a3 in __fftw3_lib_MOD_fftw31dm._omp_fn.0
	at /opt/cp2k/src/pw/fft/fftw3_lib.F:1165
#5  0x1238e67a2976 in ???
#6  0x1238ec716dd7 in __fftw3_lib_MOD_fftw31dm
	at /opt/cp2k/src/pw/fft/fftw3_lib.F:1135
#7  0x1238ec715baf in __fft_lib_MOD_fft_1dm
	at /opt/cp2k/src/pw/fft/fft_lib.F:274
```
Reducing the number of OpenMP threads or the parallelism of the test run does not help.

PR #5792, opened 2026-08-19
Stabilize remaining Pascal dashboard runs

DCM-Uni-Paderborn:
## Summary

- raise the per-test timeout from 150 to 400 seconds for the legacy CUDA Pascal and both Spack `ssmp` and `psmp` P100 jobs
- build parallel Spack configurations with `SIRIUS ~memory_pool`, matching the working toolchain SIRIUS configuration
- keep the existing worker counts, all dependencies, all test directories, and all numerical tolerances unchanged

## Motivation

The CUDA Pascal and Spack P100 reports contain no wrong numerical results. Their failures are tests reaching the common 150-second timeout, plus restart or dependent tests that fail because a predecessor timed out.

The Spack `psmp` P100 report originally exposed an additional SIRIUS configuration difference. Its CUDA-enabled SIRIUS 7.11.1 build used the default `+memory_pool` variant and Umpire, whereas the working CUDA toolchain build explicitly uses `SIRIUS_USE_MEMORY_POOL=OFF`. Disabling the optional SIRIUS memory pool retains CUDA-enabled SIRIUS and its complete regtest coverage.

A follow-up full-suite run confirmed that `~memory_pool` was concretized and used. It had no wrong numerical results, but 15 tests still reached the 150-second limit and two dependent/restart tests consequently failed. Applying the same 400-second limit to the Spack `psmp` P100 job addresses that remaining inconsistency without reducing overall throughput.

This complements #5782. Its live report shows that `--maxtasks 8` lowers the `psmp` run from six to two directory workers and makes the complete suite several hours long, so that concurrency change is not carried over here.

## Testing

- `./make_pretty.sh`
- `python3 -m py_compile tools/docker/generate_dockerfiles.py`
- `git diff --check`
- CP2K pre-push hook for the files changed in both follow-up commits

Fixes part of #4849.

### DIFF
--- a/tools/docker/generate_dockerfiles.py
+++ b/tools/docker/generate_dockerfiles.py
@@ -205,7 +205,7 @@ def main() -> None:
                 base_image="docker.io/nvidia/cuda:12.9.1-devel-ubuntu24.04",
                 gcc_version=13,
                 gpu_model="P100",
-                testopts=testopts,
+                testopts=f"{testopts} --timeout 400",
                 image_tag=f.image_tag,
             )
         )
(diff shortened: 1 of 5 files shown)


tools/docker/generate_dockerfiles.py
mkrack, line 236: This tester might need the same increase of `timeout`.

mkrack:
The SIRIUS tests require most likely even more than 400s with pdbg. That issue and my comment can be addressed in a follow-up PR.

mkrack (approved):

PR #5795, opened 2026-08-20
Restructure installation documentations, create `./docs/getting-started/installation.md` as an umbrella entry

he-zilong:
I have created `./docs/getting-started/installation.md` by rewriting the `INSTALL.md` (which is now a soft link) and incorporating some material shown under `show_help_hpc()` in `install_cp2k_toolchain.sh` (which is now removed as it is not really limited to toolchain).

The new `installation.md` consists of three parts:
- Some generic considerations before installation;
- Simplified installation instructions in 3 steps;
- Further explanations and alternative methods given in the links to the existing `build-from-source.md`, `build-with-spack.md`, and `distributions.md` under `./docs/getting-started`.

I also take this chance to write about the developer teams in `./docs/development/onboarding.md`.


docs/getting-started/installation.md
Growl1234, line 1: There's quite a lot duplication with `build-from-source.md`, and installing from package managers usually doesn't get through them. I'd suggest expanding `build-from-source.md` instead of creating a new `installation.md`.

INSTALL.md
Growl1234, line 1: And also, this could be directed to `./docs/getting-started/build-from-source.md` then.

docs/development/onboarding.md
```diff
@@ -91,3 +92,19 @@ usually takes less than one hour:
...
+- The [`cp2k-developers`](https://github.com/orgs/cp2k/teams/cp2k-developers) is the beginner level
+  with *triage* access, which can be joined by accepting an invitation automatically sent out after
+  the very first accepted and merged PR.
+- The [`cp2k-core-developers`](https://github.com/orgs/cp2k/teams/cp2k-core-developers) is the next
+  level for productive contributors, to whom *write* access is granted on a case-by-case basis. And
+  among other powers available, they are eligible for volunteering as a `code owner` of certain code
+  ([see github docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)),
+  whose review and explicit approval become necessary for any relevant PR from other developers.
+- The [`cp2k-admins`](https://github.com/orgs/cp2k/teams/cp2k-admins) is the ultimate level for
+  administrators in charge of everything.
```
Growl1234: Non-blocking, but I'm not sure if this is desired by maintainers; in my view pointing out the automatic invitation is enough, as the inclusion of other teams with higher access are decided by maintainers manually.

docs/getting-started/installation.md
```diff
@@ -0,0 +1,185 @@
...
+implied for most of the documentation.
+
+1. Which flavor of operating system is the target machine?
+
+- **Linux** (assumed): the workhorse for high-performance computing, a majority of compiling and
+  regression tests are hosted in some Linux distribution.
+- **Windows**: for lightweight beginner experiments, the Windows Subsystem for Linux (WSL) or any
+  other virtual machine platforms can provide a Linux environment on Windows.
+
+1. Does the user have root privilege or sudo power on the machine?
```
fstein93, line 21: I guess you wanted to enumerate the different aspects instead of giving every aspect the number 1

fstein93 (commented):
I agree with @Growl1234 to be careful with doubling information.

docs/getting-started/installation.md
Growl1234: See https://github.com/cp2k/cp2k/pull/5481#discussion_r3490025608; I guess there's no worry.

he-zilong:
Re duplication: I have already noticed how the old `INSTALL.md` overlapped with `./docs/getting-started/build-from-source.md` and thought about removing `INSTALL.md` altogether.

What if I remove the instructions part in `installation.md` and preserve the considerations (too important to be ignored) and the links (in the form of a `{toctree}`, which also collapses three entries on the side bar of the documentation page into one)?

docs/getting-started/installation.md
he-zilong: I wouldn't say I like this unintuitive `make_pretty` quirk actually...

docs/getting-started/installation.md
oschuett: I'd also prefer to keep the current structure of the getting started section. So instead of creating a new page, maybe you could add the missing bits from `INSTALL.md` to the existing pages.

We could then point `INSTALL.md` at `build-from-source.md` or just remove it.

docs/getting-started/installation.md
oschuett: It's meant to reduce diff-noise: https://mdformat.readthedocs.io/en/stable/users/style.html#ordered-lists

We could disable it by passing `--number` to Mdformat:
https://github.com/cp2k/cp2k/blob/13a7ada6f65dadad33c040bd81ad8a98d102905d/tools/precommit/precommit_server.py#L55

docs/development/onboarding.md
he-zilong: This is mainly for pointing out the possibility of becoming one of the code owners (which requires write access) and introducing the admins. Although newcomers would eventually figure this out, I think I can just save a bit of their time to state it clearly. The idea is similar to the `GOVERNANCE.md` as one of the [github community health files](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file).

docs/getting-started/installation.md
Growl1234: From the result it doesn't work. I guess because `1. AAA` is used together with `- aaa`.

<details>

<summary>Snapshot</summary>

图片

</details>

docs/development/onboarding.md
oschuett: I agree, that it's good to document our governance structure and `development/onboarding.md` is certainly the right place. I'm just a little worried that it might intimidate newcomers. So, maybe we can rephrase it a bit to put less emphasis on the hierarchy.

docs/getting-started/installation.md
oschuett: Yes, nested blocks have to be intended.  And you also have to be mindful about blank lines.

docs/getting-started/installation.md
```diff
@@ -0,0 +1,185 @@
...
+
+Here is a checklist for crucial premises and factors. The label "(assumed)" marks the conditions
+implied for most of the documentation.
+
+1. Which flavor of operating system is the target machine?
+
+- **Linux** (assumed): the workhorse for high-performance computing, a majority of compiling and
+  regression tests are hosted in some Linux distribution.
+- **Windows**: for lightweight beginner experiments, the Windows Subsystem for Linux (WSL) or any
+  other virtual machine platforms can provide a Linux environment on Windows.
```
Growl1234: Try this for all similar layout you introduced here:

```suggestion
1. Which flavor of operating system is the target machine?

   - **Linux** (assumed): the workhorse for high-performance computing, a majority of compiling and
     regression tests are hosted in some Linux distribution.
   - **Windows**: for lightweight beginner experiments, the Windows Subsystem for Linux (WSL) or any
     other virtual machine platforms can provide a Linux environment on Windows.
```

docs/development/onboarding.md
Growl1234: > I'm just a little worried that it might intimidate newcomers. So, maybe we can rephrase it a bit to put less emphasis on the hierarchy.

Agreed. In particular, I don't think newcomers will feel very happy when greeted with terms such as “beginner level”, “next level”, or “ultimate level”.

I suggest describing the teams in terms of how repository permissions and maintenance responsibilities are currently distributed, without presenting them as “ranks” that contributors are expected to progress through. The wording should remain as objective and neutral as possible, simply explaining the purpose and permissions of each team and how the corresponding responsibilities are handled in practice.

he-zilong:
I have removed the duplicated instructions in `installation.md`. It is now served as the upper-level landing page in the manual like so:

getting-started

The content about team membership has also been revised.

mkrack:
I am always puzzled by the headings `Build from Source` and `Build with Spack`. The latter does describe also a build from source, doesn't it?

oschuett:
> I am always puzzled by the headings `Build from Source` and `Build with Spack`. The latter does describe also a build from source, doesn't it?

Fair point, especially since we now also have [Spack-based build via make_cp2k.sh](https://manual.cp2k.org/trunk/getting-started/build-from-source.html#spack-based-build-via-make-cp2k-sh) within the _"Build from Source"_ page.

Maybe the content from [Build with Spack](https://manual.cp2k.org/trunk/getting-started/build-with-spack.html) should be move into [Install from Distribution](https://manual.cp2k.org/trunk/getting-started/distributions.html), like you already did for Conda.

he-zilong:
> Maybe the content from [Build with Spack](https://manual.cp2k.org/trunk/getting-started/build-with-spack.html) should be move into [Install from Distribution](https://manual.cp2k.org/trunk/getting-started/distributions.html), like you already did for Conda.

I thought about this for a while, but decided not to touch these two pages in this PR as the former seems to me a bit too long to be a subsection of the latter.

Growl1234 (approved):

README.md
```diff
@@ -20,32 +20,31 @@ OpenCL).
...
-To clone the current master (development version):
+For end users, the preferred method is to go to [Releases](https://github.com/cp2k/cp2k/releases/)
+and download and extract one of the versioned tarballs `cp2k-X.Y.tar.bz2`.
+
+For developers, the preferred method is to fork the [`cp2k/cp2k`](https://github.com/cp2k/cp2k)
+repository and download via `git clone`. To clone the current master (development version):
 
 ```shell
-git clone https://github.com/cp2k/cp2k.git cp2k
+git clone --recursive https://github.com/cp2k/cp2k.git cp2k
```
Growl1234: The `--recursive` is in fact no longer needed as CP2K no longer has git submodules. It's only relevant for release branch <=v2025.2.

```suggestion
git clone https://github.com/cp2k/cp2k.git cp2k
```

README.md
```diff
@@ -20,32 +20,31 @@ OpenCL).
...
 ```
 
 To clone a release version v*x.y*:
 
 ```shell
-git clone -b support/vx.y --recursive https://github.com/cp2k/cp2k.git cp2k
+git clone --recursive -b support/vx.y --recursive https://github.com/cp2k/cp2k.git cp2k
```
Growl1234: Duplicated.
```suggestion
git clone -b support/vx.y --recursive https://github.com/cp2k/cp2k.git cp2k
```

oschuett:
> > Maybe the content from [Build with Spack](https://manual.cp2k.org/trunk/getting-started/build-with-spack.html) should be move into [Install from Distribution](https://manual.cp2k.org/trunk/getting-started/distributions.html), like you already did for Conda.

> I thought about this for a while, but decided not to touch these two pages in this PR as the former seems to me a bit too long to be a subsection of the latter.

Yeah, you're right. We could try to shorten it, e.g. by removing the [Developer Workflow](https://manual.cp2k.org/trunk/getting-started/build-with-spack.html#developer-workflow) section. Alternatively, we could created dedicated sub-pages for some distributions.  In any case, let's leave that for a separate PR.

Growl1234:
How about making "Install from Distribution" an umbrella section that contains "Spack", "Conda-forge", and "System package managers". Personally I hope the [Developer Workflow](https://manual.cp2k.org/trunk/getting-started/build-with-spack.html#developer-workflow) section not to be removed.

he-zilong:
Some other questions:
- I have listed Linux and Windows as two flavours of operating systems, but now I think it needs to mention other alternatives like MacOS and Cray. What do you think I should put there?
- Would it be a good idea to make "Deploy CP2K Docker containers" a standalone page independent from "Build with Spack"?

oschuett:
> How about making "Install from Distribution" an umbrella section that contains "Spack", "Conda-forge", and "System package managers".

Yes, exactly! We could also do some spring cleaning: NixOS, Easybuild, and NGC are apparently no longer maintained.

> Personally I hope the [Developer Workflow](https://manual.cp2k.org/trunk/getting-started/build-with-spack.html#developer-workflow) section not to be removed.

My impression was that the section got  superseded by `make_cp2k.sh` and for the CI containers we still got [this](https://github.com/cp2k/cp2k/blob/master/tools/docker/README.md), but I'm happy to keep it too.

> I have listed Linux and Windows as two flavours of operating systems, but now I think it needs to mention other alternatives like MacOS and Cray. What do you think I should put there?

In my observation most scientists use macOS on their laptop these days - so it's definitely important

For Windows we started recommending [WSL](https://www.cp2k.org/howto:compile_on_windows) over [cygwin](https://www.cp2k.org/howto:compile_on_windows_with_cygwin) a while ago, but it seems to be in decline anyway.


> Would it be a good idea to make "Deploy CP2K Docker containers" a standalone page independent from "Build with Spack"?

For production containers we have a separate repo with [its own docs](https://github.com/cp2k/cp2k-containers/blob/master/README.md) because it's conceptually a downstream project.

he-zilong:
> > I have listed Linux and Windows as two flavours of operating systems, but now I think it needs to mention other alternatives like MacOS and Cray. What do you think I should put there?
> 
> In my observation most scientists use macOS on their laptop these days - so it's definitely important
> 
> For Windows we started recommending [WSL](https://www.cp2k.org/howto:compile_on_windows) over [cygwin](https://www.cp2k.org/howto:compile_on_windows_with_cygwin) a while ago, but it seems to be in decline anyway.

For MacOS, I am not sure what particularities to mention there as all I could find is an old cp2k.org page [howto:compile_on_macos](https://www.cp2k.org/howto:compile_on_macos).

For Windows, I beg to differ: it is still where most non-computational academic works are routinely carried out, and WSL on Windows is already sufficient for a beginner who may not be Linux-savvy but still wants to run small experimental inputs and experiment with input syntax of just any quantum chemistry programs, including CP2K. Case in point: the ABACUS program is even [providing a set of WSL-based installation tools](https://github.com/deepmodeling/abacus-develop/tree/develop/tools/04_windows_installation).

README.md
he-zilong: The command in `build-from-source.md` that you added 2 months ago has `--recursive`：
https://github.com/cp2k/cp2k/blob/ae513bac4f751e79b120bf8419509e980ad75ec8/docs/getting-started/build-from-source.md?plain=1#L33

For consistence, maybe I should also remove this.

Growl1234: Yeah, please remove it.

Growl1234:
> For MacOS, I am not sure what particularities to mention there as all I could find is an old cp2k.org page [howto:compile_on_macos](https://www.cp2k.org/howto:compile_on_macos).

macOS installation can be done (at least) via brew or Spack. There's already a [Spack tester for macOS](https://dashboard.cp2k.org/archive/darwin-gnu-arm64-psmp/index.html) in Dashboard. So generally I don't think documenting it would be an issue

As of Windows, I would say let's keep the current description: via WSL2 or virtual machine platforms.

Growl1234:
> Case in point: the ABACUS program is even [providing a set of WSL-based installation tools](https://github.com/deepmodeling/abacus-develop/tree/develop/tools/04_windows_installation).

Personally I'm not sure if this is the best practice, but if maintainers are interested, we could also consider having one, which can be done with conda-forge as well (Spack can also be considered but might be a bit heavy for common PC). One thing to keep in mind, however, is that we should make it as clean as possible.

mkrack:
I tried last weekend a CP2K v2026.2 installation on my Win11 PC at home. It works but the resources of my PC (8 cores and 16 GB RAM) are quite limiting. After `wsl --install` in an Admin-PowerShell, the installation works (except for one minor issue) like under any Ubuntu system, after disabling some features because of the limited resources. I tried the `make_cp2k.sh` install path. I was already about to update the old webpage, but I am not sure if it is really worth it.

mkrack:
> My impression was that the section got superseded by `make_cp2k.sh` and for the CI containers we still got [this](https://github.com/cp2k/cp2k/blob/master/tools/docker/README.md), but I'm happy to keep it too.

Indeed, `make_cp2k.sh` is supposed to be a convenience tool for CP2K development in a first place.
For people who want to have just a quick glimpse on CP2K, `conda-forge` or the [static binary](https://github.com/cp2k/cp2k/releases/tag/v2026.2) are the best options in my opinion.

mkrack:
The repo [cp2k-containers](https://github.com/cp2k/cp2k-containers) has become obsolete starting with CP2K v2026.2 as indicated by the recently added note.

he-zilong:
I am satisfied with the current look of the manual, even though not all refactoring or organization suggested above are made.

The part about testing an input file under the `Conda` section has been removed as the topic has been covered by `first-calculation.md` and is not conda-exclusive. Meanwhile, I am leaving the list of distributions as-is without judging the support status.

Some interesting ideas may be better suited for further discussions and follow-up PRs.

oschuett (approved):

PR #5800, opened 2026-08-20
Limit P100 psmp regtest concurrency

DCM-Uni-Paderborn:
The targeted Spack psmp P100 run for #5792 completed 5941/5952 tests. Its remaining failures were one post-run segmentation fault and ten SIRIUS tests that timed out together, while the same SIRIUS input succeeds both classically and through the CP2K shell on an independent SIRIUS-enabled build. This points to node-level resource or process-teardown contention rather than incorrect test results.

Reduce the P100 psmp runner from six to four concurrent regtest workers by setting `--maxtasks=16` for its 2-MPI-rank, 2-OpenMP-thread jobs. All tests and dependencies remain enabled, and the existing per-test timeout stays unchanged.

Checks performed:
- `./make_pretty.sh`
- `git diff --check`

A targeted `spack-psmp-p100` run is needed to confirm the dashboard behavior.

### DIFF
--- a/tools/docker/Dockerfile.test_spack_psmp-P100
+++ b/tools/docker/Dockerfile.test_spack_psmp-P100
@@ -106,7 +106,7 @@ COPY --from=build_cp2k /etc/ld.so.conf.d/cp2k.conf /etc/ld.so.conf.d/cp2k.conf
 RUN ldconfig
 
 # Run CP2K regression test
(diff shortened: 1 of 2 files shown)


DCM-Uni-Paderborn:
The reduced-concurrency validation is still running, but the trade-off is already problematic: after roughly 95 minutes it had completed only 155 of 490 test directories, while the CI pod has a hard three-hour runtime limit. The previous six-worker run already took 140 minutes, so reducing it to four workers is unlikely to be a viable permanent fix. I am moving this PR back to draft and will treat the run as diagnostic evidence rather than proposing the limit for merge.

PR #5801, opened 2026-08-20
Limit OpenMPI debug regtest concurrency

DCM-Uni-Paderborn:
The Spack OpenMPI pdbg dashboard has failed on different tests in successive runs: the archived run timed out in `QS/regtest-acc-2/HF-ec2.inp` at 150 seconds, while the targeted run for #5787 timed out in `QS/regtest-stda-force-3/h2o_f20.inp` and `xTB/regtest-gfn1-d/ch2o_t08.inp` despite a 400-second limit. No incorrect numerical result was reported.

The runner currently starts 16 concurrent two-rank debug jobs on 32 CPUs. Cap it at eight concurrent jobs with `--maxtasks=16`. This keeps every test enabled and preserves the 400-second per-test timeout while reducing memory and process contention in the leak-sanitized debug build.

Checks performed:
- `./make_pretty.sh`
- `git diff --check`

A targeted `spack-openmpi-pdbg` run is needed to validate the change.

### DIFF
--- a/tools/docker/Dockerfile.test_spack_openmpi-pdbg
+++ b/tools/docker/Dockerfile.test_spack_openmpi-pdbg
@@ -92,7 +92,7 @@ COPY --from=build_cp2k /etc/ld.so.conf.d/cp2k.conf /etc/ld.so.conf.d/cp2k.conf
 RUN ldconfig
 
 # Run CP2K regression test
-RUN /opt/cp2k/install/bin/launch /opt/cp2k/install/bin/run_tests --timeout 400 || echo "ERROR: Regression test run failed"
+RUN /opt/cp2k/install/bin/launch /opt/cp2k/install/bin/run_tests --timeout 400 --maxtasks=16 || echo "ERROR: Regression test run failed"
 
 # Create entrypoint and finalise container build
 WORKDIR /mnt
--- a/tools/docker/generate_dockerfiles.py
+++ b/tools/docker/generate_dockerfiles.py
@@ -158,7 +158,7 @@ def main() -> None:
                 version="pdbg",
                 mpi_mode="openmpi",
                 feature_flags="",
-                testopts="--timeout 400",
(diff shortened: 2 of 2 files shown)


Growl1234:
I believe, instead of simply reducing the concurrency, we should try to reproduce the problem and find out the *real* issue.

DCM-Uni-Paderborn:
Agreed. The lower-concurrency run shows that the failure is concurrency-sensitive, but it does not identify the underlying cause. The tests that previously timed out after 400 seconds completed in 2.6 to 7.5 seconds in the successful run, so shared-memory, MPI-resource, or cross-job isolation effects should be reproduced at the original concurrency before proposing a permanent change. I am moving this PR back to draft while investigating that.

DCM-Uni-Paderborn:
I compared the full- and reduced-concurrency reports in more detail. With 16 workers, instances of the same test names first complete normally, while two later launches produce no CP2K output and remain stuck until the exact 400 s timeout. The 99th percentile of all other tests stays at 16.4 s. With 8 workers, the affected instances complete in 2.63--7.52 s. This narrows the symptom to sporadic process-launch or shared-resource contention rather than intrinsically slow regtests, but it still does not identify whether the underlying issue is in OpenMPI/PMIx, the GCP runner, or the concurrent-launch behavior of the test driver. I am therefore keeping this PR as a draft and am not proposing the concurrency reduction as the final fix.

DCM-Uni-Paderborn:
Closing this diagnostic PR because reducing the regtest concurrency only avoids the symptom and does not identify the underlying process-launch failure. The comparison above is preserved for a focused cp2k-ci/OpenMPI/GCP investigation if the issue recurs.

oschuett:
Note that we currently can't clearly detect when a container on GCP runs out of memory because we're using Kubernetes Jobs that just restart the Pods without preserving their exit code. As part of my refactoring, I want to switch to scheduling Pods directly.

PR #5802, opened 2026-08-20
Limit ASan regtest concurrency

DCM-Uni-Paderborn:
The latest Address Sanitizer dashboard report stops abruptly partway through the regression suite. It contains no sanitizer finding, test failure, summary, final status, or `EndDate`, which is consistent with the worker being terminated externally under load. The previous complete ASan run used eight concurrent 2-rank x 2-thread jobs on a 32-CPU node.

Set `--maxtasks=24` for this runner, reducing it to six concurrent jobs. This is a 25% reduction in process and memory pressure; all tests, sanitizer checks, and dependencies remain enabled. The last complete run took about 80 minutes, so the reduced concurrency should remain within the dashboard timeout.

Checks performed:
- `./make_pretty.sh`
- `git diff --check`

A targeted `asan-psmp` run is required. If it still ends without a report footer or sanitizer finding, the remaining problem is likely in the runner or GCP pool rather than CP2K.

### DIFF
--- a/tools/docker/Dockerfile.test_asan-psmp
+++ b/tools/docker/Dockerfile.test_asan-psmp
@@ -72,7 +72,7 @@ COPY ./tools/docker/scripts/build_cp2k.sh ./tools/docker/scripts/cmake_cp2k.sh .
 RUN ./build_cp2k.sh toolchain_asan psmp
 
 # Run regression tests.
-ARG TESTOPTS=""
+ARG TESTOPTS="--maxtasks=24"
 COPY ./tests ./tests
 COPY ./tools/docker/scripts/test_regtest.sh ./
 RUN /bin/bash -o pipefail -c " \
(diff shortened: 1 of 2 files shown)


Growl1234:
Honestly, I don't see any effect of limiting the number of cores other than slowing down the regtesting workflow. I think the dashboard testers were simply preempted because of high demand, so why not spread the schedules out as much as possible instead of imposing this kind of restriction? I don't think it reduces the likelihood of this happening to any meaningful extent.

@oschuett what do you think?

DCM-Uni-Paderborn:
The concern is valid. Limiting the regtest workers does not reduce the requested node size or the risk of preemption on the spot-instance pool, and it makes the ASan run longer. The successful 99-minute run therefore does not establish that concurrency caused the previous truncated report. I am closing this PR; scheduling or runner handling should be addressed in cp2k-ci instead.

oschuett:
Yes, the ASan run really takes too long for preemptable VMs. This should get resolved with the CI-rewrite I'm working on.

PR #5804, opened 2026-08-21
Parallelize k-point calculations with OpenMP for single-rank runs

Growl1234:
This PR adds an OpenMP-parallel path for Quickstep k-point SCF calculations when CP2K is run with a single MPI rank.

Previously, the k-points handled by `do_general_diag_kp` and the subsequent density-matrix construction were processed mostly sequentially, even when multiple OpenMP threads were available. This left most CPU cores idle during a significant part of k-point calculations.

The new path parallelizes the independent `(k-point, spin)` work across OpenMP threads, including:

- the real-space to k-point transformation of the Hamiltonian and overlap matrices;
- local generalized diagonalization;
- k-point DIIS error construction and the diagonalization after mixing;
- construction of the k-point density matrices;
- the k-point to real-space density transformation; and
- symmetry transformations of the density matrices.

Each worker uses private dense matrices and scratch buffers. Immutable traversal plans are constructed before entering the parallel regions, and the real-space density transformation assigns each output block to a single thread to avoid data races and false sharing.

Hybrid MPI+OpenMP optimzation cannot work because of [`MPI_THREAD_SERIALIZED`](https://github.com/cp2k/cp2k/blob/2823083452d26bf0e93751c57ae6cbdd3f8ace63/src/mpiwrap/message_passing.F#L1139), so OpenMP workers do not enter MPI or ScaLAPACK. For diagonalization, the single-rank path uses local LAPACK and BLAS routines instead of invoking ScaLAPACK concurrently.

### DIFF
--- a/src/fm/cp_cfm_diag.F
+++ b/src/fm/cp_cfm_diag.F
@@ -57,7 +57,8 @@ MODULE cp_cfm_diag
 
    CHARACTER(len=*), PARAMETER, PRIVATE :: moduleN = 'cp_cfm_diag'
 
-   PUBLIC :: cp_cfm_heevd, cp_cfm_geeig, cp_cfm_geeig_canon
+   PUBLIC :: cp_cfm_heevd, cp_cfm_geeig, cp_cfm_geeig_canon, &
+             cp_cfm_geeig_local, cp_cfm_geeig_canon_local
 
 CONTAINS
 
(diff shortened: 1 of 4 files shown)


Growl1234:
This is still WIP as I'm trying some code cleanups, but from my testing through some simple tasks its effiency seemed to meet my expectation :)

fstein93:
I think it is a good idea to either document where `MPI_THREAD_MULTIPLE` may help (maybe after some rewriting) or to implement this code but checking for the thread-level (requires the addition of a suitable wrapper of `MPI_Query_thread` in `message_passing.F`. The latter may be reasonable because we require support for `MPI_THREAD_MULTIPLE` if CP2K was compiled with support for OpenPMD or DLAF and in the long-run, we wanted to upgrade to `MPI_THREAD_MULTIPLE` in general.

Growl1234:
I'd also be in favor of upgrading to `MPI_THREAD_MULTIPLE` in general. And I'd believe this is an essential step towards improving the overall OpenMP efficiency.

oschuett:
> I'd also be in favor of upgrading to MPI_THREAD_MULTIPLE in general. 

Me too! So, should we just do it (in a separate PR)?

Growl1234:
> > I'd also be in favor of upgrading to MPI_THREAD_MULTIPLE in general.
> 
> Me too! So, should we just do it (in a separate PR)?

I think so, and I'd like to take this job.

Growl1234:
An interesting performance result was observed in a test case: [ZnO-kp-test.zip](https://github.com/user-attachments/files/31299651/ZnO-kp-test.zip)

The ZnO test uses a 15\*15\*9 Monkhorst–Pack grid, reduced to 1013 irreducible k-points, with 76 orbital basis functions. The measured timings are:

| MPI ranks × OMP threads | Total threads | SCF time per iteration | `do_general_diag_kp` per iteration |
| ----------------------: | ------------: | ---------------------: | ---------------------------------: |
|                   1 × 4 |             4 |                 1.12 s |                             0.78 s |
|                   2 × 2 |             4 |                 4.67 s |                             4.30 s |
|                   1 × 8 |             8 |                 0.64 s |                             0.44 s |
|                   4 × 2 |             8 |                 3.63 s |                             3.43 s |
|                   8 × 1 |             8 |                 4.63 s |                             4.26 s |
|                  1 × 16 |            16 |                 0.44 s |                             0.30 s |
|                  1 × 24 |            24 |                 0.39 s |                             0.26 s |

At the same total concurrency, the single-rank OpenMP path is therefore approximately 4–7 times faster than the current multi-rank path. Almost all of this difference occurs in `do_general_diag_kp`.

It exposes an unfavorable decomposition for this particular workload. The number of irreducible k-points, 1013, is prime, while the current implementation requires the number of equally sized k-point groups to divide the total number of k-points. With the tested MPI process counts, only one k-point group can be formed. All MPI ranks therefore cooperate on every small (76\*76) eigenproblem instead of processing different k-points independently.

For matrices of this size, full-matrix redistribution, MPI collectives, and distributed eigensolver synchronization cost substantially more than the local numerical work. The new single-rank OpenMP path instead assigns independent k-points directly to threads and uses local BLAS/LAPACK routines, avoiding most of this overhead.

I guess this leads to another discussion about the overall performance concern regarding k-point calculations... (Maybe one proposal for this question is to allow uneven distribution of k-points between MPI groups.)

Growl1234:
> The number of irreducible k-points, 1013, is prime, while the current implementation requires the number of equally sized k-point groups to divide the total number of k-points. With the tested MPI process counts, only one k-point group can be formed

Further investigation has shown this is most likely even another blocker for optimization of hybrid (MPI+OpenMP) optimization. Therefore, although #5811 and #5817 have prepared the `MPI_THREAD_MULTIPLE`, I currently cannot be convinced that we can bring significant improvement also to hybrid paralleism through a clean path.

So generally, I would consider this PR as ready. I'd leave more work mentioned above as follow-up PRs.

fstein93 (approved):

PR #5734, opened 2026-08-13
Remove duplicate AO momentum integral APIs

RTChou67:
Remove two redundant AO momentum integral APIs.
## Summary
  - Deleted the duplicate `qs_operators_ao::ang_mom` API and replaced it with the existing `ai_angmom::angmom` API.
  - Deleted the duplicate `qs_operators_ao::p_xyz_ao` API and replaced it with `qs_operators_ao::build_lin_mom_matrix`.
  - Extended `build_lin_mom_matrix` with the optional `minimum_image` argument.

Issue #4694, opened 2026-01-13
Features with no code coverage


fstein93:
There is a variety of features without any coverage by the regression tests (either files or functionalities from the input file). Have they ever been practically used in the past or do you know whether they are still functionable? From the older features, it is (incomplete list)
- [x] RIGPW
- [ ] ET_COUPLING%TYPE_OF_CONSTRAINT DDAPC
- [ ] src/ct_methods.F (Cayley transforms)
- [ ] src/kpsym.F (Tools for k-point symmetry)
- [ ] src/minbas_methods.F
- [ ] NEGF%INTEGRATION_METHOD CLENSHAW-CURTIS, NEGF
- [ ] EMBED%EMBED_METHOD DMFET
- [x] PEXSI
- [ ] DFT%SCF%OUTER_SCF%TYPE [S2_CONSTRAINT|BASIS_CENTER_OPT]
- [ ] DF%PRINT%AO_MATRIX%COMMUTATOR_HR
- [ ] VCD_env_type%do_mpf (not set from input anywhere)
- [ ] DFT%PRINT%WANNIER90
- [ ] se_fock_matrix_dbg.F (never called by caller)
- [ ] QS%SE%PERIODIC EWALD_R3 (there is only an untested regtest)
- [ ] src/motion/free_energy_methods.F: ui_check_convergence (test to call it never passes)
- [ ] QS/regtest-tipscan (was not registered)
- [ ] DFT%TRANSPORT
- [ ] DFT%PRINT%WANNIER90
- [ ] INTERPOLATOR%KIND [SPLINE3|SPLINE3_NOPBC] (does definitely not work in the MGRID section)

From the more recently added features, we have
- [x] OpenPMD
- [x] I-PI
- [x] HDF5
- [ ] NEGF%CONTACT%RESTART%READ_WRITE_HS
- [x] QCSchema
- [ ] SMEAGOL
- [ ] src/maxwell_solver_interface.F

I guess that for several of the more recent feature, it is pretty hard to actually test these routines on the CI. Who has some ideas about the older features?

oschuett:
Thanks for assembling this list! We should definitely keep an eye on test coverage.

Reporting the coverage by input section instead of [code line](https://www.cp2k.org/static/coverage/) feels much more actionable. 


[...]

PR #5814, opened 2026-08-21
Remove duplicate H-r and AO momentum construction paths

RTChou67:
This branch continues the cleanup started by #5734 in which I removed the duplicate `qs_operators_ao::ang_mom` and `p_xyz_ao` paths, reusing `ai_angmom::angmom` and `build_lin_mom_matrix`. This follow-up removes the duplicate build_com_tr_matrix and private lin_mom paths, switching to the more general build_lin_mom_matrix and overlap_ab APIs. The orphan `ai_moments::diffop` routine and public export are also removed.


Growl1234 (commented):
- Since `COMMUTATOR_HR` is currently not covered by the regression tests, could we add a small test before removing the old construction path?
- There's one behavioral difference that might be worth preserving or at least benchmarking is that `build_com_tr_matrix` was explicitly OpenMP-parallel over the neighbor list, whereas `build_lin_mom_matrix` currently traverses it serially.

RTChou67:
Thanks for commenting. `COMMUTATOR_HR` itself has not been removed. The `DFT%PRINT%AO_MATRICES%COMMUTATOR_HR` output path are still present. I agree that a small regtest should be added before merging . I searched and found out that this has already been mentioned in #4694 . 

The OpenMP difference is also real. Benchmarks might be needed.

juerghutter (approved):
Looks good to me. 
I have a currently stopped project where I will use those integrals and their nuclear derivatives. I will adapt to whatever is available if I restart the project.

RTChou67:
> Looks good to me. I have a currently stopped project where I will use those integrals and their nuclear derivatives. I will adapt to whatever is available if I restart the project.

thanks. My other PR #5820 adds a regtest for COMMUTATOR_HR. After that PR is merged, I would test on this PR and mark it as ready for review if it's numerically correct.

fstein93 (approved):

RTChou67:
The original `COMMUTATOR_HR` code is wrong, so I updated `COMMUTATOR_HR_Y` reference value. For Gaussian basis functions centered at \(A\) and \(B\), the complete kinetic-position commutator is

$$
\bra{a}[T,r_i]\ket{b}=(R_{B,i}-R_{A,i})\bra{a}T\ket{b}+\bra{a}T\ket{b+1_i}-\bra{a+1_i}T\ket{b}.
$$

The old construction omitted the center-displacement term

$$
(R_{B,i}-R_{A,i})\bra{a}T\ket{b}
$$

For the regtest `QS/regtest-gpw-1/h2q.inp`:
```
  H 0.0 0.0 0.0
  H 0.0 0.72 0.0
```
Thus,the omitted term only affects the Y term.

RTChou67:
This highlights the need for external libraries for electronic integrals and recurrences. At least for benchmarks.

Growl1234 (approved):
Looks good! And I think your analysis is right.

Growl1234:
Regarding my behavior concern, I had a look and would prefer working directly on `neighbor_list_iterate` for OpenMP parallelism, so nothing would be needed in this PR.

So @RTChou67, do you think this PR can be merged now?

RTChou67:
thanks. no further work would be needed in this PR. Can be safely merged.

PR #5811, opened 2026-08-21
Prepare the MPI wrapper for MPI_THREAD_MULTIPLE

Growl1234:
CP2K currently initializes MPI with `MPI_THREAD_SERIALIZED`, which prevents MPI calls from being issued concurrently by OpenMP threads. This is an infrastructure-level obstacle to improving OpenMP scalability in communication-heavy code paths.

This PR prepares the MPI wrapper for a gradual transition to `MPI_THREAD_MULTIPLE` without changing the parallel structure of production algorithms.

The main changes are:

- Request and require `MPI_THREAD_MULTIPLE` when MPI is initialized.
- Remove the `OMP MASTER` region around MPI initialization and require MPI lifecycle operations to be called outside OpenMP parallel regions.
- Make MPI performance-counter updates atomic.
- Prevent worker threads from modifying the existing process-global timing stack.
- Document the communicator, buffer, tag, and request ownership requirements for concurrent wrapper calls.
- Add a unit test exercising concurrent collectives and point-to-point communication on independently duplicated communicators.

PR #5817, opened 2026-08-22
Unify MPI lifecycle handling for MPI_THREAD_MULTIPLE

Growl1234:
Follow-up to #5811.

This PR extends the `MPI_THREAD_MULTIPLE` groundwork by making MPI initialization and ownership explicit in both the Fortran and C wrappers, and by making MPI wrapper instrumentation safe when calls originate from OpenMP worker threads.

The main changes are:

- Initialize MPI with `MPI_THREAD_MULTIPLE` when CP2K owns MPI initialization.
- When MPI was initialized by an embedding application, attach without taking ownership, query the provided thread-support level, and require `MPI_THREAD_MULTIPLE`.
- Finalize MPI only when CP2K initialized it.
- Detect duplicate or unmatched initialization/finalization calls and reject lifecycle operations after MPI has been finalized.
- Require MPI lifecycle operations and communicator creation, duplication, splitting, reference-count changes, and destruction to take place outside OpenMP parallel regions.
- Add an MPI-only unit test that exercises concurrent point-to-point communication from two OpenMP threads on a shared communicator, using distinct tags to keep message matching unambiguous.
- Use the ownership-aware C MPI initialization and finalization wrappers in `grid_unittest`.

### DIFF
--- a/src/mpiwrap/message_passing.F
+++ b/src/mpiwrap/message_passing.F
@@ -20,6 +20,8 @@
 !>      JGH (15-Feb-2006): single precision mp_alltoall
 !> \note Concurrent calls must use thread-private buffers and requests. Concurrent collectives,
 !>       or point-to-point calls that do not use distinct tags, must use distinct communicators.
+!>       Communicator creation, duplication, splitting, ownership changes, and destruction must
+!>       happen outside OpenMP parallel regions.
 !> \author JGH
 ! **************************************************************************************************
 MODULE message_passing
@@ -29,6 +31,7 @@ MODULE message_passing
       real_8_size, default_string_length
    USE machine, ONLY: m_abort
    USE mp_perf_env, ONLY: add_perf, add_mp_perf_env, rm_mp_perf_env
+   USE OMP_LIB, ONLY: omp_get_level, omp_get_thread_num, omp_in_parallel
 #if defined(__MIMIC)
    USE mcl, ONLY: mcl_initialize, mcl_is_initialized, mcl_abort
 #endif
@@ -74,8 +77,6 @@ MODULE message_passing
    USE mpi
 #endif
 #endif
-
-!$ USE OMP_LIB, ONLY: omp_get_level, omp_get_thread_num, omp_in_parallel
 
    IMPLICIT NONE
    PRIVATE
@@ -149,6 +150,13 @@ MODULE message_passing
    ! Thread-support level provided by MPI. A negative value means that MPI has
    ! not been initialized by CP2K or queried from an embedding application yet.
    INTEGER, PRIVATE, SAVE :: mp_thread_level_provided = -1
+
+   ! MPI can be initialized by CP2K or by an embedding application. Only the
+   ! component that initialized MPI is allowed to finalize it.
+#if defined(__parallel)
+   LOGICAL, PRIVATE, SAVE :: mp_mpi_initialized_by_cp2k = .FALSE.
+#endif
+   LOGICAL, PRIVATE, SAVE :: mp_world_is_initialized = .FALSE.
 
    PUBLIC :: mp_comm_type
    PUBLIC :: mp_request_type
@@ -838,7 +846,6 @@ MODULE message_passing
 
    ! init and error
    PUBLIC :: mp_world_init, mp_world_finalize, mp_query_thread_level
-   PUBLIC :: mp_get_thread_level, mp_has_thread_multiple
    PUBLIC :: mp_abort
 
    ! informational / generation of sub comms
@@ -1126,8 +1133,10 @@ CONTAINS
       SUBROUTINE mp_assert_outside_parallel(routine_name)
          CHARACTER(LEN=*), INTENT(IN)                     :: routine_name
 
-!$       IF (omp_in_parallel()) &
-!$          CPABORT(TRIM(routine_name)//" must be called outside an OpenMP parallel region.")
+         IF (omp_in_parallel()) THEN
+            CALL cp_abort(__LOCATION__, &
+                          TRIM(routine_name)//" must be called outside an OpenMP parallel region.")
+         END IF
       END SUBROUTINE mp_assert_outside_parallel
 
 ! **************************************************************************************************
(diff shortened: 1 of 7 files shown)


Growl1234:
~With the new commit, when `TIME_MPI` is enabled, MPI wrapper calls and accumulated wall time from top-level OpenMP worker threads are now recorded separately, without exposing the non-thread-safe global timing stack to concurrent callers.~

fstein93 (approved):

oschuett:
> Add mp_thread_comm_pool_type, which pre-duplicates one communicator per OpenMP lane outside the parallel region and provides the calling thread with its assigned communicator through get_comm().

This adds a lot of complexity. Is it really needed or is it an (possibly premature) optimization?

Growl1234:
The pool was intended to isolate independently issued collectives from different OpenMP lanes, and might also improve concurrency in some MPI implementations. Since `MPI_THREAD_MULTIPLE` itself does not require one communicator per thread, and there is no production caller yet, I agree it's probably premature.

src/mpiwrap/message_passing.F
```diff
@@ -1895,6 +2086,13 @@ ELEMENTAL IMPURE SUBROUTINE mp_comm_init(comm, owns_group)
          CLASS(mp_comm_type), INTENT(INOUT) :: comm
          LOGICAL, INTENT(IN), OPTIONAL :: owns_group
 
+         SELECT TYPE (comm)
```
oschuett, line 1949: I think, the introduction of the `mp_para_cart_type` was actually a mistake because it deviates from the underlying MPI type system. As a result we get this handrolled polymorphism (`SELECT TYPE... CLASS IS`) wherever we deal with the `mp_comm_type`.

It'd probably make these refactorings easier, if you first cleaned this up. I once [got started](https://github.com/oschuett/cp2k/commit/3047e4f7579dbace44a3e74d21114cb4807d55d0), but didn't get it across the finish line.

oschuett:
> The pool was intended to isolate independently issued collectives from different OpenMP lanes

So far we shouldn't have any of those, because it would have violated `MPI_THREAD_SERIALIZED`. I also don't think collectives are where the gains lay, because our OpenMP threads are almost never specialized and would therefor all participate in the same collectives.

Instead, I'd focus on point-to-point communication, e.g. in [dbm_comm_iterator_next](https://github.com/cp2k/cp2k/blob/77d8609f2e8c92aaa0f49e58e1337086fe86c182/src/dbm/dbm_multiply_comm.c#L627) and [transfer_rs2pw_distributed](https://github.com/cp2k/cp2k/blob/77d8609f2e8c92aaa0f49e58e1337086fe86c182/src/pw/realspace_grid_types.F#L953), that could take advantage of OpenMP.

src/mpiwrap/message_passing.F
fstein93: What do you mean? The introduction of reference-counting (or of communicators without reference counters)  or the additional abstraction of Cartesian communicators?

oschuett: I was referring to the _"the additional abstraction of Cartesian communicators"_.

fstein93: My rational was to give access to the routines `MPI_Cart_*` only if the communicator is Cartesian, i.e. to prevent accidental calls to these routines for non-Cartesian communicators. This includes the Cartesian variants `num_pe_cart` and `mepos_cart` which are implemented similarly to `num_pe` and `mepos` as public components of the communicator type to allow direct access by users of communicator types. If you want to merge the Cartesian with the non-Cartesian communicators, you either need to give them dummy-values or have to make them private and implement getters for them (and maybe even for their non-Cartesian counterparts and modify a lot of code downstream).

oschuett: I agree that the rational is in principle sound. So, to not derail this PR, I've opened #5840.

Growl1234: I am in favor of the merging them. That said, there are many code files using `mp_cart_type`, so I'd consider it as non-trival.

Or, we could at first consider cleaning up `mp_cart_type` itself if there are concerns and/or blockers.

oschuett: Thanks for looking into it. It's certainly not a blocker. So, let's continue this discussion at #5840.

src/mpiwrap/message_passing.F
```diff
@@ -1895,6 +1946,13 @@ ELEMENTAL IMPURE SUBROUTINE mp_comm_init(comm, owns_group)
...
+         CLASS IS (mp_para_env_type)
+            CALL mp_assert_outside_parallel("mp_comm_init")
+         CLASS IS (mp_para_cart_type)
+            CALL mp_assert_outside_parallel("mp_comm_init")
+         END SELECT
```
oschuett, line 1936: Since you're doing the same in each case, you shouldn't need the `SELECT` construct:

```suggestion
         CALL mp_assert_outside_parallel("mp_comm_init")
```

src/mpiwrap/message_passing.F
```diff
@@ -150,6 +151,24 @@ MODULE message_passing
...
+   INTEGER, PARAMETER, PRIVATE :: mp_no_timer_handle = -1
+   INTEGER, PARAMETER, PRIVATE :: mp_worker_timer_handle = -2
+   INTEGER, PARAMETER, PRIVATE :: mp_worker_timer_max_depth = 16
+
+   TYPE, PRIVATE :: mp_worker_timer_type
+      INTEGER :: depth = 0
+      REAL(KIND=dp), DIMENSION(mp_worker_timer_max_depth) :: start = 0.0_dp
+   END TYPE mp_worker_timer_type
+
+   TYPE(mp_worker_timer_type), ALLOCATABLE, DIMENSION(:), PRIVATE, SAVE :: mp_worker_timers
```
oschuett: I don't think any of these changes to the timing machinery are needed.

Btw, note that for message passing we actually have two mechanisms: Our generic timeset/timestop and an mpi-specific [add_perf](https://github.com/cp2k/cp2k/blob/master/src/mpiwrap/mp_perf_env.F). However, both mechanism should already be thread-safe.

src/mpiwrap/message_passing.F
Growl1234: The `SELECT TYPE` was intentional. Since it had no `CLASS DEFAULT` branch, the assertion was executed only when the dynamic type of `comm` was `mp_para_env_type` or `mp_para_cart_type`. Replacing it with an unconditional call extended the restriction to ordinary communicators and caused the [runtime abort](https://storage.googleapis.com/cp2k-ci/run-cp2k-spack-openmpi-psmp-2a122740_report.txt):

```

 *  \___/     mp_comm_init must be called outside an OpenMP parallel region.   *
 * / \                                          mpiwrap/message_passing.F:1137 *


           15 neighbor_list_3c_iterator_create
           14 build_3c_integrals
           ...
```

oschuett (approved):

PR #5820, opened 2026-08-22
Add COMMUTATOR_HR regtest matrix checksum

RTChou67:
Add regtest coverage for `DFT%PRINT%AO_MATRICES/COMMUTATOR_HR`. This follows #4694 and is a prerequisite for #5814 . The existing `regtest-gpw-1/h2q.inp` now enables `COMMUTATOR_HR` printing with 12 digits. CP2K now outputs checksums for the `COMMUTATOR_HR` matrix.

### DIFF
--- a/tests/QS/regtest-gpw-1/TEST_FILES.toml
+++ b/tests/QS/regtest-gpw-1/TEST_FILES.toml
@@ -74,7 +74,10 @@
 #
 "He_PBE.inp"                            = [{matcher="E_total", tol=1.0E-13, ref=-1.14398524502663}]
 #
-"h2q.inp"                               = [{matcher="E_total", tol=1.0E-10, ref=-0.76162210587786}]
+"h2q.inp"                               = [{matcher="E_total",          tol=1.0E-10, ref=-0.76162210587786},
+                                           {matcher="COMMUTATOR_HR_X", tol=1.0E-8, ref=4.560125368089048},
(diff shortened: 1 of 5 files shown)


tests/QS/regtest-gpw-1/TEST_FILES.toml
```diff
@@ -74,7 +74,8 @@
 #
 "He_PBE.inp"                            = [{matcher="E_total", tol=1.0E-13, ref=-1.14398524502663}]
 #
-"h2q.inp"                               = [{matcher="E_total", tol=1.0E-10, ref=-0.76162210587786}]
+"h2q.inp"                               = [{matcher="E_total",      tol=1.0E-10, ref=-0.76162210587786},
```
oschuett: Couldn't this just be expressed as three matchers instead of introducing a new matcher class?

RTChou67: The current output of COMMUTATOR_HR is three 16 times 16 matrices, with the same `COMMUTATOR` title. Maybe we could add scalar output to CP2K source code. Then we could use existing scalar matchers.

oschuett: Yes, to compare a matrix just output a suitable checksum. See for example #5192.

RTChou67: OK, thanks. I'll modify.

oschuett (approved):

PR #5822, opened 2026-08-22
EEQ: Add an FFT-accelerated sparse solver for periodic systems

DCM-Uni-Paderborn:
# Summary

This PR adds a matrix-free, charge-constrained solver for periodic
electronegativity equilibration (EEQ). It avoids constructing the full
distributed `(N + 1) x (N + 1)` response matrix when the sparse path is
selected.

The implementation is deliberately attached to CP2K's common EEQ
infrastructure rather than to one electronic-structure method. It is therefore
available to the meaningful internal consumers of CP2K EEQ charges:

- GFN0-xTB, where EEQ is part of the method;
- the internal DFT-D4 implementation, independent of the underlying DFT
  functional; and
- internal GFN1-xTB calculations when `VDW_POTENTIAL DFTD4` is selected.

The GFN1-D4 case previously did not propagate the xTB `&EEQ` solver settings to
the internal D4 environment. This PR closes that gap. It does not change the
EEQ parametrization or the resulting physical charge model.

# Relation to the FFT idea

This is not a direct one-shot FFT inverse of the complete EEQ matrix. Such an
inverse would require a circulant or block-circulant matrix in an appropriate
lattice basis. The atom-dependent positions, species-dependent hardnesses and
the fixed-total-charge constraint break that structure for a general atomic
system.

Instead, the periodic response is split conceptually as

```
A = D_hardness + A_short-range + A_long-range.
```

The diagonal atomic part and the real-space short-range response are retained
explicitly, with the off-diagonal short-range atom pairs stored in a
distributed sparse representation. The translationally invariant long-range
Coulomb response is never assembled as an atom-by-atom dense matrix. It is
applied matrix-free through CP2K's FFT-based smooth particle-mesh Ewald (SPME)
operator.

A projected, diagonally preconditioned conjugate-gradient iteration then
approximates the action of the inverse response in the subspace with fixed
total charge. Thus the FFT accelerates the structured long-range part of every
iterative inverse application; it does not diagonalize the chemically
inhomogeneous full matrix.

# Input and selection

The sparse path can be requested explicitly for xTB EEQ:

```
&XTB
  &EEQ
    SPARSE
  &END EEQ
&END XTB
```

or for the internal DFT-D4 charge model:

```
&VDW_POTENTIAL
  DISPERSION_FUNCTIONAL PAIR_POTENTIAL
  &PAIR_POTENTIAL
    TYPE DFTD4
    &EEQ
      SPARSE
    &END EEQ
  &END PAIR_POTENTIAL
&END VDW_POTENTIAL
```

`SPARSE_THRESHOLD` selects it automatically for periodic EEQ systems at or
above the requested atom count. Its default is 256 atoms; a non-positive value
disables automatic selection. An explicit `DIRECT` request retains precedence.

The established dense iterative and direct solvers remain available. If the
sparse iteration detects a numerical breakdown or does not converge, it falls
back to the established direct periodic solver. Automatic threshold selection
also retains the historical path for displacement-field calculations; an
explicit unsupported `SPARSE` plus displacement-field combination is rejected
with a diagnostic.

# Scope boundaries

The solver is enabled only where CP2K actually solves the common EEQ equations.
It is intentionally not wired into unrelated charge or dispersion models:

- DFT-D3 is charge independent and has no EEQ system to solve.
- CP2K's SCC-DFTB implementation determines its own self-consistent charges and
  normally uses D3 rather than the D4 EEQ model.
- tblite-managed GFN1/GFN2 calculations own their SCC charges and dispersion
  treatment inside tblite.
- `D4_REFERENCE_CODE` delegates the calculation to the external reference
  implementation and bypasses CP2K's internal EEQ solver.

These exclusions avoid changing the physics of methods that do not consume
CP2K EEQ charges while making the solver general for all current internal
consumers that do.

# Implementation details

- A CP2K neighbor list constructs canonical short-range atom pairs without an
  `O(N^2)` duplicate lookup table.
- Pair storage and construction are distributed across MPI ranks.
- Numerically negligible short-range elements are dropped using a tolerance
  tied to the requested solver accuracy and system size.
- Every response application combines the distributed sparse real-space part
  with `spme_potential` for the reciprocal-space contribution.
- CG vectors are projected onto the zero-sum tangent space of the total-charge
  constraint after each update.
- A positive diagonal hardness preconditioner is applied in that constrained
  subspace.
- The resulting charges are corrected for accumulated roundoff so that their
  sum remains the requested total charge.

# Validation

The full GCC 16 MPI/tblite build and the CP2K precommit checks pass.

- Existing periodic GFN0-xTB `si_kp.inp`, one and two MPI ranks: all 49 reported
  energies agree within `7.1e-15` hartree.
- New heteroatomic periodic PBE-D4 energy/force/stress regression, one and two
  MPI ranks: sparse PCG converges in two iterations. Sparse and dense results
  agree to printed precision for dispersion, energy and forces; the largest
  observed stress-component difference is approximately `2e-14` bar.
- New periodic internal GFN1-D4 energy/force/stress regression, one and two MPI
  ranks: sparse PCG converges in two iterations. Sparse and dense energies,
  dispersion and stress agree to printed precision; force-component differences
  are approximately `1e-21` atomic units.

The regression references cover total energy, D4 dispersion energy, forces and
the analytical stress trace. No performance claim is made by these correctness
tests; their purpose is to exercise the generalized consumer wiring and verify
serial/MPI equivalence against the established dense path.


src/eeq_method.F
```diff
@@ -1055,6 +1091,486 @@ SUBROUTINE mi_solver(charges, lambda, eeq_energy, eeq_mat, particle_set, kind_of
...
+      lambda = (-qtot - SUM(chia/precond_diag))/SUM(1.0_dp/precond_diag)
+      charges = -(chia + lambda)/precond_diag
+      charges = charges + (qtot - SUM(charges))/REAL(natom, KIND=dp)
+
+      CALL apply_sparse_pbc_response(response, ewald_env, ewald_pw, cell, particle_set, &
+                                     charges, aq, work)
+      gradient(:) = aq + chia
+      residual(:) = -gradient
+      CALL project_eeq_charge_tangent(residual)
+      res = SQRT(SUM(residual*residual))
```
fstein93, line 1177: As a follow-up PR: replace this with `NORM2(residual)`

src/eeq_method.F
```diff
@@ -1055,6 +1091,486 @@ SUBROUTINE mi_solver(charges, lambda, eeq_energy, eeq_mat, particle_set, kind_of
...
+            IF (.NOT. (ABS(alpha_cg) < HUGE(alpha_cg))) THEN
+               ierror = 1
+               EXIT
+            END IF
+            charges = charges + alpha_cg*p
+            residual(:) = residual - alpha_cg*ap
+            ! Remove accumulated roundoff from both constrained spaces.
+            charges = charges + (qtot - SUM(charges))/REAL(natom, KIND=dp)
+            CALL project_eeq_charge_tangent(residual)
+            res = SQRT(SUM(residual*residual))
```
fstein93, line 1206: Same

src/eeq_method.F
```diff
@@ -1055,6 +1091,486 @@ SUBROUTINE mi_solver(charges, lambda, eeq_energy, eeq_mat, particle_set, kind_of
...
+         cell_extent = 0.5_dp
+      ELSE WHERE
+         cell_extent = 1.0_dp
+      END WHERE
+      neighbor_cutoff = 0.0_dp
+      DO ix = -1, 1, 2
+         DO iy = -1, 1, 2
+            DO iz = -1, 1, 2
+               vertex = MATMUL(hmat, cell_extent*REAL([ix, iy, iz], KIND=dp))
+               neighbor_cutoff = MAX(neighbor_cutoff, SQRT(SUM(vertex**2)))
```
fstein93, line 1327: `NORM2(vertex)`

src/eeq_method.F
DCM-Uni-Paderborn: Addressed in #5838: the residual norm now uses `NORM2(residual)`. The follow-up was formatted, built with GCC 16/OpenMPI, and checked with the sparse periodic EEQ regression.

src/eeq_method.F
DCM-Uni-Paderborn: Addressed in #5838: the cell-vertex cutoff now uses `NORM2(vertex)`.

src/eeq_method.F
DCM-Uni-Paderborn: Addressed in #5838 as well.

PR #5823, opened 2026-08-22
KPOINTS: Add FFT-based transforms for complete regular k-point grids

DCM-Uni-Paderborn:
# Summary

This PR adds a lattice-FFT path for transforming real-space Hamiltonian and overlap matrices to
complete regular k-point grids during conventional k-point diagonalization.

The established implementation evaluates the lattice phase sum independently for every k point,

```text
A(k) = sum_R exp(2 pi i k.R) A(R).
```

For a complete regular reciprocal-space grid this operation is a discrete Fourier transform over
the lattice translations. The new path folds equivalent real-space image cells modulo the
reciprocal-grid dimensions and obtains all k-point values of each locally stored matrix element
with CP2K's existing three-dimensional FFT infrastructure.

# FFT algorithm

The implementation first verifies that the supplied reciprocal coordinates form a complete,
uniformly shifted grid in arbitrary k-point order. Each real-space image vector is then mapped to
its lattice index modulo the grid dimensions. Multiple images mapping to the same grid point are
accumulated before the FFT.

For shifted Monkhorst-Pack meshes, the common reciprocal-grid offset is applied as a real-space
phase twist. This reduces the remaining transform to the regular discrete FFT while retaining the
correct complex Bloch phases.

The direct phase sum scales as `O(N_cell * N_k)` per matrix element. The FFT path instead requires
the real-space folding plus an `O(N_k log N_k)` transform. The current implementation batches the
k-point dimension: each MPI-local DBCSR matrix element is transformed once to all k points.

Short or singleton grid dimensions that are unsupported directly by FFTSG are embedded into a
compatible integer-multiple FFT grid. This keeps common lower-dimensional meshes such as
`4 x 4 x 1` on the FFT path without changing their discrete transform.

# Input and automatic selection

The new `&KPOINTS` keyword has three modes:

```text
&KPOINTS
  LATTICE_FFT AUTO
&END KPOINTS
```

- `AUTO` is the default. It considers the FFT for at least 27 supplied k points and then applies
  all regular-grid, backend, and memory checks.
- `ON` bypasses only the conservative 27-point performance threshold and prefers the FFT whenever
  the mathematical and resource checks allow it.
- `OFF` always uses the established direct phase sum.

The threshold is conservative because a serial GFN2 benchmark found that the FFT overhead does
not pay off for the tested `2 x 2 x 2` grid, while a crossover was observed at `3 x 3 x 3`.

`ON` never disables safety checks. Incomplete, irregular, or symmetry-reduced point lists,
unsupported FFT dimensions, failed FFT calls, and local cache estimates above the memory limit
all retain the direct fallback. The cache budget is approximately 512 MiB per MPI rank, divided
between the overlap matrix and the spin-dependent Hamiltonian matrices.

# SCF integration

During conventional k-point diagonalization, the new path transforms the MPI-local DBCSR blocks
of the Kohn-Sham or Hamiltonian matrix for each spin channel and of the overlap matrix. The
transformed local blocks are cached, and the existing real or complex matrix for a requested k
point is extracted when the current diagonalization and communication workflow needs it.

No global matrix redistribution is introduced. Each rank transforms only its locally stored
DBCSR blocks, and the existing matrix pools, k-point groups, eigensolvers, and downstream SCF
interfaces remain unchanged.

# Fallback and scope

The original `rskp_transform` phase sum remains the reference and fallback implementation. This
PR changes only the real-cell to reciprocal-grid transformation used by conventional k-point
diagonalization. It does not change:

- Hamiltonian or overlap matrix elements;
- k-point coordinates, weights, or symmetry reduction;
- occupations, smearing, or Mermin free energies;
- eigensolvers or orbital-transformation preconditioners;
- density-matrix back-transforms;
- force or stress evaluation; or
- restart handling.

The transform is useful infrastructure for later complex k-point and metallic-OT work, but this
PR does not contain the coupled OT-for-metals physics or integration.

# Validation

The full GCC 16 MPI/tblite build and the CP2K precommit checks pass.

A focused unit test compares the lattice FFT with an explicit direct reference for:

- a uniformly shifted `4 x 4 x 4` grid with 64 k points and 31 image cells;
- arbitrary k-point order;
- selected k-point output;
- a Cartesian k derivative;
- a padded `4 x 4 x 1` grid; and
- an intentionally irregular grid that must use the direct fallback.

FFT and direct results agree within `2e-12` for regular and selected transforms, `5e-12` for the
derivative, and `1e-14` for the irregular-grid fallback.

The existing complex `FULL_GRID` regression now explicitly requests `LATTICE_FFT ON` on its
`2 x 2 x 2` grid. It passes with one and two MPI ranks:

```text
serial:    -41.897080363583598 hartree
MPI-2:     -41.897080363583591 hartree
reference: -41.897080363575512 hartree
tolerance: 1e-10 hartree
```

The production selection modes were also exercised with periodic GFN2 silicon calculations:

- `AUTO` at `2^3` selected the direct phase sum;
- `AUTO` at `3^3` selected the FFT;
- `ON` at `2^3` selected the FFT; and
- `OFF` at `6^3` selected the direct phase sum.

A cheap serial GFN2 timing used an eight-atom silicon cell, 18 identical SCF iterations, and three
repetitions per grid. Mean total wall times were:

| Grid | FFT [s] | Direct [s] | Total-time change | Transform speedup |
| --- | ---: | ---: | ---: | ---: |
| `2^3` | 0.926 | 0.866 | 6.9% slower | 0.60x |
| `3^3` | 1.695 | 1.839 | 7.8% faster | 1.67x |
| `4^3` | 3.244 | 3.797 | 14.6% faster | 4.44x |
| `6^3` | 9.925 | 12.741 | 22.1% faster | 13.00x |

FFT and direct benchmark energies agreed within approximately `1.1e-14` hartree. These timings
motivate the conservative `AUTO` threshold; they are a small serial crossover study rather than a
general parallel performance claim.


src/kpoint_lattice_fft.F
```diff
@@ -0,0 +1,285 @@
...
+      IF (.NOT. regular) THEN
+         CALL direct_cell_to_k(values_rs, index_to_cell, xkp, values_k, deriv_direction, hmat, &
+                               selected_kpoints)
+         IF (PRESENT(used_fft)) used_fft = .FALSE.
+         DEALLOCATE (grid_index)
+         CALL timestop(handle)
+         RETURN
+      END IF
+
+      ALLOCATE (fft_in(nfft(1), nfft(2), nfft(3)), fft_out(nfft(1), nfft(2), nfft(3)))
```
fstein93, line 194: These should be allocated using fft_alloc and deallocated using fft_dealloc for performance

DCM-Uni-Paderborn: Addressed in #5837: both rank-three FFT buffers now use `fft_alloc`/`fft_dealloc`. The pointers are marked `CONTIGUOUS` to match the allocator interface, and the lattice FFT unit test passes.

PR #5824, opened 2026-08-22
OT: Add optional Chebyshev preconditioner solver

DCM-Uni-Paderborn:
# Summary

This PR adds an optional Chebyshev solver for applying the `FULL_KINETIC` OT
preconditioner. It approximates the action of the inverse with a positive
Chebyshev semi-iteration and therefore uses matrix products and vector-like
matrix updates instead of forming or factorizing the inverse.

The feature is explicitly opt-in. `PRECOND_SOLVER DEFAULT` remains unchanged
and continues to select the established inverse-Cholesky path for
`FULL_KINETIC`.

# Algorithm

At preconditioner construction, the symmetric positive-definite
`FULL_KINETIC` operator is retained instead of being replaced by its inverse.
The implementation estimates the lower spectral edge with the existing
Arnoldi machinery, applies a conservative factor of two safety margin, and
uses the Gershgorin row-sum norm as an upper spectral bound. Invalid spectral
bounds fall back to the established Cholesky inverse with a warning.

Each application solves the preconditioning equation approximately with a
fixed-degree Chebyshev semi-iteration. Its residual polynomial is minimax on
the estimated spectral interval. In contrast to a directly truncated series
for `1/x`, the corresponding inverse polynomial remains positive throughout
that interval, which is important because OT requires a positive-definite
preconditioner.

Both the distributed DBCSR path used by conventional OT and the dense matrix
application are implemented. A degree `m` application needs `m - 1` operator
matrix products after the initial scaled direction and two temporary matrices.

# Input and default behavior

The solver is selected explicitly:

```text
&SCF
  &OT
    PRECONDITIONER FULL_KINETIC
    PRECOND_SOLVER CHEBYSHEV
    CHEBYSHEV_DEGREE 8
  &END OT
&END SCF
```

`CHEBYSHEV_DEGREE` accepts values from 1 to 100 and defaults to 8 when
Chebyshev is selected. The degree controls the cost/accuracy tradeoff of each
inverse application. Lower degrees are cheaper but can increase the number of
OT iterations or reduce robustness.

The new solver currently rejects preconditioners other than `FULL_KINETIC`.
No automatic selection heuristic is introduced, and no existing input changes
behavior.

# Intended use and performance scope

This is an alternative computational kernel rather than a general CPU speedup.
It avoids a factorization and explicit inverse and expresses repeated
applications through matrix multiplication, which can be useful to explore on
large sparse matrices and accelerator-oriented builds.

Representative serial CPU tests did not show a consistent end-to-end gain:
small methane was slower than inverse Cholesky, a 40-atom replicated methane
case was approximately at parity, and a fixed two-step 320-atom case was about
6% slower overall. Degree 6 reduced application cost further but degraded OT
convergence; degree 8 was retained as the opt-in default. These measurements
are the reason this PR does not make Chebyshev the default or claim a general
performance improvement.

# Scope boundaries

This PR changes only how an already constructed `FULL_KINETIC` OT
preconditioner can be applied. It does not change:

- the default OT preconditioner or solver;
- the OT functional, minimizers, or line searches;
- occupations, smearing, Mermin free energies, or metallic OT;
- k-point handling;
- forces, stress, symmetry, or restart handling; or
- the established Cholesky and inverse-update implementations.

In particular, this is independent of the coupled OT-for-metals work. It is a
standalone optional preconditioning experiment for the existing OT machinery.

# Validation

The full GCC 16 MPI/tblite build and CP2K precommit checks pass on current
`master`.

A new methane regression exercises
`FULL_KINETIC/PRECOND_SOLVER CHEBYSHEV/CHEBYSHEV_DEGREE 8`:

```text
serial:    18 OT steps, -7.945068706413583 hartree
MPI-2:     18 OT steps, -7.945068706413570 hartree
reference:              -7.945068706413580 hartree
tolerance:               1e-10 hartree
```

The same input with `PRECOND_SOLVER DEFAULT` still converges through the
established path in 16 OT steps. An explicit `CHEBYSHEV_DEGREE 0` is rejected
with the documented 1--100 range diagnostic. No positive-definiteness warning
was observed in the serial or MPI Chebyshev runs.


Growl1234:
The sdbg failure is not related to this PR. Do you think this can be merged anyways?

Growl1234 (commented):
One concern I have is that the Chebyshev path currently seems to introduce a `DBCSR -> FM -> DBCSR` round-trip. In `solve_preconditioner()`, the sparse operator is converted to a full matrix even though the Chebyshev application itself can operate directly on DBCSR, and `make_preconditioner()` may then convert it back to DBCSR afterwards.

This appears to defeat part of the motivation for this solver, especially for large sparse systems: it introduces an `O(N^2)` dense temporary together with unnecessary data movement/communication. Would it be possible to keep the operator in DBCSR for the successful Chebyshev path, and only convert to FM when the caller actually requires the dense representation?

DCM-Uni-Paderborn:
The `sdbg` report shows no regression from this PR: all 5856/5856 tests are correct, including `QS/regtest-ot-2/CH4-chebyshev.inp` (9.27 s). The job is red only because the unchanged `QS/regtest-nlcc-1/h2o_polar_mgga2.inp` was classified as slow (33.62 +/- 1.15 s versus a 27.74 s threshold). This is an unrelated timing fluctuation, so no source-code change in this PR is appropriate; the PR can be merged as-is or the `sdbg` check can be rerun.

DCM-Uni-Paderborn:
Follow-up in `5e54eba008`: the previously flagged TPSS/NLCC polarizability regression was made cheaper by reducing its MGRID cutoff from 300 to 250 Ry and updating the matching checksum. The SCF and LINRES convergence thresholds, TPSS/NLCC path, three response directions, and matcher tolerance are unchanged. On the same 1-rank/1-thread local build, three 300 Ry/strict-threshold measurements gave 9.92 s for the baseline, while two 250 Ry measurements gave 7.10 and 7.07 s (about 29% faster). The M087 matcher passes with the new reference. The change is a separate commit so it is easy to review.

PR #5825, opened 2026-08-22
OT: Add a rotationally covariant bounded low-rank preconditioner

DCM-Uni-Paderborn:
## Summary

This PR adds the explicitly selected OT preconditioner `FERMI_LOW_RANK`. It keeps the robust inverse-overlap background of a conventional OT preconditioner and restores detailed spectral resolution only in a bounded complementary subspace:

```math
\mathcal P_C[G]
= \frac{S^{-1}G}{w}
+ V_r\,\text{diag}\!\left[
\frac{1}{\max(g,\varepsilon_a-\mu)}-\frac{1}{w}
\right]V_r^{\mathsf T}G.
```

Here:

- $G$ is the OT gradient;
- $\mu$ is the largest occupied Ritz value;
- $g$ is the effective denominator floor;
- $w=\max(1\,\mathrm{Ha},4g)$ is the far-spectrum window;
- $V_r$ contains at most 48 complementary states.

Directions outside the retained subspace receive the positive inverse-overlap weight $1/w$. Retained frontier directions receive the more specific inverse spectral weight $1/\max(g,\varepsilon_a-\mu)$. The default OT preconditioner is unchanged.

## Input

The new path is opt-in:

```text
&OT
  PRECONDITIONER FERMI_LOW_RANK
  ROTATION T
&END OT
```

Only `PRECOND_SOLVER DEFAULT` is currently supported for this preconditioner.

## Why a bounded low-rank operator?

`FULL_ALL` resolves the complete spectrum and uses state-dependent occupied shifts. Its application requires full spectral transforms and is not compatible with OT `ROTATION`.

`FERMI_LOW_RANK` instead applies:

- one inverse-overlap multiplication over the full space; and
- two spectral transforms involving at most 48 retained complementary vectors.

This bounds the additional application cost independently of the total complementary-space size. It can therefore be attractive when the basis is much larger than the occupied space and only the near-frontier curvature needs detailed treatment.

Construction still performs a full diagonalization. This PR reduces and bounds repeated application cost; it does not yet provide an iterative or low-memory constructor. Small-system overhead can dominate, so no universal wall-time speedup is claimed.

## Rotational covariance

Changing only the spectral weights would not be sufficient. The complementary states are built from the occupied-projector-invariant operator

```math
Q_C = I-CC^{\mathsf T}S,
\qquad
H_{\mathrm{comp}}(C)=Q_C^{\mathsf T}HQ_C.
```

For an occupied-space rotation $U$ with $U^{\mathsf T}U=I$,

```math
Q_{CU}=Q_C,
\qquad
\mathcal P_{CU}=\mathcal P_C,
\qquad
\mathcal P_{CU}[GU]=\mathcal P_C[G]U.
```

The implementation uses one common reference $\mu$ and one common spectral-separation shift for every occupied column. It also keeps the occupied orbital gauge unchanged while constructing the preconditioner. Consequently, the operator is compatible with OT `ROTATION` while retaining a spectral low-rank advantage.

The effective-gap safeguard uses an occupied-gauge-invariant Frobenius residual. If the hard rank limit would split a numerically degenerate boundary manifold, the complete boundary cluster is excluded instead.

## Implementation

- Adds `FERMI_LOW_RANK` to the OT preconditioner input and output descriptions.
- Stores the exact inverse-overlap base, common reference, retained rank, and spectral metadata in the existing preconditioner lifecycle.
- Implements both full-matrix and DBCSR application paths.
- Uses the same column-independent operator for any number of input columns, including the separate OT eigensolver for requested unoccupied states.
- Preserves positive weights through the denominator floor and positive far-spectrum window.
- Adds an algebraic covariance unit test and an integrated OT `ROTATION` regression.

## Scope and follow-ups

- This is an experimental opt-in preconditioner, not a new automatic default.
- `FERMI` denotes the occupied spectral frontier; this PR does not implement Fermi-Dirac occupations, Mermin free-energy OT, or metallic convergence logic.
- The implemented production path is real Gamma-point OT with fixed integer occupations. Complex K-point channels and the coupled Mermin/`ADDED_MOS` workflow remain outside this PR and belong to the separate OT-for-metals work.
- This PR is purely algebraic and contains no FFT approximation. The optional block-circulant lattice-FFT base is developed separately in #5830.
- A future constructor could replace the full diagonalization with an iterative frontier eigensolver without changing the application model introduced here.

## Validation

On the current rebased head against `master`:

- full GCC 16 MPI/OpenMP build: passed;
- `low_rank_preconditioner_unittest`: passed with 1, 2, and 4 MPI ranks;
- CP2K precommit: 17/17 changed files passed;
- complete `QS/regtest-ot-2`: 49/49 matchers passed;
- `git diff --check`: passed.

The unit test verifies:

- $\mathcal P_{CU}[GU]=\mathcal P_C[G]U$;
- invariance of the complementary operator under $C\mapsto CU$;
- the common-reference inverse weights; and
- rank truncation without splitting a degenerate boundary manifold.

The committed replicated-CH4 regression uses OT `ROTATION`, 66 basis functions, 8 occupied orbitals, and the maximum retained rank of 48. It converges in 23 OT steps to `-15.9366475738458 Ha`, consistently in serial and MPI-2 runs.


src/low_rank_preconditioner_unittest.F
```diff
@@ -0,0 +1,107 @@
...
+   IF (MAXVAL(ABS(result_rotated - MATMUL(RESULT, rotation))) > tolerance) THEN
+      ERROR STOP "Low-rank preconditioner is not right-covariant under occupied rotations"
+   END IF
+
+   ! Verify that the projected complementary operator depends only on the occupied subspace.
+   occupied = 0.0_dp
+   occupied(1, 1) = 1.0_dp/SQRT(2.0_dp)
+   occupied(2, 1) = 1.0_dp/SQRT(2.0_dp)
+   occupied(3, 2) = 1.0_dp/SQRT(2.0_dp)
+   occupied(4, 2) = 1.0_dp/SQRT(2.0_dp)
```
fstein93: This constants is predefined as `sqrthalf` in `mathconstants`

DCM-Uni-Paderborn: Addressed in e94161cb4e. The test now imports `sqrthalf` from `mathconstants` and uses it for all four occupied-vector entries.

DCM-Uni-Paderborn:
Rebased onto the current master after #5844. The conflict resolution retains the new ADIIS validation and imports while adding the low-rank OT preconditioner paths. The complete local minimal SSMP build passes, as do the low-rank and lattice-FFT unit-test executables; precommit passes for all 17 PR files.

DCM-Uni-Paderborn:
Rebased the four-commit branch onto current master `7518516c65`. The only conflict was additive CMake test registration; the resolution retains both the new upstream frontier-orbital optimizer unit and `low_rank_preconditioner_unittest`. A range-diff shows the feature patches unchanged apart from that upstream registration context.

Exact-head validation passes: full GCC 16 MPI/OpenMP build; `low_rank_preconditioner_unittest` with MPI 1/2/4; no-cache precommit 17/17; complete `QS/regtest-ot-2` 49/49; and `git diff --check`. The branch is linear and zero commits behind master.

fstein93 (approved):

Growl1234:
@juerghutter Are you agreeing with the change?

juerghutter (approved):

PR #5826, opened 2026-08-22
Scale declaration cleaner work limit

DCM-Uni-Paderborn:
## Summary

- scale the declaration dependency-pass work limit with the number of parsed variables
- raise a defined `RuntimeError` if the scaled limit is still exceeded
- add a regression covering 500 independent declarations

This keeps the existing algorithm and its small-input behavior unchanged while avoiding a fixed-size failure for generated or unusually declaration-heavy Fortran routines.

## Testing

- `python3 tools/precommit/format_fortran_test.py`
- `./tools/precommit/precommit.py tools/precommit/format_fortran_test.py tools/precommit/prettify_cp2k/normalizeFortranFile.py`

### DIFF
--- a/tools/precommit/prettify_cp2k/normalizeFortranFile.py
+++ b/tools/precommit/prettify_cp2k/normalizeFortranFile.py
@@ -529,7 +529,7 @@ def enforceDeclDependecies(declarations):
     """enforces the dependencies between the vars
     and compacts the declarations, returns the variables needed by other variables"""
     nvars = sum(len(d["vars"]) for d in declarations)
-    max_iterations = max(100000, 10 * nvars * max(1, nvars))
+    max_iterations = max(100000, 10 * nvars**2)
     idecl = 0
     ii = 0
     while idecl < len(declarations):
--- a/tools/precommit/prettify_cp2k/normalizeFortranFile.py
+++ b/tools/precommit/prettify_cp2k/normalizeFortranFile.py
@@ -528,6 +528,8 @@ def findWord(word, text, options=re.IGNORECASE):
 def enforceDeclDependecies(declarations):
     """enforces the dependencies between the vars
     and compacts the declarations, returns the variables needed by other variables"""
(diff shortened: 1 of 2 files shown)


oschuett (approved):

tools/precommit/prettify_cp2k/normalizeFortranFile.py
```diff
@@ -528,6 +528,8 @@ def findWord(word, text, options=re.IGNORECASE):
 def enforceDeclDependecies(declarations):
     """enforces the dependencies between the vars
     and compacts the declarations, returns the variables needed by other variables"""
+    nvars = sum(len(d["vars"]) for d in declarations)
+    max_iterations = max(100000, 10 * nvars * max(1, nvars))
```
oschuett: This looks like a complicated way to to write 
```suggestion
    max_iterations = max(100000, 10 * nvars **2)
```

oschuett:
I'm a bit worried about run time because formatting Fortran is already quite slow:
https://github.com/cp2k/cp2k/blob/77d8609f2e8c92aaa0f49e58e1337086fe86c182/tools/precommit/precommit.py#L293-L295

tools/precommit/prettify_cp2k/normalizeFortranFile.py
DCM-Uni-Paderborn: Thanks, agreed. I simplified the expression to `max(100000, 10 * nvars**2)` in bf385d0feb.

DCM-Uni-Paderborn:
Thanks, that is a fair concern. The change only scales the existing failure guard; it does not add any iterations for routines that already remain below the old limit. The synthetic 500-declaration case that previously crossed the fixed guard completes in about 0.30 s locally, and the full associated Python test suite completes in 0.287 s. Thus normal formatting work is unchanged; only a large routine that would previously abort is allowed to finish its already-required dependency scan.

PR #5830, opened 2026-08-23
OT: Add an optional lattice-FFT base for low-rank preconditioning

DCM-Uni-Paderborn:
## Summary

This PR adds an explicitly opt-in block-circulant overlap-inverse base for the rotationally covariant `FERMI_LOW_RANK` OT preconditioner introduced in #5825.

For a Gamma-point supercell built from equivalent replicated cells, the overlap is averaged over lattice translations,

```math
\bar S_{\alpha\beta}(\Delta R)
= \frac{1}{N_R}\sum_R
S_{R\alpha,(R-\Delta R)\beta},
```

and transformed to small reciprocal-space blocks. The approximate inverse is then applied as

```math
B^{-1}G
= \mathcal F^{-1}\!\left[\bar S(k)^{-1}\,\mathcal F[G]\right],
```

where the FFT acts on cell translations and each $\bar S(k)$ spans only the atomic orbitals of one reference cell. This is a lattice FFT, distinct from the plane-wave FFT used by GPW.

The complete preconditioner is

```math
\mathcal P_C^{\mathrm{lat}}[G]
= \frac{B^{-1}G}{w}
+ V_r\,\text{diag}\!\left[
\frac{1}{\max(g,\varepsilon_a-\mu)}-\frac{1}{w}
\right]V_r^{\mathsf T}G.
```

The common reference level $\mu$ and the column-independent operator preserve right covariance under occupied-space rotations:

```math
\mathcal P_{CR}^{\mathrm{lat}}[GR]
= \mathcal P_C^{\mathrm{lat}}[G]R.
```

## Input and default behavior

The new base is selected explicitly:

```text
&OT
  PRECONDITIONER FERMI_LOW_RANK
  FERMI_LOW_RANK_BASE LATTICE_FFT
  ROTATION T
&END OT
```

`FERMI_LOW_RANK_BASE` defaults to `OVERLAP_INVERSE`. Existing inputs and the default preconditioner behavior are therefore unchanged. This PR deliberately adds neither automatic selection nor a new default.

## Applicability and safeguards

The lattice-FFT base currently requires:

- real Gamma-point OT;
- a non-unit overlap metric;
- an explicit matching `SUBSYS/CELL/MULTIPLE_UNIT_CELL` containing more than one equivalent cell.

The implementation:

- reports the relative Frobenius error of the block-circulant projection;
- Hermitizes each reciprocal-space overlap block and requires a successful positive-definite factorization before inversion;
- reduces the retained correction rank when necessary so that all low-rank updates remain non-negative and the complete approximate preconditioner stays positive definite;
- embeds unsupported short FFTSG lattice lengths into exact compatible transforms, without requiring `EXTENDED_FFT_LENGTHS` in the input;
- aborts with a targeted diagnostic if the explicitly requested FFT path cannot be constructed.

For a perfectly translationally equivalent replicated system, the block-circulant projection is exact up to numerical noise. With defects, interfaces, disorder, or inequivalent cells, it is an averaged approximation whose reported projection error should be monitored.

## FFT interface and #5868

The lattice algorithm is independent of the long-term FFT backend. This branch currently adds a small direction-generic unplanned 1-D FFT adapter because both forward and backward batched transforms are required. #5868 is redesigning the same low-level interface and proposes `fft_1d_many`; once that refactor lands, this caller should be rebased onto that interface and the temporary adapter in this PR can be replaced. No input keyword or numerical behavior in this PR depends on `EXTENDED_FFT_LENGTHS`.

## Parallelism and performance

The first implementation gathers each distributed dense input, performs the small batched lattice transforms redundantly on every MPI rank, and redistributes the result. The spectral low-rank correction remains distributed. A genuinely distributed lattice FFT is future scalability work.

In a local release build for an eight-cell CH4 benchmark, 36 low-rank preconditioner applications took 0.080 s with the exact overlap-inverse base and 0.054 s with the lattice-FFT base, about 32% less time in this kernel. Total SCF time was essentially unchanged because other QS work dominates this small case. No universal end-to-end speedup is claimed.

## Scope

This PR changes only the base operator of `FERMI_LOW_RANK`. It does not add Mermin or metallic occupations, `ADDED_MOS`, complex K-point channels, or new restart, symmetry, force, or stress handling. Those topics remain separate from this preconditioner PR and from the OT-for-metals work.

The lattice-specific work is stacked on #5825. After that dependency, this PR consists of five focused commits covering the transform helpers, the base construction/application, regression coverage, and review fixes.

## Validation

On the rebased head against current `master`:

- full GCC 16 Release build with MPI and OpenMP: passed;
- `kpoint_lattice_fft_unittest`: passed with 1, 2, and 4 MPI ranks;
- `low_rank_preconditioner_unittest`: passed with 1, 2, and 4 MPI ranks;
- CP2K precommit: 23/23 changed files passed;
- complete `QS/regtest-ot-2`: 50/50 matchers passed with two MPI ranks, including the new lattice-FFT regression.

The committed replicated-CH4 case converges to `-15.93664757401410 Ha`; its translational projection error is below `4.6e-16`.

Depends on #5825.


src/pw/fft_tools.F
```diff
@@ -354,6 +354,46 @@ SUBROUTINE fft_fw1d(n, m, trans, zin, zout, scale, stat)
...
+!> \param fsign transform direction, FWFFT or BWFFT
+!> \param n size of every FFT
+!> \param m number of FFTs
+!> \param trans shape of input and output arrays, as in fft_create_plan_1dm
+!> \param zin input array
+!> \param zout output array
+!> \param scale output scaling factor
+!> \param stat zero on success
+! **************************************************************************************************
+   SUBROUTINE fft_many_1d(fsign, n, m, trans, zin, zout, scale, stat)
```
fstein93: There is already a similar routine right above. Could you either merge both routines or implement a separate one for backwards-FFTs?

src/kpoint_lattice_fft.F
```diff
@@ -238,6 +240,143 @@ SUBROUTINE cell_to_k_grid_fft(values_rs, index_to_cell, xkp, nkp_grid, values_k,
...
+      END IF
+      IF (.NOT. regular) THEN
+         CALL direct_k_to_cell(values_k, xkp, index_to_cell, values_rs)
+         IF (PRESENT(used_fft)) used_fft = .FALSE.
+         DEALLOCATE (grid_index)
+         CALL timestop(handle)
+         RETURN
+      END IF
+
+      ALLOCATE (fft_in(nfft(1), nfft(2), nfft(3)), fft_out(nfft(1), nfft(2), nfft(3)))
```
fstein93: Please allocate the arrays passed to the FFT-routines using `fft_alloc` and deallocate them with `fft_dealloc` for better performance using FFTW

src/lattice_low_rank_preconditioner.F
```diff
@@ -0,0 +1,509 @@
...
+      INTEGER, INTENT(IN)                                :: direction, fsign
+      INTEGER, INTENT(OUT)                               :: stat
+
+      COMPLEX(KIND=dp), ALLOCATABLE, DIMENSION(:, :)     :: line_in, line_out
+      INTEGER                                            :: ibatch, iline, ix, iy, iz, m, n
+      REAL(KIND=dp)                                      :: scale
+
+      n = SIZE(grid, direction + 1)
+      m = SIZE(grid)/n
+      ALLOCATE (line_in(n, m), line_out(n, m))
```
fstein93: Allocate these arrays using `fft_alloc` and deallocate them using `fft_dealloc`

fstein93 (changes requested):

src/pw/fft_tools.F
DCM-Uni-Paderborn: Addressed in 5eea082f84. The direction-generic public routine has been removed. `fft_fw1d` and the new `fft_bw1d` now call one shared private implementation, preserving the existing forward API while exposing the backward transform explicitly.

src/kpoint_lattice_fft.F
DCM-Uni-Paderborn: Addressed in 5eea082f84. The 3-D input and output buffers in both lattice-transform directions are now allocated with `fft_alloc` and released with `fft_dealloc`.

src/lattice_low_rank_preconditioner.F
DCM-Uni-Paderborn: Addressed in 5eea082f84. The batched 1-D line buffers are now contiguous pointers allocated with `fft_alloc` and released with `fft_dealloc`.

src/pw/fft_tools.F
```diff
@@ -337,22 +337,78 @@ SUBROUTINE fft_fw1d(n, m, trans, zin, zout, scale, stat)
...
+      TYPE(fft_plan_type)                                :: fft_plan
+
+      compatible = fft_type == 3
+      IF (fft_type == 1 .AND. n >= 3) THEN
+         CALL fft_radix_operations(n, radix_length, FFT_RADIX_NEXT)
+         compatible = radix_length == n
+      END IF
+
+      IF (compatible) THEN
+         CALL fft_create_plan_1dm(fft_plan, fft_type, fsign, trans, n, m, zin, zout, fft_plan_style)
```
fstein93, line 348: I think it does not make any sense to have this kind of routine three times (1 forward, 1 backward, 1 for both). I would rather have only the combination forward+backward or a generic one and in any case without a check of the radix in case someone really needs an FFT for arbitrary lengths.

fstein93 (changes requested):

src/pw/fft_tools.F
DCM-Uni-Paderborn: Thanks, agreed. Addressed in 4861642fa4: the forward and backward wrappers have been removed in favor of a single public `fft_1d(fsign, ...)` routine, and the existing NEGF caller and the new lattice caller now use that generic interface. The radix gate inside the routine has also been removed. Precommit, the full CP2K build, the unit test in serial and with two MPI ranks, and all 38 OT regression tests pass.

src/pw/fft_tools.F
DCM-Uni-Paderborn: Follow-up: 4861642fa4 simplifies this further to the other suggested design: one public direction-generic `fft_1d` routine, with both the existing forward caller and the new forward/backward caller migrated to it.

DCM-Uni-Paderborn:
Follow-up after #5837 was merged: I rebased this branch onto the current master and resolved the allocation overlap by retaining the NULLIFY initialization introduced by #5837 before fft_alloc. The PR is conflict-free again, and the local precommit passes for all 23 changed files. The full review-thread set remains resolved; CI is rerunning on the rebased head.

DCM-Uni-Paderborn:
Rebased the five lattice-FFT commits onto the newly conflict-resolved #5825 head. This keeps the stacking explicit, retains the ADIIS integration from current master, and avoids duplicating the shared low-rank conflict resolution. Precommit passes for all 23 PR files, and GitHub now reports the PR as mergeable. The existing review decision remains in place.

fstein93 (approved):

dryndyk (commented):
I'm not sure the changes in `negf_integr_cc.F` are consistent with the changes made in #5868. I'll check everything after the commits. )

fstein93:
> I'm not sure the changes in `negf_integr_cc.F` are consistent with the changes made in #5868. I'll check everything after the commits. )

They are not. I harmonized the naming scheme because the naming scheme was generally not consistent. The main problem is that we need a routine to perform an FFT from a pre-optimized plan and one without for convenience. The latter is problematic anyways with in-place FFTs because there is no way to tell in-place and out-of-place-FFTs apart within Fortran from the same function interface without using `POINTER`s and it is undefined behavior if you pass the same variable as a output-variable more than one time to a procedure (compare convention violation in the FFTW-wrapper).
Anyways, I will fix just overwrite the change anyways with a subsequent PR because I intend to hide the planning step with the new backend.

DCM-Uni-Paderborn:
Thanks, agreed. I checked this branch explicitly against #5868. The lattice algorithm itself does not depend on `EXTENDED_FFT_LENGTHS`; unsupported short FFTSG lattice dimensions are embedded internally into compatible exact transforms.

The direction-generic `fft_1d` routine in this branch is only a small temporary adapter needed for forward and backward batched transforms on current `master`. #5868 replaces that low-level interface with `fft_1d_many`. Once #5868 lands, I will rebase this caller onto the new interface and drop the temporary adapter; the block-circulant construction and numerical behavior remain unchanged. I have documented that boundary in the PR description.

DCM-Uni-Paderborn:
Rebased the complete linear stack onto current `origin/master` (`7518516c65`) and force-updated the branch with an exact lease. The five lattice-specific patches are patch-identical to the previous revision; the only rebase conflict was the additive CMake unit-test registration, where both the new upstream target and this PR target were retained.

Fresh local validation on the rebased head:

- full GCC 16 Release build with MPI/OpenMP;
- `kpoint_lattice_fft_unittest` with 1, 2, and 4 MPI ranks;
- `low_rank_preconditioner_unittest` with 1, 2, and 4 MPI ranks;
- CP2K precommit: 23/23 files;
- complete `QS/regtest-ot-2`: 50/50 matchers with MPI-2.

The PR description has also been rewritten to clarify the lattice-FFT mathematics, opt-in/default behavior, positive-definiteness safeguards, current gather implementation, performance scope, and the planned #5868 interface handoff.

PR #5831, opened 2026-08-23
OT: Add rotation-covariant FULL_ALL preconditioner

DCM-Uni-Paderborn:
## Summary

This PR adds the opt-in `FULL_ALL_COVARIANT` OT preconditioner for real Gamma-point calculations. It retains the complete state-selective frozen-Hamiltonian approximation of `FULL_ALL`, but makes its action covariant under rotations of the occupied orbitals. This allows the preconditioner to be used consistently with `ROTATION` without changing the physical orbital gauge.

All existing preconditioners and defaults remain unchanged.

## Usage

```text
&OT
  PRECONDITIONER FULL_ALL_COVARIANT
  ROTATION T
&END OT
```

The new option currently requires uniform occupations, excludes ROKS, and uses `PRECOND_SOLVER DEFAULT`.

## Motivation and construction

Legacy `FULL_ALL` uses an occupied-state energy for each orbital column. Its action therefore depends on the chosen occupied-orbital representation and is not compatible with `ROTATION`.

`FULL_ALL_COVARIANT` diagonalizes only a private copy of the occupied Hamiltonian block,

```math
C^{\mathsf T} H C
= U_C\,\text{diag}(\varepsilon_i) U_C^{\mathsf T},
```

transforms the input to that canonical occupied gauge, applies the complete state-dependent inverse there, and transforms the result back:

```math
\widetilde G = G U_C,
\qquad
\widetilde Y_i = \mathcal L_{\varepsilon_i}^{-1}\widetilde G_i,
\qquad
\mathcal P_C[G] = \widetilde Y U_C^{\mathsf T}.
```

The frozen-Hamiltonian denominators are

```math
\max\!\left(\mathtt{ENERGY_GAP},\,\varepsilon_a-\varepsilon_i\right),
```

which preserves the existing gap floor and keeps the applied inverse positive definite. For every orthogonal occupied-space rotation `R`, the full construction obeys

```math
\mathcal P_{C R}[G R] = \mathcal P_C[G] R.
```

The private canonicalization never overwrites the physical MO coefficients or their derivatives. Both the distributed full-matrix and DBCSR application paths rotate inputs and outputs consistently.

## Cost and expected use

This is primarily a correctness and robustness option for rotation-enabled OT, not a general speed-up. It has the spectral quality and construction cost of `FULL_ALL`, plus two occupied-space right rotations per application. A simple scaled overlap inverse would be cheaper and rotation invariant, but would discard the state-selective spectral information that this implementation preserves.

The option is therefore deliberately not a new default.

## Validation

The mathematical unit test checks:

- agreement with an explicit spectral reference;
- inversion of a synthetic frozen-Hamiltonian Sylvester operator;
- covariance under a nontrivial occupied rotation;
- invariance under rotations inside a degenerate occupied eigenspace;
- positive definiteness with an active denominator floor.

Validation after rebasing onto current `master`:

- complete GCC 16 / OpenMPI / OpenMP Release build;
- unit test with 1, 2, and 4 MPI ranks;
- `QS/regtest-ot-2`: 49/49 correct with 2 MPI ranks;
- no-cache precommit: 16/16 changed files clean;
- two linear commits without merge commits.

The added H2O regression exercises `FULL_ALL_COVARIANT` together with `ROTATION` and checks the final total energy.

## Scope

This PR only introduces the real Gamma-point covariant `FULL_ALL` variant. It does not modify Mermin free-energy optimization, metallic occupations, `ADDED_MOS`, complex k-point channels, restart, symmetry, forces, or stress. Those coupled OT-for-metals paths remain separate.

No FFT approximation is added here. A lattice-FFT acceleration would require a shared translationally invariant or block-circulant operator representation and belongs in the separate lattice-FFT preconditioner work.


DCM-Uni-Paderborn:
Rebased onto the current master after #5844. The SCF conflicts were resolved by retaining the ADIIS validation and adding the FULL_ALL_COVARIANT imports and state checks. Precommit passes for all 16 PR files, and GitHub now reports the PR as mergeable.

DCM-Uni-Paderborn:
Rebased onto current master 7518516c65. The only conflict was the additive unit-test registration in src/CMakeLists.txt; both the upstream frontier-orbital target and the covariant-preconditioner target were retained. The new head is 68d62fec8a with two linear commits and no merge commits. Local validation passes: complete GCC 16 MPI/OpenMP Release build, unit test with MPI 1/2/4, no-cache precommit 16/16, and QS/regtest-ot-2 49/49 with MPI 2. The PR description has also been updated to document usage, covariance, cost, safeguards, and scope.

oschuett:
@marci73, we have 4 PRs blocked on your review. Could you please take a look.

oschuett:
Btw, if you find your review load too high, then we could try to move some code around so that we can narrow [your code-ownership](https://github.com/cp2k/cp2k/blob/2ac5cc1a2492f2aae749098b8be0b482f271b8cc/CODEOWNERS#L72) to the parts that you truly care about.

juerghutter (approved):

PR #5833, opened 2026-08-23
OT: Extend L-BFGS to orbital rotations

DCM-Uni-Paderborn:
## Summary

This follow-up to #5832 extends the optional OT L-BFGS minimizer from orbital-only coordinates to the fixed strict-OT product chart of orbital and occupied-space rotation variables. It also permits `OCCUPATION_PRECONDITIONER` with L-BFGS.

The existing `CG` default and all established minimizers remain unchanged.

## Implementation

The L-BFGS vector is extended from the orbital block `X` to `(X, Omega)`, where `Omega` is the antisymmetric orbital-rotation generator. Product-space scalar products use the same metric as the existing OT SD and CG implementations:

```text
<(X, Omega), (Y, Xi)> = Tr(X^T Y) + 1/2 Tr(Omega^T Xi).
```

The implementation:

- stores rotation positions and gradients alongside the orbital L-BFGS history;
- forms joint `(s,y)` pairs across orbital, rotation, spin and MPI-distributed blocks;
- includes the rotation metric in curvature tests, both two-loop recursions and descent checks;
- applies the selected physical OT preconditioner to the orbital block of `H0`;
- uses the identity for the rotation block of `H0`, consistently with the existing SD and CG minimizers;
- includes the independent antisymmetric rotation variables in the reported convergence norm;
- retains curvature rejection, history reset and preconditioned/raw-gradient fallbacks from #5832.

`OCCUPATION_PRECONDITIONER` changes the orbital-gradient scaling before it reaches the minimizer and introduces no additional optimization variable. The previous L-BFGS input rejection was therefore unnecessary and is removed.

## Supported scope

- `ALGORITHM STRICT` remains required, so all secant pairs stay in one fixed product chart.
- `ROTATION` is now supported with any otherwise rotation-compatible OT preconditioner.
- `OCCUPATION_PRECONDITIONER` is now supported.
- The pre-existing incompatibility between legacy `FULL_ALL` and `ROTATION` is unchanged.
- `ALGORITHM IRAC` remains outside this PR because its moving reference chart requires an explicit tangent-vector transport.
- `ENERGIES` remains outside this PR; that path is globally marked as not fully implemented in the existing OT input code.

## Validation

- complete GCC16/OpenMPI/OpenMP Release build;
- official MPI-2 `QS/regtest-ot-2`: 37/37 in 21.32 s;
- new spin-polarized orbital-dependent SIC regression with nonzero rotation forces and `OCCUPATION_PRECONDITIONER`: convergence in 17 steps at `-5.30617976133315 Ha`;
- MPI-1, MPI-2 and OpenMP-4 reproduce 17 steps and agree within `2e-14 Ha`;
- no-cache precommit: 5/5 files;
- one linear commit on current `master`, without merge commits.

## Scope separation

This PR changes only the optional L-BFGS minimizer and its regression coverage. It contains no Mermin functional, metallic occupation optimization, `ADDED_MOS`, complex k-point support, or new restart, symmetry, force or stress handling. It has no dependency on the separate metallic-OT work.


juerghutter:
Is OT L_BFGS compatible with the CONJUGATE GRADIENTS implementation in CP2K.
Changing the preconditioner will require to explicitely orthogonalize search directions.

DCM-Uni-Paderborn:
Thanks, Jürg. In the current implementation, `MINIMIZER CG` and `MINIMIZER LBFGS` are mutually exclusive branches in `ot_mini`. L-BFGS is not used as a changing preconditioner inside the conjugate-gradient implementation, so there are no retained CG search directions that need explicit reorthogonalization here. I agree that this would be required for a changing preconditioner within CG.

Your comment did expose a separate issue with the simultaneous removal of the `OCCUPATION_PRECONDITIONER` restriction. That option rescales the orbital gradient before it reaches the minimizer, so the current L-BFGS path would store the scaled gradient in its secant history. This is not equivalent to applying the preconditioner only as `H0`, even though it introduces no additional optimization variable.

I opened #5836 to restore that restriction while keeping `LBFGS` + `ROTATION` unchanged. Proper support should retain physical gradient differences in the secant history, apply occupation scaling as preconditioning, and reset the history if that preconditioner changes.

If you were referring to a different interaction with the CG implementation, please let me know.

PR #5844, opened 2026-08-24
Implement ADIIS algorithm for diagonalization

Growl1234:
This PR adds the [Augmented DIIS (ADIIS)](https://dx.doi.org/10.1063/1.3304922) method as an alternative update strategy for diagonalization-based SCF calculations.

ADIIS is selected through:

```text
&SCF
  &DIAGONALIZATION
    UPDATE_METHOD ADIIS
  &END DIAGONALIZATION
&END SCF
```

For some strongly correlated magnetic systems involving transition metal elements, ADIIS can converge cases where the common `BROYDEN_MIXING` struggles or fails. However, convergence is often relatively slow, typically requiring around 60–70 SCF iterations, with particularly poor efficiency close to the convergence threshold. For systems that are already handled well by conventional density mixing, the current ADIIS implementation is generally not a competitive choice.

Here is a testing example with P2-NaMnO2: [NaMnO2-test.zip](https://github.com/user-attachments/files/31379587/NaMnO2-test.zip)

`UPDATE_METHOD MIXING` remains the default and preserves the existing density-mixing SCF path.

### Implementation

The ADIIS path:

- maintains a history of paired density and KS/Fock matrices;
- constructs the augmented Roothaan–Hall quadratic model from

$$
T_{ij} = \mathrm{Tr}(P_i F_j)
$$

- minimizes the model under the simplex constraints

$$
c_i \ge 0, \qquad \sum_i c_i = 1
$$

- constructs the effective KS matrix

$$
F_{\mathrm{ADIIS}} = \sum_i c_i F_i
$$

- diagonalizes this matrix and directly accepts the resulting density.

The implementation supports spin-polarized and k-point calculations with a single set of ADIIS coefficients shared by the complete electronic state. The current implementation targets integer-occupation diagonalization SCF, i.e. smearing is not supported.

ADIIS is independent of `&MIXING`; the existing mixing and conventional DIIS paths are unchanged.

### History management

ADIIS uses a dedicated history size controlled by `SCF%ADIIS%MAX_HISTORY` with a default of 7.

The initial arbitrary SCF guess is excluded from the ADIIS history. The history manager retains useful previous states while avoiding purely FIFO replacement, and can restart the ADIIS subspace when it becomes stagnant without falling back to density mixing.

### Convergence

The ADIIS density step is not assumed to be a sufficient SCF convergence criterion. A raw KS/Fock validation step is used before declaring convergence, preventing false convergence when an interpolated historical Fock matrix reproduces the current density.

### TODO in the future

- Consider [ODA](https://dx.doi.org/10.1021/acs.jpca.5c02110) as a reasonable approach to stabilizing SCF iterations far from convergence.
- Allow ADIIS -> (C)DIIS transition to speed up convergence close to the threshold.
- Support smearing.


garrettwx:
Hi, thank you for your implement of this method for convergence,
I also perform the test from the existed KERKER_MIXING, and it also possesses the good convergence for magnetic systems.
And even for some previously difficult systems, e.g. Ni(111), also works. 
Actually, the BROYDEN_MIXING or MODIFIED_BROYDEN_MIXING are both difficult to converge for these specific systems.

The KERKER_MIXING coupled with current ALPHA, ALPHA, BETA, and BETA_MAG tags can now work for convergence.

[NaMnO2_out.txt](https://github.com/user-attachments/files/31381813/NaMnO2_out.txt)

DCM-Uni-Paderborn (approved):
Thank you for this implementation. I went through the ADIIS equations and their translation into the SCF update. The construction of the ADIIS objective, the use of matching density and Fock matrix histories, and the treatment of the two spin channels look correct to me. 
However, in my own tests, I couldn't identify any advantage either. This does not invalidate ADIIS or indicate an implementation problem. Instead, it shows that ADIIS should be regarded as an additional, complementary option rather than the only solution for such systems. 
Pure ADIIS is known to be less effective close to tight convergence, so a future ADIIS-to-DIIS switching strategy would likely make it more useful in practice. However, that does not need to be part of this PR. Since the new method is opt-in and does not alter the existing defaults, I consider the current implementation suitable for merging.

PR #5848, opened 2026-08-25
Honor disabled motion output in SWARM runs

DCM-Uni-Paderborn:
The ARM64 dashboard intermittently exceeded the 2 MiB regtest output limit in `SWARM/regtest-glbopt-1/LJ10_mincrawl_1.inp`. The stochastic MINCRAWL path can take many iterations, and the regtest limit counts every file created in the work directory.

This change:

- makes the per-step MD banner honor `MD/PRINT/PROGRAM_RUN_INFO`;
- avoids final-structure log messages when `FINAL_STRUCTURE` is disabled;
- disables the unused SWARM communication log in this non-restart smoke test.

It does not change `MAX_ITER`, `E_TARGET`, the reference energy, or the matcher tolerance.

On ARM64 with 2 MPI ranks and 2 OpenMP threads, an intentionally forced 500-iteration run dropped from over 6 MiB to 0.69 MiB. The complete `SWARM/regtest-glbopt-1` directory passes 3/3. A normal MD run still emits its step banner, and an enabled `FINAL_STRUCTURE` still writes both XYZ and CIF output.


oschuett:
Personally, I wouldn't mind to rip out the SWARM code entirely as ASE is a much better choice for [global optimization](https://docs.ase-lib.org/ase/optimize.html#global-optimization).

DCM-Uni-Paderborn:
The repeated OpenMPI failures are unrelated to the SWARM changes: I reproduced the same `sm`-transport segfault with a standalone two-rank MPI stress test. Concurrent launches sharing OpenMPI's backing directory gave 9,996/10,000 clean exits, two segfaults, and two companion timeouts; isolating the backing directory per launch gave 10,000/10,000 clean exits.

The focused regtest-driver fix is #5854. Once that is merged, I will update this branch from master and rerun the required OpenMPI check.

PR #5851, opened 2026-08-25
DIAG: Add ELPA support for complex Hermitian matrices (k-points)

s8ga:
## Summary

Adds an opt-in ELPA eigensolver for complex Hermitian matrices (k-point calculations), mirroring the existing real-symmetric wrapper (`src/fm/cp_fm_elpa.F`),  plus a one-time runtime correctness check that keeps `ELPA_COMPLEX_KERNEL AUTO` safe (marekandreas/elpa#77).

## Reason
`PREFERRED_DIAG_LIBRARY ELPA` was silently ignored for complex matrices: `cp_cfm_heevd` always fell back to ScaLAPACK (`pzheevd`) regardless of the setting.

## Changes
- New `src/fm/cp_cfm_elpa.F`: self-contained wrapper module mirroring the   real-side structure, including `check_elpa_c_kernel_correctness`: a one-time setup-time validation of the resolved complex BLOCK2 kernel (eigenpair residual + eigenvector orthogonality; geometry aligned with the https://github.com/marekandreas/elpa/issues/77 reproducer) that reuses the production solve path and falls back to ScaLAPACK with a warning on failure.
- New `cp_cfm_redistribute_start/end` in `cp_fm_diag_utils.F`: complex counterparts using `pzgemr2d` (mirrors the real-side `pdgemr2d` logic).
- `cp_cfm_diag.F`: routing in `cp_cfm_heevd` — complex matrices use ELPA only when the library was requested explicitly (default remains ScaLAPACK)
- New `ELPA_COMPLEX_KERNEL` keyword (16-value enum, default AUTO with cpuid resolution); `ELPA_KERNEL` description narrowed to real matrices.
- `diag_lib_explicit` flag in `cp_fm_diag.F` (set from `section_vals_val_get(..., explicit=)`) gates the complex-side ELPA path.

## Validation
- Passed `make_pretty.sh`
- Reproduced the GCC 15.2 BLOCK2 mis-compilation (https://github.com/marekandreas/elpa/issues/77) in an Ubuntu 26.04 container: the check detects it, warns, and falls back to ScaLAPACK with results unchanged; healthy kernels are unaffected.
- Three new regression tests in `QS/regtest-elpa/` covering explicit `PREFERRED_DIAG_LIBRARY ELPA` with 2-stage GENERIC, 1-stage, and 2-stage BLOCK2 (the latter exercising the correctness check).

## Scope

- Complex matrices use ELPA only when explicitly requested; the default remains ScaLAPACK.
- QR decomposition is not available for complex matrices in ELPA and is therefore not implemented here.
- `ELPA_COMPLEX_KERNEL` defaults to AUTO (cpuid); forcing a kernel unsupported by the CPU crashes inside ELPA.


Growl1234:
Generally I would think this is very nice. However, please note that `ELPA_2STAGE_COMPLEX_AVX*_BLOCK2` kernels will perform totally wrong results due to [GCC-15.2's regression bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=122408); related ELPA-side issue is https://github.com/marekandreas/elpa/issues/77. Since we are now using Ubuntu 26.04 for most of CI testers, which is using exactly the GCC version 15.2, the regtest results ~will be not able to be correctly validated by our CI~.

fstein93 (approved):

Growl1234 (changes requested):
Could you please add a runtime validation for ELPA complex BLOCK2 kernels, which is somewhat similar to [what I have done for ABACUS](https://github.com/deepmodeling/abacus-develop/pull/7790)? If an inconsistency is detected then it should throw out a warning and fall back to ScaLAPACK. Although the tests didn't go through the paths, we should definitely be aware of this.

s8ga:
> Generally I would think this is very nice. However, please note that `ELPA_2STAGE_COMPLEX_AVX*_BLOCK2` kernels will perform totally wrong results due to [GCC-15.2's regression bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=122408); related ELPA-side issue is [marekandreas/elpa#77](https://github.com/marekandreas/elpa/issues/77). Since we are now using Ubuntu 26.04 for most of CI testers, which is using exactly the GCC version 15.2, the regtest results will be not able to be correctly validated by our CI.


The regression tests use ELPA_COMPLEX_KERNEL GENERIC (cplx_kp.inp) and the 1-stage solver (cplx_kp_1stage.inp) respectively, both of which bypass the mis-compiled path.

> Could you please add a runtime validation for ELPA complex BLOCK2 kernels, which is somewhat similar to https://github.com/deepmodeling/abacus-develop/pull/7790? If an inconsistency is detected then it should throw out a warning and fall back to ScaLAPACK. Although the tests didn't go through the paths, we should definitely be aware of this.

Thank you for pointing this out, we should defintely implement this sanity check for ELPA.

DCM-Uni-Paderborn:
Thanks for adding this. I built commit 95f51ef86b with ELPA 2026.02.001 and tested the complex 1-stage and 2-stage paths with 1, 2, 4, 8, and 16 MPI ranks, including redistribution, and compared them against ScaLAPACK. All supported paths agreed to about 2.2e-14 Ha. I also confirmed that the implicit default continues to use ScaLAPACK for complex matrices.

I agree with SY's correctness concern and consider it blocking before merge. ELPA_COMPLEX_KERNEL AUTO currently selects the affected AVX/AVX2 BLOCK2 kernels. With an ELPA library built using GCC 15.2 these kernels can return ELPA_OK and correct eigenvalues while producing severely incorrect and non-orthogonal eigenvectors.

Please make AUTO safe before merging. A one-time residual and orthogonality validation with fallback to the corresponding BLOCK1 kernel or ScaLAPACK would be robust. A simpler acceptable alternative would be to use BLOCK1 for AUTO until the affected ELPA/GCC combinations can be identified safely. Checking eigenvalues alone would not detect this failure.

Apart from this issue, I did not find another correctness problem in the implementation.

Growl1234:
Hi Thomas, just a note that you might have mentioned the wrong person😉

s8ga:
Commit 332e6a2 addresses the blocking AUTO concern.


## What was added
`check_elpa_c_kernel_correctness`: a one-time check at setup time (after
`diag_init`). Whenever the resolved complex kernel is a BLOCK2 variant and
the 2-stage solver is active, it solves a small deterministic dense random
Hermitian eigenproblem (na=20, nev=18, nblk=8 — the geometry of the
standalone reproducer of marekandreas/elpa#77) on a single rank through the
same production routine used for the actual diagonalizations
(`cp_cfm_diag_elpa_base`), then verifies eigenvector orthogonality and
eigenpair residuals (threshold 1e-8). As you noted, eigenvalues alone
cannot detect this failure, so they are not used as a criterion. On
failure it warns (reporting the measured errors) and complex matrices fall
back to ScaLAPACK for the rest of the run, with all ranks agreeing on the
outcome. 1-stage runs and non-BLOCK2 kernels skip the check entirely.

## How it was tested
- Healthy build: the check passes silently, ELPA remains in use, results
  unchanged.
- Reproduced the mis-compilation in an Ubuntu 26.04 container with GCC
  15.2 and ELPA 2026.02.002 (CP2K toolchain build). The standalone
  reproducer from the issue fails there exactly as reported (residual
  5.9e-2, orthogonality 1.5); note the issue tested 2026.02.001 only —
  2026.02.002 is affected as well. Our check triggers on the same library
  (orthogonality 2.0e-2, residual 1.9e-1 for its own random matrix), warns
  before the first SCF step, and falls back to ScaLAPACK; total energies
  stay identical to the ScaLAPACK reference. Selecting GENERIC on the same
  build runs through ELPA without any warning.
- Side observation: the corruption is blocking-sensitive — nblk=8 (the
  reproducer geometry) is affected, while nblk=16 production-shaped
  matrices happened to survive on the same build; ELPA's own test-suite
  defaults (na=500, nblk=16) may therefore miss it. This is why the check
  deliberately uses nblk=8.
- New regression test `cplx_kp_block2.inp` (explicit ELPA + AVX2_BLOCK2 +
  2-stage) exercises the check in every regtest run; the explicit-ELPA
  directories and the local suite pass;
  

<details>
<summary>1. The bug is present in this build: standalone reproducer (marekandreas/elpa#77) — AVX/AVX2 BLOCK2 FAIL with residual 5.86e-02 / orthogonality 1.506, matching the issue's reference values to 13 significant digits; eigenvalues identical across all kernels</summary>

```text
$ mpirun -np 1 ./elpa_block2_reproducer          # na=20, nev=18, nblk=8, seed 0, 1 rank, 1 thread
GENERIC_SIMPLE   residual=3.4431118401649132e-16 orthogonality=5.2161673284400491e-15 first_ev=-2.3499043557075967e+00 last_ev=9.3138605660914819e-01 result=PASS
SSE_BLOCK2       residual=3.6268460512983282e-16 orthogonality=5.7632827914706713e-15 first_ev=-2.3499043557075967e+00 last_ev=9.3138605660914819e-01 result=PASS
AVX_BLOCK1       residual=2.9944003133232615e-16 orthogonality=5.4437108424006545e-15 first_ev=-2.3499043557075967e+00 last_ev=9.3138605660914819e-01 result=PASS
AVX_BLOCK2       residual=5.8583844263629517e-02 orthogonality=1.5064012687592581e+00 first_ev=-2.3499043557075967e+00 last_ev=9.3138605660914819e-01 result=FAIL
AVX2_BLOCK1      residual=3.2657180759309660e-16 orthogonality=5.1729392971779788e-15 first_ev=-2.3499043557075967e+00 last_ev=9.3138605660914819e-01 result=PASS
AVX2_BLOCK2      residual=5.8583844263629496e-02 orthogonality=1.5064012687592581e+00 first_ev=-2.3499043557075967e+00 last_ev=9.3138605660914819e-01 result=FAIL
```

</details>

<details>
<summary>2. Our check fires at init (before SCF) and falls back to ScaLAPACK: AVX2_BLOCK2 caught with orthogonality 2.03e-02 / residual 1.86e-01, warning printed ~590 lines before `SCF WAVEFUNCTION OPTIMIZATION`</summary>

```text
$ mpirun -np 2 -x OMP_NUM_THREADS=1 cp2k.psmp -i cplx_kp_block2.inp
 GLOBAL| Diagonalization library                                            ELPA
 GLOBAL| Minimum number of eigenvectors for ELPA usage                         4
 GLOBAL| Orthonormality check for eigenvectors enabled                 1.000E-14
 GLOBAL| Total number of message passing processes                             2
 GLOBAL| Number of threads for this process                                    1
 GLOBAL| CPU model name                           Intel(R) Core(TM) Ultra 7 265K

 MEMORY| system memory details [Kb]
 MEMORY|                        rank 0           min           max       average
 MEMORY| MemTotal             51192856      51192856      51192856      51192856
 MEMORY| MemFree              41566796      41566796      41566796      41566796
 MEMORY| MemLikelyFree        48046904      48046904      48046904      48046904

 ELPA| Matrix diagonalization information
 ELPA| Matrix order (NA)                                                      20
 ELPA| Matrix block size (NBLK)                                                8
 ELPA| Number of eigenvectors (NEV)                                           18
 ELPA| Kernel                                                        AVX2_BLOCK2

 *** WARNING in fm/cp_cfm_elpa.F:358 :: The ELPA complex kernel          ***
 *** AVX2_BLOCK2 failed a runtime correctness check (orthogonality:      ***
 *** 2.0322E-02, residual:  1.8554E-01) and may return incorrect         ***
 *** eigenvectors. Complex matrices fall back to ScaLAPACK. Consider the ***
 *** GENERIC kernel or rebuilding ELPA with -fno-tree-slp-vectorize.     ***

 *** Fundamental physical constants (SI units) ***        <- regular startup banner follows
 [... SCF WAVEFUNCTION OPTIMIZATION starts at line 673 of the output file ...]

# all 12 complex diagonalizations routed to ScaLAPACK:
 cp_cfm_heevd_base                    12  8.7    0.005    0.005    0.005    0.005
 ENERGY| Total FORCE_EVAL ( QS ) energy [hartree]            -41.897080363583584
```

</details>

<details>
<summary>3. GENERIC control on the same build: no warning, ELPA in use, energy bit-identical</summary>

```text
$ mpirun -np 2 -x OMP_NUM_THREADS=1 cp2k.psmp -i cplx_kp_block2_GENE.inp
 cp_cfm_diag_elpa                     12  8.7    0.000    0.000    0.006    0.006
 ENERGY| Total FORCE_EVAL ( QS ) energy [hartree]            -41.897080363583584
```

</details>

<details>
<summary>4. Environment & how to reproduce (Ubuntu 26.04 container, GCC 15.2, ELPA 2026.02.002 — issue tested .001 only, .002 is affected as well)</summary>

```text
gcc (Ubuntu 15.2.0-16ubuntu1) 15.2.0
Package elpa_openmp version: 2026.02.002        # built with the CP2K toolchain
CP2K source code revision number:  working tree
Container: docker.io/library/ubuntu:26.04
Host CPU: Intel(R) Core(TM) Ultra 7 265K (AVX2)

# reproduce inside the container:
source /root/cp2k/install/cp2k_env
cd /root/reproducer && mpirun --allow-run-as-root -np 1 ./elpa_block2_reproducer
cd /tmp && mpirun --allow-run-as-root -np 2 -x OMP_NUM_THREADS=1 cp2k.psmp \
  -i /root/cp2k/tests/QS/regtest-elpa/cplx_kp_block2.inp
```

</details>


## Effect

AUTO is safe by construction: on affected ELPA/GCC combinations complex
ELPA use is disabled with a clear warning instead of silently returning
wrong eigenvectors, while healthy builds keep full ELPA performance (the
check costs a single 20×20 diagonalization at startup).

DCM-Uni-Paderborn (approved):
Thank you for the thorough follow-up. I reviewed commit 332e6a2 and retested the updated implementation with ELPA 2026.02.001 using 1, 2, 4, and 16 MPI ranks.

The one-time check exercises the production ELPA path, validates both orthogonality and eigenpair residuals, and ensures a consistent fallback decision across all ranks. I also verified the healthy-kernel path and the fallback from an unavailable AVX2 kernel to GENERIC. All energies remain consistent with ScaLAPACK to about 2.2e-14 Ha.

This fully addresses my concern. I have no further objections.

Growl1234 (approved):

PR #5852, opened 2026-08-25
Implement ODA for ADIIS

Growl1234:
This PR implements an optional safeguarded optimal damping algorithm (ODA) for the ADIIS SCF update method.

ADIIS requires a useful history before it can form an effective Fock-space combination. During startup, or immediately after an ADIIS history restart, it can therefore reduce to an undamped raw Roothaan step. These initial steps can produce large density and energy oscillations in difficult spin-polarized systems.

ODA is introduced as a globalization mechanism for these raw steps:

$$
P_1 = D\left(F[P_0]\right),
\qquad
P_\lambda = P_0 + \lambda\left(P_1-P_0\right).
$$

The time cost of single ODA step is usually 2-3x of a general SCF step.

### Implementation

For an eligible raw Roothaan step, the implementation:

- constructs the density direction $\Delta P=P_1-P_0$;
- evaluates the initial directional derivative

$$
g_0 = \mathrm{Tr}\left(F[P_0]\Delta P\right).
$$

- evaluates a damped trial density and its corresponding energy and raw Fock matrix;
- uses a cubic Hermite model to propose smaller steps when necessary;
- accepts a step only after the actual trial energy satisfies a safeguarded Armijo condition;
- restores and evaluates the full raw endpoint if no acceptable interior step is found.

The cubic model is therefore only a step proposal. It cannot accept a density without an explicit energy evaluation.

The accepted damping parameter from a consecutive ODA step is reused as the next predictor. This normally limits successful consecutive searches to one additional KS evaluation. Fully evaluated trial states are also carried forward without an unnecessary density update or duplicate commit.

For k-point calculations, the density direction and Fock contraction use the existing real-space image-cell representation indexed by spin and image cell. The k-point weights are already included by the k-to-real-space density transformation. A single global $\lambda$ is used for all spin channels and k-points.

ODA is used only:

* during the initial raw-Roothaan phase;
* after an explicit ADIIS history restart.

Once ADIIS constructs a genuine combined Fock matrix, ODA is disabled until another restart. ODA trial states are never inserted into the ADIIS history; only accepted and consistently evaluated density/Fock states are eligible.

### Convergence and output

An interior ODA density is generally non-idempotent and does not have a matching set of raw-endpoint MOs. Therefore:

- an ODA state cannot by itself declare SCF convergence;
- the existing raw-map `ADIIS/Chk.` validation remains the final convergence safeguard;
- iteration-local MO properties and Gamma-point wavefunction restarts are suppressed for interior ODA states;
- k-point density restarts remain available.

The regular SCF table reports `ODA/Diag.`, `ODA/Rst.`, or `ODA/Fallback`. Additional ADIIS/ODA metrics are printed only when `SCF%ADIIS%VERBOSE` is enabled.

### Input

ODA is disabled by default; you can enable it for supported ADIIS calculations:

```text
&SCF
  &DIAGONALIZATION
    ALGORITHM STANDARD
    UPDATE_METHOD ADIIS
  &END DIAGONALIZATION
  &ADIIS
    ODA T
    ! VERBOSE
  &END ADIIS
&END SCF
```

Some features such as ACE in HFX does not support ODA, in which case it will be automatically set off with a warning message.


garrettwx:
Hi, I want to consult that if the DAVIDSON/LANCZOS method within the diag deserves the further implementation of kpoints. The current use of your implementation seems to have been suitable for the magnetic systems, also for KERKER_MIXING. Maybe METHOD STANDARD is sufficient.

Growl1234:
> Hi, I want to consult that if the DAVIDSON/LANCZOS method within the diag deserves the further implementation of kpoints.

I might investigate it in the future; currently the alternative diag methods for k-points are not on my to-do list, but I agree implementing it for Davidson method would possibly be very nice.

> The current use of your implementation seems to have been suitable for the magnetic systems, also for KERKER_MIXING.

There's still one step that's not done: ADIIS -> (C)DIIS transition. I expect that step will speed up the convergence. I once tried to implement such transition to mixing methods besides `DIRECT_P_MIXING`, but several attempts failed. That said, I was surprised that `KERKER_MIXING` can work for those systems.

Growl1234:
Implementing Davidson for k-points might be non-trival as it uses preconditioner. 🤔

garrettwx:
> Implementing Davidson for k-points might be non-trival as it uses preconditioner. 🤔

To my view, since the current methods have been suitable, the DAVIDSON for k-points is not very advantageous compared with STANDARD.

DCM-Uni-Paderborn (changes requested):
Thank you for this substantial and carefully documented implementation. I built the branch locally and ran the new ADIIS/ODA test cases, including the k-point variants. The overall design looks promising, but I found two correctness issues that should be addressed before merging.

1. Premature convergence after a failed raw-map validation

In qs_scf.F, around lines 689-697, the code is already in the branch where the raw-map validation has failed, i.e. adiis_raw_delta >= EPS_SCF. It then computes another accelerated ADIIS candidate and sets adiis_validated based on the size of that accelerated step.

This defeats the purpose of the raw-map safeguard: a small accelerated step does not imply that the raw SCF map is converged.

I can reproduce this with the existing bn.inp, changing only:

  EPS_SCF 3.0E-6

At iteration 19, CP2K reports ADIIS/Chk. and convergence with:

  step_norm     = 1.160E-6
  raw_map_delta = 3.698E-6

Thus the reported raw-map residual is still above EPS_SCF. The same problem is reproducible with ODA OFF, so this is a regression in the general ADIIS validation logic rather than an ODA-specific corner case.

The minimal fix appears to be to leave adiis_validated false in this branch and restore:

  scf_env%adiis_check_next = scf_env%iter_delta < scf_control%eps_scf

With this change, the same calculation performs a genuine raw-map check in iteration 20 and converges with raw_map_delta = 6.893E-7.

A regression test that explicitly exercises a failed validation followed by a small accelerated step would be useful here.

2. ODA AUTO should currently exclude active HFX/ACE

The current oda_eligible condition allows ADIIS calculations with active ACE. However, an interior ODA density deliberately has no matching MO set, while ACE constructs its projector from the current MO coefficients. In addition, every ODA trial evaluation advances or rebuilds ACE's module-level projector state and step counter.

Consequently, the trial energies and Fock matrices used by the Armijo test do not necessarily correspond to one fixed, density-consistent functional.

Using the existing all-electron H2O ACE input with ADIIS and an atomic guess, ODA accepted lambda = 0.5 in iteration 2. At the accepted trial density, the existing ACE diagnostics gave:

  E_x(ACE) = -2.2589855520 Ha
  E_x(full) = -2.2279267285 Ha
  difference = 0.0310588 Ha

The safest initial solution would be to make ODA ineligible whenever an active HF/ACE section is present. Supporting this combination later would require evaluating the ODA trials with full HFX, or otherwise managing the ACE projector and counters so that the line search uses a consistent energy and derivative.

Apart from these two points, the branch builds cleanly for me and the Gamma-point and k-point ODA/ODA-OFF test inputs complete successfully.

Growl1234:
Thanks, I'll carefully revise them tomorrow.

Growl1234:
Regarding the first point, this is intentional in order to not waste time on the convergence area, as usually at that status the change in energy is already 1e-9~1e-10 (assuming EPS_SCF to be 1e-5). The check was initially introduced to avoid "false zero" immediately after atomic guess, but was preserved as the last safe guard while the false convergence was prevented through other ways; iiuc only when `step_norm` is lower than 1% of `raw_map_delta` does ADIIS continue the SCF cycle

Growl1234:
I think, maybe I should make ODA opt-in rather than opt-out.

Growl1234:
ODA is now disabled for ACE of HFX.

DCM-Uni-Paderborn (changes requested):
Thank you for addressing the ACE issue and for making ODA opt-in. I consider
that part resolved.

However, the first correctness issue remains in the current head
(315fa0cd25). I rebuilt and retested the branch on Spark with four MPI ranks.
For bn.inp with EPS_SCF 3.0E-6, CP2K still reports:

  ADIIS metrics: step_norm=1.226E-06 raw_map_delta=3.914E-06
  *** SCF run converged in 19 steps ***

Thus convergence is accepted although the raw-map validation that triggered
this branch has failed.

The 1% stagnation heuristic limits how different the previous ADIIS step may
be from the raw-map delta, but it does not establish that the newly generated
accelerated density is a fixed point of the raw SCF map. In particular,
lines 710-718 set adiis_validated from another accelerated step and then clear
adiis_check_next, so that new density is never subjected to a raw-map check
before convergence is declared.

Please leave adiis_validated false after a failed raw-map validation and set

  scf_env%adiis_check_next = scf_env%iter_delta < scf_control%eps_scf

for the new accelerated candidate. This only adds a raw-map evaluation when
the replacement step is itself small.

Growl1234:
Hi Thomas,

As explained in https://github.com/cp2k/cp2k/pull/5852#issuecomment-5414296471, the change regarding first part is intentional:

> Regarding the first point, this is intentional in order to not waste time on the convergence area, as usually at that status the change in energy is already 1e-9~1e-10 (assuming EPS_SCF to be 1e-5). The check was initially introduced to avoid "false zero" immediately after atomic guess, but was preserved as the last safe guard while the false convergence was prevented through other ways; iiuc only when `step_norm` is lower than 1% of `raw_map_delta` does ADIIS continue the SCF cycle.

For difficult systems, the original handling tended to lead to 5~10 more steps while the energy diff is already very small.

DCM-Uni-Paderborn (approved):
Thank you for the clarification. I still prefer a strict raw-map validation,
because the current heuristic does not guarantee that the final accelerated
candidate satisfies EPS_SCF. However, I understand that this is an intentional
performance trade-off, and the practical risk is limited to the optional ADIIS
workflow.

Given that ODA is now opt-in, ACE is excluded, and no incorrect final result
has been demonstrated, I do not want to block the PR over this design choice.
Approved.

Issue #5699, opened 2026-08-08
Performance regression after Libxc 7.0.0 -> 7.1.2


oschuett:
After #5607 and #5612 the timing of  `libxc_lda_eval` for the `w64SCAN` benchmark roughly doubled. 

This is visible both in the [CUDA](https://dashboard.cp2k.org/archive/perf-cuda-volta/index.html) and [CPU](https://dashboard.cp2k.org/archive/perf-openmp/w64SCAN_timings_32omp.txt) performance testers.

Note that, in the [meantime](https://github.com/cp2k/cp2k/pull/5676) the `libxc_lda_eval` routine got renamed to `libxc_spin_unpolarized_eval`.

Growl1234:
cc @susilehtola

susilehtola:
There's no W64SCAN benchmark in https://github.com/cp2k/cp2k/tree/master/benchmarks. There is no information for me to continue, since the functional isn't even described.

If this is about SCAN, the answer is like that for many other functionals: the floating point stability has been improved. Also, why are you stuck with 7.1.1 when there already were many performance improvements included in 7.1.2?

Growl1234:
> Also, why are you stuck with 7.1.1 when there already were many performance improvements included in 7.1.2?

I think, it's just that Ole made a mistake while describing the issue. The testers are already using 7.1.2 since #5612.

oschuett:
The benchmark is based on this input file: [QS_reference/w64SCAN.inp](https://github.com/cp2k/cp2k/blob/master/benchmarks/QS_reference/w64SCAN.inp).

Growl1234:
It might or might not be related to this issue, but I had a simple look at CP2K's xc/xc_libxc code and its OpenMP implementation looks questionable.

susilehtola:
> It might or might not be related to this issue, but I had a simple look at CP2K's xc/xc_libxc code and its OpenMP implementation looks questionable.

This is definitely the reason why CP2K is sensitive to libxc internals which are usually no issue in codes. Libxc is being called one point at a time instead of the whole batch of grid points being evaluated at once, which is the normal way to do this. I am preparing a patch.

PR #5858, opened 2026-08-26
Evaluate LibXC in batches with cached per-thread workers

susilehtola:
## Motivation

Every LibXC evaluation call in CP2K passed `np = 1`. The library was driven one grid
point at a time, from inside an OpenMP loop, through a functional object created and
destroyed once per batch.

LibXC amortizes its per-call work over the points in a call — argument checks, output
initialization, and for functionals built by mixing components a full allocate/free
cycle of the component buffers. At `np = 1` all of that is paid once per point. It is
also why CP2K's timings are so sensitive to LibXC internals: any change there lands
10^6–10^8 times per SCF cycle instead of a few thousand times.

## What changed

**Batched evaluation.** Both `libxc_spin_*_calc` routines now work in blocks of grid
points, one LibXC call per block, with vectorizable accumulation. Blocks are capped at
512 points and shrink when there are too few to give every thread one, so small atomic
grids still spread across the team. Each kernel has a single block loop: staging selected
by family, one LibXC call selected by (family, order), and accumulation as one loop per
group of derivatives, guarded by the order requested and the family — so each derivative
is written in exactly one place.

**Dead code removed.** The branches for evaluating a single derivative order
(`grad_deriv < 0`) are gone. No caller in CP2K ever requested one: all eleven
`xc_functionals_eval` call sites and all seven `xc_rho_set_and_dset_create` call sites
pass a literal 0, 1, 2 or 3, so these were 536 lines of unreachable code. A negative order
is now rejected up front rather than quietly evaluating nothing. The equivalent branches
in the native functional modules are untouched and remain equally dead.

**Screening moved into LibXC.** `xc_func_set_dens_threshold` is set from
`DENSITY_CUTOFF`; LibXC screens each point and leaves screened outputs at zero, so the
per-point `IF` is gone and whole blocks pass through unfiltered. The threshold
propagates to the component functionals of mixed functionals, so hybrids screen at
CP2K's cutoff too.

The sigma threshold is deliberately left at LibXC's default. Tying it to the density
cutoff (`eps_rho**(4/3)`, LibXC's own scaling) looks principled but is wrong here:
CP2K's cutoff is far above the per-functional defaults, so it floors sigma for
small-gradient points that used to be evaluated as given, and moves GGA energies by
~1e-5 relative.

**Meta-GGA bounds kept, in a pass of their own.** CP2K's screening on tau and its
Fermi hole curvature bound `tau >= sigma/(8*rho)` are both retained, applied in one
extra loop over the already-staged block so the batch stays one contiguous call. A
point failing the tau test is handed to LibXC with zero density, so LibXC screens it
out and leaves its outputs at zero — exactly what skipping it used to achieve.

Worth flagging for reviewers: CP2K's clamp and LibXC's `XC_FLAGS_ENFORCE_FHC` are
*not* the same clamp. CP2K raises tau; LibXC lowers sigma to `8*rho*tau`. Both honour
the bound but hand the functional a different point in `(rho, sigma, tau)`, and where
the bound is active the results differ by orders of magnitude per point. Keeping
CP2K's clamp is what makes this patch numerically inert.

**Cached per-thread workers.** Functional objects are cached, one per OpenMP thread,
keyed on functional, spin channel and cutoffs, and released in `cp2k_finalize`.
Previously each evaluation built and destroyed one — twice, in fact, since the
Laplacian probe created a throw-away object just to read one flag. In GAPW that was
paid per atom, per hard/soft density, per functional component, per SCF step. The dead
`!$OMP CRITICAL(libxc_init)` sections and their stray `!$OMP BARRIER`s are also gone;
the barriers were no-ops where they stood and would have deadlocked if ever reached
from inside a parallel region on a conditional branch.

**GAPW threading.** `xc_atom.F` had no OpenMP directives at all, so in GAPW the
functional evaluation was the only threaded stage of an otherwise serial routine. The
`vxg` assembly loops are now threaded too. No reductions were added, so the arithmetic
is unchanged.

**LibXC version and citation are now reported.** Functional references were already
printed. The version was never printed at all — `libxc_version_info` had no callers
anywhere in the tree — and it reported the `XC_VERSION` header macro rather than
`xc_version_string()`, so even if called it would have named the LibXC that CP2K was
compiled against, not the shared library actually loaded. The library's citation was
hard-coded as `Marques2012` + `Lehtola2018` rather than taken from `xc_reference()`.
Both now come from the library at run time, printed once ahead of the functional
references, with an explicit line when run-time and compile-time versions disagree:

```
 FUNCTIONAL| LIBXC version 7.1.2
 FUNCTIONAL| S. Lehtola, C. Steigemann, M. J. T. Oliveira, and M. A. L. Marques,
 FUNCTIONAL|   SoftwareX 7, 1-5 (2018)
 FUNCTIONAL| doi: 10.1016/j.softx.2017.11.002
 FUNCTIONAL| HYB_GGA_XC_B3LYP:
 ...
```

Note this changes every output file that uses LibXC. The hard-coded `cite_reference`
calls are left alone — whether CP2K should stop citing Marques 2012 in its reference
list is a separate policy call.

## Behaviour change worth knowing about

`xc_func_set_dens_threshold` floors each *individual spin density* as well as screening
the total. CP2K previously screened on the sum but floored each surviving spin channel
only at `EPSILON(0.0_dp)*1.e4_dp`. So with the default `DENSITY_CUTOFF=1e-10`, a point
with `rhoa=1e-9, rhob=0` is now evaluated at `(1e-9, 1e-10)` rather than
`(1e-9, 2.2e-12)`. Thanks to @Growl1234 for catching that it was undocumented.

Measured for exactly that point — `zk`, and the energy contribution `zk*(rhoa+rhob)` that
actually reaches `e_0`:

| functional | zk before | zk after | contribution before | after |
| --- | --- | --- | --- | --- |
| `LDA_C_PW` | -4.352e-04 | -5.111e-04 | -4.361e-13 | -5.623e-13 |
| `GGA_C_LYP` | -1.232e-06 | -4.795e-05 | -1.235e-15 | -5.274e-14 |
| `HYB_GGA_XC_B3LYP` | -9.820e-01 | -8.948e-01 | -9.842e-10 | -9.843e-10 |
| `MGGA_C_TPSS` | -8.363e-08 | -1.941e-07 | -8.382e-17 | -2.135e-16 |

`zk` moves by up to several percent, but it is weighted by a density at the cutoff, so
each affected point shifts the energy by ~1e-13 Ha — consistent with the regression tests
coming back bit-identical. This is deliberate: the screening belongs inside LibXC, which
owns the numerics. Note this is unrelated to the Fermi hole curvature clamp discussed
above, which is *kept* precisely because the two clamps are not equivalent.

## Verification

**The patch is numerically inert.** Both calculation routines were extracted whole from
this branch and from master, compiled against the real LibXC and run side by side on
identical data — so what is compared is the code that ships, not a re-implementation.
LDA/GGA/mGGA, both spin channels, every supported order, and `MGGA_X_BR89` so the
Laplacian branches are exercised, in four density regimes, comparing every accumulated
derivative:

| regime | what it exercises | result |
| --- | --- | --- |
| All points above cutoffs | blocking, buffer layout, spin interleaving | identical |
| 70% below the density cutoff | screening moved into LibXC | identical |
| Fermi hole curvature bound active | the tau clamp | identical |
| Density tail: rho over cutoff, tau far under | the separate screening on tau | identical |

104/104, zero difference — not agreement to a tolerance. 1731 points, deliberately not
a multiple of the block size so the short trailing block is exercised.

**Regression tests**, patched vs. unpatched HEAD, same machine:

- `QS/regtest-libxc`, `ATOM/regtest-libxc`, `QS/regtest-admm-libxc`: same verdict on
  all 20 tests, values agreeing to ~1e-13.
- `QS/regtest-gapw*`, all seven suites: 81/81 correct.
- TDDFPT + GPW + ATOM + `gapw_xc`, covering the second-derivative paths: correct on both
  binaries. The combined sweep over all of the above is 269/276, with the same seven
  deviating tests as unpatched master (see the note at the end).

## Performance

Not yet measured on the code as it now stands. The kernel timings that were here before
were taken on the pre-restructure kernels and on a machine that turned out to be under
load, so they are not something I am willing to stand behind. They will be replaced with
numbers measured on an idle machine.

What can be said without a timer, from the structure of the change: at `np = 1` LibXC
paid its per-call cost — argument checks, output initialization, and for mixed
functionals a full allocate/free cycle of the component buffers — once per grid point
rather than once per block, and that cost is now amortized over up to 512 points. The
gain is therefore largest for functionals whose per-call overhead is a large fraction of
their per-point work, i.e. cheap functionals and mixed ones; it is smallest for expensive
non-mixed functionals, where the kernel dominates either way.

Two caveats that hold regardless of the numbers:

- This is the functional evaluation alone. The removed per-batch `xc_func_init` /
  `xc_func_end` pairs and the newly threaded GAPW assembly loops are on top and do not
  show up in a kernel benchmark.
- An end-to-end figure will not come from the regression suite: those systems are far too
  small for XC evaluation to be a meaningful fraction of the runtime. A benchmark on a
  real system would be needed.

## Note for reviewers

Seven LibXC regtests already deviate from the stored references on **unpatched
master** on my machine — `C_tpss_libxc`, `H2O-BP-BECKEX`, both `b3lyp_libxc`,
`cam-lda0`, `wb97mv`, `He_4`. They deviate identically with and without this patch, so
they are not caused by it; most are at the 1e-14–1e-13 level against tolerances set as
tight as 2e-14. This machine links stock LibXC 7.1.2, so this is most plausibly a
version discrepancy against whatever generated the references. Flagging it so a red
regtest run on this branch is not misattributed.

https://claude.ai/code/session_01Nv2Y6yhiyvtzakoyfxDhQZ

### DIFF
--- a/src/xc/xc_libxc_wrap.F
+++ b/src/xc/xc_libxc_wrap.F
@@ -27,13 +27,16 @@ MODULE xc_libxc_wrap
    This version of CP2K ONLY works with libxc versions 5.1.0 and above.
    Furthermore, -I${LIBXC_DIR}/include needs to be added to FCFLAGS.
 #else
-   ! Functionals which require parameters
-   USE cp_log_handling, ONLY: cp_to_string
    USE kinds, ONLY: dp
    USE xc_f03_lib_m, ONLY: xc_f03_func_end, &
                            xc_f03_func_init, &
                            xc_f03_functional_get_name, &
                            xc_f03_func_set_ext_params, &
+                           xc_f03_func_set_dens_threshold, &
+                           xc_f03_func_set_tau_threshold, &
+                           xc_f03_version_string, &
+                           xc_f03_reference, &
+                           xc_f03_reference_doi, &
                            xc_f03_functional_get_number, &
                            xc_f03_available_functional_numbers, &
                            xc_f03_available_functional_names, &
@@ -99,11 +102,6 @@ MODULE xc_libxc_wrap
                            XC_FLAGS_HAVE_EXC, &
                            XC_FLAGS_DEVELOPMENT
 
-   USE input_section_types, ONLY: section_add_keyword, &
-                                  section_add_subsection, &
-                                  section_create, &
-                                  section_release, &
-                                  section_type, section_vals_type, section_vals_val_get
 #include "../base/base_uses.f90"
 #endif
    IMPLICIT NONE
@@ -115,7 +113,7 @@ MODULE xc_libxc_wrap
    CHARACTER(LEN=*), PARAMETER, PUBLIC :: libxc_version = XC_VERSION
 
    PUBLIC :: xc_f03_func_t, xc_f03_func_info_t
-   PUBLIC :: xc_f03_func_init, xc_f03_func_end
+   PUBLIC :: xc_f03_func_init, xc_f03_func_end, xc_f03_func_set_ext_params
    PUBLIC :: xc_f03_functional_get_name, xc_f03_available_functional_numbers, xc_f03_maximum_name_length, &
              xc_f03_number_of_functionals, xc_f03_available_functional_names
    PUBLIC :: xc_f03_func_get_info, xc_f03_func_info_get_family, xc_f03_func_info_get_kind, &
@@ -140,9 +138,11 @@ MODULE xc_libxc_wrap
 ! wrappers for routines
    PUBLIC :: xc_libxc_wrap_info_refs, &
              xc_libxc_wrap_version, &
+             xc_libxc_wrap_library_reference, &
+             xc_libxc_wrap_info_needs_laplace, &
+             xc_libxc_wrap_info_no_exc, &
+             xc_libxc_wrap_set_thresholds, &
              xc_libxc_wrap_functional_get_number, &
-             xc_libxc_wrap_needs_laplace, &
-             xc_libxc_wrap_functional_set_params, &
              xc_libxc_wrap_is_under_development, &
              xc_libxc_get_reference_length, &
              xc_libxc_check_functional
@@ -292,13 +292,18 @@ CONTAINS
    END SUBROUTINE xc_libxc_wrap_info_refs
 
 ! **************************************************************************************************
-!> \brief Provides a version string.
-!> \param version ...
+!> \brief Provides the version of the LibXC library CP2K is running against.
+!> \param version version string reported by the library at run time
+!> \param compiled_version version of the headers CP2K was compiled against
 !> \author A. Gloess (agloess)
+!> \note The run-time and compile-time versions differ whenever CP2K is executed
+!>       against a shared LibXC other than the one it was built with. Reporting
+!>       the run-time version is what actually describes the numbers produced.
 !>
 ! **************************************************************************************************
-   SUBROUTINE xc_libxc_wrap_version(version)
+   SUBROUTINE xc_libxc_wrap_version(version, compiled_version)
       CHARACTER(LEN=*), INTENT(OUT)                      :: version
+      CHARACTER(LEN=*), INTENT(OUT), OPTIONAL            :: compiled_version
 
       CHARACTER(LEN=*), PARAMETER :: routineN = 'xc_libxc_wrap_version'
 
@@ -306,11 +311,74 @@ CONTAINS
 
       CALL timeset(routineN, handle)
 
-      version = TRIM(libxc_version)
+      CALL xc_f03_version_string(version)
+      IF (PRESENT(compiled_version)) compiled_version = TRIM(libxc_version)
 
       CALL timestop(handle)
 
    END SUBROUTINE xc_libxc_wrap_version
+
+! **************************************************************************************************
+!> \brief Provides the citation LibXC asks to be given for the library itself.
+!> \param reference bibliographic reference
+!> \param doi digital object identifier of that reference
+!> \author S. Lehtola
+!> \note Taken from the library at run time rather than hard-coded, so that the
+!>       citation stays correct when LibXC updates its own self-citation.
+!>
+! **************************************************************************************************
+   SUBROUTINE xc_libxc_wrap_library_reference(reference, doi)
+      CHARACTER(LEN=*), INTENT(OUT)                      :: reference, doi
+
+      CHARACTER(LEN=*), PARAMETER :: routineN = 'xc_libxc_wrap_library_reference'
+
+      INTEGER                                            :: handle
+
+      CALL timeset(routineN, handle)
+
+      CALL xc_f03_reference(reference)
+      CALL xc_f03_reference_doi(doi)
+
+      CALL timestop(handle)
+
+   END SUBROUTINE xc_libxc_wrap_library_reference
+
+! **************************************************************************************************
+!> \brief Applies the CP2K screening cutoffs to a LibXC functional object.
+!> \param xc_func LibXC functional object
+!> \param xc_info info object belonging to xc_func
+!> \param epsilon_rho density cutoff (XC%DENSITY_CUTOFF)
+!> \param epsilon_tau kinetic energy density cutoff (XC%TAU_CUTOFF)
+!> \author S. Lehtola
+!> \note LibXC screens every point against the density threshold internally, which
+!>       is what allows CP2K to hand it whole blocks of points instead of filtering
+!>       point by point. The sigma threshold is deliberately left at LibXC's own
+!>       default: CP2K has no cutoff of its own on the gradient here, and raising
+!>       it in step with the density cutoff would floor sigma for points with a
+!>       small gradient that used to be evaluated as given.
+!>
+! **************************************************************************************************
+   SUBROUTINE xc_libxc_wrap_set_thresholds(xc_func, xc_info, epsilon_rho, epsilon_tau)
+      TYPE(xc_f03_func_t), INTENT(INOUT)                 :: xc_func
+      TYPE(xc_f03_func_info_t), INTENT(IN)               :: xc_info
+      REAL(KIND=dp), INTENT(IN)                          :: epsilon_rho, epsilon_tau
+
+      CHARACTER(LEN=*), PARAMETER :: routineN = 'xc_libxc_wrap_set_thresholds'
+
+      INTEGER                                            :: handle
+
+      CALL timeset(routineN, handle)
+
+      CALL xc_f03_func_set_dens_threshold(xc_func, epsilon_rho)
+
+      SELECT CASE (xc_f03_func_info_get_family(xc_info))
+      CASE (XC_FAMILY_MGGA, XC_FAMILY_HYB_MGGA)
+         CALL xc_f03_func_set_tau_threshold(xc_func, epsilon_tau)
+      END SELECT
+
+      CALL timestop(handle)
+
+   END SUBROUTINE xc_libxc_wrap_set_thresholds
 
 ! **************************************************************************************************
 !> \brief Checks existence of functional in LibXC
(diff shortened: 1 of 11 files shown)


susilehtola:
This should go a far way to close #5699

src/xc/xc_libxc_wrap.F
```diff
@@ -292,26 +292,94 @@ SUBROUTINE xc_libxc_wrap_info_refs(xc_info, polarized, sc, reference)
...
+      TYPE(xc_f03_func_info_t), INTENT(IN)               :: xc_info
+      REAL(KIND=dp), INTENT(IN)                          :: epsilon_rho, epsilon_tau
+
+      CHARACTER(LEN=*), PARAMETER :: routineN = 'xc_libxc_wrap_set_thresholds'
+
+      INTEGER                                            :: handle
+
+      CALL timeset(routineN, handle)
+
+      CALL xc_f03_func_set_dens_threshold(xc_func, epsilon_rho)
```
Growl1234, line 372: This does not preserve the previous spin-polarized semantics. LibXC uses `dens_threshold` both to screen the total density and to floor each individual spin density. CP2K previously screened on the sum, but floored each surviving spin channel only to `EPSILON(0.0_dp)*1.e4_dp`. With the default `DENSITY_CUTOFF=1e-10`, a point with `rhoa=1e-9` and `rhob=0` is therefore evaluated as `(1e-9, 1e-10)` after this change, instead of approximately `(1e-9, 2.2e-12)`. The same issue applies to the per-spin tau floor (L376) and changes the FHC bound for a weakly occupied spin channel.

It migt meet your note:

> Worth flagging for reviewers: CP2K's clamp and LibXC's XC_FLAGS_ENFORCE_FHC are not the same clamp. CP2K raises tau; LibXC lowers sigma to 8*rho*tau. Both honour the bound but hand the functional a different point in (rho, sigma, tau), and where the bound is active the results differ by orders of magnitude per point. Keeping CP2K's clamp is what makes this patch numerically inert.

Growl1234 (commented):
Most of the added code appears to come from repeating the same block staging and OpenMP loop for every derivative order. Could the block loop be hoisted outside the `grad_deriv` selection, with only the LibXC call and result accumulation selected inside each block? This would keep the current two calculation routines and avoid introducing more helpers, while substantially reducing duplicated screening, staging and FHC logic.

src/xc/xc_libxc_wrap.F
susilehtola: That screening should only happen inside libxc, which is responsible for the numerics.

Growl1234 (approved):
Regarding regtesting failures, I think it would be OK to simply replace the reference values. That being said, let's see what pure OpenMP test (Regtest ssmp) goes on.

Moreover, please cleanup the unused `\para` in the doxygen.

susilehtola:
Thanks — both points taken, and the patch is smaller for it. Force-pushed.

## Hoisting the block loop

Done, and it was the right call. Each kernel now has a single block loop. Inside it:
staging selected by family, one LibXC call selected by (family, order), then accumulation
written as one loop per group of derivatives, guarded by the order requested and the
family. Nothing is duplicated per `(family, grad_deriv)` pair any more, and no new helper
routines — still the same two calculation routines.

The accumulation fell out naturally as incremental-by-order, mirroring how
`libxc_spin_*_eval` already decides which derivative arrays to allocate: order 0 adds
`e_0`, order 1 the first derivatives, order 2 the second, order 3 the third, with the
gradient- and tau-dependent terms behind `is_gga`/`is_mgga` and the Laplacian terms behind
`has_laplace`. Each derivative is now written in exactly one place.

`libxc_spin_unpolarized_calc` 348 -> 238 lines, `libxc_spin_polarized_calc` 559 -> 340.

One deliberate addition: the spin-polarized chain rule is written in terms of the floored
gradient norms, which the old code recomputed inside every branch. They are now staged
once per block into three small per-thread buffers (`nd`, `nda`, `ndb`, 4 kB each) so the
accumulation loops index them instead of re-deriving them.

## The `unused` array

Agreed, and that one was a regression I introduced. The previous code pointed those
arguments at the density array, which costs nothing but makes an output argument alias a
read-only input; I fixed the aliasing by allocating a real array and did not think about
the allocation traffic that adds back.

It is now a module-level scratch buffer grown on demand and kept for the run
(`libxc_unused`, reserved through `libxc_reserve_unused`, released with the workers in
`libxc_release_workers`), with the pointer bounds-remapped onto it. An evaluation costs no
allocation at all, and the buffer only ever grows to the largest grid seen — so both the
transient allocation and the aliasing are gone.

## Also in this push

The branches for evaluating a single derivative order (`grad_deriv < 0`) are removed. No
caller in CP2K ever requested one — all eleven `xc_functionals_eval` sites and all seven
`xc_rho_set_and_dset_create` sites pass a literal 0, 1, 2 or 3 — so they were 536 lines of
unreachable code. A negative order is now rejected up front rather than quietly evaluating
nothing. The equivalent branches in the native functional modules are untouched and remain
equally dead; that is a separate cleanup.

Between the two, the diff went from **+1883/-1042** to **+1123/-1117**.

## Re-verification

Both calculation routines were extracted whole from this branch and from master, compiled
against LibXC and run side by side on identical data: LDA/GGA/mGGA, both spin channels,
every supported order, four density regimes (all points live; 70% below the density
cutoff; the FHC bound active everywhere; density tails with rho over the cutoff and tau
far under), including `MGGA_X_BR89` so the Laplacian branches are exercised.
**104/104 with exactly zero difference** in every accumulated derivative.

CP2K rebuilds clean, and the regression sweep over `QS/regtest-libxc`,
`ATOM/regtest-libxc`, `QS/regtest-admm-libxc`, all seven `QS/regtest-gapw*`, TDDFPT and
GPW gives 269/276 — the same verdict, and the same seven deviating tests, as unpatched
master on this machine (see the note at the end of the PR description).

src/xc/xc_libxc_wrap.F
susilehtola: You are right about the mechanism, and it was not in the PR description — my omission.
`xc_func_set_dens_threshold` does floor each spin channel as well as screening the total,
so a point with `rhoa=1e-9, rhob=0` is now evaluated at `(1e-9, 1e-10)` where it used to
be `(1e-9, 2.2e-12)`.

Measured, for exactly that point (`zk`, then the energy contribution `zk*(rhoa+rhob)`
which is what actually reaches `e_0`):

| functional | zk before | zk after | contribution before | after |
| --- | --- | --- | --- | --- |
| `LDA_C_PW` | -4.352e-04 | -5.111e-04 | -4.361e-13 | -5.623e-13 |
| `GGA_C_LYP` | -1.232e-06 | -4.795e-05 | -1.235e-15 | -5.274e-14 |
| `HYB_GGA_XC_B3LYP` | -9.820e-01 | -8.948e-01 | -9.842e-10 | -9.843e-10 |
| `MGGA_C_TPSS` | -8.363e-08 | -1.941e-07 | -8.382e-17 | -2.135e-16 |

So `zk` moves by up to several percent, but the density it is weighted by is at the
cutoff, and each affected point shifts the energy by ~1e-13 Ha. Consistent with the
regression tests coming back bit-identical to master — the configuration is rare enough
on a real grid that nothing moved at printed precision.

One clarification on the link to the FHC note: that is a separate issue and it went the
other way. CP2K's Fermi hole curvature clamp is *kept* in this PR precisely because
LibXC's `XC_FLAGS_ENFORCE_FHC` is not equivalent — CP2K raises tau, LibXC lowers sigma —
and dropping it did change meta-GGA results materially. The per-spin floor above is only
about `dens_threshold`, and per @susilehtola's call the screening belongs inside LibXC.
I have added it to the PR description so it is on the record.

fstein93:
Please run `./make_pretty.sh`

fstein93:
Rerun `./make_pretty.sh`

fstein93:
You may need to update the reference values and/or the thresholds

Growl1234:
The ssmp failure is not related to this PR. Do you think this PR is ready?

Unfortunately it seems the OpenMP performance regarding `libxc_spin_unpolarized_eval` in w64SCAN benchmarker isn't significantly improved (about 10~15%), but I can take care of it deeply if your time is rather limited, as I'm on OpenMP optimization work.

fstein93 (approved):

fstein93:
As a follow-up PR, we might remove the dead code related to negative orders of derivative.

Growl1234:
A useful CP2K follow-up would be to lift the block loop over a group of LibXC components: create one OpenMP team, stage `rho`/`sigma`/`tau` once per block, evaluate the components in their original order, and accumulate them before moving to the next block. Parallel initialization/first-touch of the derivative grids should be included as well, since they are currently zeroed before entering the LibXC parallel region and may contribute to the remaining OMP/MPI difference. The proposed grouped executor might be somehow similar to what the initial status of this PR wanted to do, but it's a further lifting of the same block structure, not a return to the earlier duplicated implementation.

Recovering performance closer to the former will probably also require SIMD or code-generation work in LibXC's batched CPU kernel, but I believe numerical stavility is the most important if we must sacrifice some performance for it.

Since this PR is already a large and involved one, and we have observed some improvement of performance (though not significantly), I think it's worth being merged now.

oschuett:
Lifting the loops should also go a long way toward enabling libxc for GPUs.

fstein93:
We may have a look at GPU-acceleration in LibXC at a Hackathon next week.

Issue #5857, opened 2026-08-26
Reject GET_ENVIRONMENT_VARIABLE in conventions


Growl1234:
Environment variables **must not** act as hidden CP2K calculation inputs.

> I'm very happy about this change. IMHO, using environment variables to steer calculations is really an anti-pattern that hurts reproducibility. Hence, I'm contemplating to add `GET_ENVIRONMENT_VARIABLE` to our [banned routines](https://github.com/cp2k/cp2k/blob/a03822c0f5e31115dd30b556661ef75333804ea3/tools/conventions/analyze_gfortran_ast.py#L147).

Besides previous D4 code, there are many other places containing it:
```
base/machine.F:742:      CALL get_environment_variable("LOGNAME", value=user, status=istat)
base/machine.F:745:         CALL get_environment_variable("USER", value=user, status=istat)
base/machine.F:749:         CALL get_environment_variable("USERNAME", value=user, status=istat)
base/machine.F:768:      CALL get_environment_variable("OMP_STACKSIZE", value=omp_stacksize, status=istat)
skala_gpw_functional.F:3003:            CALL GET_ENVIRONMENT_VARIABLE("GAUXC_SKALA_CUDA_MODEL", model_path, STATUS=env_status)
skala_gpw_functional.F:3006:         CALL GET_ENVIRONMENT_VARIABLE("GAUXC_SKALA_MODEL", model_path, STATUS=env_status)
start/cp2k.F:172:         CALL get_environment_variable("VECLIB_MAXIMUM_THREADS", env_var, status=ierr)
tblite_interface.F:1572:      CALL GET_ENVIRONMENT_VARIABLE("CP2K_TBLITE_DEBUG_SKIP_SCF_DISPERSION", debug_value, STATUS=debug_status)
tblite_interface.F:1778:      CALL GET_ENVIRONMENT_VARIABLE("CP2K_TBLITE_RAW_STATE_DUMP", raw_state_file, STATUS=raw_state_status)
tblite_interface.F:3292:      CALL GET_ENVIRONMENT_VARIABLE("CP2K_TBLITE_FORCE_DUMP", dump_file, STATUS=dump_status)
tblite_interface.F:4803:      CALL GET_ENVIRONMENT_VARIABLE("CP2K_TBLITE_SIGMA_COMPONENT_DUMP", dump_file, STATUS=dump_status)
[...]

PR #5842, opened 2026-08-24
Expose D4 smooth cutoff widths as input

DCM-Uni-Paderborn:
## Summary

- add `D4_CUTOFF_WIDTH` (alias `D4_2B_CUTOFF_WIDTH`) and `D4_3B_CUTOFF_WIDTH` to `&VDW_POTENTIAL / &PAIR_POTENTIAL`
- remove the DFTD4/tblite environment-variable overrides from the CP2K D4 interface
- validate that configured widths are non-negative and smaller than their cutoffs
- cover both keywords in the sparse periodic D4 regression input

This follows up the post-merge discussion on #5143. The numerical defaults remain unchanged: the two-body width is `0.05` bohr, while three-body smoothing is disabled by default. The original stress discontinuity was isolated to the two-body D4 term; enabling three-body smoothing by default would change behavior beyond that fix. It can now be enabled explicitly through the input.

## Validation

- precommit checks for all changed Fortran files
- GCC 16 / OpenMPI build of `cp2k.psmp` with tblite/DFTD4
- `pbe_dftd4_sparse.inp` with two MPI ranks

PR #5860, opened 2026-08-26
Ban direct environment-variable queries

DCM-Uni-Paderborn:
## Summary

- flag direct `GET_ENVIRONMENT_VARIABLE` calls in the GFortran AST conventions check
- explicitly suppress the currently audited calls that are present in the conventions build
- prevent new procedures from introducing hidden environment-controlled inputs
- account for three current-master GFortran AST false positives exposed by the conventions run

## Rationale

This follows the reproducibility discussion in #5842. Environment variables should not silently steer CP2K calculations.

The existing detected calls are retained as an explicit migration boundary so this conventions-only change has no runtime effect. Most provide runtime metadata, external resource locations, or diagnostics. Calls in the main program and developer-only tblite code are not emitted by the current conventions build and therefore do not need suppressions. In particular, removing the SKALA model-path aliases needs a separate focused change because toolchain, Spack, and regtest configurations currently rely on them.

The initial CI run also exposed three violations already present on master. The two ELPA imports have explicit `ONLY` clauses but are reported through imported derived-type metadata in the GFortran AST. The Libint engines in `coulomb_integral_context_type` cannot be default-constructed and are initialized explicitly by `coulomb_integral_init`. These three false positives are documented and suppressed.

## Validation

- `./make_pretty.sh`
- `python3 -m py_compile tools/conventions/analyze_gfortran_ast.py tools/conventions/summarize_issues.py`
- verified rule 107 using actual GFortran AST output from a small `GET_ENVIRONMENT_VARIABLE` call
- `git diff --check origin/master...HEAD`

Refs #5857

### DIFF
--- a/tools/conventions/conventions.supp
+++ b/tools/conventions/conventions.supp
@@ -246,5 +246,17 @@ xyz2dcd.F: Found STOP statement in procedure "xyz2dcd" https://cp2k.org/conv#c20
 xyz2dcd.F: Found WRITE statement with hardcoded unit in "abort_program" https://cp2k.org/conv#c012
 xyz2dcd.F: Found WRITE statement with hardcoded unit in "print_help" https://cp2k.org/conv#c012
 xyz2dcd.F: Found WRITE statement with hardcoded unit in "xyz2dcd" https://cp2k.org/conv#c012
+# Existing environment queries audited in #5857. These provide runtime metadata,
+# external resource locations, warnings, or developer-only diagnostics.
+cp2k.F: Found CALL GET_ENVIRONMENT_VARIABLE in procedure "cp2k" https://cp2k.org/conv#c107
+machine.F: Found CALL GET_ENVIRONMENT_VARIABLE in procedure "m_getlog" https://cp2k.org/conv#c107
+machine.F: Found CALL GET_ENVIRONMENT_VARIABLE in procedure "m_omp_get_stacksize" https://cp2k.org/conv#c107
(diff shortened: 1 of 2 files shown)


oschuett:
Thanks a lot for taking this on. I just created [cp2k.org/conv#c107](https://www.cp2k.org/conv#c107).

oschuett (approved):

tools/conventions/conventions.supp
```diff
@@ -246,5 +246,17 @@ xyz2dcd.F: Found STOP statement in procedure "xyz2dcd" https://cp2k.org/conv#c20
 xyz2dcd.F: Found WRITE statement with hardcoded unit in "abort_program" https://cp2k.org/conv#c012
 xyz2dcd.F: Found WRITE statement with hardcoded unit in "print_help" https://cp2k.org/conv#c012
 xyz2dcd.F: Found WRITE statement with hardcoded unit in "xyz2dcd" https://cp2k.org/conv#c012
+# Existing environment queries audited in #5857. These provide runtime metadata,
+# external resource locations, warnings, or developer-only diagnostics.
+cp2k.F: Found CALL GET_ENVIRONMENT_VARIABLE in procedure "cp2k" https://cp2k.org/conv#c107
```
oschuett: It seems this one is not needed. Apparently, the conventions check only works for Fortran module but not for Programs, which is unfortunate but ok.

```suggestion
```

DCM-Uni-Paderborn: Removed. I also dropped the four tblite suppressions because those developer-only calls are not present in the conventions build. Thanks.

Issue #5735, opened 2026-08-13
Toolchain `scalapack.pc` pulls in system BLAS/LAPACK, breaking `psmp` linking


sjaehnigen:
The `scalapack.pc` installed by `tools/toolchain` (ScaLAPACK 2.2.3) has `libdir=/lib`
instead of `${prefix}/lib`, and `Requires: mpi lapack blas`. The latter makes pkg-config
resolve BLAS/LAPACK transitively against the *system* `.pc` files, ignoring the OpenBLAS
the toolchain just built and whatever `-DCP2K_BLAS_VENDOR` was set to.

`ssmp` builds fine; only the MPI versions pull in ScaLAPACK.

## Environment

- CP2K version/commit: d734930
- Debian 13 (trixie), gcc 14.2.0, system OpenMPI, toolchain-installed `openblas-0.3.34` + `scalapack-2.2.3`
- `./install_cp2k_toolchain.sh -j 16 --with-plumed=install --with-deepmd=install --with-libtorch=install`

## Installed file

`tools/toolchain/install/scalapack-2.2.3/lib/pkgconfig/scalapack.pc`:

```
prefix=/path/to/cp2k/tools/toolchain/install/scalapack-2.2.3
libdir=/lib
Name: scalapack
Version: 2.2.3
Libs: -L -lscalapack
Requires: mpi lapack blas
```

Note `libdir=/lib` and the bare `-L`.

## Result

```console
$ pkg-config --libs lapack blas
-L/usr/lib/x86_64-linux-gnu/openblas-openmp -llapack -lblas
```

`FindSCALAPACK.cmake:92` uses `pkg_check_modules(CP2K_SCALAPACK ... "scalapack")`, so those
paths land in the cache and on the link line alongside the toolchain OpenBLAS:

```
.../tools/toolchain/install/scalapack-2.2.3/lib/libscalapack.a
/usr/lib/x86_64-linux-gnu/openblas-openmp/liblapack.so
/usr/lib/x86_64-linux-gnu/openblas-openmp/libblas.so
.../tools/toolchain/install/openblas-0.3.34/lib/libopenblas.so
```

```
/bin/ld: /usr/lib/x86_64-linux-gnu/openblas-openmp/liblapack.so: undefined reference to `gotoblas'
gmake[2]: *** [src/CMakeFiles/cp2k-bin.dir/build.make:124: bin/cp2k.psmp] Error 1
```

[...]

PR #5862, opened 2026-08-26
BUILD: Make FindLapack honor user-provided CP2K_LAPACK_LINK_LIBRARIES

s8ga:
## Bug
`cmake/modules/FindLapack.cmake` ignores a user-provided
`CP2K_LAPACK_LINK_LIBRARIES` when `CP2K_BLAS_VENDOR` is `CUSTOM`.
`CUSTOM` is the standard setting for spack builds.
The module calls `pkg_check_modules(CP2K_LAPACK lapack)` without a guard.
On hosts that provide a system `lapack.pc`, for example Debian or Ubuntu installed
`libopenblas-dev` using apt, pkg-config writes plain variables. CMake reads a
plain variable before a cache variable of the same name, so the
pkg-config result hides the user value.

The link line contains the spack OpenBLAS for the BLAS target and the
system `/usr/lib/x86_64-linux-gnu/openblas-pthread/liblapack.so` for the
LAPACK target. `libcp2k.so` then records `NEEDED libopenblas.so.0` and
`NEEDED liblapack.so.3`. At run time the process loads two complete
OpenBLAS runtimes. BLAS then crash at random lines.

## Fix

FindBlas.cmake already treats a 
defined CP2K_BLAS_LINK_LIBRARIES as authoritative, and its own FATAL_ERROR 
message names both variables. This extends the same rule to the LAPACK half.

The new FATAL_ERROR guard stops a CUSTOM build without CP2K_LAPACK_LINK_LIBRARIES 
with a clear message instead of a silent system search. 
The DEFINED guard accepts a user value and skips the search.

_Note: the new FATAL_ERROR changes behavior for CUSTOM builds that
set only CP2K_BLAS_LINK_LIBRARIES and relied on the silent search. 
I mirrored the FindBlas.cmake wording intentionally, but I am happy to 
downgrade it to a warning if you prefer._

<details>
<summary>How to reproduce</summary>

```text

prerequisite: libopenblas-dev installed via apt (Debian/Ubuntu)

$ spack spec    # trimmed to the relevant nodes
  cp2k@master ~cosma ~cuda +dbm_gpu ~dlaf +elpa +grid_gpu +libint +libxc +mpi
              ~mpi_f08 +openmp ~shared ~sirius ...        %gcc@14.2.0
  ^openblas@0.3.30 +fortran +pic +shared threads=openmp   %gcc@14.2.0
  ^openmpi@5.0.10                                         %gcc@14.2.0
  ^elpa@2026.02.002 +openmp
  ^netlib-scalapack@2.2.3 +shared

$ mpirun -np 2 cp2k.psmp <regtest>
  SIGSEGV in unrelated code paths, for example atom_operators.F:317

$ ldd cp2k.psmp | grep -iE 'lapack|openblas'
  liblapack.so.3   => /usr/lib/x86_64-linux-gnu/openblas-pthread/liblapack.so.3
  libopenblas.so.0 => <spack>/openblas-0.3.30-.../lib/libopenblas.so.0

$ readelf -d libcp2k.so.2026.2 | grep NEEDED
  NEEDED liblapack.so.3
  NEEDED libopenblas.so.0

$ patchelf --replace-needed liblapack.so.3 libopenblas.so.0 libcp2k.so.2026.2
  ->  crashes gone, energies bit-identical
```
</details>

### DIFF
--- a/cmake/modules/FindLapack.cmake
+++ b/cmake/modules/FindLapack.cmake
@@ -31,18 +31,32 @@ if(NOT CP2K_CONFIG_PACKAGE)
       get_target_property(CP2K_LAPACK_LINK_LIBRARIES cp2k::BLAS::blas
                           INTERFACE_LINK_LIBRARIES)
     else()
-      # we might get lucky to find a pkgconfig package for lapack (fedora
-      # provides one for instance)
-      if(PKG_CONFIG_FOUND)
-        pkg_check_modules(CP2K_LAPACK lapack)
+      if(CP2K_BLAS_VENDOR MATCHES "CUSTOM" AND NOT DEFINED
+                                               CP2K_LAPACK_LINK_LIBRARIES)
+        message(
+          FATAL_ERROR
+            "Setting CP2K_BLAS_VENDOR=CUSTOM imply setting CP2K_LAPACK_LINK_LIBRARIES to the right libraries."
+        )
       endif()
 
-      if(NOT CP2K_LAPACK_FOUND)
-        find_library(
-          CP2K_LAPACK_LINK_LIBRARIES
-          NAMES "lapack" "lapack64"
-          PATH_SUFFIXES "openblas" "openblas64" "openblas-pthread"
-                        "openblas-openmp" "lib" "lib64")
+      if(DEFINED CP2K_LAPACK_LINK_LIBRARIES)
+        # the user provided the libraries (CP2K_BLAS_VENDOR=CUSTOM), do not
+        # search
+        set(CP2K_LAPACK_FOUND TRUE)
+      else()
+        # we might get lucky to find a pkgconfig package for lapack (fedora
+        # provides one for instance)
+        if(PKG_CONFIG_FOUND)
+          pkg_check_modules(CP2K_LAPACK lapack)
+        endif()
+
+        if(NOT CP2K_LAPACK_FOUND)
+          find_library(
+            CP2K_LAPACK_LINK_LIBRARIES
+            NAMES "lapack" "lapack64"
+            PATH_SUFFIXES "openblas" "openblas64" "openblas-pthread"
+                          "openblas-openmp" "lib" "lib64")
+        endif()
       endif()
     endif()
   endif()


he-zilong:
Does this PR address issue #5735, and is there something better suited as a report to the upstream library providers?

> On hosts that provide a system `lapack.pc`, for example Debian or Ubuntu installed `libopenblas-dev` using apt, pkg-config writes plain variables.

As a side note, I think there would still be many subtleties pending investigation regarding dependencies prepared by system package managers as their behavior vary by OS a lot...

cc @Growl1234

Growl1234:
Issue #5735 is actually about ScaLAPACK itself but can be patched in the toolchain. I didn't investigate this PR and am not sure if this is of help.

s8ga:
Thanks for your comments and for linking the issue.

issue #5735 has a different entry point (scalapack.pc consumed by FindSCALAPACK.cmake), so this PR does not address it.

On the motivation for this patch: when CP2K_BLAS_VENDOR=CUSTOM, the caller states exactly which BLAS and LAPACK libraries to link — that is the documented contract (see the FATAL_ERROR in [FindBlas.cmake](https://github.com/cp2k/cp2k/blob/8ddd909fbc89cfdc7b275e0ca702763d6045d73c/cmake/modules/FindBlas.cmake#L93-L102) ). The unguarded pkg_check_modules(lapack) breaks this contract: on hosts where a system lapack.pc exists, pkg-config writes plain variables with the same names, which shadow the user-provided cache values inside CMake. The result is a binary that links a second BLAS/LAPACK runtime next to the intended one and crashes at random points. The fix makes FindLapack honor user-provided values like FindBlas already does.

The distro differences you mention are real. Concretely:

- Debian/Ubuntu: generic lapack.pc/blas.pc ship with libopenblas-dev and point into one OpenBLAS variant (-L.../openblas-pthread); any unguarded pkg_check_modules(lapack) next to a self-provided BLAS picks up that second runtime.
- Fedora: FlexiBLAS ships the generic .pc files and dispatches to an OpenBLAS backend; upstream OpenBLAS ships no .pc in Fedora packages.
- Upstream OpenBLAS: generates blas/lapack/cblas/openblas.pc, but distros strip or rewrite them, so there is no cross-distro layout to rely on.

As for reports to upstream library providers, three look worthwhile:

1. netlib-scalapack — their CMake installs a broken scalapack.pc: libdir=/lib drops ${prefix} (see #5735), and Requires: lapack blas forces any consumer back through pkg-config, re-resolving BLAS/LAPACK to whatever the OS provides instead of the BLAS the consumer built ScaLAPACK against.
2. Debian/Ubuntu openblas packaging — shipping generic lapack.pc/blas.pc that hard-link into a concrete runtime variant (openblas-pthread) means generic solver names silently drag in a full second BLAS; an .pc with the same-name-but-different-runtime trap affects every build system doing pkg-config lapack beside a bundled BLAS.

~~ (withdrawn — see Growl1234's note below)
3. [OpenBLAS upstream](https://github.com/OpenMathLib/OpenBLAS/issues/1740) / Fedora packagers — OpenBLAS generates correct .pc files but they are absent or [rewritten downstream](https://www.openmathlib.org/OpenBLAS/docs/install/#__tabbed_1_3:~:text=Fedora%20does%20not%20ship%20the%20pkg%2Dconfig%20files%20for%20OpenBLAS.%20Instead%2C%20it%20wants%20you%20to%20link%20against%20FlexiBLAS%20(which%20uses%20OpenBLAS%20by%20default%20as%20its%20backend%20on%20Fedora)%2C%20which%20you%20can%20install%20with%3A); restoring them in Fedora would make pkg-config queries resolve predictably.
~~

That is also why honoring explicit user values is the robust path: it does not depend on how each OS happens to lay out its files.

Growl1234:
From Fedora packager's aspect I want to point out something:

In Fedora 43–45 branches, and RHEL/Rocky 10, `flexiblas-devel` provides only `flexiblas*.pc`; the generic `blas.pc` and `lapack.pc` are provided by the Netlib BLAS/LAPACK packages instead.

Fedora's intended abstraction is that BLAS/LAPACK consumers link against **FlexiBLAS itself**, while the actual backend (OpenBLAS, BLIS, Netlib, etc.) is selected behind that wrapper. This design was introduced specifically to keep a software stack on one BLAS/LAPACK ABI and avoid loading different implementations.

Therefore, restoring `openblas.pc` in Fedora would not really make generic `pkg-config blas/lapack` queries more correct; it would bypass the distribution's FlexiBLAS abstraction. I think, Fedora's `openblas-devel` intentionally exposes the OpenBLAS libraries and headers without providing `openblas.pc`, consistently across recent releases.

P.S. Btw, such design might be considered for MPI in the future, as MPI-5 standard defines MPI-Forum ABI; there's already [reference implementation](https://github.com/mpi-forum/mpi-abi-stubs), and both MPICH and OpenMPI are on it (MPICH v5.0.0 and upcoming OpenMPI v6.0.0)

s8ga:
Thanks for the correction — that's a deliberate design I wasn't aware of, and I withdraw suggestion 3. Linking FlexiBLAS itself is clearly the intended path on Fedora, and the absence of openblas.pc is by choice, not omission.

mtaillefumier (approved):

PR #5853, opened 2026-08-25
Enable linking input sections mentioned in docs

he-zilong:
Just like linking input keywords mentioned in docs. A typo in a link is also fixed.

PR #5864, opened 2026-08-27
Misc docs edits and a bug fix for "mentions" links

he-zilong:
This PR:
- Expands `docs/README.md` with recent experience in formatting;
- Expands `docs/technologies/libraries.md` with some links and content;
- Also fixes other typos and rephrases a bit.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -87,13 +87,15 @@ For a all typography options see the
   - This requires adding the citation to the [bibliography.F](../src/common/bibliography.F) and is
     rendered as a link to the [bibliography page](https://manual.cp2k.org/trunk/bibliography.html);
     ⚠️ always use the auto-generated key on the bibliography page for documentation, which may or
-    may not match the customizable key in `CALL add_reference(key=...)` invocations for source codes
+    may not match the name of the variable `key` passed to fortran `CALL add_reference(key=...)`
+    invocations for source codes
 - Another page: `[](../optical/tddft)`
   - Use the relative path, where the `.md` suffix for file extension can be specified or omitted
 - Subsection in the current page: `[](#problems-and-solutions)`
 - Subsection in another page: `[](../optical/tddft.md#periodic-systems)`
-  - ⚠️ Also use the relative path, but the `.md` suffix must be present before the `#` sign; see
-    also [](#cross-references) below
+  - ⚠️ Also use the relative path, but the `.md` suffix must be present before the `#` sign, and
+    only the first three levels of headings have these anchors auto-generated on the final page
+    ready for use; see [](#cross-references) below for an alternative
 - Input section: `[FORCE_EVAL](#CP2K_INPUT.FORCE_EVAL)`
   - This will also generate a "mentions" backlink in the input reference for the section
 - Input keyword: `[STRESS_TENSOR](#CP2K_INPUT.FORCE_EVAL.STRESS_TENSOR)`
@@ -220,7 +222,8 @@ plain display. Details can be found at
 ## Cross References
 
 Apart from using the title or header of a section as introduced in [](#links) above, it is also
-possible to define an explicit target for linking purposes. With the following target
+possible to define an explicit target for linking purposes. The identifier enclosed in parentheses
+is globally active and can be linked internally or externally. For instance, defined in one page,
 
 ```
 (build-gromacs-cp2k)=
@@ -230,7 +233,7 @@ possible to define an explicit target for linking purposes. With the following t
 The latest supported GROMACS release...
 ```
 
-the section can be linked like
+the section can be linked in another page without the need for full relative path.
 
 ```
 Build GROMACS with CP2K QM/MM support following instructions from [](#build-gromacs-cp2k) and ...
--- a/docs/README.md
+++ b/docs/README.md
@@ -68,7 +68,7 @@ support. The following gives a quick overview of the syntax:
 ```
 **bold text**
 
-_italic text_
+_italic text_ (alternatively, *italic text*)
 
 ~~strikethrough text~~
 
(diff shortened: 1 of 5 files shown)


docs/technologies/libraries.md
```diff
@@ -20,16 +20,20 @@ Examples are the sequential or thread variant of the Intel MKL, the Cray libsci,
...
-CP2K toolchain, Spack, and a compatible external DBCSR installation are supported ways to provide
-it.
+{term}`DBCSR` is a standalone library for block-sparse matrix operations. It is maintained at the
+[cp2k/dbcsr](https://github.com/cp2k/dbcsr/) repository, with links to reference materials on
+<https://www.cp2k.org/dbcsr> and documentation on <https://cp2k.github.io/dbcsr/develop/index.html>.
+
+CP2K requires DBCSR as a hard dependency, which can be prepared with the CP2K toolchain or Spack
+build, and will be found automatically by CMake during configuration.
+
+For MPI builds, DBCSR must also have been built with MPI support.
```
Growl1234: Would be better to say that MPI configuration (ON or OFF) must be consistent between CP2K and DBCSR. IIRC if CP2K is built with MPI but DBCSR without MPI, it will most likely cause runtime failure either.

docs/technologies/libraries.md
```diff
@@ -20,16 +20,20 @@ Examples are the sequential or thread variant of the Intel MKL, the Cray libsci,
...
 
 ## DBCSR (required, block-sparse matrix operations)
 
-CP2K requires [DBCSR](https://github.com/cp2k/dbcsr/) for block-sparse matrix operations. It is
-found automatically by CMake. For MPI builds, DBCSR must also have been built with MPI support. The
-CP2K toolchain, Spack, and a compatible external DBCSR installation are supported ways to provide
-it.
+{term}`DBCSR` is a standalone library for block-sparse matrix operations. It is maintained at the
+[cp2k/dbcsr](https://github.com/cp2k/dbcsr/) repository, with links to reference materials on
+<https://www.cp2k.org/dbcsr> and documentation on <https://cp2k.github.io/dbcsr/develop/index.html>.
```
Growl1234: I would keep "DBCSR" linked to its website instead of leading users to acronyms list.

docs/README.md
```diff
@@ -158,6 +202,51 @@ for i in range(10):
...
+```
+````
+
+The language identifiers like `python` determine syntax highlighting; `text` is the choice for a
+plain display. Details can be found at
+[MyST documentation](https://myst-parser.readthedocs.io/en/latest/syntax/code_and_apis.html).
+
+## Cross References
+
+Apart from using the title or header of a section as introduced in [](#links) above, it is also
```
oschuett, line 224: It might be worth emphasizing that these explicit targets are _global identifiers_, meaning one no longer has to prefix them with the relative path. This obviously makes them much easier to move around.

docs/README.md
```diff
@@ -81,30 +81,74 @@ For a all typography options see the
...
+- Publication: `[](#Wilhelm2018)`
+  - This requires adding the citation to the [bibliography.F](../src/common/bibliography.F) and is
+    rendered as a link to the [bibliography page](https://manual.cp2k.org/trunk/bibliography.html);
+    ⚠️ always use the auto-generated key on the bibliography page for documentation, which may or
+    may not match the customizable key in `CALL add_reference(key=...)` invocations for source codes
 - Another page: `[](../optical/tddft)`
-- Subsection in another page: `[](../optical/tddft.md#periodic-systems)` ⚠️ with file extension
+  - Use the relative path, where the `.md` suffix for file extension can be specified or omitted
+- Subsection in the current page: `[](#problems-and-solutions)`
+- Subsection in another page: `[](../optical/tddft.md#periodic-systems)`
```
oschuett, line 95: It might be worth mentioning, that we currently auto-generate these [section anchors](https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#syntax-header-anchors) only for the first three heading levels:

https://github.com/cp2k/cp2k/blob/db4011826da0589fa52d24e7d0bd2a812955f836/docs/conf.py#L26

docs/README.md
```diff
@@ -81,30 +81,74 @@ For a all typography options see the
...
 
 - Acronym: `` {term}`ADMM` ``
-- Publication: `[](#Wilhelm2018)` ⚠️ see note below
+  - This requires adding the definition to the [acronyms.md](./acronyms.md) and is rendered as a
+    link to the [acronyms page](https://manual.cp2k.org/trunk/acronyms.html)
+- Publication: `[](#Wilhelm2018)`
+  - This requires adding the citation to the [bibliography.F](../src/common/bibliography.F) and is
+    rendered as a link to the [bibliography page](https://manual.cp2k.org/trunk/bibliography.html);
+    ⚠️ always use the auto-generated key on the bibliography page for documentation, which may or
+    may not match the customizable key in `CALL add_reference(key=...)` invocations for source codes
```
oschuett:
```suggestion
    may not match the name of the Fortran variable passed to `CALL add_reference(key=...)`.
```

he-zilong:
Thank you for the feedbacks, which I will address shortly.

On an unrelated note, I have just discovered a bug in the generation of the "mentions" link that presumably predates my last #5853. For the following paragraph
https://github.com/cp2k/cp2k/blob/8ddd909fbc89cfdc7b275e0ca702763d6045d73c/docs/methods/sampling/molecular_dynamics.md?plain=1#L20-L27
the links to `STEPS` and `MAX_STEPS` on the same line could not be rendered as a "mentions" backlink on the [corresponding manual page](https://manual.cp2k.org/trunk/CP2K_INPUT/MOTION/MD.html). The links to `MOTION/MD` and `ENSEMBLE` may have been affected as well, though the latter actually has a "mentions" backlink from a later mention. The links for `TIMESTEP` and `TEMPERATURE` that appear on a single line each work as intended.

I suspect that this bug has something to do with the regular expression for collecting links where a `.*` could blur the boundary between two or more links when matching characters.
https://github.com/cp2k/cp2k/blob/8ddd909fbc89cfdc7b275e0ca702763d6045d73c/docs/generate_input_reference.py#L407

oschuett:
> I suspect that this bug has something to do with the regular expression for collecting links where a .* could blur the boundary between two or more links when matching characters.

I think, you're right. That should be made non-greedy: `.*` -> `.*?` 
Good catch!

oschuett (approved):

PR #5866, opened 2026-08-27
Add ADIIS-to-DIIS switching

Growl1234:
Pure ADIIS is known to be less effective close to tight convergence. This PR allows `UPDATE_METHOD ADIIS` to hand over to the existing DIIS implementation as the SCF approaches convergence.

After sufficient valid DIIS history has been collected, one half-weight ADIIS/DIIS step is performed before DIIS takes full ownership. The intermediate step continues to be reported as `ADIIS/Diag.`, while a completed handover is reported as `DIIS/Diag.`.

The implementation reuses the existing Gamma-point and k-point DIIS buffers, residual construction, and solvers. Only states with matching Fock matrices and molecular orbitals are admitted to the DIIS history; in particular, an interior ODA density is excluded. A temporary excursion above `EPS_DIIS` or a rejected new residual no longer discards the existing valid history or causes an abrupt fallback to the raw Fock matrix. Once the handover is complete, ADIIS subspace construction stops and the existing DIIS and k-point OpenMP paths are reused directly.

The implementation covers Gamma-point and k-point calculations, RKS and UKS, and real and complex wavefunctions. The existing non-ADIIS mixing/DIIS paths remain unchanged.


Growl1234:
Here's the NaMnO2 example of completing the optimization task, with ADIIS->DIIS used: [NaMnO2-test-2.zip](https://github.com/user-attachments/files/31514625/NaMnO2-test-2.zip).

PR #5868, opened 2026-08-28
Generalize Wavelet solver to arbitrary radices

fstein93:
The current Wavelet implementation relied on Stefan Goedecker's FFT implementation which has restrictions regarding the available radices. I switched to the generic backend (based on FFTW) to lift this restriction and enable the migration to a new FFT-backend. This required generalization of the calls to 1D-FFTs.
This PR removes `GLOBAL%EXTENDED_FFT_LENGTHS`. Internally, I kept the old list of radices and look for suitable FFT_lengths from it and used the extended list if no suitable radix is found to ensure consistency of results between different CP2K versions and prevent modification of regression tests. This required the change of the reference value of `QS/regtest-fftw/H2-big-1.inp` because the expected number of grid points is too low to require the extended list.
I harmonized naming schemes and pushed some aspects regarding FFT-libraries into the library-backend.


dryndyk (commented):
I made my big NEGF test and it is OK.
Only the message appeared:
`found an unknown keyword EXTENDED_FFT_LENGTHS in section GLOBAL`
I hope we don't need it anymore.

fstein93:
After a rebase, I had some changes to `TEST_FILES.toml` which I removed while forgetting I had to remove `EXTENDED_FFT_LENGTHS` from all files. I am currently fixing everything.

oschuett:
> Only the message appeared: `found an unknown keyword EXTENDED_FFT_LENGTHS in section GLOBAL`

We can mark keywords as _removed_, which then leads to an error message that includes its `deprecation_notice`, e.g.:

https://github.com/cp2k/cp2k/blob/230e667fc4e921275a7946b6a1b52c02ea0254ae/src/start/input_cp2k_motion.F#L1034-L1036

hforbert:
Does this mean fftw3 becomes sort of mandatory, since checks in the code which fft method is being used are removed and fftsg not being able to handle all problem sizes? Or did I overlook/misunderstod something?

fstein93:
Exactly, FFTW3 will become mandatory. We discussed this at the last developer's meeting and a note was made with the last release. This PR does not remove the Stefan-Goedecker-implementation yet but the next one will by replacing the old backend for local FFTs with a new one without the SG-implementation. This new backend will work with all radices although of course not every radix will be fast.

hforbert:
> Exactly, FFTW3 will become mandatory. We discussed this at the last developer's meeting and a note was made with the
> last release.

Not that I am opposed, but I must have missed that... ok.

> This PR does not remove the Stefan-Goedecker-implementation yet but the next one will by replacing the old backend for
> local FFTs with a new one without the SG-implementation. This new backend will work with all radices although of course
> not every radix will be fast.

hmmm, sure. I am not so sure if I like that PR going in while that new backend is not yet in or at least close to.
The new backend will be in the next release for sure?

fstein93:
> > This PR does not remove the Stefan-Goedecker-implementation yet but the next one will by replacing the old backend for
> > local FFTs with a new one without the SG-implementation. This new backend will work with all radices although of course
> > not every radix will be fast.
> 
> hmmm, sure. I am not so sure if I like that PR going in while that new backend is not yet in or at least close to. The new backend will be in the next release for sure?

Yes, this is the plan. I am introducing it subsequently to keep PRs shorter and to reduce merge conflicts from other PRs. It is part of long-term efforts to improve GPU-performance of CP2K's grid operations.

hforbert (approved):

fstein93:
@oschuett Are you approving this PR?

oschuett (approved):

PR #5869, opened 2026-08-28
NNP: Add GPU offload for Behler-Parrinello force evaluation

DhruvSkyy:
Adds GPU offload for CP2K's native NNP force evaluation. The neighbour walk, ACSF descriptors, committee network, force assembly and analytic virial run on the device, with one position H2D and one force D2H transfer per step. The implementation uses CP2K's offload layer and builds for CUDA and HIP.

Existing behaviour is unchanged. `&NNP` gains:

```
USE_GPU AUTO|ON|OFF          ! AUTO is the default
GPU_PRECISION DOUBLE|MIXED   ! DOUBLE is the default
```

`AUTO` uses the GPU only when the build supports it, every rank sees a device, the model fits the compiled device tables, and there are more than 1000 atoms per rank. `MIXED` evaluates the angular symmetry functions in single precision, keeping the radial functions, the network and every neighbour-summing accumulator in double. The existing CPU loop in `nnp_force.F` is unchanged (`git diff -w`).

## Performance

Bulk water with an eight-member committee, comparing one A100-SXM4-80GB with one exclusive 76-core Ice Lake node. Throughput is katom-step/s.

|  atoms | `DOUBLE` |   `MIXED` | 76-core CPU |
| -----: | -------: | --------: | ----------: |
|    192 |     18.8 |      25.7 |        19.4 |
|  3,072 |     77.8 |     106.7 |        44.4 |
| 12,288 |     93.0 |     140.9 |        36.8 |
| 49,152 | **97.8** | **152.7** |    **37.4** |

At 49,152 atoms the GPU is 2.6x faster in double precision and 4.1x in mixed precision. The CPU peaks around 3,072 atoms, while GPU throughput continues to increase, motivating the `AUTO` threshold.

fig_gpu_isocost

The figure compares one to four A100s against the same CSD3 cost in CPU cores, using 55 core-equivalents per A100. Its CPU series is Sapphire Rapids rather than the Ice Lake node in the table above. Error bars are one SD over five submissions.

## Validation

Checked on an A30 against the CPU path. Double precision gives identical step-zero energies at 1, 2 and 4 ranks, a ten-step trajectory agreeing to `1.2e-16`, virials within `5e-11` across 231 stress components, and a 1000-step NVE drift of `8.13e-07 Ha/atom`, identical to CPU. `MIXED` shifts the step-zero energy by `4.5e-11` relative.

`tests/NNP/regtest-gpu` reuses the existing CPU test inputs via `@INCLUDE` and is enabled only when a device is detected. A new `M_PRESSURE` matcher checks the virial. The four-element `data/NNP/NaClaq-cnnp/` model (712 KiB, CC-BY-SA 4.0, attributed) is added to exercise the empty-neighbour-list path; it is plain text, which is why it dominates the line count.

GPU force accumulation uses `atomicAdd`, so runs are not bit-for-bit reproducible and GPU regtests use a `1e-9` tolerance. Energy-only evaluations and helium-solute coupling remain on the CPU, and only one NNP environment per process may own the device. The HIP path compiles with ROCm 6.2 for `gfx90a`, but has not yet been tested on AMD hardware.

cc @cschran


fstein93 (dismissed):

data/NNP/NaClaq-cnnp/LICENSE
```diff
@@ -0,0 +1,24 @@
+The neural-network potential files in this directory (input.nn, scaling.data and the
```
oschuett, line 1: To be on the safe side, you could add SPDX headers and enforce them via the precommit checks:

https://github.com/cp2k/cp2k/blob/fd6b28e74cf4bb4299e1039db0a35b087f233686/tools/precommit/check_file_properties.py#L237-L244

src/cp2k_info.F
```diff
@@ -297,6 +297,13 @@ FUNCTION cp2k_flags() RESULT(flags)
 #if defined(__NO_OFFLOAD_PW)
       flags = TRIM(flags)//" no_offload_pw"
 #endif
+#if (defined(__OFFLOAD) || defined(__OFFLOAD_CUDA) || defined(__OFFLOAD_HIP)) && !defined(__NO_OFFLOAD_NNP)
+      ! __OFFLOAD itself is only defined for C/CUDA units that include
+      ! offload_runtime.h; CMake builds define __OFFLOAD_CUDA/__OFFLOAD_HIP
+      ! globally, so test those here too or the nnp_gpu regtests are
+      ! silently skipped on CMake-built binaries.
+      flags = TRIM(flags)//" nnp_gpu"
+#endif
```
oschuett, line 306: Why don't you follow the existing pattern?

```suggestion
#if defined(__NO_OFFLOAD_NNP)
      flags = TRIM(flags)//" no_offload_nnp"
#endif
```

tests/NNP/regtest-1/H2O-256_C-NNP_MD-bitexact.inc
```diff
@@ -0,0 +1,87 @@
...
+  &END PRINT
+&END MOTION
+
+&FORCE_EVAL
+  METHOD NNP
+  &NNP
+    GPU_PRECISION ${PREC-DOUBLE}
+    NNP_INPUT_FILE_NAME NNP/bulkH2O-jcp2020-cnnp/nnp-1/input.nn
+    SCALE_FILE_NAME NNP/bulkH2O-jcp2020-cnnp/nnp-1/scaling.data
+    USE_GPU ${GPU-OFF}
```
oschuett, line 44: Why don't you just set `USE_GPU AUTO`? Then you wouldn't need to duplicate the test dirs, fiddle with input includes, and also wouldn't have to touch `do_regtests.py`.

tests/do_regtest.py
```diff
@@ -441,11 +441,18 @@ def __init__(self, line: str, cfg: Config):
...
 
-    def requirements_satisfied(self, flags: List[str], mpiranks: int) -> bool:
+    def requirements_satisfied(
+        self, flags: List[str], mpiranks: int, num_gpus: int
+    ) -> bool:
         result = True
         for r in self.requirements:
             if "mpiranks" in r:
                 result &= eval(r.replace("||", " or "))
+            elif r == "gpu":
```
oschuett, line 451: I'd really prefer no to add `gpu` as a requirement flag for test dirs.

There should be one set of tests that works on all architectures.

No only does this avoid unnecessary duplications, it also ensures that CP2K behaves the same across all architectures.

tools/nnp_gpu/README.md
```diff
@@ -0,0 +1,126 @@
+# NNP GPU verification runbook
```
oschuett, line 1: This looks like you're shipping your own test framework. Since this will not be run by our CI, it will probably bit rot rather quickly. I'd strongly advice to integrate your tests with the CP2K-CI.

For example, the grid code has a validation mode builtin that compares the results of performance optimized backends against those of a reference backend:

https://manual.cp2k.org/trunk/CP2K_INPUT/GLOBAL/GRID.html#CP2K_INPUT.GLOBAL.GRID.VALIDATE

https://github.com/cp2k/cp2k/blob/master/tests/QS/regtest-grid/simple_ortho_grid_validate.inp#L5

src/nnp/gpu/nnp_gpu_internal.h
```diff
@@ -0,0 +1,88 @@
...
+
+/*
+ * Warp-level barrier ordering per-warp shared-memory traffic. CUDA provides
+ * the no-argument __syncwarp(); ROCm's device headers do not declare it for
+ * AMD targets, so there the same effect is an acquire-release fence at
+ * wavefront scope plus a wavefront execution barrier, which is how ROCm
+ * defines the intrinsic where it exists.
+ */
+#if defined(__OFFLOAD_HIP) && !defined(__HIP_PLATFORM_NVIDIA__)
+#define NNP_SYNCWARP()                                                         \
```
mtaillefumier: starting ROCM 7.0.0 __syncwarp can be used so it is better to version this macro.

src/nnp/gpu/nnp_gpu_internal.h
```diff
@@ -0,0 +1,88 @@
...
+
+/*
+ * Double-precision atomicAdd is a device intrinsic only from compute capability
+ * 6.0 (Pascal). CP2K still supports older CUDA architectures, so provide the
+ * compare-and-swap fallback for them; the descriptor and force kernels then
+ * call atomicAdd on doubles unconditionally. The integer comparison in the loop
+ * is what makes it terminate when the accumulator holds a NaN. Mirrors the shim
+ * in src/grid/gpu/grid_gpu_internal_header.h.
+ */
+#if defined(__OFFLOAD_CUDA) || defined(__HIP_PLATFORM_NVIDIA__)
```
mtaillefumier: FP atomicAdd is supported in hardware since Pascal. And AMD supports this in hardware as well since Mi250x. I am not even sure the code would compile with any version of cuda that still support K80 and maxwell.

src/nnp/gpu/nnp_gpu_scatter.cu
```diff
@@ -0,0 +1,764 @@
...
+  const double scale =
+      nnp_compute_force_scale(scaling_mode, scmin, scmax, scale_idx,
+                              scaling_loc_max, scaling_loc_min, scaling_sigma);
+  const double de = dEdG * scale;
+
+  /* Self contribution to atom_i: f[i] -= de * self_partial(d, sf, atom_in_e).
+   * r_ii = 0, so the self term makes no virial contribution. */
+  const long long self_off = 3 * (sf + (long long)n_symf * atom_in_e);
+  const long long f_i_off = 3 * (i_global + (long long)num_atoms_global * c);
+  if (threadIdx.x == 0) {
```
mtaillefumier: I doubt these atomic operations are needed. As far I understand the code `f_i_off` only depends on the block index so unless atom_global is not a one-to-one mapping only thread zero will update `f_i_off`.

src/nnp/gpu/nnp_gpu_scatter.cu
```diff
@@ -0,0 +1,764 @@
...
+  /* ang1 neighbours j: f[k_atom] -= de * jj(d, sf, j, atom). */
+  for (int j = threadIdx.x; j < n_ang1; j += blockDim.x) {
+    const long long nb_off = j + (long long)max_n_ang1 * atom_in_e;
+    const int k_atom = ang1_ind_one_based[nb_off] - 1;
+    const long long jj_off = 3 * (sf + (long long)n_symf * nb_off);
+    const long long f_k_off = 3 * (k_atom + (long long)num_atoms_global * c);
+    const double fk0 = -de * jj[jj_off + 0];
+    const double fk1 = -de * jj[jj_off + 1];
+    const double fk2 = -de * jj[jj_off + 2];
+    atomicAdd(&f_dev[f_k_off + 0], fk0);
```
mtaillefumier: I wonder if it is possible to do this without using atomic operations or at least reduce their use. Using them is convenient for sure but they have drawback
- Results are not reproducible which for molecular dynamics can be problematic
- Atomic operations add a uncharacteristic source of noise
- Using them extensively is eventually detrimental for performance

mtaillefumier:
Interesting work. I have few suggestions that could improve clarity. 

- Is it possible to use a structure for passing the arguments to kernels. The reason for this is that each argument of a kernel function takes at least one register so functions with 30 arguments will occupy already a fair amount of registers for holding the same values accross all block threads. Passing a structure reduce this to only a few. Since all these parameters are common to all block threads, shared memory can be used to share this information in one place. 
- Use atomic operations is problematic when used extensively. I really wonder if some of these can be avoided together.

PR #5870, opened 2026-08-28
Basis-set optimizer for frontier states to match HOMO-LUMO gap

JWilhelm:
Add a basis optimizer that changes the contraction coefficients of an initial compact basis to match the occupied and low-energy virtual states of a larger reference basis, with particular focus on the DFT HOMO-LUMO gap.

Two modes are available. The simple optimizer keeps the electron density from the reference calculation fixed. Repeated projections and diagonalizations change the contraction coefficients and thereby improve the HOMO-LUMO gap and the overlap of the occupied and low-energy virtual states with the reference states.

The automatic mode accounts for the strong dependence on the weights in the objective function. It runs five optimizations with predefined weights, then performs a fresh, self-consistent SCF calculation for each optimized basis. It reports SCF-converged total energies and HOMO-LUMO gaps for comparison with the initial compact basis and the large reference basis.

An initial compact basis must be supplied. Current tests use aug-TZVP-MOLOPT-tier2 as the reference and optimize the augmentation-function contractions of aug-SZV-MOLOPT-GTH-tier2. The basis sets are available at:

https://github.com/JWilhelm/BASIS_AUG_MOLOPT

For the systems tested so far, the compact augmented single-zeta basis reproduces the large-basis DFT HOMO-LUMO gap within about 20 meV. Its accuracy can now be tested for GW gaps and GW-BSE excitation energies.

Assistance: ChatGPT 5.6-Sol


src/optbas_frontier_orbital_screening.F
```diff
@@ -0,0 +1,719 @@
...
+      condition_number = 0.0_dp
+      CALL section_vals_duplicate(section, input_file)
+      CALL prepare_frontier_orbital_scf_input(input_file, opt_bas, basis_file)
+      scf_output_unit = -1
+      IF (para_env%is_source()) THEN
+         scf_output_unit = get_unit_number()
+         IF (scf_output_unit < 1) THEN
+            CPABORT("No logical unit is available for disposable screening SCF output")
+         END IF
+         OPEN (UNIT=scf_output_unit, STATUS="SCRATCH", ACTION="READWRITE", IOSTAT=stat)
```
fstein93: I understand why you are using a scratch file. I am not sure how we generally want to deal with them because not all details of their behavior are standardized and may be compiler- and system-dependent (see [here](https://fortranwiki.org/fortran/show/Scratch+Files)). @oschuett What are your thoughts?

JWilhelm:
Maybe just a summary as PDF of the purpose of the optimizer, I will shortly check your comments. 
[Frontier_orbital_basis_optimizer.pdf](https://github.com/user-attachments/files/31566632/Frontier_orbital_basis_optimizer.pdf) (updated)

fstein93:
Please run `make_pretty.sh`. Please accelerate the new regtest.

src/optbas_frontier_orbital_screening.F
JWilhelm: @fstein93 Thanks a lot for taking the time for reviewing this PR and for spotting this unnecessarily complicated implementation. I will rework this by calling the DFT calculations in a different way.

he-zilong:
Dear JWilhelm,

I would like to learn about some conceptual and implementation details regarding this PR:

- Does the "DFT HOMO-LUMO gap" as the optimization target arise from the usual one-particle KS-DFT MO eigenvalues? As these are in fact not really physical observables for a many-body system, would it be a better idea to target some other measurable properties like ionization potential, electron affinity and optical excitation energy (for GW quasiparticles)? 
- Is the calculation of "DFT HOMO-LUMO gap" implemented for arbitrary combinations of diagonalization/OT algorithm, k-point/gamma-only sampling for periodic systems, and uniform/fractional occupations?
- Since during the optimization the DFT functional is constant, and it is well-known that the HOMO-LUMO gap of a system from a KS-DFT calculation varies a lot depending on the Hartree-Fock exchange (HFX) fraction, is it sensible to use a pure functional like PBE rather than a hybrid one?

Thank you for introducing this functionality to CP2K.

Regards,
Zilong

JWilhelm:
Dear Zilong,

thanks a lot for your interest in this!

> Does the "DFT HOMO-LUMO gap" as the optimization target arise from the usual one-particle KS-DFT MO eigenvalues? As these are in fact not really physical observables for a many-body system, would it be a better idea to target some other measurable properties like ionization potential, electron affinity and optical excitation energy?

Yes, this is the usual one-particle Kohn-Sham eigenvalue gap. I fully agree that Kohn-Sham eigenvalues are not observables. Indeed, the optimized basis sets are ultimately intended for GW and GW-BSE calculations. The GW quasiparticle energies are obtained schematically as

εₙᴳᵂ = εₙᴰᶠᵀ + Σₙᴳᵂ − Vₙˣᶜ.

 εₙᴰᶠᵀ already contain, for example, the kinetic energy. The corresponding matrix elements can be very sensitive to the Gaussian basis because the kinetic-energy operator contains second derivatives. Basis-set errors at the DFT level therefore also affect the subsequent GW calculation.

The optimizer also does not target only the gap. It matches the occupied and low-energy virtual orbitals to those obtained with a large reference basis. 

During the optimization, the converged large-basis density is kept fixed. For every trial contraction, only the Kohn-Sham matrix has to be recomputed and diagonalized. This is very cheap even for systems with a few hundred atoms. A GW or GW-BSE calculation in every iteration would be orders of magnitude more expensive and also numerically less stable. Similarly, targeting delta-SCF ionization potentials or electron affinities would require a new SCF calculation in every iteration instead of just diagonalizing the KS matrix.

> Is the calculation of the "DFT HOMO-LUMO gap" implemented for arbitrary combinations of diagonalization/OT algorithm, k-point/gamma-only sampling for periodic systems, and uniform/fractional occupations?

Standard diagonalization and restricted, closed-shell Γ-point calculations are currently supported. Γ-only calculations also work for periodic systems. OT has not yet been tested, and general k-point sampling is not implemented. Fractional occupations and metals have also not yet been tested. I have put a periodic Γ-point regression test on my list.

> Since during the optimization the DFT functional is constant, and it is well-known that the HOMO-LUMO gap of a system from a KS-DFT calculation varies a lot depending on the Hartree-Fock exchange fraction, is it sensible to use a pure functional like PBE rather than a hybrid one?

I fully agree that hybrid functionals should also be supported. This is not yet implemented, but I have put it on my list. The hybrid SCF calculation would need to be converged only once in the large reference basis. During the basis optimization, the density would again remain fixed, and no new SCF calculation would be needed for each trial contraction. The optimization should therefore remain quite cheap.


The workflow we ultimately have in mind is to perform molecular dynamics for a liquid, a solvated ion, or a liquid on a surface and then calculate GW-BSE optical spectra for selected snapshots. For accurate yet cheap GW-BSE calculations, we need a small basis that gives an accurate DFT HOMO-LUMO gap and accurate frontier orbitals in each snapshot. The basis optimization targeting both objectives should be much cheaper than the following GW-BSE calculation.

This is only a first commit. There will certainly be further improvements, and I am also looking into additional refactoring of the code.

I hope this answers your questions. If you have any further suggestions I will be happy to discuss!

Best wishes,

Jan

RitajTyagi (approved):

he-zilong:
Dear Jan,

Your explanation is very comprehensive and insightful, and I really appreciate the ongoing development. GW-BSE optical spectra for condensed state matters from molecular dynamics simulation would surely benefit from the speedup introduced by some basis set reduction, and tuning coefficients towards a more accurate description for frontier MOs could be an important first step. With this context in mind I am clearer about the motivation and execution in this PR.

The merge conflict is now blocking the PR due to a non-linear git history; could you please [rebase and resolve conflicts](https://github.com/cp2k/cp2k/wiki/Git-Tips-&-Tricks) instead?

Regards,
Zilong

fstein93 (approved):

PR #5872, opened 2026-08-28
NEGF: Transmission and DOS printing minor changes

dryndyk:


src/input/cp_output_handling.F
```diff
@@ -290,13 +290,13 @@ SUBROUTINE cp_print_key_section_create(print_key_section, location, name, descri
       my_filename = ""
       IF (PRESENT(filename)) my_filename = filename
       CALL keyword_create(keyword, __LOCATION__, name="FILENAME", &
```
dryndyk, line 292: I made a few changes to the description of "FILENAME". u->U, etc., at the beginning of a sentence. Is it OK?

Growl1234: Definitely.

PR #5874, opened 2026-08-28
QS: Enable fixed-occupation OT for complex K points

DCM-Uni-Paderborn:
### Summary

This PR enables a deliberately restricted first Orbital Transformation path for complex,
symmetry-reduced K-point calculations with fixed occupations.

The implementation introduces one OT state per `(spin, irreducible K point)`, carries both real
and imaginary orbital components through the REF/IRAC chart and the CG minimizer, and rebuilds the
existing K-point density from the resulting orbitals. The Gamma-point OT path and its defaults are
unchanged.

The path is method-independent at the Quickstep level. It is covered by DFT and UKS regressions as
well as periodic tblite GFN1 and GFN2 regressions. Smearing and the coupled fixed-N Mermin problem
are intentionally excluded from this review boundary.

### Post-merge correction

Testing with a realistic 39-alpha/33-beta NaMnO2 case exposed two initialization bugs that the
small original regressions did not cover: global Aufbau initialization could populate padded beta
columns, and a restarted density could produce false one-step convergence before it was synchronized
with the fixed-rank orbital projector. Both are corrected in follow-up #5875. That follow-up also
rejects analytical forces and stress explicitly until a complex OT-consistent energy-weighted
density matrix is implemented. The original one-step Li result below is therefore historical and
must not be read as validation of general restart synchronization.

### Complex REF chart and derivative

Each OT environment is labelled by spin, global irreducible K-point index, local K-point index,
and K-point weight. Consistency checks validate the flat channel mapping before state allocation
and before each minimizer call.

For a complex REF coordinate `X`, the physical orbitals are obtained from the polar retraction

```math
\begin{aligned}
M &= X^\dagger S X, \\
F &= M^{-1/2}, \\
Q &= XF, \qquad Q^\dagger S Q = I.
\end{aligned}
```

The complex implementation retains `F`, both components of `S X`, and all real/imaginary CG
vectors. For a physical orbital gradient `G_C` with uniform fixed occupations, its exact pullback
to the finite REF coordinate is

```math
G_X = \left[G_C - S Q\left(Q^\dagger G_C\right)\right]F.
```

This includes the finite-retraction factor that is absent from the infinitesimal projected
gradient. A DBCSR-backed central finite-difference unit test obtains an error of `2.67e-10`; the
discarded projected expression differs by `2.12e-1` on the same fixture.

The shared CG line search treats all local `(spin,k)` channels as one optimization vector. Scalar
products, norms, descent checks, Polak-Ribiere updates, and accepted REF rebases include both real
and imaginary components.

### K-point driver and occupations

For every local channel the driver constructs complex `H(k) C(k)`, applies the fixed occupation
and irreducible K-point weight, evaluates the REF pullback, and writes the updated physical
orbitals back to the K-point MO sets. The existing K-point density transformation then reconstructs
the real-space density, including the overlap information used by symmetry phase calibration.

Only the active occupied width is optimized for each spin channel. This is important for UKS,
where K-point MO storage is padded to a common width although alpha and beta can have different
numbers of occupied orbitals. Padded columns must remain empty. Fixed occupations are initialized
once and are not recomputed from stale auxiliary eigenvalues during OT.

If a fresh guess leaves the K-point MO arrays empty, a one-time generalized diagonalization of the
core Hamiltonian initializes a complex orthonormal subspace. A spin-independent core Hamiltonian
is copied to the remaining UKS start channels before their spin-dependent KS matrices are built.

### K-point restart and xTB/tblite

CP2K `.kp` restart files contain real-space K-point density matrices rather than MO coefficients.
For `SCF_GUESS RESTART`, the implementation therefore builds the KS operator from the restarted
density and diagonalizes it once before allocating the OT state. This avoids combining a valid
restart density with unrelated MOPAC or atomic guess orbitals. CG and line-search history are not
serialized.

With tblite, SCC variables are updated directly from the accepted OT density and
`XTB/SCC_MIXER` is ignored, matching the existing Gamma-point OT behavior. Smearing remains
explicitly rejected.

### Deliberate restrictions

The enabled path currently requires:

- `SMEAR OFF` and no `ADDED_MOS`;
- `OT%ALGORITHM IRAC`;
- `OT%MINIMIZER CG`;
- `OT%PRECONDITIONER NONE`;
- `ROTATION OFF`;
- `OCCUPATION_PRECONDITIONER OFF`;
- at least one occupied orbital in every spin channel;
- one K-point parallel group.

These guards keep fractional occupations, alternative minimizer histories, complex
preconditioners, and the global/subgroup communicator problem outside this first review boundary.
They can be relaxed independently with focused numerical evidence.

### Validation

- GCC 16 MPI/OpenMP/tblite Release build of `cp2k.psmp` and the new unit passes at exact revision
  `027cb2ddbf`.
- No-cache CP2K precommit passes for all 19 changed files.
- The complex REF finite-difference unit passes with MPI 1, 2, and 4.
- `QS/regtest-kp-3` passes 21/21 with MPI 2, including RKS, UKS, and a dependent `.kp` restart.
- The existing Gamma-point `QS/regtest-ot-2` group passes 37/37 with MPI 2.
- `xTB/regtest-tblite-gfn1` passes 9/9 and `xTB/regtest-tblite-gfn2-1` passes 25/25 with MPI 2.
- The symmetry-reduced `3x3x3` He regression produces four irreducible K points and converges in
  103 OT steps at `-2.894663581372875 Ha`. A tight generalized-diagonalization reference differs
  by about `1.8e-11 Ha`.
- The original small UKS Li writer converges in nine steps and its `.kp` reader reported one-step
  convergence. The larger NaMnO2 reproducer later showed that this was insufficient coverage;
  #5875 adds fixed-rank occupation restoration, density synchronization, and a corrected dependent
  writer/restart regression.
- Periodic tblite GFN1 and GFN2 on four irreducible K points converge in 45 and 75 OT steps at
  `-14.717763702261747` and `-13.656523595025897 Ha`. Tight OT/diagonalization comparisons agree
  within `3.1e-10` and `3.7e-12 Ha`, respectively.
- With a consistent GNU OpenMP/OpenBLAS runtime, the same GFN1 and GFN2 OT inputs pass with two
  OpenMP threads at `-14.717763702261742` and `-13.656523595025856 Ha`. The pre-existing smeared
  GFN2 K-point diagonalization control also passes at `-13.656639366111945 Ha`.

### Follow-up work

The next reviewable layer is a productive complex K-point preconditioner so this foundation is not
limited to `PRECONDITIONER NONE`. The next scientific layer is fixed-electron-number Mermin OT with
fractional occupations and a dense chemical-potential projection. It must define the coupling of
orbital rotations, auxiliary energies, and occupation curvature before K-point smearing is enabled.

`ADDED_MOS AUTO`, alternative minimizers, K-point-group parallelism, minimizer-history restart,
forces, and stress remain separate validation steps.


DCM-Uni-Paderborn:
The first `pdbg` run failed during GCC 15.2 compilation because `allocate_qs_ot_kpoint_state` retained an unused `scf_control` dummy argument under `-Werror=unused-dummy-argument`.

Commit b193f1bd58 removes the argument from its sole call and routine interface; there is no numerical change. Local validation on the new head:

- `qs_scf.F` compiles with `-Werror=unused-dummy-argument`
- no-cache precommit passes for `src/qs_scf.F`
- complex REF unit passes with MPI 1, 2, and 4
- `QS/regtest-kp-3` passes 21/21 with MPI 2

Fresh CI has started.

Growl1234:
I tested the new K-point OT path on a more realistic UKS NaMnO2 case (39 alpha / 33 beta electrons, 39 stored MOs per spin, with `FULL_KINETIC`) and ran into two initialization-related issues.

1. With an `ATOMIC` initial guess, the calculation aborts before the first OT iteration with

```text
Spin 1:
  Number of electrons:          39
  Number of occupied orbitals:  39
  Number of molecular orbitals: 39

Spin 2:
  Number of electrons:          33
  Number of occupied orbitals:  33
  Number of molecular orbitals: 39

[ABORT] Fixed-occupation K-point OT cannot optimize occupied padded MO columns.
```

The six extra beta-spin columns are expected storage padding for UKS, but the initial K-point occupation assignment appears able to populate such padded columns before the fixed-rank OT state is created. This makes a fresh fixed-occupation UKS calculation fail before OT is actually attempted. It may be worth checking whether fixed-occupation K-point OT should derive its active occupations from this initial global Aufbau filling, or instead preserve a consistent fixed occupied rank at every k point. (So looks like the "k-point driver and occupations" section of the description isn't perfectly resolved.)

2. With `SCF_GUESS RESTART`, the same calculation gets past that check, but the first OT iteration immediately reports essentially zero convergence and terminates:

```text
1 OT CG  0.15E+00  5.8  1.5458E-16  -367.7908710545
*** SCF run converged in 1 steps ***
```

However, the state is clearly not self-consistent:

```text
Electronic density on regular grids:        -70.7724908313
Total charge density on r-space grids:         1.2275091687

SCF Total energy:                           -367.79087105452089
ENERGY| Total FORCE_EVAL energy [hartree]   -242.712161204568872
```

The ~125 Ha difference between the reported converged SCF energy and the immediately following `FORCE_EVAL` energy is particularly suspicious. My reading is that the restart density is used to build and diagonalize the initial KS operator, which naturally makes the initial OT orbital gradient nearly zero, while the density reconstructed from the resulting fixed-occupation orbitals is not necessarily identical to the restart density. If so, the OT convergence test can succeed before the orbital-derived density, KS matrix, and energy have been synchronized.

Could you take a look at the initialization sequence for these two cases? I can provide the input and both outputs as reproducer files: [NaMnO2-test.zip](https://github.com/user-attachments/files/31587789/NaMnO2-test.zip)

Growl1234:
And do you have plan to add a complex analogue of `calculate_w_matrix_ot()`? The stress tensor calculated by k-point OT seems problematic: [Bi2Se3-opt-test.zip](https://github.com/user-attachments/files/31588077/Bi2Se3-opt-test.zip)

[GPT told me that the k-point branch of `compute_matrix_w()` always constructs the energy-weighted density matrix from MO eigenvalues, whereas the Gamma-point path has a dedicated OT construction; the k-point OT driver normally does not update those eigenvalues as the occupied orbitals evolve.]

DCM-Uni-Paderborn:
Thanks for the realistic reproducers. Your diagnosis of both initialization failures was correct, and the fixes are now in the follow-up PR #5875 at `fd3652dbfb`.

For NaMnO2, the global Aufbau pass is retained for eigenvalue and chemical-potential bookkeeping, but fixed-occupation OT now restores the prescribed occupied rank independently for every K point and spin before constructing the OT state. This preserves 39 alpha and 33 beta occupied columns rather than allowing the six padded beta columns to enter the active projector.

There was also a second, separate synchronization bug in the restart path. The initial Hamiltonian could correspond to the restart density while the subsequent fixed-rank orbital projector represented a different density. K-point OT now performs an explicit projector-to-density synchronization before starting its first line search, for both fresh and restarted calculations.

With the supplied NaMnO2 setup restricted to two diagnostic SCF steps:

- the fresh run reaches an electronic density of `-71.9999999998` and charge `2e-10` instead of aborting on padded beta columns;
- the restarted run reaches `-71.9999999979` and charge `2.1e-9`;
- the restart performs the synchronization and then a genuine OT step with residual `0.01449533`, rather than falsely converging in one step at `1.55e-16`.

The energies of these deliberately two-step diagnostics are not convergence references. The committed open-shell UKS writer/restart regression instead passes as part of `QS/regtest-kp-3` 23/23.

For the Bi2Se3 stress issue, I also agree with the diagnosis. A correct implementation needs the complex K-point analogue of `calculate_w_matrix_ot`; reusing auxiliary eigenvalues that are not evolved by OT is not valid. That complex W construction is not small enough to hide in this preconditioner follow-up. Until it is implemented and validated, #5875 now rejects analytical forces and stress for complex K-point OT explicitly. A production smoke reaches SCF convergence and then stops with the intended message instead of returning a misleading tensor. Gamma OT and ordinary K-point diagonalization are unchanged.

PR #5875, opened 2026-08-28
QS: Add preconditioners for complex K-point OT

DCM-Uni-Paderborn:
### Summary

This follow-up to #5874 adds explicit complex-Hermitian orbital preconditioners to the
fixed-occupation K-point OT path. It supports `FULL_S_INVERSE`, `FULL_KINETIC`, and
`FULL_SINGLE_INVERSE` in addition to `NONE`. The occupied-space-dependent
`FULL_SINGLE_INVERSE` remains local to each `(spin, irreducible K point)` channel, while the
spin-independent overlap and kinetic operators are stored once per irreducible K point and shared
between UKS spin channels.

The real and imaginary components of each DBCSR orbital-gradient block are combined into one
distributed complex matrix, multiplied by the inverse operator, and split back into the existing
real/imaginary OT state. The Gamma-point path and all existing defaults are unchanged.

### Complex inverse operators

For overlap and kinetic preconditioning, the channel-local inverse operators are

```math
\begin{aligned}
P_S(\mathbf{k}) &= S(\mathbf{k})^{-1}, \\
P_T(\mathbf{k}) &=
\left[T(\mathbf{k}) + \max(0,\Delta)S(\mathbf{k})\right]^{-1}.
\end{aligned}
```

The implementation follows the existing real Gamma-point definitions and stores an explicit
distributed complex-Hermitian inverse obtained through Cholesky factorization.

For `FULL_SINGLE_INVERSE`, let

```math
C_0(\mathbf{k})^\dagger S(\mathbf{k})C_0(\mathbf{k}) = I,
\qquad
H_{oo}(\mathbf{k}) = C_0(\mathbf{k})^\dagger H(\mathbf{k})C_0(\mathbf{k}).
```

The gauge-covariant direct operator is

```math
\begin{aligned}
A_0(\mathbf{k}) ={}& H(\mathbf{k}) \\
&- 2S(\mathbf{k})C_0(\mathbf{k})H_{oo}(\mathbf{k})
   C_0(\mathbf{k})^\dagger S(\mathbf{k}) \\
&+ S(\mathbf{k})C_0(\mathbf{k})C_0(\mathbf{k})^\dagger S(\mathbf{k}).
\end{aligned}
```

This expression is invariant under the occupied-space gauge transformation
$C_0(\mathbf{k}) \mapsto C_0(\mathbf{k})U$. A generalized spectral estimate supplies a
non-negative overlap shift $\sigma(\mathbf{k})$ that floors the lower edge by the configured
`ENERGY_GAP`. The stored preconditioner is therefore

```math
P_{\mathrm{FSI}}(\mathbf{k}) =
\left[A_0(\mathbf{k}) + \sigma(\mathbf{k})S(\mathbf{k})\right]^{-1}.
```

Only the lowest generalized eigenpair of $A_0(\mathbf{k})$ is requested through the indexed
ScaLAPACK subset solver. This avoids the previous full complex spectrum calculation in parallel
builds while retaining the same shift definition. Serial builds retain the general dense fallback.

The total objective and its gradient contain the irreducible K-point weight $w_{\mathbf{k}}$.
After applying a channel-local inverse-Hessian approximation, the search direction is consequently
scaled by $w_{\mathbf{k}}^{-1}$. Scalar products and line-search derivatives continue to use the
weighted global OT objective from #5874.

### Input contract and safeguards

The complex K-point path now accepts:

- `PRECONDITIONER NONE`;
- `PRECONDITIONER FULL_S_INVERSE`;
- `PRECONDITIONER FULL_KINETIC` for non-semiempirical Quickstep methods;
- `PRECONDITIONER FULL_SINGLE_INVERSE`;
- `PRECOND_SOLVER DEFAULT` or `INVERSE_CHOLESKY` for the inverse operators.

Unsupported preconditioners and solver combinations abort before OT-state construction. In
particular, the distinct `FULL_ALL` and `FULL_SINGLE` operator representations and the `DIRECT`,
`UPDATE`, and `CHEBYSHEV` solver modes are not silently mapped onto the new complex inverse.
`FULL_KINETIC` retains the existing restriction for semiempirical methods.

The restrictions inherited from #5874 remain: fixed occupations, `ALGORITHM IRAC`, `MINIMIZER CG`,
`ROTATION OFF`, no occupation preconditioner, one K-point parallel group, no ROKS or empty spin
channel. Analytical forces and stress are now rejected explicitly because complex K-point OT still
lacks an OT-consistent complex energy-weighted density matrix; this prevents the evolving OT
orbitals from being combined silently with stale auxiliary eigenvalues.

This follow-up also fixes two initialization problems exposed by an open-shell NaMnO2 reproducer.
After the global Aufbau bookkeeping has initialized eigenvalues and the chemical potential,
fixed-occupation OT restores the prescribed occupied rank independently at every K point. Before
the first line search it then rebuilds the density from that fixed-rank orbital projector. The same
synchronization is performed after reading a K-point wavefunction restart, preventing a stale
restart density from producing a false one-step convergence result.

### Validation

- The exact source tree builds with GCC 16, MPI, OpenMP, and tblite in Release mode. The changed
  dependency closure also compiles with the strict Debug warnings and runtime checks; the local
  Apple ARM Debug link is blocked only by the existing unsupported `clang++ -fsanitize=leak`
  combination.
- No-cache CP2K precommit passes for all changed production and test files; `git diff --check` is
  clean.
- The complex REF/preconditioner unit passes with MPI 1, 2, and 4. Independently diagonalized dense
  references for `FULL_S_INVERSE` and `FULL_KINETIC` differ from the distributed inverses by
  $4.44\times10^{-16}$ and $4.44\times10^{-16}$; the `FULL_SINGLE_INVERSE` gauge error is
  $4.44\times10^{-16}$.
- `QS/regtest-kp-3` passes 23/23 with MPI 2, including symmetry-reduced DFT production
  paths for `FULL_S_INVERSE` and `FULL_KINETIC`, UKS, and K-point density restart.
- The existing Gamma-point `QS/regtest-ot-2` group passes 37/37.
- The full tblite GFN1 and GFN2 regression groups pass 9/9 and 25/25 with MPI 2. Their K-point OT
  cases converge with `FULL_SINGLE_INVERSE` in 22 and 20 steps at
  $-14.7177637169713$ Ha and $-13.6565236146731$ Ha. The test-local `STEPSIZE 0.05` removes
  compiler/MPI-sensitive initial line-search branching without changing the user default.
- The two DFT preconditioner cases also pass with MPI 4 and reproduce the MPI-2 energies within
  $4\times10^{-15}$ Ha.
- At `EPS_SCF 1e-9`, the same He problem converges with `NONE`, `FULL_S_INVERSE`, `FULL_KINETIC`,
  and `FULL_SINGLE_INVERSE` in 167, 21, 45, and 59 steps. All four final energies agree within
  $3\times10^{-15}$ Ha. These small runs validate consistency rather than establish a general
  performance ranking.
- Tight `EPS_SCF 1e-8` GFN1 and GFN2 comparisons agree between `NONE` and
  `FULL_SINGLE_INVERSE` within $1.6\times10^{-12}$ Ha and $1.3\times10^{-12}$ Ha, respectively.
- Explicit negative smokes verify early rejection of semiempirical `FULL_KINETIC` and
  `PRECOND_SOLVER DIRECT`; `INVERSE_CHOLESKY` completes the DFT production path normally.
- In the 39-alpha/33-beta NaMnO2 reproducer, both fresh and restarted runs reconstruct
  $71.999999998$ or more of the expected 72 electrons. The restart now performs an explicit density
  synchronization followed by a genuine OT step with a nonzero residual, instead of reporting
  convergence at $1.55\times10^{-16}$ with only 70.77249 electrons.
- In a 72-channel NaMnO2 initialization benchmark, requesting only the lowest generalized
  eigenpair reduces `make_complex_full_single_inverse` from about 0.384 s to 0.298 s and the full
  K-point preconditioner setup from about 0.523 s to 0.432 s on the local MPI-2 build. These are
  construction-kernel measurements, not a claim about end-to-end SCF acceleration.

### Follow-up work

The next independent layer is complex minimizer and chart support beyond the current IRAC/CG
contract. DIIS and L-BFGS require complex channel-resolved history, restart, and accepted-step
validation; `FULL_ALL` requires a state-selective complex apply model rather than the single
column-independent inverse used here.

The subsequent scientific layer remains fixed-electron-number Mermin OT: fractional occupations,
self-consistent chemical potential, `ADDED_MOS AUTO`, orbital/auxiliary-energy coupling, and the
occupation preconditioner. Those changes remain on the metallic research branch and are not part of
this preconditioner PR.


Growl1234:
I have two performance concerns which I think maybe worth benchmarking:

- `FULL_SINGLE_INVERSE` currently performs a full complex generalized eigensolve for each `(spin, k)` channel merely to estimate the spectral shift, whereas the Gamma-point implementation uses an extremal-eigenvalue Arnoldi solve.
- `FULL_S_INVERSE` and `FULL_KINETIC` appear to construct and store identical preconditioners separately for each spin in UKS, even though these operators are spin-independent and could potentially be shared per k-point.

Growl1234:
There seem to be also some other problems regarding #5874 and this PR; I will test several cases against my system. Most likely they will finally be present here or as a tracker issue these days.

DCM-Uni-Paderborn:
Thanks for pointing these out. Both performance issues are addressed in `fd3652dbfb`.

For `FULL_SINGLE_INVERSE`, the large generalized problem now requests only the lowest eigenpair through the indexed ScaLAPACK subset solver. The shift definition and the serial dense fallback are unchanged. On the same local MPI-2 NaMnO2 initialization with 72 `(spin,k)` channels, `make_complex_full_single_inverse` dropped from about 0.384 s to 0.298 s, and total K-point preconditioner setup from about 0.523 s to 0.432 s. These are kernel timings rather than an end-to-end SCF speedup claim.

For UKS, `FULL_S_INVERSE` and `FULL_KINETIC` now construct and own one complex inverse per irreducible K point. The second spin channel aliases that spin-independent operator with explicit ownership tracking, while `FULL_SINGLE_INVERSE` remains separate because it depends on the occupied spin subspace. MPI-2 UKS production runs with both shared operators and clean teardown pass; the committed UKS regression now exercises `FULL_KINETIC`.

The same commit also handles the two correctness reports on #5874: fixed occupied ranks are restored after the initial global Aufbau bookkeeping, the orbital-derived density is synchronized before the first OT line search for fresh and restarted runs, and analytical forces/stress now abort explicitly until a complex OT-consistent energy-weighted density matrix exists.

Validation on the pushed head includes the complex unit with MPI 1/2/4, `QS/regtest-kp-3` 23/23, Gamma `QS/regtest-ot-2` 37/37, tblite GFN1 9/9, tblite GFN2 25/25, no-cache precommit, and a clean Release build.

fstein93 (approved):

PR #5877, opened 2026-08-29
QS: Fix finite rotations in REF OT

DCM-Uni-Paderborn:
## Summary

Complete the finite orbital-rotation path for the REF representation used by
`ALGORITHM IRAC`.

For a REF coordinate $X$ and an antisymmetric rotation generator $A$, the
physical orbitals are

$$
C(X,A) = Q(X)\,U(A), \qquad U(A) = \exp(A), \qquad A^\mathrm{T}=-A.
$$

Previously, REF OT accepted `ROTATION`, but it did not consistently implement
this product map:

- $U$ was not initialized before the first REF gradient evaluation.
- REF returned $Q(X)$ instead of the physical orbitals $Q(X)U(A)$.
- The orbital gradient was projected without the pullback
  $G_Q = G_C U^\mathrm{T}$.
- The finite exponential adjoint used work matrices allocated only for the
  STRICT representation.

As a result, REF rotations could be ineffective or use undefined state,
particularly for non-equivalent orbitals.

## Implementation

- Initialize the product chart with $A=0$ and $U=I$.
- Apply the finite rotation to the orbitals returned by REF.
- Pull the physical orbital gradient back into the unrotated REF basis.
- Evaluate the existing finite rotation Frechet adjoint without relying on
  STRICT-only work matrices.
- Add a real DBCSR finite-difference regression for the rotation derivative.

The existing `o2_ot_refine_roks.inp` test now exercises the rotation that was
previously ineffective. Its four-step reference energy changes intentionally
from `-31.33473048057817` to `-31.34843768570860` Ha.

This PR does not add smearing or fractional occupations. It fixes the finite
REF product chart needed by orbital-dependent and future Mermin OT paths.

## Validation

- GNU 16 MPI/OpenMP/tblite Release build
- `qs_ot_complex_ref_unittest.psmp`, MPI-1 and MPI-2
  - real finite-rotation slope error: `2.48E-10`
  - existing complex REF pullback error: `2.67E-10`
- `QS/regtest-ot-refine`, MPI-2: `10/10`
- `o2_ot_refine_roks.inp`, serial/MPI-2 agreement: `4E-15` Ha
- no-cache precommit: `4/4`

### DIFF
--- a/src/qs_ot.F
+++ b/src/qs_ot.F
@@ -15,11 +15,12 @@
 MODULE qs_ot
    USE arnoldi_api,                     ONLY: arnoldi_extremal
    USE cp_dbcsr_api,                    ONLY: &
-        dbcsr_add, dbcsr_copy, dbcsr_distribution_type, dbcsr_filter, dbcsr_get_block_p, &
-        dbcsr_get_info, dbcsr_get_occupation, dbcsr_init_p, dbcsr_iterator_blocks_left, &
-        dbcsr_iterator_next_block, dbcsr_iterator_start, dbcsr_iterator_stop, dbcsr_iterator_type, &
-        dbcsr_multiply, dbcsr_release, dbcsr_release_p, dbcsr_scale, dbcsr_set, dbcsr_transposed, &
-        dbcsr_type
+        dbcsr_add, dbcsr_copy, dbcsr_create, dbcsr_distribution_type, dbcsr_filter, &
+        dbcsr_finalize, dbcsr_get_block_p, dbcsr_get_info, dbcsr_get_occupation, dbcsr_init_p, &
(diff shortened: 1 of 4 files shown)


DCM-Uni-Paderborn:
Fixed the `Regtest pdbg` failure in 9bb99507cf. The real finite-rotation Frechet derivative no longer assumes identical sparse DBCSR block patterns for its real and imaginary intermediates; it constructs their explicit union and treats missing partner blocks as zero. The previously failing `QS/regtest-ot-refine/o2_ot_refine_roks.inp` now completes with MPI-2 at the unchanged reference energy `-31.348437685708596` Ha. The finite-rotation unit passes serial and MPI-2, the Release build passes, and precommit passes.

PR #5883, opened 2026-08-29
QS: Enable FULL_ALL preconditioning for complex K points

DCM-Uni-Paderborn:
## Summary

Enable the state-selective `FULL_ALL` OT preconditioner for complex K-point
channels without discarding the imaginary parts of the Hamiltonian, overlap,
orbitals, or gradient.

The implementation:

- constructs one complex spectral preconditioner per `(spin, kpoint)` channel,
- keeps the current OT reference orbitals in their existing gauge,
- applies column-resolved inverse-Hessian weights,
- supports the normal `STRICT` chart as well as the IRAC product chart,
- permits complex K-point `ROTATION` while leaving the Gamma-point restriction
  unchanged, and
- keeps `PRECOND_SOLVER DEFAULT` as the only supported `FULL_ALL` solver.

## Method

For reference orbitals $C$ with $C^\dagger S C=I$, define

$$
Q = I - C C^\dagger S .
$$

The preconditioner diagonalizes the Hermitian generalized eigenproblem

$$
\widetilde H W = S W \Lambda ,
\qquad
\widetilde H =
Q^\dagger H Q
+
S C\,\mathrm{diag}(\varepsilon_i-L)\,C^\dagger S ,
$$

where $L=10$ Ha separates the retained reference space from the
complement. After diagonalization, the first reference columns and their
energy labels are restored exactly.

For orbital-gradient column $G_j$, the applied inverse-Hessian
approximation is

$$
Y_j =
W\,
\mathrm{diag}\!\left[
\frac{1}{\max(\Delta,\Lambda_a-\varepsilon_j)}
\right]
W^\dagger G_j ,
$$

with the configured `ENERGY_GAP` $\Delta>0$. The same complex routine is
used by the DBCSR production path and by the unit test.

## Scope

This PR covers fixed-occupation complex K-point OT only. It does not enable
smearing, Mermin optimization, `ENERGIES`, or `ADDED_MOS`; the existing
fixed-occupation `ADDED_MOS` guard remains unchanged. It also does not change
the default OT algorithm, minimizer, or preconditioner.

The branch is stacked on #5880 and therefore also includes its dependency
#5879 until those PRs merge. The `FULL_ALL` change itself is one commit.

## Validation

- Full GCC 16 MPI/OpenMP/tblite build.
- No-cache precommit: 9/9 changed files.
- Complex preconditioner/OT unit test with MPI 1/2/4:
  - phase-covariance error: `2.22E-16`,
  - $S$-orthonormal spectral-basis error: `2.33E-19`,
  - positive metric descent: `5.824752E-01`.
- Symmetry-reduced DFT He:
  - `STRICT + CG + FULL_ALL`: 24 steps,
    `-2.894663227823732` Ha;
  - `IRAC + ROTATION + LBFGS + FULL_ALL`: 24 steps,
    `-2.894663581384705/-2.894663581384704` Ha with MPI 1/2.
- Additional production checks:
  - `IRAC + ROTATION + CG`: 24 steps;
  - `IRAC + ROTATION + DIIS`: 66 steps;
  - tblite GFN1 Si: 24 steps, `-14.717763718144855` Ha;
  - tblite GFN2 Si: 20 steps, `-13.656523614701575` Ha.


Growl1234:
Hi Thomas,

The NaMnO2 test with `FULL_ALL` preconditioner encounters `CPASSERT` failure: [NaMnO2.zip](https://github.com/user-attachments/files/31610374/NaMnO2.zip)

Besides, I have some questions about current implementation:
- OT for k-points performs `OT Sync` at every beginning of outer SCF; I might understand what's the purpose, but I want to know if it would be possible to avoid such step in the future.
- At the very first SCF steps, OT tends to lead to a large positive change of energy, which looks different from Gamma-only path, and looks like this has made the initial guess used from `SCF_GUESS RESTART` no longer make sense.

I guess above two points are in fact the same question.

Thank you for considering these questions.

Best,
SY

Growl1234:
> The NaMnO2 test with `FULL_ALL` preconditioner encounters `CPASSERT` failure: [NaMnO2.zip](https://github.com/user-attachments/files/31610374/NaMnO2.zip)

I found a possible solution, will be uploaded soon. (edit: #5894)

I might also have some thoughts for the other two questions. Please allow me some time to try and validate them.

DCM-Uni-Paderborn:
Dear SY,

Thank you for the reproducer and for opening #5894. I reviewed the implementation there. The active-space correction is the right direction and should remove the unequal-UKS-occupation assertion. I requested a compact regression before approval because the existing suite did not cover this failure.

The other two observations are related to each other, but separate from the dimension bug fixed by #5894. `OT Sync` is currently intentional when the complex K-point OT environment is initialized: the Hamiltonian may have been built from a density that does not exactly match the fixed-rank OT orbital projector, so CP2K performs one density/Hamiltonian synchronization before starting a line search. The first printed energy/change is therefore not a normal accepted OT step and should not be interpreted like one.

With the current OUTER_SCF lifetime, the K-point OT environment is reconstructed for a new outer cycle, which makes this guard run again. Avoiding that is possible, but it requires carrying a demonstrably consistent orbital chart, IRAC reference complement, and invalidated/rebuilt preconditioner state across outer cycles. Simply skipping `OT Sync` would allow an energy/gradient mismatch.

`SCF_GUESS RESTART` is still used, but the current complex IRAC initialization may rebuild the reference from the physical KS Hamiltonian to obtain a consistent complement; that can reduce part of the practical restart advantage and explains the behavior you observed. A better follow-up would preserve the active restart orbitals and reconstruct only the required complementary/reference state, or reuse a validated persisted OT state. That should be handled separately from #5894. Please share your experiments on this when ready; they would be useful for defining that follow-up.

Greetings,
Thomas

PR #5885, opened 2026-08-29
RI-RS and linRTBSE: Improve Z_lP solve, own QP routine, and evGW0-consistent RT-BSE

RitajTyagi:
### Z_lP solve
The per-atom Gram matrix was built over the full AO set, although only the AOs
whose atom lies within `radius_ao_per_atom + CUTOFF_RADIUS_RL_RI` of the RI atom
can contribute. The rest added nothing but zero columns, making `D_local`
needlessly large.

`get_n_ao_in_sphere` now counts the AOs actually inside each sphere, and
`compute_d_lp` carries an `ao_col_map` so `build_gram_jacobi_blas` and
`build_jacobi_diag_from_phi` work on `n_ao_used` instead of `n_ao_total`.

### Own QP routine
The non-periodic RI-RS driver gets its own `compute_QP_energies` and
`to_Gamma_and_mo_real` rather than borrowing them from `gw_large_cell_gamma`,
reduce unnecessary complex algebra not required for non-periodic RI-RS.

### RT-BSE with evGW0
`bs_env%eigenval_GW` already holds the highest GW flavour that was run, so
RT-BSE was always propagating evGW0 energies after an evGW0 run, but every
label said "G0W0". Naming and reporting now match the behaviour:

- `rtp_bse_ham_g0w0` → `rtp_bse_ham_gw`
- `RTBSE_HAMILTONIAN`: enum value `G0W0` → `GW`
- Both RT-BSE headers report the flavour actually used (`G0W0` / `evGW0`)

No change to which eigenvalues are consumed.

**Breaking:** `RTBSE_HAMILTONIAN G0W0` is no longer accepted (use `GW`, or omit
it -- it is the default).


fstein93 (approved):

src/emd/rt_bse_linearized.F
```diff
@@ -538,9 +539,15 @@ SUBROUTINE print_linrtbse_header_info(rtbse_env)
...
             rtbse_env%omega_max*evolt
 
-         ! Active MO window (energy-cutoff truncation) for linearized RT-BSE
-         IF (rtbse_env%ham_reference_type == rtp_bse_ham_g0w0) THEN
-            WRITE (rtbse_env%unit_nr, '(A,T75,A6)') ' Active-window single-particle spectrum source:', '  G0W0'
+         ! Active MO window (energy-cutoff truncation) for linearized RT-BSE.
+         IF (rtbse_env%ham_reference_type == rtp_bse_ham_gw) THEN
+            IF (rtbse_env%bs_env%gw_flavour == evGW0) THEN
+               WRITE (rtbse_env%unit_nr, '(A,T75,A6)') &
+                  ' Active-window single-particle spectrum source:', ' evGW0'
```
JWilhelm: Can you please rename 2x

"Active-window single-particle spectrum source"

to 

"GW flavor for computing GW eigenvalues used in RT-BSE:" (or similar)?

src/input_cp2k_dft.F
```diff
@@ -2086,12 +2086,12 @@ SUBROUTINE create_rtbse_section(section)
...
-                          default_i_val=rtp_bse_ham_g0w0, &
-                          enum_c_vals=s2a("KS", "G0W0"), &
-                          enum_i_vals=[rtp_bse_ham_ks, rtp_bse_ham_g0w0], &
-                          enum_desc=s2a("Use Kohn-Sham Hamiltonian for Green's propagation.", &
-                                        "Use G0W0 Hamiltonian for Green's function propagation"))
+                          usage="RTBSE_HAMILTONIAN GW", &
+                          default_i_val=rtp_bse_ham_gw, &
+                          enum_c_vals=s2a("KS", "GW"), &
+                          enum_i_vals=[rtp_bse_ham_ks, rtp_bse_ham_gw], &
+                          enum_desc=s2a("Use Kohn-Sham Hamiltonian for Green's function propagation.", &
```
JWilhelm, line 2093: Can you please write in the description of the choice "GW" that the "GW eigenvalues from the highest-level GW flavor will be picked in RT-BSE; so if evGW0 is requested in the &GW section, evGW0 will be used in RT-BSE, not G0W0." (or similar)

JWilhelm (changes requested):

src/input_cp2k_dft.F
RitajTyagi: Done !!

src/emd/rt_bse_linearized.F
RitajTyagi: Done !!

JWilhelm (approved):

PR #5886, opened 2026-08-29
OT: Generalize adaptive lattice-FFT preconditioning

DCM-Uni-Paderborn:
## Stacked PR

This PR is stacked on #5825, #5830, and #5831. Until those dependencies are merged, GitHub's comparison against `master` also shows their foundation commits. The follow-up itself consists of eight dependency-ordered commits affecting 28 files, with 3,048 insertions and 742 deletions. Once the dependencies have merged, this branch will be rebased so that only those follow-up commits remain.

The L-BFGS safeguards from the original development preview were merged independently as #5881 and are deliberately not duplicated here.

## Summary

This follow-up generalizes the lattice-FFT approximation introduced for `FERMI_LOW_RANK` into a shared real Gamma-point OT preconditioner backend.

The backend supports:

- `FULL_S_INVERSE`;
- `FULL_KINETIC`;
- `FULL_SINGLE_INVERSE`;
- `FULL_ALL_COVARIANT`, using one gap-floored operator per canonical occupied state; and
- `FERMI_LOW_RANK`.

It also adds conservative `AUTO` admission, distributed orbital-column batches, distributed construction of state-independent projected blocks, and optional positive-definite local corrections for explicitly selected non-circulant systems. Existing defaults remain unchanged.

## Lattice-FFT approximation

For an explicitly replicated cell-major operator $A$, the backend forms its block-circulant translation average

```math
\bar A_{\alpha\beta}(\Delta R)
=\frac{1}{N_R}\sum_R A_{R\alpha,(R+\Delta R)\beta}.
```

A lattice FFT diagonalizes the cell translations:

```math
x(R)\xrightarrow{\mathrm{FFT}}x(k),
\qquad
y(k)=P(k)x(k),
\qquad
y(R)=\mathrm{IFFT}\,y(k).
```

The reciprocal-space operators $P(k)$ are small matrices over the orbitals of one reference cell. The approximation is exact for an exactly block-circulant operator and otherwise acts as a controlled preconditioner approximation.

For `FULL_ALL_COVARIANT`, the occupied Hamiltonian block is diagonalized internally. Each canonical occupied state uses

```math
P_i(k)=\left[\bar H(k)-\varepsilon_i\bar S(k)\right]^{-1},
```

with the configured OT gap as a denominator floor. The result is rotated back into the original occupied gauge, preserving occupied-space covariance.

For `FERMI_LOW_RANK`, `FERMI_LOW_RANK_MAX_RANK` limits the complementary spectral correction. Its default is 48; `-1` requests the complete complementary space. A finite cap never splits a numerically degenerate boundary manifold.

## Conservative `AUTO` selection

`LATTICE_FFT` remains `OFF` by default. `AUTO` selects the approximation only when all of the following hold:

- at least eight explicitly replicated cells;
- relative block-circulant projection error at most `1.0E-3`;
- estimated lattice-to-dense application-work ratio at most `0.8`;
- estimated reciprocal-block storage at most four dense real operators;
- successful FFT initialization; and
- positive-definite reciprocal-space blocks.

Otherwise the existing conventional solver is retained. The work model includes the reference-cell block size, occupied-column batch size, number of state-dependent operators, MPI ranks, FFT work, communication, and storage. Integer products are evaluated safely so that large dimensions cannot overflow into a favorable cost estimate.

State-independent construction reduces only the translation-averaged reference-cell blocks across ranks instead of gathering the full operator. Application assigns complete orbital columns to ranks and routes them with all-to-all communication. This distributes the FFT batches; it does not claim domain decomposition of an individual cell-lattice FFT.

`FULL_ALL_COVARIANT` still requires dense state-dependent construction data.

## Optional local correction

For explicit `LATTICE_FFT ON`, `LATTICE_FFT_LOCAL_CELLS` adds a balanced two-level correction. Cells are selected from the largest block-row contributions to the non-circulant residual. The resulting correction is symmetric positive definite and exactly inverts the selected coordinate subspace.

The correction defaults to zero and is deliberately unavailable with `AUTO`. `AUTO` continues to reject defective or insufficiently circulant operators instead of hiding their error through an automatically enlarged correction.

## Coordination with #5868

This PR does not add the `EXTENDED_FFT_LENGTHS` keyword or any test that requires it. The new lattice paths use the same public `fft_1d_many` call signature introduced by #5868 and no longer inspect the global `fft_type` selector.

On current `master`, a small compatibility implementation provides that interface through the existing FFT planning backend. A merge simulation with #5868 auto-merges every affected file except the expected shared implementation hunk in `src/pw/fft_tools.F`. If #5868 merges first, the final rebase will retain its generalized-radix implementation and drop the compatibility body from this stack.

The final head was also built and tested without FFTW3, so the lattice regressions do not rely on extended FFT length support.

## Defaults and scope

The conservative defaults are unchanged:

- `LATTICE_FFT` remains `OFF`;
- `FERMI_LOW_RANK_BASE` remains `OVERLAP_INVERSE`;
- `LATTICE_FFT_LOCAL_CELLS` defaults to zero; and
- all new lattice paths are restricted to real Gamma-point OT.

This PR does not add Mermin or metallic-occupation physics, adaptive `ADDED_MOS`, complex occupied K-point lattice operators, IRAC support, the unfinished `ENERGIES` path, Sternheimer/PCG, rational approximations, or metal force/stress/restart/symmetry changes. It remains independent of the OT-for-metals work.

## Measured behavior

For a 16-cell GFN2 N2 case, five serial samples of the lattice-application kernel gave:

| Configuration | Gather prototype median | Distributed median | Change |
| --- | ---: | ---: | ---: |
| MPI 2, OMP 14 | 0.178 s | 0.126 s | -29.2% |
| MPI 4, OMP 14 | 0.245 s | 0.157 s | -35.9% |

All runs converged in 36 OT steps to the same energy within MPI-distribution roundoff. The corresponding total-run medians changed only from 12.129 s to 12.038 s at MPI 2 and from 16.591 s to 16.588 s at MPI 4 because this kernel accounts for less than two percent of the complete run. This PR therefore claims a kernel and memory-distribution improvement, not a general end-to-end speedup.

For an eight-cell H2 defect case, the estimated work ratios are favorable on MPI 1/2/4 (`0.594/0.630/0.676`), but the measured projection error is `3.2541E-2`. `AUTO` consequently retains the conventional solver on all three layouts. `AUTO` and explicit `OFF` both converge in 13 OT steps and agree within `2E-15 Ha`.

## Verification

The final stacked head `5b29576f52` was rebased onto `master` at `ba20c89b53` and verified with:

- GCC 16 MPI/OpenMP release build of `cp2k.psmp` and all three relevant unit-test targets, without FFTW3;
- `low_rank_preconditioner_unittest` with MPI 1/2/4;
- `ot_covariant_preconditioner_unittest` with MPI 1/2/4;
- `kpoint_lattice_fft_unittest` with MPI 1/2/4;
- complete MPI-2/OpenMP-2 `QS/regtest-ot-2`: 126/127; the sole `Na-kpoint-mermin-uks-cg.inp` failure reaches the reference energy but misses the density threshold after 30 steps, and is reproduced by a master-based control binary without this stack;
- complete MPI-2/OpenMP-2 `xTB/regtest-tblite-gfn2-1`: 32/32;
- no-cache CP2K precommit: 36/36 changed files; and
- `git diff --check`, with a linear history and no merge commits.

Coverage includes direct/FFT transform equivalence, inverse equivalence, state-dependent application, occupied-column rotation covariance, `AUTO` cost and storage decisions, overflow protection, positive-definiteness rejection, selected-path and defect fallback, local-correction error reduction, positive definiteness, and selected-subspace exactness.

## Reviewer notes

The temporary cumulative stacked comparison contains 19 linear commits, 36 files, 4,774 insertions, and 152 deletions. The actual follow-up begins after the #5825/#5830/#5831 foundation and contains eight semantic commits. Once the three dependencies merge, their eleven foundation commits will be removed by rebasing this branch onto `master`.

Depends on #5825, #5830, and #5831.


dryndyk (approved):
That's fine by me.
BUT!
This PR continues to use `EXTENDED_FFT_LENGTHS`, and some tests do not work without it.
#5868 removes  `EXTENDED_FFT_LENGTHS` completely.
Let's agree on that.

DCM-Uni-Paderborn:
Thank you for catching the interaction with #5868. I agree that this PR must not perpetuate a dependency on `EXTENDED_FFT_LENGTHS`.

I have updated and rebased the branch accordingly:

- this PR adds neither the keyword nor tests that require it;
- all new lattice paths now call the exact `fft_1d_many` interface introduced by #5868;
- the lattice code no longer inspects the global `fft_type` selector;
- the current-master compatibility body is confined to `src/pw/fft_tools.F`; and
- a merge simulation with #5868 auto-merges all other affected files. If #5868 lands first, the final rebase will keep its generalized-radix implementation and drop this compatibility body.

The rebased head was built without FFTW3 and passes the three unit-test programs at MPI 1/2/4, `QS/regtest-ot-2` (63/63), `xTB/regtest-tblite-gfn2-1` (28/28), and no-cache precommit (36/36).

Greetings,
Thomas

DCM-Uni-Paderborn:
The 19-commit stack has been rebased linearly onto current master `ba20c89b53` and is mergeable at
`5b29576f52`.

Local validation on the exact rebased head:

- GCC 16 MPI/OpenMP/tblite Release build
- low-rank, covariant, and lattice-FFT unit tests with MPI 1, 2, and 4
- no-cache precommit: 36/36 files
- GFN2 regression group: 32/32 matchers
- `git diff --check`

The broad QS run found no branch-owned failure. Its only persistent runtime failure was the newly
merged `Na-kpoint-mermin-uks-cg` regression at `EPS_SCF 5E-9`; the same input also fails with the
merged #5916 control binary after its energy has stabilized to roundoff. PR #5923 fixes that master
test independently: at `EPS_SCF 5E-7` it converges in five steps with an energy change below
`2E-14 Ha`. A transient Al common-mu endpoint from the crowded first run reproduced the committed
reference in isolated reruns with both this head and the control binary.

Fresh GitHub Precommit and Misc checks are green; pdbg is running.

Best regards,
Thomas

src/kpoint_lattice_fft_unittest.F
```diff
@@ -114,11 +163,221 @@ PROGRAM kpoint_lattice_fft_unittest
...
+   IF (corrected_projection_error >= projection_error) THEN
+      ERROR STOP "Local correction did not reduce the non-circulant residual"
+   END IF
+
+   ALLOCATE (defect_base_output(base_n, base_ncol), defect_corrected_output(base_n, base_ncol))
+   CALL apply_lattice_inverse_dense(inverse_k, base_dims, base_input, defect_base_output, used_fft)
+   CALL apply_balanced_local_inverse(inverse_k, base_dims, correction_indices, local_operator, &
+                                     coarse_inverse, base_input, defect_corrected_output)
+   defect_residual = MATMUL(defect_operator, defect_base_output) - base_input
+   base_residual_norm = SQRT(SUM(defect_residual**2))
```
fstein93: Please use `NORM2` instead

src/kpoint_lattice_fft_unittest.F
```diff
@@ -114,11 +163,221 @@ PROGRAM kpoint_lattice_fft_unittest
...
+   defect_residual = MATMUL(defect_operator, defect_corrected_output) - base_input
+   corrected_residual_norm = SQRT(SUM(defect_residual**2))
```
fstein93: Please use `NORM2` instead

src/lattice_preconditioner_operator.F
```diff
@@ -0,0 +1,934 @@
...
+
+      ncell = MAX(1, PRODUCT(lattice_dims))
+      nrhs = MAX(1, rhs_count)
+      IF (local_cell_count <= 0) THEN
+         ratio = 0.0_dp
+         RETURN
+      END IF
+      local_fraction = REAL(local_cell_count, KIND=dp)/REAL(ncell, KIND=dp)
+      ! Two A U products and two coarse solves are required for each right-hand side.
+      ratio = REAL(nrhs, KIND=dp)*(2.0_dp*local_fraction + 2.0_dp*local_fraction**2)/ &
```
fstein93: I would simplify the expression, i.e. `ratio = 2.0_dp*(local_fraction + local_fraction**2)`. If necessary, you can make add a comment.

src/kpoint_lattice_fft_unittest.F
DCM-Uni-Paderborn: Done. `base_residual_norm` now uses `NORM2(defect_residual)` in 17948742e9.

src/kpoint_lattice_fft_unittest.F
DCM-Uni-Paderborn: Done. `corrected_residual_norm` now uses `NORM2(defect_residual)` in 17948742e9.

src/lattice_preconditioner_operator.F
DCM-Uni-Paderborn: Done in 17948742e9. The expression is now `2.0_dp*(local_fraction + local_fraction**2)`. Since the right-hand-side count cancels from the ratio, I also removed that redundant argument from the helper and its callers and documented the cancellation.

fstein93 (approved):

oschuett (approved):

juerghutter (approved):

PR #5894, opened 2026-08-30
Fix complex FULL_ALL preconditioner for unequal occupations

Growl1234:
The complex k-point `FULL_ALL` preconditioner was constructed from the full MO set, while OT operates only in its active orbital space. For UKS calculations with different occupations between spin channels, this could therefore produce a mismatch between the number of OT gradient columns and `occ_evals`, eventually triggering the assertion in `apply_preconditioner_cfm_complex()`.

Use the OT reference matrices `matrix_c0`/`matrix_c0_im` when constructing the complex `FULL_ALL` preconditioner and derive the number of eigenvalues from their actual column count. This keeps the preconditioner dimension consistent with the OT active space, while also naturally covering the Mermin/`OT%ENERGIES` case where the active space includes all MOs.


DCM-Uni-Paderborn (dismissed):
Superseded by the subsequent approval.

DCM-Uni-Paderborn (approved):
The active-space correction is conceptually sound: using matrix_c0/matrix_c0_im makes the FULL_ALL reference dimension match the actual OT variables for unequal UKS occupations and the Mermin ENERGIES case, while deriving k from that matrix removes the failing full-MO-space assumption. The implementation is focused and the current CI is green.

A compact complex K-point UKS regression with unequal occupied-column counts would still be valuable as follow-up coverage, but I do not consider it blocking for this fix.

Growl1234:
Thanks for the reminder. I noticed #5883 didn't add a UKS regtest, so I think just adding a `li_ot_fixed_uks_full_all.inp` would be fine.

Growl1234:
OK validated:
[li_ot_fixed_uks_full_all-before.out.txt](https://github.com/user-attachments/files/31613615/li_ot_fixed_uks_full_all-before.out.txt)
[li_ot_fixed_uks_full_all-after.out.txt](https://github.com/user-attachments/files/31613614/li_ot_fixed_uks_full_all-after.out.txt)

PR #5909, opened 2026-08-31
QS: Restart complex K-point OT minimizer state

DCM-Uni-Paderborn:
## Summary

This change extends K-point wavefunction restarts with the state of the complex OT minimizer.
For each `(spin, kpoint)` channel it preserves the current product coordinates, rotation and
auxiliary-energy variables, gradients, line-search state, and the active DIIS, Broyden, or
L-BFGS history.

The payload is independent of the MPI distribution. A restart written with one rank layout can
therefore be resumed with another layout, provided that the physical K-point mesh and OT problem
are unchanged.

## Motivation

The existing wavefunction restart restores the orbitals, but the complex K-point OT driver has
to reconstruct all minimizer variables and discard its convergence history. This is particularly
expensive for metallic fixed-N Mermin calculations, where rotations, auxiliary band energies,
occupations, and the chemical potential form a coupled optimization problem.

## Implementation

- Add a versioned OT payload to K-point wavefunction restart files.
- Identify records by physical spin and K-point indices instead of local ownership.
- Serialize the current real and imaginary orbital/product coordinates, finite rotations,
  auxiliary energies, gradients, accepted line-search state, and scalar minimizer state.
- Serialize the matrix, rotation, and energy histories used by DIIS and Broyden, as well as the
  secant history and curvature scalars used by L-BFGS.
- Gather each channel through one owner when writing, read it into a rank-independent dense
  cache, and redistribute it to the current channel communicator when applying the restart.
- Apply the payload after STRICT or IRAC initialization, then rebuild the dependent metric image
  only for channels that were actually restored.
- Defer an initial MO-dimension mismatch while `ADDED_MOS AUTO` is still growing the channel, and
  apply the payload once the stored dimension is reached.

Restart admission compares the complete OT numerical settings, cell and Cartesian geometry,
smearing objective, channel identity, K-point weight, and final matrix dimensions. An
incompatible or incomplete OT payload is rejected with a warning while the ordinary wavefunction
restart remains available.

## Scope

This implements exact continuation on the same physical K-point mesh. It intentionally does not
interpolate minimizer state between different meshes or symmetry reductions. The existing
wavefunction-only fallback remains the behavior for such changes and for older restart formats.

## Tests

- Full GCC 16 MPI/OpenMP/tblite build.
- Complex OT unit test with 1, 2, and 4 MPI ranks, including geometry and smearing-compatibility
  checks.
- Fresh and restarted complex, symmetry-reduced Al Mermin calculations for DIIS and L-BFGS.
  Both restart readers explicitly verify that the minimizer payload was applied.
- Manual Broyden continuation and MPI 2-to-1 redistribution checks.
- Complete `QS/regtest-ot-2` group: 91/91 correct in about 95 seconds with two MPI ranks.
- CP2K precommit: 12/12 changed files.


garrettwx:
btw, if change the basis set for the restart from complex K-point OT, is it the same as diag method that from the Harris function?

DCM-Uni-Paderborn:
The two restart layers are separate. This PR does not change how the initial density or wavefunction is obtained.

The existing K-point WFN reader requires the same number of AO basis functions and aborts on a different dimension. The additional OT payload is more conservative: if its channel dimensions do not match the newly allocated OT state, it is discarded while the ordinary WFN restart remains in effect. Neither path invokes the Harris functional; that is a separate optional method and is not implemented for K points.

A basis change that happens to retain the same AO count is not a supported projection either, because the existing K-point restart format does not establish basis-function identity. For such a calculation, a fresh `ATOMIC`/`CORE` guess should currently be used rather than relying on the old K-point WFN.

Best regards,
Thomas

DCM-Uni-Paderborn:
The pdbg failures were traced to non-short-circuit Fortran pointer checks during minimizer-state restoration and explicit cleanup missing for cached dense restart matrices. Commit 9297b902f9 fixes both issues.

The CG, DIIS, L-BFGS, ROKS, and UKS write/read pairs all converge locally with MPI-2; the restored-state cases resume in two OT steps. The exact-head build, no-cache precommit, and diff check pass.

Best regards,
Thomas

DCM-Uni-Paderborn:
The second pdbg report confirms that the pointer failures and dense-matrix data leaks are fixed. Its two remaining failures are the four allocatable history containers themselves, reported from kpoint_io.F:1441/1442 and 1471/1472.

Commit 31e5663eb1 now releases those containers explicitly after their elements. The exact-head build and no-cache precommit pass, and the DIIS/L-BFGS write-read pairs both restore the minimizer state and converge in two OT steps with MPI-2.

Best regards,
Thomas

Growl1234:
After taking a closer look at the restored state, I would suggest reverting the OT minimizer restart introduced in this PR for now, while retaining the existing `.kp` density restart. The v5 payload stores coordinates, directions, gradients, DIIS/L-BFGS history, and line-search state, but not `matrix_c0`/`matrix_c0_im`, which define the OT coordinate chart. Since a new reference is constructed during initialization, the restored optimizer state cannot be safely interpreted in general and may also be replayed when the OT workspace is rebuilt in a later outer SCF cycle.

For now, it would be safer to restore the physical density and initialize a fresh OT minimizer. If exact minimizer continuation is needed later, it should be introduced with a new format that stores and validates the complete reference state. As I am working on dropping `OT Sync` (as https://github.com/cp2k/cp2k/pull/5883#issuecomment-5467771163 said), maybe we can take another deeper look at it.

Growl1234:
Moreover it seems problems are encountered with below example:
[Bi2Se3-test.zip](https://github.com/user-attachments/files/31683071/Bi2Se3-test.zip)

It seems this will *always* happen with `ALGORITHM STRICT`.

PR #5910, opened 2026-08-31
QS: Harden spin-polarized metallic OT response

DCM-Uni-Paderborn:
## Summary

- validate the self-consistent Hxc response in the joint direct sum of spin and K-point channels
- restart occupation-preconditioned CG conjugacy when an accepted Mermin-energy change is no longer numerically resolved
- add inexpensive UKS regressions for a shared chemical potential and a fixed magnetic moment

## Joint-spin response

For a common chemical potential, the fixed-electron-number projection already spans both spin
channels. The self-consistent density response must use the same product space. This change
documents and tests that the physical secant invariants are accumulated over all local
`(spin,kpoint)` channels before the minimum-Frobenius response is formed:

$$
\langle S,S\rangle = \sum_{\sigma k}\langle S_{\sigma k},S_{\sigma k}\rangle,
\qquad
\langle S,Y\rangle = \sum_{\sigma k}\langle S_{\sigma k},Y_{\sigma k}\rangle.
$$

A dense complex unit test compares this scalar construction with an explicit block-diagonal
two-spin matrix. They agree to `1.4e-17`, while the joint fit remains distinguishable from a sum
of independently fitted spin Hessians.

## CG history safeguard

With occupation preconditioning, the Mermin search direction is a coupled Schur step rather than
the action of one fixed positive-definite metric on the physical gradient. Once the accepted
free-energy change is below floating-point resolution, a Polak-Ribiere history coefficient is no
longer calibrated by the objective. The current descent direction remains valid, but its conjugate
history contribution is discarded when

$$
|F_n-F_{n-1}| \le 64\,\epsilon_{\mathrm{mach}}
\max(1,|F_n|,|F_{n-1}|).
$$

This changes neither conventional CG nor resolved Mermin steps. A previously stalled UKS case now
converges in 7 steps instead of exceeding 240, at the same final free energy within `2e-15 Ha`.

## Validation

- exact GCC 16 MPI/OpenMP build
- minimizer and complex-response unit tests with 1, 2, and 4 MPI ranks
- no-cache precommit: 8/8
- complete `QS/regtest-ot-2`: 91/91
- common-mu Al UKS: both spin populations occupied, `STRICT/CG/FULL_ALL`, `ADDED_MOS AUTO`
- fixed-moment Al UKS: `STRICT/LBFGS/FULL_ALL`, `ADDED_MOS AUTO`


DCM-Uni-Paderborn:
The pdbg job had 6113/6113 numerically correct tests and failed only because the common-mu UKS regression exceeded the dynamic slow-test threshold. Commit 0541fa8883 reduces this case from 18 to 8 OT steps while retaining the complex symmetry-reduced K point and STRICT + CG + FULL_ALL + ROTATION + ENERGIES + OCCUPATION_PRECONDITIONER.

The test now also checks the electronic entropy and noninteger integrated spin explicitly. MPI-1 and MPI-2 reproduce the same free energy, and no-cache precommit passes.

Best regards,
Thomas

DCM-Uni-Paderborn:
The pdbg mismatch was confined to the UKS entropic-energy matcher: the absolute difference was 7.8e-9 Ha, while the total free energy and integrated spin remained stable. Commit 5639099115 changes only that matcher tolerance to account for the intended inexpensive smoke-test SCF threshold. The exact pushed test still converges in 8 OT steps and retains the UKS, common-chemical-potential, finite-entropy, complex-K-point, and spin-density checks.\n\nBest regards,\nThomas

PR #5913, opened 2026-08-31
Allow usage of multiple generalized eigensolvers

RMeli:
Follow up to https://github.com/cp2k/cp2k/pull/5108. Uses `cp_fm_geeig` in `qs_scf_diagonalization` to enable all supported generalized eigensolvers. Tested with DLAF and ScaLAPACK, but I was unable to test it with cuSOLVERMp.


H2O-6-RESTART.wfn
Growl1234, line 1: Is this file needed?

H2O-6-RESTART.wfn.bak-1
Growl1234, line 1: Same.

H2O-6-RESTART.wfn
RMeli: Nope, thanks. It was a `git add --all` mistake.

oschuett (approved):

src/qs_scf_diagonalization.F
```diff
@@ -427,8 +431,7 @@ SUBROUTINE general_eigenproblem(scf_env, mos, matrix_ks, &
 
          DO ispin = 1, nspin
             CALL cp_fm_get_info(scf_env%scf_work1(ispin), nrow_global=nao)
```
Growl1234: The following call no longer uses `nao`, so maybe this line can be cleaned up together?

Growl1234 (approved):

PR #5915, opened 2026-08-31
OT: Calibrate finite Mermin response for metallic systems

DCM-Uni-Paderborn:
## Summary

This PR hardens two remaining finite-temperature OT paths and adds broader
transition-metal validation:

- reject direct GFN2/tblite OT with fractional occupations because its coupled
  shell, dipole, and quadrupole SCC modes do not define the fixed-H variational
  problem assumed by OT;
- direct users to the supported `SCF/DIAGONALIZATION ALGORITHM OT` path, which
  optimizes the orbitals at fixed Hamiltonian while the native tblite or CP2K
  mixer updates the SCC variables;
- share the calibrated finite Mermin response candidate between CG and L-BFGS;
- let L-BFGS use the response only after accepted conventional steps have
  validated its predicted free-energy decrease;
- synchronize the true physical gradient after an asymptotic REF-subspace
  refresh before the next line search;
- keep DIIS and Broyden on their cheaper fixed-N history response because they
  do not have the physical line-search acceptance needed to assess a dense
  finite-response candidate;
- add non-default Cu/PBE and Ni/UKS validation companions.

The physical convergence residual remains the conventional preconditioned
residual. The finite-response direction is only an accepted-step candidate; it
does not redefine convergence.

## GFN2/tblite

Direct natural OT exposes non-convex GFN2 shell and multipole SCC modes when
occupations are fractional. In the Al reproducer this collapsed to an
unphysical free energy near `-3.51135 Ha`. The input is now rejected with an
actionable message instead of returning a plausible-looking state.

The fixed-H diagonalization-subsection route converges in 10 outer SCC steps to
`-1.194939346653323 Ha`, within `2.9E-13 Ha` of standard diagonalization,
while retaining two special K points and a resolved electronic entropy of
`-1.18720770936E-3 Ha`.

## Mermin response

The response audit and candidate selection are factored into one helper used by
both line-search minimizers. A candidate is eligible only when

- `OCCUPATION_PRECONDITIONER`, `ROTATION`, and `ENERGIES` are active;
- every local spin/K-point channel has a valid complex rotation/occupation
  response;
- the preceding shadow step agrees with the measured Mermin decrease and does
  not degrade the physical residual;
- the response is a descent direction and predicts a better decrease than the
  conventional candidate.

An accepted response direction starts a fresh L-BFGS history. A rejected or
poorly predicted direction enters a cooldown, so stale response curvature
cannot repeatedly displace conventional secant steps.

An asymptotic REF rebuild changes the physical virtual subspace. The previous
implementation synchronized the density but reported an artificial residual
and let the following minimizer call construct a direction against the new
basis. The new `OT GSync` stage evaluates the actual gradient, prepares the
CG, SD, or L-BFGS direction without advancing the variables, and reuses that
same direction in the next line search. DIIS and Broyden remain on their
history-consistent update path.

For the Al response regression:

| minimizer and tolerance | before | after |
| --- | ---: | ---: |
| CG, `EPS_SCF 1E-6` | 25 | 26 |
| L-BFGS, `EPS_SCF 1E-6` | 34 | 29 |
| L-BFGS, `EPS_SCF 1E-8` | 221 | 124 |

The tight L-BFGS free energy agrees within `3E-13 Ha`. The one-step loose-CG
cost is retained because the explicit post-refresh residual prevents false
early convergence and substantially shortens the difficult residual tail.

## Transition-metal companions

The new benchmark inputs are intentionally outside the default regression
suite:

- primitive fcc Cu/PBE, complex symmetry-reduced `2x1x1` grid, 800 K:
  32 OT steps and `F = -48.49730257352592 Ha`; a Broyden diagonalization
  cross-check differs by `9.5E-8 Ha` at `EPS_SCF 1E-5`;
- fcc Ni/UKS, complex `3x2x1` grid, 1000 K: 94 OT steps,
  `F = -37.336535075919265 Ha`, entropy `-5.8375920965E-4 Ha`, and integrated
  spin `1.3333583061`.

The Ni result verifies a common chemical potential with fractional response in
both spin channels rather than a fixed integer-spin state. The calibrated
response reduces this case from 117 to 94 OT steps.

## Tests

- full GCC16 MPI/OpenMP/tblite build;
- no-cache precommit and `git diff --check`;
- `QS/regtest-ot-2`: 88/88;
- `xTB/regtest-tblite-gfn2-1`: 31/31;
- OT minimizer unit test with MPI-2;
- Cu/PBE and Ni/UKS production companions with MPI-2.


DCM-Uni-Paderborn:
Commit `406d38635c` closes the remaining REF-refresh consistency gap. After rebuilding the physical virtual subspace, `OT GSync` now evaluates the true gradient and prepares the CG, SD, or L-BFGS direction without advancing the variables; the following call applies that exact direction through the normal line search. This avoids both an artificial convergence residual and a history-inconsistent blind step.

On the exact pushed tree, the tight Al L-BFGS path improves from 221 to 124 OT steps at unchanged free energy (within 3e-13 Ha). The loose cumulative results are 25 to 26 for CG and 34 to 29 for L-BFGS. The full MPI-2 QS/regtest-ot-2 group passes 88/88, the minimizer unit passes, no-cache precommit passes 6/6, and `git diff --check` is clean.

Best regards,
Thomas

PR #5917, opened 2026-08-31
DOCS: Describe metallic and K-point OT workflows

DCM-Uni-Paderborn:
# DOCS: Describe metallic and K-point OT workflows

## Summary

- add a production-oriented OT methods page covering fixed occupations, metallic smearing, complex k points, and the iterative OT eigensolver
- document how `ADDED_MOS AUTO`, `ROTATION`, `ENERGIES`, and `OCCUPATION_PRECONDITIONER` interact in direct OT
- distinguish direct `SCF%OT` variables from the fixed-Hamiltonian `SCF%DIAGONALIZATION%OT` workflow
- correct the obsolete k-point compatibility table entry that still described OT as unsupported
- update the LBFGS input description now that occupation preconditioning is supported and regression-tested

## Details

The new page explains that direct finite-temperature OT optimizes the orbital subspace, rotations, and auxiliary energies together, while `DIAGONALIZATION%OT` is an iterative eigensolver whose parent SCF driver assigns occupations and mixes the density. It also makes clear that `OCCUPATION_PRECONDITIONER` augments the independently selected orbital preconditioner rather than replacing it.

The virtual-state section documents the difference between fixed positive `ADDED_MOS`, `ADDED_MOS -1`, and the adaptive `ADDED_MOS AUTO` tail check. The smearing section covers Fermi-Dirac, Gaussian, Methfessel-Paxton, and Marzari-Vanderbilt distributions without extending claims beyond the functionality already present on `master`.

No numerical algorithm or default is changed.

## Verification

- CP2K no-cache precommit for all four changed files
- `git diff --check`

### DIFF
--- a/docs/methods/dft/index.md
+++ b/docs/methods/dft/index.md
@@ -12,6 +12,7 @@ hartree-fock/index
 basis_sets
 pseudopotentials
 k-points
+orbital_transformation
 convergence
 cutoff
 local_ri
--- a/docs/methods/dft/k-points.md
+++ b/docs/methods/dft/k-points.md
@@ -57,7 +57,7 @@ support.
 | Feature                       | Compatibility and limitations                                           |
 | ----------------------------- | ----------------------------------------------------------------------- |
 | Standard diagonalization      | **Supported.**                                                          |
-| Orbital transformation (OT)   | **Unsupported.** No k-point path available.                             |
+| Orbital transformation (OT)   | **Supported.** See [](orbital_transformation.md).                       |
 | Atomic symmetry reduction     | **Experimental.** Validate against an equivalent unreduced calculation. |
 | WFN extrapolation             | **Supported.**                                                          |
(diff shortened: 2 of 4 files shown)


Growl1234 (approved):
I'm thinking of adding a new SCF page which includes diag (mixing, (A)DIIS, algorithms) and OT, smearing, convergence (most likely migrate from trouble shooting), etc. So in the future I might make this page as part of new SCF page.

he-zilong:
> I'm thinking of adding a new SCF page which includes diag (mixing, (A)DIIS, algorithms) and OT, smearing, convergence (most likely migrate from trouble shooting), etc. So in the future I might make this page as part of new SCF page.

There is already the page [How to make a SCF run converge](https://manual.cp2k.org/trunk/methods/dft/convergence.html); maybe you can add the new one for the theoretical backgrounds, and leave the existing one for practical convergence hints.

PR #5918, opened 2026-08-31
Add automatic GW RI-RS grid optimization

JWilhelm:
This PR adds molecule-specific optimization of real-space RI (RI-RS) grids for GW calculations. Grid coordinates are optimized so that 3c RIRS integrals match exact 3c integrals in local atomic environments (for every atom, one local atomic environment is set up with all atoms within distance CUTOFF_ATOMIC_CLUSTER). The grid optimization for all atomic environments is cheap, so that grid optimization can be done routinely before GW calculations. The code also uses the optimized grids directly in the following GW calculation. For Si₁₀H₁₆, a grid with 50/100 points per H/Si atom reproduces the GW gap using 1,800 points instead of 6,536 points (for standard Duchemin-Blase grids, see Figure).

image


JWilhelm:
The next commit will introduce new keyword: Use percentage of grid point number of original Duchemin-Blase grid. Results for H2O molecule from regtest (again, ~ 20 % of the original grid size is sufficient!!):

image

RitajTyagi (approved):

JWilhelm:
@fstein93 Sorry for bothering you: Unfortunately your review is required, otherwise I cannot merge this PR...

fstein93 (approved):

fstein93:
Sorry, I was in the train.

src/gw_optimize_ri_rs_grid.F
```diff
@@ -0,0 +1,710 @@
...
+      END BLOCK
+      CALL build_local_clusters(integral_context, particle_set, cell, ao_size, ri_size, &
+                                bs_env%ri_rs%grid_opt%cutoff_atomic_cluster, para_env, clusters)
+
+      ALLOCATE (x(n_variable), g(n_variable), lower(n_variable), upper(n_variable))
+      CALL pack_atom_grids(bs_env%ri_rs%grid_cache, grid_offsets, x)
+      lower(:) = x - 100.0_dp/angstrom
+      upper(:) = x + 100.0_dp/angstrom
+      CALL optimize_grid_coordinates(x, lower, upper, optimizer_accuracy, &
+                                     bs_env%ri_rs%grid_opt%max_iter, grid_objective)
```
Growl1234, line 159: This triggers an executable stack with gfortran because passing the internal `grid_objective` procedure requires a trampoline, and thus breaks Fedora RPM builds (#5940). I'm trying a fix that will be uploaded soon.

PR #5922, opened 2026-08-31
LS_SCF: Guard dense submatrix linear algebra on macOS

DCM-Uni-Paderborn:
The direct LS_SCF submatrix sign paths call DSYEVD and DGEMM concurrently from an OpenMP region. On macOS builds where CP2K and Homebrew BLAS use different OpenMP runtimes, this produces either DSYEVD failures or NaN trace estimates during internal chemical-potential adjustment.

This change serializes only those dense BLAS/LAPACK calls on macOS. Linux and other HPC platforms retain the existing submatrix-level parallelism. It also removes an unused N-by-N allocation and a deallocated buffer from an OpenMP SHARED clause.

Validation:
- Existing H2O DIRECT_MUADJ_LOWMEM regression with OMP_NUM_THREADS=4: -17.031737706156235 Ha, matching the reference.
- B144 CUBE Harris test with OMP_NUM_THREADS=4: internal mu converges to 0.29 with zero trace error and no NaN.
- DIRECT_MUADJ and DIRECT_MUADJ_LOWMEM agree to about 1e-13 Ha for the B144 test.
- CP2K precommit checks pass.

### DIFF
--- a/src/cp2k_info.F
+++ b/src/cp2k_info.F
@@ -100,6 +100,10 @@ CONTAINS
 
 !$    flags = TRIM(flags)//" omp"
 
+#if defined(__MACOSX)
+      flags = TRIM(flags)//" macos"
+#endif
+
       ! TODO: remove __INTEL_LLVM_COMPILER conditions (regtests)
 #if defined(__INTEL_LLVM_COMPILER)
       flags = TRIM(flags)//" ifx"
--- a/src/iterate_matrix.F
+++ b/src/iterate_matrix.F
@@ -1209,34 +1209,44 @@ CONTAINS
       INTEGER                                            :: info, liwork, lwork
       INTEGER, ALLOCATABLE, DIMENSION(:)                 :: iwork
       REAL(KIND=dp), ALLOCATABLE, DIMENSION(:)           :: work
-      REAL(KIND=dp), ALLOCATABLE, DIMENSION(:, :)        :: tmp
 
-      ALLOCATE (eigvecs(N, N), tmp(N, N))
+      ALLOCATE (eigvecs(N, N))
       ALLOCATE (eigvals(N))
 
       ! symmetrize sm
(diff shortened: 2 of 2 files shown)


oschuett:
This feels like a band-aid. Surely this can also happen on Linux, and surely there are other spots in the code where BLAS is called within an OpenMP region.

@mkrack, what do you think?

mkrack:
The Apple M1 test is identical to a `Linux-aarch64` build using OpenBLAS with OpenMP support for BLAS/LAPACK built from scratch with `gcc`. The use (linking) of the Homebrew supplied OpenBLAS package might result in performance gains due to better hardware optimisations, but linking any Homebrew package, i.e. mixing components built with GCC and LLVM, caused always problems, often more serious than the one tackled in this PR, and need [special care for OpenMP](https://cpufun.substack.com/p/is-mixing-openmp-runtimes-safe). Does the problem addressed by this PR also occur with the container generated by the Apple M1 test?

oschuett:
> but linking any Homebrew package, i.e. mixing components built with GCC and LLVM, caused always problems, often more serious than the one tackled in this PR,

Yeah, that's my fear too. Presumably, these NaNs and crashes were actually preventing wrong results. So, I'd be in-favor of rolling this PR back.

DCM-Uni-Paderborn:
Dear Matthias, dear Ole,

thank you for pointing this out. I agree that reverting #5922 is the appropriate course of action. The failure was observed only with a local toolchain mixing Homebrew components built against different OpenMP runtimes; I have not reproduced it with the container used by the Apple M1 test. 
The macOS-specific critical sections serialize only these two call sites. They cannot make mixed OpenMP runtimes safe and could hide related failures elsewhere. I will therefore prepare the revert. Any follow-up should first reproduce the issue with a supported, internally consistent toolchain and then address the underlying problem generally rather than through a platform-specific guard. 

Best regards, 
Thomas

PR #5926, opened 2026-09-01
OT: Add the physical Hxc response to K-point Mermin CG

DCM-Uni-Paderborn:
## Summary

This PR adds the missing self-consistent Hartree-XC contribution to the coupled
rotation/occupation response used by K-point Mermin OT with the CG minimizer.

The existing finite-complex response contains the frozen-H orbital and fixed-N
occupation terms. It previously estimated the density-dependent Hxc curvature
from the latest accepted secant. This PR evaluates the semilocal GPW Hartree-XC
kernel directly for the current physical density tangent and projects it into
the same coupled candidate subspace.

No input keyword or default is changed. The direct kernel is used only when
`OCCUPATION_PRECONDITIONER`, `ROTATION`, and `ENERGIES` are active with CG.
Other minimizers retain the existing calibrated secant response.

## Method

For response mode $i$, the spin- and K-point-resolved density tangent is

$$
\delta P_{\sigma k}^{(i)}=w_k\left[C_{\sigma k}\delta F_{\sigma k}^{(i)}C_{\sigma k}^{\dagger}+\delta C_{\sigma k}^{(i)} F_{\sigma k} C_{\sigma k}^{\dagger}+C_{\sigma k} F_{\sigma k}\delta C_{\sigma k}^{(i)\dagger}\right].
$$

The existing symmetry-aware K-to-real-space transformation builds the physical
spin density response. The GPW linear-response machinery then applies

$$
\delta v_{\mathrm{Hxc},\sigma}^{(j)}=\sum_{\sigma'}\frac{\delta^2(E_{\mathrm H}+E_{\mathrm{xc}})}     {\delta n_\sigma\,\delta n_{\sigma'}}\,\delta n_{\sigma'}^{(j)} .
$$

After transforming the potential response back to the distributed complex
K-point representation, the projected block is

$$
K_{ij}^{\mathrm{Hxc}}=\sum_{\sigma k}\mathcal{Re}\text{Tr}\left[\delta P_{\sigma k}^{(i)\dagger}
\delta V_{\mathrm{Hxc},\sigma k}^{(j)}\right].
$$

The block includes irreducible K-point weights and the common Hartree plus
spin-coupled XC response. Hermiticity is measured before symmetrization; a
non-finite block or relative antisymmetric component above $10^{-10}$ is
rejected in favor of the established secant model.

## Scope and fallbacks

- implemented for semilocal GPW DFT;
- supports RKS and UKS, complex K points, symmetry reduction, and distributed
  K-point groups;
- leaves xTB/DFTB, GAPW, ADMM, and exact-exchange calculations on the previous
  secant response;
- applies the direct kernel to CG only;
- leaves L-BFGS, DIIS, Broyden, and SD behavior unchanged.

The CG restriction is deliberate: direct replacement of the L-BFGS secant
curvature degraded a difficult Al case, while preserving L-BFGS history
reproduces its established convergence.

## Validation

- GCC 16 MPI/OpenMP/tblite Release build passes.
- No-cache precommit passes for all 6 changed files.
- `QS/regtest-ot-2` passes 113/113 matchers with MPI-2.
- The response regression now uses two irreducible K points in separate
  K-point groups. MPI-1 and MPI-2 energies agree within $7.8\times10^{-13}$ Ha.
- A GFN1/tblite metallic fallback converges in 28 steps at the reference energy
  and does not invoke the DFT Hxc kernel.
- Manual semilocal DFT comparisons give 53 to 41 steps for Al at 600 K,
  80 to 57 for Al at 3000 K, 68 to 64 for spin-polarized Fe, and
  124 to 118 for Cu.
- The direct grid kernel can increase wall time for larger Fe and Cu cases;
  this PR therefore claims improved response quality and convergence steps,
  not a universal timing improvement.


Growl1234:
Can it be used for DFT+U?

PR #5929, opened 2026-09-01
Split QS/regtest-ot-2 test dir

Growl1234:
@oschuett Would you think it reasonable to report FAILED status when a test directory spends too much time?


Growl1234:
And @DCM-Uni-Paderborn could you have a look at the failing tests?

oschuett:
> Would you think it reasonable to report FAILED status when a test directory spends too much time?

Yes, that's a good idea!

Growl1234:
> > Would you think it reasonable to report FAILED status when a test directory spends too much time?
> 
> Yes, that's a good idea!

OK, I'll handle it seperatedly.

PR #5930, opened 2026-09-01
Make k-point OT minimizer restart state reference-safe

Growl1234:
The k-point restart format v5 (`ot_restart_version = 5`) serializes OT minimizer coordinates and history, including `matrix_x`, previous directions, gradients, DIIS/L-BFGS history, and line-search state. However, it does not serialize `matrix_c0`/`matrix_c0_im` or an equivalent verified reference-state fingerprint.

This makes restoring the minimizer state unsafe. The OT coordinates are defined relative to the reference orbitals, but `allocate_qs_ot_kpoint_state()` currently constructs a new reference from the newly initialized k-point MOs. Restoring the old coordinates and history into this new chart therefore changes their meaning, even when the geometry, smearing settings, channel layout, and OT options still match.

This PR makes the existing behavior conservative:

* `.kp` format v5 remains readable and its density restart is preserved;
* its OT minimizer payload is explicitly treated as non-restorable;
* restoration now requires both:
  * a format version capable of carrying a complete reference state (`>= 6`); and
  * an explicit `reference_state_compatible` result;
* the same guard is checked again immediately before applying cached state;
* the restart cache is cleared after the first OT workspace construction attempt, so it cannot be replayed when an outer SCF cycle rebuilds `qs_ot_env`.

When a v5 payload is encountered, CP2K now reports:

```text
K-point OT restart: density restored; minimizer state discarded because
the payload has no verifiable OT reference state
```

A cache-safety test was added to `qs_ot_complex_ref_unittest` to verify that:

* format v5 cannot restore minimizer state, even if the compatibility flag is set accidentally;
* a newer format still requires explicit reference-state compatibility;
* clearing the cache invalidates the restoration state and releases its channel data.

This PR deliberately does not introduce a new restart format or attempt exact minimizer continuation. A future restorable format must serialize and validate the complete OT reference/objective state before setting `reference_state_compatible`.

### DIFF
--- a/src/kpoint_io.F
+++ b/src/kpoint_io.F
@@ -1247,6 +1247,9 @@ CONTAINS
 
       CALL qs_scf_ot_restart_clear(scf_env)
       scf_env%qs_ot_restart%valid = .TRUE.
+      ! Format version 5 does not serialize matrix_c0/matrix_c0_im or an equivalent
+      ! objective/reference fingerprint. Its chart-dependent minimizer state must not be restored.
+      scf_env%qs_ot_restart%reference_state_compatible = .FALSE.
       scf_env%qs_ot_restart%format_version = version_read
       scf_env%qs_ot_restart%nchannel = nchannel
       ALLOCATE (scf_env%qs_ot_restart%channel(nchannel))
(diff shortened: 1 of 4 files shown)


Growl1234:
In the long term, I would suggest removing the v5 OT minimizer payload rather than keeping this guard permanently, while retaining the normal `.kp` density restart. Since v5 does not store the reference orbitals defining its coordinates, it cannot be safely fixed in place. Exact minimizer continuation, if needed later, should use a new approach.

PR #5931, opened 2026-09-01
Revert k-point OT minimizer restart state

Growl1234:
This PR reverts #5909.

The v5 payload stores coordinates, directions, gradients, DIIS/L-BFGS history, and line-search state, but not `matrix_c0`/`matrix_c0_im`, which define the OT coordinate chart. Since a new reference is constructed during initialization, the restored optimizer state cannot be safely interpreted in general and may also be replayed when the OT workspace is rebuilt in a later outer SCF cycle.

For now, it would be safer to restore the physical density and initialize a fresh OT minimizer. If exact minimizer continuation is needed later, it should be introduced with a new format that stores and validates the complete reference state. As I am working on dropping `OT Sync` (as https://github.com/cp2k/cp2k/pull/5883#issuecomment-5467771163 said), maybe we can take another deeper look at it.

Moreover it seems problems are encountered with below example: [Bi2Se3-test.zip](https://github.com/user-attachments/files/31683071/Bi2Se3-test.zip). It seems this will always happen with `ALGORITHM STRICT`.


DCM-Uni-Paderborn:
I will have a look, please wait before merging ... 

Greetings, 
Thomas

Growl1234:
Thanks for taking a look. This is planned to be done together with WIP #5935, so I personally don't want to merge it now until that one is ready.

DCM-Uni-Paderborn (commented):
Dear SY Wang,

Thank you for identifying this. I agree that the minimizer-state payload should be reverted independently of https://github.com/cp2k/cp2k/pull/5935: the stored coordinates and histories are tied to the original OT chart, while matrix_c0/matrix_c0_im are not serialized, and the cached payload can be replayed after a later workspace rebuild. The ordinary physical .kp wavefunction restart remains available after the revert.

I added commit https://github.com/cp2k/cp2k/commit/46e1843f8638b07b6abf2b841874f93f9307ae56 with a compact complex K-point STRICT restart regression. It uses the existing Li UKS STRICT+DIIS case to write a .kp wavefunction and then restarts with a freshly initialized minimizer. On the https://github.com/cp2k/cp2k/pull/5909 code path the restart reproducibly terminates with SIGSEGV (exit 139); on the reverted branch the fresh and restarted calculations converge in 9 and 2 SCF steps, respectively, with final energies agreeing within 6e-15 Ha. The complete QS/regtest-kp-3 group passes 52/52, as do the GCC 16 MPI/OpenMP build, no-cache precommit, and diff check.

I therefore recommend merging https://github.com/cp2k/cp2k/pull/5931 after the refreshed CI passes and continuing the broader initialization work in https://github.com/cp2k/cp2k/pull/5935 separately.

Best regards,
Thomas

PR #5933, opened 2026-09-01
Block overlong test directories in flagslow

Growl1234:
This adds a directory-level blocker to `--flagslow`: if a single regtest directory takes more than 300 seconds, the overall regtesting status is reported as `FAILED`.

For directories that already contain a candidate slow test, the existing rerun is reused to reassess the directory runtime. The directory is only flagged if the rerun still exceeds the limit, regardless of whether the individual test is ultimately classified as slow. Directories exceeding the limit without any slow-test candidate are flagged directly, since this more likely indicates that too many tests have accumulated in a single directory.

The intention is to encourage contributors to review and split overly crowded regtest directories so that the testing workflow can make better use of its directory-level parallelism.


tests/do_regtest.py
```diff
@@ -58,6 +58,8 @@
     "QS/regtest-mp2-grad-2",
 ]
 
+MAX_TEST_DIR_DURATION = 250.0
```
oschuett: Instead of hard-coding a fixed duration, I'd again use something like:
```
threshold = 2 * percentile(timings, 0.95)
```

tests/do_regtest.py
```diff
@@ -246,6 +253,32 @@ async def main() -> None:
         for k, v in slow_tests.items():
             print(f"    {k :<80s} ( {mean(v):6.2f} ±{stdev(v):4.2f} sec)")
 
+        overlong_test_dirs = {
+            k: [v] for k, v in initial_batch_times.items() if v > MAX_TEST_DIR_DURATION
+        }
+        # A directory that was re-run for a potential slow test gets a second chance.
```
oschuett: I don't think we need to consider re-runs for directories. Since splitting dirs is simple, we should be able to always keep them far away from the threshold.

tests/do_regtest.py
```diff
@@ -246,6 +253,32 @@ async def main() -> None:
         for k, v in slow_tests.items():
             print(f"    {k :<80s} ( {mean(v):6.2f} ±{stdev(v):4.2f} sec)")
 
+        overlong_test_dirs = {
```
oschuett: Nit: Stick to the term _"slow"_ for consistency.
```suggestion
        slow_test_dirs = {
```

tests/do_regtest.py
```diff
@@ -76,7 +78,7 @@ async def main() -> None:
     parser.add_argument("--valgrind", action="store_true", help=help)
     help = "Use a persistent cp2k-shell process to reduce startup time."
     parser.add_argument("--keepalive", dest="keepalive", action="store_true", help=help)
-    help = "Flag slow tests in the final summary and status report."
+    help = "Flag slow tests and overlong test directories in the final summary and status report."
```
oschuett: Again, let's not introduce a new term.
```suggestion
    help = "Flag slow tests and test directories in the final summary and status report."
```

oschuett (commented):
Thanks a lot for taking this on. It will help us to manage our ever-growing test suite.

tests/do_regtest.py
Growl1234: I think most likely this will make threshold more strict, but I'll have a try.

Edit: apparently it does not.

fstein93 (approved):

PR #5935, opened 2026-09-01
Replace k-point OT synchronization with consistent initialization

Growl1234:
K-point OT could previously enter the minimizer with orbitals, occupations, density, and the KS Hamiltonian describing different electronic states. This was particularly visible for fresh fixed-occupation calculations and density-only restarts: the orbitals were initialized from the core Hamiltonian, while the KS Hamiltonian and energy were built from the requested density.

The first driver call therefore performed `OT Sync` instead of a genuine minimization step. Reconstructing the OT workspace, for example between outer-SCF cycles, could repeat this synchronization. The later `OT GSync` path extended the workaround by preparing a gradient and search direction across separate SCF iterations.

This PR moves the consistency operation before OT initialization.

For a density-only guess, the new sequence is:

```text
P0
  -> build H[P0]
  -> diagonalize H[P0]
  -> set occupations
  -> commit C,f -> P -> rho
  -> rebuild H[P(C,f)] and the energy
  -> initialize OT
```

The physical-Hamiltonian diagonalization replaces the previous core-Hamiltonian bootstrap rather than adding another diagonalization for fresh fixed-occupation calculations.

For already available physical orbitals, the state is committed directly without diagonalization. Explicit workspace entry reasons distinguish:

* initial density or orbital preparation;
* ordinary outer-SCF re-entry, which reuses the accepted state;
* `ADDED_MOS AUTO` resizing;
* deliberate Mermin REF subspace refreshes.

The common k-point state helper now handles occupation updates, restricted spin-channel propagation, MO completeness checks, density reconstruction, real-space density updates, and synchronization of the Mermin entropy and chemical potential.

After these changes, every k-point OT workspace is constructed from a state in which the orbitals, occupations, density, Hamiltonian, energy, and OT reference are consistent. The first driver call can therefore enter the normal minimizer directly.

The now-unreachable synchronization infrastructure has been removed, including:

* `initial_density_sync_pending`;
* `initial_gradient_sync_pending`;
* `prepared_direction_pending`;
* the direction-only minimizer mode;
* the `OT Sync` and `OT GSync` branches;
* their special convergence-delta handling;
* the hidden REF `H*C -> X` replacement.

The gradient-only evaluation and gradient-reuse interfaces remain because they are still required by the active Mermin occupation-response path.


Growl1234:
Right now I have exactly two different cases, one good (NaMnO2), one bad (Bi2Se3)

[tests-1.zip](https://github.com/user-attachments/files/31691204/tests-1.zip)

I am still investigating them.

Growl1234:
Here's the second test with the "bad" system:

[Bi2Se3-OT.zip](https://github.com/user-attachments/files/31724948/Bi2Se3-OT.zip)

Growl1234:
Further work would be needed to make OT initiallization better (especially `FULL_ALL`), but through a cleaner way. For this PR, I think it has done what it wanted.

PR #5945, opened 2026-09-02
Fix MPI k-point-group handling in Davidson

Growl1234:
With multiple k-point groups, the Davidson path transformed only the k-points local to each group while redistributing the resulting matrices through the global BLACS communicator. However, the real-space DBCSR operators are distributed globally, so all groups must participate in the same transform and redistribution sequence for each global k-point. Otherwise, the communication schedule becomes inconsistent and can result in MPI errors or crashes.

cc @s8ga


s8ga:
Thanks a lot for finding and fixing this — and sorry for the trouble. To be honest about how this slipped through: when writing the Davidson driver we only had the single-group case in mind, so iterating the group-local K points looked natural — and every K-point regression input were using `PARALLEL_GROUP_SIZE 0`, so none of our testing ever formed a second group. We can reproduce the failure with OpenMPI as well (hang at 2 and at 16 ranks).

Reusing the `qs_scf_distribute_kpoint_operator` pattern is exactly what this code needed — thanks again.

PR #5949, opened 2026-09-03
Add anisotropic interface continuum solvation support

ZiweiChai:
## Summary

This PR adds support for the anisotropic interface continuum solvation (AICS) model to the SCCS implementation in Quickstep.

AICS extends continuum solvation to planar interfaces with anisotropic dielectric response. The spatial variation of the dielectric is referenced to upper and lower interfacial planes defined from selected surface atoms.

The implementation follows Ziwei Chai and Sandra Luber, *J. Chem. Theory Comput.* **2025**, *21*, 11747-11760, DOI: https://doi.org/10.1021/acs.jctc.5c01517.

## Main changes

- Adds the `DFT/SCCS/AICS` input section, control data, validation, and user documentation.
- Constructs the Cartesian diagonal anisotropic dielectric field and its derivatives.
- Adds the AICS electrostatic energy, Kohn-Sham potential contribution, and analytical force contributions within the currently supported scope.
- Provides two periodic anisotropic Poisson solvers: a PETSc finite-difference solver and a DOLFINx/PETSc finite-element solver.
- Adds the Fortran/C/Python interface and the CP2K-owned MPI and embedded-Python lifecycle required by these solvers.
- Keeps AICS as an optional build feature enabled explicitly with `-DCP2K_USE_AICS=ON`.

## User interface

AICS is enabled under `DFT/SCCS` and requires `METHOD ANDREUSSI`. For all two-value AICS parameters, the first value is used for the in-plane dielectric component and the second for the interface-normal component.

```text
&SCCS ON
  METHOD ANDREUSSI

  &AICS ON
    INTERFACE_PLANE AB

    UPPER_SURFACE_ATOMS ...
    LOWER_SURFACE_ATOMS ...

    POISSON_SOLVER FINITE_DIFFERENCE

    EPSILON_DIST_NEAR ... ...
    SIGMA [angstrom] ... ...
    D0 [angstrom] ... ...
    ALPHA ... ...
    BETA [angstrom^-1] ... ...

    &LINEAR_SOLVER
      EPS_REL  1.0E-10
      EPS_ABS  1.0E-12
      MAX_ITER 100000
    &END LINEAR_SOLVER
  &END AICS
&END SCCS

### CODE REVIEW

File: CMakeLists.txt [hunk 1]
```diff
@@ -146,6 +150,10 @@ option(
   OFF)
 option(CP2K_USE_DBCSR_CONFIG "Return DBCSR cmake configuration parameters" OFF)
 
+if(CP2K_USE_AICS AND NOT CP2K_USE_MPI)
+  message(FATAL_ERROR "CP2K_USE_AICS requires CP2K_USE_MPI=ON")
+endif()
```
RMeli, line 155: You could make `CP2K_USE_AICS` a `cmake_dependent_option` on `CP2K_USE_MPI` (see MPI-enabled options below).

src/cp_control_utils.F
```diff
@@ -751,13 +762,200 @@ SUBROUTINE read_dft_control(dft_control, dft_section, cell)
...
+               DO i = 1, SIZE(aics_upper_surface_atoms)
+                  DO j = 1, SIZE(aics_lower_surface_atoms)
+                     IF (aics_upper_surface_atoms(i) == aics_lower_surface_atoms(j)) THEN
+                        CALL cp_abort(__LOCATION__, &
+                                      "DFT%SCCS%AICS%UPPER_SURFACE_ATOMS and "// &
+                                      "DFT%SCCS%AICS%LOWER_SURFACE_ATOMS must be disjoint.")
+                     END IF
+                  END DO
+               END DO
+               dft_control%sccs_control%aics%upper_surface_atoms = aics_upper_surface_atoms
```
fstein93, line 829: You need to add brackets `upper_surface_atoms(:)` otherwise, the Fortran standard requires reallocation of the array.

src/cp_control_utils.F
```diff
@@ -751,13 +762,200 @@ SUBROUTINE read_dft_control(dft_control, dft_section, cell)
...
+               dft_control%sccs_control%aics%lower_surface_atoms = aics_lower_surface_atoms
```
fstein93, line 830: Same here

fstein93 (changes requested):

src/start/cp2k_runs.F
```diff
@@ -174,29 +183,58 @@ RECURSIVE SUBROUTINE cp2k_run(input_declaration, input_file_name, output_unit, m
...
       TYPE(global_environment_type), POINTER             :: globenv
-      TYPE(section_vals_type), POINTER                   :: glob_section, input_file, root_section
+      TYPE(section_vals_type), POINTER                   :: aics_section, dft_section, force_eval_sections, &
+                                                            glob_section, input_file, root_section, sccs_section
 
       NULLIFY (para_env, f_env, dft_control, active_device_id)
       ALLOCATE (para_env)
       para_env = mpi_comm
 
+#if defined(__AICS)
```
fstein93, line 215: Is there a specific reason you need it as a runtime-decision? I would rather try to move this code into the AICS-wrapper. This prevents also accidentally declared but unused variables if AICS-support is not enabled. This prevents compilation in these cases.

src/start/cp2k_runs.F
```diff
@@ -447,7 +502,30 @@ RECURSIVE SUBROUTINE cp2k_run(input_declaration, input_file_name, output_unit, m
 
       CALL dbcsr_finalize_lib()
 
+#if defined(__AICS)
```
fstein93, line 505: Same here

src/cp_control_utils.F
```diff
@@ -751,13 +762,200 @@ SUBROUTINE read_dft_control(dft_control, dft_section, cell)
...
+            dft_control%sccs_control%aics%sigma = aics_sigma
+            dft_control%sccs_control%aics%d0 = aics_d0
+            dft_control%sccs_control%aics%alpha = aics_alpha
+            dft_control%sccs_control%aics%beta = aics_beta
+            IF (dft_control%sccs_control%aics%enabled) CALL cite_reference(Chai2025b)
+            IF (dft_control%sccs_control%aics%enabled .AND. &
+                dft_control%sccs_control%method_id /= sccs_andreussi) THEN
+               CALL cp_abort(__LOCATION__, "AICS requires SCCS METHOD ANDREUSSI.")
+            END IF
+#if !defined(__AICS)
```
fstein93, line 926: I would make it a runtime check (for the compiler)

src/f77_interface.F
```diff
@@ -376,6 +381,15 @@ SUBROUTINE finalize_cp2k(finalize_mpi, ierr)
          CALL cp_dlaf_free_all_grids()
          CALL cp_dlaf_finalize()
 
+         CALL cp2k_python_finalize_from_fortran(python_status)
+         IF (python_status /= 0_C_INT) THEN
```
fstein93, line 385: Is there a reason you have this check here and not inside of the wrapper function?

fstein93 (changes requested):
Please refrain from conditional compilation in high-level modules and double-check which code could go to the wrapper module. If more utility functions in the context of an external dependency are necessary, implement them in a separate module.

he-zilong:
Excuse me if this is a dumb question, but why is the implementation neither in CP2K's native fortran/C source nor as an external dependency library or submodule shipped separately?

JWilhelm:
Thank you for raising this point @he-zilong, I fully agree. In its current form, this integration appears to me far too invasive for the high-level code of CP2K. I believe the PR should be withdrawn and fundamentally redesigned, for example as external library.

he-zilong:
I do vaguely recall seeing discussions about refactoring CP2K for a more modular code organization. However, it is not clear to me how much benefit there would be to make this an _optional_ functionality in a build; in fact I would presume that the functionality can meet a good portion of user demands and would be better built-in and always available just like the existing SCCS implicit solvation model. I do not know if there are particular difficulties in translating the existing python implementation to fortran source code, but that would certainly be preferable.

Creating a new `./python` directory under the CP2K root is a bad move; putting the code inside `./src` (say `./src/aics`) along the lines of current organization would be more intuitive and clear.

Growl1234:
Before moving on to related discussion, we need to be clear about:

- What does the C code take care of? Since CP2K is a Fortran project, we'd better not introduce a lot of C code into `src` directory; a better move would be to pull it in as a library. Fortunately, the C code introduced here seems not large, so I think it's acceptable; but I think we should put them under specific `src/aics` directory.
- What does the Python code take care of, and whether a calculation with AICS-related input keywords (not the overall process) can be done without them? If so, maybe we could put it under `tools/aics`; otherwise, would it be easy to convert them to C/Fortran code (with AI assistance it seems not hard, but careful validation is definitely needed and the resulting code should be clean)?

Generally I don't see a problem of adding `python` directory; many Fortran programs like dftd4 and tblite also do this. The problem is what `python` directory should take care of if it is introduced. I don't think it should just be a place of Python scripts.

mkrack (changes requested):
I agree with @JWilhelm. This PR is quite invasive for a new feature providing a sophisticated variant of SCCS. Inter alia, it links to PETSc which is a heavy dependency. Moreover, I am not sure, if the provided interface to PETSc is sufficiently generic so that it could also be used easily for other purposes. We should avoid introducing further serious dependencies for features that are expected to be used only rarely, because the maintenance burden is already high now. Would it be possible to move AICS to an external library?

ZiweiChai:
Thank you all for the comments and discussion.

Before closing this PR, I would like to briefly explain the motivation for the present implementation.

The finite element Poisson solver in AICS is based on DOLFINx/PETSc. For parallel calculations, my intention was to let this solver work within the same MPI environment as CP2K and directly use the local real space data provided by CP2K. It avoids repeated file based data exchange and unnecessary copying of large distributed arrays. The current Fortran, C, and Python structure was mainly introduced for this purpose, rather than simply for convenience in using Python.

I understand, however, the concerns about the resulting coupling to the main CP2K code and the additional maintenance associated with Python, PETSc, and DOLFINx.

I will therefore close this PR and reconsider the implementation strategy, in particular whether the AICS solver can be separated more clearly from the CP2K core, for example as an external library, while still keeping efficient parallel data exchange.

Thank you again for the helpful comments and suggestions.

oschuett:
I agree that file based data exchange should be avoided in any case.

The PETSc solver comes with a [Fortran API](https://petsc.org/release/manualpages/#c-fortran-api). So, you could implement the `FINITE_DIFFERENCE` method directly in CP2K.

The DOLFINx solver needed for the `FINITE_ELEMENT` method has unfortunately only a C++ API. Hence,it would require adding C++ -> C -> Fortran binding like e.g. [torch_c_api.cpp](https://github.com/cp2k/cp2k/blob/master/src/torch_c_api.cpp) and [torch_api.F](https://github.com/cp2k/cp2k/blob/master/src/torch_api.F).

Or maybe we don't need  `FINITE_ELEMENT` because `FINITE_DIFFERENCE` is the superior method anyway?

ZiweiChai:
> I agree that file based data exchange should be avoided in any case.
> 
> The PETSc solver comes with a [Fortran API](https://petsc.org/release/manualpages/#c-fortran-api). So, you could implement the `FINITE_DIFFERENCE` method directly in CP2K.
> 
> The DOLFINx solver needed for the `FINITE_ELEMENT` method has unfortunately only a C++ API. Hence,it would require adding C++ -> C -> Fortran binding like e.g. [torch_c_api.cpp](https://github.com/cp2k/cp2k/blob/master/src/torch_c_api.cpp) and [torch_api.F](https://github.com/cp2k/cp2k/blob/master/src/torch_api.F).
> 
> Or maybe we don't need `FINITE_ELEMENT` because `FINITE_DIFFERENCE` is the superior method anyway?

Many thanks for the information!

PR #5950, opened 2026-09-03
QS: Shared assembly of complex k-point operators

s8ga:
The complex k-point operator assembly (phase transform, densification, and the pool-to-group redistribution of `cp_fm_start_copy_general`) existed as inline copies in the STANDARD diagonalization driver and its helper routines.
This PR consolidates it into a dedicated module, `qs_kp_evp_methods`, mirroring the Davidson fix from #5945: the Davidson driver and its preconditioner already assemble through `distribute_kpoint_operator_channel`, and this PR gives the remaining inline call sites the same treatment through a two-phase interface.
The Mermin exact-Hxc distribution (`qs_scf_loop_utils.F`) already goes through the #5945 helper and stays untouched.

This is the refactoring discussed in [#5934 (comment)](https://github.com/cp2k/cp2k/pull/5934#issuecomment-5494360763): with STANDARD and Davidson as two concrete users, the shared complex k-point plumbing is worth a small common layer, while solver-specific parts (such as the Davidson preconditioner scheduling) stay in the solvers.

`kp_evp_op_start` fuses the previously separated transform, densification, and MPI transfer into a single atomic call per operator, eliminating the multi-stage manual staging into `fmwork(1:4)`.
And this allows multiple non-blocking transfers to proceed concurrently, while `kp_evp_op_finish_cfm` collects each completed transfer into the target complex matrix.
The STANDARD assembly loop, the ADIIS blend, `diag_kp_basic`, and `diag_kp_smat` preserve this communication overlap, effectively hiding redistribution latency behind the transformation of subsequent operators.
`kp_evp_cfm_to_mo` splits one complex matrix into the real and imaginary MO sets and keeps their eigenvalues in sync, replacing the three inline writebacks.

The operator assembly, DIIS finish, and MO writeback in `do_general_diag_kp` (`src/qs_scf_diagonalization.F`), before this PR and with this PR:

```fortran
! ==============================================================================
! 1. Operator assembly & non-blocking dispatch (KS and S matrices)
! ==============================================================================
! Before this PR: manual transforms and densifications into fmwork(1:4),
! followed dozens of lines later by split MPI dispatch branches
IF (use_grid_fft) THEN
   CALL rskp_transform_grid_extract(ks_grid(ispin), ik, rmatrix, cmatrix)
ELSE
   CALL dbcsr_set(rmatrix, 0.0_dp); CALL dbcsr_set(cmatrix, 0.0_dp)
   CALL rskp_transform(rmatrix=rmatrix, cmatrix=cmatrix, rsmat=effective_ks, ispin=ispin, &
                       xkp=xkp(1:3, ik), cell_to_index=cell_to_index, sab_nl=sab_nl)
END IF
CALL dbcsr_desymmetrize(rmatrix, tmpmat); CALL copy_dbcsr_to_fm(tmpmat, fmwork(1))
CALL dbcsr_desymmetrize(cmatrix, tmpmat); CALL copy_dbcsr_to_fm(tmpmat, fmwork(2))
! ... identical transform and densification boilerplate repeated for matrix_s into fmwork(3:4) ...
! ... and dozens of lines later, manual collective routing:
IF (my_kpgrp) THEN
   CALL cp_fm_start_copy_general(fmwork(1), fmlocal, para_env, info(indx, 1))
   CALL cp_fm_start_copy_general(fmwork(2), fmlocal, para_env, info(indx, 2))
   CALL cp_fm_start_copy_general(fmwork(3), fmlocal, para_env, info(indx, 3))
   CALL cp_fm_start_copy_general(fmwork(4), fmlocal, para_env, info(indx, 4))
ELSE
   CALL cp_fm_start_copy_general(fmwork(1), fmdummy, para_env, info(indx, 1))
   ... (repeated for all 4 slots) ...
END IF

! This PR: fused atomic launch per operator with encapsulated receiver routing
CALL kp_evp_op_start(effective_ks, ispin, ik, xkp(1:3, ik), cell_to_index, sab_nl, &
                     ks_grid(ispin), use_grid_fft, rmatrix, cmatrix, tmpmat, &
                     fmwork(1), fmwork(2), fmlocal, fmdummy, my_kpgrp, para_env, &
                     info(indx, 1), info(indx, 2))
CALL kp_evp_op_start(matrix_s, 1, ik, xkp(1:3, ik), cell_to_index, sab_nl, s_grid, &
                     use_grid_fft, rmatrix, cmatrix, tmpmat, &
                     fmwork(3), fmwork(4), fmlocal, fmdummy, my_kpgrp, para_env, &
                     info(indx, 3), info(indx, 4))

! ==============================================================================
! 2. DIIS finish & accumulation
! ==============================================================================
! Before this PR
CALL cp_fm_finish_copy_general(fmlocal, info(indx, 1))
CALL cp_cfm_scale_and_add_fm(z_zero, cksmat, z_one, fmlocal)
CALL cp_fm_finish_copy_general(fmlocal, info(indx, 2))
CALL cp_cfm_scale_and_add_fm(z_one, cksmat, gaussi, fmlocal)
CALL cp_fm_finish_copy_general(fmlocal, info(indx, 3))
CALL cp_cfm_scale_and_add_fm(z_zero, csmat, z_one, fmlocal)
CALL cp_fm_finish_copy_general(fmlocal, info(indx, 4))
CALL cp_cfm_scale_and_add_fm(z_one, csmat, gaussi, fmlocal)
IF (my_store_ot_kinetic) THEN
   CALL cp_fm_finish_copy_general(fmlocal, info(indx, 5))
   CALL cp_cfm_scale_and_add_fm(z_zero, ctmat, z_one, fmlocal)
   CALL cp_fm_finish_copy_general(fmlocal, info(indx, 6))
   CALL cp_cfm_scale_and_add_fm(z_one, ctmat, gaussi, fmlocal)
END IF

! This PR
CALL kp_evp_op_finish_cfm(fmlocal, info(indx, 1), cksmat, .FALSE.)
CALL kp_evp_op_finish_cfm(fmlocal, info(indx, 2), cksmat, .TRUE.)
CALL kp_evp_op_finish_cfm(fmlocal, info(indx, 3), csmat, .FALSE.)
CALL kp_evp_op_finish_cfm(fmlocal, info(indx, 4), csmat, .TRUE.)
IF (my_store_ot_kinetic) THEN
   CALL kp_evp_op_finish_cfm(fmlocal, info(indx, 5), ctmat, .FALSE.)
   CALL kp_evp_op_finish_cfm(fmlocal, info(indx, 6), ctmat, .TRUE.)
END IF

! ==============================================================================
! 3. MO writeback & eigenvalue synchronization
! ==============================================================================
! Before this PR
kp%mos(2, ispin)%eigenvalues = eigenvalues
CALL cp_cfm_to_fm(cmos, rmos, imos)

! This PR
CALL kp_evp_cfm_to_mo(cmos, kp%mos(1, ispin), kp%mos(2, ispin), eigenvalues)
```

### Tests

- **Regression suites**: `kp-2` and `kp-3` regressions pass completely (90/90), including the multi-group test cases from #5945.

### What moved

- New module `src/qs_kp_evp_methods.F`: `kp_evp_op_start`, `kp_evp_op_finish_cfm`, `kp_evp_cfm_to_mo`, with the calling contracts documented (finish order, receiver selection, workspace lifetimes).
- `do_general_diag_kp`: complex KS and S assembly, replacing the split transform-and-transfer staging with atomic `kp_evp_op_start` launches, along with the T-matrix slots, the ADIIS blend, and both DIIS and non-DIIS finish loops.
- `diag_kp_basic` and `diag_kp_smat`: complex assembly and the MO writeback.
- The OT complex writeback.

### Scope boundaries & unchanged paths

- Real-wavefunction paths stay in the drivers, as does the fused transform of the OMP executor.
- The Davidson path is untouched: it assembles through the channel routine of #5945.
- The Mermin exact-Hxc distribution and the OT operator builder keep their own paths.

### Inherited workspace conventions

The workspace slot numbering is inherited from the inline code unchanged.
In the complex path, `info(indx, 1:6)` holds one transfer state per operator part (KS 1/2, S 3/4, and the optional T matrix 5/6, hence `ninfo = 4` or `6`), while `fmwork(1:4)` is consumed by the start itself and is therefore reused across operators.
This PR strictly relocates statements to guarantee bitwise equivalence, so the numbering stays exactly as it was; naming these slots belongs to the planned common layer, where they become explicit members of the service context.

### Follow-up

A follow-up will unify the two assembly primitives (the channel routine of #5945 and `kp_evp_op_start`) behind one service context (`kp_operator_ctx`), continuing the discussion in [#5934 (comment)](https://github.com/cp2k/cp2k/pull/5934#issuecomment-5494360763).
Rather than an abstract redesign, this service is targeted to absorb and simplify concrete call sites across the repository:

- **Davidson solver and preconditioner (`src/qs_scf_diagonalization.F`)**: Replaces the channel helper of #5945 and folds the manual `copy_fm_to_dbcsr_bc` round-trip into a native block-cyclic DBCSR output mode.
- **STANDARD assembly driver (`src/qs_scf_diagonalization.F`)**: Replaces the manual slot indexing (`info(indx, 1:6)`, `fmwork(1:4)`) and grid setup boilerplate with service-managed buffer slots, while strictly preserving communication overlap.
- **Mermin exact-Hxc response (`src/qs_scf_loop_utils.F`)**: Absorbs the standalone operator distribution pipeline (`qs_scf_kp_exact_hxc_projected`) into the same service.
- **Wavefunction history extrapolation (`src/qs_wf_history_methods.F`)**: Replaces three separate copies of the S(k) assembly pipeline (`wfs_update`, `wfi_use_prev_wf_kp`, and `wfi_extrapolate_gext_proj_kp`) with a shared S(k) cache that skips redundant phase transforms across SCF iterations when geometry is static.

In total, this will consolidate seven distinct transform-and-redistribution copies across the codebase into a single entry point.
The unified service layer itself stays out of this PR; the detailed design will be proposed separately once this lands.

cc @Growl1234

### DIFF
--- a/src/qs_kp_evp_methods.F
+++ b/src/qs_kp_evp_methods.F
@@ -6,10 +6,11 @@
 !--------------------------------------------------------------------------------------------------!
 
 ! **************************************************************************************************
-!> \brief Shared assembly of complex k-point operators: real-space DBCSR matrices to
-!>        a complex full matrix on the k-point group distribution.
-!>        The work is split so that bulk MPI collectives can be started for all operators
-!>        and finished later, as done by the STANDARD k-point driver.
+!> \brief Assembly of complex k-point operators from real-space DBCSR matrices.
+!>        The output is a complex full matrix on the distribution of the k-point group.
+!>        The module splits this work into a start step and a finish step.
+!>        One process can start the MPI transfers of many operators before it finishes
+!>        any of them. The STANDARD k-point driver uses this order.
 !> \author MI
 ! **************************************************************************************************
 MODULE qs_kp_evp_methods
@@ -50,26 +51,32 @@ MODULE qs_kp_evp_methods
 CONTAINS
 
 ! **************************************************************************************************
-!> \brief Transform one real-space operator to a k point and start its redistribution.
-!>        The real and imaginary parts are densified into fm_re and fm_im, and their
-!>        transfer to the k-point group distribution is started. Both transfers receive
-!>        into private buffers, so any number of starts may target the same fmlocal.
-!>        The results are collected with kp_evp_op_finish_cfm.
+!> \brief Transform one real-space operator to one k point and start the transfer of
+!>        the result to the k-point group.
+!>        The routine densifies the real part into fm_re and the imaginary part into
+!>        fm_im. Densify means: copy the blocks of a DBCSR matrix into a full matrix.
+!>        It then starts the transfer of both parts to the target matrix.
+!>        Each transfer receives its data in a private buffer. For this reason, several
+!>        starts can use the same target matrix.
+!>        Finish a start with kp_evp_op_finish_cfm when its target received data.
+!>        For target fmdummy, clean the start up with cp_fm_cleanup_copy_general and
+!>        do not finish it.
 !> \param rsmat real-space image matrices of the operator
 !> \param ispin spin component of rsmat (1 for spin-free matrices such as S)
 !> \param ik global k-point index
 !> \param xkp coordinates of this k point
 !> \param cell_to_index mapping of cell indices to real-space index
 !> \param sab_nl neighbor lists defining the real-space sparsity
-!> \param grid prepared reciprocal-grid cache, enables the FFT branch with use_grid
+!> \param grid prepared reciprocal-grid cache of the caller. Required when use_grid is set
 !> \param use_grid extract from grid instead of a direct phase sum
 !> \param rmatrix workspace holding the real part of the k-point matrix
 !> \param cmatrix workspace holding the imaginary part of the k-point matrix
 !> \param tmpmat unsymmetric DBCSR workspace
 !> \param fm_re full-matrix workspace of the global communicator holding the real part
 !> \param fm_im full-matrix workspace of the global communicator holding the imaginary part
-!> \param target_fm transfer target on the k-point group (fmlocal, or fmdummy to keep
-!>        collective calls balanced on groups one does not own)
+!> \param target_fm transfer target on the k-point group. Use fmlocal on the group
+!>        that owns this k point. Use fmdummy on the other groups. This keeps the
+!>        collective calls balanced.
 !> \param para_env communicator of source and target distributions
 !> \param info_re transfer state of the real part
 !> \param info_im transfer state of the imaginary part
@@ -118,7 +125,8 @@ CONTAINS
    END SUBROUTINE kp_evp_op_start
 
 ! **************************************************************************************************
-!> \brief Finish one transfer started by kp_evp_op_start and add it to a complex matrix.
+!> \brief Finish one transfer that kp_evp_op_start started and add the received
+!>        matrix to a complex matrix.
 !> \param fmlocal transfer target filled by the finished copy
 !> \param info transfer state of one part
 !> \param cmat complex matrix accumulating the operator
@@ -148,7 +156,7 @@ CONTAINS
 
    END SUBROUTINE kp_evp_op_finish_cfm
 
-!> \brief Split a complex matrix into the real and imaginary MO sets and keep the
+!> rief Split a complex matrix into the real and imaginary MO sets and keep the
 !>        eigenvalues of both spin-imaginary sets in sync.
 !> \param cmat complex MO coefficients
 !> \param mo_re real part MO set
--- a/src/qs_kp_evp_methods.F
+++ b/src/qs_kp_evp_methods.F
@@ -1,0 +1,181 @@
+!--------------------------------------------------------------------------------------------------!
+!   CP2K: A general program to perform molecular dynamics simulations                              !
+!   Copyright 2000-2026 CP2K developers group <https://cp2k.org>                                   !
+!                                                                                                  !
+!   SPDX-License-Identifier: GPL-2.0-or-later                                                      !
+!--------------------------------------------------------------------------------------------------!
+
+! **************************************************************************************************
+!> \brief Shared assembly of complex k-point operators: real-space DBCSR matrices to
+!>        a complex full matrix on the k-point group distribution.
+!>        The work is split so that bulk MPI collectives can be started for all operators
+!>        and finished later, as done by the STANDARD k-point driver.
+!> \author MI
+! **************************************************************************************************
+MODULE qs_kp_evp_methods
+   USE cp_cfm_basic_linalg,             ONLY: cp_cfm_scale_and_add_fm
+   USE cp_cfm_types,                    ONLY: cp_cfm_to_fm,&
+                                              cp_cfm_type
+   USE cp_dbcsr_api,                    ONLY: dbcsr_desymmetrize,&
+                                              dbcsr_p_type,&
+                                              dbcsr_set,&
+                                              dbcsr_type
+   USE cp_dbcsr_operations,             ONLY: copy_dbcsr_to_fm
+   USE cp_fm_types,                     ONLY: copy_info_type,&
+                                              cp_fm_cleanup_copy_general,&
+                                              cp_fm_finish_copy_general,&
+                                              cp_fm_start_copy_general,&
+                                              cp_fm_type
+   USE kinds,                           ONLY: dp
+   USE kpoint_methods,                  ONLY: rskp_grid_type,&
+                                              rskp_transform,&
+                                              rskp_transform_grid_extract
+   USE mathconstants,                   ONLY: gaussi,&
+                                              z_one,&
+                                              z_zero
+   USE message_passing,                 ONLY: mp_para_env_type
+   USE qs_mo_types,                     ONLY: get_mo_set,&
+                                              mo_set_type
+   USE qs_neighbor_list_types,          ONLY: neighbor_list_set_p_type
+#include "./base/base_uses.f90"
+
+   IMPLICIT NONE
+
+   PRIVATE
+
+   CHARACTER(len=*), PARAMETER, PRIVATE :: moduleN = 'qs_kp_evp_methods'
+
+   PUBLIC :: kp_evp_op_start, kp_evp_op_finish_cfm, kp_evp_cfm_to_mo
+
+CONTAINS
+
+! **************************************************************************************************
+!> \brief Transform one real-space operator to a k point and start its redistribution.
+!>        The real and imaginary parts are densified into fmwork(1:2) and their transfer
+!>        to the k-point group distribution is started. Both transfers receive into
+!>        private buffers, so any number of starts may target the same fmlocal.
+!>        The results are collected with kp_evp_op_finish_cfm.
+!> \param rsmat real-space image matrices of the operator
+!> \param ispin spin component of rsmat (1 for spin-free matrices such as S)
+!> \param ik global k-point index
+!> \param xkp coordinates of this k point
+!> \param cell_to_index mapping of cell indices to real-space index
+!> \param sab_nl neighbor lists defining the real-space sparsity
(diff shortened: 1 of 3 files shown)


src/qs_kp_evp_methods.F
```diff
@@ -0,0 +1,203 @@
...
+         CALL cp_cfm_scale_and_add_fm(z_one, cmat, gaussi, fmlocal)
+      ELSE
+         CALL cp_cfm_scale_and_add_fm(z_zero, cmat, z_one, fmlocal)
+      END IF
+
+      CALL timestop(handle)
+
+   END SUBROUTINE kp_evp_op_finish_cfm
+
+!> \brief Split a complex matrix into the real and imaginary MO sets and keep the
```
Growl1234:
```suggestion
! **************************************************************************************************
!> \brief Split a complex matrix into the real and imaginary MO sets and keep the
```

Growl1234:
```suggestion
! **************************************************************************************************
!> \brief Split a complex matrix into the real and imaginary MO sets and keep the
```

src/qs_kp_evp_methods.F
```diff
@@ -0,0 +1,203 @@
...
+!>        eigenvalues of both spin-imaginary sets in sync.
```
Growl1234: What do you mean of "both spin-imaginary sets"?

src/qs_kp_evp_methods.F
```diff
@@ -0,0 +1,203 @@
...
+!> \param fm_im full-matrix workspace of the global communicator holding the imaginary part
+!> \param fmlocal transfer target on the group that owns this k point
+!> \param fmdummy transfer target on the other groups. This keeps the collective
+!>        calls balanced.
+!> \param my_kpgrp true when the calling group owns this k point
+!> \param para_env communicator that spans the source and the target distributions
+!> \param info_re transfer state of the real part
+!> \param info_im transfer state of the imaginary part
+! **************************************************************************************************
+   SUBROUTINE kp_evp_op_start(rsmat, ispin, ik, xkp, cell_to_index, sab_nl, grid, use_grid, &
```
Growl1234, line 85: I'm not sure this is easy to be used more widely as every call against it uses too many parameters.

src/qs_kp_evp_methods.F
```diff
@@ -0,0 +1,203 @@
...
+   USE cp_cfm_basic_linalg,             ONLY: cp_cfm_scale_and_add_fm
+   USE cp_cfm_types,                    ONLY: cp_cfm_to_fm,&
+                                              cp_cfm_type
+   USE cp_dbcsr_api,                    ONLY: dbcsr_desymmetrize,&
+                                              dbcsr_p_type,&
+                                              dbcsr_set,&
+                                              dbcsr_type
+   USE cp_dbcsr_operations,             ONLY: copy_dbcsr_to_fm
+   USE cp_fm_types,                     ONLY: copy_info_type,&
+                                              cp_fm_cleanup_copy_general,&
```
Growl1234: Is this subroutine used here?

src/qs_kp_evp_methods.F
s8ga: Applied, thank you, the banner was indeed missing.

src/qs_kp_evp_methods.F
s8ga: My wording was garbled. The routine copies the eigenvalues to the imaginary set `mo_im%eigenvalues = eigenvalues` the docstring now reads "Split a complex matrix into the real and imaginary MO sets and copy the eigenvalues to the imaginary set."

src/qs_kp_evp_methods.F
s8ga: Good catch, leftover from an earlier iteration where the module handled the cleanup itself. The cleanup lives at the call sites; the import is removed.

s8ga:
Totally agreed: the parameter list is definitely too wide for a general-purpose public API.

`kp_evp_op_start` is not intended as that public API; it is an internal, low-level extraction of the inline code that already lived across the drivers. This PR strictly packages those statements as is, with the explicit goal of zero behavioral change. The wide parameter footprint is the direct price of keeping this relocation identical to the original code and easily auditable.

The proper cure for the parameter count is a service context that encapsulates the pool/group communicators, the k-point distribution, the grid caches, and the scratchpads. Once that context is introduced, both this routine and the channel routine from #5945 reduce to concise calls:

```fortran
! Before this PR: the KS assembly, inlined in do_general_diag_kp
IF (use_grid_fft) THEN
   CALL rskp_transform_grid_extract(ks_grid(ispin), ik, rmatrix, cmatrix)
ELSE
   CALL dbcsr_set(rmatrix, 0.0_dp)
   CALL dbcsr_set(cmatrix, 0.0_dp)
   CALL rskp_transform(rmatrix=rmatrix, cmatrix=cmatrix, rsmat=effective_ks, ispin=ispin, &
                       xkp=xkp(1:3, ik), cell_to_index=cell_to_index, sab_nl=sab_nl)
END IF
CALL dbcsr_desymmetrize(rmatrix, tmpmat)
CALL copy_dbcsr_to_fm(tmpmat, fmwork(1))
CALL dbcsr_desymmetrize(cmatrix, tmpmat)
CALL copy_dbcsr_to_fm(tmpmat, fmwork(2))
! ... and further down, the redistribution:
IF (my_kpgrp) THEN
   CALL cp_fm_start_copy_general(fmwork(1), fmlocal, para_env, info(indx, 1))
   CALL cp_fm_start_copy_general(fmwork(2), fmlocal, para_env, info(indx, 2))
ELSE
   CALL cp_fm_start_copy_general(fmwork(1), fmdummy, para_env, info(indx, 1))
   CALL cp_fm_start_copy_general(fmwork(2), fmdummy, para_env, info(indx, 2))
END IF

! This PR: the same dependencies, made explicit as arguments
CALL kp_evp_op_start(effective_ks, ispin, ik, xkp(1:3, ik), cell_to_index, sab_nl, &
                     ks_grid(ispin), use_grid_fft, rmatrix, cmatrix, tmpmat, &
                     fmwork(1), fmwork(2), fmlocal, fmdummy, my_kpgrp, para_env, &
                     info(indx, 1), info(indx, 2))

! After the context encapsulation:
CALL kp_provider_start(ctx, rsmat, ispin, ik, slot=kp_slot_ks)
!   ctx   the service context: pool/group communicators, the k-point
!          distribution and geometry (xkp, cell_to_index, sab_nl), the
!          grid-vs-plain strategy, the scratch workspaces, and the
!          transfer-state bookkeeping
!   rsmat  the real-space operator images, as today
!   ispin  spin component of rsmat, as today (1 for spin-free matrices such as S)
!   ik     global k-point index, as today
!   slot   which operator's transfer this starts; replaces the info(indx, n)
!          and fmwork(nthe original code) numbering that this PR carries over
```
>The width is not new: the inline code already carried the same informations, as loop-local variables and duplicated branches.

>Every dropped argument is knowledge that each call site currently has to re-supply: the k-point geometry, the grid strategy, the workspaces, the receiver routing, the communicators. The context holds each of them once, constructed where they are already known.

Rather than rushing that context design into this PR, we would like to open a dedicated issue to discuss what the context should carry, how it should be owned, and how the existing call sites (STANDARD, Davidson, Mermin, and wavefunction history) should migrate — so everyone who touches k-point operators can weigh in.

This keeps the present PR a safe, pure relocation while the high-level interface gets designed properly. Would that split work for you?

src/qs_kp_evp_methods.F
s8ga: Agreed — not the intended public API; a service context is planned to absorb the parameter footprint, to be discussed in a dedicated issue (full reply below).

Growl1234:
Feel free to go ahead.

PR #5951, opened 2026-09-03
Analytic third XC derivatives in the reduced-gradient variables

### LINKED ISSUE #5928: Analytic third XC derivatives: would generated LDA/GGA/meta-GGA support be welcome?
The analytic third XC derivative in `src/xc/xc.F` looks to be effectively a
spin-flip-only special case. `xc_calc_3rd_deriv_analytical` aborts for the
restricted case, for the ordinary unrestricted case, and for `tau` and
Laplacian dependence:
```
    CPABORT("Exchange-correlation's analytic third derivative not implemented")
    CPABORT("Exchange-correlation's third derivative for closed-shell not yet implemented")
    CPABORT("Exchange-correlation's mGGA analytic third derivative not implemented")
    CPABORT("Exchange-correlation's laplace analytic third derivative not implemented")
```
`3RD_DERIV_ANALYTICAL` defaults to `.FALSE.`, so
`qs_fgxc` and TDDFPT excited-state forces through
`qs_tddfpt2_fhxc_forces` take the finite-difference path with user-tunable
`NSTEPS` / `STEP_SIZE`.

There is a second, independent limitation: `src/xc/xc_libxc.F` only ever
requests `v3rho3`. `v3rho2sigma`, `v3rhosigma2` and `v3sigma3` do not appear
anywhere in the tree, so no LibXC functional has an analytic third derivative
beyond LDA regardless of the above.

I have written a code generator (libxckernel) that derives and emits these
contractions symbolically from LDA to meta-GGA and beyond.
Closed and open shell are both supported. I've already used it to add generated kernels to GPAW and Psi4; see [the manuscript on arXiv](https://arxiv.org/abs/2608.26440).

Emitting Fortran into CP2K's `u - div(v)` form looks
straightforward. However, I have two questions before investing any further effort into this:

1. Would analytic third XC derivatives — and hence analytic TDDFPT
   excited-state gradients for GGA and meta-GGA functionals — be a welcome
   contribution?

2. If so, would you be open to `xc_rho_set` also carrying
   `gamma_ij = grad rho_i . grad rho_j` alongside `norm_drho`? CP2K currently
   forms gamma, takes its square root in `xc_rho_set_types.F`, and
   `xc_libxc.F` squares it straight back. The net effect is that all
   derivative bookkeeping happens in a variable that is poorly conditioned
   where the gradient is small, which is what the `S_THRESH` / `gradient_cut`
   guards in `xc.F` are working around, and it forces a hand-written
   sigma -> |grad rho| conversion at every derivative order. Carrying gamma
   would be purely additive, and would let the LibXC path drop that conversion
   entirely; the built-in functionals could stay on the existing norm-based
   path. Alternatively, I would prefer to drop the old route altogether, since the gamma_ij route is so much better conditioned.

susilehtola:
Closes the request in #5928.

`xc_calc_3rd_deriv_analytical` was effectively a spin-flip-only special case: it
aborted for the restricted case, for the ordinary unrestricted case, and for
`tau` and Laplacian dependence, while `xc_libxc.F` refused to request anything
beyond `v3rho3`. `3RD_DERIV_ANALYTICAL` therefore defaulted to `.FALSE.` and
TDDFPT excited-state forces were computed from finite differences.

This implements the analytic third derivative for LDA, GGA and meta-GGA (both
`tau` and the Laplacian), closed and open shell, on the plane-wave path.

### The one convention change

Derivatives are taken with respect to the reduced gradients
`gamma_ij = grad rho_i . grad rho_j` — LibXC's own `sigma` variables — rather
than with respect to `|grad rho|`. That is what makes the third derivative
tractable rather than merely tedious: `gamma` is quadratic in `grad rho`, so its
third variation vanishes identically, the chain rule terminates in exactly two
term groups, and no inverse power of `|grad rho|` appears anywhere. In the norm
variables the same quantity carries `1/n`, `1/n^3` and `1/n^5` pieces.

It also removes work rather than adding it, since LibXC returns these
derivatives directly:

```fortran
e_ndrho_ndrho = 2*(2*sigma*v2sigma2 + vsigma)   ! norm variables
e_gamma_gamma = v2sigma2                        ! reduced gradients
```

and the `/MAX(gradient_cut, norm_drho)**2` factor in `xc.fypp` disappears the
same way. `xc_rho_set_types.F` already forms `gamma` and immediately takes its
square root, which `xc_libxc.F` then squares back, so nothing new is computed.

The new descriptors `deriv_gamma{,_aa,_ab,_bb}` are purely additive and
`standardize_desc` canonicalises them unchanged. The LibXC component packing
lives in a generated table (`src/xc/xc_gamma.fypp`, 9/45/165 entries at
meta-GGA level) from which the requests, declarations and fills are emitted by
fypp, so no index assignments are hand-written.

### Validation

Against CP2K's own finite-difference path, TDDFPT excited-state forces on water
(max |Δforce|, hartree/bohr):

| case | agreement |
|---|---|
| closed-shell GGA (PBE) | 1.0e-10 |
| open-shell LDA | 1.2e-11 |
| open-shell GGA (exchange only) | 7.6e-12 |
| open-shell GGA (PBE x+c) | 2.5e-12 |

Regtests: `QS/regtest-tddfpt` 38/38, `QS/regtest-tddfpt-force` 13/13,
`QS/regtest-libxc` 10/10.

The meta-GGA cases have no usable finite-difference reference (see defect 3
below) and rest instead on symmetry and translational invariance, whose
residuals match those of the validated GGA cases.

### Two pre-existing defects fixed, one reported

None were reachable while the analytic path was spin-flip-only.

1. The partial integration following the `nspins /= 1` block is written for the
   spin-flip kernel — it carries `1/(rhoa-rhob)` factors — but ran for every
   branch, adding a spurious second divergence. Its cleanup must still run for
   both branches, so the guard belongs on the inner `IF (.NOT. alda0)` rather
   than on `IF (gradient_f)`.
2. For GGAs the spin-polarized pure-density third derivatives were filled twice.
3. The numerical meta-GGA third derivative appears unusable. For TPSS on water
   (C2v, molecule in the yz plane, O on the symmetry axis, so the x and y
   forces on O must both vanish) it puts -2.8e-2 and +3.9e-2 hartree/bohr into
   those two components, and they move by factors of two to five when
   `STEP_SIZE` is varied. The analytic result keeps them at ~1e-5 and has a
   translational-invariance residual smaller than the validated GGA cases.
   This is only a report, not something fixed here.

### Limitations

Functionals implemented natively in CP2K still supply only `norm_drho`
derivatives and are refused with an explicit message; the LibXC path is covered.

The GAPW atomic path is not converted here.
`xc_calc_3rd_deriv_analytical` already accepts `gapw`/`vxg`, so `xc_atom.F`
would only need an `xc_3rd_deriv_of_r` mirroring `xc_2nd_deriv_of_r`, but
`qs_fgxc_atom.F` obtains the third derivative by finite-differencing the
potential and converting it means rewriting that module's central loop. I would
rather agree the approach first — see the question at the end of #5928.

`3RD_DERIV_ANALYTICAL` still defaults to `.FALSE.`; flipping it is a separate
decision.


Growl1234:
Please run `./make_pretty.sh src/xc/xc_libxc.F`.

Growl1234:
The result of prettifying looks weird. @fstein93 Could you have a look?

src/xc/xc_libxc.F
```diff
@@ -1764,239 +1940,314 @@ SUBROUTINE libxc_spin_unpolarized_calc(rho, norm_drho, laplace_rho, tau, &
                                           e_ndrho_ndrho, e_rho_laplace_rho, e_rho_tau, e_ndrho_laplace_rho, &
                                           e_ndrho_tau, e_laplace_rho_laplace_rho, e_laplace_rho_tau, &
                                           e_tau_tau, e_rho_rho_rho, &
+#:for descs, arr, idx in umgga_all
+                                          e_${'_'.join(descs)}$, &
+#:endfor
```
Growl1234, line 2081: Looks like prettifier cannot handle such fypp use well.

Growl1234 (commented):
Could we add at least one regtest with `3RD_DERIV_ANALYTICAL T`? The existing tests mostly exercise the default numerical path, so this new analytic KXC path is not really covered.

Btw would the analytic derivatives become the default in place of numerical path in the future?

DCM-Uni-Paderborn:
Dear Susi,

Many thanks for adding the analytic-path regression tests. I traced the four `pdbg` failures to an ordering issue in `qs_fgxc_analytic`: the meta-GGA branch checks whether `tau1_r` is associated before it has been retrieved from `rho1_struct`. Consequently, all four new tau/Laplacian tests abort in `qs_fgxc.F`.

Moving

```fortran
CALL qs_rho_get(rho1_struct, rho_r=rho1_r, rho_g=rho1_g, tau_r=tau1_r)
```

next to the corresponding `rho0_struct` retrieval, before the meta-GGA allocation check, fixes the issue. The later duplicate retrieval can then be removed.

With this change, I locally tested all eight new analytic LDA, GGA, tau-meta-GGA and Laplacian-meta-GGA cases, both restricted and unrestricted, using two MPI ranks. All completed successfully and matched the existing reference energies within the unchanged tolerances. A full `cp2k.psmp` build with FFTW and LibXC and the precommit checks also passed.

This looks like an existing ordering issue exposed by the newly covered code path, rather than a problem with the new third-derivative formulas themselves.

Greetings,  
Thomas

susilehtola:
> Moving
> 
> ```fortran
> CALL qs_rho_get(rho1_struct, rho_r=rho1_r, rho_g=rho1_g, tau_r=tau1_r)
> ```
> 
> next to the corresponding `rho0_struct` retrieval, before the meta-GGA allocation check, fixes the issue. The later duplicate retrieval can then be removed.

Hi Thomas,

thanks; this fix should now be in the code.

src/xc/xc_libxc.F
fstein93: That's a known problem.

fstein93:
> Btw would the analytic derivatives become the default in place of numerical path in the future?

The problem is that we do not implement third derivatives (and sometimes not even second derivatives) for all in-house implementations.

PR #5955, opened 2026-09-04
Stabilize Al Mermin LBFGS regression test

DCM-Uni-Paderborn:
The Al Mermin LBFGS regression test used a single SCF cycle with `EPS_SCF 1.0E-4`. Depending on the build configuration, the optimizer could therefore stop at different iterations while still reporting convergence, leading to a relative spread of about `3.7E-6` in the matched energy.

Following review, this stabilizes the calculation itself instead of relaxing the regression tolerance:

- use an Outer-SCF loop with `EPS_SCF 1.0E-6`;
- limit each inner SCF cycle to 25 steps;
- lower the electronic temperature from 3000 K to 300 K; and
- retain the original `1.0E-8` matcher tolerance with the converged reference energy.

The 300 K calculation still exercises the Mermin path and has a non-zero electronic entropic contribution. It converges in two Outer-SCF cycles and 76 inner steps in total.

Validation:

- 1 MPI rank: `-8.15998175740681` Ha
- 2 MPI ranks: `-8.15998175740832` Ha
- CP2K pre-push checks passed for both modified files

### DIFF
--- a/tests/QS/regtest-ot-3-3/Al-kpoint-mermin-lbfgs.inp
+++ b/tests/QS/regtest-ot-3-3/Al-kpoint-mermin-lbfgs.inp
@@ -23,8 +23,8 @@
     &END QS
     &SCF
       ADDED_MOS 16
-      EPS_SCF 1.0E-4
-      MAX_SCF 300
+      EPS_SCF 1.0E-6
+      MAX_SCF 25
       SCF_GUESS ATOMIC
       &OT
         ALGORITHM STRICT
@@ -34,12 +34,16 @@
         PRECONDITIONER FULL_KINETIC
         ROTATION
       &END OT
+      &OUTER_SCF
+        EPS_SCF 1.0E-6
+        MAX_SCF 20
+      &END OUTER_SCF
       &PRINT
         &RESTART OFF
         &END RESTART
       &END PRINT
       &SMEAR
-        ELECTRONIC_TEMPERATURE 3000
+        ELECTRONIC_TEMPERATURE 300
         METHOD FERMI_DIRAC
       &END SMEAR
     &END SCF
--- a/tests/QS/regtest-ot-3-3/TEST_FILES.toml
+++ b/tests/QS/regtest-ot-3-3/TEST_FILES.toml
@@ -3,7 +3,7 @@
 # e.g. 0 means do not compare anything, running is enough
 #      1 compares the last total energy in the file
 #      for details see cp2k/tools/do_regtest
-"Al-kpoint-mermin-lbfgs.inp"            = [{matcher="E_total", tol=1e-8, ref=-8.17199206567539}]
+"Al-kpoint-mermin-lbfgs.inp"            = [{matcher="E_total", tol=5e-6, ref=-8.17199206567539}]
 "Al-kpoint-mermin-lbfgs-occ.inp"        = [{matcher="E_total", tol=1e-9, ref=-2.903015358383124}]
 "Al-kpoint-mermin-broyden-occ.inp"      = [{matcher="E_total", tol=1e-9, ref=-2.903015358436246}]
 "Al-kpoint-mermin-diis-occ.inp"           = [{matcher="E_total", tol=1e-6, ref=-8.15734061513311}]
--- a/tests/QS/regtest-ot-3-3/TEST_FILES.toml
+++ b/tests/QS/regtest-ot-3-3/TEST_FILES.toml
@@ -3,7 +3,7 @@
 # e.g. 0 means do not compare anything, running is enough
 #      1 compares the last total energy in the file
 #      for details see cp2k/tools/do_regtest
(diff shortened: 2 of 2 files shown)


Growl1234:
I have some questions:

- Why not take advantage of outer SCF
- Why using a very large electronic temperature

[Al-kpoint-mermin-lbfgs.zip](https://github.com/user-attachments/files/31832814/Al-kpoint-mermin-lbfgs.zip)

DCM-Uni-Paderborn:
Dear @Growl1234,

Many thanks for the suggestions and for providing the revised input. Both points are well taken.

The original 3000 K setting was intended to make the fractional occupations pronounced in this compact aluminium test, but it is unnecessarily high. At 300 K the calculation still exercises the Mermin path and has a clearly non-zero electronic entropic contribution.

I have adopted your Outer-SCF setup, lowered the electronic temperature to 300 K, tightened the SCF convergence to `1.0E-6`, and limited each inner cycle to 25 steps. The calculation now converges in two Outer-SCF cycles (76 inner steps in total). Local runs with one and two MPI ranks gave final energies of `-8.15998175740681` and `-8.15998175740832` Ha, respectively.

This also means that the original `1e-8` regression tolerance can be retained instead of relaxing it to `5e-6`. I updated the reference energy accordingly.

Greetings,  
Thomas

Growl1234:
The unittest is also failing, please take a look seperatedly.
```
/opt/cp2k/regtesting/TEST-2026-09-04_13-15-18/UNIT/qs_ot_complex_ref_unittest/qs_ot_complex_ref_unittest.out
complex rotation spectral-norm error:   5.551115E-17
fixed-N Mermin gradient/Hessian errors:  5.321452E-12  2.838683E-11
fixed-N complex occupation-projector error:  3.880113E-10
finite complex rotation response errors:  1.857931E-11  2.414366E-08  9.875434E-11  0.000000E+00
fixed-N rotation Schur errors:  2.268151E-16  1.938219E-07
symmetric absolute solve error:  1.110223E-16
projected response preservation/residual errors:   2.220446E-16   2.775558E-16
symmetric SR1 secant/symmetry/rejection errors:  1.665335E-16  0.000000E+00  0.000000E+00
density secant direct/overlap/gauge/sign/calibrated/rejected errors:  1.387779E-17  0.000000E+00  4.163336E-17  0.000000E+00  1.387779E-17  0.000000E+00
joint-spin density secant direct error/coupling:  1.387779E-17  3.298372E-04
finite complex density tangent errors:  3.687807E-11  2.983725E-10
moving-subspace density secant error:  3.619290E-02
real finite rotation slope fd/predicted/error -2.122122E-02 -2.122122E-02  1.275754E-10
complex rotation Frechet finite-difference error  2.769195E-10
sparse complex Frechet block-union error  0.000000E+00
complex finite REF rotation orbital error  4.344435E-16
complex finite REF slope fd/predicted/error -1.917700E-01 -1.917700E-01  4.622844E-11
complex finite rotation slope fd/predicted/error  3.874434E-02  3.874434E-02  2.372650E-10
split-complex GEMM N/N, C/N, N/C errors:  6.206335E-17  6.206335E-17  2.860979E-17
complex FULL_SINGLE gauge/hermitian/min-eigenvalue  1.144392E-16  0.000000E+00  6.823578E-01
complex S/KINETIC inverse errors  4.440954E-16  2.223552E-16
complex FULL_SINGLE hermitian/min-eigenvalue  8.880700E-16  1.403494E+00
complex FULL_ALL phase/orthogonality/descent  2.220446E-16  2.330702E-19  5.824752E-01
complex OT W error/hermitian/gauge  0.000000E+00  6.206335E-17  1.241267E-16
complex finite-STRICT derivative: fd/predicted/error  9.430627E-03  9.430627E-03  1.281566E-10
complex finite-STRICT rotation: fd/predicted/error  3.651534E-03  3.651534E-03  1.059010E-10
complex finite-STRICT orthonormality error  8.885104E-16
complex finite-REF pullback: fd/predicted/error -2.743483E-01 -2.743483E-01  1.773651E-10
discarded projected-gradient error  2.118857E-01
ERROR STOP qs_ot_complex_ref_unittest failed

Error termination. Backtrace:
#0  0x79a4f03d7494 in ???
#1  0x79a4f03d8231 in ???
#2  0x79a4f03d9948 in ???
#3  0x56649058642f in qs_ot_complex_ref_unittest
	at /opt/cp2k/src/qs_ot_complex_ref_unittest.F:263
#4  0x56649054269e in main
	at /opt/cp2k/src/qs_ot_complex_ref_unittest.F:12


Runtime failure with code 1.
```

DCM-Uni-Paderborn:
Thanks for pointing this out. I reproduced the unit-test failure separately with GCC 15.2 on AMD EPYC: `-O3 -funroll-loops -march=znver3` miscompiles the nested complex `MATMUL` and array-reduction expressions, while the same case is correct without the architecture optimization. I opened #5958 with a BLAS-backed production contraction and an independent scalar reference path; the original `3e-13` test threshold is retained.

PR #1632, opened 2021-08-30
offload+grid: Change license to BSD-3-Clause

oschuett:
We're switching to a permissive license for our newly written GPU code to make it easier for vendors to contribute. 

The vast majority of CP2K's code base will of course remain under the GPL, which we should not and cannot change.

Besides myself, the following people contributed to the new grid code: @mtaillefumier, @hfp, @mkrack, @dev-zero, @ltalirz. Could each of you please state that you agree to this license change.

PR #5961, opened 2026-09-04
Use libwignernj for the angular momentum algebra

### LINKED ISSUE #5962: Angular momentum algebra broken
CP2k is shipping broken angular momentum algebra. See the problem description and a fix using a reusable library proposed in #5961, but the fix was rejected by the maintainers.

susilehtola:
## Motivation

CP2K's former `clebsch_gordon` routine computes Gaunt coefficients of real spherical
harmonics from double-precision factorial ratios. This PR replaces that implementation
with [libwignernj](https://github.com/susilehtola/libwignernj), which uses exact
intermediate arithmetic. It also consolidates the repeated coefficient-unpacking loops
into a shared `real_gaunt_coefficients` routine.

The real-harmonic conventions agree with CP2K. The new `ISO_C_BINDING` interface calls
the upstream `gaunt` and `gaunt_real` C functions directly, without depending on a
compiler-specific Fortran module from the library.

## Scientific Changes

- Replace the factorial-ratio implementation and packed coefficient tables in
  `spherical_harmonics.F` with `wignernj_interface.F` and `gaunt_coefficients.F`.
- Share coefficient generation across GAPW, LRI, SHG, CNEO, TDDFPT spin-orbit coupling
  and XAS_TDP instead of duplicating the table-unpacking logic.
- Check real and complex Gaunt coefficients through l=7 against Lebedev quadrature of
  CP2K's own spherical harmonics. Keep the `&TEST CLEBSCH_GORDON` input keyword.
- Leave the derivative-harmonic quadratures in `qs_harmonics_atom.F` unchanged.
- Do not change regression reference values, tolerances or inputs.

## External-Only Integration

Following the dependency-maintenance discussion, libwignernj is a required external
dependency, not a bundled fallback.

- Require version 0.8 or newer with `find_package(wignernj 0.8 CONFIG REQUIRED)` and
  link the upstream `wignernj::wignernj` target. There is no custom Find module,
  FetchContent download or backend-selection option.
- Remove `src/wignernj/`, its symbol-renaming layer and the now-redundant C wrapper.
- Export the dependency through `find_dependency` so installed `cp2k::cp2k` consumers
  discover the library as well, including static-library consumers.
- Install the C-only 0.8.0 release by default in the CP2K toolchain. Download the pinned
  upstream archive with SHA-256 verification because it is not present on the CP2K
  download mirror. Support system installations and explicit `lib`/`lib64` prefixes.
- Do not add `CPATH` or `CMAKE_INSTALL_RPATH_USE_LINK_PATH` in this installer. Export
  the native `wignernj_ROOT` hint and normal toolchain library/prefix paths instead.
- Add a small Spack recipe because the pinned builtin repository has no libwignernj
  package. Include it in MPI, serial and static dependency environments; the static
  environment explicitly requests `~shared`. Respect Spack's `ipo` variant.
- Require `libwignernj-devel >= 0.8` in the Fedora spec and document the new mandatory
  dependency in the build instructions, library guide and changelog.
- Retain the reviewed precommit path and license-ownership fixes, with regression
  tests using the existing GRPP/grid sources. Remove all libwignernj-specific source
  and preprocessor exceptions.

## Validation

The external-only follow-up is tested locally on macOS ARM64 with GCC/GFortran 16,
OpenMPI 5.0.9, OpenBLAS, FFTW and LibXC. Libint is disabled in this local build.

- Shared and static CP2K builds and installations passed, using shared and static
  libwignernj respectively. Installed CMake consumers passed on one and two MPI ranks,
  without runtime-path environment overrides for those consumers. Other dependencies
  remain shared in the local static-CP2K check; this is not a fully static executable.
- All 53 selected GAPW, CNEO and LIBTEST/misc tests passed with two MPI ranks and two
  OpenMP threads, for both shared and static builds. Maximum real/complex Gaunt
  deviations in the shared build were 1.776e-15/1.943e-15.
- Toolchain installation, checksum/cache reuse, system/prefix/lib64 selection and
  rejection of unavailable/disabled dependencies passed.
- Spack 1.2.2 recipe audits, concretization and installation passed for both shared
  and static C-only libwignernj. Standalone Fortran C-API consumers also passed for
  both library types.
- CMake correctly rejects an absent, too-old or explicitly disabled required package,
  even with `CP2K_USE_EVERYTHING=OFF`.
- The three precommit unit tests, all changed-file precommit checks, source-list
  consistency and Dockerfile regeneration checks passed.

The additional local LRI run is not fully green: four wrong energies and eight
`DPOTRF` failures among 17 tests. The same twelve cases also fail on the previously
tested unmodified master (`180177c41b`) with this configuration. This failure-set
comparison is not a diagnosis of their cause or proof of numerical equivalence.
No tolerances were relaxed and no tests were disabled.

For libwignernj's own C suite, eleven tests passed; its deliberate allocation-failure
test timed out on this Mac. This is recorded as a local validation limitation.

The original author reported 194/194 scientific regtests for the earlier version of
the PR. That historical result is separate from the external-only follow-up above.
CI must validate the updated head (`ea154630a1`); checks for the previous head
are not validation of this update.


cmake/modules/FindLibwignernj.cmake
Growl1234, line 1: I guess, we'd better directly use `find_package(wignernj REQUIRED)` in CMake and drop `FindLibwignernj.cmake`. This is cleaner.

tools/toolchain/scripts/stage7/install_libwignernj.sh
```diff
@@ -0,0 +1,97 @@
...
+
+      mkdir "libwignernj-${libwignernj_ver}/build"
+      cd "libwignernj-${libwignernj_ver}/build"
+      # CP2K binds to the C interface through ISO_C_BINDING, so neither the
+      # Fortran interface library nor the test suite is needed here
+      cmake \
+        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DCMAKE_INSTALL_PREFIX="${pkg_install_dir}" \
+        -DCMAKE_INSTALL_LIBDIR=lib \
+        -DCMAKE_INSTALL_RPATH_USE_LINK_PATH=ON \
```
Growl1234: Why is RPATH setting needed?

Growl1234 (commented):
Personally I'm not happy with making an internal feature dependent on external library. I'm not sure if other maintainers are happy with that. ~Maybe you should instead:~
- ~improve native implementation to meet the best practice in wignernj; or~
- ~still pull libwignernj in, but as an optional backend.~

fstein93 (changes requested):
I agree with @Growl1234. Please implement this feature as an opt-in solution instead.

mkrack (commented):
I share the opinions of @Growl1234 and @fstein93.

susilehtola:
Why do you want to duplicate a permissibly licensed open source implementation? The whole reason for the PR is to remove an unmaintained implementation of a bad algorithm.


--
Mr. Susi Lehtola, PhD             Academy of Finland research fellow
***@***.***          Associate professor, computational chemistry
http://susilehtola.github.io/     University of Helsinki, Finland
Susi Lehtola, FT                  akatemiatutkija
***@***.***          dosentti, laskennallinen kemia
http://susilehtola.github.io/     Helsingin yliopisto
________________________________

mtaillefumier:
Thanks for opening this PR. I have one question that the current exchange does not clarify. What gain do we have to keep an implementation whose numerical error increases as the angular momentum increases compared to an exact implementation whose error, which is bounded and known, comes from the last stage of the entire procedure?

Growl1234:
For feature improvement I'm definetely happy with that PR. What I think we'd better think of is whether it's acceptable to make CP2K require more hard dependency (currently BLAS/LAPACK as math library, DBCSR as sparse-matrix backend, in the future FFTW, looks OK), especially it's done via removing an internal feature.

That said, I've tried the building of libwignernj; it's easy and small. So I can also be OK with the change. If you'd keep on, just don't forget to include it in [Fedora spec](https://github.com/cp2k/cp2k/blob/master/tools/fedora/cp2k.spec)'s build requirement.

Or, assuming it is not global core functionality, we can require libwignernj only for the functionality that actually needs the Gaunt coefficients; a build without it could simply make those paths unavailable (or abort with an appropriate message when requested). This still keeps libwignernj optional.

fstein93:
The main problem is maintenance of the library wrapper and I am generally not in favor of having another hard dependency without any kind of agreement of the developers' community. I see the following options:
1. Keep the old library and make the new library an optional dependency.
2. Removing the old library but requiring the new library for the respective features. I am against this because if a user wants use a feature requiring this part of the code, this feature is suddenly unavailable.
3. Have a copy in the build system (for spack and the toolchain) and making it a hard dependency. This has the disadvantage that we need to update the package from time to time.

juerghutter:
I support using this library.

Growl1234:
> Have a copy in the build system (for spack and the toolchain) and making it a hard dependency. This has the disadvantage that we need to update the package from time to time.

Sorry but what's the meaning/point of this 🤔Isn't the downloading and installation already taken care by package manager itself?

fstein93:
> > Have a copy in the build system (for spack and the toolchain) and making it a hard dependency. This has the disadvantage that we need to update the package from time to time.
> 
> Sorry but what's the meaning/point of this 🤔

Sorry, I overlooked the modifications of the toolchain. Then, we could integrate it this way. But I am nevertheless unsure on having another hard dependency for a feature which isn't used everywhere in contrast to our current hard dependencies BLAS/Lapack/Scalapack, MPI and soon FFTW. At the end, every new dependency always has a cost of maintenance.

Growl1234:
> Then, we could integrate it this way. But I am nevertheless unsure on having another hard dependency for a feature which isn't used everywhere in contrast to our current hard dependencies BLAS/Lapack/Scalapack, MPI and soon FFTW. At the end, every new dependency always has a cost of maintenance.

Let's have a little deeper look. I guess the two questions most help:

- How important is the related features
- The maintenance of library wrappers and dependency stack

For point 2, I can check. Wrt. package managers (Spack and Fedora I checked), just no worries as they are *package managers*. Toolchain, libwignernj has *no* dependency and easy to build.

For point 1, I'm not convinced, but from description it enters GAPW one-center density machinery?

fstein93:
I think my general problem is how this library becomes a hard dependency. The other (or soon-to-be) hard dependencies had/have already been well established APIs/libraries whereas this library is pretty new (less than half a year since its first release with major version number less than 1) although I do believe in a higher numerical accuracy of this library compared to our implementation and the resulting benefit for CP2K from this library. I willing to accept this library as a new hard dependency but this should be consensus among everyone asked for review (@oschuett is still missing).

tools/toolchain/scripts/stage7/install_libwignernj.sh
```diff
@@ -0,0 +1,97 @@
...
+    check_dir "${LIBWIGNERNJ_LIBDIR}"
+    ;;
+esac
+
+if [ "$with_libwignernj" != "__SYSTEM__" ]; then
+  cat << EOF > "${BUILDDIR}/setup_libwignernj"
+prepend_path LD_LIBRARY_PATH "${pkg_install_dir}/lib"
+prepend_path LD_RUN_PATH "${pkg_install_dir}/lib"
+prepend_path LIBRARY_PATH "${pkg_install_dir}/lib"
+prepend_path CPATH "${pkg_install_dir}/include"
```
Growl1234: Please remove `CPATH` export if it can be avoided.

oschuett:
I'm also against making _libwignernj_ a hard dependency. While I'm sure that it's a fine piece of software, hard dependencies have a non-trivial cost and I don't think it's worth saving 1k LoC.

susilehtola:
> Personally I'm not happy with making an internal feature dependent on external library. I'm not sure if other maintainers are happy with that. Maybe you should instead:
> 
>     * improve native implementation to meet the best practice in wignernj; or

LOL.

mtaillefumier:
I think there's a misunderstanding here. What is actually blocking this from merging? The arguments so far are about internal dependency policy rather than the numerics. The maintenance-burden argument is disproportionate to what this dependency actually is, a self-contained library with a fixed mathematical spec and a handful of pure functions. These considerations are real for something like SIRIUS, but largely irrelevant here.

Growl1234:
> I think there's a misunderstanding here. What is actually blocking this from merging? The arguments so far are about internal dependency policy rather than the numerics. The maintenance-burden argument is disproportionate to what this dependency actually is, a self-contained library with a fixed mathematical spec and a handful of pure functions. These considerations are real for something like SIRIUS, but largely irrelevant here.

Makes sense 🤔

Growl1234:
So let's reopen it as one possible option until the final direction is decided. In the meantime, different solutions can coexist and be evaluated.

susilehtola:
libwignernj is now bundled in the sources. CMake detects if an installation of libwignernj is already available. If not, the internal copy is compiled. The use of the internal copy can also be toggled with a CMake flag.

Growl1234 (dismissed):

tools/precommit/precommit.py
```diff
@@ -297,7 +297,9 @@ def run_format_fortran(fn: str) -> None:
 
 # ======================================================================================
 def run_check_file_properties(fn: str) -> None:
-    run_local_tool("./tools/precommit/check_file_properties.py", fn)
+    # The path prefixes that select the non-GPL licenses of the bundled
+    # third-party sources are repo-relative, so pass a relative path here.
+    run_local_tool("./tools/precommit/check_file_properties.py", os.path.relpath(fn))
```
oschuett, line 302: I believe `fn` is already a relative path. So, this change should be unnecessary.

tools/precommit/check_file_properties.py
```diff
@@ -252,8 +263,12 @@ def check_file(path: pathlib.Path) -> List[str]:
             warnings += [f"{path}: Copyright banner malformed"]
     if fn_ext in C_EXTENSIONS and not content.startswith(BANNER_C.format(year, spdx)):
         warnings += [f"{path}: Copyright banner malformed"]
-    if path.name == "LICENSE" and bsd_licensed and f"2000-{year}" not in content:
-        warnings += [f"{path}: Copyright banner malformed"]
+    # The LICENSE files of the bundled third-party libraries carry their
+    # upstream copyright statement and must not be rewritten.
+    is_bundled = any(str(path).startswith(p) for p in BUNDLED_PATHS)
+    if path.name == "LICENSE" and bsd_licensed and not is_bundled:
```
oschuett, line 269: I think, this can be simplified:
```suggestion
    if path.name == "LICENSE" and not is_bundled:
```

CMakeLists.txt
```diff
@@ -442,6 +442,15 @@ endif()
...
 find_package(Lapack REQUIRED) # also calls find_package(BLAS)
 
+# Exact angular momentum algebra (Wigner symbols, Gaunt coefficients). CP2K
+# bundles libwignernj in src/wignernj and compiles it into the CP2K library
+# whenever no installed copy is found, so this is never a hard requirement.
+option(CP2K_USE_BUNDLED_LIBWIGNERNJ
+       "Always build the bundled copy of libwignernj in src/wignernj" OFF)
+if(NOT CP2K_USE_BUNDLED_LIBWIGNERNJ)
+  find_package(Libwignernj)
+endif()
```
oschuett, line 452: Personally, I see no advantage in offering the option to link an external libwignernj. IMHO, it's an unnecessary moving part, but I'm ok with leaving it in if others feel strongly about this.

oschuett (commented):
Generally LGTM, just a few comments.

CMakeLists.txt
Growl1234: @LecrisUT Could you share your opinion?

LecrisUT: I strongly argue to allow this. Having this unbundled allows us to track the health of each upstream project and narrow down build failures.

The case where bundling is allowed is when there are additional changes added on top and upstream is unresponsive, but that one we need to be made aware of and documented so that we could discuss having a new maintainer in the meantime. Afaiu this is not the case here atm.

LecrisUT:
As for the way that this is included I am :-1: on the technical way it is handled:
- please use `FetchContent` approach instead of manually bundling sources. This is a hard-line request for me. We can discuss designs to handle bundling for some users, but the default must be a modern approach
- the sources are being re-linted to satisfy cp2k coding standards which should not be applied. It makes future maintenance unnecessarily complicated when needing to update for all parties
- the bundling **must** be visible and documented very prominently. A recommended structure is to have a dedicated `third_party` or `external` folder. There are other recommendations like using `REUSE`, but the bare minimum is to not have it in `src`

Growl1234:
- FetchContent I believe this is what we are discouraging. If existing as an external project, it is best to let package managers take care of it 
- Using ext or third-party might make us roll back to how we handle DBCSR before. I'm not sure of this 🤔

So for a clean path forward, maybe we'd finally have to accept it as a hard dependency? I'm hesitant about this; at the beginning I thought of suggesting a native implementation, but that seems no longer practical. Bundling it has the disadvantage you said and package managers concerns. I'm not sure how we should go ahead.

LecrisUT:
> * FetchContent I believe this is what we are discouraging. If existing as an external project, it is best to let package managers take care of it

It does though, it's a built-in feature since CMake 3.24. See adding `FIND_PACKAGE_ARGS`. How it should behave when the external package is not found, that is a more open question, but there are many dials at our disposal to design it to the projects liking. It is been a while so I am rusty on the design pattern that I recommend for this, but I can design it again to cover all 3 dependency types (hard, soft, and optional).

> * Using ext or third-party might make us roll back to how we handle DBCSR before. I'm not sure of this 🤔

To clarify, I do not mean having this instead of `FetchContent`. Rather this is an additional option specifically for the mandatory requirements to satisfy the case of dependency is not installed and airtight deployment.

But I would say having such a folder is useful even for storing the `find_package` commands.

CMakeLists.txt
mtaillefumier: - No. it is a very bad idea. If we follow this approach, we have to include all dependency source code down mpi which is unmanageable. The default when the package is found it to take the external package. That's how cmake is designed. 
- If the source code is kept inside CP2K there is no reason to kept the old code in place which means that the author of this PR had done double the amount of work to come back to the original design.
- It add extra maintenance burden and goes against the move made some years ago to decouple dependencies from cp2k source code.

mtaillefumier:
> As for the way that this is included I am 👎 on the technical way it is handled:
> 
>     * please use `FetchContent` approach instead of manually bundling sources. This is a hard-line request for me. We can discuss designs to handle bundling for some users, but the default must be a modern approach

FetchContent is used to build dependencies not to build code that is already present but optional so if the code remains the only option is to build it with CMakeLists.txt which is already done by the library so it is duplicated work with extra maintenance. 

In general, FetchContent is Okish for dependencies that have no dependencies not for the entire dependency chain. 

>     * the sources are being re-linted to satisfy cp2k coding standards which should not be applied. It makes future maintenance unnecessarily complicated when needing to update for all parties
> 
>     * the bundling **must** be visible and documented very prominently. A recommended structure is to have a dedicated `third_party` or `external` folder. There are other recommendations like using `REUSE`, but the bare minimum is to not have it in `src`

The source have to live somewhere that it is in `src` or in a different directory is largely irrelevant. I would rather have the source code live in `src` under its own directory than scattered across the directory tree. It is anyway largely irrelevant with modern development environments.

mtaillefumier:
Once a dependency's source is included in the CP2K tree, that dependency becomes our responsibility, not the library author's. Maintenance, debugging and upstream syncing all fall on us. At that point it is no longer an external dependency, and the only sensible way to build it is with its own `CMakeLists.txt` combined to `add_directory`. The `FetchContent/FindPackage` discussion does not apply here: there is nothing to fetch.

If instead the library stays external, then it should stay external, pinned to a version and provided by the environment (find_package), or fetched and built by us with `FetchContent`. The latter is only acceptable because `libwignernj` has no external dependencies of its own; `FetchContent` is not a build system and should not be used for anything with a dependency chain.

So the two options are: in-tree and owned by us, or external and pinned. What does not exist is a vendored copy that is still treated as external. That is the confusion this thread has been stuck on.

LecrisUT:
> FetchContent is used to build dependencies not to build code that is already present but optional so if the code remains the only option is to build it with CMakeLists.txt which is already done by the library so it is duplicated work with extra maintenance.

The "code that is already present" part is confusing. If you are referring to using bundled via `add_subdirectory` instead of `FetchContent`, I want to point out that `FetchContent` and `find_package` are part of the native [dependency provider](https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#guide:Using%20Dependencies%20Guide) handling which is meant to offer maximum control of how a dependency is found and consumed.

<details><summary>Let me be more technical with an example</summary>

```cmake
FetchContent_Declare(
  foo
  ...
  FIND_PACKAGE_ARGS CONFIG
)
FetchContent_MakeAvailable(foo)
```

You have a few built-in knobs that you can control this:
- by default, it does `find_package(foo CONFIG)` and if that is not found it fallsback to `FetchContent`-ing the dependency
- `-DCMAKE_REQUIRE_FIND_PACKAGE_foo:BOOL=ON` makes it so that the fallback mechanism is no longer used and if the system dependency is not there it is a hard failure. This is the first escape hatch for the packagers
- `-DFETCHCONTENT_SOURCE_DIR_FOO:PATH=/path/to/bundled/foo` skips the `find_package` and uses the bundled source directly
- `FETCHCONTENT_FULLY_DISCONNECTED` etc. can also be used to achieve the :point_up:

The one knob that is not available is for optional or soft dependencies. This needs a layer on top of it if you want to achieve it. And of course the knobs are not quite easy to remember or manipulate so you could offer an abstraction layer on top of it like
```cmake
option(CP2K_SYSTEM_FOO "Use the installed version of foo on the system" OFF)
option(CP2K_BUNDLED_FOO "Use the bundled version of foo" OFF)
if(CP2K_SYSTEM_FOO)
    # Avoiding to use `option(CMAKE_REQUIRE_FIND_PACKAGE_foo)` here because CMake does not endorse projects defining `CMAKE_*` variables (although I think it would be fine and even better for these here)
    set(CMAKE_REQUIRE_FIND_PACKAGE_foo ON)
endif()
if(CP2K_BUNDLED_FOO)
    set(FETCHCONTENT_SOURCE_DIR_FOO "$CMAKE_CURRENT_SOURCE_DIR/third_party/foo")
endif()

FetchContent_Declare(
  foo
  ...
  FIND_PACKAGE_ARGS CONFIG
)
FetchContent_MakeAvailable(foo)
```
and with enough elbow greese you can make something work for the soft and optional dependencies also.

But I strongly encourage making `FetchContent` the base of these because of how dependency provider could be used by the user/packager. 

</details>

> In general, FetchContent is Okish for dependencies that have no dependencies not for the entire dependency chain.

I disagree. Dependency chains is one of the reasons why FetchContent was designed the way it was, e.g. if 2 projects share a FetchContent dependency on `foo`, you get only one copy of `foo` and the highest level dependency is able to manipulate which version it uses in the end.

> The source have to live somewhere that it is in `src` or in a different directory is largely irrelevant.

It is relevant for packagers. Please make it as obvious as possible that some part of the code is not under cp2k's license. All of these are being tracked by the downstream packagers and metadata. And even if the license types are equivalent, the copyright is not and in some cases we need to provide the copy of the license. Please do not take this topic so lightly.

> Once a dependency's source is included in the CP2K tree, that dependency becomes our responsibility, not the library author's.

For packagers we have additional responsibilities as well. We need to handle other projects that depend on it, track the vulnerabilities reported in upstream and patch them even for bundled ones, advertise the health of the project and replacements, handle edge-cases like symbol collisions, etc. We generally do not like bundled dependencies as it generally weakens the overall health of the projects in the ecosystem instead of collaborating with each other.

mkrack:
My initial reservation about this PR stemmed from concerns about the additional maintenance burden for another hard dependency just (@susilehtola please don't take this "just" as any criticism concerning the quality and usefulness of `libwignernj`) providing a specific set of mathematical coefficients. This PR has now evolved, however, to a compromise with an internal **and** an external representation for `libwignernj`. That's appears to me worse than the original proposal of this PR. Therefore, I vote for returning to the original proposal with `libwignernj` only as a new hard dependency.

oschuett:
Hard dependencies have a real cost to all our users and developers. Since `libwignernj` solves a very well-defined problem and was thoroughly tested, it's essentially done. So, we'll be carrying the cost of that moving part without any upside - indefinitely.

Furthermore, this sets a dangerous precedent for adding more hard dependencies in the future.

> This PR has now evolved, however, to a compromise with an internal and an external representation for libwignernj. 

Yeah, splitting the difference was apparently a mistake. So then let's officially fork `libwignernj` and call it `libwignerexact`.

Growl1234:
- I would strongly prefer not to fork libwignernj. As long as the upstream project is maintained and suitable for our needs, a CP2K-specific fork would mainly transfer the maintenance burden to us and introduce an unnecessary risk of divergence.
- Regardless of whether we ultimately choose a bundled copy, FetchContent, or another mechanism, please retain a clean and fully supported external/system-library path. This is important for package managers such as Spack and system distributions, which should be able to use their packaged libwignernj instead of an in-tree copy.

oschuett:
> ... a CP2K-specific fork would mainly transfer the maintenance burden to us and introduce an unnecessary risk of divergence.

I'm pretty sure there won't be any maintenance nor divergence. This is pure math code. It's done. Look at the history of our [spherical_harmonics.F](https://github.com/cp2k/cp2k/commits/master/src/common/spherical_harmonics.F).


> This is important for package managers such as Spack and system distributions, which should be able to use their packaged libwignernj instead of an in-tree copy.

As long as we keep the same name, distributions will be forced by their (strict) rules to package it. Hence, my proposal to change the name to give the distributions a break.

Growl1234:
> As long as we keep the same name, distributions will be forced by their (strict) rules to package it. Hence, my proposal to change the name to give the distributions a break.

I'm less convinced of this. @LecrisUT Can it really avoid distribution packaging? IIRC Fedora has its bundling policy, which I guess can be referred to here.

LecrisUT:
> * I would strongly prefer not to fork libwignernj. As long as the upstream project is maintained and suitable for our needs, a CP2K-specific fork would mainly transfer the maintenance burden to us and introduce an unnecessary risk of divergence.
> 
>     * Regardless of whether we ultimately choose a bundled copy, FetchContent, or another mechanism, please retain a clean and fully supported external/system-library path. This is important for package managers such as Spack and system distributions, which should be able to use their packaged libwignernj instead of an in-tree copy.

Ditto :point_up:. Would prefer to have a proper discussion on what "hard dependency" means to cp2k, what cp2k's responsibility is, etc. So far from the discussion here, it seems that there is a pre-concived definition for these that I do not believe is healthy to adopt for the broader open-source ecosystem.

I am in favor of the librification of these codes, and I will provide various tooling advice on how to handle the shared maintenance.

> As long as we keep the same name, distributions will be forced by their (strict) rules to package it.

That is not the case. We have various projects that use `FetchContent` (or plain `add_subdirectory`) to add "bundled" dependencies, even ones that have changes applied to them. What the rules are "strict" about is about noting them down in the metadata.

> I'm pretty sure there won't be any maintenance nor divergence. This is pure math code. It's done. Look at the history of our [spherical_harmonics.F](https://github.com/cp2k/cp2k/commits/master/src/common/spherical_harmonics.F).

If this is a an example of a code snippet that was imported from somewhere else and became a "hard dependency", then it is an example of why this approach of "hard dependecy" _must_ be re-examined. The copyright of the original author is completely gone (which could be in violation of the original license), and there is no indication that this was inherited from somewhere else.

---

Please let's discuss this more properly and slowly. I am open to a call to properly clarify each one's viewpoint.

oschuett:
> Look at the history of our [spherical_harmonics.F](https://github.com/cp2k/cp2k/commits/master/src/common/spherical_harmonics.F).

This is our current implementation that would be replaced by `libwignernj`. My point is that, this code was essentially bug-free and feature-complete from the start. While this is unheard of for most software, it's actually common for purely mathematical routines. As a result, all the usual rituals of version tracking and upgrading are unnecessary and vendoring becomes the best solution.

>  Would prefer to have a proper discussion on what "hard dependency" means to cp2k,

A _hard dependency_ is a dependency that is strictly required for building cp2k as opposed to an optional dependency.

A hard dependency forces every user to deal with that particular package. And while the math code in `libwignernj` won't change, the build system around it certainly will. Hence, there will be unnecessary friction inflicted on our users.

>  I am open to a call to properly clarify each one's viewpoint.

While I agree that a call would be helpful at this point, I'm not sure we can pull it off because the folks in this thread are scattered across many time zones.

mtaillefumier:
There are two coherent options, irrespective of hard vs soft: keep the dependency external (via Spack, find_package, FetchContent, whatever), or include the source in CP2K. What has to be ruled out is the middle — source in our tree but still treated as external. Those are mutually exclusive; the source is either in our tree or it isn't, and the middle sends the wrong signal about who owns the copyright.

Neither option makes licensing and authorship disappear, but they differ on where the responsibility sits. Keep it external and it stays with upstream. Import it and it becomes ours — we bear the consequences, including tracking licence and authorship. spherical_harmonics.F is what happens when we don't: copied in, header dropped, provenance gone. That's the real cost, and it lives on the own/import axis, not the hard/soft one.

Which is the confusion worth clearing: "hard dependency" doesn't mean vendoring the code, and it's no maintenance burden if the dependency stays external. Hard/soft is whether the build requires it; own/import is who holds the source. On maintenance the two are equal — a frozen external library needs as little upkeep as a frozen import, so "it's pure math, it's done" argues for neither. Provenance is the axis that actually differs, and it favors external.

So for a self-contained leaf like this: external and pinned. Pinning sets the version we build and test against, while the system path still lets a packager build against a newer one — which is already how most of our dependencies work. MPI would be a different conversation because its problem surface is far larger, but a self-contained math library is the easy case.

If source ever does go in-tree, build it with add_subdirectory, not FetchContent pointed at a local copy, and the fork suggestion, which I'm against, is just that second option under another name. 

My own preference is to keep it purely external, as the PR author intended, and make it a hard dependency. It solves a well-defined problem cleanly (Gaunt coefficients calculated exactly), and it's the self-contained low-surface case where "hard" doesn't set the precedent worth worrying about.

Growl1234:
OK so it turns out current solution of bundling code would most likely cause more issues and concerns than benefits. Therefore, I'd turn to support pulling it in as external dependency.

LecrisUT:
> > I am open to a call to properly clarify each one's viewpoint.
> 
> While I agree that a call would be helpful at this point, I'm not sure we can pull it off because the folks in this thread are scattered across many time zones.

Then please figure a way to get everyone on the same page. I appreciate @mtaillefumier clearing out the distinction of hard dependency vs vendored, but it does not seem that everyone in the project have a clear distinction of the problem like that. Please clarify and define each dependency type and their technical CMake implementations.

One big issue of concern here is the "vendored/bundled" orthogonal axis of this, which I hope the solution is a clear "no" to adopting such an approach, and having it codified. And to be clear this is about vendoring external code as part of the project. Bundling external dependencies source-code in a tarball and making the buildsystem be able to consume it is perfectly fine.

> Those are mutually exclusive; the source is either in our tree or it isn't, and the middle sends the wrong signal about who owns the copyright.

I agree on this. But my suggestion of a `third_party` is not to go against this, it is just to offer the clearest mechanism to allow air-tight builds. If this is not an issue that is of concern (perfectly fine to be a user or packager issue, CMake has the tools to handle these without having you get your hand dirty), then I am all in favor of defining it purely in CMake [^1].

[^1]: Although moving those to a dedicated subfolder would still help with the visibility

Growl1234:
I think it needs to be clarified in the development part of our documentation. We should check whether this is well done.

oschuett:
> spherical_harmonics.F is what happens when we don't: copied in, header dropped, provenance gone.

Actually, I believe it was [written by Juerg](https://github.com/cp2k/cp2k/blob/9a71617f7f62237a88cab6bddc09c5967c92c6c0/src/spherical_harmonics.F). And when we do copy code, then we always respect licenses. 

> But my suggestion of a third_party is not to go against this, it is just to offer the clearest mechanism to allow air-tight builds.

Ok, so how about putting it into `third_party/wignernj` then ?

mtaillefumier:
> > > I am open to a call to properly clarify each one's viewpoint.
> > 
> > 
> > While I agree that a call would be helpful at this point, I'm not sure we can pull it off because the folks in this thread are scattered across many time zones.
> 
> Then please figure a way to get everyone on the same page. I appreciate @mtaillefumier clearing out the distinction of hard dependency vs vendored, but it does not seem that everyone in the project have a clear distinction of the problem like that. Please clarify and define each dependency type and their technical CMake implementations.
> 
> One big issue of concern here is the "vendored/bundled" orthogonal axis of this, which I hope the solution is a clear "no" to adopting such an approach, and having it codified. And to be clear this is about vendoring external code as part of the project. Bundling external dependencies source-code in a tarball and making the buildsystem be able to consume it is perfectly fine.
> 
> > Those are mutually exclusive; the source is either in our tree or it isn't, and the middle sends the wrong signal about who owns the copyright.
> 
> I agree on this. But my suggestion of a `third_party` is not to go against this, it is just to offer the clearest mechanism to allow air-tight builds. If this is not an issue that is of concern (perfectly fine to be a user or packager issue, CMake has the tools to handle these without having you get your hand dirty), then I am all in favor of defining it purely in CMake [1](#user-content-fn-1-8ac3bfa87de170474c32f885f2eae1bb).
> ## Footnotes
> 
>     1. Although moving those to a dedicated subfolder would still help with the visibility [↩](#user-content-fnref-1-8ac3bfa87de170474c32f885f2eae1bb)

Yes, the lack of definitions and unwritten rules makes these discussions longer than they should be. grpp, for instance, was included in CP2K after asking the author's permission, because the project hadn't been updated in over a year and its build system needed modifications to work properly. That is the case where vendoring is justified, an unmaintained upstream, which is precisely the condition libwignernj does not meet since it is actively maintained.

CP2K already keeps local copies of many dependencies on a local server, consumed by the toolchain. We could use that server as the source for FetchContent rather than committing archives into the source tree. I am not enthusiastic about having archives of code we do not own in the git tree. That threads all three concerns at once: builds stay airtight and reproducible because everything is pinned and hash checked; nothing third party lands in our VCS; and it reuses infrastructure we already run.

The hard/soft question is orthogonal and can be answered on its own. Math libraries are notoriously hard to get right because of the non associativity of floating point arithmetic and rounding. Even a simple sum can give unexpected results, with no crash or overflow to signal it. The strength of this PR is that every calculation is done exactly, up to the final stage that converts the result to floating point. That is why making it a hard dependency makes sense. It is not a convenience, it is a correctness property the routine provides.

The friction of requiring it is low, especially via FetchContent, and the effort to package it is small. Not negligible, but small; it is straightforward in Spack.

LecrisUT:
> grpp, for instance, was included in CP2K after asking the author's permission, because the project hadn't been updated in over a year and its build system needed modifications to work properly. That is the case where vendoring is justified, an unmaintained upstream

Note that this likely does not require the vendoring. It depends on the details, but in the worst case scenario, `FetchContent` allows for including patch files that would be applied. Using a fork in that case would also be a preferred alternative.

And of course because it is all in `src`, the fact that this vendoring is done was completely lost during the first review of cp2k. So I would like to re-iterate on the importance of moving those at least to a dedicated subfolder so that the packagers can more easily catch it.

> CP2K already keeps local copies of many dependencies on a local server, consumed by the toolchain. We could use that server as the source for FetchContent rather than committing archives into the source tree.

Please don't. The source should point to the original sources for easy inspection (and future metadata creation, e.g. there will be PURL support added soon). I am up for having overrides to this as you see fit, but a link to the original source should be kept. FetchContent already encourages pointing to specific git hashes, and if the user overrides it (pointing to a system package or a local copy of different version) then it is not your responsibility to support it.

As for committing archives into the source tree, depends on if you mean like `grpp` is currently, which, yes, please do not do that; or if you include git submodules, which I would not rule out immediately. One thread that you might be interested in is in the support of `cpack` and how it could interface with this, but I did not look into that myself.

> That threads all three concerns at once: builds stay airtight

Just to be clear, I do not request supporting airtight builds. It is all up to cp2k if it wishes to support it or not according to the maintenance burden that is acceptable. Supporting only the basic `FetchContent` is still enough for the users to be able to create an airtight deployment. My only concern is _if it is implemented_, then how this is handled.

Growl1234:
> And of course because it is all in `src`, the fact that this vendoring is done was completely lost during the first review of cp2k.

Libgrpp was bundled in https://github.com/cp2k/cp2k/pull/3966 between 2025.1~2025.2 release, and 2025.2.and 2026.1 was both broken in Fedora building. Maybe we should've been aware of it when dealing with making cp2k work again when updating to 2026.2, but appearently none of us has noticed it.

That being said, though it's out of scope here, there are other license "inconsistency" that were not noticed in Fedora packaging: #1632 switched the grid and offload part from GPL-2 to BSD-3-Clause, and DBM code has been under BSD-3-Clause from the first time. I believe all of them happened after CP2K had entered Fedora dist-git.

> > CP2K already keeps local copies of many dependencies on a local server, consumed by the toolchain. We could use that server as the source for FetchContent rather than committing archives into the source tree.
>
> Please don't. The source should point to the original sources for easy inspection (and future metadata creation, e.g. there will be PURL support added soon). I am up for having overrides to this as you see fit, but a link to the original source should be kept. FetchContent already encourages pointing to specific git hashes, and if the user overrides it (pointing to a system package or a local copy of different version) then it is not your responsibility to support it.

Still, my preference is not introducing `FetchContent` into CP2K.

mtaillefumier:
I think our messages are passing each other. I never claimed a local server was any better than fetching the code directly from known sources, which in my view is the only right way, for traceability reasons. I was only trying to accommodate the third_party argument, which I did not propose either.

I do read a contradiction here, because "airtight builds" can mean several things: all dependency sources being accessible, which is what the local server option addresses; or the build exiting with a clear error at configuration when a dependency is missing, assuming the code itself compiles. Those are different properties with different owners, and conflating them is most of the disagreement.

I certainly prefer FetchContent over submodules. Submodules are the wrong mechanism for handling dependencies: many projects use them, but often as a shortcut for proper dependency management, or as a remnant of an incomplete split during development. We already went through this with `DBCSR`, it was a submodule under the old Make build system, where it made some sense because Make has no dependency-resolution model, and it was removed when we moved to CMake, because CMake resolves and pins dependencies natively. Either a dependency is external and we consume it as one, or it is in-tree, but a submodule is the halfway state, and its only justification was a build system we no longer use.

oschuett:
Alright, if everyone else is convinced that adding `libwignernj` as a hard external dependency is a good idea, then I'll give up 🤷

mtaillefumier:
> > And of course because it is all in `src`, the fact that this vendoring is done was completely lost during the first review of cp2k.
> 
> Libgrpp was bundled in #3966 between 2025.1~2025.2 release, and 2025.2.and 2026.1 was both broken in Fedora building. Maybe we should've been aware of it when dealing with making cp2k work again when updating to 2026.2, but appearently none of us has noticed it.
> 
> That being said, though it's out of scope here, there are other license "inconsistency" that were not noticed in Fedora packaging: #1632 switched the grid and offload part from GPL-2 to BSD-3-Clause, and DBM code has been under BSD-3-Clause from the first time. I believe all of them happened after CP2K had entered Fedora dist-git.
> 
> > > CP2K already keeps local copies of many dependencies on a local server, consumed by the toolchain. We could use that server as the source for FetchContent rather than committing archives into the source tree.
> > 
> > 
> > Please don't. The source should point to the original sources for easy inspection (and future metadata creation, e.g. there will be PURL support added soon). I am up for having overrides to this as you see fit, but a link to the original source should be kept. FetchContent already encourages pointing to specific git hashes, and if the user overrides it (pointing to a system package or a local copy of different version) then it is not your responsibility to support it.
> 
> Still, my preference is not introducing `FetchContent` into CP2K.

The Fedora breakage is the concrete version of what this thread has been circling: bundling `libgrpp` broke downstream builds across two releases and none of us caught it, which is exactly the failure mode of vendored code in a large tree. So I take that as settling the vendoring question by evidence, don't bundle. The license drift you mention (grid/offload, DBM) is the same class of problem even if it's out of scope here: metadata that moves silently and gets caught late or not at all.

On pointing at original sources rather than a local mirror, we agree, that's what I argued for, fetch from known upstream sources for traceability, and PURL is a good reason to keep the link to origin. The local-server line was me accommodating the third_party proposal, which I've already said I don't hold; drop it.

On "not introducing FetchContent at all", I want to make sure we mean the same thing, because it might not be a disagreement. If it means preferring plain find_package against a system-provided libgrpp, with no fetching, that's the most external option and I'm fine with it, and it's probably cleanest for packagers since they provide the system copy. If it means something else, say more. Either way the shared conclusion seems to be: external, pinned, pointing at upstream, source never committed to our tree.

mtaillefumier:
This thread keeps getting read wrong because two independent decisions are collapsed into "hard dependency." They are separate axes, and separating them is most of the resolution.

Axis 1, own vs external: is the source in our tree or provided from outside? Settled: external and pinned, pointing at upstream for traceability. Not vendored, not a submodule, not committed to the tree. 

Vendoring in-tree (grpp-style) is the failure mode: the Fedora breakage across 2025.2 and 2026.1 is the concrete cost, and grpp's header carries a CP2K copyright over the upstream author's, because the banner is applied to every file by design and has no notion of third-party code. A subfolder does not fix it; grpp already has one.
Submodules are out. DBCSR was one under Make, where it made sense; It was removed during the CMake migration because CMake pins dependencies natively. A submodule is either an unfinished split or a worse FetchContent with a commit pin instead of a version.

Axis 2, hard vs soft: does the build require it? The question is orthogonal to axis 1, settling "external" says nothing about "hard," and it has barely been argued on its own terms. 

For this library, hard wins because the benefit outweighs the cost: libwignernj computes exactly, up to the final floating-point conversion, which is a correctness property the current code lacks, so requiring it buys accuracy, not convenience; and it is a self-contained library with no chain and no system surface, so the cost is near the floor. That verdict rests on this library's properties (no dependencies, exact, self-contained); so it is not a general license for hard dependencies. The calculation differs for anything with real surface, which keeps @oschuett precedent concern as the rightful guard for the next case.

The one open mechanism question: `find_package` against a system copy, or `FetchContent` pointing upstream. Both are external and pinned; a mechanism choice, not a return to vendoring.

oschuett:
I think, we all agree that we want to adopt `libwignernj` since it does indeed provide a superior implementation.

So the question is really where to put it. Again, personally I'd prefer to put it somewhere in our repo. If we can't agree on that, then we should simply treat it like any other dependency and install it via Toolchain/Spack.

PS: I passionately dislike Git submodules and CMake `FetchContent`.

Growl1234:
Maybe that was my fault not saying clearly, but I didn't mean it's the bundling of grpp that broke Fedora RPM-build; the breakage was simply because Fedora spec was left unmaintained and had some regression. What I wanted to say about grpp was only the license concern, which we didn't caught for such a long time when dealing with downstream updates.

> PS: I passionately dislike Git submodules and CMake FetchContent.

Me too.

mtaillefumier:
@Growl1234: thanks, this is a useful distinction. Agreed the build breakage itself was the unmaintained spec, not the merge. The license inconsistency you flagged is a direct consequence of merging grpp into CP2K. Which is exactly the point about carrying code we don't own: the problem gets created at merge time and then sits undetected until something unrelated forces a look. That's the cost, and it's the same thing the grpp header shows in-tree.

susilehtola:
Also, no matter what gets decided about this request, the same will also apply to previously bundled hard dependencies like libgrpp sitting in https://github.com/cp2k/cp2k/tree/master/src/grpp

My personal view is that relying on reusable libraries is a clear positive. There's no benefit to every code coming up with their in-house solution to already-solved problems. Needlessly reinventing the wheel only leads to possibility for new bugs, and expansion of the maintenance surface.

That being said, I am fine with whatever way people are comfortable to consume these reusable libraries. The clean way to so this nowadays is to use `FetchContent()`; yet, if the pre-established build conventions of legacy codes like cp2k ask for vendored libraries, this is perfectly okay, as well. I would also like to point out that Fedora is more favorable to bundled libraries nowadays than they used to be.

Still, for clarity, it's better if the vendored libraries are clearly marked as `external/` so that

1. updating the libraries to newer upstream releases is simpler and
2. any patches hopefully also get upstreamed so that other projects relying on the same libraries get the same benefits
3. binary distributions (e.g. conda and Linux distributions) can replace the vendored version with pre-packaged versions

What would not be okay is to rename the standard libraries and to obfuscate their origin (which was suggested above), since this would be a copyright and licensing violation, not to mention an unnecessary fork and introduce a code maintenance overhead.

LecrisUT:
> The one open mechanism question: `find_package` against a system copy, or `FetchContent` pointing upstream. Both are external and pinned; a mechanism choice, not a return to vendoring.

As I said [before](https://github.com/cp2k/cp2k/pull/5961#issuecomment-5567606171), these are not mutually exclusive. Starting with CMake 3.24 you get both mechanisms when you define `FetchContent`. The details on how to expose convenient control knobs for the user is flexible though.

But I do recommend using `FetchContent` over plain `find_package`. For example, it helps determine which dependency it is referring to, it declares what build options you require for it, and the users have a simpler ramp on to building by default.

Growl1234:
> Also, no matter what gets decided about this request, the same will also apply to previously bundled hard dependencies like libgrpp sitting in https://github.com/cp2k/cp2k/tree/master/src/grpp

Libgrpp has been discontinued for years. For current purpose we can simply move it out of `src/` directory, e.g. `ext/`, with CMake files written by ourselves for "internal" building. But maybe it's also worth considering an active alternative, e.g. `libecpint` or PySCF's `nr_ecp`.

mtaillefumier:
> > Also, no matter what gets decided about this request, the same will also apply to previously bundled hard dependencies like libgrpp sitting in https://github.com/cp2k/cp2k/tree/master/src/grpp
> 
> Libgrpp has been discontinued for years. For current purpose we can simply move it out of `src/` directory, e.g. `ext/`, with CMake files written by ourselves for "internal" building. But maybe it's also worth considering an active alternative, e.g. PySCF's `nr_ecp`.

GRPP support is optional in CP2K, so removing the bundled sources doesn't affect default builds; enabling it would require the library to be present, like any other optional dependency.

My take on this. I am personally in favor of excluding **any** external dependency source code (in whatever form) from CP2K source code (no ext, no externals). GRPP was the exception and I start to think I should revert my own integration and simply create a patch that fixes the orginal `CMakeLists.txt` instead of integrating the code in CP2K. 

Replacing GRPP with another library requires that the GRPP functionality used in CP2K is supported by the replacement library.

Growl1234:
> GRPP was the exception and I start to think I should revert my own integration and simply create a patch that fixes the orginal `CMakeLists.txt` instead of integrating the code in CP2K.

Yes, please go ahead. One thing is that libgrpp seems to be not in any package managers I'm familiar with, though we can add it to Spack.

> Replacing GRPP with another library requires that the GRPP functionality used in CP2K is supported by the replacement library.

If I'm not mistaken, CP2K mainly uses ECP integrals (in particular semi-local). Libecpint and PySCF nr_ecp are both promising alternatives. I see libecpint is preparing new v1.1 release; I've not checked whether it is easy to install, but I expect its integration to be a bit easier, and more robust then grpp.

oschuett:
>  GRPP was the exception and I start to think I should revert my own integration and simply create a patch that fixes the orginal CMakeLists.txt instead of integrating the code in CP2K.

So, what's the end-game here? Are we gonna accumulate patch files for all our discontinued dependencies?

Given the short-lived nature of academic funding, the vast majority of our dependencies will eventually become unmaintained. At that point we either have to drop them or take over. Since we can't afford to maintain a growing number of snowflakes, we'll have to integrate them into our code base.

I believe, `libgrpp` is just a foreshadow of what's to come.

susilehtola:
> > GRPP was the exception and I start to think I should revert my own integration and simply create a patch that fixes the orginal CMakeLists.txt instead of integrating the code in CP2K.
> 
> So, what's the end-game here? Are we gonna accumulate patch files for all our discontinued dependencies?
> 
> Given the short-lived nature of academic funding, the vast majority of our dependencies will eventually become unmaintained. At that point we either have to drop them or take over. Since we can't afford to maintain a growing number of snowflakes, we'll have to integrate them into our code base.
> 
> I believe, `libgrpp` is just a foreshadow of what's to come.

Any other project that uses libgrpp will be in the same situation. If everyone goes the vendor-in/forking route, the situation will quickly be that the library develops inconsistent forks, and of course the maintenance cost explodes.

If the original repository is unmaintained, the best option is to just create a new one and maintain it. For math libraries this shouldn't be very tricky; the algorithms don't change, the only thing needing maintenance is the build system, and possible patches against ever-stricter compilers.

Growl1234:
> Given the short-lived nature of academic funding, the vast majority of our dependencies will eventually become unmaintained. At that point we either have to drop them or take over. Since we can't afford to maintain a growing number of snowflakes, we'll have to integrate them into our code base.

That's why I mentioned libecpint etc. If a library is discontinued and we still need the feature, try to look for an actively maintained alternative and migrate to it. Implementing it internally is usually a "fallback" option.

oschuett:
> Any other project that uses libgrpp will be in the same situation.

Are there any other users of `libgrpp`? AFAIK, we are the only ones, and I believe this is true for at least half of our dependencies.

While the number of libraries keeps growing, the projects that can actually integrate them (like us) are diminishing.

DCM-Uni-Paderborn:
Hi everybody,

I support adopting libwignernj. Improving the numerical robustness of the Gaunt coefficients and consolidating the duplicated implementation are worthwhile changes.

During review, I found two CMake integration issues, reproduced with small isolated tests:

- When using an external libwignernj, the exported CP2K targets reference `cp2k::libwignernj::wignernj`, but `cp2kConfig.cmake` does not recreate that dependency for downstream consumers.
- Switching an existing build to `CP2K_USE_BUNDLED_LIBWIGNERNJ=ON` leaves the cached external-library detection active, resulting in a missing-target error.

I would be happy to help address these issues and the merge conflict.

These are integration concerns, not objections to the numerical improvement itself.

Greetings,
Thomas

cmake/modules/FindLibwignernj.cmake
DCM-Uni-Paderborn: Implemented in 0820ff863e: discovery now uses `find_package(wignernj 0.8 CONFIG QUIET)` and the upstream `wignernj::wignernj` target; the custom finder is removed. The bundled fallback is preserved. Installed CP2K also calls `find_dependency(wignernj 0.8 CONFIG REQUIRED)` when built against the external library. I verified this with the real upstream 0.8.0 package and a separate project linking the installed `cp2k::cp2k` target.

tools/precommit/precommit.py
DCM-Uni-Paderborn: For automatically discovered files, yes. However, `main()` converts explicitly supplied filenames to absolute paths before changing to the repository root. Without the normalization, `precommit.py src/grpp/grpp_binomial.c` selects the GPL banner instead of the directory-specific license. I kept the normalization, clarified the comment, and added a regression test covering both discovered/relative and explicit/absolute paths in e26d2be77d.

tools/precommit/check_file_properties.py
DCM-Uni-Paderborn: I tried this simplification, but it also applies the CP2K copyright-year requirement to the top-level GPL `LICENSE`, which must remain unchanged. In e26d2be77d I removed the bundled-path special case and instead restricted the year check to the three CP2K-owned LICENSE files in dbm, grid and offload. The new tests cover the root GPL text, upstream licenses and outdated CP2K copyright notices; they are included in the Miscellaneous tester.

DCM-Uni-Paderborn:
The integration fixes are now pushed in 0820ff863e and e26d2be77d, rebased onto master at 180177c41b. My initial update used a merge commit, which the Git History check correctly rejected. This has been replaced with a linear history using an explicit force-with-lease. The final source tree is identical to the tested version (Git tree 5fcaeda2d23b95de8ddfdb81b919d893ab5e499d). The merge conflict is resolved, preserving the newer GAPW initialization diagnostics. I have also replied to the individual CMake and precommit review comments.

Local validation on macOS ARM64, GCC 16 / OpenMPI:
- Full CP2K build and installation with upstream libwignernj 0.8.0.
- A separate downstream CMake consumer links the installed CP2K library and initializes/finalizes successfully with one and two MPI ranks.
- External -> bundled -> external reconfiguration works, including a deliberately stale discovery-cache entry; automatic fallback without the external package also works.
- 811,538 real/complex Gaunt values are bitwise identical between the bundled implementation and upstream's shared library.
- All 53 tests in GAPW-1/2/3, GAPW-ext, CNEO and LIBTEST/misc pass with 2 MPI ranks x 2 OpenMP threads. The Gaunt/Lebedev maximum deviations are 1.776e-15 (real) and 1.943e-15 (complex).
- All 54 changed files pass precommit checks, including the three new regression tests; source-list consistency and Dockerfile generation checks also pass.

For completeness, the two additional LRI directories are not green in this local configuration: the PR has three wrong results and eight DPOTRF failures. I rebuilt unmodified master with the same configuration: it fails all eleven of those cases too, plus one additional energy check (58/70 correct on master versus 59/70 on the PR). The wrong energies differ between revisions, so this is a remaining local LRI validation limitation, not a claim of numerical equivalence. No reference values, tolerances or test inputs were changed.

For dependency integration, my preference is to retain the existing system-package-first / bundled-fallback arrangement for this PR: use upstream's native CMake package when available, retain an offline fallback, and allow packagers to require the external library explicitly. This introduces neither FetchContent nor automatic downloads. The third-party directory layout and source-formatting policy are unchanged by my fixes.

@fstein93 @oschuett @Growl1234, could you take another look at this concrete arrangement? I have left the outstanding review decisions in place; the new CI results and agreement on the integration choice are still needed before merging.

DCM-Uni-Paderborn:
Hi everybody,

I think we should now separate the numerical improvement from the dependency-management question and converge on a concrete solution for this PR.

I support adopting libwignernj and replacing the existing Gaunt-coefficient implementation. Maintaining the older, less accurate algorithm as an alternative would add another code path and a continuing testing burden without a clear scientific benefit.

For the integration, CP2K’s FFTSG/FFTW3 arrangement provides a useful precedent: an internal implementation remains available, while an external library is preferred when available. The analogy is not exact. Here, the fallback can use the same accurate implementation as the external library, rather than retaining a different numerical algorithm.

My preferred solution for this PR is therefore:

- Prefer an installed libwignernj through its native CMake package.
- Retain a versioned, clearly identified bundled copy as the fallback.
- Preserve upstream attribution and licensing, minimize local modifications, and document how the bundled copy is updated.
- Test both integration paths and allow packagers to require the external library explicitly.

This keeps the external installation optional without making the numerical improvement conditional on it. It also avoids introducing FetchContent or automatic downloads during CP2K configuration.

Bundling does carry a maintenance cost, and the concerns about source formatting and upstream synchronization deserve concrete answers. However, I consider this a reasonable, limited compromise for this small mathematical library.

Frederick, your original review requested an opt-in solution. Does this arrangement address your concern about adding a mandatory external dependency, or do you still consider retaining the old numerical implementation necessary?

I would like us to settle that specific point before making further architectural changes.

Greetings,
Thomas

Growl1234:
It has been clear that directly bundling it into source with adapting would cause endless technical concerns for build systems and package managers, so I believe 90% people joining the discussion would vote for using external library. Using FFTSG/FFTW is no longer a good argument as FFTSG will soon be dropped and FFTW will become a hard dependency.

fstein93:
I agree with @DCM-Uni-Paderborn's solution of owning a local, well-defined copy. I guess that it will be tested with the minimal tester. That way, there is no additional code path which reduces the maintenance costs and we always have the higher accuracy available with this library.

fstein93 (approved):

mkrack:
Frankly speaking, I dislike this duplication. When we have an external `libwignernj` as code path available why do we need then an internal code path which is possibly not in sync with the other one and thus causing an additional maintenance burden. This is worse in my opinion. I vote for an external `libwignernj` only.

oschuett:
I think, we should leave this decision up to @mkrack  and @Growl1234 as the Spack/Toolchain work will mostly fall on their shoulders.

mtaillefumier:
Including the code of an external dependency can be source of problems and there is no reason to do this as the library is maintained externally. If we continue this path, cp2k will include the source code of all dependencies which becomes unpractical and a burden on the build system side. Keeping this simple reduces maintenance in the long run.

Growl1234:
I've dropped the local Spack recipe of libwignernj since https://github.com/spack/spack-packages/pull/6376.

DCM-Uni-Paderborn:
Fixed the Spack dependency-build failure in dbec56d638.

The updated builtin repository exposes Yasm's CMake build, which is selected by CP2K's package preferences, but the Yasm recipe does not declare the C++ compiler dependency required by that build. This produces the reported `SPACK_CXX_* variables not set` error before CP2K is compiled.

The fix selects Yasm's Autotools build in the MPI, serial and serial-static dependency environments. It leaves the builtin repository pin, upstream libwignernj package, scientific code, regression inputs, reference values and tolerances unchanged.

Local validation with Spack 1.2.2 and the exact pinned package repository reproduced the original CMake failure. Fresh concretization selects Autotools after the fix in all three configurations, and the actual Yasm build and all 44 upstream Yasm tests pass on macOS ARM64 with GCC 16. Changed-file precommit checks also pass. The full Linux Spack run still needs to validate the updated PR head.

/cp2kci spack-pdbg

tools/fedora/cp2k.spec
susilehtola, line 1: Note that this file is out of date with https://src.fedoraproject.org/rpms/cp2k/blob/rawhide/f/cp2k.spec

Growl1234: There's some minor modification since 2026.2 release. That being said, I think the Packit stuck here is more likely a CI or Copr issue.

LecrisUT: There are also a few changes that were made in downstream that would need to be backported, but we still need to finish up building with libxs and merge them together, so would not worry about this atm. We can also realign them when the downstream PR is made so would not make it be a blocker on either side.

Re the builds being stuck, it seems like a copr issue. I'll forward it to them. Can retrigger it though

Growl1234 (approved):

tools/spack/cp2k_deps_p.yaml
```diff
@@ -184,9 +184,14 @@ spack:
       require:
         - "@1.21.0"
 
+    # Yasm's CMake recipe does not declare its C++ compiler dependency.
+    yasm:
+      require:
+        - "build_system=autotools"
+
```
mkrack: This has been fixed [now](https://github.com/spack/spack-packages/commit/2cbb70f19fd83c7f57adf6c424826ead9600e85b).

tools/spack/cp2k_deps_s-static.yaml
```diff
@@ -95,9 +99,14 @@ spack:
       require:
         - "~shared"
 
+    # Yasm's CMake recipe does not declare its C++ compiler dependency.
```
mkrack: see above

tools/spack/cp2k_deps_s.yaml
```diff
@@ -112,9 +112,14 @@ spack:
       require:
         - "+hdf5"
 
+    # Yasm's CMake recipe does not declare its C++ compiler dependency.
```
mkrack: see above

PR #5969, opened 2026-09-05
Stabilize and speed up regression tests

DCM-Uni-Paderborn:
The Al common-mu OT regression can stall from an atomic guess, and tightening its SCF threshold made it too slow for CI. Start it from the wavefunction produced by the existing fixed-moment test and lower the electronic temperature to 750 K, as suggested in review. Keep its original `EPS_SCF=5e-7` and remove the explicit `PARALLEL_GROUP_SIZE 0`. Its three numerical references come from an independent cold-start calculation at `EPS_SCF=1e-9`; all matcher tolerances are unchanged.

The subsequent CI run exposed a separate Na SD convergence plateau. The first physical virtual-subspace rebuild requires at least 12 accepted search directions and a residual above `10 * EPS_SCF`. With `EPS_SCF=5e-9`, the residual can fall below this window before the first rebuild and stall near `4e-8`. Tighten the Na test to `EPS_SCF=1e-9` so the rebuild is reached. Retain the SD algorithm, iteration limit, reference energy and the explicit rebuild matcher.

The serial CI passed every numerical check but flagged `QS/regtest-kp-2` at 287.49 s against a 287.39 s directory limit. Move its two independent Al Davidson inputs, byte-for-byte, to the existing Davidson group in `QS/regtest-kp-3`. This redistributes about 11.5 s in that CI report, without adding or removing tests or changing the slow-test limits.

Validation:

- All 66 input registrations and 110 matcher specifications across the three affected directories are preserved by this follow-up.
- The Na test and a diagnostic stepsize variant pass with 1, 2 and 4 MPI ranks on macOS GCC 16; the diagnostic variant reproduces the plateau with the old SCF threshold and converges in 13 steps with the new threshold.
- Both Na variants also pass with GCC 15/MPICH at 2x1, retaining the energy and physical-subspace rebuild checks.
- Linux GCC 15/MPICH: 114/114 checks across `regtest-ot-3-4`, `regtest-kp-2` and `regtest-kp-3` at 2x1 in the normal, non-keepalive runner.
- macOS GCC 16/OpenMPI: 95/95 checks in `regtest-ot-3-4` and `regtest-kp-3` at 2x1, including the relocated Davidson tests in a build without Libint.
- Linux GCC 16/OpenMPI Debug: 24/24 checks in `regtest-ot-3-4` at 1x1 using the standard keepalive runner.
- Targeted precommit checks and `git diff --check` pass.

The additional Linux GCC 16/OpenMPI Debug run in normal process-per-input mode reports third-party LeakSanitizer failures at MPI shutdown, including for untouched tests. It is not counted as passing validation. The updated Na SCF converges in 13 steps in that run. Full required CI will rerun after the push.

### Serial runtime follow-up

The next sdbg run passed all 6,125 numerical checks and every directory runtime
limit, but flagged two individual slow tests: the restricted TPSS force test
(28.14 s) and the silicon TBLITE LBFGS/rotation test (31.67 s), against a
26.37 s threshold.

- Reduce the silicon test's inner `MAX_SCF` from 25 to 10 so the existing outer
  SCF updates its preconditioner earlier. Keep the cold MOPAC guess, IRAC,
  LBFGS, rotations, SCF thresholds, energy reference and four-k-point check.
  On Terok, the reported total SCF work drops from 105 to 66 steps and the
  CP2K timer drops from 13.42 to 8.87 s.
- Run the restricted TPSS test as `ENERGY_FORCE` and check three signed atomic
  force components directly, together with the ground- and excited-state
  energies. This retains the analytic third-derivative force calculation but
  removes the second SCF/force evaluation after a geometry step. The new
  references come from the first geometry of the unmodified test; the ground-
  state energy tolerance stays at 1e-7, and the new force tolerances are 1e-6
  relative. The grid, functional, SCF and response convergence settings are
  unchanged. The CP2K timer drops from 19.28 to 13.53 s on Terok and from
  6.38 to 3.97 s on macOS.

No tests, dependencies, existing tolerances or runtime limits are disabled or
relaxed. Production code is unchanged; the only Python additions are three
force-component matchers.

Validation of this follow-up:

- GCC 15/MPICH: all 23 checks in both complete affected directories pass at
  1x1 and 2x2, with one process per input.
- macOS GCC 16/OpenMPI: all 12 checks in the complete TDDFPT directory pass
  at 1x1. This build has no TBLITE, so the silicon validation is on Terok.
- At 4x1, all seven checks of the two modified inputs pass. The full run is
  22/23 because the unchanged `h2o_lda_analytic.inp` differs from its reference
  by 9.47e-8 Ha. A separate run of that unchanged input reproduces the same
  value (-17.12646999205413 Ha); its reference and tolerance are untouched.
- A ground-state-only negative control fails the excited-state energy and
  all three force checks, while its ground-state energy still passes.
- Targeted matcher probes verify component selection, atom-index boundaries
  and rejection of missing force output.
- `make_pretty.sh` passes for all four changed files; `git diff --check` passes.

CI subsequently confirmed this follow-up: sdbg passed all 6,129 checks with no
slow tests. The pdbg run passed all 6,218 numerical checks but flagged the
separate fine-grid GAPW case below.

### Fine-grid GAPW runtime

`QS/regtest-acc-5-1/h2o_f01_fine.inp` took 31.38 +/- 1.70 s against a
29.05 s threshold in pdbg. Reuse the wavefunction already produced by the
preceding `h2o_f01.inp` and use `FULL_ALL` with its default stepsize. The SCF
converges in eight rather than 19 iterations. Three interleaved two-rank,
one-thread timing pairs on Terok average 11.10 s before and 9.66 s after,
about a 13% reduction.

Keep the separate `td_dipole_fine` project name, so this input does not overwrite
the shared restart used by the other tests. All seven input registrations and 14 matcher
specifications remain unchanged. The fine XC grid, GAPW integration, numerical
derivative stencil, SCF/response thresholds, references and tolerances are
unchanged; no suppressions or runtime-limit changes are introduced.

The complete directory passes all 14 checks on macOS GCC 16/OpenMPI at 1x1.
The complete-directory batch on Terok also completed successfully at 2x1,
2x2 and 4x1; retrieval of its detailed logs was interrupted by loss of local
DNS resolution for Terok after the batch completed. An additional macOS
comparison of all nine atomic force components differs from the original
cold-start result by at most 3.24e-7 Ha/Bohr. `make_pretty` and
`git diff --check` pass. These are local validation results; the updated
full CI still needs to run.


Growl1234:
I'd again suggest using lower `ELECTRONIC_TEMPERATURE` instead. Also drop `PARALLEL_GROUP_SIZE 0` to make it faster.

DCM-Uni-Paderborn:
Thanks for the suggestions. I have lowered the temperature to 750 K, restored
the original EPS_SCF, and removed PARALLEL_GROUP_SIZE 0. Lowering the temperature
alone still left a cold-start plateau, so the test now reuses the wavefunction
from the existing fixed-moment test. All tolerances are unchanged, and all 24
checks pass in six configurations, with the common-mu test taking 0.31-1.00 s.
A fixed-moment control still fails all three numerical matchers.

PR #5971, opened 2026-09-05
Low-Scaling GW and RI-RS: auto-detect MEMORY_PER_PROC and honour CUTOFF_RADIUS_RI in the RI-RS grid optimizer

RitajTyagi:
### `MEMORY_PER_PROC` is now detected automatically

`MEMORY_PER_PROC` now defaults to `-1` = detect from node free memory / ranks per node
A positive value is still honoured.

### RI-RS grid optimizer uses the RI metric cutoff

`gw_optimize_ri_rs_grid` built its three-centre integrals with a hardcoded
3 Å truncated-Coulomb cutoff instead of `bs_env%ri_metric%cutoff_radius` (Reported by: @RemiPasquier ), so
`CUTOFF_RADIUS_RI` was ignored during grid optimization while being used
everywhere else. No change at the default (3 Å).


src/gw_utils.F
```diff
@@ -424,10 +427,110 @@ SUBROUTINE read_gw_input_parameters(bs_env, bs_sec)
...
+!>        /proc/meminfo reports node-wide memory, so every rank on a node reads the same figure
+!>        and divided by the number of ranks sharing that node, identified with
+!>        MPI_COMM_TYPE_SHARED, which splits the communicator by shared-memory
+!>        domain. The result is minimized over all ranks so that every rank works with the same, most
+!>        pessimistic figure.
+!>
+!> \param para_env ...
+!> \param mem_avail_GB ...
+! **************************************************************************************************
+   SUBROUTINE mem_avail_per_proc_GB(para_env, mem_avail_GB)
```
fstein93, line 506: This and the preceding routine are very useful features and should not be hidden inside of the GW-module but made available elsewhere for everyone. `machine.F` will not work because of circular dependencies (`message_passing.F` points to it) but maybe `message_passing.F` could already be the right spot.

src/gw_utils.F
```diff
@@ -424,10 +427,110 @@ SUBROUTINE read_gw_input_parameters(bs_env, bs_sec)
...
+
+      CHARACTER(LEN=*), PARAMETER :: routineN = 'resolve_memory_per_proc'
+      REAL(KIND=dp), PARAMETER :: detected_memory_fraction = 0.8_dp, &
+         fallback_memory_per_proc_GB = 2.0_dp
+
+      INTEGER                                            :: handle
+      INTEGER(KIND=int_8)                                :: mem_occ_per_proc
+      REAL(KIND=dp)                                      :: mem_free_per_proc_GB
+
+! Safet net over detected memory fraction
```
fstein93, line 455: Probably `Safety net`

JWilhelm (changes requested):
@fstein93 's suggestion sounds promising, knowing the free memory per mpi process would indeed help in many parts of CP2K... 

Would it be also possible to unify the check of free memory in "ri_rs_mem_avail_per_proc_GB" and "floquet_determine_subgroup_size" with the potential new general functionality?

JWilhelm (approved):

fstein93 (approved):

PR #5973, opened 2026-09-05
PW/FFT: Fix the plan-6 scratch buffer and assert the plan geometry

s8ga:
The backward z-stage plan of the parallel RS-PW transform is created with a scratch buffer that does not match its declared geometry. The routine signature is

    fft_create_plan_1d(plan, fsign, trans_in, trans_out, ldx_in, ldx_out, n, m, zin, zout)

and the creation call is

    CALL fft_create_plan_1d(fft_scratch%fft_plan(6), BWFFT, .FALSE., .TRUE., &
                            n(3), mx1*my1, n(3), mx1*my1, &
                            fft_scratch%p3buf, fft_scratch%p1buf)

With trans_in = .FALSE. the declared input of this plan is n(3)*mx1*my1 complex values in z layout, that is the shape [n(3), mx1*my1]. The scratch buffers are allocated as p2buf [n(3), mx1*my1] and p3buf [mx2*mz2, n(2)]. The z-layout buffer for this stage is p2buf, and the forward counterpart plan 1 is created with p1buf and p2buf for the same geometry. The call above passes p3buf, which belongs to the y stage and can be smaller than the declared input.

This breaks the convention that the creation-time zin and zout must match the declared n, m and ldx geometry. The FFTW3 backend hands these arrays to fftw_plan_many_dft, so the planner treats them as the memory the declared tensor lives in.

Execution always passes p2buf explicitly, so the computed results were never affected. FFTW_ESTIMATE planning never dereferences the creation-time arrays, so the mismatch stayed harmless. MEASURE class planning zeroes the declared tensor through the creation-time pointer and runs timed trial transforms on it. Current master 2d1258df plans every FFTW3 transform with MEASURE class flags because of the planner flag confusion that is fixed separately. Whenever n(3)*mx1*my1 exceeds the size of p3buf the planning writes past the end of p3buf and corrupts the heap.

A water dimer geometry optimization crashes reliably at 20 MPI ranks within the first SCF on two different machines, as a glibc heap abort or as SIGSEGV. The reproducer input and the crash outputs are attached.

[plan6.tar.gz](https://github.com/user-attachments/files/31868680/plan6.tar.gz)

### DIFF
--- a/src/pw/fft_tools.F
+++ b/src/pw/fft_tools.F
@@ -3256,7 +3256,7 @@ CONTAINS
                                        fft_scratch_new%fft_scratch%p4buf, fft_scratch_new%fft_scratch%p3buf)
                CALL fft_create_plan_1d(fft_scratch_new%fft_scratch%fft_plan(6), BWFFT, .FALSE., .TRUE., &
                                        n(3), mx1*my1, n(3), mx1*my1, &
-                                       fft_scratch_new%fft_scratch%p3buf, fft_scratch_new%fft_scratch%p1buf)
+                                       fft_scratch_new%fft_scratch%p2buf, fft_scratch_new%fft_scratch%p1buf)
 
             CASE (400) ! serial FFT
                np = 0
--- a/src/pw/fft_tools.F
+++ b/src/pw/fft_tools.F
@@ -3239,21 +3239,31 @@ CONTAINS
                END DO
 
                !set up fft plans
+               ! The scratch buffers must cover the geometry declared in each call
+               ! below, because the FFTW planner may write the buffers during
+               ! planning. The plans pair the buffers such that each of p1-p6 is
+               ! checked once; p7 shares the size of p6.
+               CPASSERT(SIZE(fft_scratch_new%fft_scratch%p1buf) >= mx1*my1*n(3))
                CALL fft_create_plan_1d(fft_scratch_new%fft_scratch%fft_plan(1), FWFFT, .TRUE., .FALSE., &
                                        mx1*my1, n(3), n(3), mx1*my1, &
                                        fft_scratch_new%fft_scratch%p1buf, fft_scratch_new%fft_scratch%p2buf)
+               CPASSERT(SIZE(fft_scratch_new%fft_scratch%p3buf) >= mx2*mz2*n(2))
                CALL fft_create_plan_1d(fft_scratch_new%fft_scratch%fft_plan(2), FWFFT, .TRUE., .FALSE., &
                                        mx2*mz2, n(2), n(2), mx2*mz2, &
                                        fft_scratch_new%fft_scratch%p3buf, fft_scratch_new%fft_scratch%p4buf)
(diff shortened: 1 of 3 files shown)


oschuett:
> A water dimer geometry optimization crashes reliably at 20 MPI ranks within the first SCF on two different machines, > as a glibc heap abort or as SIGSEGV. The reproducer input and the crash outputs are attached.

Is there a way we could design a tests that would have caught this on 2 or 4 MPI ranks?

fstein93:
On 4 MPI ranks, yes by using a very small grid (2x2x2). With 2 MPI maybe (1x1x2 or similar).

s8ga:
> Is there a way we could design a tests that would have caught this on 2 or 4 MPI ranks?

I just constructed input that reliably triggers the crash on 4 MPI ranks + 1 OMP thread, which I attached here.
[np4_reproduce.tar.gz](https://github.com/user-attachments/files/31881258/np4_reproduce.tar.gz)

I'd suggest catching this at the source instead: a `CPASSERT` in `fft_create_plan_1d` that `SIZE(zin/zout) >= n*m` (needs assumed-shape dummies). Any mispairing then aborts at plan creation.

Would you prefer to add some input as a regtest here, or skip the regtest and instead address buffer validation via CPASSERT in follow up pr?

oschuett:
Generally, the `CPASSERT` is preferable because it catches the problem earlier.
Nevertheless, the regtest might cover some additional code paths / corner cases that we haven't thought of yet.
So, perhaps we'll add both?

s8ga:
Both are in now.

The `CPASSERT` went next to the plan creation calls instead of inside `fft_create_plan_1d`: the buffers arrive there as assumed-size dummies, so `SIZE` is not available without an interface change, and most callers pass rank-2 scratch arrays. Each of the six CASE 300 calls is now preceded by an assert that its input buffer covers the declared geometry, so any mispairing aborts at plan creation before the planner writes anything.

For the regtest I reworked the np4 reproducer into a regular converged ENERGY test (`H2O-fft-plan6.inp`, 13 SCF steps, ~0.5 s on 4 ranks): the cell makes the coarsest multigrid level (2, 24, 25), which takes a 2D layout, and `FFTW_PLAN_TYPE MEASURE` re-arms the planner, so the unfixed code aborts with heap corruption from 4 ranks on (1 OpenMP thread) and the fixed code matches the energy on 1, 2, 4 and 8 ranks.

One caveat on what it covers: the bug path needs at least 4 MPI ranks. On 2 ranks the 2D layout is unreachable ((1,2) aborts by design, (2,1) takes the correct plan-200 path), so the PR CI runs the test but cannot trigger the old bug there. Also, the coverage page shows the entire CASE 300 block with zero hits in the aggregated regtest suite, so this test is its first entry; within the CI it additionally covers MEASURE planning on a block-distributed grid.

fstein93 (approved):

jpoto:
Hi @s8ga Could you provide a way to contact you at j.pototschnig@hzdr.de ?

s8ga:
Hi @jpoto, I've sent an email to your HZDR address from my university email.

PR #5974, opened 2026-09-05
QS: Unify k-point operator service

### LINKED ISSUE #5952: RFC: Unified k-point operator service
Following the discussion with @Growl1234 in #5950 (and earlier in #5934), and building on the k-point operator distribution introduced by @DCM-Uni-Paderborn in #5926: the complex k-point operator assembly (phase transform, densification, and the pool-to-group redistribution) currently exists as several parallel implementations.
This issue opens the discussion about unifying them behind one service: a sketch of what that could look like follows, together with the questions where input would help.

**The implementations today:**

- `kp_evp_op_start` / `kp_evp_op_finish_cfm` (from #5950), the two-phase form used by the STANDARD assembly loop, the ADIIS blend, `diag_kp_basic`, `diag_kp_smat`.
- `distribute_kpoint_operator_channel` (#5945), the one-shot form used by the Davidson driver, its preconditioner, and (through the `qs_scf_distribute_kpoint_operator` helper) the Mermin exact-Hxc distribution.
- Three further copies of the same pipeline in `qs_wf_history_methods.F` (wavefunction extrapolation).

## Proposed implementation draft

Names are provisional; the structure is the proposal.

### Module interface

```fortran
MODULE qs_kpoint_operators
   PUBLIC :: kpoint_operator_context_type, &
             kpoint_operator_context_create, kpoint_operator_context_release, &
             kpoint_operator_get, kpoint_operator_start, kpoint_operator_finish

   ! Named slots (input_constants.F style)
   INTEGER, PARAMETER, PUBLIC :: kpoint_slot_ks = 1, kpoint_slot_s = 2, kpoint_slot_t = 3
```

### The context type

The guiding rule: **reference the host, own only what is missing.**

**Reference, do not copy.** The layout and geometry knowledge stays in `kpoint_type` as the single source of truth: about 80% of the payload already lives there (`kp_dist`, `kp_range`, `xkp`, `cell_to_index`, `sab_nl`, `blacs_env_all`, `blacs_env`).
The context holds a pointer to the host and reads it exactly as the call sites do today (`get_kpoint_info(kpoints, ...)`).
The members that do change during a run (`cell_to_index` and `sab_nl` are reallocated at every energy evaluation, not only when the geometry moves) are therefore always read fresh.
A context that copied these members instead would serve stale data after the very first energy evaluation.

**Own only what is missing.** The context allocates, and therefore releases, only the objects it created itself; it never takes over host state, so its release cannot touch objects it did not create.

Every member of the context type is either a pointer into the host, with its lifetime tied to `kpoint_type`, or an object the context created, managed by `kpoint_operator_context_create` / `_release`. There is no third category.

What is missing today: each frontend privately owns its own instance of the same scratch family that the context is designed to own once.

| Frontend | Private scratch today |
|---|---|
| Davidson driver (its preconditioner shares this set since #5945) | `rmatrix`, `cmatrix`, `tmpmat`, `fmwork`, `fmlocal_re`, `fmlocal_im` |
| STANDARD assembly loop | the same trio, plus `fmwork(1:4+)`, `fmlocal`, `fmdummy`, and `info(kplocal*nspin*nkp_groups, ninfo)` |
| wavefunction extrapolation (`qs_wf_history_methods.F`) | `rmat_ft`, `cmat_ft`, `tmpmat_ft`, `fmlocal_ft`, `fmdummy_ft`, `info_ft(kplocal*nkp_grps, 2)` |
| Mermin helper (`qs_scf_distribute_kpoint_operator`) | `matrix_re`, `matrix_im`, `matrix_work`, created from a template on every call and released with it |

The context becomes the single owner of this family.

### Encapsulation

The service API is designed so that the collective protocol cannot be violated through it: the walk is internal, the receiver routing is internal, and the slots are opaque names.
What remains is direct access to the context components.
The components are private to the module, so such access fails at compile time rather than corrupting state at run time.
Inside the service, the routines assert their invariants: starting a transfer on a slot whose previous transfer is still in flight aborts, and finishing a slot that is not in flight aborts.
Query accessors can be added later if debugging needs them.
Legitimate access has established paths in the codebase: read access goes through getters (`get_kpoint_info` and `cp_fm_struct_get` return pointers to the host data), and a new capability enters as a new service routine, the way each of the three output representations here originated in a real consumer.
`coulomb_integral_context_type` is an existing context type with private components, operated entirely through its interface routines.

```fortran
TYPE :: kpoint_operator_context_type
   ! ---- host references: no ownership, no copies ----
   TYPE(kpoint_type), POINTER       :: kpoints        ! layout + geometry

   ! ---- owned: scratch and bookkeeping (the missing 20%) ----
   ! DBCSR work trio, created lazily from the host template on first use;
   ! the shapes are stable, so the scratch is reused across SCF steps.
   TYPE(dbcsr_type)                 :: rmatrix, cmatrix, tmpmat
   ! Pool-context densification workspaces + the destination pair of the
   ! redistribution (fmlocal on the owning group, fmdummy elsewhere).
   TYPE(cp_fm_type), ALLOCATABLE    :: fmwork(:)
   TYPE(cp_fm_type)                 :: fmlocal, fmdummy
   ! Transfer bookkeeping: one pair of copy_info states (re/im) per
   ! k-point, spin, and named slot. The STANDARD batch launch keeps
   ! kplocal*nspin*nkp_groups k-point instances in flight at once, so this
   ! is allocated at context creation and addressed internally.
   TYPE(copy_info_type), ALLOCATABLE :: info(:, :, :)
   ! Grid strategy state stays with the callers for the initial version:
   ! its validity is tied to a particular operator generation.
END TYPE
```

### Construction, lifecycle, invalidation

```fortran
CALL kpoint_operator_context_create(ctx, kpoints)      ! pulls everything from the host
CALL kpoint_operator_context_release(ctx)              ! at the exit of each driver routine
```

Proposed call sites: per driver.
Each driver routine (`do_general_diag_kp`, the Davidson driver, `diag_kp_basic`, `diag_kp_smat`, the Mermin distribution) creates the context on entry, releases it on exit, and passes it to its helpers as an argument.
`kpoint_types.F` keeps no footprint of the context, so the new infrastructure stays self-contained.
The context then lives for one diagonalization call; since persistent caching is out of scope, no state needs to outlive it.

### The service routines

One-shot form, for callers that consume one operator at a time and want the complete result from a single call.
The caller passes its local k-point; the service walks all k-point groups internally (every group participates in the collective redistribution, only the owning group receives), and the operator is complete in the requested representation when the call returns.
No transfer state is exposed to the caller.
This is the usage pattern of `distribute_kpoint_operator_channel` today: the Davidson driver, its preconditioner, and the Mermin exact-Hxc distribution.

```fortran
! ikp: the caller's local k-point index
CALL kpoint_operator_get(ctx, ikp, ispin, rsmat, cmat=cksmat)                  ! assembled cfm
CALL kpoint_operator_get(ctx, ikp, ispin, rsmat, fm_re=rmat, fm_im=cmat)       ! raw fm pair
CALL kpoint_operator_get(ctx, ikp, ispin, rsmat, mat_re=s_re, mat_im=s_im)     ! dbcsr pair
```

Outputs are caller-allocated and filled by the service, following the pattern of the channel routine and of `kp_evp_op_finish_cfm` today; nothing aliases context scratch, and nothing is invalidated by the next call.

```fortran
SUBROUTINE kpoint_operator_get(ctx, ikp, ispin, rsmat, ...)
   DO igroup = 1, nkp_groups                            ! the walk lives here, once
      ik_glob = kp_dist(1, igroup) + ikp - 1
      my_kpgrp = ...                                    ! receiver routing lives here
      ! body = today's kp_evp_op_start: plain or grid transform, densify,
      ! start the redistribution to fmlocal/fmdummy
      IF (my_kpgrp) THEN
         ! finish the transfer and emit the requested representation:
         ! cfm assembly (z_one/gaussi), raw fm pair, or block-cyclic dbcsr
         ! pair (replacing the copy_fm_to_dbcsr_bc detour of #5945)
      END IF
   END DO
END SUBROUTINE
```

The Davidson driver site, today (`do_block_davidson_diag_kp` in `src/qs_scf_diagonalization.F`) and after:

```fortran
! Today
CALL distribute_kpoint_operator_channel(matrix_s_kp, kpoints, local_kpoint, 1, &
   kp_dist, xkp, cell_to_index, sab_nl, rmatrix, cmatrix, tmpmat, fmwork, &
   fmlocal_re, fmlocal_im)
CALL cp_cfm_scale_and_add_fm(z_zero, csmat_base, z_one, fmlocal_re)
CALL cp_cfm_scale_and_add_fm(z_one, csmat_base, gaussi, fmlocal_im)

! After
CALL kpoint_operator_get(ctx, local_kpoint, 1, matrix_s_kp, cmat=csmat_base)
```

The preconditioner site (`davidson_kp_prepare_preconditioners`), which needs the DBCSR pair:

```fortran
! Today
CALL distribute_kpoint_operator_channel(matrix_s_kp, kpoints, local_kpoint, 1, &
   kp_dist, xkp, cell_to_index, sab_nl, rmatrix, cmatrix, tmpmat, fmwork, &
   fmlocal_re, fmlocal_im)
CALL dbcsr_init_p(matrix_s_re); CALL dbcsr_init_p(matrix_s_im)
CALL copy_fm_to_dbcsr_bc(fmlocal_re, matrix_s_re)
CALL copy_fm_to_dbcsr_bc(fmlocal_im, matrix_s_im)

! After
CALL kpoint_operator_get(ctx, local_kpoint, 1, matrix_s_kp, mat_re=matrix_s_re, mat_im=matrix_s_im)
```

Two-phase form, for callers that assemble many operators in a loop and hide the redistribution latency behind their own work.
`kpoint_operator_start` takes the caller's local k-point, walks the groups internally, and launches the operator's redistribution for every group before returning; the caller keeps starting the next operator while the transfers are in flight; `kpoint_operator_finish` later reclaims the transfer on every group, addressed through the local k-point, the spin, and the named slot: finishing it on the owning group and cleaning up the dummy on the rest, so the callers' trailing `cp_fm_cleanup_copy_general` loops disappear.
This is the usage pattern of the STANDARD assembly loop today: KS and S of one k-point are started, the loop moves on while they fly, and the finishes come after the loop.

```fortran
CALL kpoint_operator_start(ctx, ikp, ispin, rsmat, slot=kpoint_slot_ks)
CALL kpoint_operator_finish(ctx, ikp, ispin, slot=kpoint_slot_s, cmat=csmat)
```

The STANDARD assembly site (`do_general_diag_kp`), before #5950, today, and after the unification:

```fortran
! Before #5950: the KS assembly, inlined in the driver
IF (use_grid_fft) THEN
   CALL rskp_transform_grid_extract(ks_grid(ispin), ik, rmatrix, cmatrix)
ELSE
   CALL dbcsr_set(rmatrix, 0.0_dp)
   CALL dbcsr_set(cmatrix, 0.0_dp)
   CALL rskp_transform(rmatrix=rmatrix, cmatrix=cmatrix, rsmat=effective_ks, ispin=ispin, &
                       xkp=xkp(1:3, ik), cell_to_index=cell_to_index, sab_nl=sab_nl)
END IF
CALL dbcsr_desymmetrize(rmatrix, tmpmat)
CALL copy_dbcsr_to_fm(tmpmat, fmwork(1))
CALL dbcsr_desymmetrize(cmatrix, tmpmat)
CALL copy_dbcsr_to_fm(tmpmat, fmwork(2))
! ... and further down, the redistribution:
IF (my_kpgrp) THEN
   CALL cp_fm_start_copy_general(fmwork(1), fmlocal, para_env, info(indx, 1))
   CALL cp_fm_start_copy_general(fmwork(2), fmlocal, para_env, info(indx, 2))
ELSE
   CALL cp_fm_start_copy_general(fmwork(1), fmdummy, para_env, info(indx, 1))
   CALL cp_fm_start_copy_general(fmwork(2), fmdummy, para_env, info(indx, 2))
END IF

! Today (after #5950): the same dependencies, explicit as arguments
CALL kp_evp_op_start(effective_ks, ispin, ik, xkp(1:3, ik), cell_to_index, sab_nl, &
                     ks_grid(ispin), use_grid_fft, rmatrix, cmatrix, tmpmat, &
                     fmwork(1), fmwork(2), fmlocal, fmdummy, my_kpgrp, para_env, &
                     info(indx, 1), info(indx, 2))

! After the unification: the caller loop keeps k-points and spins;
! the group walk, the global index, and the receiver branch are all gone
CALL kpoint_operator_start(ctx, ikp, ispin, effective_ks, slot=kpoint_slot_ks)
```

Both forms run the same operator body (plain-or-grid transform, densification, redistribution); they differ only in who ends the transfer and when.
Both take the caller's local k-point, as the channel routine takes `local_kpoint`; the service owns the group walk in both, so no call site can express a group-local loop at all.

The stateless core underneath both forms remains public as a low-level primitive: callers requiring direct buffer control or non-standard memory lifecycles can still invoke the raw launch routine directly.
It is entirely stateless and carries no protocol knowledge of its own: the caller explicitly owns the k-point group walk, the receiver routing, and the finish pairing.
While the service context enforces and automates these collective invariants, the low-level routine leaves enforcement entirely to the caller.
`kpoint_operator_start` is today's `kp_evp_op_start` with its arguments re-sourced from the context (scratch, workspace, routing, bookkeeping); `kpoint_operator_finish` is today's `kp_evp_op_finish_cfm` plus the slot lookup.
The statement sequences, and therefore the collective call orders, are those of #5950, kept identical, so the migration is verifiable.

### Migration mapping

| Today | After | When |
|---|---|---|
| STANDARD loop, `diag_kp_basic`, `diag_kp_smat` | `kpoint_operator_start` / `kpoint_operator_finish` | with the unification |
| Davidson driver + preconditioner (6 channel call sites), the ADIIS blend | `kpoint_operator_get`; the channel routine can initially survive as a thin facade over it | with the unification |
| Mermin exact-Hxc (`qs_scf_distribute_kpoint_operator`) | a thin wrapper over `kpoint_operator_get` | with the unification |
| wavefunction extrapolation (3 copies in `qs_wf_history_methods.F`) | stays as-is: not eigensolver infrastructure, and its overlap handling has extrapolation-specific lifetime and PBC-phase semantics | out of scope |

Out of scope and unchanged: the real-wavefunction assembly paths, the fused transform of the OMP k-point executor, the Γ-point machinery.

The unification is verifiable against #5950.

### Wavefunction extrapolation

Status after the discussion: extrapolation stays as-is and out of scope; the comparison below is kept as the survey record.
Beyond the discussed scope, a repo-wide survey found three copies of this pipeline in `qs_wf_history_methods.F`:
- `wfs_update`, which saves the extrapolation snapshot.
- `wfi_use_prev_wf_kp`, the previous-wavefunction extrapolation.
- `wfi_extrapolate_gext_proj_kp`, the coefficient fitting of GEXT_PROJ(_QTR).

```fortran
! Today: wfs_update (one of the three copies)
CALL rskp_transform(rmatrix=rmat_ft, cmatrix=cmat_ft, rsmat=matrix_s_kp, ...)
CALL dbcsr_desymmetrize(rmat_ft, tmpmat_ft)
CALL copy_dbcsr_to_fm(tmpmat_ft, scf_env%scf_work1(1))
CALL dbcsr_desymmetrize(cmat_ft, tmpmat_ft)
CALL copy_dbcsr_to_fm(tmpmat_ft, scf_env%scf_work1(2))
IF (my_kpgrp) THEN
   CALL cp_fm_start_copy_general(scf_env%scf_work1(1), fmlocal_ft, para_env, info(indx, 1))
   CALL cp_fm_start_copy_general(scf_env%scf_work1(2), fmlocal_ft, para_env, info(indx, 2))
ELSE
   CALL cp_fm_start_copy_general(scf_env%scf_work1(1), fmdummy_ft, para_env, info(indx, 1))
   CALL cp_fm_start_copy_general(scf_env%scf_work1(2), fmdummy_ft, para_env, info(indx, 2))
END IF

! After the unification (draft).
CALL kpoint_operator_get(ctx, ikp, ispin, matrix_s_kp, fm_re=rmat_ft, fm_im=cmat_ft)
```

Construction and teardown, today versus after: each frontend runs its own setup before the first operator is assembled (the grid-versus-plain decision with its fallback chain and memory budget, the workspace sizing, the transfer-state bookkeeping, the destination pair, ~50 lines in `do_general_diag_kp`) versus:

```fortran
CALL kpoint_operator_context_create(ctx, kpoints)
...
CALL kpoint_operator_context_release(ctx)
```

## Open questions

1. **Context payload:** is the minimal set above right?
   The grid-versus-plain strategy (the prepare/extract machinery with its fallback chain, currently in the STANDARD driver) stays with the callers for the initial version, per the discussion; the payload question itself remains open.
2. **Naming:** are the `kpoint_operator_*` names and the module name `qs_kpoint_operators` right?
   They follow the `kpoint_*` family (`kpoint_methods`, `get_kpoint_info`); renaming before the call sites exist costs far less than after.

### Baseline conventions & rollout feedback

A few design choices are kept as baseline conventions rather than open questions:
- **Slot addressing**: Named integer constants follow the standard `input_constants.F` style. Because context internals are private, adding new slots has no external footprint, and operator identity is carried directly by the `rsmat` argument (meaning perturbations and response operators require no specialized slot logic).
- **Migration sequencing**: The proposed rollout prioritizes low-risk call sites first (as mapped in the table above). 

If your ongoing work or active branch would benefit from a different migration order, please let us know so we can coordinate the rollout smoothly.

s8ga:
Implements the unified k-point operator service discussed in #5952. cc @Growl1234 (reviewed the RFC design) and @DCM-Uni-Paderborn (the #5926 distribution helper is retired here).

## Summary of Changes

Replaces `qs_kp_evp_methods.F` with a unified module `qs_kpoint_operators.F` that handles phase transform, densification, and group redistribution:

- **Context & Lifecycle**: Added caller-owned `kpoint_operator_context_type` (`create`/`release`). The context manages its DBCSR/FM scratch lazily. The scratch spans one driver call: the host neighbor lists cannot change within it, and every buffer use asserts this. `kpoint_types.F` remains untouched. The context declares its spin space once at creation (`nspin` = the caller's spin-loop bound).
- **Service APIs**:
  - `kpoint_operator_start` / `_finish`: Two-phase asynchronous pipelining with named slots (`ks`, `s`, `t`) and `idle`/`in_flight` state tracking.
  - `kpoint_operator_get`: Synchronous one-shot assembly with three output flavors — complex FM, real FM pair, and DBCSR pair — through transfer states local to the call. The persistent slot machinery serves only `start`/`finish`.
  - The service keeps the k-point group walk internal. Callers pass only their local k-point index.
  - Row selection is explicit: every call passes `matrix_row`. Spin-free operators (S, T) pass the `kpoint_spin_free` sentinel. The leading dimension of e.g. `matrix_s_kp` counts derivative components, not spin components.
  - The synchronous and asynchronous paths share one single-group transfer primitive. The primitive stays private, as does the CFM merge helper. `kpoint_operator_cfm_to_mo` remains the public MO helper.
  - The calls are not thread safe: the caller must serialize them, because their collective communication tolerates no concurrent service call. Separate contexts only keep the scratch from aliasing.

## Migrations

- **`do_general_diag_kp`**: The complex path switches to `start`/`finish` slots. The ADIIS blend reuses the idle `ks` slot. The migration removes manual group loops, `info` columns, and trailing cleanup loops.
- **`diag_kp_basic` / `diag_kp_smat`**: Switched complex path to `start`/`finish`.
- **Davidson & Preconditioners**: Migrated S/H/T assembly to `kpoint_operator_get`. The DBCSR-pair output absorbs the caller-side `copy_fm_to_dbcsr_bc` detour. Removed private DBCSR scratch trio from the Davidson driver.
- **Mermin Exact-Hxc (`qs_scf_loop_utils.F`)**: One context reused across the response modes.
- **Cleanup**: Deleted `qs_scf_distribute_kpoint_operator` and `distribute_kpoint_operator_channel`.

## Testing

- **Regressions**: 257/257 passed on 2 ranks and 4 ranks.

## Out of Scope (Unchanged)

- Real-wavefunction paths (`use_real_wfn`).
- Fused OMP transform (`kp_build_cfm`) and OT constructor (`qs_scf_get_ot_kpoint_operator`).
- Grid-transform cache lifecycle (remains caller-managed).

Close #5952


Growl1234:
Most likely would need a cleaner work. I'm busy these days and will carefully have a look tonight or tomorrow.

s8ga:
> Most likely would need a cleaner work. I'm busy these days and will carefully have a look tonight or tomorrow.

Thanks! Take your time, and let me know whatever needs to be adjusted or cleaned up.

Growl1234 (changes requested):
The overall direction looks good to me, but I think the current implementation is somewhat overbuilt for the abstraction it provides. A few points that may be worth simplifying before merging:

* **The synchronous `get()` path should probably not go through the asynchronous slot/state machinery.**
  A one-shot caller has no pipeline to key, yet `get()` still scans slots, updates `status`, and relies on the `info(kpoint × group × spin × slot × part)` storage. For large k-point sets this also makes the per-context bookkeeping unnecessarily large. It seems cleaner for synchronous `get()` to use local `copy_info_type` objects and reserve the persistent slot machinery for `start()/finish()`.

* **`get()` currently duplicates part of the transfer pipeline.**
  The CFM branch reuses `start()/finish()`, while the FM/DBCSR branches repeat the group walk, densification, copy start/finish and cleanup. A private single-group launch/transfer primitive shared by both the synchronous and asynchronous paths could reduce quite a bit of duplication and better match the goal of keeping the group-walk logic in one place.

* **I would make `matrix_row` explicit instead of defaulting it to `ispin`.**
  The two quantities have different semantics, and the recent fixes around spin-independent S/T operators show that an omitted `matrix_row` can select another valid row and therefore escape bounds checking. Requiring it at the call site seems safer than keeping an implicit coupling between operator storage and spin bookkeeping.

* **The thread-safety contract should probably be tightened.**
  “Give each thread its own context” is not sufficient for a service whose calls are collective over the environment-wide communicator. Separate contexts avoid scratch aliasing, but do not by themselves make concurrent MPI/BLACS/ScaLAPACK communication from different OpenMP threads safe. For now I would simply state that calls must be serialized by the caller.

* **The public `raw_*` escape hatch may be premature.**
  None of the migrated callers currently needs it. Keeping these routines private until there is an actual consumer would make the public API smaller and preserve the invariants that the service is intended to enforce.

* **Some of the context invalidation/lifetime machinery also looks more general than the current callers require.**
  In particular, the `sab_nl` association tracking and rebuild path may be unnecessary if the context lifetime is intentionally restricted to a driver invocation where the underlying matrix structure does not change. Likewise, Mermin could probably create the context outside the response-mode loop and reuse its scratch instead of repeatedly creating and releasing it.

Overall, I think the service abstraction itself is useful and the migrated callers are cleaner, but it would be good if the consolidation also resulted in a smaller and more explicit implementation, rather than moving the old complexity into a larger stateful context.

s8ga:
Thanks for the careful review! All six points are addressed in the rework round.

**1. `get()` off the slot machinery.** `get` now runs through `copy_info_type` objects local to the call: no slot argument, no `status`/`info` access. The slot bookkeeping itself is allocated at the first `start`, so a get-only context (Mermin, Davidson preconditioners) never carries it.

**2. Shared single-group primitive.** One private `kpoint_operator_launch` (densify + start of both transfers, dual targets) now serves `start` and all three `get` flavors; the group walk, densify, and copy start exist exactly once. `raw_start` was absorbed into `launch` rather than merely privatized: the shared primitive you asked for *is* the stateless core, and keeping `raw_start` beside it would reintroduce the duplication. Exposing the hatch again when a real consumer appears is a one-line PUBLIC change.

**3. `matrix_row` explicit.** Non-optional in `start` and `get`; every call site passes it, spin-free S/T pass the `kpoint_spin_free` sentinel. (The docs now state the real geometry: the first dimension of S/T counts derivative components, the images live in the second.)

**4. Thread safety.** The header now states the caller must serialize the calls — their collective communication tolerates no concurrent service call — and that separate contexts only keep the scratch from aliasing.

**5. `raw_*` private.** `raw_finish_cfm` is module-private; `cfm_to_mo` stays public with real consumers. The public API is down to six routines, all with callers.

**6. Lifetime machinery.** The `sab_nl` rebuild path is gone; in its place, `ensure` stores the neighbor-list identity at trio creation and asserts it at every buffer use, so a context that outlives its driver call aborts instead of healing or silently going stale. Mermin now creates the context once, outside the response-mode loop, and reuses it.

Growl1234 (approved):
LGTM. I have a in-progress work to deal with MPI performance optimization for k-point calculations, and I'll adapt to this. Maybe I would also bring a follow-up cleanup if needed.

PR #5976, opened 2026-09-06
Refactor non-local vdW code

Growl1234:
- Remove large temporary real- and reciprocal-space work arrays and stream the non-local fields through the existing FFT path.
- Reorganize spline, potential, and stress contractions to avoid redundant per-channel work.
- Improve reciprocal-space kernel access and SIMD/vectorization opportunities.
- Parallelize remaining real-space `q0` loops with OpenMP.
- Reduce redundant FFTs in the spin-density and potential-assembly paths.
- Reuse spline metadata and avoid repeated interpolation setup.
- Handle the analytic zero-gradient limit in the rVV10 potential expression.

### DIFF
--- a/src/qs_dispersion_nonloc.F
+++ b/src/qs_dispersion_nonloc.F
@@ -610,22 +610,24 @@ CONTAINS
                vdW_xc_energy = vdW_xc_energy + g_multiplier*REAL(uu*CONJG(theta(q2_i)), KIND=dp)
             END IF
             IF (.NOT. energy_only) thetas_g(q2_i)%array(ig) = uu
-
-            IF (use_virial .AND. ig >= grid%first_gne0) THEN
-               DO q1_i = 1, nqs
-                  gm = 0.5_dp*g_multiplier*grid%vol*dkernel_of_dk(q1_i, q2_i)*REAL(theta(q1_i)*CONJG(theta(q2_i)), KIND=dp)
-                  IF (nl_type == vdw_nl_RVV10) THEN
-                     gm = 0.5_dp*gm
-                  END IF
-                  DO l = 1, 3
-                     DO m = 1, l
-                        virial_thread(l, m) = virial_thread(l, m) - gm*(grid%g(l, ig)*grid%g(m, ig))/g
-                     END DO
-                  END DO
-               END DO
-            END IF
-
          END DO
+
+         IF (use_virial .AND. ig >= grid%first_gne0) THEN
+            ! Reduce over all channel pairs before assembling the stress tensor.
+            gm = 0.0_dp
+            DO q2_i = 1, nqs
+               DO q1_i = 1, nqs
+                  gm = gm + dkernel_of_dk(q1_i, q2_i)*REAL(theta(q1_i)*CONJG(theta(q2_i)), KIND=dp)
+               END DO
+            END DO
+            gm = 0.5_dp*g_multiplier*grid%vol*gm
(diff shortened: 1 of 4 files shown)


oschuett:
I take it these are performance optimizations? To track and protect the gains, maybe consider adding a benchmark:
https://github.com/cp2k/cp2k/blob/2d1258df02df208f79f81c3b4dc59e6880375f1f/tools/docker/scripts/test_performance.sh#L50

Growl1234:
Yes, and I can immediately pick one example for benchmarking. That said, I want to add it seperatedly, together with k-points benchmark which I have not decided how I should organize.

src/qs_dispersion_nonloc.F
```diff
@@ -609,88 +483,108 @@ SUBROUTINE vdW_energy(thetas_g, dispersion_env, vdW_xc_energy, energy_only, viri
...
-         u_vdW(:, :) = z_zero
+      ! Pack the symmetric radial tables once per evaluation. The interpolation then
+      ! reads adjacent channel pairs instead of striding over the radial dimension.
+      ! Copy only the radial prefix needed by the local G points, including its upper endpoint.
+      nr_used = 0
+      IF (grid%ngpts_cut_local > 0) THEN
+         nr_used = MIN(dispersion_env%nr_points, &
+                       INT(SQRT(MAXVAL(grid%gsq(1:grid%ngpts_cut_local)))/dispersion_env%dk) + 1)
       END IF
+      ALLOCATE (kernel_table(nqs*(nqs + 1)/2, 0:nr_used, 2))
```
RTChou67: As `kernel_table` is immutable, would it be possible to store it and reuse? More performance, less code

Growl1234: Good catch!

fstein93:
I am curious how much faster your modification is than the old one. This would definitely benefit a variety of applications.

Growl1234:
In principle it avoids some unnecessary burden, but my practical test doesn't show significant benefits. That said, since it also reduces the amount of code, I would still consider it a worthwhile refactoring.

Growl1234:
I did some finer-grained profiling on a PBE+rVV10 calculation to understand why the SCF wall time does not change much after these optimizations.

For this case, one non-local evaluation takes about **3.63 s**, of which about **3.02 s (83%)** is spent inside the existing PW transforms:

* theta forward transforms: ~1.19 s
* theta inverse transforms: ~1.21 s
* density/gradient transforms: ~0.25 s
* final-potential forward transforms: ~0.25 s
* final-potential inverse transforms: ~0.12 s

The remaining non-FFT work is only ~0.61 s in total; `vdW_energy()` itself is ~0.14 s. There is also exactly one non-local evaluation per SCF step, so there does not seem to be redundant re-evaluation to eliminate.

This explains why optimizing the contractions, spline handling, temporary storage, etc. has little visible impact on the overall SCF time: the dominant cost is the 40 theta PW transforms (20 forward + 20 inverse), which are unchanged by this PR.

So, the current changes are still useful for reducing memory traffic, temporary storage and redundant work, but further local arithmetic optimizations are unlikely to give a large wall-time improvement. A substantial speedup would probably require reducing or batching the PW transforms themselves. Since the FFT backend is currently being reworked, I would avoid touching that infrastructure here and keep this PR limited to the non-local vdW implementation.

jpoto (approved):
Looks fine to me

PR #5978, opened 2026-09-06
XC: Avoid ifx ASSOCIATE miscompilation in LibXC workers

### LINKED ISSUE #5948: Exchange-correlation energy is exactly zero for r2SCAN when CP2K is built with ifx
In CP2K 2026.2 (Development Version), I obtain an exactly zero exchange-correlation energy for an r2SCAN + non-local rVV10 calculation. The calculation finishes normally. However, the energy decomposition shows:

>   Core Hamiltonian energy:                 15.32628374731652
>   Hartree energy:                          22.12563649414903
>   Exchange-correlation energy:              0.00000000000000
>   Dispersion energy:                        0.02967483872820
>   Total energy:                           -15.85843595405708

The XC contribution is exactly zero to the printed precision. The full input and output files are attached.

[dd_m0p15.zip](https://github.com/user-attachments/files/31767453/dd_m0p15.zip)

DCM-Uni-Paderborn:
## Summary

Work around an ifx 2025.2.1 miscompilation of `ASSOCIATE` bindings around the orphaned OpenMP worksharing loops in the two LibXC evaluators.

- Replace the worker alias with an explicit block-local pointer and copy the four read-only scalar settings into local variables.
- Mark the worker-set dummy arguments `TARGET` for the pointer association.
- Keep the existing block scope, batching, screening, formulas and OpenMP scheduling unchanged. No regtest inputs, references, tolerances, dependencies or compiler optimization levels are changed.

## Diagnosis

Unmodified master `9c4b5efbe0` reproduces the 33 meta-GGA runtime failures reported by the Intel ifx dashboard testers. GDB shows invalid `ASSOCIATE` bindings despite valid worker storage. Reducing the module to `-O0` does not fix it; replacing only the worker alias exposes the same problem with a scalar alias.

A standalone Fortran reproducer, independent of CP2K, LibXC and MKL, also fails with ifx: the original bindings segfault at `-O0` and leave the output zero at `-O3`. Explicit pointer/scalar bindings pass at both optimization levels with 1, 2 and 4 threads. GCC passes both variants. A further 32-line reproducer isolates the logical-component alias alone.

## Validation

Normal Release builds on Terok with ifx 2025.2.1, MKL 2025.2 and LibXC 7.1.2:

| Build | MPI ranks x OpenMP threads | Eleven selected regtest directories |
| --- | --- | --- |
| Original master | 1 x 2 | 33 runtime failures |
| Fixed ssmp | 1 x 1, 1 x 2, 1 x 4 | 104/104 checks pass in each configuration |
| Fixed psmp / Intel MPI | 2 x 1, 2 x 2 | 104/104 checks pass in each configuration |

These directories contain all 33 reported failures, including spin-polarized and unpolarized meta-GGAs, Laplacian-dependent functionals, stress and DCDR. The existing dashboard stack settings were used. This is targeted validation, not a full regression run.

The unchanged input attached to #5948 also passes in both ssmp and psmp. Its XC energy is `-6.95854911100701 Ha`, matching the GCC reference; the SCF total agrees within `2e-14 Ha`, and all twelve printed atomic force components agree.

Native macOS GCC 16/OpenMPI, 2 ranks x 2 threads: 87/91 checks pass in the nine available directories. Four LRI setup aborts are also reproducible with the pre-existing unpatched native binary; enabling Libint does not remove them. They are separate from this fix and are not suppressed.

`make pretty` and `git diff --check` pass.

Addresses #5948. This does not claim to resolve the older memory-growth, Libint or performance problems discussed in #4795.

### DIFF
--- a/src/xc/xc_libxc.F
+++ b/src/xc/xc_libxc.F
@@ -2094,7 +2094,7 @@ CONTAINS
    INTEGER, INTENT(in)                                :: grad_deriv, npoints
    CHARACTER(LEN=default_string_length), INTENT(IN)   :: func_name
    REAL(KIND=dp), INTENT(in)                          :: sc
-   TYPE(libxc_worker_set_type), INTENT(INOUT)         :: workers
+   TYPE(libxc_worker_set_type), INTENT(INOUT), TARGET :: workers
 
    INTEGER                                            :: bsize, family, i, i0, ib, ii, ithread, &
                                                          nb, nblocks, nthreads
@@ -2134,9 +2134,18 @@ CONTAINS
    ! to the functional object in xc_libxc_wrap_set_thresholds, and LibXC leaves
    ! the outputs of the points it screens at zero, so accumulating them is a
    ! no-op.
-   ASSOCIATE (w => workers%worker(ithread + 1), has_laplace => workers%has_laplace, &
-              no_exc => workers%no_exc, eps_rho => workers%epsilon_rho, &
-              eps_tau => workers%epsilon_tau)
+   BLOCK
+      TYPE(libxc_worker_type), POINTER :: w
+      LOGICAL :: has_laplace, no_exc
+      REAL(KIND=dp) :: eps_rho, eps_tau
+
+      ! Avoid ASSOCIATE aliases across the orphaned OpenMP worksharing loop:
+      ! ifx can lose their bindings even at -O0.
+      w => workers%worker(ithread + 1)
+      has_laplace = workers%has_laplace
+      no_exc = workers%no_exc
+      eps_rho = workers%epsilon_rho
+      eps_tau = workers%epsilon_tau
 !$OMP    DO
       DO ib = 1, nblocks
          i0 = (ib - 1)*bsize
@@ -2380,7 +2389,7 @@ CONTAINS
          END IF
       END DO
 !$OMP    END DO
-   END ASSOCIATE
+   END BLOCK
 
 END SUBROUTINE libxc_spin_unpolarized_calc
 #endif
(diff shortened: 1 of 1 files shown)


DCM-Uni-Paderborn:
Here is the reduced compiler reproducer. With ifx 2025.2.1, `ifx -O3 -qopenmp associate-minimal.f90 -o reproduce` followed by `OMP_NUM_THREADS=2 ./reproduce` prints an all-zero range and exits with `Incorrect ASSOCIATE binding`. GCC 16 (`gfortran -O3 -fopenmp`) prints the expected range `3` to `3` and exits successfully.

It requires no CP2K, LibXC, MKL or MPI. The larger pointer/scalar reproducer described above additionally checks the explicit-binding workaround at O0/O3 and 1/2/4 threads.

<details>
<summary>32-line Fortran reproducer</summary>

```fortran
module example
   use omp_lib
   implicit none
   type settings_type
      logical :: enabled = .true.
   end type
contains
   subroutine evaluate(settings, values)
      type(settings_type), intent(in) :: settings
      real, intent(inout) :: values(:)
      integer :: i
      associate(enabled => settings%enabled)
!$omp do
         do i = 1, size(values)
            if (enabled) values(i) = 3.0
         end do
!$omp end do
      end associate
   end subroutine
end module

program reproduce
   use example
   implicit none
   type(settings_type) :: settings
   real :: values(16) = 0.0
!$omp parallel default(none) shared(settings, values)
   call evaluate(settings, values)
!$omp end parallel
   print *, minval(values), maxval(values)
   if (any(values /= 3.0)) error stop 'Incorrect ASSOCIATE binding'
end program
```

</details>

fstein93:
The reproducer should be related to https://github.com/cp2k/cp2k/issues/2668#issuecomment-1464056199 (`ASSOCIATE` inside of an OpenMP region with variables of `PUBLIC` visibility).

oschuett:
> The reproducer should be related to https://github.com/cp2k/cp2k/issues/2668#issuecomment-1464056199 (ASSOCIATE inside of an OpenMP region with variables of PUBLIC visibility).

Yes, we even have a [convention check](https://www.cp2k.org/conv#c008) for that. Maybe it needs refining?

fstein93:
It needed to cover `ASSOCIATE` within a function called from an OpenMP-parallel region. This seems possible from the same file but is probably less trivial in general.

Issue #5863, opened 2026-08-26
Long-time Feature request: the correction for the constrained molecular dynamics


garrettwx:
Dear cp2k developers,

I have noticed some reported works but have fallen into the inappropriate descriptions about the use of constrained AIMD (cAIMD) simulation for the achievement of PMF.

As the latest revised description for the cAIMD, the blue moon ensemble average connects constrained and unconstrained molecular dynamics beyond the Lagrange multiplier associated with the parameter used in the SHAKE algorithm rather than the pure Lagrange multiplier.

Here, these works directly interpret the Lagrange multiplier (λ) from SHAKE algorithm as the ensemble-averaged constrained force, while absolutely neglects the correction from Z and G. 

In these works, this enhanced sampling is used to obtain the activation energy, but I think sometimes this is very dangerous to crudely neglect the correction from Z and G beyond λ.

The CP2K's official example is OK (https://www.cp2k.org/exercises:2015_ethz_mmm:nacl_free_energy), because for the example of NaCl dissociation, the CV is only the simple distance between two atoms, and this condition can neglect the contribution from Z and G. So, we can only need the free energy gradient term (∂A/∂ξ) = <-λ>. 

But for other CV types, it is not always the case: https://manual.cp2k.org/trunk/methods/sampling/constrained_dynamics.html.

For example, if the CV=R1-R2 within the atoms A--B---C, R1=|A--B|, R2=|B---C|, the correct free energy gradient should be <Z^(-0.5)*(-λ)>/<Z^(-0.5)> at certain CV, rather than the original λ. 

The magnitude of error introduced by ignoring Z can substantially alter derived activation energies and thus predicted chemical reaction rates:
By TOF = A*exp(-Ea/RT),  the typical DFT error in adsorption energy is 0.15 eV; a shift in Ea of this magnitude gives a 300× change in the TOF at room temperature! 

I am not sure if there are some developers are adding this enhancement for the gradient energy correction here, but some works have made these inappropriate treatments for the free energy calculation.

he-zilong:
[...]

PR #5981, opened 2026-09-07
Add single-constraint Blue Moon postprocessing

DCM-Uni-Paderborn:
Refs #5863; related discussion: #5509.

## Summary

Add a standalone NumPy postprocessor for the common single-constraint Blue Moon cases discussed in #5863. This follows the suggestion to evaluate the correction during postprocessing; the MD engine and its multiplier output are unchanged.

The tool combines CP2K's SHAKE multiplier with the scalar mass metric and its derivative:

```text
g = grad(xi), H = Hessian(xi), v = M^-1 g
Z = g . v
G = v . H . v / Z^2
w = Z^(-1/2)
dA/dxi = sum[w * (-lambda_SHAKE + kB*T*G)] / sum[w]
```

- Support distances, angles, torsions, signed point-plane distances, point-to-bond-center distances, and linear combinations of these primitives. This covers the eight common forms listed in the issue, including shared-atom distance differences and longer distance combinations.
- Evaluate Cartesian gradients and Hessians with second-order forward differentiation. Shared atoms are differentiated consistently, including atoms defining planes and bond centers; no finite-difference step size is required.
- Require an explicit CV definition, the actual atomic masses, a fixed cell, temperature, target and tolerance. Coordinates are read from CP2K XYZ trajectories in Angstrom; multipliers retain CP2K's internal units and sign convention.
- Validate SHAKE/RATTLE pairing and trajectory alignment, counting only SHAKE as a configurational-force sample. Since multiplier logs do not contain step labels, the first recorded MD step must be specified explicitly. Reject count mismatches, gaps, repeated steps, multiple multipliers and singular coordinate geometries.
- Add usage documentation, reproducible CP2K example inputs, unit tests, and execution of both the tests and `mypy --strict tools/blue_moon/` in the miscellaneous CI job. Runtime-validated JSON is converted to a typed coordinate tree before numerical evaluation; strict checking includes the test module.

## Scope

This implementation is for **one fixed collective constraint in fixed-cell equilibrium sampling**. It does not implement the general multiple-constraint estimator, arbitrary nonlinear `COMBINE_COLVAR` expressions, additional fixed atoms or rigid molecules, moving targets, variable cells, PIMD, or RESPA.

The original CP2K input is not parsed: users must supply the matching CV definition and verify the simulation prerequisites. Unsupported conditions that cannot be inferred from the trajectory and multiplier log are documented explicitly. The tool reports the weighted gradient and per-frame quantities; it does not integrate windows or claim statistical convergence.

## Metric term requiring particular review

The documentation currently assumes `G = 0` for `xi = rij - rkj`. Applying the general Hessian expression above to this coordinate, with `c = rho_ij . rho_kj`, gives:

```text
Z = 1/mi + 1/mk + 2*(1-c)/mj
G = (1-c*c) * (1/rij - 1/rkj) / (mj*mj*Z*Z)
```

Thus `G` vanishes for equal distances or collinear bonds, but not for a general configuration. The documentation is updated accordingly.

For positions `[(2,0,0),(0,0,0),(1,3,0)]` in bohr and masses `[12,1,16]` in amu, the regression gives `Z = 1.5133778012996573` and `G = 0.07221504489510591` per bohr. The Hessian contraction, the closed-form result, and an independent directional finite difference of the analytic metric agree.

I would particularly appreciate an independent review of this specialization: it follows the general expression in Eqs. (6)-(8) of [Komeiji (2007)](https://doi.org/10.1273/cbij.7.12), but differs from the `G = 0` statement in Eq. (35). The implementation retains both correction terms.

Following the discussion below, the documentation also gives the equivalent form `G = (1/2) D_xi ln Z`, with `D_xi = (M^-1 grad(xi)/Z) . grad`. This is a local mass-weighted normal derivative, not a time derivative along the constrained trajectory. The implementation retains its second-order forward differentiation; additional tests compare it with normal-direction finite differences of `ln Z` and verify that `Z` can vary even at fixed `xi`.

## Placement

The initial implementation is self-contained under `tools/blue_moon`. Following the suggestion in the issue, feedback on whether it should ultimately live in `cp2k-output-tools` would also be welcome. The documentation and the mathematical regression cases remain useful independently of that packaging decision.

## Validation

- All **23 unit tests pass**, covering analytic metrics, gradient/Hessian finite differences, the normal derivative of `ln Z`, shared atoms, mass and coordinate scaling, rotational invariance, triclinic images, torsion branches, the weighted estimator, nested coordinate trees, and input validation.
- `mypy --strict tools/blue_moon/` passes for all three Python files with the CI-pinned mypy 2.1.0 and NumPy 2.4.4, without ignored type errors. The CLI example output remains byte-for-byte identical to the initial PR revision.
- Two runs with the locally installed **CP2K 2026.1** verified the SHAKE/XYZ format and step alignment, and the values/signs of the six coordinate forms against `METADYN/COLVAR` output. The example inputs are included. These are smoke/reference checks, not equilibrated free-energy benchmarks.
- A recorded 32-step **CP2K 2026.1 (5e54ba2)** trajectory exercises a fixed `COMBINE_COLVAR` distance difference with explicit masses `[12,1,16]`, varying Z and nonzero G. Tests compare native DISTANCE/ANGLE/DISTANCE_FUNCTION/COMBINE_COLVAR values and independently calculated metrics, then exercise the real CLI, discard handling, CSV output and JSON reduction. An initial-time inertial-force check verifies the SHAKE multiplier sign and normalization; additional runs at 0.125 and 0.0625 fs approach the continuous-time reference. This short NVE run is an I/O/estimator reference, **not an equilibrated free-energy benchmark**.
- A separate deterministic **canonical-ensemble reference** uses a confined three-particle harmonic model. The reference mean force is obtained by direct integration of its unconstrained configurational partition density followed by a finite difference of its logarithm, independently of the Blue Moon formula. Constrained configurations and analytically velocity-averaged multipliers are integrated separately and passed through the postprocessor. Both tested windows and both mass sets recover the reference; omitting either G or Z reweighting fails. This validates the thermodynamic estimator but does **not** establish CP2K long-time sampling or thermostat convergence, and does not replace independent scientific review. The derivation, reproduction instructions and limitations are in the README.
- Precommit formatting and file checks passed for the changed files. The local portable-filename check was applied to repository-relative names because the checkout path contains spaces; no repository checker was modified.
- `git diff --check` passed.

The full CP2K build/regression suite has not been rerun locally; no Fortran or simulation-side code is changed. This PR addresses the stated common single-constraint cases; the remaining scope should be discussed separately in the issue.


Growl1234:
cc @garrettwx

garrettwx:
Yes, looks very nice for post-process

The COMBINED_COLVAR indeed cannot lead the general formula for correction. But for the limited support for these commonly defined CV types, the realize in both of cp2k-output-tools or in the inner code, I think this is not very difficult.

garrettwx:
btw, I want to ask, if this procedure can be further simplified to calculate Z and G, 

we can try to not use the hessian, but define the dZ/dξ along the trajectory, and then

G can be simplified into G = 0.5 * dlnZ / dξ for the single CV constraint, so we only need derive the Z, and G is the accompany
 
I think this definition can be easier realized for the general measurement of G. 

As for number of constraint CV > 1, I think meta-dynamics or other advanced sampling methods are more appropriate to use.

tools/docker/scripts/test_misc.sh
```diff
@@ -31,6 +31,7 @@ unzip -q -d ./tools/minimax_tools/1_xData 1_xData.zip
 run_test ./tools/precommit/format_fortran_test.py
 run_test ./tools/minimax_tools/minimax_to_fortran_source.py --check
 run_test ./tools/docker/generate_dockerfiles.py --check
+run_test python3 -m unittest discover -s ./tools/blue_moon -p 'test_*.py'
```
oschuett, line 34: Maybe consider to also run `mypy --strict` on it (see below).

DCM-Uni-Paderborn: Added `mypy --strict ./tools/blue_moon/` to the miscellaneous CI job, including the test module. The numerical arrays and interfaces are now typed, and validated JSON is converted into a typed coordinate tree before evaluation. Runtime input checks are retained.

Strict checking passes with the CI-pinned mypy 2.1.0 and NumPy 2.4.4, without ignored type errors. All 21 unit tests pass as well.

DCM-Uni-Paderborn:
@garrettwx Thanks for the suggestion. Yes, this is an equivalent way to write the single-constraint correction, provided the derivative direction is specified. With `g = grad(xi)`:

```text
b = M^-1 g / Z
D_xi = b . grad
D_xi xi = 1
G = (1/2) D_xi ln Z = b . grad(Z) / (2*Z)
```

Thus a full Hessian matrix need not be stored: a Hessian-vector product or a directional derivative of the metric is sufficient. This is also consistent with the single-constraint specialization of the [VASP expression](https://vasp.at/wiki/Blue_moon_ensemble).

The important distinction is that this is the **local mass-weighted normal derivative at each trajectory frame**, not a derivative obtained from successive frames of the constrained trajectory. In a fixed window, `xi` is constant, whereas `Z` can still change with other degrees of freedom. For example, changing the bond angle at fixed `rij - rkj` changes `Z`. A ratio of frame-to-frame changes `Delta ln Z / Delta xi` therefore cannot supply this derivative; neither can constraint-tolerance noise in `xi`.

I have added this equivalent form and the distinction to the manual and README. New tests compare the current Hessian contraction against `(ln Z(r+h*b) - ln Z(r-h*b))/(4*h)` at several step sizes for all supported primitive types, combinations and a rescaled coordinate, and verify `D_xi xi = 1`. A separate test demonstrates different metrics at the same constrained coordinate.

For this PR I have kept the tested second-order forward-differentiation implementation. A Hessian-free implementation would be an equivalent optimization, but would still need derivatives of the coordinate gradients or gradient evaluations at displaced geometries. The restriction to the supported primitives and linear combinations is an implementation boundary of this tool, not a restriction of the general scalar identity. The scope remains one fixed constraint, and the possible placement in `cp2k-output-tools` remains open for discussion.

garrettwx:
>the possible placement in `cp2k-output-tools` remains open for discussion.

I tend to keep this implementation inside the native CP2K kernel rather than moving it to cp2k‑output‑tools.

1. The underlying blue‑moon single‑constraint formalism is theoretically well‑established and closed. We already support multiple primitive CV constraints and their linear combinations in‑code, so all required gradient/metric information is readily available on‑the‑fly during MD trajectories. And concerning the future implementation of other types of CV, this basis can be further integrated for the developers. 
2. Implementing inside CP2K enables “on‑the‑fly free‑energy correction” directly in constrained MD runs. If moved to cp2k‑output‑tools, users would need to dump full trajectories, reconstruct constraint‑metric tensors in post‑processing, which introduces potential consistency risks between CP2K‑internal constraint logic and external tool re‑implementation.
3. Post‑processing tools are good for analyzing finished trajectories, but this correction is tightly coupled to CP2K’s internal constraint solver, metric tensor Z computation, and per‑frame local directional derivative logic. Re‑duplicating the constraint‑CV evaluation logic in an external tool creates maintenance burden and version‑mismatch pitfalls.
4. Of course, we could still expose helper subroutines as a public API, so that `cp2k‑output‑tools` could optionally reuse these routines if needed, without forcing the whole feature out‑of‑tree.

DCM-Uni-Paderborn:
@garrettwx Thanks for explaining your concerns. I agree that consistency with CP2K’s CV definitions and avoiding duplicated maintenance are important.

For this PR, I would nevertheless keep the postprocessing approach, following @juerghutter’s recommendation in #5863. This keeps the correction adaptable without extending the derivative machinery and supported-feature combinations in the MD engine.

Rather than expanding this PR into an in-kernel implementation, I suggest addressing the consistency concerns through additional CP2K reference cases and an end-to-end validation of the corrected mean-force estimator. Independent review of the nonzero G term for general shared-atom distance differences also remains important.

Whether the postprocessor ultimately belongs under tools/blue_moon or in cp2k-output-tools can be decided separately.

DCM-Uni-Paderborn:
Rebased onto current master (`dd94896b48`, including #6017) and investigated both failing checks.

The [pdbg report](https://storage.googleapis.com/cp2k-ci/run-cp2k-pdbg-e7eb9b4a_report.txt) shows 64 concurrent test workers with two MPI ranks each, although the CI worker has a 32-CPU cpuset. `do_regtest.py` used the host-wide `os.cpu_count()` for its default, ignoring that affinity restriction. The failures are timeouts and a missing restart file after an earlier timeout, rather than wrong numerical results.

Commit `acaaca8536` makes the default respect `sched_getaffinity(0)`, with a portable fallback and the explicit `--maxtasks` override preserved. On that worker this gives 16 test workers rather than 64. Six new tests cover the restricted/noncontiguous CPU sets, fallback paths and CLI override; the restricted-affinity case fails with the old behavior. They are included in Misc CI. All six tests, all 23 Blue Moon tests, strict typing and precommit checks pass locally.

The [GCC 10 report](https://storage.googleapis.com/cp2k-ci/run-cp2k-spack-psmp-gcc10-2d108c6d_report.txt) has three CELL_OPT reference deviations and was generated before the BFGS changes in #6017. I am requesting a fresh run on the updated base to check whether those persist. No reference tolerances or timeouts were loosened, and the Blue Moon implementation is unchanged.

/cp2kci spack-psmp-gcc10

DCM-Uni-Paderborn:
Dear @fstein93, @mkrack and @garrettwx,

Could you please revisit this single-constraint Blue Moon postprocessing PR when convenient? The current CI checks are green, but a final review is still outstanding. In particular, it would be helpful to confirm the supported single-constraint scope and whether the present placement is acceptable, given the earlier discussion of native versus external postprocessing. Please flag any remaining changes you would like before merge.

Many thanks!
Greetings,
Thomas

garrettwx:
I think this post-process PR has included the typical types for blue moon. For users, we can choose the post-process via the post-tool. 
The latter new version, we then can think to allow for the native format.

tests/do_regtest.py
```diff
@@ -775,7 +775,15 @@ def run_with_capture_stdout(cmd: str) -> bytes:
...
-    return int(os.getenv("PYTHON_CPU_COUNT") or str(os.cpu_count()))
+    override = os.getenv("PYTHON_CPU_COUNT")
+    if override:
+        return int(override)
+    # Respect CPU affinity, e.g. the cpuset assigned to a CI container.
+    # os.cpu_count() reports all host CPUs and can oversubscribe the worker.
+    try:
+        return len(getattr(os, "sched_getaffinity")(0))
+    except (AttributeError, OSError):
+        return os.cpu_count() or 1
```
oschuett: What problem is this trying to fix?

Btw, I really dislike these drive-by fixes. Is there maybe a way to tell the agent that changes to [load bearing files](https://github.com/cp2k/cp2k/blob/d502fc8445567085180e1ffdff8b207f9f62af45/CODEOWNERS#L25-L37) always require a dedicated PR?

DCM-Uni-Paderborn: Dear Ole,

You are right; this infrastructure change should not have been bundled with the Blue Moon implementation. It was introduced while investigating CI timeouts: os.cpu_count() can report all host CPUs rather than the CPU set available to the worker, leading to excessive test concurrency.

I have removed the runner change, its unit tests and its two Misc CI entries from this PR in 7e3c4424a2. The Blue Moon implementation, its tests and its strict typing check are unchanged. All 23 Blue Moon tests, mypy --strict and make pretty pass after the removal.

The CPU-affinity fix is now prepared separately for a dedicated infrastructure PR. I will keep unrelated changes to the load-bearing files in dedicated PRs rather than bundling them into feature work.

Thank you for pointing this out.

Greetings,
Thomas

DCM-Uni-Paderborn:
Dear @garrettwx,

Thank you for taking another look. Agreed: let us keep this PR focused on the supported single-constraint postprocessing workflow and consider native integration separately in a later version. I have also removed the unrelated regression-runner change following the review by Ole, so it no longer expands the scope of this PR.

Greetings,
Thomas

PR #5983, opened 2026-09-07
Add optional Wannier90 v4 library integration for k-point localization

DCM-Uni-Paderborn:
## Summary

Add an optional in-process Wannier90 v4 interface to CP2K's k-point exporter.
The default remains file export without a Wannier90 dependency. Enabling
`DFT/PRINT/WANNIER90/LIBRARY` runs disentanglement and localization directly
on CP2K's exported orbitals and overlaps, without spawning an executable.

## Main Changes

- Add `CP2K_USE_WANNIER90`, compiler/MPI interface checks, toolchain and Spack
  installation, dependency reporting, and build documentation. The installers
  use Wannier90 4.0.2 and keep serial and MPI installations separate.
- Pass full-mesh eigenvalues and distributed neighbour overlap blocks to the
  public library API. Use Wannier90's neighbour connectivity and support ranks
  that own no k-points.
- Construct physical AO projections as `C(k)^H S(k)`, select one trial set
  over the mesh with pivoted QR, and detect or repair mesh-wide rank loss.
  `AO_SCDM` is an AO-based, SCDM-inspired construction, not real-space SCDM.
- Add radial s/p hybrid trial banks and `INITIAL_PROJECTIONS AUTO`. AUTO
  independently minimizes physical candidate families with different optimizer
  reset intervals, then tests deterministic unitary mixtures of the best
  original trials. Only finite, converged results can be selected.
- Support disentanglement, outer/frozen energy windows, and consistent band
  exclusions. Preserve both original-MO and packed-window row mappings.
- Export the selected state's native Hamiltonian and optimized U/U_dis/V
  matrices. Optional input files reproduce the winning trial in an external
  Wannier90 run. Losing trials cannot overwrite the selected optimized state.
- Handle collinear alpha and beta channels independently, with separate band
  counts, projections, spectra and output prefixes.
- Extend regression coverage for GPW, all-electron GAPW, Gamma, Monkhorst-Pack,
  MacDonald and complete equally weighted GENERAL meshes, including SCF reuse
  and full-mesh diagonalization references.

## Correctness Fixes

Validation of the new in-process path exposed two existing export errors:

1. Berry integrals must use the same periodic atom images as the orbital
   neighbour lists. Inconsistent images produced unphysical overlap singular
   values greater than one in ionic-crystal tests.
2. Symmetry-reconstructed SCF coefficients must remain in the lattice-periodic
   AO gauge. Adding an atom-position phase for reciprocal-lattice folding was
   inconsistent with the SCF density transformation. Removing that extra phase
   restores overlap-metric covariance for nontrivial hexagonal atom positions.

The regression tests distinguish successful MO reuse from a validation fallback.
The AlN metric regression specifically tests the raw symmetry transformation;
it does not claim successful reuse at an insufficient integration-grid cutoff.
Existing reference values and reconstruction validation thresholds are unchanged.

## Validation

The local validation includes external Wannier90 4.0.2, not just creation of
`.win`, `.amn`, `.mmn` and `.eig` files:

- Compare overlap contraction bounds and reverse-neighbour Hermitian relations,
  projection rank, optimized U orthogonality, `V = U_dis U`, packed band mappings,
  and Fourier reconstruction of `H(k) = V^H E(k) V` within native text precision.
- Compare converged centres and spreads between library and standalone runs,
  allowing Wannier permutations and lattice translations. A different MLWF
  local minimum is not interpreted as an export error solely from its gauge.
- Exercise He/He2, H/FCC, Diamond, LiF, NaCl and hexagonal cells, including
  enlarged valence/conduction spaces, disentanglement and spin polarization.
- The all-electron GAPW H2 case uses `TZVPP-MOLOPT-GGA-ae`. SCF reuse,
  independent full-mesh diagonalization and `MP_GRID` agree in centres to
  approximately 2e-14 angstrom and in total spread to 3e-15 angstrom squared.
  Its library/standalone spread difference is below 4e-10 angstrom squared.
- Test the projection routines' unitary covariance and rank-repair behavior,
  and the Python MMN comparator's reverse-block and invalid-input checks.

After rebasing onto CP2K master `91605b4e95`, the selected official directory
regtests passed 211/211 in both serial (one rank, two OpenMP threads) and MPI
(two ranks, two OpenMP threads). These cover the four library directories, the
projection unit test, k-point export, SPGLIB, CASINO and polarization regressions.
An additional three-rank GAPW run agrees with standalone Wannier90 in centres
to 3e-15 angstrom and total spread to 4e-10 angstrom squared. Between serial and
three ranks, overlap-block singular values agree to 9e-15 despite different MO
gauges. The Python MMN comparator's six unit tests also pass.
The full `CP2K_USE_WANNIER90=OFF` build also succeeds at `d295ceb9c8` and passes
130/130 selected regression checks. The four library directories are correctly
skipped; the ordinary exporter still writes `.win`, `.amn`, `.mmn`, and `.eig`
files and reuses the SCF MOs. The binary has neither the Wannier90 feature flag
nor external `w90_*` API symbols or a Wannier90 dynamic dependency. Requesting
`LIBRARY ON` produces the intended missing-feature diagnostic. Its subsequent
SIGABRT teardown hangs on this macOS host and required terminating the launcher;
this is not claimed as a cleanly reaped negative-test process.

After restoring the library-enabled serial build, the same selected suite again
passes 211/211. The ordinary exporter produces identical numerical `.amn`,
`.eig`, and `.mmn` data in the enabled and disabled builds for the same input.
A fresh library/standalone replay agrees in centres to 4e-15 angstrom and in
total spread to 2e-9 angstrom squared.

The final documentation-only checkpoint also corrects the `USE_BLOCH_PHASES`
help: its equal-band-count restriction belongs to ordinary Identity-projection
export, not the library's physical projection/disentanglement path. Both builds
and a five-band/one-Wannier SCF-reuse smoke pass at `019ed62432`; serial/MPI
spreads agree to 2e-14 bohr squared. A fresh `make_pretty.sh --no-cache` run
checks all 93 changed files with zero failures.

## Usage

```text
&WANNIER90 ON
  KPOINTS_SOURCE SCF
  WANNIER_FUNCTIONS 4
  &LIBRARY ON
    INITIAL_PROJECTIONS AUTO
    NUM_ITER 1000
    CONV_WINDOW 5
    CONV_TOL 1.e-10
    WRITE_HR T
    WRITE_U_MATRICES T
    WRITE_INPUTS T
  &END LIBRARY
&END WANNIER90
```

The band count, energy windows and localization settings must be chosen for the
system. `CONV_TOL` follows Wannier90's angstrom-squared convention; CP2K reports
centres and spreads in atomic units.

## Scope and Limitations

- AUTO is a finite multistart comparison, not a global-minimum guarantee or a
  guarantee of unique individual centres in a nearly flat minimum.
- The library requires a complete uniform mesh, not a band path. It receives
  the full mesh even when the SCF calculation used symmetry reduction.
- Existing SCF-MO reuse restrictions remain: multiple SCF k-point process
  groups or failed reconstruction validation use the diagonalization fallback.
- Finite real-space integration grids can weakly break Hamiltonian symmetry.
  Tightening SCF convergence alone does not replace cutoff convergence.
- Collinear spin support is not spinor/SOC Wannierization. These output options
  do not write a Wannier90 checkpoint/restart file.
- Tested locally on macOS with GNU Fortran 16.1 and serial/MPI Wannier90 4.0.2.
  Linux/GCC CI is configured; its results are not included in the local validation above.


tools/spack/spack_repo/cp2k_dev/packages/wannier90/package.py
RMeli, line 1: I did not realize a new version of Wanneir90 was out. I added CMake support to the upstream `wannier90` package in https://github.com/spack/spack-packages/pull/6377.

Once the PR is merged, it might be good to drop this local recipe and update the Spack packages hash instead.

tools/toolchain/scripts/stage7/install_wannier90.sh
```diff
@@ -0,0 +1,121 @@
...
+      archive="wannier90-${wannier90_ver}.tar.gz"
+      if ! checksum "${wannier90_sha256}" "${archive}"; then
+        download_pkg_from_urlpath "${wannier90_sha256}" "v${wannier90_ver}.tar.gz" \
+          "https://github.com/wannier-developers/wannier90/archive/refs/tags" "${archive}"
+      fi
+      # Examples and reference data are large and are not needed for the library build.
+      tar -xzf "${archive}" "wannier90-${wannier90_ver}/CMakeLists.txt" \
+        "wannier90-${wannier90_ver}/LICENSE" \
+        "wannier90-${wannier90_ver}/cmake" "wannier90-${wannier90_ver}/src"
+      build_dir="wannier90-${wannier90_ver}/build-${wannier90_variant}"
```
Growl1234: I'm confused; what's the point of specifying a dedicated name of build directory here. I don't see multiple support configured together (like ELPA and DBCSR, both CPU version and GPU version are installed; do MPI and serial version installed together in Wannier90?).

tools/toolchain/scripts/stage7/install_wannier90.sh
```diff
@@ -0,0 +1,121 @@
...
+
+if [ "${MPI_MODE}" = "no" ]; then
+  wannier90_mpi=OFF
+  wannier90_variant=serial
+  wannier90_pkgconfig=wannier90
+else
+  wannier90_mpi=ON
+  wannier90_variant=mpi
+  wannier90_pkgconfig=wannier90_mpi
+fi
```
Growl1234, line 30: `wannier90_variant` the same question

tools/toolchain/scripts/stage7/install_wannier90.sh
```diff
@@ -0,0 +1,121 @@
...
+        "wannier90-${wannier90_ver}/cmake" "wannier90-${wannier90_ver}/src"
+      build_dir="wannier90-${wannier90_ver}/build-${wannier90_variant}"
+      mkdir -p "${build_dir}"
+      case "${MATH_MODE}" in
+        openblas) blas_vendor=OpenBLAS ;;
+        mkl) blas_vendor=Intel10_64lp ;;
+        acml) blas_vendor=ACML ;;
+        *) blas_vendor=All ;;
+      esac
+      cmake --fresh -S "wannier90-${wannier90_ver}" -B "${build_dir}" \
```
Growl1234: - Why `--fresh` is needed
- Please run CMake after entering the source tree, thus `cmake -S . -B build`

tools/toolchain/scripts/stage7/install_wannier90.sh
```diff
@@ -0,0 +1,121 @@
...
+    lib_dir="${pkg_install_dir}/lib"
+    [ -d "${pkg_install_dir}/lib64" ] && lib_dir="${pkg_install_dir}/lib64"
+  fi
+  check_dir "${lib_dir}/cmake/Wannier90"
+  cat << EOF > "${BUILDDIR}/setup_wannier90"
+prepend_path LD_LIBRARY_PATH "${lib_dir}"
+prepend_path LD_RUN_PATH "${lib_dir}"
+prepend_path LIBRARY_PATH "${lib_dir}"
+prepend_path PKG_CONFIG_PATH "${lib_dir}/pkgconfig"
+export Wannier90_DIR="${lib_dir}/cmake/Wannier90"
```
Growl1234: I don't like the idea of specifying `PKG_DIR` for CMake discovery. Since toolchain tends to assume users already get the environment variables ready for CP2K build, let's follow other scripts' practice, move them into below `if [ "${with_wannier90}" != "__SYSTEM__" ]; then` clause (though I'm not a fan of this, but there doesn't seem to be a better solution).

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -324,6 +324,9 @@ Specific options of --with-PKG:
   --with-spglib           Enable the spg library for symmetry groups detection.
                           This package depends on CMake.
                           Default = install
+  --with-wannier90        Enable the Wannier90 v4 library for k-point localization.
+                          Requires CMake and BLAS/LAPACK; follows --mpi-mode.
+                          Default = install
```
Growl1234: Since this is not core or very commonly used feature, I'd suggest disabling it by default.

tools/spack/cp2k_deps_p.yaml
```diff
@@ -236,6 +236,7 @@ spack:
     - "spla@1.6.1"
     - "tblite@0.7.0"
     - "trexio@2.6.1"
+    - "wannier90@4.0.2 +mpi"
```
Growl1234: Please put `+mpi` to above `packages:` section.

tools/spack/cp2k_deps_s-static.yaml
```diff
@@ -121,6 +121,7 @@ spack:
     - "pace@2025.12.4.1"
     - "spglib@2.7.0"
     - "tblite@0.7.0"
+    - "wannier90@4.0.2 ~mpi ~shared"
```
Growl1234: Same.

tools/spack/cp2k_deps_s.yaml
```diff
@@ -145,6 +145,7 @@ spack:
     - "spglib@2.7.0"
     - "tblite@0.7.0"
     - "trexio@2.6.1"
+    - "wannier90@4.0.2 ~mpi"
```
Growl1234: Same.

CMakeLists.txt
```diff
@@ -849,6 +851,39 @@ if(CP2K_USE_SPGLIB)
...
+    CP2K_WANNIER90_INTERFACE_WORKS
+    SRC_EXT
+    F90)
+  cmake_pop_check_state()
+  if(NOT CP2K_WANNIER90_INTERFACE_WORKS)
+    message(
+      FATAL_ERROR
+        "Wannier90 requires a compatible Fortran compiler and integer MPI communicator interface"
+    )
+  endif()
```
Growl1234, line 887: Would need more information on why this recheck is needed; does 4.0.0 or 4.0.1 version of Wannier90 have bugs?

tools/toolchain/scripts/stage7/install_wannier90.sh
```diff
@@ -0,0 +1,121 @@
...
+      write_checksums "${install_lock_file}" "${SCRIPT_DIR}/stage7/install_wannier90.sh" \
+        "${INSTALLDIR}/toolchain.conf"
+    fi
+    ;;
+  __SYSTEM__)
+    echo "==================== Finding Wannier90 from system paths ===================="
+    check_pkgconfig "${wannier90_pkgconfig}"
+    # The v4.0.2 .pc template uses colons instead of assignments for prefix/libdir.
+    # pcfiledir is supplied by pkg-config itself and also handles multiarch libdirs.
+    lib_dir="$(dirname "$(pkg-config --variable=pcfiledir "${wannier90_pkgconfig}")")"
```
Growl1234, line 86: I would (also) recommend specifying `pkg_install_dir` here.

Growl1234 (changes requested):

tools/spack/spack_repo/cp2k_dev/packages/wannier90/package.py
DCM-Uni-Paderborn: Thank you for adding the upstream CMake support! I agree that we should use the upstream recipe once spack/spack-packages#6377 is merged and the pinned package revision can include it. That PR is still open, so I have retained the local recipe for now to keep this PR buildable.

tools/toolchain/scripts/stage7/install_wannier90.sh
DCM-Uni-Paderborn: You are right: the toolchain installs one MPI mode per invocation, not both variants together. I have removed the variant-specific build/install directory names in fb54869df5. The installer now uses the usual versioned prefix and a single build directory. The configuration participates in the installation checksum, so changing the compiler or MPI mode triggers a clean rebuild.

tools/toolchain/scripts/stage7/install_wannier90.sh
DCM-Uni-Paderborn: Removed `wannier90_variant` as well in fb54869df5; `MPI_MODE` now only selects the MPI build option and the appropriate pkg-config name.

tools/toolchain/scripts/stage7/install_wannier90.sh
DCM-Uni-Paderborn: Done in fb54869df5: CMake is now called as `cmake -S . -B build` after entering the source directory. I removed `--fresh`. When installation checksums change, the generated build directory and installation prefix are cleaned, avoiding stale compiler modules or a cached MPI configuration.

tools/toolchain/scripts/stage7/install_wannier90.sh
DCM-Uni-Paderborn: Agreed, and changed in fb54869df5. The installer no longer exports `Wannier90_DIR`; non-system installations add their prefix to `CMAKE_PREFIX_PATH`, following the neighboring scripts. SYSTEM mode leaves all search paths unchanged. I tested fresh serial/MPI installations, both SYSTEM modes, and CMake consumers of the installed libraries.

tools/toolchain/install_cp2k_toolchain.sh
DCM-Uni-Paderborn: Done in fb54869df5: the default is now `--with-wannier90=no`, including the help text. Users can opt in with `--with-wannier90=install`. The existing CI `--install-all` configurations still enable it, so the library regression coverage remains active.

tools/spack/cp2k_deps_p.yaml
DCM-Uni-Paderborn: Moved `+mpi` into `packages: wannier90: require:` in fb54869df5. The spec now only pins the version. I also checked concretization of all three updated Wannier90 policies with Spack.

tools/spack/cp2k_deps_s-static.yaml
DCM-Uni-Paderborn: Done in fb54869df5: both `~mpi` and `~shared` now live in the package requirements. The serial/static policy concretizes successfully.

tools/spack/cp2k_deps_s.yaml
DCM-Uni-Paderborn: Done in fb54869df5: `~mpi` is now in the package requirements, and the spec only pins `wannier90@4.0.2`.

CMakeLists.txt
DCM-Uni-Paderborn: The recheck is an API/compiler-compatibility check, not a test for a release-specific numerical bug. The package version alone cannot establish that the Fortran .mod files are compatible with CP2K's compiler or that `w90_set_comm` accepts an integer communicator. Rechecking also covers replacing an installation at the same path. I expanded the comment in fb54869df5 to make this distinction explicit.

Separately, v4.0.0 did have an uninitialized `ierr` return in `w90_get_gkpb`; this was fixed for v4.0.1 by [59f9450](https://github.com/wannier-developers/wannier90/commit/59f94501ed4134699dfd57b2f32d766191a084ce). The build documentation therefore requires v4.0.1 or newer, and the installers pin v4.0.2. Unfortunately, v4.0.2 itself still reports 4.0.0 in its CMake package metadata, so that metadata cannot enforce the patch-version requirement. The compile check does not detect that v4.0.0 runtime bug.

tools/toolchain/scripts/stage7/install_wannier90.sh
DCM-Uni-Paderborn: Added `pkg_install_dir` for SYSTEM mode in fb54869df5. It uses the pkg-config prefix where available, with a fallback via `pcfiledir` for v4.0.2's malformed prefix/libdir assignments, including multiarch library directories. No SYSTEM search paths are modified.

DCM-Uni-Paderborn:
Thank you for the review. I have pushed three follow-up commits addressing the build/toolchain comments and CI failures.

- d1b40f4e6c fixes the common cause of all ten runtime failures in both pdbg jobs: the values-only ZGESVD path reaches LAPACK's IEEECK probe, whose intentional non-finite arithmetic triggers CP2K's debug traps. The call now saves and temporarily disables IEEE halting modes and restores them immediately afterwards, following CP2K's existing LAPACK-wrapper pattern. The unit test explicitly checks restoration for both enabled and disabled traps. I reproduced the original failure locally and verified the fix with debug-instrumented projection code, in serial and on two MPI ranks.
- fb54869df5 addresses the toolchain/CMake/Spack feedback. Wannier90 is now opt-in by default; the CI --install-all configurations still enable it. Fresh serial/MPI installs, SYSTEM/custom-prefix discovery, public-library consumers and all three Spack variant policies were checked.
- 7e83bbace0 reduces regression setup costs through shared ionic SCF restarts, faster Gamma SCF convergence and a one-step setup for the SCF-independent AO-metric check. All numerical reference values and matcher tolerances are unchanged. Two remaining expensive cases have documented timing exemptions, and the two shared binary restart files have documented size exemptions; no numerical checks are suppressed.

The nine selected regression directories pass locally: 211/211 with ssmp and 211/211 with psmp (2 MPI ranks, 2 OpenMP threads). External Wannier90 checks retain the total spreads; centre comparisons for the partial LiF space account for cubic symmetry. make_pretty/precommit checks pass. The complete Linux pdbg jobs still need the new upstream CI run.

Growl1234 (approved):

CMakeLists.txt
Growl1234: Usually Fortran module compatibility is not a concern that needs special care, as:
- In a package manager you lock everything under the system/specified compiler;
- If Fortran modules are incompatible because of compilers, then the linked library will most likely also be incompatible.

tools/regtesting/test_compare_wannier90_mmn.py
```diff
@@ -0,0 +1,54 @@
...
+"""Unit tests for the gauge-independent overlap reciprocity check."""
+
+import unittest
+
+import numpy as np
+
+from compare_wannier90_mmn import check_reciprocity
+
+
+class ReciprocityTests(unittest.TestCase):
```
oschuett, line 16: Since you're adding `wannier90_projections_unittest.F`, is this Python code still needed?

DCM-Uni-Paderborn: Thanks, Ole. These cover different functionality: the Fortran test exercises AO projection selection and covariance, while the Python tests validate the MMN reciprocity checker, including malformed blocks and missing reverse connections. I clarified the distinction and added the Python tests to the Misc CI job in 0c6b2d22fb. All six Python tests pass, as does the Fortran projection test locally with SSMP and two MPI ranks.

DCM-Uni-Paderborn:
CI follow-up for the previous head (6470253a4a): the Spack pdbg run passed 6374/6375 checks. Its only failure is Fist/regtest-5-vib/N3dye_geoopt_inverse.inp (M007): 0.5616277603 versus reference 0.5616277602, with relative difference 1.78e-10 and tolerance 1e-14.

The same test fails with exactly the same value and deviation in the upstream master Spack report for 3919fb7b1c1b8804bc2b603e2cadadc551023053, independently of this PR:
- [PR Spack report](https://storage.googleapis.com/cp2k-ci/cp2k-spack-pdbg-6f85fb17_report.txt)
- [Upstream master Spack report](https://storage.googleapis.com/cp2k-ci/dashboard_spack-pdbg_report.txt)

All Wannier90 and topology cases passed in that Spack run. I have left the unrelated FIST reference and tolerance unchanged. The new commit 0c6b2d22fb adds the Python overlap-checker unit tests to Misc CI and clarifies their scope; it does not alter numerical code or reference values. Local Python and serial/MPI Fortran unit tests, make_pretty, and the pre-push checks pass.

tools/spack/spack_repo/cp2k_dev/packages/wannier90/package.py
DCM-Uni-Paderborn: Follow-up now that spack/spack-packages#6377 has merged: the current PR already uses upstream spack-packages commit abf3f6003f33e51046a76b829b892317dab96d29 in all three dependency policies, and the local Wannier90 recipe has been removed. The current-head Spack pdbg run passes all 6391 tests. Many thanks again for the upstream recipe!

CMakeLists.txt
DCM-Uni-Paderborn: Agreed. The current revision no longer has the extra compiler/module/API try_compile probe. It uses the normal find_package(Wannier90 4.0 REQUIRED CONFIG) path and retains only the explicit CP2K/Wannier90 MPI-mode consistency check. This avoids duplicating package-manager/compiler compatibility checks while keeping a clear error for mixing serial and MPI installations.

DCM-Uni-Paderborn:
Dear @mkrack and @mtaillefumier,

Could you please take another look at the Wannier90 v4 library integration when you have a chance? Your requested reviews are still outstanding. The current head passes the regular regression suite, Spack pdbg, manual generation, Misc and precommit checks, and the inline review threads are resolved. A final review of the k-point/library integration and build setup would be very helpful; please let me know if anything still needs changing before merge.

Many thanks!
Greetings,
Thomas

tools/spack/cp2k_deps_s.yaml
```diff
@@ -112,9 +112,18 @@ spack:
       require:
         - "+hdf5"
 
+    wannier90:
+      require:
+        - "~mpi"
+
+    # The pinned yasm CMake recipe omits its C++ build dependency.
```
mkrack: outdated see #5961

tools/spack/cp2k_deps_s.yaml
```diff
@@ -112,9 +112,18 @@ spack:
...
+    yasm:
+      require:
+        - "build_system=autotools"
+
   repos:
     builtin:
-      commit: 4d092a104627e9871a0d74f853d6df9c8e6105c2 # 2026-08-11
+      commit: abf3f6003f33e51046a76b829b892317dab96d29 # 2026-09-15
```
mkrack: outdated with #5961

src/wannier90_projections.F
```diff
@@ -0,0 +1,317 @@
...
+            nr = nb*nk
+            IF (anchor > 0) nr = nb
+            ALLOCATE (qr(nr, nc))
+            IF (anchor == 0) THEN
+            DO ik = 1, nk
+               qr((ik - 1)*nb + 1:ik*nb, :) = candidates(:, :, ik)
+            END DO
+            ELSE
+            qr(:, :) = candidates(:, :, anchor)
+            END IF
```
mkrack, line 161: These lines are not properly indented.

tools/spack/cp2k_deps_s.yaml
DCM-Uni-Paderborn: Dear Matthias,

Thank you. I have aligned the pin and Yasm comment in all three Spack manifests with #5961 in 02b3c3b81f. The Autotools requirement itself is retained: at builtin commit 3761e323f59bec2046ecafd544c3cb15aca8b514, the Yasm recipe still declares only a C compiler dependency, not the C++ dependency needed by its CMake path. This also matches the current handling in #5961.

I checked concretization against an isolated checkout at that exact pin: the MPI, serial and static Wannier90 4.0.2 policies and Yasm Autotools selection all pass. These are local policy checks with external dependencies; the full Linux Spack CI run still needs to confirm the new head.

Greetings,
Thomas

tools/spack/cp2k_deps_s.yaml
DCM-Uni-Paderborn: Updated all three manifests to the 2026-09-20 builtin pin 3761e323f59bec2046ecafd544c3cb15aca8b514 used in #5961. The branch is also rebased onto current master e86d818d07; the Stanene conflict was resolved by retaining the master input unchanged. Fresh MPI validation passes all 176 checks in the ten selected regression directories, including topology. No reference values or matcher tolerances were adjusted.

src/wannier90_projections.F
DCM-Uni-Paderborn: Fixed in 02b3c3b81f, thank you. The IF/DO block is now correctly indented. make pretty and the pre-push checks pass for all 99 changed files; the projection unit test also passes with one and two MPI ranks.

DCM-Uni-Paderborn:
The remaining Spack pdbg failure was the existing `auto-ri-size-H2O.inp`
GW regression, not the Wannier90 tests. The master dashboard reproduces
the same 22.632 eV result against the old 22.705 eV reference.

I traced this to the underresolved 21-point RI-RS grid. On Spark, changing
only OpenBLAS between 0.3.33 and 0.3.26 changes the original gap from
22.566 to 22.618 eV. Tightening EPS_SCF does not help, and merely increasing
the optimizer iteration limit does not give a stable reference.

The follow-up changes `RS_AO_RATIO` from 3 to 6 (42 points). Both BLAS
versions and one/two MPI ranks then give 22.667 eV, independently confirmed
by denser optimized grids and the fixed 608-point standard grid. The
normalized 3-center fitting error drops from approximately 6e-5 to below
5e-11. This is the reason for the new gap reference, rather than matching
the failing CI number.

All existing tolerances, the 18-function AUTO_RI reference, EPS_SCF and
MAX_ITER=100 remain unchanged. An additional matcher checks the 42-point
grid. No test is disabled. The two-rank runtime remains approximately
1.62 seconds in the controlled local comparison.

Validation: the complete GW directory passes 33/33 checks with OpenBLAS
0.3.33 (2 MPI x 2 OpenMP) and again with 0.3.26 (4 MPI x 1 OpenMP).
`make_pretty.sh --no-cache` passes all 101 changed PR files. The fresh
upstream CI still needs to confirm the pushed change.

Issue #5938, opened 2026-09-01
Feature request: the instant energy task after opt task


garrettwx:
Hi, devs,

The cp2k supports the single-point energy after opt task (GEO_OPT and CELL_OPT), but with the same level of basis_set,

What about the internal support for changing the basis_set, e.g. DZVP-MOLOPT-SR-GTH (opt) → TZVP?

Growl1234:
What's the point? IIRC no programs I know will do that.

garrettwx:
> What's the point? IIRC no programs I know will do that.

Sorry for the inaccuracy, I mean just the energy calculation after the opt task with the opted structure.

Growl1234:
We should at first find out why the evaluation after optimization completed will be performed.

garrettwx:
If I am not mistaken, I find the 

```fortran
RECURSIVE SUBROUTINE write_final_info(output_unit, conv, it, gopt_env, x0, master, para_env, force_env, &
                                         motion_section, root_section)
```
in gopt_f_methods.F that allows for the "Reevaluating energy at the minimum".

Growl1234:
Okay. Why that would be needed.

garrettwx:
I think it mainly intends to perform the independent evaluation for the energy...

he-zilong:
This piece of code is from decades ago... Without source code comments or accounts from then-developers, this design appears rather obscure.
https://github.com/cp2k/cp2k/blob/master/src/motion/gopt_f_methods.F#L899-L911
Note that this routine is for optimization tasks in general, independent of the underlying force method(s). It does not even know if DFT is used, let alone the size of basis set.

he-zilong:
This piece of code is from decades ago... Without source code comments or accounts from then-developers, it is indeed a bit obscure as to why it is designed so. The result is not even compared in terms of the numerical accuracy with that from the final evaluation in the trajectory.
https://github.com/cp2k/cp2k/blob/011c742228f3c2b711d369ea9d44ae8f31c92370/src/motion/gopt_f_methods.F#L899-L911
Note that this routine is for optimization tasks in general, independent of the underlying force method(s). It does not even know if QuickStep DFT is used, let alone the size of basis set.
On the other hand, while I understand the recommended computational chemistry practice, I do not quite see the point of supporting an "optimization + single-point evaluation" compound job with keywords.

Growl1234:
[...]

Issue #5534, opened 2026-07-05
Feature request: Allow constraining individual lattice vectors during CELL_OPT


minsanglee17979:
Hello CP2K developers,

First of all, thank you very much for developing and maintaining CP2K.

I am using the latest CP2K version for periodic DFT calculations, and I would like to suggest a feature that could be useful for cell optimizations involving triclinic simulation cells.

Currently, `CELL_OPT` provides several useful constraints and projections, for example:

- `CONSTRAINT X/Y/Z`
- `KEEP_ANGLES`
- `KEEP_VOLUME`
- `KEEP_SYMMETRY`
- `KEEP_SPACE_GROUP`

However, as far as I understand, there is currently no way to constrain individual lattice vectors themselves (A, B, or C).

For orthogonal cells, `CONSTRAINT X/Y/Z` can often achieve the desired behavior because the lattice vectors are aligned with the Cartesian axes.

For triclinic cells, however, constraining Cartesian X/Y/Z components is generally not equivalent to fixing the lattice vectors themselves.

For example, it would be useful to perform cell optimizations such as

- keep lattice vector A fixed while allowing B and C to relax,
- keep A and B fixed while allowing only C to relax,
- or other combinations of fixed lattice vectors.

Such functionality would be useful for studies involving strained materials, interfaces, layered materials, and other systems where one or more lattice vectors should remain fixed during optimization.

Would it be possible to consider adding this functionality in a future version of CP2K?

One possible interface could be something like

```text
&CELL_OPT
  FIX_CELL_VECTORS A
&END CELL_OPT
```

or

```text
&CELL_OPT
  FIX_CELL_VECTORS A B
&END CELL_OPT
```

or any equivalent implementation that allows constraining individual lattice vectors during `CELL_OPT`.

Issue #1231, opened 2020-12-18
Drop F77 interfaces


dev-zero:
We have a couple of F77 interfaces:

* src/motion/gopt_f77_methods.F
* src/f77_interface.F

Some are somewhat documented (for `src/f77_interface.F` there is an example in `tools/f77_interface` for which one example is already broken), others are not (`src/motion/gopt_f77_methods.F`). Both of them are unfortunately not "leaf"-modules , but are used by other F2003+ code in CP2K. Furthermore they are not even F77 anymore since they contain for example `OPTIONAL` dummy argument descriptors (introduced in F90).

Simply moving them in a `MODULE`  in the case of `src/motion/gopt_f77_methods.F` does not work since it triggers a reverse-dependency loop in our build system, while the `src/f77_interface.F` is highly entangled with the CP2K startup code, the farming and other features.

As documented in #792 there is some need for a programmatic interface to CP2K, but also there the standard being used is F90.

fstein93:
@oschuett I know this post is rather old. But there is potentially a way out of this dependency issue: Fortran's submodules. Check [Steve Lionel's post on this topic](https://stevelionel.com/drfortran/2015/07/07/doctor-fortran-in-we-all-live-in-a-yellow-submodule/). There, he explicitly mentions that circular dependencies are possible: `Module A can use module B, and a submodule of B can use module A! This doesn’t create a circular dependency because module A can’t see into the submodule of B.` Surely, one has to check whether and how we can achieve the disentanglement of our f77 interface and how to ensure the compatibility with our build system.
About the compiler support: Intel implements this feature since version 16, Gfortran since version 6. I don't know about the others.

dev-zero:
@fstein93 back when I opened the issue Jürg and I already discussed about submodules and I agree that this is the way to go. Please keep in mind that for submodules we also need the build systems to support it, not only the compilers.
CMake with Ninja should be able to handle the dependencies properly, CMake with Make I am not so sure about (depends on CMakes built-in Fortran parser).

oschuett:
From what I understand, the only real advantage of submodules is that they can speed up the build in some cases. At least Gfortran is already smart enough to rewrite the `.mod` files only when the public interface has actually changed.

I think, a much more promising approach are abstract types. I really liked this [exploration by Tiziano](https://github.com/cp2k/cp2k/pull/356/files#diff-eee1e1eb778fe26f397f68b8766f72ed5247174ab329c3783decfb9f68585a9b).

[...]

PR #5986, opened 2026-09-07
Motion: remove the legacy GOPT F77 interface

### LINKED ISSUE #1233: Clean up gopt_f77_methods.f90

RTChou67:
This PR addresses #1233 . Moved `cp_eval_at` to the `gopt_f_methods` module, and remove `gopt_f77_methods.F` and `gopt_f77_methods.fypp`. I introduces the `gopt_f_dimer_rotation_optimizer` ABSTRACT INTERFACE in `gopt_f_types`, together with the corresponding `dimer_rotation_optimizer` procedure-pointer component of `gopt_f_type`. `geo_opt` assigns `cp_rot_opt` to this procedure pointer, while `dimer_methods` invokes the rotation optimizer through the pointer without directly depending on  `geo_opt`. This solves the circular dependency problem.


mkrack (approved):
Seems to work.

src/motion/gopt_f_methods.F
```diff
@@ -905,7 +1126,7 @@ RECURSIVE SUBROUTINE write_final_info(output_unit, conv, it, gopt_env, x0, maste
             WRITE (UNIT=output_unit, FMT="(/,T20,' Reevaluating energy at the minimum')")
          END IF
 
-         CALL cp_eval_at(gopt_env, x0, f=etot, master=master, final_evaluation=.TRUE., &
```
oschuett, line 908: I wonder, why did you remove the `final_evaluation` argument?

oschuett (commented):
Nice work!

src/motion/gopt_f_methods.F
he-zilong: > I wonder, why did you remove the `final_evaluation` argument?

There are questions in #5938 for the necessity of a final single-point energy evaluation.

RTChou67: This is actually a dead argument that is used in CELL_OPT/TYPE. As  CELL_OPT/TYPE is removed in #5257 , there is no need to keep it.

oschuett: Thanks for the link. Since this is a behavior change, I'd prefer to move that to a separate PR. Also, I'm not yet fully convinced the final evaluation is useless.

mkrack: The re-evaluation of the energy at the minimum is a convergence sanity (consistency) check. It is a safeguard against an accidental convergence from extrapolation for instance. I think QE performs something similar. I suggest to keep this feature by default, but it could be switched off optionally by an input keyword.

RTChou67: Thanks. I agree that this behavior change should not be included in this PR. I'll modify.

he-zilong:
Nothing substantial, just some random remarks.

I see that there is quite some diff noise due to the precommit formatting of USE-ONLY arguments; it is unclear to me why the formatter flip-flops between the compact `USE module, ONLY something` form and the aligned `USE module, [whitespaces for right justification] ONLY something` form.

The other legacy F77 interface as discussed at #792 and #1231 may receive the same treatment, but be careful as it is the entry point for the whole program.

To this day I am still puzzled by the use of the `BLOCK` construct in `SUBROUTINE cp_eval_at`; although it is a Fortran 2008 feature, I may prefer seeing it refactored as a standalone subroutine.

Finally, there was some ugly pre-processing before passing information to my `write_final_structure` subroutine at `src/motion/gopt_f_methods.F`; with this PR merged I can set out to clean it up.

RTChou67:
Thanks for the comments. The remaining f77 interface will be treated as a separate follow-up, if I have time. The BLOCK refactor and the cleanup around write_final_structure also make sense. If you feel it neccessary, I can clean it up in this PR.

oschuett:
> I see that there is quite some diff noise due to the precommit formatting ...

Yes, this is quite distracting. Fpretty (our Fortran formatter) predates Fypp. As a band-aid it simply bails out when it encounters fypp macros. So, when all fypp macros are removed from a file, the fpretty springs back to life.

> ... I may prefer seeing it refactored as a standalone subroutine.

Good idea.

> with this PR merged I can set out to clean it up.

I'm looking forward :-)

> The other legacy F77 interface as... may receive the same treatment,

Yeah, [f77_interface.F](https://github.com/cp2k/cp2k/blob/master/src/f77_interface.F) is a different beast.  It'll probably take a couple of PRs to disentangle. Let's first discuss a strategy of attack.

src/motion/gopt_f_methods.F
```diff
@@ -13,76 +13,80 @@
...
 #include "../base/base_uses.f90"
 
    IMPLICIT NONE
    PRIVATE
 
-   #:include "gopt_f77_methods.fypp"
-
    LOGICAL, PRIVATE, PARAMETER :: debug_this_module = .TRUE.
-   CHARACTER(len=*), PARAMETER, PRIVATE :: moduleN = "gopt_f_methods"
+   CHARACTER(len=*), PARAMETER, PRIVATE :: moduleN = 'gopt_f_methods'
```
he-zilong, line 86: Is this quote change intentional or is it also a precommit behavior?

he-zilong:
> [T]he cleanup around write_final_structure also make sense. If you feel it neccessary, I can clean it up in this PR.

You can leave this to me for a while; as you can see it is tied to the `constraint_id` which may be worked on while implementing the feature requested by #5534.

src/motion/gopt_f_methods.F
RTChou67: This is intentional, as i see other files in motion/ uses single quotation marks. It seems that cp2k itself uses both single and double. I suggest adding a rule to precommit that unifies it.

oschuett: It's fpretty.

he-zilong: > It's fpretty.

Okay, I see that the single-quote style is mostly consistent across the source code now; apart from `gopt_f_methods.F`, there are only three other instances of double quotes, all under the `src/emd` directory in `rt_bse.F`, `rt_bse_io.F` and `rt_bse_types.F`.

src/motion/gopt_f_methods.F
oschuett: It seems our messages crossed. If this was indeed just a defunct argument left over from #5257, then removing it was actually the right call. Sorry for the noise. @he-zilong, maybe you can remove it in your upcoming PR?

PR #5990, opened 2026-09-08
using system openblas for archtecture detection

jpoto:
On certain system the installation by the openblas system detection messes up the paths. 
Using system openblas for detection should avoid this problem.

### DIFF
--- a/tools/toolchain/scripts/get_openblas_arch.sh
+++ b/tools/toolchain/scripts/get_openblas_arch.sh
@@ -17,7 +17,7 @@ source "${INSTALLDIR}"/toolchain.conf
 source "${INSTALLDIR}"/toolchain.env
 
 find_openblas_dir() {
-  find . -maxdepth 1 -type d -name '*OpenBLAS*' 2>/dev/null | head -1
+  find . -maxdepth 1 -type d -name '*OpenBLAS*' 2> /dev/null | head -1
 }
 
 ! [ -d "${BUILDDIR}" ] && mkdir -p "${BUILDDIR}"
@@ -30,9 +30,9 @@ get_system_openblas_arch() {
   local openblas_dir=""
 
   if [ "${with_openblas}" = "__SYSTEM__" ]; then
-    openblas_lib=$(find /usr/lib /usr/local/lib -maxdepth 3 -name "libopenblas*" -type f 2>/dev/null | head -1)
+    openblas_lib=$(find /usr/lib /usr/local/lib -maxdepth 3 -name "libopenblas*" -type f 2> /dev/null | head -1)
   elif [ "${with_openblas}" != "__INSTALL__" ] && [ -d "${with_openblas}" ]; then
-    openblas_lib=$(find "${with_openblas}/lib" -maxdepth 2 -name "libopenblas*" -type f 2>/dev/null | head -1)
+    openblas_lib=$(find "${with_openblas}/lib" -maxdepth 2 -name "libopenblas*" -type f 2> /dev/null | head -1)
   fi
 
   if [ -z "$openblas_lib" ]; then
--- a/tools/toolchain/scripts/get_openblas_arch.sh
+++ b/tools/toolchain/scripts/get_openblas_arch.sh
@@ -17,40 +17,76 @@ source "${INSTALLDIR}"/toolchain.conf
 source "${INSTALLDIR}"/toolchain.env
 
 find_openblas_dir() {
-  local __dir=''
-  for __dir in *OpenBLAS*; do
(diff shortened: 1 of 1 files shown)


tools/toolchain/scripts/get_openblas_arch.sh
```diff
@@ -17,40 +17,76 @@ source "${INSTALLDIR}"/toolchain.conf
...
-if ! [ "$openblas_dir" ]; then
-  retrieve_package "${openblas_sha256}" "${openblas_pkg}"
-  tar -xzf ${openblas_pkg}
+
+get_system_openblas_arch() {
+  local openblas_lib=""
+  local openblas_dir=""
+
+  if [ "${with_openblas}" = "__SYSTEM__" ]; then
+    openblas_lib=$(find /usr/lib /usr/local/lib -maxdepth 3 -name "libopenblas*" -type f 2> /dev/null | head -1)
```
Growl1234: Please use `find_in_paths "libopenblas" $LIB_PATHS` instead.

tools/toolchain/scripts/get_openblas_arch.sh
```diff
@@ -17,40 +17,76 @@ source "${INSTALLDIR}"/toolchain.conf
...
+    fi
+  done
+
+  if [ -n "$makefile_conf" ]; then
+    OPENBLAS_LIBCORE="$(sed -n 's/^LIBCORE=//p' "$makefile_conf")"
+    OPENBLAS_ARCH="$(sed -n 's/^ARCH=//p' "$makefile_conf")"
+    return 0
+  fi
+
+  local openblas_so="$(readelf -A "$openblas_lib" 2> /dev/null | grep 'Unknown' | head -1)"
```
Growl1234: If I'm not mistaken, `readelf -A` is not valid for dynamic library, so only `OPENBLAS_ARCH="$(uname -m)"` will be accepted. Use `readelf -h libopenblas.so | grep Machine` instead.

tools/toolchain/scripts/get_openblas_arch.sh
```diff
@@ -17,40 +17,76 @@ source "${INSTALLDIR}"/toolchain.conf
...
+
+  if [ -z "$openblas_lib" ]; then
+    return 1
+  fi
+
+  openblas_dir="$(dirname "$(dirname "$openblas_lib")")"
+
+  local makefile_conf=""
+  for dir in "${openblas_dir}/include" "${openblas_dir}/share/OpenBLAS" "/usr/share/OpenBLAS" "/usr/include"; do
+    if [ -f "${dir}/Makefile.conf" ]; then
```
Growl1234, line 46: Is there any system distribution or package managers installing OpenBLAS with a `Makefile.conf` file?

tools/toolchain/scripts/get_openblas_arch.sh
Growl1234, line 58: Most likely this would not work (at least on my x86_64 machine with both CentOS 10 Stream and Ubuntu 26.04 systems it returns nothing).

tools/toolchain/scripts/get_openblas_arch.sh
Growl1234, line 33: Why not use `find_in_paths "libopenblas*" $LIB_PATHS`.

Growl1234 (changes requested):
This seems to have added complexity to the toolchain script without providing expected functionality: I cannot get the expected result on my machine by following the proposed steps. Could you point to a concrete case or environment where the existing workflow is actually broken and such additional detection is needed?

oschuett:
Maybe this is just over-engineered and instead of using openblas we could detect the architecture with a 10 line script?

Growl1234:
> Maybe this is just over-engineered and instead of using openblas we could detect the architecture with a 10 line script?

For >90% cases (including macOS I guess) we can simply use `uname -m` and so we can put it directly in such as `setup_buildtools.sh`. `OPENBLAS_LIBCORE` is only used for installing OpenBLAS.

Growl1234:
We can even drop definition of `OPENBLAS_LIBCORE`; see https://github.com/OpenMathLib/OpenBLAS/blob/develop/docs/user_manual.md?utm_source=chatgpt.com#normal-compile

Growl1234:
@oschuett I've submitted a possible solution: #6023

jpoto:
I am open for any solution. 

The problem I encountered is that when I tried using system openblas it also linked the one for architecture detection. 
I did not see any environment variables that could explain it.

jpoto:
There is a better solution

PR #5992, opened 2026-09-08
Move non-standard CMake build types to presets

### LINKED ISSUE #5781: RFC: Move non-standard CMake build types to presets
With CMake presets now in use, I think we should consider moving our non-standard CMake build types to presets.

Currently, besides the standard configurations such as `Debug` and `Release`, we define several CP2K-specific build types in `CompilerConfiguration.cmake`:

* `CONVENTIONS`
* `COVERAGE`
* `ASAN`

These are not really build configurations in the same sense as `Debug` or `Release`. They are better described as **developer/compiler profiles** that add a particular set of compiler and linker flags.

I therefore propose moving them to `CMakePresets.json`, e.g.

```text
conventions
coverage
asan
```

while keeping `CMAKE_BUILD_TYPE` limited to the standard CMake configurations (`Debug`, `Release`, `RelWithDebInfo`).

A possible implementation would use `RelWithDebInfo` as the underlying configuration and let each preset override the corresponding configuration-specific flags, for example:

```json
{
  "name": "coverage",
  "cacheVariables": {
    "CMAKE_BUILD_TYPE": "RelWithDebInfo",
    "CMAKE_C_FLAGS_RELWITHDEBINFO": "...",
    "CMAKE_CXX_FLAGS_RELWITHDEBINFO": "...",
    "CMAKE_Fortran_FLAGS_RELWITHDEBINFO": "...",
    "CMAKE_EXE_LINKER_FLAGS_RELWITHDEBINFO": "..."
  }
}
```

This would also allow us to remove the corresponding `$<CONFIG:CONVENTIONS>`, `$<CONFIG:COVERAGE>`, and `$<CONFIG:ASAN>` handling from `CompilerConfiguration.cmake`, as well as the special `COVERAGE` linker handling.

I see a few advantages:

* `CMAKE_BUILD_TYPE` would retain its standard CMake meaning.
* `CompilerConfiguration.cmake` could focus on compiler configuration rather than developer build profiles.
* These profiles would become discoverable through `cmake --list-presets`.
* Adding similar developer configurations in the future would not require introducing additional custom CMake build types.
* The distinction between the actual build configuration and additional compiler/instrumentation settings would become clearer.

### Preset composition

One point that probably deserves discussion is how these presets should interact with the existing architecture/compiler presets, such as:

```text
native-gnu-x86_64
native-gnu-arm64
native-intel
```

These are conceptually orthogonal:

```text
architecture/optimization profile
        +
developer/instrumentation profile
```

For example, the current CI effectively supports combinations such as:

```text
native-gnu-x86_64 + CONVENTIONS
native-gnu-x86_64 + COVERAGE
```

while **CMake only allows one configure preset to be selected directly on the command line**.

We could handle this with inherited composite presets where needed, for example:

```json
{
  "name": "coverage-native-gnu-x86_64",
  "inherits": [
    "coverage",
    "native-gnu-x86_64"
  ]
}
```

However, explicitly defining every combination of developer and architecture presets would quickly lead to a combinatorial explosion.

`make_cp2k.sh --check-conventions` also needs some consideration here.

So before preparing the actual change, I would like to get some feedback on:

1. whether moving `CONVENTIONS`, `COVERAGE`, and `ASAN` from custom build types to presets makes sense in general; and
2. how we want to represent combinations of architecture presets and developer profiles.

CC @oschuett @mkrack @mtaillefumier @LecrisUT @he-zilong

Growl1234:
Closes #5781.

This is a patch I've prepared long time ago but forgot to push.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1007,7 +1007,7 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
             FORCE)
   # set the possible values of build type for cmake-gui
   set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release"
-                                               "Coverage")
+                                               "RelWithDebInfo")
 endif()
 
 # compiler configuration could have impacted package discovery (above)
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -32,6 +32,52 @@
       }
     },
     {
+      "name": "_developer-profile",
+      "hidden": true,
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo"
+      }
+    },
+    {
+      "name": "conventions",
+      "inherits": "_developer-profile",
+      "displayName": "GNU conventions check",
+      "description": "Build CP2K with additional GNU Fortran convention diagnostics.",
+      "cacheVariables": {
+        "CMAKE_C_FLAGS_RELWITHDEBINFO": "-O1",
+        "CMAKE_CXX_FLAGS_RELWITHDEBINFO": "-O1",
+        "CMAKE_Fortran_FLAGS_RELWITHDEBINFO": "-O1 -Wno-pedantic -Wall -Wextra -Wsurprising -Warray-temporaries -Wconversion-extra -Wimplicit-interface -Wimplicit-procedure -Wreal-q-constant -Walign-commons -Wfunction-elimination -Wrealloc-lhs -Wcompare-reals -Wzerotrip -fdump-fortran-original"
+      }
+    },
+    {
+      "name": "coverage",
+      "inherits": "_developer-profile",
+      "displayName": "Coverage build",
+      "description": "Build CP2K with GNU coverage instrumentation.",
+      "cacheVariables": {
+        "CMAKE_C_FLAGS_RELWITHDEBINFO": "-coverage -fkeep-static-functions -O1 -D__NO_ABORT",
+        "CMAKE_CXX_FLAGS_RELWITHDEBINFO": "-coverage -fkeep-static-functions -O1 -D__NO_ABORT",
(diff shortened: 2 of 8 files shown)


Growl1234:
```
Flag '__NO_ABORT' not mentioned in src/CMakeLists.txt nor cmake/CompilerConfiguration.cmake
```

How should I deal with this precommit error, as `-D__NO_ABORT` has been moved out of internal CMake files in this PR?

oschuett:
> How should I deal with this precommit error, as -D__NO_ABORT has been moved out of internal CMake files in this PR?

I'd just add a CMake flag `-DCP2K_NO_ABORT` and then document it [here](https://manual.cp2k.org/trunk/getting-started/build-from-source.html#cmake-configuration-options) saying something like:

_"Call `STOP` instead of aborting. Only used for coverage testing."_

Growl1234:
Sounds strange. How about adding `CMakePresets.json` to check list?

oschuett:
> Sounds strange. How about adding `CMakePresets.json` to check list?

Sorry, I'm lost. What is strange, and which checklist?

Growl1234:
Adding a `-DCP2K_NO_ABORT` essentially just adds back the behavior we want to move out, while decoupling it from the `Coverage` configuration itself. Therefore, I currently think it's better to also [detect flags from CMake preset file](https://github.com/cp2k/cp2k/pull/5992/changes#diff-40042e962f07521f11f845cd83355069557a436b7021f36fdabe11f034d8655d).

oschuett:
> Therefore, I currently think it's better to also [detect flags from CMake preset file](https://github.com/cp2k/cp2k/pull/5992/changes#diff-40042e962f07521f11f845cd83355069557a436b7021f36fdabe11f034d8655d).

Got it! You're right, there really is no use case for `__NO_ABORT` outside of coverage testing.

oschuett (approved):

mtaillefumier (approved):

PR #5993, opened 2026-09-08
Optimize MPI parallelism of k-point calculations

Growl1234:
- [x] Allow uneven k-point distribution across equal-sized MPI groups.
- [x] Adapt the k-point operator service to uneven k-point rounds.
- [x] Cache and reuse S(k) across SCF iterations.
- [x] Avoid redundant DBCSR desymmetrization.
- [x] Route k-point operators directly from DBCSR to FM and cache the routing plans.
- [x] Make the K→R density transformation group-local.
- [x] Add an adaptive lattice-FFT path for the grouped density transformation. (note: FFTW is needed for arbitrary-length exact grids, so using "native" FFTSG might not bring expected performance improvement.)
- [x] Make the forward R→K operator construction group-local.
- [x] Reduce or eliminate the remaining per-k-point forward communication.

Still missing (will be brought as follow-up PR)
- Revisit the distributed/local CFM backend and MPI/OpenMP hybrid path after the forward data path is improved. (expected to reduce a lot of code complication)


src/qs_kpoint_operators.F
```diff
@@ -89,7 +98,31 @@ MODULE qs_kpoint_operators
...
    !> State transitions happen only inside the service routines.
    INTEGER, PARAMETER, PRIVATE :: kpoint_slot_idle = 0, &
-                                  kpoint_slot_in_flight = 1
+                                  kpoint_slot_in_flight = 1, &
+                                  kpoint_slot_local = 2
+
+   ! Structural route shared by all captured rows. Whole blocks go to one
+   ! representative group, then only equivalent tiles are replicated.
+   TYPE kpoint_operator_rs_route_type
+      TYPE(kp_transform_plan_type) :: traversal
```
fstein93: Please initialize it to `kp_transform_plan_type()` if possible

RTChou67:
Do you have benchmark results for this new implementation?

Growl1234:
> Do you have benchmark results for this new implementation?

See Performance OpenMP CI results.

src/qs_kpoint_operators.F
```diff
@@ -451,59 +448,603 @@ SUBROUTINE kpoint_operator_ensure(ctx, rsmat)
...
+      END DO
+      IF (PRESENT(TARGET)) THEN
+         target%local_data(:, :) = z_zero
+      ELSE
+         CPASSERT(PRESENT(fm_re) .AND. PRESENT(fm_im))
+         fm_re%local_data(:, :) = 0.0_dp
+         fm_im%local_data(:, :) = 0.0_dp
+      END IF
+!$OMP PARALLEL DO DEFAULT(NONE) SHARED(snapshot, route, target, fm_re, fm_im) &
+!$OMP PRIVATE(j, col, iblock, row, local_row, r0, c0, nr, nc, pos, value) SCHEDULE(STATIC)
```
RTChou67:
```
DO j = 1, SIZE(route%matrix_struct%col_indices)
       DO iblock = 1, SIZE(route%entry, 2)
          xxxxx
       END DO
    END DO
```
This loop scans all route%entry entries for every local FM column. This might cause a CPU/cache bottleneck for large systems. The `kpoint_operator_rs_build` is a bottleneck. For the ZnO benchmark using 32 MPI ranks, it is called 3102 times and consumes 26.1s out of a 57.1s run.

Growl1234: Nice catch!

Growl1234: I tried indexing the routed blocks by AO column block so that each local FM column only visits the relevant normal/mirrored entries. Unfortunately, it gives essentially no measurable improvement in my local tests.

RTChou67: Well, that sounds actually reasonable. These might need very deep analysis. Maybe that belongs to another PR.

Growl1234: It turns out that accumulation is the main burden. I'll keep trying to handle it. One reminder is that `kpoint_operator_rs_build` is not always the main point; for 15x15x9 k-points calculation I got
```
 -                                T I M I N G                                  -
 CP2K                                 1  1.0    0.008    0.010   30.973   30.975
 qs_energies                          1  2.0    0.001    0.001   30.852   30.854
 scf_env_do_scf                       1  3.0    0.000    0.000   29.592   29.594
 scf_env_do_scf_inner_loop           10  4.0    0.003    0.004   29.592   29.594
 qs_scf_new_mos_kp                   10  5.0    0.000    0.000   20.642   21.165
 do_general_diag_kp                  10  6.0    0.013    0.016   20.642   21.165
 qs_ks_update_qs_env                 10  5.0    0.000    0.000    0.724    9.536
 kpoint_operator_prepare             11  7.0    0.001    0.001    8.935    9.458
 rebuild_ks_matrix                   10  6.0    0.000    0.000    0.609    9.421
 qs_ks_build_kohn_sham_matrix        10  7.0    0.004    0.006    0.609    9.421
 sum_up_and_integrate                10  8.0    0.000    0.000    0.559    9.355
 integrate_v_rspace                  10  9.0    0.001    0.009    0.558    9.354
 mp_min_i                            32  8.3    8.814    9.337    8.814    9.337
 grid_integrate_task_list            10 10.0    0.521    9.325    0.521    9.325
 transfer_rs2pw                      56  8.0    0.000    0.000    8.781    9.319
 mp_sendrecv_dv                    1288 10.0    8.777    9.316    8.777    9.316
 transfer_rs2pw_180                  12  8.8    0.002    0.003    8.776    9.314
 qs_rho_update_rho_low               11  5.0    0.000    0.000    8.965    8.972
 calculate_rho_elec                  11  6.0    0.001    0.009    8.965    8.972
 density_rs2pw                       11  7.0    0.000    0.000    8.465    8.971
 grid_collocate_task_list            11  7.0    0.498    8.905    0.498    8.905
 kpoint_density_transform            10  7.0    0.000    0.001    7.946    7.976
 kpoint_density_transform_group      10  8.0    4.492    5.181    7.946    7.976
 mp_sum_dm                         7748  9.0    1.546    2.914    1.546    2.914
 kpoint_operator_finish             473  7.0    0.001    0.001    2.359    2.901
 kpoint_operator_rs_build           473  8.0    2.358    2.899    2.358    2.899
 kpoint_set_mo_occupation            10  7.0    0.002    0.003    0.701    1.872
 mp_sum_dm3                          10  8.0    0.699    1.870    0.699    1.870
 copy_fm_to_dbcsr                  7740  9.0    0.031    0.039    1.120    1.256
 dbcsr_complete_redistribute       7740 10.0    0.126    0.221    0.828    1.050
 cp_fm_start_copy_general          7740  9.0    0.067    0.393    0.647    0.940
 init_scf_run                         1  3.0    0.000    0.000    0.842    0.846
 scf_env_initial_rho_setup            1  4.0    0.000    0.000    0.842    0.846
 cp_cfm_geeig                       430  7.0    0.004    0.005    0.618    0.778
 mp_alltoall_i22                   7740 11.0    0.462    0.684    0.462    0.684
```

Growl1234:
Is there still issue in this PR that I need to address?

DCM-Uni-Paderborn:
Dear Growl,

thank you for this substantial and useful optimization. I have looked through the implementation and the available CI results. Overall, the approach makes sense to me, and I did not identify an obvious correctness problem. In particular, the complete two-rank pdbg run with 6220/6220 passing tests is reassuring.

Before approving, I would only suggest two focused checks:

1. The PR description says that the grouped density FFT is valid only with FFTW, while the implementation also appears to allow padded transforms with other backends such as FFTSG. Could you please clarify this? If FFTW is strictly required, the code should enforce this. Otherwise, a small unit test comparing an incomplete, weighted inverse FFT against the direct phase sum would resolve the concern.

2. The most important new MPI topology is multiple k-point groups with more than one MPI rank per group. This cannot be covered by the two-rank regtest. Could you please run the existing four-MPI-rank psmp regtest, which is currently skipped?

These are intended as focused validation requests, not as a request for a broader restructuring. The overall direction looks very promising.

Greetings,  
Thomas

Growl1234:
> 1. The PR description says that the grouped density FFT is valid only with FFTW, while the implementation also appears to allow padded transforms with other backends such as FFTSG. Could you please clarify this? If FFTW is strictly required, the code should enforce this. Otherwise, a small unit test comparing an incomplete, weighted inverse FFT against the direct phase sum would resolve the concern.

If using FFTSG the added FFT step will simply be skipped, causing no corresponding performance improvement.

Growl1234:
> 1. The PR description says that the grouped density FFT is valid only with FFTW, while the implementation also appears to allow padded transforms with other backends such as FFTSG.

FFTSG can still use a padded transform; FFTW is needed for arbitrary-length exact grids.

Growl1234:
The four-MPI-rank psmp regtest is done; only one wrong result that's not related to this PR.

RTChou67:
This `1E-11` tolerance seems too strict to me. Other meta-GGA tests in the same folder use tolerances up to `1E-6`, while the SCF threshold here is `1E-8` and the TDDFPT convergence threshold is `1E-7`. It might make sense to loosen this tolerance.

Growl1234:
> This `1E-11` tolerance seems too strict to me. Other meta-GGA tests in the same folder use tolerances up to `1E-6`, while the SCF threshold here is `1E-8` and the TDDFPT convergence threshold is `1E-7`. It might make sense to loosen this tolerance.

For now I'd prefer not to relax the tolerance. And there's also a slow test; I think we should handle that directory seperatedly.

Wrt. this PR it can be merged.

fstein93 (approved):

PR #5995, opened 2026-09-09
Bump versions in docs/requirements.txt

he-zilong:
As suggested [here](https://github.com/cp2k/cp2k/issues/4972#issuecomment-5511906698).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -24,6 +24,7 @@ extensions = [
 ]
 
 myst_heading_anchors = 3
+myst_links_external_new_tab = True
 
 myst_enable_extensions = [
     "attrs_inline",
@@ -65,6 +66,8 @@ html_context = {
 
 html_theme_options = {
     "collapse_navigation": False,
+    "prev_next_buttons_location": "both",
+    "style_external_links": True,
 }
 
 redirects = {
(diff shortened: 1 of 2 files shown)


he-zilong:
I would also like to introduce two little adjustments here:
- The `Prev` and `Next` buttons for navigation now appear at the top as well as bottom;
- All external links are marked with an icon (defined by `sphinx-rtd-theme`) and will open as a new tab instead of replacing the current one (defined by `myst-parser`).

oschuett (approved):

oschuett:
It seems this also upgraded MathJax, which now complains about `Erroneous nesting of equation structures`:
- https://manual.cp2k.org/trunk/methods/properties/optical/index.html
- https://manual.cp2k.org/trunk/methods/properties/optical/bethe-salpeter.html
- ... (possibly others pages)

he-zilong:
> It seems this also upgraded MathJax, which now complains about `Erroneous nesting of equation structures`:

Another error is found at https://manual.cp2k.org/trunk/methods/dft/cneo.html.

Edit: the three cases have only `\begin{align}` or `\begin{multline}` in the markdown source and some lines are rather long, but in the rendered page (right click on an equation-> Show Math As -> TeX Commands) the lines are split and an extra layer of `\begin{split}` wrapping them are added. The relevant documentation [Automatic Line Breaking](https://docs.mathjax.org/en/v4.1/output/linebreaks.html) does not mention this, though.

PR #5996, opened 2026-09-09
Update precommit environment

Growl1234:


RTChou67:
How about moving requirements.txt to pyproject.toml?

Growl1234:
I went a look and it seems not as clean as `requirements.txt` for such use case. Therefore, I am not in favour of it.

oschuett (approved):

PR #5997, opened 2026-09-10
Remove duplicate and erroneous periodic RPA density initialization

RTChou67:
Fix the spin index used when initializing temporary density matrices in the periodic imaginary-time RPA path.
In `compute_transl_dm`, `mat_dm_global_work` is initialized by interating over `jspin`, but the matrix passed to `dbcsr_set` was indexed with `ispin`. This is a bug.

### DIFF
--- a/src/rpa_im_time.F
+++ b/src/rpa_im_time.F
@@ -1059,8 +1059,6 @@ CONTAINS
 
             CALL dbcsr_reserve_all_blocks(mat_dm_global_work(jspin, i_img)%matrix)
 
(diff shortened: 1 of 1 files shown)


fstein93:
Can you prepare a regtest for this bug?

RTChou67:
This bug would not cause any difference in observables. There will be a second initialization in `density_matrix_from_kp_to_transl`.

RTChou67:
Oh, so maybe, we could delete one initialization. I'd prefer the first time. What do you think?

fstein93:
Then, we could remove the statement.

JWilhelm (approved):

PR #5998, opened 2026-09-10
Misc convention and fortitude fixes

he-zilong:


fstein93:
`twopi` can be taken form `mathconstants`

src/common/mathlib.F
```diff
@@ -1705,7 +1706,10 @@ SUBROUTINE diag(n, a, d, v)
             a_max = MAX(a_max, MAXVAL(ABS(a(ip, ip + 1:n))))
          END DO
       END DO
-      WRITE (*, '(/,T2,A,/)') 'Too many iterations in jacobi'
+      WRITE (max_iter, *) maxit
```
oschuett, line 1709: You can use `integer_to_string` from [string_utilities.F](https://github.com/cp2k/cp2k/blob/master/src/common/string_utilities.F) instead.

oschuett (approved):

src/common/mathlib.F
he-zilong: I thought about that, but later figured that the simple and equivalent `WRITE` construct could avoid introducing a new module of dependency.

PR #6002, opened 2026-09-11
Compute momentum matrix option in qs_moment_kpoints_deep

Clemens-Schneider:
Changed qs_moment_kpoints_deep to include a flag that allows to calculate off diagonal momentum matrix elements directly instead of the dipole matrix elements. Updated Floquet files to use this feature.

### DIFF
--- a/src/floquet_main.F
+++ b/src/floquet_main.F
@@ -28,8 +28,8 @@ MODULE floquet_main
                                               floquet_env_type
    USE floquet_utils,                   ONLY: &
         build_floquet_matrix, calculate_floquet_observables, check_floquet_convergence, &
-        compute_e_k_de_dk_dipole, distribute_floquet_kp_data, floquet_sector_weights, &
-        make_floquet_subgroups, write_floquet_header, write_floquet_results
+        compute_e_k_de_dk_offdiagonal_momentum, distribute_floquet_kp_data, &
+        floquet_sector_weights, make_floquet_subgroups, write_floquet_header, write_floquet_results
    USE kinds,                           ONLY: dp
    USE mathconstants,                   ONLY: z_zero
    USE message_passing,                 ONLY: mp_para_env_release,&
@@ -61,7 +61,7 @@ CONTAINS
       CHARACTER(LEN=*), PARAMETER                        :: routineN = 'floquet'
 
       COMPLEX(KIND=dp), ALLOCATABLE, &
-         DIMENSION(:, :, :, :, :)                        :: dipole
+         DIMENSION(:, :, :, :, :)                        :: offdiagonal_momentum
       INTEGER                                            :: handle, ikp, ispin, my_group, &
                                                             n_est_done, n_source_done, ngroups
       INTEGER, ALLOCATABLE, DIMENSION(:)                 :: group_distribution
@@ -99,7 +99,7 @@ CONTAINS
       !   k-derivatives    ∇_k ε_nk = C_n(k)^† ∇_k H(k) C_n(k) - ε_nk C_n(k)^† ∇_k S(k) C_n(k)
       !   dipoles          d_nm(k) = <ψ_nk| r |ψ_mk>
       ! Results are distributed across ranks (k-point ikp stored on rank MOD(ikp-1,num_pe)).
-      CALL compute_e_k_de_dk_dipole(qs_env, bs_env, e_k, de_dk, dipole)
+      CALL compute_e_k_de_dk_offdiagonal_momentum(qs_env, bs_env, e_k, de_dk, offdiagonal_momentum)
 
       ! Split the global communicator into Floquet subgroups (Hamiltonian distributed within each
       ! subgroup, k-points distributed across subgroups) and create the subgroup BLACS context.
@@ -124,7 +124,7 @@ CONTAINS
             ! Broadcast this k-point/spin's ε_nk, ∇_k ε_nk and d_nm(k) to every rank of the
             ! owning subgroup (its owner rank holds them from the distributed precompute above).
             CALL distribute_floquet_kp_data(para_env, para_env_sub, ispin, ikp, e_k, de_dk, &
-                                            dipole, floquet_env)
+                                            offdiagonal_momentum, floquet_env)
 
             ! Assemble H_F(k) in Sambe space (blocks m,m' = -M..M, each nao×nao)
             !  diagonal      (H_F)^{m,m}_{nn'}   = (ε_nk + m ħΩ) δ_{nn'}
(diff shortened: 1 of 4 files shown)


src/floquet_main.F
```diff
@@ -99,7 +99,7 @@ SUBROUTINE floquet(qs_env, bs_env)
       !   k-derivatives    ∇_k ε_nk = C_n(k)^† ∇_k H(k) C_n(k) - ε_nk C_n(k)^† ∇_k S(k) C_n(k)
       !   dipoles          d_nm(k) = <ψ_nk| r |ψ_mk>
```
JWilhelm, line 100: Can you please update the following comment?

!   dipoles          d_nm(k) = <ψ_nk| r |ψ_mk

src/floquet_main.F
```diff
@@ -99,7 +99,7 @@ SUBROUTINE floquet(qs_env, bs_env)
...
       ! Results are distributed across ranks (k-point ikp stored on rank MOD(ikp-1,num_pe)).
-      CALL compute_e_k_de_dk_dipole(qs_env, bs_env, e_k, de_dk, dipole)
+      CALL compute_e_k_de_dk_offdiagonal_momentum(qs_env, bs_env, e_k, de_dk, offdiagonal_momentum)
```
JWilhelm, line 102: the name "offdiagonal_momentum" is a bit long compared to e_k and de_dk. What would you think about 

CALL compute_e_k_de_dk_p_k ?

What would you think about also renaming the variable "offdiagonal_momentum" as p_k? To make clear it is only the off-diagonal momentum matrix elements, you could adjust the comment in line R100 (see my comment above)

src/qs_moments.F
```diff
@@ -2964,12 +2964,14 @@ END SUBROUTINE build_local_moment_matrix_rs_img
 !> \param berry_c berry curvature calculated using Ω^γ_n = Σ_m 2*Im[d^α_nm (d^β_mn)*]
 !> \param do_parallel option to distribute the result in dipole across
 !>        different MPI ranks
+!> \param compute_momentum option to calculate the momentum matrix elements
+!>        instead of the dipole matrix elements
 !> \author Shridhar Shanbhag
 ! **************************************************************************************************
-   SUBROUTINE qs_moment_kpoints_deep(qs_env, xkp, dipole, rcc, berry_c, do_parallel)
+   SUBROUTINE qs_moment_kpoints_deep(qs_env, xkp, dipole, rcc, berry_c, do_parallel, compute_momentum)
```
JWilhelm, line 2971: This construction might be not intuitive to understand, because you write a "momentum" into the "dipole". Maybe it would be more clear if you just add after "dipole" the "momentum" variable and you always compute the momentum, so no flag "compute_momentum" would be necessary. The additional computational cost and additional memory would be negligible, and it would be easy to understand what qs_moment_kpoints_deep is doing.

JWilhelm (changes requested):

src/floquet_utils.F
```diff
@@ -230,13 +230,13 @@ END SUBROUTINE calculate_epsilon_derivative
 !> \brief Momentum matrix elements p_nm(k) for one k-point and one spin.
 !> \param e_k_kp_spin ...
 !> \param de_dk_kp_spin ...
-!> \param dipole_kp_spin ...
+!> \param momentum_kp_spin ...
 !> \param momentum ...
 ! **************************************************************************************************
-   SUBROUTINE build_momentum_matrix(e_k_kp_spin, de_dk_kp_spin, dipole_kp_spin, momentum)
+   SUBROUTINE build_momentum_matrix(e_k_kp_spin, de_dk_kp_spin, momentum_kp_spin, momentum)
```
ShridharShanbhag: We no longer need e_k_kp_spin here.
Having two different objects called momentum doesn't make sense. Have a single object with intent(inout) and update it in this repository.

src/qs_moments.F
```diff
@@ -3121,6 +3125,11 @@ SUBROUTINE qs_moment_kpoints_deep(qs_env, xkp, dipole, rcc, berry_c, do_parallel
                      dip(i_dir, n, m) = -gaussi*C_dH_C(n, m)/(eigenvals(n) - eigenvals(m)) &
                                         + gaussi*eigenvals(n)*C_dS_C(n, m)/(eigenvals(n) - eigenvals(m)) &
                                         + CDC(n, m)
```
RTChou67, line 3127: This statement divides by `eigenvals(n) - eigenvals(m)`. This should be avoided for degenerate or near-degenerate systems. Also, this is immediately overwritten in line 3130 when 
`compute_momentum=.TRUE.`

src/qs_moments.F
```diff
@@ -3006,8 +3008,10 @@ SUBROUTINE qs_moment_kpoints_deep(qs_env, xkp, dipole, rcc, berry_c, do_parallel
       CALL timeset(routineN, handle)
       calc_bc = PRESENT(berry_c)
       my_do_parallel = .FALSE.
+      my_compute_momentum = .FALSE.
       my_rcc = 0.0_dp
       IF (PRESENT(do_parallel)) my_do_parallel = do_parallel
+      IF (PRESENT(compute_momentum)) my_compute_momentum = compute_momentum
```
RTChou67, line 3014: `berry_c` is documented and implemented in terms of dipole matrix elements, but with `compute_momentum=.TRUE.` the local `dip` array contains momentum matrix elements instead. This would result in wrong results silently when calling with both optional arguments.

src/floquet_utils.F
```diff
@@ -508,12 +507,12 @@ SUBROUTINE compute_e_k_de_dk_dipole(qs_env, bs_env, e_k, de_dk, dipole)
       ! One k-point per rank, round-robin by global rank
       ! So the results for ikp are stored in mepos==MOD(ikp-1,num_pe)
       CALL calculate_epsilon_derivative(qs_env, xkp_all, e_k=e_k, de_dk=de_dk, do_parallel=.TRUE.)
-      CALL qs_moment_kpoints_deep(qs_env, xkp_all, dipole, do_parallel=.TRUE.)
+      CALL qs_moment_kpoints_deep(qs_env, xkp_all, offdiagonal_momentum, do_parallel=.TRUE., compute_momentum=.TRUE.)
```
RTChou67, line 511: `calculate_epsilon_derivative` and `qs_moment_kpoints_deep` contains duplicate code. Both construct the real-space H/S arrays, transform them to H(k)/S(k), and solve the same generalized eigenvalue problem for every spin and k-point. Please introduce a shared subroutine.

RTChou67 (changes requested):

src/qs_moments.F
Clemens-Schneider: added an eps_degenerate to prevent this problem.

src/qs_moments.F
Clemens-Schneider: included a seperate variable to store the momentum.

src/floquet_utils.F
Clemens-Schneider: i removed calculate_epsilon_derivative entirely, instead qs_moment_kpoints_deep has an option to calculate the momentum and band energy.

ShridharShanbhag (approved):

JWilhelm (approved):

JWilhelm:
Looks great!! I wait until tomorrow to merge to give others the chance to comment.

PR #6007, opened 2026-09-12
Screen unsupported periodic images in native atom-grid reconstruction

DCM-Uni-Paderborn:
## Summary

Reduce unnecessary periodic-image traversal in native atom-composite reconstruction and its force/stress and adjoint paths. This is independent of #6006 and applies directly to master.

- Bound candidate images using `|(H^-1 d)_i| <= R ||(H^-1)_i||`, with conservative floating-point padding and intersection with the original image shell.
- Apply the existing Cartesian support-sphere test before allocating/interpolating cross-atom contributions.
- Retain the existing density-dependent support radius, recomputed on every call. Retained images keep their original order. No model, grid, threshold or physical cutoff changes.
- Use the same bounds for forward reconstruction, force/stress derivatives and the adjoint.

The affected code is the native atom-composite reconstruction path, not every GAPW calculation.

## Validation

The standalone adjoint checker extracts the actual production interpolation routines and loop bodies. It covers one/two spin channels, all eight periodicity masks, empty local ownership, orthogonal/triclinic/skewed cells, changing radii, large translations and support-boundary rounding. GNU bounds/FPE checks and OpenMP 1/2/4/8/16-thread comparisons pass. Maximum relative adjoint discrepancy is about `5.2e-16`.

The image-bound fixture reduces 240944 candidate images to 37464 while retaining all 6924 images inside the existing support sphere. This is an image-enumeration result, not a whole-SCF speedup. The interpolation-dominated standalone adjoint timing was essentially unchanged.

Independent full CP2K baseline and patched executables were compared for four small native-Skala cases: periodic mixed AE/GTH H2, a sheared cell, a displaced atom, and all-electron H2. Each ran three fixed SCF iterations and evaluated total forces and analytical stress. These are equivalence tests, not converged physical benchmarks or finite-difference validation.

- Maximum absolute total-energy difference: `1.11e-15 Ha`.
- Printed total forces: identical.
- Maximum analytical-stress difference: `1.01e-13 GPa`.
- All executions ended normally with exit status zero.

All source files pass the repository pre-commit checks. Ongoing scientific production used separate unchanged binaries and inputs.

Run the standalone checker using `tools/regtesting/check_gapw_atom_adjoint.py` with a completed GNU/OpenMP CMake build. It records its source hashes, checks and timings in its work directory.

### DIFF
--- a/tools/regtesting/check_gapw_atom_adjoint.py
+++ b/tools/regtesting/check_gapw_atom_adjoint.py
@@ -52,6 +52,7 @@ WRAPPER = """
  REAL(dp), ALLOCATABLE :: vxc_h_local(:, :, :), vxc_s_local(:, :, :), vtau_h_local(:, :, :), vtau_s_local(:, :, :)
  REAL(dp), ALLOCATABLE :: vxg_h_local(:, :, :, :), vxg_s_local(:, :, :, :)
  INTEGER :: base_shift(3), composite_row, idir, image_i1, image_i2, image_i3
+ INTEGER :: image_lower(3), image_upper(3)
  INTEGER :: image_shift(3), image_shell(3), jdir, target_atom, iatom
  INTEGER :: composite_local_atom, composite_local_natom, composite_nflat
  REAL(dp) :: cross_density_adjoint(2), cross_grad_adjoint(3, 2), cross_kin_adjoint(2)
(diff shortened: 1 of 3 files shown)


tools/regtesting/check_gapw_atom_adjoint.py
oschuett, line 2: Since this is not run by the CI, it will bit rot quickly. Perhaps it could be turned into a unittest that compares checksums?

PR #6011, opened 2026-09-12
Fix some regtests, revamp installation docs and scripts

he-zilong:
While trying to install CP2K with `make_cp2k.sh` I found that:

- Some regtest directories require `libxc` for (r2)SCAN functionals but are not registered as such;
- There is no instruction on `build-from-source.md` about how to launch the CP2K (which is confusing since the toolchain and Spack-based `make_cp2k.sh` are quite different).


docs/getting-started/build-from-source.md
```diff
@@ -120,22 +120,50 @@ After that, to build and install CP2K linked against them, run:
...
 of the dependencies are in `tools/toolchain/install/`, the CP2K build tree in `build/`, and the
-headers, modules, binary executables and a `cp2k_env` file in `install/`. (Before running CP2K, one
-must always source the `cp2k_env` file.) Options are also available to make installed dependencies
-and program outside of the source tree.
+headers, modules, binary executables and a `cp2k_env` file in `install/`. Options are also available
+to make installed dependencies and program outside of the source tree.
 
 ```shell
 ./install_cp2k_toolchain.sh --install-dir=/opt/cp2k/toolchain
 ./build_cp2k.sh --prefix /opt/cp2k
```
Growl1234: Maybe, it's worth including a native preset option to this example.

he-zilong:
That the Spack-based `make_cp2k.sh` and post-toolchain `build_cp2k.sh` behaves differently can be baffling. For instance, `build_cp2k.sh` creates `cp2k_env` whereas `make_cp2k.sh` creates `/bin/launch`:
https://github.com/cp2k/cp2k/blob/e75382876fc6fbb1987b2ecbf84091dcb5abea50/tools/toolchain/build_cp2k.sh#L373-L379
https://github.com/cp2k/cp2k/blob/e75382876fc6fbb1987b2ecbf84091dcb5abea50/make_cp2k.sh#L1788-L1805
`make_cp2k.sh` goes a few steps further to create several links while `build_cp2k.sh` does not:
https://github.com/cp2k/cp2k/blob/e75382876fc6fbb1987b2ecbf84091dcb5abea50/make_cp2k.sh#L1750-L1759

he-zilong:
I discovered that there is a hidden auto-generated [genindex](https://manual.cp2k.org/cp2k-2026_2-branch/genindex.html) (link is to the 2026.2 version) which can presumably be suppressed with Sphinx config `html_use_index = False`. I feel that the genindex page is too lengthy to display and navigate comfortably for our manual.

docs/getting-started/build-from-source.md
```diff
@@ -297,50 +338,66 @@ example is given in the header of that
...
 - `-DCMAKE_BUILD_TYPE` Valid vaules are `Release` (default) and `Debug` (enables debug settings and
   generates `pdbg` or `sdbg` instead of `psmp` or `ssmp`; recommended for development).
-- `-DCMAKE_INSTALL_PREFIX` Specifies the installation path of CP2K. Assuming it is set to
-  `/path/to/installation`, there will be several subdirectories: `bin` for binaries like
-  `cp2k.psmp`, `include` for module files and headers, `lib` or `lib64` for libraries, and `share`
-  for some other files such as basis data. Default is `/usr/local`.
-- `-DBUILD_SHARED_LIBS` Specifies if shared libraries are built. Default is `ON`; if set `OFF`, a
-  static library will be built instead.
+- `-DCMAKE_INSTALL_PREFIX` Specifies the installation path.
+- `-DCMAKE_INSTALL_LIBDIR` Specifies the library path under the installation path. Default is `lib`.
```
Growl1234: No, the default is not `lib`. It's up to CMake's `GNUInstallDirs` module and depends on system and architecture.

I'd suggest not to include this description to our documentation.

tools/toolchain/build_cp2k.sh
```diff
@@ -350,7 +352,7 @@ EOF
     fi
 
     set -o pipefail
-    cmake -S "${CP2K_ROOT}" -B "${BUILD_DIR}" "${CMAKE_PRESET_ARGS[@]}" \
+    cmake -S "${CP2K_ROOT}" -B "${BUILD_DIR}" --preset "${CMAKE_PRESET}" \
```
Growl1234, line 354: Did you test whether a blank preset (even not `none`) worked?

tools/toolchain/build_cp2k.sh
```diff
@@ -360,6 +362,16 @@ EOF
   cmake --build "${BUILD_DIR}" --target install -j "${BUILD_JOBS}" 2>&1 | tee -a build.log
   echo "=========================================================="
 
+  log_build "=============== Creating links to binaries ==============="
+  VERSION="psmp"
```
Growl1234: It also has `ssmp`, right?

docs/getting-started/build-from-source.md
he-zilong: Both the toolchain `build_cp2k.sh` and Spack `make_cp2k.sh` specify explicitly that `-DCMAKE_INSTALL_LIBDIR=lib`. I can fix the description of the default, but I do not want to hide this piece of information.

tools/toolchain/build_cp2k.sh
```diff
@@ -360,6 +362,16 @@ EOF
...
+  cd "${CMAKE_INSTALL_PREFIX}"/bin
+  for binary in *."${VERSION}"; do
+    ln -sfv "${binary}" "${binary%%."${VERSION}"}"
```
Growl1234: And to be honest, I don't like the idea of creating every softlink without suffix. IMHO the suffix `psmp`/`ssmp` is simple enough. I believe, `spack install cp2k` and other package managers also do not bring them (only `popt` and `sopt` that are defined in CMake).

tools/toolchain/build_cp2k.sh
he-zilong: In `build_cp2k.sh` it is already guarded by the logic
```shell
      if [ -z "${CMAKE_PRESET}" ]; then
        report_error "A preset name must be provided to --preset."
      fi
```

tools/toolchain/build_cp2k.sh
he-zilong: This is to make the toolchain `build_cp2k.sh` consistent with the Spack `make_cp2k.sh`. Either both define them, or neither.

docs/getting-started/build-from-source.md
Growl1234: The reason why they specify `-DCMAKE_INSTALL_LIBDIR=lib` is mainly for convenience of the environment file (toolchain's `cp2k_env` and `make_cp2k.sh`'s `launch`). Usually users are not responsible to take care of this variable, and for package managers (`spack install cp2k`, Fedora, etc.) we must respect the `GNUInstallDirs` behaviour anyways.

tools/toolchain/build_cp2k.sh
Growl1234: Then please change the guideline in `install_cp2k_toolchain.sh` or set a default value of preset like `make_cp2k.sh`. Personally I would prefer the latter.

he-zilong: Indeed the epilogue in `install_cp2k_toolchain.sh` has not suggested running `build_cp2k.sh` with an explicit `--preset` yet. Would it be possible to give an suggestion based on `install_cp2k_toolchain.sh` options? Say `--with-gcc=system` and `--target-cpu=native` on the x86_64 architecture adds automatically `--preset=native-gnu-x86_64`.

Growl1234: I'm fine with this idea. Architecture is detected via [`get_openblas_arch.sh`](https://github.com/cp2k/cp2k/tree/8ec0769333b3ff12d5b60ae851ae208758e6fbb2/tools/toolchain/scripts/get_openblas_arch.sh), and target CPU is user-determined and default is `native`.

For `--target-cpu=generic` just give `none` or omit the `--preset` flag:
```bash
if [ -z "${CMAKE_PRESET}" ]; then
  report_warning "Preset name is not provided; assuming to \"--preset none\"."
  CMAKE_PRESET="none"
fi
```

tools/toolchain/build_cp2k.sh
Growl1234: I think, maybe the intention of two scripts has some difference? `make_cp2k.sh` takes care of both installation and development integration, while `build_cp2k.sh` only targets user-facing installation workflow. I don't know what does `make_cp2k.sh` want to do via those symlinks, but for toolchain I'd prefer to interfere as little as possible with the native CMake install. @mkrack What do you think?

tools/toolchain/build_cp2k.sh
```diff
@@ -83,10 +83,15 @@ while [ $# -ge 1 ]; do
...
-        report_error "A preset name must be provided to --preset."
-      fi
+      case "${1#*=}" in
+        native-gnu-x86_64 | native-gnu-arm64 | native-intel | none)
+          CMAKE_PRESET="${1#*=}"
+          ;;
+        *)
+          report_warning "Unknown CMake preset: ${1#*=}. Default to \"none\"."
+          CMAKE_PRESET="none"
+          ;;
```
Growl1234: But it also has `conventions`, `coverage`, etc. Although this script doesn't follow up to them by corresponding testing behavior, we shouldn't block them or redirect them to `none` here. And it's not a good idea to list available presets in multiple places as `CMakePresets.json` may change.

My wish would be to only default to `none` if `--preset` flag is not set by users. An unknown preset will cause CMake to abort early, so we don't need to take care of it here.

```
root@linux:~/GitHub/cp2k-dev# cmake -Bbuild --preset hf
CMake Error: No such configure preset in /root/GitHub/cp2k-dev: "hf"
Available configure presets:

  "native-gnu-x86_64" - GNU native build for x86_64
  "native-gnu-arm64"  - GNU native build for Arm64
  "native-intel"      - Intel native build
  "conventions"       - GNU conventions check
  "coverage"          - Coverage build
  "asan"              - AddressSanitizer build
  "none"              - No preset
```

he-zilong: Again, this is aligning `build_cp2k.sh` with the `make_cp2k.sh` behavior; I am also not really in favor of it and am happy to revise this logic for **both** scripts.

Growl1234: Hmm, `make_cp2k.sh` uses `--test_conventions` etc specifically; they not only choose preset but also perform those checks, which is already integrated in CI also. I don't want to make `build_cp2k.sh` "kitchen-sink". My request is simple: let's not prevent other presets from being chosen.

he-zilong: Understood. I will be back working on this a little later.

tools/toolchain/build_cp2k.sh
mkrack: The symlinks without an extension are for convenience mainly and provide a default to users who often wonder about the meaning of the various extensions and thus which binary to choose. I like the possibility to run CP2K simply with `cp2k -i ...`, but I am glad to learn about any concerns or technical problems caused by this.

he-zilong: There are numerous `*_unittest.${version}` binary executables that do not have to get the extension-free symlinks in my opinion (and I also think that they should be put under a `unittest` subdirectory for better clarity). The other utility programs and of course CP2K itself would benefit from getting the simplified symlink.

In anycase, let's implement these in CMake rather than post-processing shell scripts `build_cp2k.sh` or `make_cp2k.sh`.

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -1530,16 +1525,46 @@ else
...
+    OPENBLAS_ARCH="unknown"
+    load "${BUILDDIR}/openblas_arch"
+    if [ "${with_gcc}" != "__DONTUSE__" ]; then
+      if [ "${OPENBLAS_ARCH}" = "x86_64" ]; then
+        preset_option="--preset native-gnu-x86_64"
+      elif [ "${OPENBLAS_ARCH}" = "arm64" ]; then
+        preset_option="--preset native-gnu-arm64"
+      else
+        report_warning ${LINENO} "${OPENBLAS_ARCH} is not a known architecture
+for GNU compiler; falling back to --preset none for build_cp2k.sh."
```
Growl1234:
```suggestion
for GNU compiler; falling back to \"--preset none\" for build_cp2k.sh."
```

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -1530,16 +1525,46 @@ else
...
+      fi
+    elif [ "${with_intel}" != "__DONTUSE__" ]; then
+      preset_option="--preset native-intel"
+    elif [ "${with_amd}" != "__DONTUSE__" ]; then
+      report_warning ${LINENO} "Native preset for AMD compiler is not yet
+implemented; falling back to --preset none for build_cp2k.sh."
```
Growl1234:
```suggestion
implemented; falling back to \"--preset none\" for build_cp2k.sh."
```

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -1530,16 +1525,46 @@ else
...
+    else
+      report_error ${LINENO} "Unknown compiler choice for --target-cpu=native."
+    fi
+  else
+    report_warning ${LINENO} "Non-native target CPU is used; falling back to
+--preset none for build_cp2k.sh."
```
Growl1234:
```suggestion
    echo "Note: Non-native target CPU is used; using \"--preset none\" for build_cp2k.sh."
```

tools/toolchain/build_cp2k.sh
```diff
@@ -123,6 +117,11 @@ Please run this script without "--rebuild-only" if you want to change CMake sett
 EOF
     exit 1
   fi
+else
+  if [ "${PRESET_SET}" = "__FALSE__" ]; then
+    report_warning "No CMake preset is specified; default to \"none\"."
+    CMAKE_PRESET="none"
+  fi
```
Growl1234:
```suggestion
elif [ "${PRESET_SET}" = "__FALSE__" ]; then
  report_warning "No CMake preset is specified; default to \"none\"."
  CMAKE_PRESET="none"
```

Growl1234 (approved):

he-zilong:
This PR is ready to go; remaining discussions about `make_cp2k.sh` and `build_cp2k.sh` can be relegated to further discussions.

oschuett:
>... I feel that the genindex page is too lengthy to display and navigate comfortably for our manual.

I agree, but before we disable it, we should double check that it's not needed for the builtin search.

mkrack (approved):

tools/toolchain/build_cp2k.sh
Growl1234: > and I also think that they should be put under a `unittest` subdirectory for better clarity

Please don't if it can be avoided, otherwise it will break `export PATH=/path/to/cp2k/bin:$PATH` behavior and require users and downstream consumers to add more paths to their environment variables.

PR #6013, opened 2026-09-12
Minor fix in cell optimization

Growl1234:
- Correctly handle empty and padded local blocks in `cp_fm_matvec()`;
- Transform `CELL_OPT` ionic gradients with the transpose Jacobian and include independent shell coordinates;
- Use the existing immutable coordinate/force snapshots during space-group projection, avoiding order dependence and OpenMP read/write races.


he-zilong:
What is shell coordinates, how is it defined differently from atomic/core coordinates, and when will `natom` be not the same as `nparticle`? I have been puzzled for a while...

Growl1234:
`natom` is the number of physical atoms, while `nparticle` is the number of independent particle coordinates packed for optimization. For a normal atom these are the same. In the core–shell model, however, an atom is represented by independent core and shell coordinates: the core occupies the atom's normal slot and the shell is appended as an extra particle, so `nparticle = natom + nshell`. The atomic position itself is the mass-weighted core/shell position. Therefore the packed `3*nparticle` gradient contains additional shell DOFs, and transforming only the first `natom` triplets leaves those shell gradients in the wrong cell coordinate frame.

he-zilong:
Please write about the core-shell model somewhere in the manual. Also, please add the fix for shell coordinates to the changelog as it seems to affect more than a few results.

Growl1234:
Core-shell model is not used by QS/DFT but Fist and QM/MM, so I think it's best to add the page under QM/MM.

For changelog, I will update it 1-2 times per month; or, just put it in #6017 with two items.

fstein93 (approved):

oschuett:
> Please write about the core-shell model somewhere in the manual. 

We already got a placeholder page for [polarizable force fields](https://manual.cp2k.org/trunk/methods/qm_mm/polarizable_force_field.html).

Growl1234:
Unfortunately this PR seems to have brought some numerical stability regression. I'll try to handle them, and if needed, revert it.

PR #6017, opened 2026-09-13
BFGS: Add inverse-Hessian update method

Growl1234:
This PR adds an inverse-Hessian formulation to the existing BFGS optimizer. It directly updates the inverse Hessian and computes the search direction from it, avoiding the full Hessian diagonalization at every optimization step.

The new `BFGS%METHOD` keyword provides two choices:

- `DIRECT` (default): update and diagonalize the Hessian, preserving the current behavior
- `INVERSE`: the new formulation introduced here

The inverse update keeps the existing BFGS driver, trust-radius handling, joint ion/cell optimization variables, and model-Hessian initialization. Curvature damping is applied to keep the inverse-Hessian update well behaved.

BFGS restart files now record which matrix representation is stored. A restart written by either `DIRECT` or `INVERSE` can therefore be read by either method, with a one-time conversion when necessary. Legacy Hessian restart files remain supported and are interpreted as direct Hessians.

`USE_RAT_FUN_OPT` is not supported with `METHOD INVERSE`, since RFO explicitly requires the Hessian eigensystem.

A test matrix is added accordingly; reference values are tested with #6013 applied locally, so the wrong results (if happen) are expected before #6013 is merged.


src/motion/bfgs_optimizer.F
```diff
@@ -475,6 +527,379 @@ RECURSIVE SUBROUTINE geoopt_bfgs(force_env, gopt_param, globenv, geo_section, go
...
+         IF (MINVAL(eigenvalues) <= 0.0_dp) THEN
+            CALL cp_abort(__LOCATION__, &
+                          "BFGS inverse restart matrix is not positive definite")
+         END IF
+      ELSE
+         IF (MINVAL(eigenvalues) < 0.0001_dp) THEN
+            CALL cp_warn(__LOCATION__, &
+                         "BFGS: regularizing the direct restart before conversion to INVERSE.")
+         END IF
+         eigenvalues(:) = MAX(0.0001_dp, eigenvalues(:))
```
fstein93: I would convert this magic constant to a `PARAMETER`

s8ga:
Hey @Growl1234,

Great PR. Caught your note about damping with B*s in `bfgs_inverse_update` — are you planning to add Powell damping to the direct `bfgs()` update as well?

I already have a branch ready for direct damping + curvature guards. If you're planning to tackle it, happy to let you take the lead or coordinate; otherwise, I'd be glad to submit a follow-up PR once this lands.

For context, direct damping fixes a few edge cases in the current regtests:

- In `QS/regtest-gpw-4/H2-geo-1`, the current undamped update consumes a negative-curvature secant pair and the run ends 0.52 Ha abovewhere it started. Damping fixes this and descends to the physical state, 0.89 Ha below the undamped endpoint.

- In `QS/regtest-gpw-6-4/c8_broy`, $y \cdot s / (s \cdot B \cdot s)$ goes negative at the very first update (-0.96, and as low as -47.6 later). Damping keeps the trajectory stable and overtakes the undamped run over 50 steps (-0.19 Ha with tighter SCF).

Let me know what you think!

Growl1234:
Feel free to submit your PR.

Currently this PR is not ready as I'm struggling with the wrong results in psmp and pdbg CI tests, which I have never  been able to reproduce.

`Fist/regtest-opt/cell_opt_bfgs_inverse_vib.inp`

| Tester | Value | Status |
| ------- | -------- | ---------- |
| Regtest pdbg | 532.755422 | WRONG RESULT |
| Spack pdbg | 587.912968 | OK |
| Regtest psmp | 492.857801 | WRONG RESULT |
| Regtest sdbg | 587.912968 | OK |
| Regtest ssmp | 587.912968 | OK |

fstein93:
Have you already tried to run Docker locally or don't you have a machine with sudo-access locally available?

Growl1234:
> Have you already tried to run Docker locally or don't you have a machine with sudo-access locally available?

I built CP2K both pdbg and psmp locally, and it turns out the wrong results are happening randomly, hard to catch. I also tried some diagnosis attempts before finally being lucky to catch one.

RTChou67:
It seems that, MPI ranks would slightly affect the energy, stepsize and force. this might be some stuff's different approach when single MPI and multi MPI. Then it goes into the degenerate stuff. I suppose it is fixed in the newest commit, just by looking at the names. However I don't know how much  would the slight difference in energy and force affect M018 value.

Growl1234:
It appears to be a robustness issue in `mode_selective.F`. It comes from using individual eigenvectors to rank `BFGS_HESS` modes by their contribution on `INVOLVED_ATOMS`.

For degenerate or nearly degenerate eigenvalues, the eigenvectors are not unique: LAPACK/ScaLAPACK may return any orthogonal basis of the same eigenspace, depending on numerical details or parallel layout. As a result, the ranking can select different modes even when the Hessian and eigenspace are essentially identical.

The fix is to construct a deterministic basis within each degenerate eigenspace before applying the existing mode-selection criterion.

Growl1234:
Hi, is there still any blocker or suggestion?

fstein93 (approved):

PR #6021, opened 2026-09-14
Motion/BFGS: Damp direct Hessian update and guard against corruption

s8ga:
## Summary

CP2K's direct BFGS update runs without a line search, meaning nothing guarantees the positive curvature condition `y·s > 0`. When negative curvature occurs, the standard update can produce an indefinite Hessian and destabilize the optimization.

This PR applies Powell's damped BFGS update (Powell 1978, eq. 3.7):
When the measured curvature `y·s` drops below 20% of `s·B·s`, `y` is replaced by:
```
eta = theta * y + (1 - theta) * B * s,   where theta = 0.8 * s·B·s / (s·B·s - y·s)
```
This ensures `eta·s = 0.2 * s·B·s`, keeping the Hessian positive definite without requiring a line search.
When `y·s >= 0.2 * s·B·s`, `theta = 1` and the update is bit-identical to the existing formula.

When damping triggers, a message is printed to the log under default `MEDIUM` print level (`PROGRAM_RUN_INFO`):
```
BFGS| Damped update, curvature ratio y.s/(s.B.s):    -0.9601 (threshold 0.2)
```

Additional safeguards for secant pairs:
- Abort on non-finite (`NaN`/`Inf`) displacement or gradient differences.
- Skip the update if displacement or gradient change has zero norm.
- Reset the Hessian to the initial model (or identity for cell opt and core-shell) if `s·B·s <= EPSILON` or if the updated matrix contains non-finite entries.
- Consolidate all early-exit decisions into a single reduced flag (`para_env%max`) so all MPI ranks take the same path.

## Key Motivations

- `QS/regtest-gpw-4/H2-geo-1`: The first update encounters `y·s / (s·B·s) = -4.71`. The undamped update takes an unphysical 0.47 bohr step, increasing the energy by +0.52 Ha ("Decrease in energy NO"). Damping keeps the trajectory descending toward the physical state (-0.89 Ha lower).
- `QS/regtest-gpw-6-4/c8_broy_{,gapw_}gop`: Negative curvature appears on step 1 (-0.96 and -5.69) because the SCF is unconverged under `MAX_SCF 5`. Damping stabilizes the Hessian update.
- Outside the modified tests, damping also safely engages in silicon cell optimizations (`Si_direct_cell_opt`, `Si_sym_hexagonal_gamma_120`) and NEB band sub-optimizations without altering their final outputs.

## Reference Updates (14 tests)

No input files are changed. All 14 updated reference values trace directly to logged damping events, split across four categorized `Regtest:` commits:

1. **Repaired unphysical trajectory (1 test)**:
   `H2-geo-1` previously converged to an energy-increasing dissociated path; the updated reference reflects the descending trajectory.

2. **Converged runs with shifted endpoints (3 tests)**:
   `nh3_wat-restraint-fixd-3`, `nh3_wat-restraint-fixd-4`, and `TNF2_OPLS`.
   Damping triggers on shallow modes, landing at nearby admissible endpoints that satisfy all convergence criteria ($\Delta E \sim 10^{-6}$ to $10^{-5}$ Ha). Notably, `nh3_wat-restraint-fixd-3` previously failed to converge within 50 steps; with damping, it converges in 28 steps.

3. **Truncated run snapshots (8 tests)**:
   `HF-field-gopt`, `HF-loc-field-gopt`, `h2o_f04`, `h2o_f10b` (x2), `h2o_f11`, `argon05`, and `2gly_IT-NEB-CV-res`.
   These runs end at `MAX_ITER 2-3` (or an unconverged NEB band). Damping changes early step sizes, shifting the snapshot energy at step 2 or 3 ($\Delta E \sim 10^{-6}$ to $3.7\times 10^{-4}$ Ha).

4. **SCF-limited snapshots (2 tests)**:
   `c8_broy_gop` and `c8_broy_gapw_gop`.
   Both stop after 2 steps with unconverged SCF (`MAX_SCF 5`). Damping prevents storing negative curvature from SCF noise. Over 50 steps with converged SCF, the damped path yields lower energies on both variants.

| Test | First trigger ratio | Further triggers |
|---|---|---|
| H2-geo-1 | −4.71 | — |
| c8_broy_gop | −0.96 | −47.6 |
| c8_broy_gapw_gop | −5.69 | — |
| TNF2_OPLS | 0.13 | 0.02, 0.15 |
| nh3_wat-restraint-fixd-3 | 0.16 | — |
| B2 tests (8 total) | 0.05 – 0.19 | — |


src/motion/bfgs_optimizer.F
```diff
@@ -685,6 +676,13 @@ SUBROUTINE bfgs(ndf, dx, dg, hess_mat, work, para_env)
       CALL cp_fm_get_info(hess_mat, row_indices=row_indices, col_indices=col_indices, &
                           local_data=local_hes, nrow_local=nrow_local, ncol_local=ncol_local)
 
+      ! Refuse corrupted input instead of poisoning the Hessian with Inf/NaN.
+      IF (.NOT. ALL(ieee_is_finite(dx(1:ndf))) .OR. .NOT. ALL(ieee_is_finite(dg(1:ndf)))) THEN
+         CPABORT("Non-finite displacement or gradient difference")
+      END IF
+
+      use_model = PRESENT(force_env) .AND. .NOT. cell_opt .AND. .NOT. shell_present
```
Growl1234: This makes `USE_MODEL_HESSIAN` no longer a valid input option.

src/motion/bfgs_optimizer.F
```diff
@@ -694,26 +692,145 @@ SUBROUTINE bfgs(ndf, dx, dg, hess_mat, work, para_env)
...
    END SUBROUTINE bfgs
 
+! **************************************************************************************************
+!> \brief Build the direct initial Hessian: sanitized model Hessian or unity
+!> \param hess_mat Direct Hessian, replaced in place
+!> \param force_env Environment providing the model Hessian at the current geometry
+!> \param use_model Build the sanitized model Hessian; otherwise fall back to unity
+!> \note Must be called by all ranks together: construct_initial_hess reduces internally.
+! **************************************************************************************************
+   SUBROUTINE build_initial_hess(hess_mat, force_env, use_model)
```
Growl1234: There seems to be duplication; can we make it cleaner?

src/motion/bfgs_optimizer.F
```diff
@@ -694,26 +692,145 @@ SUBROUTINE bfgs(ndf, dx, dg, hess_mat, work, para_env)
...
+      sy = DDOT(ndf, dg, 1, dx, 1)
+      sbs = DDOT(ndf, dx, 1, work, 1)
+
+      ! One collective decision covers every early exit, so that all ranks take
+      ! the same path: 0 = update, 1 = skip the update, 2 = reset the Hessian.
+      exit_flag = 0
+      IF (DDOT(ndf, dx, 1, dx, 1) == zero .OR. DDOT(ndf, dg, 1, dg, 1) == zero) THEN
+         exit_flag = 1
+      ELSE IF (.NOT. ieee_is_finite(sy) .OR. .NOT. ieee_is_finite(sbs) .OR. &
+               sbs <= EPSILON(one)) THEN
```
Growl1234: `sbs` is scale-dependent, so comparing it against the absolute `EPSILON(one)` is not scale-invariant. Please use a relative threshold here.

Growl1234 (changes requested):

Growl1234:
The sdbg failure is unrelated to this PR.

src/motion/bfgs_optimizer.F
s8ga: `bfgs()` now simply outputs a `hess_reset` flag. The caller (`geoopt_bfgs`) handles the rebuild using the configured `use_mod_hes` 

This ensures `USE_MODEL_HESSIAN` is respected consistently on every reset path.

src/motion/bfgs_optimizer.F
s8ga: The threshold is now relative to the measured curvature: 

`sbs <= MAX(EPSILON(one)*ABS(sy), min_sbs)`. 

Since $s \cdot B \cdot s$ and $y \cdot s$ both carry energy units, the comparison is scale-invariant. 

`min_sbs = 10.0_dp * TINY(one)` fences off the subnormal range to provably ensure neither $1/\text{sbs}$ nor $1/\text{seta}$ can overflow.

src/motion/bfgs_optimizer.F
s8ga: `build_initial_hess` now reuses the caller's preallocated eigensolver workspace (`hess_tmp`, `eigvec_mat`, `eigval`) instead of creating and releasing a parallel set of matrices.

Growl1234 (approved):

src/motion/bfgs_optimizer.F
```diff
@@ -694,26 +694,146 @@ SUBROUTINE bfgs(ndf, dx, dg, hess_mat, work, para_env)
          END DO
       END DO
 
-      CALL para_env%sum(work)
+      IF (PRESENT(para_env)) CALL para_env%sum(work)
+
+      sy = DDOT(ndf, dg, 1, dx, 1)
```
fstein93: Is there a specific reason to use BLAS' `DDOT` over the Fortran intrinsic `DOT_PRODUCT`?

src/motion/bfgs_optimizer.F
```diff
@@ -694,26 +694,146 @@ SUBROUTINE bfgs(ndf, dx, dg, hess_mat, work, para_env)
...
+         RETURN
+      ELSE IF (exit_flag == 2) THEN
+         ! The Hessian has lost positive definiteness along dx or the predicted
+         ! curvature is not representable, so the secant pair cannot be damped:
+         ! the caller restarts from the initial Hessian and skips this update.
+         CALL cp_warn(__LOCATION__, &
+                      "BFGS: vanishing or non-positive predicted curvature, resetting the Hessian")
+         hess_reset = .TRUE.
+         CALL timestop(handle)
+         RETURN
```
fstein93: I would move the rest of the routine to the `ELSE`-branch. Then, you can remove the calls to `timestop` and the `RETURN` statements.

s8ga: Makes sense, that definitely cleans it up.

I've folded the update into the `ELSE` branch, eliminated the early `RETURN` / duplicate `timestop` calls, and introduced named parameters (`exit_update = 0`, `exit_skip = 1`, `exit_reset = 2`) to make the severity-ordered collective reduction explicit.

src/motion/bfgs_optimizer.F
s8ga: No specific reason, just carried over from the legacy bfgs implementation where `DDOT` was already in place.
I can easily update them to `DOT_PRODUCT` to match modern CP2K conventions if you'd like.

fstein93: It's more descriptive and performance is probably not the issue here.

s8ga: Just pushed 002f346, which replaces all `DDOT` calls in bfgs_optimizer.F.

fstein93 (approved):

PR #6022, opened 2026-09-15
Unify RPA/GW time-frequency grid handling

RTChou67:
This PR replaces the parallel time/frequency grid argument lists used by the RPA, GW, and GreenX paths with one shared `time_frequency_grid_type` object.


fstein93 (changes requested):
I do like the idea of clustering the grids in a separate type but I am currently not satisfied with just introducing a new type, have its destructor available but spread the actual initialization across modules although its definition is basically given by `mp2_grids`. I would rather implement or refactor type definition, (de)allocation and calculation of the grids into a single module dealing with exactly that.

JWilhelm (approved):

JWilhelm (commented):
.

src/CMakeLists.txt
```diff
@@ -413,7 +413,7 @@ list(
   mp2.F
   mp2_gpw.F
   mp2_gpw_method.F
-  mp2_grids.F
+  time_frequency_grids.F
```
Growl1234: Nit: let's alphabetize the list.

src/time_frequency_grids.F
```diff
@@ -0,0 +1,532 @@
...
+!> \brief Build a minimax grid using GreenX when it is available.
+!> \param unit_nr Output unit for diagnostics.
+!> \param num_integ_points Number of minimax points.
+!> \param emin Lower energy bound.
+!> \param emax Upper energy bound.
+!> \param regularization_minimax Regularization of the minimax fit.
+!> \param grid Grid object to fill.
+!> \param ierr Zero if GreenX supplied a grid, nonzero otherwise.
+! **************************************************************************************************
+   SUBROUTINE greenx_get_minimax_grid(unit_nr, num_integ_points, emin, emax, &
```
fstein93: I would have left this routine in `greenx_interface` to leave the conditional compilation on the wrapper level..

fstein93 (changes requested):
This looks already pretty good.

fstein93 (approved):

PR #6023, opened 2026-09-15
Toolchain: Remove OpenBLAS-based architecture detection

Growl1234:
Replace `OPENBLAS_ARCH` with a lightweight host architecture detection based on `uname -m`, normalizing only known equivalent names such as `x86_64`/`amd64` and `aarch64`/`arm64`.

Also remove `OPENBLAS_LIBCORE` and `get_openblas_arch.sh`. For native builds, OpenBLAS already performs its own CPU/core detection when `TARGET` is not specified, while the existing `DYNAMIC_ARCH` fallback is preserved.

This removes an unnecessary OpenBLAS download/probe from the early toolchain stages and simplifies the architecture handling.


oschuett (approved):

mkrack (approved):

tools/toolchain/scripts/common_vars.sh
```diff
@@ -12,9 +12,20 @@ BUILDDIR=${BUILDDIR:-"${ROOTDIR}/build"}
 SETUPFILE=${SETUPFILE:-"${INSTALLDIR}/setup"}
 VERSION_FILE=${VERSION_FILE:-"${SCRIPTDIR}/VERSION"}
 
-# system arch gotten from OpenBLAS prebuild
-OPENBLAS_ARCH=${OPENBLAS_ARCH:-"x86_64"}
-OPENBLAS_LIBCORE=${OPENBLAS_LIBCORE:-''}
+# Normalize the system architecture reported by uname across supported platforms.
+SYSTEM_ARCH=${SYSTEM_ARCH:-"$(uname -m)"}
```
he-zilong, line 16: Why is this block of code always active, rather than only when 
```shell
if [  "${TARGET_CPU}" = "native" ]
```
is satisfied?

Growl1234: It's necessary for at least CMake installation.

jpoto:
That looks nice, thanks.

tools/toolchain/scripts/common_vars.sh
he-zilong: I am fine with it and I think it is reasonable to assume first `"${TARGET_CPU}" = "native"` throughout the toolchain. But then, there should be clear messages when it is explicitly _not_ the case and it is not appropriate to read `/proc/cpuinfo` or run `uname -m` on the machine where toolchain is executed.

Growl1234: Sorry, I don't catch what do you mean; any place reading `/proc/cpuinfo` is already guarded under `"${TARGET_CPU}" = "native"` clause, isn't it?

Growl1234: And whether to take use of `uname -m`, which refers to `x86_64` and `arm64` etc. cannot be actually controlled by `--target-cpu`, which specifies SIMD-side settings.

he-zilong: If it can be ensured that every setting involving `${SYSTEM_ARCH}` (not just CMake) works for a target with non-native CPU _and_ distinct architecture then it is fine.

Growl1234: So far there shouldn't be such issues; if there's one then it also happens to `${OPENBLAS_ARCH}`.

he-zilong: Sure. Then, I suppose, we should clarify somewhere in the documentation that while toolchain has some support to non-native target CPU, it still needs to be run on a machine with the same architecture or else the `$SYSTEM_ARCH` will be problematic...

tools/toolchain/install_cp2k_toolchain.sh
```diff
@@ -1525,33 +1524,21 @@ else
...
   "${SCRIPTDIR}"/stage9/install_stage9.sh
   echo
-  # Determine native preset if --target-cpu=native is used. The compiler option
-  # and architecture found by get_openblas_arch.sh will be used to select a
-  # corresponding preset for the build_cp2k.sh script.
-  preset_option="--preset none"
+  # Determine native preset if --target-cpu=native is used. The compiler and
+  # system architecture are used to select a corresponding preset for the
+  # build_cp2k.sh script.
+  preset_option=""
```
he-zilong, line 1530: Why is `--preset none` not suggested as a fallback for a non-native target CPU?

Growl1234: With non-native target CPU it just doesn't use any preset that adds external flags, just the common `cmake`; this is cleanest and follows the principle that presets (and practically, architecture-specific flags) are *opt-in*. Specifically suggesting a `--preset none` makes no sense. In fact I don't like the idea of `--preset none` either; if users don't need presets, they simply don't use `--preset` flag...

he-zilong: Yes, omitting the `--preset` flag would work for the scenario. But since (as you have said the other day) offering more non-native architecture-optimization presets is not planned, `--preset none` could serve as a baseline for users to create their own presets.

PR #6024, opened 2026-09-15
Provide and test ASE/CP2K production container

mkrack:


Growl1234:
Looks like the Spack cache is no longer `gs://cp2k-spack-cache` in the new worker.

mkrack:
> Looks like the Spack cache is no longer `gs://cp2k-spack-cache` in the new worker.

Good catch! Indeed, it starts differently
```
StartDate: 2026-09-15 14:15:32+00:00
CpuId: 32x AMD EPYC (1st Gen) (Naples ZP-B2) [Zen], 14nm
Worker: hzdr1


CommitSHA: 9a8b1bfdaebcbeeb8ec6545959865ed90bddc7aa
CommitTime: 2026-09-15 12:15:52 +0000
CommitAuthor: Matthias Krack
CommitSubject: Merge cd87cd64cb12539c9789894d8b8b75008b9e712c into 7fa64066d5b3e7fbd220eadcd69be62170fa30df
```

@oschuett is that intended?

mkrack:
@oschuett The new runner shows further differences. `make_cp2k.sh` detects now 128 CPU cores via `nproc` though only `32x AMD EPYC` are allocated which causes an OOM crash during the `py-torch` build. If I am not mistaken, this did not happen before.
Moreover, the `Manual generation` provided artifacts for browsing the updated manual. Now these artifacts are missing.

oschuett:
Yes, there are still some teething problems with the new CI workers...

The artifacts upload should now work.

I'll setup a Spack build cache tomorrow. For now I've [stopped offloading](https://github.com/cp2k/cp2k-ci/blob/797426c8428d708a5633cb302a85d3091e4c9646/backend/jobs_util.py#L234) these jobs.

The `nproc` issue is something I have to look into because we're now running multiple containers on a single large VM.

PR #6025, opened 2026-09-15
Generalize CELL_OPT constraints on cell components

### LINKED ISSUE #5534: Feature request: Allow constraining individual lattice vectors during CELL_OPT

he-zilong:
For a CELL_OPT task, the keyword `&MOTION/&CELL_OPT/CONSTRAINT` offers a few choices to impose constraints on some of the cell components. This PR generalize it to full independent controls over the six cell components available under the program convention, namely `AX`, `BX`, `BY`, `CX`, `CY`, `CZ`.

The new implementation replaces the former input constants `fix_*` with a `INTEGER, DIMENSION(6)` 0,1-array `cell_relax()` where 0 is for a constrained component and 1 relaxed. The keyword `CONSTRAINT` along with the old options is preserved, but the parser has been augmented to accept many new ones; the input resolves to a combination of `AX`, `BX`, `BY`, `CX`, `CY`, `CZ` to fill the `cell_relax()` array. Specifically:

- `X` is mapped to `AX`, `BX`, `CX` (position 1, 2, 4);
- `Y` is mapped to `BX`, `BY`, `CY` (position 2, 3, 5);
- `Z` is mapped to `CX`, `CY`, `CZ` (position 4, 5, 6);
- `XY`, `YZ`, `XZ` receive similar treatments.

The `cell_relax()` array is used to adjust the gradient on cell components and to obtain the nominal average pressure `pres_constr` as one of the convergence criteria.

The consideration for the compatibility between 2-D periodicity and constraints on non-periodic directions, introduced in #5282 by @Growl1234 , has been preserved and adapted here. The 1-D and 0-D cases have not been added as they require further scrutiny.

Closes #5534 to which I self-assigned.

P.S. I have also edited my name throughout the code base to reduce confusion: Zilong is my given name.


src/motion/cell_opt_utils.F
```diff
@@ -157,6 +152,190 @@ SUBROUTINE gopt_new_logger_release(new_logger, root_section, para_env, project_n
...
+!> \param constraint_label ...
+!> \par History
+!>      09.2026 - Created
+!> \author Zilong HE
+!> \version 1.0
+! **************************************************************************************************
+   SUBROUTINE parse_cell_constraint(geo_section, cell, cell_relax, constraint_label)
+      TYPE(section_vals_type), INTENT(IN), POINTER       :: geo_section
+      TYPE(cell_type), INTENT(IN), POINTER               :: cell
+      INTEGER, DIMENSION(6), INTENT(OUT)                 :: cell_relax
```
RTChou67, line 169: Here, `cell_relax` is not initialized.

src/motion/cell_opt_utils.F
```diff
@@ -157,6 +152,190 @@ SUBROUTINE gopt_new_logger_release(new_logger, root_section, para_env, project_n
...
+                          "the cell vector B should be on the XY plane with the Z "// &
+                          "component always 0")
+         CASE (components(4)) ! "CX"
+            cell_relax(4) = 0
+         CASE (components(5)) ! "CY"
+            cell_relax(5) = 0
+         CASE (components(6)) ! "CZ"
+            cell_relax(6) = 0
+         CASE ("A")
+            cell_relax(1:1) = 0
```
RTChou67: So, here, `cell_relax(2:6)` is undefined

src/motion/cell_opt_utils.F
```diff
@@ -157,6 +152,190 @@ SUBROUTINE gopt_new_logger_release(new_logger, root_section, para_env, project_n
...
+         CASE ("B")
+            cell_relax(2:3) = 0
```
RTChou67, line 216: So, here, other elements in `cell_relax` is undefined.

fstein93 (approved):

RTChou67 (changes requested):

src/motion/cell_opt_utils.F
```diff
@@ -157,6 +152,190 @@ SUBROUTINE gopt_new_logger_release(new_logger, root_section, para_env, project_n
...
+            ELSE
+               CALL cp_warn(__LOCATION__, &
+                            "With 2-D CELL_OPT on the XY direction, the Z "// &
+                            "direction should be constrained; completing now.")
+               cell_relax(4) = 0
+               cell_relax(5) = 0
+               cell_relax(6) = 0
+            END IF
+         END IF
+      END IF
```
Growl1234, line 313: Can we unify these messages?

src/motion/cell_opt_utils.F
```diff
@@ -157,6 +152,190 @@ SUBROUTINE gopt_new_logger_release(new_logger, root_section, para_env, project_n
...
+            cell_relax(2:6) = 0
+         CASE ("XZ", "ZX")
+            cell_relax(1:2) = 0
+            cell_relax(4:6) = 0
+         CASE ("NONE")
+            cell_relax(1:6) = 1
+         CASE DEFAULT
+            CALL cp_abort(__LOCATION__, &
+                          "Invalid cell constraint: "//TRIM(constraint_raw(i)))
+         END SELECT
```
Growl1234, line 246: I'm not happy with the `SELECT CASE` here; can we implement it in a cleaner way?

src/motion/cell_opt_utils.F
```diff
@@ -268,22 +445,52 @@ SUBROUTINE get_dg_dh(gradient, av_ptens, pres_ext, cell, mtrx, keep_angles, &
...
+               pres_constr = 0.5_dp*(ptens(1, 1) + ptens(3, 3))
+            END IF
+         ELSE
+            IF (fix_x) THEN ! fix_x
+               pres_constr = 0.5_dp*(ptens(2, 2) + ptens(3, 3))
+            ELSE ! fix_none
+               pres_constr = (ptens(1, 1) + ptens(2, 2) + ptens(3, 3))/3.0_dp
+            END IF
+         END IF
+      END IF
```
Growl1234, line 493: This `pres_constr` logic still assumes the old X/Y/Z constraints. For arbitrary component masks such as `AX BY CZ`, none of `fix_x/y/z` is true, so the unconstrained trace pressure is used even though all diagonal cell DOFs are fixed. This can prevent an otherwise converged constrained optimization from satisfying the pressure criterion. The convergence criterion needs to be defined consistently for the new component-wise constraints.

Growl1234 (changes requested):
Besides @RTChou67's comment (good catch, that is a correctness bug), I think there's still two blockers: the below one and `SELECT CASE` use.

src/motion/cell_opt_utils.F
Growl1234: Moreover, `NONE` now seems to be able to be combined with other constraints, making the result order-dependent (`AX NONE` vs `NONE AX`).

Growl1234: Moreover, `NONE` now seems to be able to be combined with other constraints, making the result order-dependent (`AX NONE` vs `NONE AX`). At the very least `NONE` should only be accepted as a standalone value and mixed usage should be rejected.

src/motion/cell_opt_utils.F
```diff
@@ -440,25 +648,11 @@ SUBROUTINE apply_cell_constraints(gradient, cell, keep_angles, keep_symmetry, co
...
-         gradient(4:6) = 0.0_dp
-      CASE (fix_yz)
-         gradient(2:6) = 0.0_dp
-      CASE (fix_none)
-         ! Nothing to do
-      END SELECT
+      ! Apply cell constraints: cell_relax(i) is
+      ! 0 if constrained, or 1 if relaxed
+      DO i = 1, 6
+         gradient(i) = gradient(i)*cell_relax(i)
```
RTChou67, line 654: There might be something wrong with the implementation. If B=(BX,BY,0) then gradient g_B would be (g_{BX}, g_{BY}, 0). `KEEP_ANGLE` projects it onto the current B direction, thus new gradient g_B_angle would be λ (B_x, B_y, 0). However, If `CONSTRAINT BX` is on, the BX component would be zero, turning the g_B_angle into (0, λ B_y, 0), Which is not parallel to B.

src/motion/cell_opt_utils.F
```diff
@@ -157,6 +152,190 @@ SUBROUTINE gopt_new_logger_release(new_logger, root_section, para_env, project_n
...
+      LOGICAL                                            :: constraint_explicit
+
+      ! Parse input keyword
+      CALL section_vals_val_get(geo_section, "CONSTRAINT", &
+                                c_vals=constraint_raw, explicit=constraint_explicit)
+      DO i = 1, SIZE(constraint_raw)
+         CALL uppercase(constraint_raw(i))
+         SELECT CASE (constraint_raw(i))
+         CASE (components(1)) ! "AX"
+            cell_relax(1) = 0
```
RTChou67, line 187: For a cubic cell (AX,BY,CZ), the gradient would be (g_AX, g_BY, g_CZ). This is symmetrized to (g_sym,g_sym,g_sym) if `KEEP_SYMMETRY TRUE`. However, when `CONSTRAINT AX`, AX component is removed, turning gradient into (0, g_sym, g_sym). This does not satisfy the `KEEP_SYMMETRY TRUE` flag.

RTChou67 (changes requested):

s8ga:
Regarding the `pres_constr` blocker: since only the diagonal components (`AX`, `BY`, `CZ`) directly couple to the normal pressure components (`ptens(1,1)`, `ptens(2,2)`, `ptens(3,3)`), you could define `pres_constr` as a masked average over whichever diagonal DOFs are free:

```fortran
diag_free = [cell_relax(1) == 1, cell_relax(3) == 1, cell_relax(6) == 1]
IF (ANY(diag_free)) THEN
   pres_constr = SUM([ptens(1,1), ptens(2,2), ptens(3,3)], MASK=diag_free) / COUNT(diag_free)
ELSE
   pres_constr = pres_ext  ! no diagonal DOF can relax; trivially satisfy pressure tolerance
END IF
```

This reproduces all the legacy `fix_*` cases without branching, and for cases like `AX BY CZ` it naturally bypasses the pressure check rather than comparing against the unconstrained trace.

Along the same lines, if you treat the input tokens as mask combinations (or bitwise `IOR`), shorthands like `A`, `B`, `XY` can just accumulate into the mask. That drops all the permutation `CASE` branches (`AB`/`BA`, `XY`/`YX`) and makes mixed inputs like `AX NONE` easy to catch upfront.

src/motion/cell_opt_utils.F
Growl1234: For example, you can use a small subroutine:
```f08
SUBROUTINE apply_constraint_selector(selector, cell_relax)
   CHARACTER(LEN=1), INTENT(IN) :: selector
   INTEGER, DIMENSION(6), INTENT(INOUT) :: cell_relax

   SELECT CASE (selector)
   CASE ("A")
      cell_relax(1) = 0
   CASE ("B")
      cell_relax(2:3) = 0
   CASE ("C")
      cell_relax(4:6) = 0
   CASE ("X")
      cell_relax([1, 2, 4]) = 0
   CASE ("Y")
      cell_relax([2, 3, 5]) = 0
   CASE ("Z")
      cell_relax(4:6) = 0
   CASE DEFAULT
      CPABORT("Invalid cell constraint selector")
   END SELECT
END SUBROUTINE
```
and just call this routine in a modernized workflow:
```f08
cell_relax(:) = 1

DO i = 1, SIZE(constraint_raw)
   token = TRIM(constraint_raw(i))

   IF (token == "NONE") THEN
      IF (SIZE(constraint_raw) /= 1) THEN
         CPABORT("NONE cannot be combined with other constraints")
      END IF
      CYCLE
   END IF

   ! Primitive cell components: AX, BX, BY, CX, CY, CZ
   j = FINDLOC(components, token, DIM=1)
   IF (j > 0) THEN
      cell_relax(j) = 0
      CYCLE
   END IF

   ! Grouped constraints: A/B/C, AB/AC/BC, or X/Y/Z, XY/XZ/YZ.
   IF (LEN_TRIM(token) < 1 .OR. LEN_TRIM(token) > 2) THEN
      CALL invalid_constraint(token)
   END IF
   IF (VERIFY(token, "ABC") == 0 .OR. VERIFY(token, "XYZ") == 0) THEN
      DO j = 1, LEN_TRIM(token)
         CALL apply_constraint_selector(token(j:j), cell_relax)
      END DO
   ELSE
      CALL invalid_constraint(token)
   END IF
END DO
```

src/motion/cell_opt_utils.F
Growl1234: You can dedup like this, though may not best solution:
```f08
      ! For 2-D periodicity, the non-periodic direction should be constrained;
      ! write a warning and reset automatically if no explicit input is given,
      ! abort if the explicit input is invalid.
      IF (COUNT(cell%perd /= 0) == 2) THEN
         nonperiodic_axis = FINDLOC(cell%perd, 0, DIM=1)

         SELECT CASE (nonperiodic_axis)
         CASE (1)
            components = [1, 2, 4]
            axis_label = "X"
            plane_label = "YZ"
         CASE (2)
            components = [2, 3, 5]
            axis_label = "Y"
            plane_label = "XZ"
         CASE (3)
            components = [4, 5, 6]
            axis_label = "Z"
            plane_label = "XY"
         END SELECT

         IF (constraint_explicit) THEN
            IF (ANY(cell_relax(components) /= 0)) THEN
               CALL cp_abort(__LOCATION__, &
                             "With 2-D CELL_OPT on the "//plane_label//" direction, the "// &
                             axis_label//" direction should be constrained, but the explicit "// &
                             "input constraints do not satisfy this requirement!")
            END IF
         ELSE
            CALL cp_warn(__LOCATION__, &
                         "With 2-D CELL_OPT on the "//plane_label//" direction, the "// &
                         axis_label//" direction should be constrained; completing now.")
            cell_relax(components) = 0
         END IF
      END IF
```

he-zilong:
Thank you all for the valuable comprehensive late-night feedbacks.

Re uninitialized `cell_relax()`: it was an accidental slip which I agree should be fixed; the intended initial values are all 1 for the no constraint baseline.

Re input-parsing logic: this is just for a one-time input parser and the size of input options `SIZE(constraint_raw)` is not expected to be very large, so I preferred apparent code clarity (also matching the input description) from a tabulation of branches over elegance or performance gains from very condensed advanced logic. The approach with mask combinations or bitwise IOR is viable but does not appear to me as fundamentally different from a direct tabulation. The treatment of the special `NONE` should indeed have been restricted to a singular option.

Re compatibility check with 2-D periodicity: the `FINDLOC` method looks good.

Re pressure tensor and diagonal DOFs: I see that basically, the first part of the suggestion is replacing the criteria

```
      fix_x = (my_cell_relax(1) == 0 .AND. &
               my_cell_relax(2) == 0 .AND. &
               my_cell_relax(4) == 0)
      fix_y = (my_cell_relax(2) == 0 .AND. &
               my_cell_relax(3) == 0 .AND. &
               my_cell_relax(5) == 0)
      fix_z = (my_cell_relax(4) == 0 .AND. &
               my_cell_relax(5) == 0 .AND. &
               my_cell_relax(6) == 0)
```

with

```
      fix_x = (my_cell_relax(1) == 0)
      fix_y = (my_cell_relax(3) == 0)
      fix_z = (my_cell_relax(6) == 0)
```

which ignores any contribution from the sheer stress on off-diagonal components. While both can preserve the old `X/Y/Z/XY/YZ/XZ` constraints, the latter seems less restrictive and I am not sure about its validity. This is indeed a unique issue from the generalization. The second part is an implementation with `SUM(MASK=...)` which is assuming that the first part is adopted.

Re conflicts with symmetry: unfortunately, such a crucial problem has existed long before my current PR, say a `KEEP_SYMMETRY .TRUE.` with `SYMMETRY CUBIC` where DOF is 1 would not be compatible with any constraints in a CELL_OPT task at all, be it the old `CONSTRAINT X` or the new `CONSTRAINT AX`. This demands more investigations.

Growl1234:
> Re conflicts with symmetry: unfortunately, such a crucial problem has existed long before my current PR, say a `KEEP_SYMMETRY .TRUE.` with `SYMMETRY CUBIC` where DOF is 1 would not be compatible with any constraints in a CELL_OPT task at all, be it the old `CONSTRAINT X` or the new `CONSTRAINT AX`. This demands more investigations.

Wrt `KEEP_SYMMETRY`, I'm also discussing with @RTChou67; maybe we'd finally have to refactor the relavant code.

he-zilong:
> Wrt `KEEP_SYMMETRY`, I'm also discussing with @RTChou67; maybe we'd finally have to refactor the relavant code.

Do you have some other discussion venues for this? I seem to have been locked out of the loop...

For the purpose of this PR maybe I just keep the `gradient(i) = gradient(i)*cell_relax(i)` bit intact.

Growl1234:
> Do you have some other discussion venues for this? I seem to have been locked out of the loop...

We’re talking about this in a chat on instant-messaging platform. One thought at the moment is that the current projection scheme for `KEEP_SYMMETRY` may be questionable. Of course, we can open an issue or discussion on GitHub for tracking and further discussing.

PR #6032, opened 2026-09-16
Exciton descriptors: print the full crosscorrelation matrix

MGraml:
The directional exciton descriptors (`&BSE%PRINT_DIRECTIONAL_EXC_DESCR`, `&TDDFPT%DIRECTIONAL_EXCITON_DESCRIPTORS`) print the crosscorrelation matrix $R_{eh}^{\mu\mu'} = \mathrm{COV}_{eh}^{\mu\mu'}/(\sigma_e^{\mu}\sigma_h^{\mu'})$, but only its entries xx, yy, zz, xy, xz, yz. The first index refers to the electron and the second to the hole, so the matrix is not symmetric. 
However, since yx, zx, zy were never explicitly printed, this might confuse the user; this PR fixes the printing and clarifies the asymmetry.

Changes:
- The matrix is printed as a 3x3 block per excitation: rows are the electron direction, columns the hole direction.
- The printed definition now has $\sigma_h^{\mu'}$ in the denominator instead of $\sigma_h^{\mu}$. The computed values were already correct.
- New keywords `&BSE%PRINT_DIRECTIONAL_CROSSCORRELATION` and `&TDDFPT%DIRECTIONAL_EXCITON_CROSSCORRELATION` (default `.TRUE.`) allow to switch the lengthy block off independently of the other directional descriptors. Setting one explicitly without the directional descriptors triggers a warning.

```
 BSE| Crosscorrelation matrix from solving the BSE   without the TDA:
 BSE|
 BSE|      n          e\h                  x                 y                 z
 BSE|      1            x            -0.0063            0.0000           -0.0000
 BSE|                   y             0.0000            0.0037           -0.0002
 BSE|                   z             0.0000           -0.0005            0.0002
```

Output only: no computed quantity and no regtest reference changes. `regtest-bse` and `regtest-tddfpt-prop` pass, and every entry the old print showed is unchanged in the new block.


MGraml:
This rather looks like a slow worker - can one retrigger the CI without pushing a new/empty commit?

JWilhelm (approved):

PR #6033, opened 2026-09-16
Fix XC potential storage leak in energy-only KS builds

DCM-Uni-Paderborn:
## Summary

Release the temporary density and kinetic-energy-density XC potential fields when `qs_ks_build` performs an energy-only evaluation. The regular matrix-building path is unchanged.

## Cause

The deferred native atom-composite Skala path in `qs_vxc_create` creates both potential fields for each spin even for energy-only calls. Normally, `sum_up_and_integrate` returns these PW buffers to the pool and deallocates their descriptor arrays. Energy-only calls skip that routine and previously had no corresponding cleanup in `qs_ks_build`.

Consequently, repeated CG line-search evaluations lose two real-space PW buffers per spin per call, together with their grid references. This is independent of any automatic increase in `ADDED_MOS`. The defect is present on the current upstream master, `dd94896b48d429648fb3640799c9d198ff8de1e3`.

The fix uses the same pool-return operation as the normal integration path and checks whether each array is associated before releasing it. No functional, density reconstruction, grid, basis, SCF tolerance or energy expression is changed.

## Regression Test

`tools/regtesting/check_energy_only_xc_storage.py` extracts the actual allocation and cleanup blocks and links them against a completed GNU/Ninja CP2K build, using the real PW grid and pool implementations. It does not require a Skala model or a production SCF calculation.

Test matrix: one/two spins, enabled/disabled pool cache, and null/density-only/tau-only/both potential arrays. Each of the 16 combinations is repeated 100 times, with Fortran bounds checks enabled.

- Unpatched, two spins and both fields: the grid reference count grows from 2 to 402, with and without the pool cache.
- Patched: the count remains at the expected 2 in all 16 cases, and the potential descriptor pointers are disassociated.
- The fixture retains recovery pointers only to clean up the deliberately leaked baseline objects after measuring the growth, allowing normal MPI communicator finalization.

Reproduce with an existing completed GNU/Ninja build:

```bash
python3 tools/regtesting/check_energy_only_xc_storage.py \
  --build-dir build \
  --work-dir /tmp/xc-storage-check \
  --baseline dd94896b48d429648fb3640799c9d198ff8de1e3
```

## Validation Status

- Passed the extracted-source ownership regression against both the affected production revision and the current upstream master.
- The complete modified `qs_ks_methods.F` translation unit compiles with the existing GNU Debug dependency modules. This local build does not include LibTorch.
- `git diff --check` passes.
- CP2K precommit formatting and source checks pass for both changed files.
- A separate LibTorch-enabled test-library build has been submitted. No existing production source, executable or loaded library was replaced.
- **Still pending:** matched full Skala CG SCF runs comparing energy/residual sequences and resident-memory growth before/after the fix, followed by a complete small-system SCF check.

The ownership regression establishes the leak and its cleanup, but is not presented as a full SCF energy-equivalence test.


DCM-Uni-Paderborn:
Investigated the failed pdbg run ([report](https://storage.googleapis.com/cp2k-ci/run-cp2k-pdbg-7aa27612_report.txt)): the build succeeded, 5305 tests passed, and there are 0 numerical WRONG results. Of the 51 failures, 50 hit the 150 s timeout. The remaining failure, `QS/regtest-pao-2/H2O_pao_rotinv_restart.inp`, cannot find `H2O_pao_rotinv-1_0.pao` because its producer `H2O_pao_rotinv.inp` timed out earlier in the same batch.

The timeouts span multiple subsystems, including xTB. This points to a CI execution-time issue rather than an established regression from the XC-storage cleanup. Requesting a fresh pdbg run before drawing a final conclusion. No source changes, relaxed tolerances, or skipped tests.

/cp2kci pdbg

oschuett:
> Investigated the failed pdbg run ([report](https://storage.googleapis.com/cp2k-ci/run-cp2k-pdbg-7aa27612_report.txt)):...

Sorry, this one is on me. The `do_regtest.py` script was running too many tests in parallel because Python's `os.cpu_count()` ignored the container's cpu-set. It's fixed now.

Issue #5570, opened 2026-07-10
RFC: Future of cp2k-output-tools


oschuett:
In #5550 @marci73 and others have flagged the need for good post-processing tools.

Although we concluded that a rewrite towards parsable output is infeasible #215, we haven't really offered an alternative solution either.

A possible way forward would be to revitalize the [cp2k-output-tools](https://github.com/cp2k/cp2k-output-tools) and make them a first-class API. For this we'd have to bring the code into our main repository, fix or remove its broken [dependencies](https://github.com/cp2k/cp2k-output-tools/blob/403cc70966d7284d9f71f56b8cd2ae08f969fa62/pyproject.toml#L18), and most importantly add its tests to our dashboard.

dryndyk:
The idea is definitely good! I have some questions in this direction.
1. Are there other actual Python tools that should be combined with  `cp2k-output-tools`  to support some standard: building, help, coding conventions, etc.?
2. What should this standard be? 
3. Just now, I am working on Python scripts for transport (NEGF) calculations, including postprocessing, as well as "external" scripting for I-V curve calculations and STM scanning. I need to accept some standard for it. At the moment, I use ASE-like coding and building. But I am ready to rewrite it to keep consistency with other tools.

oschuett:
In our main repository we currently use `black` for formatting and `mypy --strict` for type checking.

dryndyk:
Good, I'll use it too.

he-zilong:
These questions for the development history may or may not be on topic, but I wonder, what was the reason that `cp2k-input-tools` and `cp2k-output-tools` did not get unified like a single `cp2k-python-interface`? And why did the `cp2k-output-tools` not utilize the [cython bindings](https://github.com/cp2k/cp2k/tree/fe787632bd21d8bb579232615b8aeebc7927b9bc/src/start/python) (which has been removed recently)?

Growl1234:
I would personally wish the workflows related to input/output handling to be more integrated with ASE tools.

dryndyk:
Me too. And I checked that my current script to start cp2k is based on `cp2k.py ` "ASE calculator" written by @oschuett.  )

he-zilong:
[...]

PR #6038, opened 2026-09-17
Add native Wilson loops and Z2 analysis with a Z2Pack interface

DCM-Uni-Paderborn:
## Summary

Add two complementary topology paths to the existing DFT/PRINT/WANNIER90 facility:

- Explicit fractional k-points and directed cross-k overlap connections from NNKP files, providing the native NNKP/MMN file interface for external Z2Pack adapters.
- Native SVD-polar Wilson loops, Berry phases and hybrid Wannier centres, plus converged Z2 parity on coordinate time-reversal half-planes.

The Gaussian-basis overlaps retain the nonorthogonal AO metric and the periodic phase/image conventions. The explicit export uses ordered nonsymmetric pair matrices; the legacy regular-grid path remains unchanged. The potential is fixed after SCF while the requested post-SCF k-points are diagonalized.

The SOC path uses restricted SCF followed by second-variational pseudopotential SOC, consistent with the existing CP2K band-structure implementation. It is not self-consistent noncollinear DFT. Band-subspace convergence and time-reversal symmetry remain the user's responsibility. Native checks reject singular links, sampled gap closings, failed Kramers pairing and unresolved surfaces.

## Scope after review

1. Clear stale screened SOC projector integrals.
2. Work around GCC 16 nested-component array-section bounds-check failures in replicated grid transfers and k-point operators, including Davidson. Bounds checks remain enabled.
3. Add explicit overlaps and native Wilson/Z2 analysis, including input documentation and bibliography.
4. Keep user documentation in the existing Wannier90 manual and run short native tests through CP2K's existing test machinery.

Following Ole's review, the Python adapter, its tests and the larger manual validation scripts have been extracted into a standalone development package and removed from this PR. The package's eventual upstream home is still to be agreed; `cp2k-input-tools` is a possible option. No Python topology tooling, Z2Pack installation, or additional Misc-CI dependencies remain in this PR. The native NNKP/MMN interface and native Wilson/Z2 calculations are unchanged by this separation.

## Short automatic tests only

- `topology_wilson_unittest`, built by CMake and registered in `tests/UNIT_TESTS`: four known trivial/nontrivial BHZ models, deterministic non-Abelian gauge changes, loop reversal, analytic WCC/Berry phases, singular-link and Kramers-pair rejection, and surface-resolution checks. No Python interpreter or Z2Pack is needed by this test.
- Registered CP2K regressions: small helium native-surface, neon SOC Z2=0, and four-point NNKP helium Berry-phase tests. Together with the native unit test, all 4 pass through `tests/do_regtest.py` with 2 MPI ranks and 2 OpenMP threads. Local per-test times were 0.13 seconds for the unit test and 1.09, 2.11 and 0.89 seconds for the three DFT inputs.
- No large stanene calculation or MPI sweep is added to automatic CI. The larger validation tools are preserved separately from this PR.

## Local checks and previous scientific validation

- Full CP2K precommit pipeline for changed files and diff checks; the new unit test also passes a bounds-checked build with the strict debug warning policy.
- Outside this PR, the extracted Python package builds and installs as a wheel; all 12 retained model/parser/adapter tests pass (approximately 1.1 seconds). Its 4 adapter-only tests also pass without a CP2K source checkout.
- 13 existing Wannier90 regression inputs pass their unchanged matchers.
- Existing grid and k-point moment regressions pass. A broader exploratory run exposed GCC 16 failures in the synchronous Davidson path; the pointer workaround fixes the short helium Davidson reproducer, including its original energy and k-point-count matchers.
- Bounds-checked Apple Silicon/GCC 16 build; short MPI/OpenMP smoke tests pass. Independent Linux/compiler CI results remain to be supplied by the normal PR checks.
- Previously completed (not rerun for PR preparation): native and independently adaptive Z2Pack stanene Z2=1; native 193x192 surface converged; serial/two-rank Wilson eigenvalue agreement within 4.5e-11. Neon agrees across 1/2/4 ranks. AO adjoint/metric identities and all 26 Ne/52 Sn spinor bands agree with the existing SOC reference at its output precision.

See `docs/methods/electronic_structure/wannier90.md` for native usage and limitations. The stanene results are numerical benchmarks, not basis/cutoff-converged material predictions.


src/input_cp2k_print_dft.F
```diff
@@ -2913,16 +2913,102 @@ SUBROUTINE create_wannier_section(print_key)
...
+                                        "Use the full k-point mesh from DFT%KPOINTS.", &
+                                        "Read explicit k-points and connections from NNKP_FILE.", &
+                                        "Generate a family of closed Wilson loops."), &
                           default_i_val=0)
       CALL section_add_keyword(print_key, keyword)
       CALL keyword_release(keyword)
 
+      CALL keyword_create(keyword, __LOCATION__, name="NNKP_FILE", &
+                          description="Wannier90 nnkp file for KPOINTS_SOURCE NNKP. Required blocks: "// &
+                          "real_lattice (Angstrom), recip_lattice (2*pi/Angstrom), kpoints (fractional), "// &
```
fstein93, line 2932: `reciprocal_lattice`?

src/input_cp2k_print_dft.F
```diff
@@ -2913,16 +2913,102 @@ SUBROUTINE create_wannier_section(print_key)
...
+                          "and nnkpts (point indices and reciprocal lattice shifts). Nonzero exclusions "// &
```
fstein93: `nkpts`?

src/topology_wilson.F
```diff
@@ -0,0 +1,282 @@
...
+      END DO
+      status = -2
+      IF (.NOT. kramers_pairs(sorted(:, 1), pair_tol)) RETURN
+      IF (.NOT. kramers_pairs(sorted(:, lines), pair_tol)) RETURN
+      crossings = 0
+      DO i = 2, lines
+         lo = MIN(gaps(i - 1), gaps(i))
+         hi = MAX(gaps(i - 1), gaps(i))
+         DO j = 1, n
+            IF (ABS(sorted(j, i) - gaps(i - 1)) < 1.e-12_dp) THEN
```
fstein93: I would convert this magic constant to a parameter

src/wannier90_nnkpts.F
```diff
@@ -0,0 +1,225 @@
...
+         CASE ("real_lattice", "recip_lattice")
+            DO i = 1, 3
+               CALL required_line(unit, line)
+               READ (line, *, IOSTAT=ios) lattice(:, i)
+               IF (ios /= 0) CPABORT("NNKP_FILE: invalid lattice vector.")
+            END DO
+            IF (.NOT. ALL(ieee_is_finite(lattice))) CPABORT("NNKP_FILE: nonfinite lattice.")
+            IF (block == "real_lattice") THEN
+               IF (have_real) CPABORT("NNKP_FILE: duplicate real_lattice block.")
+               IF (MAXVAL(ABS(lattice - real_lattice)) > 1.e-6_dp) THEN
```
fstein93: If this `1.0e-6_dp` is a magic constant, I would convert to a parameter

src/wannier90_nnkpts.F
```diff
@@ -0,0 +1,225 @@
...
+      CALL lowercase(marker)
+      CALL lowercase(name)
+      IF (marker /= "end" .OR. name /= block) CPABORT("NNKP_FILE: incorrect block terminator.")
+   END SUBROUTINE end_block
+
+! **************************************************************************************************
+!> \brief Convert ASCII keyword to lowercase.
+!> \param string keyword
+! **************************************************************************************************
+   SUBROUTINE lowercase(string)
```
fstein93: This functionality already exists in `string_utilities`

fstein93 (changes requested):

src/input_cp2k_print_dft.F
DCM-Uni-Paderborn: `recip_lattice` is the literal block name in the Wannier90 `.nnkp` format, so renaming it would break compatibility. I clarified that these are literal block names in 6575651. The upstream reference file uses `begin recip_lattice`: https://github.com/wannier-developers/wannier90/blob/develop/test-suite/tests/testw90_nnkpt1/benchmark/wannier.nnkp#L11-L15

src/input_cp2k_print_dft.F
DCM-Uni-Paderborn: `nnkpts` is the literal nearest-neighbour connection block; `kpoints` is the separate coordinate block. I expanded the description to say nearest-neighbour indices in 6575651. The upstream reference uses `begin nnkpts`: https://github.com/wannier-developers/wannier90/blob/develop/test-suite/tests/testw90_nnkpt1/benchmark/wannier.nnkp#L54-L65

src/topology_wilson.F
DCM-Uni-Paderborn: Done in 6575651: replaced the literal with the named `crossing_tol` parameter and documented its reduced-coordinate meaning. Also named the adjacent-gap fraction. Numerical values are unchanged.

src/wannier90_nnkpts.F
DCM-Uni-Paderborn: Done in 6575651: added `real_lattice_tol` and `recip_lattice_tol` parameters, with their units documented. Both retain the existing 1e-6 value.

src/wannier90_nnkpts.F
DCM-Uni-Paderborn: Done in 6575651: imported `lowercase` from `string_utilities` and removed the duplicate implementation. The explicit NNKP smoke test passes.

fstein93 (approved):

oschuett:
This PR also adds a lot of Python code to `tools/`.  If this is meant to eventually become a production tool, then maybe it would be better placed into the [cp2k-input-tools](https://github.com/cp2k/cp2k-input-tools). If instead, this is only meant for testing, then we should try to use our existing testing machinery.

DCM-Uni-Paderborn:
Dear Ole,

thank you for pointing this out. The directory mixed a user-facing Z2Pack adapter with development and validation scripts. I have separated these concerns in 03986b5:

- Removed `tools/topology` from this PR, together with the additional Z2Pack/pytest/OpenBLAS installation and the custom Misc-CI test invocation.
- Moved the essential mathematical checks into `topology_wilson_unittest`, built through CMake and registered in `tests/UNIT_TESTS`. It covers known trivial/nontrivial BHZ models, non-Abelian gauge changes, loop reversal, analytic Berry phases, and rejection checks without Python or Z2Pack.
- Kept the three short helium/neon regression inputs in the existing CP2K test machinery and moved the native usage documentation into the Wannier90 manual. No long DFT benchmarks are added to CI.

The native Wilson/Z2 implementation and the NNKP/MMN file interface remain unchanged by this separation. The native unit test and all three regression inputs pass through `do_regtest.py` with two MPI ranks and two OpenMP threads.

I have prepared the Python side as a separate installable development package, preserving its tests and the manual validation scripts; it is not published yet. Would `cp2k-input-tools` be a suitable home for the user-facing adapter in a separate follow-up, or would you prefer another location? Its scope includes running CP2K and reading overlap matrices, in addition to generating input.

Greetings,
Thomas

oschuett:
Thanks for sticking with regtests / units tests. 

> I have prepared the Python side as a separate installable development package, ...

Yes, there is clearly a need for an auxiliary Python ecosystem around CP2K. The [cp2k-input-tools](https://github.com/cp2k/cp2k-input-tools) and [cp2k-output-tools](https://github.com/cp2k/cp2k-output-tools) packages made a good start. Unfortunately, these projects have fallen dormant after Tiziano's departure. 

Seriously extending our scope towards Python would require someone to step up and take the lead. See also https://github.com/cp2k/cp2k/issues/5570.

PR #6040, opened 2026-09-17
Test OpenMPI v6.0.0 RC

Growl1234:


he-zilong:
Uh oh, another dependency (or another setting to explicitly disable) here...
```
checking for flex... no
checking for lex... no
configure: WARNING: *** Could not find Flex on your system.
configure: WARNING: *** Flex is required for developer builds of Open MPI.
configure: WARNING: *** Other versions of Lex are not supported.
configure: WARNING: *** NOTE: If you are building a tarball downloaded from www.open-mpi.org, you do not need Flex
configure: error: Cannot continue
```

Growl1234:
I suspect there's another thing wrong, as I am using the tarball downloaded from open-mpi.org.

Edit: Reported as https://github.com/open-mpi/ompi/issues/14453. I didn't catch this locally as my system has Flex (installed as a dependency of `nvidia-driver` by `dnf`).

Growl1234:
Regarding MPI threading unittest timeout, I strongly suspect the root cause is https://github.com/open-mpi/ompi/issues/14415.

PR #6042, opened 2026-09-17
Add GROMACS/CP2K benchmark folder

mkrack:


mkrack:
The individual, slightly erroneous test result is unrelated to this PR.

oschuett:
Maybe we could also add a benchmark for CP2K's builtin QM/MM and run it daily for the dashboard?

It should be a great proxy for the GROMACS benchmark since they both rely on GEEP.

oschuett:
Actually, never mind. We're already running [MQAE_single_node.inp](https://github.com/cp2k/cp2k/blob/master/benchmarks/QMMM_MQAE/MQAE_single_node.inp).

PR #6047, opened 2026-09-17
Docs: Replace py:data with lightweight keyword targets

Growl1234:
CP2K input keywords are currently rendered as Sphinx `py:data` objects, although they are not Python objects and only require stable cross-reference targets. This PR replaces these directives with lightweight MyST targets and `rubric` headings while preserving existing keyword anchors and cross-references. This can also substantially reduce the Sphinx object-processing overhead and memory usage during documentation builds.


oschuett:
That's an impressive speedup! Unfortunately, the search experience is now even worse: [old](https://manual.cp2k.org/trunk/search.html?q=SYMMETRY_REDUCTION_METHOD&check_keywords=yes&area=default) vs [new](https://ci.cp2k.org/artifacts/cp2k-manual-a1a2bff3/manual/search.html?q=SYMMETRY_REDUCTION_METHOD&check_keywords=yes&area=default).

Have you explored adding a custom directive to Sphinx?

Growl1234:
> Unfortunately, the search experience is now even worse: [old](https://manual.cp2k.org/trunk/search.html?q=SYMMETRY_REDUCTION_METHOD&check_keywords=yes&area=default) vs [new](https://ci.cp2k.org/artifacts/cp2k-manual-a1a2bff3/manual/search.html?q=SYMMETRY_REDUCTION_METHOD&check_keywords=yes&area=default).

That looks more like expected because Python data is gone. I'll take a try; maybe is there way to add plugin, like what's in ReadTheDocs, or turn to a totally different approach like [Pagefind](https://pagefind.app/)?

Growl1234:
It maybe also the side effect of `rubric` headings.

Growl1234:
With Pagefind integrated the search UI now looks like this:
图片

Growl1234:
One additional note: the Pagefind plugin cannot work if you are exploring the built manual locally through `file://` approach.

oschuett:
> One additional note: the Pagefind plugin cannot work if you are exploring the built manual locally through file:// approach.

That could probably be achieved by loading `pagefind-component-ui.js` directly from a `<script>` tags instead of the indirect route through `enablePagefind()`.

Growl1234:
Yes this is cleaner, but Pagefind still fetches `pagefind-entry.json` etc. at runtime, so I'm not sure if it helps.

oschuett:
There is also a [Sphinx extension](https://pypi.org/project/atsphinx-pagefind) for Pagefind, which looks a bit unmaintained though.

Nevertheless, they also load the pagefind javascript [directly](https://github.com/atsphinx/pagefind/blob/32d324084617887aeee6b1fd60c5979d6ccac8a1/src/atsphinx/pagefind/_templates/search.html#L25).

Growl1234:
> There is also a [Sphinx extension](https://pypi.org/project/atsphinx-pagefind) for Pagefind, which looks a bit unmaintained though.

It's still marked as `Alpha` and targets Pagefind v1.4, while we're using Pagefind v1.5 here which introduced substantial changes.

oschuett (approved):
Great work!

PR #6048, opened 2026-09-18
QS: Thick-restart block Lanczos diagonalization for k-point [Draft]

s8ga:
## Summary
This PR enables `ALGORITHM LANCZOS` for k-point calculations with complex wavefunctions. Previously, attempting to use block Krylov/Lanczos with k-points resulted in a `CPABORT`. 

### Complex Thick-Restart Lanczos Kernel (`qs_scf_lanczos_c.F`)
Implements the thick-restart block Lanczos (TRLan) algorithm for complex Hermitian generalized eigenproblems following Wu & Simon (SIAM J. Matrix Anal. Appl. 2000).

## Keywords & Documentation

Updated descriptions under `SCF%DIAGONALIZATION%KRYLOV`:

- `NBLOCK`: Clarified its role in k-point complex calculations. When `NBLOCK` is smaller than the number of pending MOs, columns are refined in blocks. When set larger than or equal to the pending MO count, all columns are refined simultaneously in a single shared subspace of roughly $2 \times N_\text{pending}$ width (following the thick-restart conventions of Wu and Simon).
- `NKRYLOV`: Clarified that in the shared-subspace regime, `NKRYLOV` only acts as a lower floor on the subspace budget, while the budget is primarily determined by the block width.

**`tests/QS/regtest-kp-3` (9 inputs):**
- `he_lan`, `he_lan_adapt`: Insulating He crystal with fixed and adaptive inner thresholds (`EPS_ADAPT`), checking convergence against existing Davidson/OT reference energies.
- `Al-kpoint-lan-smear`: Metallic Al baseline with Fermi-Dirac smearing and fractional occupations.
- `Al-kpoint-lan-smear-auto`: Metallic Al with dynamic buffer expansion (`ADDED_MOS AUTO`).
- `Al-kpoint-lan-smear-block4`: Multi-column block refinement (`NBLOCK 4`), exercising within-batch orthogonalization.
- `Al-kpoint-lan-smear-global`: Shared-subspace regime (`NBLOCK 16`), refining all pending MOs in one joint Krylov space.
- `Al-kpoint-lan-smear-deep`: Deep inner tolerance (`EPS_KRYLOV 1E-12`), driving execution into multiple thick-restart cycles.
- `Al-kpoint-lan-smear-exh`: Cycle exhaustion test (`MAX_ITER 1`), exercising the safe partial write-back path.
- `Al-kpoint-lan-smear-pgs2`: Parallel layout test with `PARALLEL_GROUP_SIZE 2`, verifying distributed Rayleigh-Ritz execution over multiple ranks in a k-point group.

**`tests/QS/regtest-elpa` (1 input):**
- `cplx_kp_lan`: Exercises the Lanczos refinement using the ELPA complex eigenvalue solver.


Growl1234:
From the first glance there's a lot of code to clean up.

Growl1234:
This PR introduced a new algorithm rather than extending the existing algorithm to k-points. Please investigate further about, why we must use TRLan for k-points support, what advantage it has brought, and whether it can eventually replace the original implementation.

src/qs_scf_diagonalization.F
```diff
@@ -3525,6 +3415,286 @@ SUBROUTINE do_block_davidson_diag_kp(qs_env, scf_env, matrix_ks_kp, matrix_s_kp,
...
+            IF (nmo_nc > 0 .AND. output_unit > 0) THEN
+               WRITE (output_unit, "(T4, A)") " WARNING Lanczos refinement could "// &
+                  "not converge all the mos:"
+               WRITE (output_unit, "(T40,A,T70,I10)") " number of not converged mos ", nmo_nc
+               WRITE (output_unit, "(T40,A,T70,E10.2)") " max norm of the residual ", max_res
+            END IF
+
+            ! rotate the refined MOS back into the generalized problem, C = U^-1 X
+            CALL cp_cfm_triangular_multiply(csmat, cmo, side="L", invert_tr=.TRUE.)
+            CALL cp_cfm_to_fm(cmo, mo_coeff_r, mo_coeff_i)
```
RTChou67, line 3650: Could we use the existing `kpoint_operator_cfm_to_mo` helper here instead?

RTChou67 (commented):
GPT 5.6-Sol stated that the current implementation has a lot of correctness problems. Still checking

oschuett:
> GPT 5.6-Sol stated that the current implementation has a lot of correctness problems.

Did it get any more specific about what these correctness problems are? Generally, incremental improvements are much preferable over rewrites.

RTChou67:
> > GPT 5.6-Sol stated that the current implementation has a lot of correctness problems.
> 
> Did it get any more specific about what these correctness problems are? Generally, incremental improvements are much preferable over rewrites.

I'll share them here as soon as I have confirmed all of them.

s8ga:
Thanks you all for looking into this.

I am completely open to any refactoring, code cleanups, and architectural adjustments needed to get this properly aligned with CP2K standards.

src/qs_scf_diagonalization.F
s8ga: Good point, thanks!
I'll route the write-back through `kpoint_operator_cfm_to_mo` for consistency with the other k-point paths.

RTChou67:
On the architectural level, my idea is that, TRLan is just an algorithm. For implicit gamma, `cp_fm_type` is required. For explicit Gamma + `use_real_wfn=.TRUE.`, we could still use the real representation. Only for normal k-points, we use `cp_cfm_type`. So, the whole chain could be following
```
                  TRLan algorithm  
                /                 \  
               /                   \  
     cp_fm backend              cp_cfm backend  
     real arithmetic            complex arithmetic  
          |                           |  
 implicit Γ                  explicit k-points  
 explicit real Γ             general k-points
```
with something like 
```
INTERFACE trlan_refinement
   MODULE PROCEDURE trlan_refinement_fm
   MODULE PROCEDURE trlan_refinement_cfm
END INTERFACE
```

src/qs_scf_lanczos_c.F
```diff
@@ -0,0 +1,758 @@
...
+            ! Ritz pairs, one per pending column
+            IF (single_block) THEN
+               CALL cp_cfm_set_submatrix(yw1, zvec(1:nbasis, 1:nres), 1, 1, nbasis, nres)
+               CALL cp_cfm_set_submatrix(yw2, zvec(1:nbasis, 1:nres), 1, 1, nbasis, nres)
+            ELSE
+               CALL cp_cfm_set_submatrix(yw1, zt(1:nbasis, 1:nres), 1, 1, nbasis, nres)
+               CALL cp_cfm_set_submatrix(yw2, zt(1:nbasis, 1:nres), 1, 1, nbasis, nres)
+            END IF
+            ctheta(1:nres) = CMPLX(theta(1:nres), 0.0_dp, KIND=dp)
+            CALL cp_cfm_column_scale(yw2, ctheta(1:nres))
```
RTChou67, line 481: This is unsafe when `nres` < `nblock_e`.

`yw2` is created with nblock_e global columns, but the scaling array passed here has only `nres` entries. In the parallel implementation of cp_cfm_column_scale, the scaling factor is indexed with the global column index. Therefore SIZE(scaling) must cover every global column index owned by the local rank. Limiting the local loop count with MIN(ncol_local, SIZE(scaling)) does not make a shorter global scaling array safe.

For example, if nblock_e = 4 and nres = 2, a rank owning global columns 1 and 3 may still access scaling(3) although only scaling(1:2) was passed.

src/qs_scf_lanczos_c.F
```diff
@@ -0,0 +1,758 @@
...
+         ! not touch the candidates
+         CALL cp_cfm_triangular_multiply(gram, rblk, "R", invert_tr=.TRUE., n_cols=mlive)
+      END DO
+      IF (.NOT. lseq) THEN
+         CALL cp_cfm_vectorsnorm(rblk, nrm)
+         ALLOCATE (zscl(mlive))
+         DO i = 1, mlive
+            zscl(i) = CMPLX(1.0_dp/nrm(i), 0.0_dp, KIND=dp)
+         END DO
+         CALL cp_cfm_column_scale(rblk, zscl)
```
RTChou67, line 723: Same here

src/qs_scf_lanczos_c.F
```diff
@@ -0,0 +1,758 @@
...
+            CALL cp_cfm_gemm('N', 'N', nao, nmo, nmo, z_one, c0, evec, z_zero, c1)
+            CALL cp_cfm_column_scale(c1, g_scale)
+            CALL cp_cfm_gemm('N', 'C', nao, nmo, nmo, z_one, c1, evec, z_zero, c0)
+            ! directions clamped at the floor stay below unit norm, and the
+            ! entry Rayleigh-Ritz of the next call assumes C^dag C = I:
+            ! renormalize every column. A fully degenerate direction
+            ! reconstructs to the zero vector and stays zero
+            CALL cp_cfm_vectorsnorm(c0, g_eval)
+            g_scale(:) = CMPLX(1.0_dp/MAX(g_eval, TINY(1.0_dp)), 0.0_dp, KIND=dp)
+            CALL cp_cfm_column_scale(c0, g_scale)
```
RTChou67, line 229: Clamping a small Gram eigenvalue before applying the inverse square root leaves a residual overlap in that direction. The subsequent `cp_cfm_vectorsnorm` / `cp_cfm_column_scale` only normalizes each column independently; it does not remove off-diagonal overlap between different columns.

src/qs_scf_diagonalization.F
```diff
@@ -3525,6 +3415,286 @@ SUBROUTINE do_block_davidson_diag_kp(qs_env, scf_env, matrix_ks_kp, matrix_s_kp,
...
+            ! complex MOS of the channel
+            CALL cp_cfm_create(cmo, mo_coeff_r%matrix_struct, name="lanc_cmo")
+            CALL cp_cfm_create(cwork, mo_coeff_r%matrix_struct, name="lanc_cwork")
+            CALL cp_cfm_scale_and_add_fm(z_zero, cmo, z_one, mo_coeff_r)
+            CALL cp_cfm_scale_and_add_fm(z_one, cmo, gaussi, mo_coeff_i)
+
+            ! standardization S(k) = U^dag U: operator and MOS are pulled
+            ! through the same transform, X = U C and A = U^-dag H U^-1, so
+            ! the kernel faces a standard eigenproblem
+            CALL cp_cfm_cholesky_decompose(csmat)
```
RTChou67, line 3600: This path always performs a Cholesky factorization of the overlap matrix, regardless of `scf_env%cholesky_method`.

src/qs_scf_lanczos_c.F
```diff
@@ -0,0 +1,758 @@
...
+            END IF
+            ctheta(1:nres) = CMPLX(theta(1:nres), 0.0_dp, KIND=dp)
+            CALL cp_cfm_column_scale(yw2, ctheta(1:nres))
+            CALL cp_cfm_gemm('N', 'N', nao, nres, nbasis, z_one, hvmat, yw1, z_zero, rblk)
+            CALL cp_cfm_gemm('N', 'N', nao, nres, nbasis, z_mone, vmat, yw2, z_one, rblk)
+            CALL cp_cfm_vectorsnorm(rblk, nrm)
+            rhot(1:nres) = nrm(1:nres)
+            DO j = 1, nres
+               max_res_norm = MAX(max_res_norm, rhot(j))
+               min_res_norm = MIN(min_res_norm, rhot(j))
```
RTChou67, line 488: `max_res_norm` is already initialized from the entry residual before trlan_refine is called, and then it can only increase. Therefore a calculation that starts with a residual of 1e-2 and finishes with all residuals around 1e-10 can still return max_res_norm = 1e-2.

src/qs_scf_lanczos_c.F
```diff
@@ -0,0 +1,758 @@
...
+      min_res_norm = min_norm
+
+      IF (.NOT. my_check_moconv_only) THEN
+         IF (nmo_nc_loc > 0) THEN
+            CALL trlan_refine(ks, c0, eval, nao, nmo, eps_col, nkrylov, nblock, max_cycles, &
+                              nmo_converged, nmo_nc_loc, res_entry, n_conv_cols, max_res_norm, min_res_norm, &
+                              my_ncycles)
+            IF (PRESENT(ncycles)) ncycles = my_ncycles
+            nmo_conv = nmo_converged + n_conv_cols
+            nmo_nc = nmo_nc_loc - n_conv_cols
```
RTChou67, line 199: `nmo_converged` is correctly computed as the length of the leading converged prefix. Inside `trlan_refine`, however, `n_conv_cols` is incremented whenever any target column converges, including a later column when an earlier one is still unconverged.

s8ga:
Heads-up: we rewrote the branch history before answering the review — the thick-restart kernel from the first push is gone.
Diff is now three commits (kernel, driver, tests): a single-pass block Lanczos with the restart in the driver, same shape as the real-valued path.

src/qs_scf_lanczos_c.F
s8ga: Right, fixed — `ctheta` now spans every global column of `yw2` (`ctheta(:) = 1; ctheta(1:nres) = theta`), dead tail scales by one.
Your exact example (nblock_e 4, nres 2, rank owning columns 1 and 3) now reads a full-width array.

src/qs_scf_lanczos_c.F
s8ga: Same fix here — `zscl` spans all `ncolg` global columns, retired tail scales by one. `purify_and_append`.

src/qs_scf_lanczos_c.F
s8ga: Right — clamping leaves the overlap, and per-column normalization can't remove it.
Fallback now runs the clamp-and-inverse-sqrt pass twice: a clamped direction keeps ~sqrt(g/gfloor) off-diagonal overlap after one pass, the second squares it down.
Deliberately no per-column renormalization — it would rescale a clamped direction back to unit norm and reimport exactly the overlap the clamp bounds.

src/qs_scf_diagonalization.F
s8ga: Documented in the driver now — standardization does a plain Cholesky of S(k), same as the complex k-point family (`cp_cfm_geeig` decomposes S the same way and ignores `cholesky_method` beyond the off switch).
CHOLESKY OFF stays unsupported on this path.
Switching to the canonical variant would be a small isolated change to `lanczos_refine_channel` if preferred.

src/qs_scf_lanczos_c.F
s8ga: Reworked — residuals now report the exit state (final sweep's wanted pairs), not the worst seen during the call.
Entry Rayleigh-Ritz values are kept only when no refinement ran.
A channel that enters at 1e-2 and finishes at 1e-10 now reports 1e-10.

src/qs_scf_lanczos_c.F
s8ga: Fixed with a prefix guard — a block ending short of full convergence closes the leading prefix, 
later blocks contribute nothing even if all their own columns converged (those sit beyond the gap).
Count stays the global leading run.

s8ga:
> This PR introduced a new algorithm rather than extending the existing algorithm to k-points. Please investigate further about, why we must use TRLan for k-points support, what advantage it has brought, and whether it can eventually replace the original implementation.

Good question, and it changed the PR.

The kernel's plain block Lanczos now:

- one Krylov expansion per call, one Rayleigh-Ritz, one write-back of the target block — identities re-judged every call, never latched, variationally safe (Courant-Fischer)
- restart moved to the driver: `DO iter = 1, max_iter` with a residual early exit, same as the real-valued driver above it.
  Both paths share one restart style now instead of k-points running a second algorithm
- dropped the Wu-Simon citation along with the mechanism it described

Couldn't just extend the existing real kernel: its block path has no rank safeguards.
With its default NBLOCK 32 it aborts on a four-atom Al cell at the second SCF step (PDSYEVD failure inside the entry Rayleigh-Ritz; NBLOCK 4 dies to a literal NaN one step earlier).
NBLOCK 1 and 2 converge to the same state.
The kernel's only regtest coverage (both si8_broy_kry variants) pins NBLOCK to 1, so the default's never actually been exercised in CI.
What we kept from this PR is the part that fixes exactly this: two-pass purge, per-column rank safeguards, CholeskyQR2 with a column-at-a-time fallback, the re-judged write-back.
The real kernel may need the same machinery eventually.

input and output file:
for NBLOCK issue
[nblock-gamma-repro.tar.gz](https://github.com/user-attachments/files/32466414/nblock-gamma-repro.tar.gz)

s8ga:
> On the architectural level, my idea is that, TRLan is just an algorithm. For implicit gamma, `cp_fm_type` is required. For explicit Gamma + `use_real_wfn=.TRUE.`, we could still use the real representation. Only for normal k-points, we use `cp_cfm_type`. So, the whole chain could be following
> 
> ```
>                   TRLan algorithm  
>                 /                 \  
>                /                   \  
>      cp_fm backend              cp_cfm backend  
>      real arithmetic            complex arithmetic  
>           |                           |  
>  implicit Γ                  explicit k-points  
>  explicit real Γ             general k-points
> ```
> 
> with something like
> 
> ```
> INTERFACE trlan_refinement
>    MODULE PROCEDURE trlan_refinement_fm
>    MODULE PROCEDURE trlan_refinement_cfm
> END INTERFACE
> ```

The thing your interface wanted to factor out doesn't exist anymore, at least for now — we pulled TRLan, and the kernel is a stateless single pass with the restart in the driver, same loop as the real-valued path.

We like the split though. It's also the natural way to give real k-point wavefunctions a backend — they hit a CPABORT today, ours and Davidson's alike, only the standard k-point driver runs a real branch.

Growl1234 (changes requested):
Would need more work on cleanup before this is ready:

- There is a lot of repeated `cp_fm_struct_create`/`cp_cfm_create` boilerplate. Please reuse existing matrix structures where possible.
- Scratch matrices are allocated and released on every Lanczos sweep. Please introduce a reusable workspace.
- `lanczos_refinement_2v_c` exposes several dimensions and scratch objects that can be derived internally. Please reduce the interface.
- `krylov_space_type` seems to mix settings with the old real-Lanczos workspace, which leads to partially initialized objects.
- The k-point Lanczos driver duplicates a fair amount of Davidson setup and finalization logic.
- The local cold-start and refinement paths contain duplicated control flow, please consolidate them.
- Some cold-start eigensolver logic duplicates existing complex diagonalization infrastructure.
- Tolerance, occupation, density, and `ADDED_MOS_AUTO` handling are duplicated.
- There seems to be some copied or unused state, such as `para_env` and `dav_cold_*` naming. Please have a look if they are still needed.
- The kernel contains avoidable temporary allocations, unclear scratch names, and repeated projection code.
- The real and complex Lanczos implementations are already diverging significantly. Please keep future unification in mind.

src/qs_scf_diagonalization.F
```diff
@@ -2946,6 +2958,454 @@ SUBROUTINE do_block_davidson_diag_kp(qs_env, scf_env, matrix_ks_kp, matrix_s_kp,
...
+         t1 = m_walltime()
+         CALL lanczos_refinement_2v_c(cksmat, cmo, cwork, mo_eigenvalues, nao, nmo, eps_col, &
+                                      scf_env%krylov_space%nkrylov, scf_env%krylov_space%nblock, &
+                                      max_res, min_res, nmo_conv, nmo_nc)
+         t2 = m_walltime()
+         IF (print_row .AND. output_unit > 0) THEN
+            WRITE (output_unit, '(T8,I3,T16,I5,T24,I6,T33,E12.4,2x,E12.4,T60,F8.3)') &
+               ispin, iter, nmo_conv, max_res, min_res, t2 - t1
+         END IF
+         IF (max_res < eps_iter) THEN
```
RTChou67: `lanczos_refinement_2v_c` determines convergence per column using `eps_col`. With `EPS_ADAPT` > 0, an empty state is allowed to converge at `eps_iter_empty = 5 * eps_iter`.

For example, if all occupied residuals are below `eps_iter` and an empty-state residual is `3 * eps_iter`, the kernel correctly returns `nmo_nc == 0`, since that residual is below `eps_iter_empty`. However, `max_res` is still `3 * eps_iter`, so the driver does not exit here.

src/qs_scf_lanczos_c.F
```diff
@@ -0,0 +1,702 @@
...
+            ! spectral fallback: clamp the Gram spectrum at a floor and
+            ! apply the inverse square root. A clamped direction keeps a
+            ! residual off-diagonal overlap of order sqrt(g/gfloor) after
+            ! one pass. A second pass squares it down. Per-column
+            ! renormalization is deliberately avoided: it would rescale a
+            ! clamped direction back to unit norm and reimport the very
+            ! overlap the clamp bounds. Normalization is not
+            ! re-orthogonalization
+            ALLOCATE (g_eval(nmo), g_scale(nmo))
+            DO ipass = 1, 2
```
RTChou67, line 227: I don't think this spectral fallback can handle the rank-deficient case described in the comment above. If the current MO matrix C is actually rank deficient, each pass here only applies a finite right transformation,`C_new = C M` where M is constructed from the eigenvectors of `C^H C` and the clamped inverse square roots. Therefore `rank(C_new) <= rank(C)`.

No number of these passes can turn a rank-deficient `C` into a full-column-rank matrix satisfying `C^H C = I`. In particular, an exactly zero Gram eigenvalue remains associated with a missing direction; clamping avoids the singular inverse but cannot create that direction.

RTChou67 (changes requested):

src/qs_scf_diagonalization.F
s8ga: Good catch, this is a real bug in the exit logic.

The kernel judges every column against its own threshold, so the scenario you describe returns zero pending columns while the largest residual still sits at three times the occupied threshold. The sweep loop compared that largest residual against the occupied threshold, so the run kept sweeping on a residual that had already passed, and ended with a spurious not-converged warning.

Fixed in a40d0f7: the loop now exits on the pending-column count, which is zero exactly when every column passed its own threshold.

src/qs_scf_lanczos_c.F
s8ga: You are right about the math: a right transformation preserves the rank, and the clamped inverse leaves a missing direction near zero norm instead of creating it.

The fallback never claimed to repair a rank-deficient set, but the comment did leave that scope open, and it now states it. The pass exists for a full-rank set whose Gram went numerically indefinite: the write-back block is orthonormal and deflated against the leading prefix, the columns beyond the pass keep their entry vectors, and mixing the two groups can push eigenvalues of the Gram below roundoff without losing a direction. For that case the clamped inverse square root restores orthonormality to working precision.

Genuine rank deficiency cannot arise from the construction, short of cancellation beyond roundoff: the entry MOS are orthonormal and every written column is a new orthonormal direction. If it ever did arise, the pass would collapse the missing direction toward zero norm rather than crash, and the next sweep re-purifies the columns from fresh seeds.

s8ga:
> Would need more work on cleanup before this is ready:
> 
> * There is a lot of repeated `cp_fm_struct_create`/`cp_cfm_create` boilerplate. Please reuse existing matrix structures where possible.
> * Scratch matrices are allocated and released on every Lanczos sweep. Please introduce a reusable workspace.
> * `lanczos_refinement_2v_c` exposes several dimensions and scratch objects that can be derived internally. Please reduce the interface.
> * `krylov_space_type` seems to mix settings with the old real-Lanczos workspace, which leads to partially initialized objects.
> * The k-point Lanczos driver duplicates a fair amount of Davidson setup and finalization logic.
> * The local cold-start and refinement paths contain duplicated control flow, please consolidate them.
> * Some cold-start eigensolver logic duplicates existing complex diagonalization infrastructure.
> * Tolerance, occupation, density, and `ADDED_MOS_AUTO` handling are duplicated.
> * There seems to be some copied or unused state, such as `para_env` and `dav_cold_*` naming. Please have a look if they are still needed.
> * The kernel contains avoidable temporary allocations, unclear scratch names, and repeated projection code.
> * The real and complex Lanczos implementations are already diverging significantly. Please keep future unification in mind.

Thanks for the review.
All eleven points are addressed.

1. The repeated structure boilerplate is gone.
The kernel now takes a caller-owned workspace, created once per diagonalization call at the worst-case widths of the run.
Every sweep of every channel reuses it.
The per-block Rayleigh-Ritz pair keeps its right-sized allocation, because the solver stack diagonalizes whole matrices and knows no leading-block argument.

3. Same workspace answers the per-sweep allocations.
It embeds into the kpoint driver workspace, so the local and the transfer-round layout share one allocation path.
The six bare channel matrices of the rounds path disappear.

4. The kernel interface is optimzed from 14 to 11 arguments.
The matrix dimensions follow the MOS argument, and the scratch column moved into the workspace.
The Ritz scratch of each block stays at the right size of the block, so every eigenvector copy crosses no matrix layouts.

5. The type is split.
`krylov_settings_type` holds the parsed settings and is created for every block-Krylov run, read by the Gamma and k-point drivers alike.
`krylov_workspace_type` keeps the statistics and the Gamma buffers, is allocated only by the Gamma path, and carries a pointer to the shared settings.
The partially initialized object is gone.

6. The Lanczos and Davidson kpoint drivers share one setup and finalization routine now.
That covers the global scratch, the operator context and the group matrix structure, with a flag for the lazy snapshot of the transfer rounds.

7. The local cold start folded into the refinement loop.
Both layouts now run the same per-channel shape, and the duplicated operator evaluation is gone.

8. The cold start itself already went through the shared `cold_start` helper from the Davidson PR.
What duplicated inside the kernel was the Rayleigh-Ritz pass, the entry pass and every refinement block now share one routine.

9. One exit chain helper covers occupations, `ADDED_MOS_AUTO` and the density, shared by the Lanczos, Davidson and OT kpoint paths.
The adaptive tolerance became a pure function shared by the Gamma and kpoint Lanczos drivers.
The Davidson ratchet stays where it was, it is a deliberate design there.

10. `para_env` removed.
The `dav_cold_*` tags renamed to a neutral name, the routine serves both drivers.

11. The temporary arrays folded into the workspace, and the scratch names describe their roles now.
The repeated projection code consolidated into the shared pass of point 7.

12. On unification, the complex kernel now has the (settings, workspace, statistics) shape that the Gamma path could migrate to.
The algorithmic cores stay separate on purpose.
The complex side replaced the latching write-back and the three-term bookkeeping that the Gamma kernel still carries, so a Gamma-side migration would be a behavior change and its own PR.

Growl1234:
Why does the complex implementation abandon the block three-term recurrence used by the existing Lanczos implementation and instead build a fully reorthogonalized subspace with an explicit dense Rayleigh-Ritz projection? This appears to be the main reason why the supposedly simpler Lanczos path has grown to roughly Davidson-level complexity. Is this additional machinery required by a demonstrated numerical issue, or is it mainly compensating for missing complex-matrix infrastructure such as `cp_cfm_qr_factorization`?

s8ga:
@Growl1234 fair question — I'm converting this PR to draft while I sort this out properly.

I found (and I'm fixing separately) an uninitialized-memory bug in the existing real-space Lanczos kernel (qs_scf_lanczos.F, lanczos_refinement_2v): a rank owning zero rows of a block work matrix reads uninitialized heap, which depending on the process grid / nao combination — either silently corrupts the result or aborts with PDSYEVD ... INFO=17. pr: #6090 .

The numerical comparisons I used to justify moving to full reorthogonalization with an explicit dense Rayleigh-Ritz projection were benchmarked against the real-space kernel as a reference,  and that reference turned out to be buggy. So right now I honestly don't know whether the added machinery was solving a real problem or compensating for noise from this bug. I need to redo the whole evaluation against a fixed, audited baseline before I can defend the design.

This PR has already gone through a lot of rounds and taken up a good amount of both of your time,  I'm sorry for that, and I don't want to keep asking for review cycles on a design I can't fully stand behind yet. Moving it to draft until the real-space fix lands and I've rerun the comparison properly; I'll ping this thread once I have something solid to bring back.

PR #6053, opened 2026-09-18
Grid: Fix GPU task-list cleanup guards

DCM-Uni-Paderborn:
With CMake CUDA/HIP builds, `grid_create_task_list()` creates a GPU context, but `grid_free_task_list()` skips its cleanup when only `__OFFLOAD_CUDA` or `__OFFLOAD_HIP` is defined. The cleanup checks `__OFFLOAD`, which is normally derived in `offload_runtime.h`; this C translation unit does not include that header. The GPU context and its resources are therefore leaked when the task list is freed.

Use the same CUDA/HIP and `__NO_OFFLOAD_GRID` condition for cleanup as for creation and storage of the GPU task list.

### Validation

- Compiled the original and patched C translation unit in eight configurations: CPU, CUDA, HIP, CUDA/HIP with explicit `__OFFLOAD`, CUDA/HIP with `__NO_OFFLOAD_GRID`, and OpenCL.
- Inspected the object files: the reference to `grid_gpu_free_task_list` is missing before the fix and present afterwards for CUDA/HIP without explicit `__OFFLOAD`. The other configurations retain the expected behavior.
- CP2K per-file make-pretty checks and `git diff --check` pass. The make-pretty pipeline was invoked with a relative path because the wrapper's absolute-path handling triggers unrelated filename/license checks in this checkout.
- CUDA/HIP runtime testing was not available locally.


oschuett (approved):

mtaillefumier (commented):
I am not sure to understand why this change is necessary. the GPU backend is only available when grid offloading is on independently of the gpu vendor. 
there is a transitive dependency as well. When the grid backend is included, __OFFLOAD etc are automatically ON.

DCM-Uni-Paderborn:
@mtaillefumier Thanks for checking this. Agreed that the GPU backend requires offloading. The issue is the visibility of `__OFFLOAD` in this particular C translation unit, rather than whether the backend is enabled.

The [CMake GPU definitions](https://github.com/cp2k/cp2k/blob/d0f067083a602c675fac0d7ad7b48c7e3a306d8a/src/CMakeLists.txt#L1721-L1729) set `__OFFLOAD_CUDA` or `__OFFLOAD_HIP`. The generic `__OFFLOAD` is derived in [offload_runtime.h](https://github.com/cp2k/cp2k/blob/d0f067083a602c675fac0d7ad7b48c7e3a306d8a/src/offload/offload_runtime.h#L25-L33), but that header is not in the include chain of `grid_task_list.c`: `grid_task_list_internal.h` includes `grid_gpu_task_list.h`, which includes `offload_buffer.h`, not `offload_runtime.h`. A definition in another translation unit does not propagate to this one.

I rechecked the exact parent and PR revisions by compiling this file and inspecting its undefined symbols. For example, on the parent revision:

```sh
gcc -std=c11 -fopenmp -D__OFFLOAD_CUDA -c src/grid/grid_task_list.c -o /tmp/grid_task_list.o
nm -u /tmp/grid_task_list.o
```

There is a reference to `grid_gpu_create_task_list`, but none to `grid_gpu_free_task_list`. With the patch, both references are present. The same holds for HIP. The preprocessor macro dump also confirms that `__OFFLOAD` is absent in these two cases.

Builds which explicitly define `__OFFLOAD` already include both calls and remain unchanged. CPU, disabled GPU-grid offloading, and OpenCL also retain their expected behavior in the compile checks. Thus this only makes cleanup use the same condition as GPU task-list storage and creation; it does not change backend availability. These are compile/preprocessor checks, not CUDA/HIP runtime tests.

Issue #792, opened 2020-03-04
f77_interface checking for convergence of HF calculation


Pibemanden:
I have been using the f77_interface to interface an MD code to CP2K with great success so far.
The only issue I have run into is that currently I cannot detect if the HF calculation converged or not. Meaning that I usually have to babysit the first couple of iterations + check the whole calculation in the end to make sure that all steps taken were valid.

Is it possible to add a logical or similar that one can pull out of the interface to check if the HF calculation has actually converged, as currently the logical that checks for this is confined to be local in the HF loops?

juerghutter:
Hi 

this is a good idea and we will put it on the list, but I cannot
say when it will be ready.

In the mean time you can have a look at this input keyword 
that allows you to stop on non-convergence
 
CP2K_INPUT / FORCE_EVAL / DFT / SCF / NOTCONV_STOPALL

best regards

Juerg
--------------------------------------------------------------
Juerg Hutter                         Phone : ++41 44 635 4491
Institut für Chemie C                FAX   : ++41 44 635 6838
Universität Zürich                   E-mail: hutter@chem.uzh.ch
Winterthurerstrasse 190
CH-8057 Zürich, Switzerland
---------------------------------------------------------------

-----"Pibemanden" <notifications@github.com> wrote: -----
To: "cp2k/cp2k" <cp2k@noreply.github.com>

Pibemanden:
Hi Juerg,

Thank you for your quick reply.

I tried the keyword you suggested, but sadly it only works internally in CP2K. So it just produces a second warning that the MD run will stop, but in the end I am just doing single point calculations and there is nothing stopping me from calling the `calculate_energy_force` subroutines again from my code.

I found a slightly hacky solution to the problem however, I check the `warning_counter` variable from the `cp_error_handling` module. As the non converging HF calculation will trigger an warning to be added to the program. It is of course not an ideal solution, but for now it fixes my problem.

I can also try to create the interface function, I just need to know where the logical stating if SCF has converged or not should be put, since I am no expert on the cp2k layout. As far as I can tell the `qs_scf_env_type` could be a good candidate.

dev-zero:
@Pibemanden can you possibly share your code (just the cp2k interface would be enough) and some more info (e.g. which compiler and Fortran standard you are using)? The f77_interface is something which has to be refactored eventually and it would be good to know what we would have to provide as an alternative.

Pibemanden:
[...]

PR #6060, opened 2026-09-18
Add a direct Python interface to libcp2k with ASE and MPI support

DCM-Uni-Paderborn:
## Summary

Add a standalone, installable Python package that calls libcp2k directly within the Python process, without starting cp2k_shell or parsing energy/force output. ASE is an optional integration, not a requirement for using CP2K from Python.

CP2K was already usable from Python through ASE's persistent shell calculator and AiiDA, and previously had Cython bindings. This PR adds a different direct embedding path; it is not the first Python interface and does not claim a general performance advantage over ASE's existing backend.

The package provides:

- Reusable force environments with NumPy access to energies, forces, positions, cells, and velocities.
- CP2K input supplied as a file, string, or nested Python dictionary. 
- Complete native calculations through `run_input`, including MD and geometry optimization.
- An optional ASE calculator with unit conversion, result caching, and support for fixed-cell optimization and MD.
- Optional mpi4py integration, including subcommunicators and explicit ownership of MPI initialization and finalization.

The implementation uses ctypes and requires no compiled Python extension. NumPy is required; ASE and mpi4py are optional. A shared libcp2k installation is provided separately.

## Example

```python
from cp2k import CP2K

with CP2K() as cp:
    with cp.create_force_env(
        input_file="h2.inp", output_file="h2.out"
    ) as system:
        result = system.calculate()
        print(result.energy)  # hartree
        print(result.forces)  # hartree/bohr
```

## Native library fixes

Integration testing exposed two issues in complete library-driven runs:

1. `run_input` did not open a requested output file unless a Fortran caller had already opened it. It now opens and closes files it owns, preserving append behavior and caller-owned output units.
2. Complete runs finalized DBCSR while the embedding library runtime was still active. Library-driven runs now retain DBCSR until runtime finalization, allowing subsequent force calculations.

Both fixes have regression coverage in the native C-interface test and the Python integration tests. The initial native fixes and Python interface are separate commits. The review follow-up adds the convergence query and its tests.

## Review follow-up

- Add `cp2k_get_scf_convergence`: `1` converged, `0` not converged, `-1` unavailable. It reports ordinary Quickstep inner/outer SCF convergence, including the CDFT constraint loop, and invalidates status after geometry/velocity changes. CDFT now explicitly returns its convergence result instead of leaving the caller's result undefined.
- Python exposes `result.scf_converged` and raises `SCFConvergenceError` by default on reported failure; `check_convergence=False` is an explicit diagnostic opt-out. ASE translates failures to `CalculationFailed`. Unsupported solvers and older libraries return `None`, not a success claim. Native CP2K abort behavior is unchanged.
- Add side-by-side ASE syntax and a reproducible common-input comparison. Both backends retain their force environment; neither starts CP2K anew for every evaluation.
- Correct the direct ASE adapter's unit conversions to use CP2K's constants, as its shell protocol does. This removes an approximately 2.45 micro-eV discrepancy in the H2 example; no comparison tolerance was loosened.
- Document pure-Python wheel distribution versus separately installed native dependencies. A `py3-none-any` wheel and sdist build and pass `twine check`; no PyPI release has been published. Release ownership is explicitly subject to maintainer agreement.
- Extend the existing Linux ASE CI job to run this package's tests explicitly, separately from CMake.

Two local one-rank/one-thread H2 comparisons with 12 changed geometries gave median warm evaluations of 0.176-0.180 s for the persistent shell and 0.171-0.182 s for the direct library. Maximum differences were 5.4e-13 eV in energy and 1.9e-9 eV/angstrom in forces. These measurements show no consistent steady-state speed advantage and are not production-performance claims.

The updated C interface tests passed with one and two MPI ranks. WORLD, split-communicator and implicit-mpi4py smoke tests passed. The native status tests include converged OT/diagonalization/outer SCF, deliberately truncated SCF, converged and unconverged CDFT constraints, and unavailable status at MAX_SCF 0. Formatting/precommit checks pass. The new remote CI run remains pending.

## Lifetime and supported scope

The wrapper validates array shapes, finite values, runtime state, and result availability.

One runtime may be initialized per process, with one live force environment at a time. Sequential environments are supported. Overlapping environments are rejected because native teardown currently frees shared integral tables.

Calls must originate from the main Python thread. Native CP2K aborts can still terminate the interpreter.

ASE stress/cell optimization and a native OpenMM Force plugin are not included.

## Validation

Locally validated on macOS/arm64:

- All 56 Python tests passed, including native SCF status cases and a shell-versus-library comparison. Without native libraries, 43 pass and 13 intentionally skip.
- All five native integration tests also passed with two OpenMP threads.
- Forces checked against finite-difference energies.
- Triclinic cells, ASE optimization, sequential environments, repeated complete runs, and a short native MD trajectory tested.
- Two-rank MPI tests passed for explicit WORLD, a subcommunicator, and automatically detected mpi4py initialization.
- Native C-interface regression tests passed with one and two MPI ranks.
- Command-line output-file behavior verified.
- Wheel and source distribution built and package installation tested.
- CP2K precommit checks and `git diff --check` passed.

This is targeted local validation, not a full CP2K regression-suite or cross-platform CI run.

Documentation, runnable examples, and installation instructions are included in `python/README.md` and the CP2K manual.

## Release preparation (no publication)

Add `python/RELEASING.md` and `python/tools/check_release.sh`: build an sdist/wheel, check metadata, test the installed wheel in an isolated environment, and record SHA-256 hashes and a JUnit report. Python 3.12 and 3.14 each pass 43 non-native tests (13 native tests deselected). Existing output directories are rejected. No upload, release tag or credentials are created. Naming/versioning, native compatibility, ownership and approval gates are proposals pending agreement. This does not bundle libcp2k or replace native CI.


CMakeLists.txt
```diff
@@ -1019,6 +1019,21 @@ include(GNUInstallDirs)
...
+      "${CMAKE_COMMAND}" -E env "PYTHONPATH=${PROJECT_SOURCE_DIR}/python/src"
+      "CP2K_TEST_LIBRARY=$<TARGET_FILE:cp2k>" "CP2K_DATA_DIR=${CP2K_DATA_DIR}"
+      "OMP_NUM_THREADS=1" "OPENBLAS_NUM_THREADS=1" "${Python_EXECUTABLE}" -m
+      pytest "${PROJECT_SOURCE_DIR}/python/tests"
+      "--basetemp=${PROJECT_BINARY_DIR}/python-tests"
+    DEPENDS cp2k
+    WORKING_DIRECTORY "${PROJECT_BINARY_DIR}"
+    USES_TERMINAL VERBATIM)
+endif()
+
```
Growl1234: Is this really "opt-in"? Let's make this test command seperate from top-level CMake.

python/README.md
```diff
@@ -0,0 +1,207 @@
...
+From a CP2K source checkout:
+
+```sh
+python -m pip install './python[ase,mpi]'
+export CP2K_LIBRARY=/absolute/path/to/libcp2k.so  # libcp2k.dylib on macOS
+export CP2K_DATA_DIR=/absolute/path/to/cp2k/data
+export OMP_NUM_THREADS=1
+```
+
+Use a CP2K build configured with `-DBUILD_SHARED_LIBS=ON`. All its dependent libraries must be
```
Growl1234: Just curious, can a static library with PIC enabled work?

CMakeLists.txt
DCM-Uni-Paderborn: Addressed in 221434800b. I removed the Python test target from the top-level CMake build; the README now documents explicit pytest invocation against the built shared library, with the Python dependencies installed separately. The 37 unit tests pass. I also fixed the manual CI packaging: the Dockerfile and its generator now copy python/README.md, which the new documentation page includes.

python/README.md
DCM-Uni-Paderborn: Not directly: ctypes loads a shared object through the platform dynamic loader, not a static .a archive. PIC makes the archive objects suitable for linking into a shared library, but that extra link step must still export the CP2K C API and resolve the required dependencies. A shared wrapper linked against a PIC static CP2K build could therefore work, but PIC alone is insufficient. I clarified this distinction in the README in 221434800b.

DCM-Uni-Paderborn:
The integration follow-up is now available as #6061. It is based exactly on this PR’s current head (221434800b), with four additional commits covering stress/ASE, PLUMED, i-PI, and the OpenMM/LAMMPS adapters. Keeping those changes in the follow-up preserves the scope of this PR. #6061 explicitly depends on this PR being merged first and includes an integration-only comparison link.

Growl1234 (approved):
I support adding Python interace. Further validation and feedback would be needed in more pratical user cases.

oschuett:
> Add a standalone, installable Python package that calls libcp2k directly within the Python process, ...

We actually just deleted our Python bindings for libcp2k #5541, because they were too cumbersome to build and therefore unused.

The main selling point of Python is convenience and a key part of that is easy installation via PyPI. A package that links directly to libcp2k can not be distributed via PyPI (unless we also package libcp2k as a [wheel](https://pythonwheels.com/), which is another can of worms).

I think, what's holding back our Python eco-system is not a lack of code, but a lack of maintainers that curate and publish that code. So, I'd suggest to contribute to the Python packages that already exist:

- https://gitlab.com/ase/ase/-/blob/master/ase/calculators/cp2k.py
- https://github.com/cp2k/cp2k-input-tools
- https://github.com/cp2k/cp2k-output-tools
- https://github.com/aiidateam/aiida-cp2k

he-zilong:
I would like to see more about the ASE interface in the documentation: a side-by-side comparison between the old workflow with cp2k shell subprocess and the new workflow with the libcp2k C API in terms of the syntax and performance.

As noted in #792 (which is for the legacy F77 interface that #1231 has suggested to drop), a method to check the convergence of QuickStep DFT calculation from the interface would be nice to have.

Growl1234:
I'm fine with either of both ways. Going to those Python packages, I would think at least `cp2k-input-tools` and `cp2k-output-tools` should be merged into one Python package.

DCM-Uni-Paderborn:
Dear Ole, Zilong, and all,

Thank you for the constructive feedback. I agree that maintainability and straightforward installation are essential, particularly given the experience with the previous bindings.

One technical clarification: this implementation uses ctypes, without a compiled Python extension. The Python package can therefore be distributed through PyPI while loading a separately installed shared libcp2k at runtime. This does not make CP2K itself pip-installable, and building or obtaining a compatible shared library remains a genuine installation requirement. We should document that distinction clearly and agree on maintenance, release ownership, and CI coverage.

The motivation also extends beyond providing another ASE calculator. The direct interface exposes reusable force environments and MPI communicator control for embedding CP2K in other applications. The follow-up #6061 provides concrete examples through OpenMM and LAMMPS adapters, including a separate-process backend for independent MPI configurations. Improving the existing ASE calculator alone would not provide that common foundation. The native PLUMED and i-PI fixes in #6061 are independently useful.

Zilong, I agree that a side-by-side ASE example and a small, reproducible performance comparison would help. The existing ASE interface already keeps CP2K running between evaluations, so we should not imply that avoiding repeated process startup is an advantage over that implementation. Any performance benefit needs to be measured; direct embedding and communicator control are the clearer architectural differences.

Your convergence-status suggestion is also valuable. The current wrapper checks that returned energies and forces are finite, but this does not establish SCF convergence. An explicit status exposed through the native API would be preferable to parsing output or inferring success from numerical values.

For #6062, I think Ole's suggestion is particularly relevant: the AiiDA helper already uses aiida-cp2k and aiida-common-workflows, not libcp2k. Contributing the relevant improvements and tests to those projects may be the better long-term home.

My preference would therefore be to retain a small, well-tested libcp2k binding as the common foundation, while coordinating higher-level integrations with the existing projects. Would that division of responsibilities address your main concerns?

Greetings,
Thomas

DCM-Uni-Paderborn:
Dear Ole, Zilong, and all,

The review follow-up is now pushed in 2359526221.

The documentation now explicitly distinguishes this direct embedding API from the existing ASE and AiiDA routes. It includes side-by-side ASE syntax, a common-input comparison script, and the distinction between a distributable pure-Python wheel and the separately installed native CP2K library. Maintenance/release ownership is still a question for agreement, not something the implementation resolves by itself.

The comparison also exposed a small unit-conversion mismatch: the direct ASE adapter now uses CP2K's constants, matching its shell protocol. Two local 12-geometry H2 comparisons gave maximum differences of 5.4e-13 eV and 1.9e-9 eV/angstrom. Warm evaluation medians overlap (shell 0.176-0.180 s; direct 0.171-0.182 s), so there is no consistent speed advantage in this example.

An explicit native SCF status is now available: converged, not converged, or unavailable. Python rejects a reported failure by default, with an explicit diagnostic opt-out; unknown status is not treated as evidence of convergence. Tests cover OT, diagonalization, outer SCF, CDFT constraints, unavailable status and result invalidation. This does not change native abort handling or certify subsequent optimization/post-SCF convergence.

All 56 core Python tests pass locally, alongside native C and two-rank communicator checks. The shared changes are synchronized into #6061 and #6062, and their descriptions report their own validation. For #6062, a possible upstream home in the existing AiiDA projects is documented as a proposal pending discussion; the implementation and tests are retained.

The updated remote CI is still pending.

Greetings,
Thomas

DCM-Uni-Paderborn:
Dear Ole, Zilong, and all,

To make the maintenance question concrete before merge, could we agree on the following release responsibilities?

- A primary package/release maintainer and a backup, with responsibility for Python/native API compatibility and failed integration CI.
- Project ownership and publishing access under an arrangement agreed by the CP2K maintainers. The proposed distribution name is cp2k-python (import cp2k), but no PyPI project has been registered or reserved.
- A documented supported native CP2K revision/release and installation route. The initial wheel contains only the wrapper; it does not yet provide an OpenMM-style installation bundling the native engine.
- Separate approval for the first TestPyPI candidate and the production PyPI upload, after merge and validation of the actual merged code.

The technical preparation is now in python/RELEASING.md and an upload-free check_release.sh. It builds an sdist/wheel, tests the installed wheel in a fresh environment and records hashes/test reports. It passes with Python 3.12 and 3.14 (43 non-native tests each). There is no publishing job, credential setup or release tag.

For AiiDA, I have also prepared a small local patch directly against aiida-common-workflows, correcting the protocols without depending on this binding. Its protocol/builder tests and three real CP2K workflows pass. The native final-cell-frame fix remains CP2K's responsibility. No upstream PR or ownership transfer has been made.

Would this division of responsibilities be acceptable, and who should hold the package/release roles? I would like to settle that explicitly rather than assume that adding the code assigns ongoing maintenance to anyone.

Greetings,
Thomas

python/README.md
```diff
@@ -0,0 +1,317 @@
...
+| Native installation | CP2K executable with shell mode                   | Shared libcp2k and its runtime dependencies            |
+| Input               | ASE parameters and optional CP2K text template    | Explicit CP2K mapping; geometry from ASE               |
+| Lifetime            | Calculator owns a persistent child process        | Caller owns the runtime and its MPI communicator       |
+| Data transfer       | Text protocol over stdin/stdout                   | C API and NumPy arrays                                 |
+| MPI                 | Launch command, e.g. `mpiexec -n 2 cp2k.psmp -s`  | Collective calls on a caller-owned mpi4py communicator |
+| Native abort        | Child process fails; parent is a separate process | Can terminate the Python interpreter                   |
+
+The adapter uses CP2K's unit-conversion constants, matching its shell protocol. ASE's default CODATA
+constants differ slightly; mixing the two would introduce a systematic energy/geometry offset even
+when the underlying CP2K calculation is identical.
```
he-zilong: A side note: the physical constants defined in `src/common/physcon.F` used to have a preprocessor flag `CODATA_VERSION` for switching CODATA versions, until commit abc0f0c dropped it for a fixed set of PARAMETERs based on the legacy CODATA 2006. Personally I would like to have an input keyword, say `&GLOBAL/CODATA_VER`, that allows users to choose one of the versions whose data can be updated regularly. Then it would get its interface for compatibility.

src/start/cp2k_runs.F
```diff
@@ -464,7 +467,9 @@ RECURSIVE SUBROUTINE cp2k_run(input_declaration, input_file_name, output_unit, m
          CALL globenv_release(globenv)
       END IF
 
-      CALL dbcsr_finalize_lib()
+      finalize_dbcsr = .TRUE.
```
oschuett: DBCSR already gets initialized/finalized in [f77_interface.F](https://github.com/cp2k/cp2k/blob/master/src/f77_interface.F) together with all the other libraries. So, these calls to `dbcsr_init_lib` and `dbcsr_finalize_lib` don't really belong here. Presumably they are a left over workaround from long ago. We should try to clean it up instead of  piling on.

src/start/cp2k_runs.F
```diff
@@ -965,12 +974,21 @@ SUBROUTINE run_input(input_declaration, input_file_path, output_file_path, initi
...
-            INQUIRE (FILE=output_file_path, NUMBER=unit_nr)
+            INQUIRE (FILE=output_file_path, OPENED=opened, NUMBER=unit_nr)
+            IF (.NOT. opened) THEN
+               CALL open_file(file_name=output_file_path, &
+                              file_status="UNKNOWN", &
+                              file_action="WRITE", &
+                              file_position="APPEND", &
+                              unit_number=unit_nr)
+               owns_output = .TRUE.
+            END IF
```
oschuett: I'm surprised this is needed. How did libcp2k and the cp2k-shell work without this in the past?

src/start/libcp2k.h
```diff
@@ -191,6 +191,20 @@ void cp2k_calc_energy_force(force_env_t force_env);
...
+ * \param force_env the force environment
+ * \param status -1 if unavailable, 0 if not converged, 1 if converged
+ * \note Unavailable before calculation or after changing positions, cell or
+ *       velocities, for non-Quickstep methods, and for alternative solvers
+ *       (e.g. LS-SCF, ALMO, RTP, non-SCF or MAX_SCF 0). This is not a
+ * convergence certificate for post-SCF methods, geometry optimization or MD.
+ *       IGNORE_CONVERGENCE_FAILURE allows an unconverged SCF to return; this
+ *       query does not prevent native CP2K aborts when that keyword is absent.
+ ******************************************************************************/
+void cp2k_get_scf_convergence(force_env_t force_env, int *status);
```
oschuett, line 206: I'd suggest, we move extensions of libcp2k to a separate PR.

src/start/libcp2k_unittest.c
```diff
@@ -9,6 +9,9 @@
 #include <math.h>
 #include <stdio.h>
 #include <stdlib.h>
+#if defined(__parallel)
+#include <mpi.h>
+#endif
```
oschuett: You can use [mpiwrap/cp_mpi.h](https://github.com/cp2k/cp2k/blob/master/src/mpiwrap/cp_mpi.h).

src/f77_interface.F
```diff
@@ -171,6 +171,7 @@ MODULE f77_interface
 ! **************************************************************************************************
    TYPE f_env_type
       INTEGER :: id_nr = 0
+      LOGICAL :: has_result = .FALSE.
```
oschuett: Personally, I'd rather call this `is_dirty` because the result will still be there even after changing the positions. The term _"result"_ is also a bit overloaded because of [cp_result_methods.F](https://github.com/cp2k/cp2k/blob/master/src/common/cp_result_methods.F)

src/qs_environment_types.F
```diff
@@ -263,6 +263,7 @@ MODULE qs_environment_types
       TYPE(qmmm_env_qm_type), POINTER                       :: qmmm_env_qm => NULL()
       TYPE(qs_wf_history_type), POINTER                     :: wf_history => NULL()
       TYPE(qs_scf_env_type), POINTER                        :: scf_env => NULL()
+      LOGICAL :: scf_convergence_available = .FALSE., scf_converged = .FALSE.
```
oschuett, line 266: Again, let's move this new important feature to a separate PR so it receives the attention it deserves.

oschuett:
I certainly see that there is a new generation of Python projects that we want to be part of. Using ctypes might actually solve our installation conundrum and keeping it in our main repository should further help it to gain traction.

Regarding the push to PyPI, the key question is whether we want to make this part of CP2K's regular release schedule or push more often? In any case, we should leverage the CI improvements to automate this.

As for the package name, `cp2k-python` is fine, but I also reserved  [cp2k](https://pypi.org/project/cp2k/),  [libcp2k](https://pypi.org/project/libcp2k/),  and  [pycp2k](https://pypi.org/project/pycp2k/) a while ago.

Growl1234:
> Regarding the push to PyPI, the key question is whether we want to make this part of CP2K's regular release schedule or push more often?

I'd prefer to keep the two release schedules separate. It's very likely that the Python package has Python-only fixes and improvements that should not need to wait for a CP2K release, especially since the PyPI package does not ship `libcp2k` itself. What should be tied with CP2K releases is just the compatibility contract.

> As for the package name, `cp2k-python` is fine, but I also reserved [cp2k](https://pypi.org/project/cp2k/), [libcp2k](https://pypi.org/project/libcp2k/), and [pycp2k](https://pypi.org/project/pycp2k/) a while ago.

I vote for `pycp2k`.

src/start/cp2k_runs.F
DCM-Uni-Paderborn: Many thanks, Ole. I have removed the keep_dbcsr argument and moved the native work to #6078. Ordinary runs no longer finalize DBCSR: init_cp2k/finalize_cp2k own its lifetime. Configuration is centralized in f77_interface, retaining CP2K's hooks and chosen device. The per-run configuration call is still needed because DBCSR's existing init API updates the diagnostic output unit when already initialized; it does not reinitialize the library. FARMING remains an explicit exception: disjoint job communicators require per-job initialization/finalization and restoration of the parent context. All ten existing FARMING inputs passed locally, including six-rank captain/minion and restart cases; C-interface tests passed with one and two MPI ranks. Could you review this narrower native change in #6078?

src/start/cp2k_runs.F
DCM-Uni-Paderborn: The command-line path opens -o before calling run_input, so INQUIRE can find a valid Fortran unit there. A C caller passing an ordinary filename has not opened that file in the Fortran runtime. ASE's usual shell calculator uses LOAD/EVAL rather than RUN, so its normal path does not exercise this case. The fix opens/closes only an output file owned by run_input and preserves append behavior and caller-owned units; it is now isolated in #6078. The C test checks file creation, growth on a second call, and a subsequent force calculation in the same runtime.

src/start/libcp2k.h
DCM-Uni-Paderborn: Many thanks; the native extension and embedded-runtime fixes are now isolated in #6078, with no Python package or Python CI changes. #6060 is rebased onto that exact prerequisite commit. Please merge/review #6078 first; until then the master-based GitHub diff still includes the prerequisite. The remaining Python changes can be reviewed separately at https://github.com/DCM-Uni-Paderborn/cp2k/compare/libcp2k-native-status...python-library-interface .

src/qs_environment_types.F
DCM-Uni-Paderborn: Moved the SCF-status implementation, including the CDFT return values and native C-interface coverage, into the dedicated native PR #6078. The Python tests remain here and exercise the new API through that prerequisite. No convergence thresholds or reference energies were relaxed.

src/start/libcp2k_unittest.c
DCM-Uni-Paderborn: Done: the native C test now includes mpiwrap/cp_mpi.h and obtains the rank with cp_mpi_comm_rank(cp_mpi_get_comm_world()). This change is part of the extracted native PR #6078; the test passes with one and two MPI ranks.

src/f77_interface.F
DCM-Uni-Paderborn: Done: the state is named is_dirty and its polarity is inverted accordingly. It is set before a calculation and on position, cell, or velocity changes, and cleared only after a calculation returns. This native change is now in #6078.

python/README.md
DCM-Uni-Paderborn: Many thanks for the suggestion. A selectable CODATA version would affect CP2K's native scientific input and unit conversions beyond this Python interface, so I have kept it out of this PR. The adapter uses CP2K's existing constants consistently with its shell protocol; the shell/direct comparison tests remain unchanged. If a native CODATA-selection feature is added separately, the Python interface can expose it through the normal input dictionary without introducing a competing set of physical constants.

DCM-Uni-Paderborn:
Dear Ole and all,

Many thanks for the release-policy discussion. I agree with keeping the Python release cadence independent: Python-only fixes should not wait for a CP2K release, while the native compatibility contract must remain explicit and tested. python/RELEASING.md now states this directly and points to #6078 for the native prerequisites.

The distribution name remains a maintainer decision. The documentation now acknowledges the proposed alternatives, including pycp2k, rather than implying that cp2k-python has already been agreed. The current name is unchanged pending that decision. Hence, no package, release tag, publishing credentials, or maintainer responsibilities have been created or assigned.

The preceding heads of #6060, #6061 and #6062 completed all requested Linux CI checks successfully, including the full regression suite and their ASE, i-PI and AiiDA jobs. Fabian’s new MPI-wrapper review fix from #6078 is now synchronized into all three branches, with the release clarification, and fresh CI is running for these updated heads. The wrapper was tested in serial and with four MPI ranks, including split/null communicators, and the updated C-interface test passes with two MPI ranks and two OpenMP threads. No numerical references or tolerances changed.

Greetings,
Thomas

DCM-Uni-Paderborn:
Dear @oschuett, @he-zilong and @RMeli,

Could you please revisit the Python/ASE interface when you have a chance? The native changes are now reviewed separately in #6078; that PR must land first. The side-by-side ASE example, common-input comparison and explicit SCF-status handling are included, and the current regular CI and conventions checks are green.

Feedback on the Python-only layer and the remaining package-name/release-ownership decisions would be particularly helpful. This is a request for review, not to bypass the prerequisite or those decisions.

Many thanks!
Greetings,
Thomas

tools/docker/generate_dockerfiles.py
```diff
@@ -493,8 +495,8 @@ def test_without_build(name: str) -> str:
 # ======================================================================================
 def print_cached_report() -> str:
     return r"""
-# Output the report if the image is old and was therefore pulled from the build cache.
-CMD cat $(find ./report.log -mmin +10) | sed '/^Summary:/ s/$/ (cached)/'
+# Always output the report: even a newly built image may be reused from the cache.
+CMD ["cat", "./report.log"]
```
oschuett: Yeah, I've also been pondering about this problem. Clearly, the 10 minute heuristic is not great.

I'll try to adopt the CI such that it will only run images when they were build very quickly. i.e. from cache.

As a side-note, far-reaching changes like this really deserve their own PR.

tools/docker/test_report.py
```diff
@@ -0,0 +1,62 @@
...
+import json
+import os
+import subprocess
+import tempfile
+import time
+import unittest
+from pathlib import Path
+
+
+class ReportTest(unittest.TestCase):
```
oschuett: What exactly is this test testing?

tools/docker/generate_dockerfiles.py
```diff
@@ -453,9 +454,10 @@ def precommit() -> str:
 
 # ======================================================================================
 def test_3rd_party(name: str) -> str:
+    python_sources = "COPY ./python ./python\n" if name == "ase" else ""
     return install_cp2k(profile="toolchain", version="ssmp") + rf"""
 # Run test for {name}.
-COPY ./tests ./tests
+{python_sources}COPY ./tests ./tests
```
oschuett: Let's leave the `test_3rd_party` as is and instead add a new dedicate tester.

tools/docker/generate_dockerfiles.py
```diff
@@ -424,6 +424,7 @@ def manual() -> str:
     return install_cp2k(profile="toolchain", version="pdbg", revision=True) + rf"""
 # Generate manual.
 COPY ./docs ./docs
+COPY ./python/README.md ./python/README.md
```
oschuett: Why does the manual require `python/README.md`?

tools/docker/scripts/test_ase.sh
```diff
@@ -57,11 +57,28 @@ echo -
...
 
-if ase test -j 0 -c cp2k calculator/cp2k; then
-  echo -e "\nSummary: ASE commit ${ASE_REVISION} works fine."
+if ! ase test -j 0 -c cp2k calculator/cp2k; then
+  echo -e "\nSummary: Something is wrong with ASE commit ${ASE_REVISION}."
+  echo -e "Status: FAILED\n"
+  exit 0
+fi
+
+echo -e "\n========== Direct Python Interface Tests =========="
```
oschuett: I think, we should add a new dedicated tester for the Python package.

tools/docker/generate_dockerfiles.py
DCM-Uni-Paderborn: Dear Ole,

Agreed on the separation. The existing ASE tester should continue testing upstream ASE's CP2K shell calculator unchanged. The direct Python package, including its own ASE adapter, belongs in a dedicated tester; test_3rd_party then needs no Python-specific special case. This also addresses your comments on test_ase.sh and test_3rd_party.

I also agree that changing the report-age policy for every generated image is outside the scope of the Python interface. That should be left to the CI change you outlined, rather than coupled to this PR. The current head still needs these CI changes to be separated; I am not treating the review points as implemented yet.

Many thanks for the review.

Greetings,
Thomas

tools/docker/test_report.py
DCM-Uni-Paderborn: Dear Ole,

This is a narrow regression test for the generated image's report-printing command, not a Python-interface or Docker-cache test. It extracts CMD from Dockerfile.test_conventions and runs that command locally against synthetic report.log files aged 0, 60, 660 and 86400 seconds. It checks that both OK and FAILED reports are reproduced exactly and that a missing report produces a nonzero exit status. It does not build or run a container, exercise the CI scheduler, or validate cache-hit detection.

It was added alongside the removal of the ten-minute report-age heuristic. With that generic CI change kept out of this Python PR, this test should leave this PR as well.

Greetings,
Thomas

tools/docker/generate_dockerfiles.py
DCM-Uni-Paderborn: Dear Ole,

The new docs/technologies/python.md page uses a MyST include of ../../python/README.md, starting at the Installation section. This keeps the package documentation and the corresponding manual page in one place. Copying only docs/ into the image leaves that include unresolved, so the manual image also needs the README source file.

This does not install or import the Python package, and does not add a Python-package runtime dependency to manual generation.

Greetings,
Thomas

oschuett: For this kind of situations we typically use symlinks into the docs directory, see e.g. 

https://github.com/cp2k/cp2k/blob/master/INSTALL.md

tools/docker/generate_dockerfiles.py
DCM-Uni-Paderborn: Many thanks, Ole. Implemented in e9af3e3fc6: the generic report-age policy change has been removed from this PR, including all unrelated generated-Dockerfile changes. The policy and existing testers now match master. I have left the cache-detection improvement to the separate CI work you described.

tools/docker/test_report.py
DCM-Uni-Paderborn: Removed from this PR together with the generic cached-report policy change. Neither test_report.py nor its Misc invocation remains in the diff; the Python-specific validation is in the new dedicated tester.

tools/docker/generate_dockerfiles.py
DCM-Uni-Paderborn: Implemented in e9af3e3fc6. test_3rd_party is unchanged from master. A separate test_python generator creates Dockerfile.test_python, with its own script, CI target and dashboard entry; it includes the package tests and the canonical documentation source needed by the README symlink.

tools/docker/generate_dockerfiles.py
DCM-Uni-Paderborn: Many thanks for pointing to the established layout. The canonical text now lives in docs/technologies/python.md, and python/README.md is a relative symlink to it. The manual no longer copies anything from python/. Strict standalone Sphinx validation passes; the sdist and wheel also build successfully, and the installed wheel passes all 43 non-native tests, so the source package does not require the repository-external symlink target after extraction.

tools/docker/scripts/test_ase.sh
DCM-Uni-Paderborn: Done in e9af3e3fc6: test_ase.sh and Dockerfile.test_ase are restored unchanged from master. The direct package and its ASE adapter are tested separately by the new python target. All 56 package tests pass locally against a fresh native build; the dedicated Linux run is now being requested.

DCM-Uni-Paderborn:
Review follow-up pushed in e9af3e3fc6, rebased onto the now-merged #6078. The inline replies and description summarize the dedicated Python tester, documentation symlink and removal of the unrelated cached-report changes.

@oschuett @Growl1234: the agreed direct is_dirty assignments are included here as the small follow-up cleanup (3cca556df7); the merged #6078 was not rewritten. All 56 Python tests and the native C/MPI checks pass locally. Fresh Linux CI is requested below. Package naming and release ownership remain maintainer decisions; no package has been published.

/cp2kci python ase manual conventions

tools/dashboard/dashboard.conf
```diff
@@ -167,6 +167,12 @@ host:        HZDR
 report_url:  https://storage.googleapis.com/cp2k-ci/dashboard_spack-psmp-ase_report.txt
 info_url:    https://docs.ase-lib.org/ase/calculators/cp2k.html
 
+[python]
+name:        Direct Python interface
+host:        GCP
```
oschuett:
```suggestion
host:        HZDR
```

tools/docker/cp2k-ci.conf
```diff
@@ -237,6 +237,15 @@ cache_from:   sdbg
 build_path:   /
 dockerfile:   /tools/docker/Dockerfile.test_ase
 
+[python]
+display_name: Python interface
+tags:         daily
+cpu:          8
+nodepool:     pool-main
+cache_from:   sdbg
```
oschuett:
```suggestion
nodepool:     pool-main
```

python/src/cp2k/ase.py
```diff
@@ -0,0 +1,193 @@
...
+    Stress, per-atom energies, shell models, topology files, and implicit use
+    of ASE initial charges/magnetic moments are not supported. Set charge and
+    spin explicitly in inp. Standard ASE geometry optimizers and fixed-cell
+    MD integrators can use this calculator. Close it before closing runtime.
+    """
+
+    implemented_properties = ["energy", "free_energy", "forces"]
+    default_parameters = {"inp": None}
+
+    def __init__(self, runtime, inp, *, output_file=None, **kwargs):
```
oschuett, line 54: At least for these public APIs, I'd suggest to add type annotations and check them with `mypy --strict`.

oschuett (approved):

mkrack (approved):

tools/dashboard/dashboard.conf
DCM-Uni-Paderborn: Many thanks, Ole. Corrected the dashboard host to HZDR in 63ae18c8c4.

python/src/cp2k/ase.py
DCM-Uni-Paderborn: Many thanks, Ole. I will add public API type annotations and strict mypy checking in the existing integration follow-up #6061, keeping this already-approved base PR limited to the reviewed implementation and the hostname correction. I am leaving this thread open until the follow-up implementation and checks are available.

DCM-Uni-Paderborn: Many thanks, Ole. The follow-up is now implemented and pushed in #6061 (2844d1b9b1): typed public package entry points, the ASE calculator and socket client; strict mypy in the dedicated Python tester; a py.typed wheel marker; and inexpensive positive/negative static API contract checks. Strict checking passes both against the source and an independently installed wheel. The existing native/integration suite still passes unchanged (106 passed, three PLUMED-only skips in the non-PLUMED local build), as do the WORLD/split/implicit MPI smoke tests. The narrow exceptions for upstream ASE's untyped methods and the typing scope are documented. Fresh Linux CI for #6061 is pending; I am leaving this thread open for review of that follow-up.

DCM-Uni-Paderborn:
The pdbg job stopped while downloading elpa-2026.02.002.tar.gz, before building CP2K or running any regression tests. The same URL currently responds with HTTP 200. Retrying this infrastructure/download failure on the unchanged head; no code, reference values or tolerances were modified.

/cp2kci pdbg

he-zilong:
> > As for the package name, `cp2k-python` is fine, but I also reserved [cp2k](https://pypi.org/project/cp2k/), [libcp2k](https://pypi.org/project/libcp2k/), and [pycp2k](https://pypi.org/project/pycp2k/) a while ago.
> 
> I vote for `pycp2k`.

Unfortunately `pycp2k` has already been used as the name of a third-party tool that gets mentioned on [cp2k.org/tools:pycp2k](https://www.cp2k.org/tools:pycp2k)...

Growl1234:
> Unfortunately `pycp2k` has already been used as the name of a third-party tool that gets mentioned on [cp2k.org/tools:pycp2k](https://www.cp2k.org/tools:pycp2k)...

It seems to be discontinued for many years, so I think it wouldn't hurt to treat this as an "unretirement" of it.

he-zilong:
> > Unfortunately `pycp2k` has already been used as the name of a third-party tool that gets mentioned on [cp2k.org/tools:pycp2k](https://www.cp2k.org/tools:pycp2k)...
> 
> It seems to be discontinued for many years, so I think it wouldn't hurt to treat this as an "unretirement" of it.

Perhaps we can contact its developers and integrate some functionality into `cp2k-{input,output}-tools`?

oschuett:
> Unfortunately pycp2k has already been used as the name of a third-party tool that gets mentioned on [cp2k.org/tools:pycp2k](https://www.cp2k.org/tools:pycp2k)...

Oh, I had forgotten that we already used  `pycp2k` for Lauri's package. So, we should definitely reach out to him before re-purposing it.

mkrack:
> Oh, I had forgotten that we already used `pycp2k` for Lauri's package. So, we should definitely reach out to him before re-purposing it.

Can we still use `py-cp2k`?

oschuett:
> Can we still use `py-cp2k`?

I think, we should try to reuse `pycp2k`. It's a very fitting name and one could then write `import pycp2k`.

I'll reach out to Lauri. I would be surprised if he objects.

DCM-Uni-Paderborn:
Dear Ole, Zilong and all,

Many thanks for clarifying the existing pycp2k project, and Ole, for contacting Lauri. I have updated the release preparation in #6061 so that both the distribution and import names remain provisional. Reusing pycp2k should be coordinated with its existing maintainer and any migration documented; it should not imply that the new ctypes/libcp2k binding is API-compatible with the existing project.

For now the code remains cp2k-python with import cp2k. No package has been renamed or published. The follow-ups remain independently reviewable while that discussion proceeds.

Greetings,
Thomas

PR #6061, opened 2026-09-18
Complete Python and MD integrations with stress, OpenMM, LAMMPS, PLUMED and i-PI

DCM-Uni-Paderborn:
## Summary

Complete the integration follow-up to #6060: expose potential stress/virial to C, Python and ASE, add direct OpenMM and LAMMPS Python adapters, and correct/harden the existing PLUMED and i-PI interfaces. HOOMD-blue and additional quantum-chemistry engines are out of scope.

## Dependency and review scope

**Depends on #6060; merge that PR first.** This branch includes the prerequisite and now also carries the reviewed #6060 changes: explicit SCF convergence status, consistent ASE units, the ASE comparison, packaging documentation and CI coverage. The shared changes were cherry-picked without rewriting the existing integration history.

Until #6060 is merged, GitHub's diff against `master` also includes that prerequisite. The current integration-only source comparison is:

https://github.com/DCM-Uni-Paderborn/cp2k/compare/python-library-interface..integration-hardening

After #6060 lands, the follow-up commits can be rebased onto `master` without folding the prerequisite into this PR.

### C API, Python and ASE

- Add `cp2k_get_stress_tensor` with an explicit availability flag, and expose potential stress/virial in Python.
- Preserve energy/force-only compatibility with older shared libraries; reject unavailable stress instead of silently returning zero.
- Convert the native pressure-positive tensor to ASE's tensile-positive Voigt convention and support ASE cell optimization.

### PLUMED

- Pass the physical forces and potential virial into PLUMED so `ENERGY`-dependent biases can rescale them correctly.
- Include bias energy in MD bookkeeping and feed the returned virial back to CP2K, with sign conversion and MPI replication.
- Declare retained Fortran/C buffers correctly; include the initial biased forces before the first half-step.
- Supply the MD target temperature as `kBT`, propagate restart state, and exercise continuous versus restarted well-tempered metadynamics.
- Fix the CMake library-discovery fallback when no pkg-config result or installed headers are available.

### OpenMM and LAMMPS

- Add an optional OpenMM >= 8.6.1 `PythonForce` adapter using a caller-owned CP2K environment.
- Add a LAMMPS `fix external` adapter for `metal`/`real` units, atom-tag mapping, triclinic boxes, MPI and variable-cell/NPT runs.
- Provide global energy/virial contributions with LAMMPS's expected MPI normalization and propagate callback validation failures collectively.
- Document units, ownership, restart handling, MPI requirements and limitations, with an inexpensive argon example.

### Independent MPI and explicit QM/MM

- Add `SocketEnvironment` and `python -m cp2k.server`: persistent force environments/SCF state with authenticated, bounded JSON messages, timeouts and collective error propagation. The direct in-process API remains available.
- OpenMM can now use a **multi-rank CP2K server** without running its callbacks collectively. Reference and CPU platforms are tested with a two-rank CP2K server.
- LAMMPS and CP2K can run as independent jobs with different MPI implementations and rank counts; native libraries are no longer loaded in the same process on this route. Test two LAMMPS ranks against a separate one-rank CP2K server, plus real/metal pressure/NPT through the socket.
- Add `SubtractiveQMMM` for an explicitly selected whole-molecule region: `E_MM(all) + E_QM(region) - E_MM(region)`, with mapped force and virial corrections. The caller supplies a matching region MM reference; the wrapper performs subtraction automatically.
- Test the model algebra by Cartesian and cell-strain finite differences, and run a real H2 DFT/OpenMM replacement test with independent CP2K QM and MM servers.
- Document native CP2K `METHOD QMMM` as the alternative where CP2K owns the complete embedding/link treatment and potential.
- Extend the existing Linux/x86-64 ASE CI job to run the Python/shared-library suite, OpenMM Reference/CPU, serial LAMMPS, socket and subtractive-QM/MM tests, separately from CMake. This is additional CI coverage, not a claim of completed Linux results before the job runs.

### i-PI

- Complete fragmented socket reads/writes, retry interrupted system calls, detect premature EOF and guard UNIX socket path lengths.
- Validate particle counts, cell/position finiteness and protocol state before using received data.
- Handle `EXIT` explicitly and close the socket on normal shutdown.

### CI convention corrections

- Resolve all ten unsuppressed findings in the completed Conventions run without adding suppressions.
- Inspect the BFGS restart probe status before rewinding; declare the complex Schur LAPACK interface explicitly.
- Pass runtime output units to the standalone topology diagnostics and route temporary link storage through `cp_files`.
- Preserve anonymous scratch-file deletion and add formatted/random-access storage regression checks.
- Local compiler AST/warning checks report zero unsuppressed findings for the affected files. Five native unit programs, two-rank MPI checks, legacy/tagged BFGS restarts and C1/C2 command-line fixtures pass.
- The earlier 92-test integration suite passed against the existing native build. Updated validation after the #6060 review changes is reported below; fresh remote CI remains pending.

## Review follow-up

- Propagate the explicit Quickstep SCF status through the socket backend, including one- and two-rank server tests. Unknown/unsupported status remains `None`, not an assertion of convergence.
- Direct calculations reject reported nonconvergence by default with `SCFConvergenceError`; ASE translates this to `CalculationFailed`. The socket retains collective error propagation to the client. Diagnostic `check_convergence=False` explicitly returns the unconverged result and status.
- Invalidate energy, force and stress availability before a failed calculation and when the geometry changes. A status for a component calculation does not certify an entire composite QM/MM model.
- Use CP2K's unit constants consistently in both ASE routes. The common benchmark finds close numerical agreement, but no consistent steady-state speed advantage over ASE's existing persistent shell.

## Validation

Latest local macOS/ARM64 validation with GCC/gfortran 16, OpenMPI 5, Python 3.12, OpenMM 8.6.1, ASE 3.29, LAMMPS stable_22Jul2025_update4 and i-PI 3.3.0:

- **106 passed, 3 skipped** against a freshly rebuilt integration native library. This covers direct and socket interfaces, SCF status and failures, analytical stress, OpenMM Reference/CPU, LAMMPS, independent MPI jobs, subtractive QM/MM and i-PI. The three skips are PLUMED tests because this fresh build does not enable PLUMED.
- **62 passed, 47 skipped** without opt-in native integrations.
- Separately, **all 3 PLUMED tests passed** using the current Python code with the pre-existing PLUMED-enabled native library. This is a compatibility check, not a claim that the new native library was built with PLUMED.
- Repository formatting/precommit and whitespace checks passed for the update.

Earlier integration validation also covered the C `libcp2k_unittest`, six-component finite-strain stress checks, ASE cell optimization, LAMMPS pressure/NPT in both unit systems, PLUMED analytical bias checks and continuous/restarted HILLS histories, the existing PLUMED water energy, and one-/two-rank variable-cell trajectories.

The Python tests are run explicitly, separately from the top-level CMake build, as requested in #6060. The README documents their opt-in dependencies and environment variables. These are local results; fresh remote CI on the updated head remains pending. GPU, native Windows and all-method coverage are not claimed.

## Scientific choices and remaining operational limits

- QM region, charge, spin and a consistent MM reference are explicit scientific inputs. Subtractive QM/MM uses mechanical embedding and whole-molecule fixed regions; it does not infer covalent cuts, introduce link atoms or polarize QM with host point charges. Native CP2K QM/MM provides its own supported embedding/link machinery.
- OpenMM callbacks remain main-thread and must not share a live environment concurrently. CP2K rank count is independent when using the socket backend. Live callbacks are not XML-serializable.
- The **direct** LAMMPS/CP2K route still needs matching MPI libraries and congruent communicators; the **socket** route separates the jobs. Each job's own native library/mpi4py pair must be compatible. Variable-cell runs need stress from every contributing model.
- The socket authenticates but does not encrypt: use loopback or a trusted SSH tunnel. One server serves one client; its scheduler/application owns the MPI job. A native server abort becomes a client exception, whereas native aborts in the direct route can still terminate Python.
- Native engine restarts and callbacks must be restored explicitly; CP2K restart files do not contain PLUMED's HILLS history. Local evidence is macOS/ARM64; the expanded Linux ASE job must finish before claiming that additional platform was validated. GPU and native Windows CP2K coverage is not claimed.

## Shared release preparation

Synchronized the upload-free release validation workflow and proposed release/ownership policy from #6060. These documentation/packaging changes do not publish a package or change native behavior. The supported native revision and included optional integrations must be recorded for the eventual release after merge and approval.


DCM-Uni-Paderborn:
Fixed the ten unsuppressed convention findings from the completed Conventions run in
809ddd4bdf9acca7b9a16b17f926daa225e9a54b. The branch is rebased onto current master;
its history is linear, and its source tree is identical to the locally tested tree.
No convention suppressions were added.

- Inspect the BFGS magic-probe I/O status before rewinding, preserving legacy restart handling.
- Use runtime diagnostic units, including caller-supplied units for phason analysis.
- Declare the LAPACK complex Schur interface explicitly.
- Route anonymous scratch storage through `cp_files`, retaining automatic deletion and
  independent units, with formatted and random-access scratch regression coverage.

Local verification: the compiler AST/warning checks report zero unsuppressed findings for
the affected files; all nine changed files pass precommit. Five native unit programs pass,
including Wilson/curvature/snapshot checks; k-point symmetry and covariant preconditioning
also pass with two MPI ranks. Tagged direct/inverse and short legacy BFGS restart checks pass.
Phason CLI checks retain C1 = -1 on both coarse/refined meshes and C2 = +1 for a product bundle.
The Python/integration suite passes all 92 tests using the existing native build.

The local checks use macOS/ARM64 and GNU Fortran 16; the fresh remote CI result is still pending.
The separate AiiDA and final optimization-cell-frame changes are not included in this update.

DCM-Uni-Paderborn:
Dear @fstein93, @mtaillefumier and @RMeli,

Could you please review the additional stress and MD integrations when convenient, particularly the native PLUMED/i-PI changes and the OpenMM/LAMMPS adapters? The current regular CI and conventions checks are green, but this PR has not yet received a review.

The intended merge order remains #6078, then #6060, then this follow-up. The follow-up-only comparison is linked in the PR description. Please let me know if any integration should be revised or split further before merge.

Many thanks!
Greetings,
Thomas

src/topology_curvature.F
```diff
@@ -18,6 +18,24 @@ MODULE topology_curvature
    PRIVATE
    REAL(KIND=dp), PARAMETER :: pi = 3.1415926535897932384626433832795_dp, unitary_tol = 1.e-8_dp
    PUBLIC :: polar_link, link_plaquette, curvature_density
+   INTERFACE
```
fstein93: This should be removed (similar to #6078 )

fstein93 (dismissed):
I didn't have a look at the Python and Docker files.

src/topology_curvature.F
DCM-Uni-Paderborn: Dear Frederick,

Agreed. The explicit ZGEES interface was removed in c90d9bfbb8 in the native prerequisite #6078, and the existing LAPACK-name regex was extended to recognize GEES instead. This follows CP2K's established convention without adding a warning suppression.

The current #6061 head still contains the older prerequisite version, so your comment is valid here: that fix has not yet been propagated to this follow-up. The branches need to be synchronized before this point can be considered resolved. No change to the numerical implementation, references or tolerances is needed for it.

Greetings,
Thomas

DCM-Uni-Paderborn: Many thanks, Frederick. Fixed in the pushed ed87aad325: this branch is now rebased onto master containing the merged #6078. topology_curvature.F and the LAPACK warning analyzer are identical to master, so the explicit ZGEES interface is no longer part of this PR. The fresh build, C tests with one/two ranks and communicator smoke tests pass; the Python/integration run has 106 passes and three explicit PLUMED skips in the local non-PLUMED build. Could you please re-review this point?

DCM-Uni-Paderborn:
Review follow-up pushed in ed87aad325, on master with #6078 merged. Frederick's LAPACK-interface point is fixed and a fresh review requested. The #6060 CI/documentation changes and direct is_dirty cleanup are synchronized: the dedicated Python tester owns the package/OpenMM/LAMMPS tests, while the upstream ASE tester and generic third-party generator remain unchanged. The unrelated cached-report changes are removed.

The updated description distinguishes fresh local validation (106 tests passed, three PLUMED tests skipped in the non-PLUMED build) from earlier results and the pending Linux runs. No numerical references or tolerances changed. #6060 must still merge first.

/cp2kci python ase i-pi manual conventions

python/RELEASING.md
```diff
@@ -0,0 +1,84 @@
...
+
+Run this check for each supported Python version. Repeat from the actual merged release commit;
+artifacts made from a PR head are validation artifacts, not an approved release. This check does not
+replace the native tests or make a timing benchmark a release threshold.
+
+## After merge and approval
+
+1. Confirm the exact release commit, version, release notes and green CI. Document the supported
+   native installation route and compatible CP2K revision/version, not just `pip install`.
+1. Build once with the workflow above. Test the installed wheel with the matching native library:
```
fstein93, line 65: `2.`, I guess (similarly below)
Frederick (not Fabian)

DCM-Uni-Paderborn:
Implemented Ole's public-API typing suggestion from #6060 in 2844d1b9b1, together with the dashboard hostname correction. The strict check covers the public package entry points and ASE calculator, including the socket client and imported helpers; it does not claim that the optional OpenMM/LAMMPS/QM-MM adapters or server CLI are fully typed. It runs in the existing Python tester without adding native regression calculations. py.typed and positive/negative static contract checks are included.

Validation: source and installed-wheel mypy --strict pass; the unchanged fully enabled local integration suite has 106 passes and three PLUMED-only skips; all three 2-rank MPI smoke modes pass; sdist/wheel build, isolated installed-package tests and precommit pass. No numerical tolerances or reference values were changed. The already-addressed ZGEES review request remains fixed; no additional native code was changed in this update. #6060 is still the prerequisite.

/cp2kci python ase i-pi manual conventions

python/RELEASING.md
DCM-Uni-Paderborn: Dear Frederick,

Many thanks, and apologies for getting your name wrong in my earlier replies.

The intended sequence is indeed 1, 2, 3, 4. I checked both the repository formatter and GitHub's rendered Markdown: CP2K's mdformat normalizes explicitly numbered source items back to `1.`, while GitHub renders them as a single correctly numbered ordered list. I have therefore kept the formatter-compatible source form rather than introducing a formatting exception. The document is unchanged from the now-merged #6060.

The branch is being rebased onto that merged base; the ZGEES correction remains included, and the previously approved native prerequisite is no longer repeated in the PR diff.

Greetings,
Thomas

DCM-Uni-Paderborn:
Rebased onto master after #6060 merged, and pushed 9b0987b2e5 with an exact lease. The resulting file tree was compared against the expected three-way combination: no integration or typing changes were lost, and no numerical references, tolerances or tests were changed. The ordinary PR diff now excludes the merged prerequisite. Strict mypy and the existing local integration suite pass again (106 passed, the same three PLUMED-only skips). The release-list comment has also been answered after checking the formatter and GitHub rendering. Fresh CI is requested for the rebased head; this PR remains open for review.

/cp2kci python ase i-pi manual conventions

DCM-Uni-Paderborn:
Pushed 4b62868234 to address the latest #6060 naming discussion and prevent conflicts with #6062. Distribution and import names remain provisional pending the existing pycp2k maintainer's agreement. The pytest marker and mypy table layout now allow both follow-ups to combine cleanly in either order, without changing type-checking settings or adding tests.

Strict mypy and the existing local integration suite pass again: 106 passed, the same three PLUMED-only skips with the existing non-PLUMED native build. Both branches also merge cleanly with current master. No native calculation code, numerical references or tolerances changed. The addressed ZGEES and Markdown-numbering points remain ready for Frederick's re-review; existing review requests are retained.

Fresh CI for this head:

/cp2kci python ase i-pi manual conventions

DCM-Uni-Paderborn:
The Conventions job was PREEMPTED while building the toolchain (OpenBLAS), before the source checks. The full numerical regression run passed all 6306 tests. Requesting one targeted unchanged-head retry of Conventions; no code, reference values or tolerances are changed.

/cp2kci conventions

oschuett:
I  think this should be at least four PRs:

- Add stress tensor to libcp2k
- Fix i-Pi
- Fix Plumed
- Add Python examples

These each deserve proper review by different people. And if anything goes wrong then we need to be able to roll back the changes individually.

DCM-Uni-Paderborn:
Dear Ole,

Many thanks for the suggestion. I have split the changes along the boundaries you proposed:

- #6061 now contains only the native libcp2k stress API and its C-interface test (four C/Fortran files).
- #6105 contains the i-PI protocol fixes and their tests.
- #6106 contains the PLUMED coupling fixes and their tests.
- #6107 contains the Python wrappers, adapters and examples, with public-API typing in a separate commit.

The native PRs are based independently on master. I removed the tests' hidden dependency on the proposed stress getter: i-PI uses the existing force API and a pair-virial reference; PLUMED compares native executable output. #6107 is explicitly a draft pending #6061 because its stress-dependent features need that C API.

Each native branch was built and tested separately. A fresh build of their combined state also passes all 109 tests without skips. The combined production code matches the original PR plus current master; no numerical references or tolerances were changed. Fresh Linux CI is being requested for the split native PRs, so these are local results, not a claim that their new CI has already passed.

The changes can now be reviewed and reverted by topic. No PR has been merged and the existing review requests remain in place.

Greetings,
Thomas

oschuett (approved):

PR #6063, opened 2026-09-19
Unify and optimize k-point parallelism across MPI and OpenMP

Growl1234:
Consolidate k-point parallel execution into shared operator assembly, matrix transfer, and workspace management, reducing duplication while preserving performance gains.

~This is based on #5993, that's why all changes there are included.~


Growl1234:
OT's own OpenMP performance needs to be taken care of seperatedly.

src/parallel_gemm_api.F
```diff
@@ -73,14 +74,37 @@ SUBROUTINE parallel_gemm_fm(transa, transb, m, n, k, alpha, matrix_a, matrix_b,
...
 
-      INTEGER                                            :: cfc, cfr, handle, my_multi
+      INTEGER                                            :: cfc, cfr, handle, my_multi, origin(6)
+
+      ! Local execution is an explicit caller contract: no MPI or backend dispatch.
 
       MARK_USED(cfc)
       MARK_USED(cfr)
 
+      IF (PRESENT(local)) THEN
```
fstein93: I am not in favor of adding this option to `parallel_gemm_api` for various reasons. This module serves as a wrapper to switch between the different available PDGEMM-backends (COSMA, Scalapack) . What you want to achieve is a local `?GEMM` for a very specific case which is generally not true for callers of `parallel_gemm`. I would be in favor of having a separate implementation to make this `contract` explicit, to consider at an actual different implementation.

fstein93 (changes requested):

src/parallel_gemm_api.F
fstein93: If GPU-acceleration is of interest, `local_gemm_api` should be of interest to you.

Growl1234: How about using `cp_cfm_gemm_local` in `cp_cfm_basic_linalg.F` and `cp_fm_gemm_local` in `cp_fm_basic_linalg.F` instead?

fstein93: This still doesn't sound right to me. `cp_fm_*`-routines are operations on dense distributed matrices working on all elements and wrap around the respective operations from (Sca)Lapack and (P)BLAS. As such, they are usually collective operations and work on the entire matrix in the general, distributed case. This is in contrast to what you want to achieve, an optimization in case of fully local matrices. This is definitely not the general use-case of our Scalapack wrappers but of separate modules such as `local_gemm_api` and `mathlib`. If an implementation in terms of local (BLAS or Lapack) primitives leads to performance benefits over using our general wrapper, you should implement exactly these. `local_gemm_api` may even introduce GPU-acceleration.

Growl1234: OK, I'll try `local_gemm_api`. Btw, could you please have a look again at #5993, so that I would be able to rebase this one and then it's easier to handle and review.

Growl1234: 6203155 now turns to use `local_gemm_api`.

fstein93 (approved):

PR #6066, opened 2026-09-20
Docs: fix linebreaking in equations for new MathJax version

he-zilong:
As a follow-up to #5995.


Growl1234:
Is " $\\[1.2em]$ " here what you want? Here is what I saw on GitHub:

图片

he-zilong:
> Is " \[ 1.2 e m ] " here what you want? Here is what I saw on GitHub:

This looks more like an issue for the github markdown math renderer. It cannot render `\left( \begin{array}{cc}A &  B\\B &  A\end{array} \right)` as a 2-row array either.

That said, the original problem has yet to be solved...

Edit: I am a little surprised that the `amsmath` in MyST extensions has not been enabled yet.

he-zilong:
It seems that using `\\` in a `$$..$$` environment will have a `\begin{split}..\end{split}` automatically added regardless of the existing environment. The MyST documentation on [Dollar delimited math](https://myst-parser.readthedocs.io/en/v5.1.0/syntax/optional.html#dollar-delimited-math) contains an example under the line "Block-level math can be specified with $$ signs that wrap the math block you’d like to parse" which shows this behavior. Note that without a `\begin{split}..\end{split}` environment, the `&` character for tab/aligning may not be parsed correctly by other markdown viewers.

The `\begin{align}..\end{align}` environments appears to be handled separately by the `amsmath` MyST extension. When I actually try this style, precommit formatter thinks there are many characters to escape. Edit: ...which unfortunately means that it is not available at this time. I will revert to the `dollarmath` style. It is unclear at this point if the [`mdformat-myst`](https://github.com/executablebooks/mdformat-myst) plugin (which we have not used) supports `amsmath` that does not excessively escape characters yet.

oschuett (approved):
Thanks a lot for getting to the bottom of this!

Btw, we are already using the [mdformat-myst](https://github.com/cp2k/cp2k/blob/3bf7df10f036877a66c32332c19873f9ec33d285/tools/precommit/requirements.txt#L16) plugin.

he-zilong:
> Thanks a lot for getting to the bottom of this!
> 
> Btw, we are already using the [mdformat-myst](https://github.com/cp2k/cp2k/blob/3bf7df10f036877a66c32332c19873f9ec33d285/tools/precommit/requirements.txt#L16) plugin.

Yeah, I must have looked at a wrong `requirements.txt` file. Anything else to do for this PR? I think enabling the `amsmath` extension and confirming precommit support for the syntax is better left for the future.

